PR #78, opened 2019-07-24
Drop me down and pop me over these actions

darryltec:
## Changes

- reorder button props to sort alphabetically

### Added

- Added arias on button
- added menu component

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Menu/Menu.mdx
+++ b/packages/components/src/Menu/Menu.mdx
@@ -6,7 +6,7 @@ route: /components/menu
 
 import { Playground, Props } from "docz";
 import { ComponentStatus } from "@jobber/docx";
-import { Menu, SectionProps } from ".";
+import { Menu } from ".";
 
 # Menu
 
--- a/packages/components/src/Menu/Menu.mdx
+++ b/packages/components/src/Menu/Menu.mdx
@@ -18,12 +18,18 @@ In Jobber a Menu is used to …
 
 <Playground>
   <Menu
-    actions={[
+    items={[
       {
-        header: "File options",
-        items: [
-          { label: "Import file", icon: ImportMinor },
-          { label: "Export file", icon: ExportMinor }
+        actions: [
+          { label: "Import file", icon: "download" },
+          { label: "Export file", icon: "upload" }
+        ]
+      },
+      {
+        header: "Mark as...",
+        actions: [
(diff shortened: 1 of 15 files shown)


seanhealy:
Are these supposed to be grey?

image

packages/components/src/Menu/Menu.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { Menu } from ".";
+
+# Menu
+
+<ComponentStatus stage="pre" responsive="no" accessible="yes" />
+
+A toggleable menu that holds a list of actions.
```
seanhealy: It might be nice to add a little line here as to when this should be used vs a select.

packages/components/src/Menu/Menu.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { Button } from "../Button";
+import { Typography } from "../Typography";
+import { Icon, IconNames } from "../Icon";
+import styles from "./Menu.css";
+
+interface MenuProps {
+  /**
+   * Collection of action items.
+   */
+  readonly items: SectionProps[];
```
seanhealy: I'm wondering if this should be an array or, if we should just use `children`…

packages/components/src/Menu/Menu.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    {
+      header: header,
+      actions: [{ label: actionLabel, onClick: clickHandler }],
+    },
+  ];
+
+  const { getByRole, queryByTestId } = render(<Menu items={actions} />);
+
+  fireEvent.click(getByRole("button"));
+  expect(queryByTestId("menu-popup")).toBeTruthy();
```
JordanLos: I think if we use `queryByRole("More")` instead of `queryByTestId("menu-popup")` we can remove the `data-testid` on the menu div.

packages/components/src/Menu/Menu.tsx
darryltec: 🤔 Children allows people to put non menu things inside the menu. What's your concern?

PR #81, opened 2019-08-02
Create input number

darryltec:
## Changes

- Make form field return a number if type is number
- Make value return as number instead of string

### Added

- Input number component

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -77,13 +77,13 @@ export interface FormFieldProps {
   /**
    * Set the component to the given value.
    */
-  readonly value?: string;
+  readonly value?: string | number;
 
   /**
    * Simplified onChange handler that only provides the new value.
    * @param newValue
    */
-  onChange?(newValue: string): void;
+  onChange?(newValue: string | number): void;
 }
 
 export function FormField({
@@ -113,8 +113,13 @@ export function FormField({
       | ChangeEvent<HTMLTextAreaElement>
       | ChangeEvent<HTMLSelectElement>,
   ) => {
-    const newValue = event.currentTarget.value;
+    let newValue: string | number;
+    newValue = event.currentTarget.value;
     setHasMiniLabel(newValue.length > 0);
+
+    if (type === "number" && newValue.length > 0) {
+      newValue = parseFloat(newValue);
+    }
     onChange && onChange(newValue);
   };
 
--- a/packages/components/src/InputNumber/InputNumber.mdx
+++ b/packages/components/src/InputNumber/InputNumber.mdx
@@ -7,13 +7,14 @@ route: /components/input-number
 import { Playground, Props } from "docz";
 import { ComponentStatus } from "@jobber/docx";
 import { InputNumber } from ".";
+import { Text } from "../Text";
 import { useState } from "react";
 
 # InputNumber
 
 <ComponentStatus stage="pre" responsive="no" accessible="no" />
 
-In Jobber a InputNumber is used to …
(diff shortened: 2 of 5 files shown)


packages/components/src/InputNumber/InputNumber.mdx
```diff
@@ -0,0 +1,94 @@
...
+    invalid={true}
+    name="wholeNumber"
+  />
+</Playground>
+
+## Sizes
+
+<Playground>
+  <InputNumber name="smallField" size="small" placeholder="Duration" />
+  <InputNumber name="largeField" size="large" placeholder="Your username" />
```
kkimble006: Should this be a different placeholder? 'Your username' is confusing

packages/components/src/InputNumber/InputNumber.mdx
```diff
@@ -0,0 +1,94 @@
...
+      name="durationInDays"
+      value={2}
+      size="small"
+      inline={true}
+      maxLength={4}
+      align="center"
+    />
+    days
+  </Text>
+</Playground>
```
kkimble006: For all the other fields up and down work, here the up and down arrows do nothing. I am guessing because it is wrapped in text?

darryltec: oh, no because it's using `value` instead of `defaultValue`... will explain in person

packages/components/src/InputNumber/InputNumber.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+## Properties
+
+<Props of={InputNumber} />
```
kkimble006: For this, it always puts undefined as an acceptable value. Is there a way we can set a default to ensure this this is the expected type? Or are we okay with undefined?

kkimble006:
@darryltec lookin good, just a couple of questions.

packages/components/src/FormField/FormField.tsx
```diff
@@ -77,13 +77,13 @@ export interface FormFieldProps {
...
    */
-  readonly value?: string;
+  readonly value?: string | number;
 
   /**
    * Simplified onChange handler that only provides the new value.
    * @param newValue
    */
-  onChange?(newValue: string): void;
+  onChange?(newValue: string | number): void;
```
seanhealy: Should this always be a number?

packages/components/src/InputNumber/InputNumber.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+afterEach(cleanup);
+
+it("renders an input type number", () => {
+  const tree = renderer.create(<InputNumber value="123" />).toJSON();
+  expect(tree).toMatchInlineSnapshot(`
+    <div
+      className="wrapper"
+      style={
+        Object {
+          "--formField-maxLength": undefined,
```
seanhealy: Any idea how this is getting set to undefined?

darryltec: because it ain't specified. I think the tests still renders undefined stuffs

packages/components/src/InputNumber/InputNumber.mdx
darryltec: Talked in person, we're okay with undefined because we has no control over it... yet

packages/components/src/FormField/FormField.tsx
seanhealy: Can we restrict these appropriately on text and number to string and number respectively? (If it's easy?)

PR #84, opened 2019-08-07
Tweaks to Card and Text component

gregb-jobber:
## Changes

CSS change and added a type to `onClick()`.

### Added

- Added a param for the onClick event to the Card. I needed to do a preventDefault on the onClick of the Card I'm working with, but TypeScript complains because the types don't match. ( `() => void` vs. `(e: MouseEvent<HTMLElement>) => void`);

### Changed

- `width: 100%` needed to make cards all the same width when they have different text while in a flex row/column.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Card/Card.css
+++ b/packages/components/src/Card/Card.css
@@ -9,6 +9,7 @@
 
 .card {
   display: block;
+  width: 100%;
   padding: var(--card--base-padding);
   border: var(--border-base) solid var(--card--border-color);
   border-radius: var(--radius-base);
--- a/packages/components/src/Card/Card.tsx
+++ b/packages/components/src/Card/Card.tsx
@@ -51,7 +51,7 @@ interface LinkCardProps extends CardProps {
 }
 
 interface ClickableCardProps extends CardProps {
-  onClick(e: React.MouseEvent<HTMLElement>): void;
+  onClick(event: React.MouseEvent<HTMLElement>): void;
 }
 
 type CardPropOptions = XOR<CardProps, XOR<LinkCardProps, ClickableCardProps>>;
--- a/packages/components/src/Card/Card.tsx
+++ b/packages/components/src/Card/Card.tsx
@@ -51,7 +51,7 @@ interface LinkCardProps extends CardProps {
 }
 
 interface ClickableCardProps extends CardProps {
-  onClick(): void;
+  onClick(e: React.MouseEvent<HTMLElement>): void;
 }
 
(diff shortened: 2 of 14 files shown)


packages/components/src/Card/Card.tsx
```diff
@@ -51,7 +51,7 @@ interface LinkCardProps extends CardProps {
 }
 
 interface ClickableCardProps extends CardProps {
-  onClick(): void;
+  onClick(e: React.MouseEvent<HTMLElement>): void;
```
darryltec: Still can't find a direct answer to how this mouse event _reacts_ on a mobile environment where there's no mouse

gregb-jobber: Is there a way to test it? I don't know much about mobile.

darryltec: I think running atlantis on your phone should do the trick

darryltec: works fine, soooo... 🤷‍♂

packages/components/src/Card/Card.css
```diff
@@ -9,6 +9,7 @@
 
 .card {
   display: block;
+  width: 100%;
```
darryltec: interesting.. the `display: block` didn't add a width of 100%?

darryltec: Correction: The display block didn't take up the full width of the parent?

packages/components/src/Card/Card.tsx
```diff
@@ -70,7 +70,7 @@ export function Card({ accent, children, url, onClick }: CardPropOptions) {
     href?: string;
     role?: "button";
     tabIndex?: 0;
-    onClick?(): void;
+    onClick?(e: React.MouseEvent<HTMLElement>): void;
```
darryltec: thoughts on naming this `event` instead of `e`? I know `e` is also used to convey event but it could also mean a lot of other things. Remove the FUD even before it happens

> The best icon to represent edit is to put the word edit on it

packages/components/src/Card/Card.css
darryltec: talked in person. layout is a pain

packages/components/src/Card/Card.tsx
darryltec: @gregb-jobber Also, you forgot to change this to event

PR #87, opened 2019-08-08
Time formatter

kkimble006:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/TimeFormatter/TimeFormatter.mdx
+++ b/packages/components/src/TimeFormatter/TimeFormatter.mdx
@@ -1,0 +1,33 @@
+---
+name: TimeFormatter
+menu: Components
+route: /components/time-formatter
+---
+
+import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { TimeFormatter } from ".";
+
+# TimeFormatter
+
+<ComponentStatus stage="pre" responsive="no" accessible="no" />
+
+In Jobber a TimeFormatter is used to …
+
+## Simple Example
+
+<Playground>
+  <TimeFormatter civilTime={"11:30:00.000000000"} useMilitaryTime={false} />
+  <br />
+  <TimeFormatter civilTime={"17:45:00.000000000"} useMilitaryTime={false} />
+</Playground>
+
+## 24 hour Example
+
+<Playground>
+  <TimeFormatter civilTime={"17:45:00.000000000"} useMilitaryTime={true} />
+</Playground>
+
+## TimeFormatter Properties
+
+<Props of={TimeFormatter} />
--- a/packages/components/src/TimeFormatter/TimeFormatter.test.tsx
+++ b/packages/components/src/TimeFormatter/TimeFormatter.test.tsx
@@ -14,5 +14,5 @@ it("renders a TimeFormatter", () => {
       />,
     )
     .toJSON();
-  expect(tree).toMatchInlineSnapshot();
+  expect(tree).toMatchInlineSnapshot(`"11:30 AM"`);
 });
--- a/packages/components/src/TimeFormatter/TimeFormatter.test.tsx
+++ b/packages/components/src/TimeFormatter/TimeFormatter.test.tsx
@@ -1,0 +1,18 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import { cleanup } from "@testing-library/react";
+import { TimeFormatter } from ".";
+
+afterEach(cleanup);
+
+it("renders a TimeFormatter", () => {
+  const tree = renderer
+    .create(
+      <TimeFormatter
+        civilTime={"11:30:00.000000000"}
+        useMilitaryTime={false}
+      />,
+    )
+    .toJSON();
+  expect(tree).toMatchInlineSnapshot();
+});
--- a/packages/components/src/TimeFormatter/TimeFormatter.tsx
+++ b/packages/components/src/TimeFormatter/TimeFormatter.tsx
@@ -31,7 +31,7 @@ function getAmPmLabel(hour: number, useMilitaryTime: boolean) {
     return "";
   }
 
-  return hour > 12 ? "PM" : "AM";
+  return hour >= 12 ? "PM" : "AM";
 }
 
 function setCorrectHourFormat(hour: number, useMilitaryTime: boolean) {
--- a/packages/components/src/TimeFormatter/TimeFormatter.tsx
+++ b/packages/components/src/TimeFormatter/TimeFormatter.tsx
@@ -1,0 +1,43 @@
+import React from "react";
+
+interface TimeFormatterProps {
+  /**
+   * Civil Time string to be formatted
+   * 17:45:00.000000000
+   * 11:30:00.000000000
+   */
+  readonly civilTime: string;
+
+  /**
+   * Decide to show 24 hour vs.
+   * 12 hour display
+   */
+  readonly useMilitaryTime: boolean;
+}
+
+export function TimeFormatter(props: TimeFormatterProps) {
+  const timeSplit = props.civilTime.split(":");
+  const hour = parseInt(timeSplit[0]);
+  const minute = parseInt(timeSplit[1]);
+
+  const amOrPm = getAmPmLabel(hour, props.useMilitaryTime);
+  const formattedHour = setCorrectHourFormat(hour, props.useMilitaryTime);
+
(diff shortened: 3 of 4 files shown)


packages/components/src/TimeFormatter/TimeFormatter.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+interface TimeFormatterProps {
+  /**
+   * Civil Time string to be formatted
+   * 17:45:00.000000000
+   * 11:30:00.000000000
+   */
+  readonly civilTime: string;
```
JordanLos: @kkimble006 Maybe not directly related to the work in your PR. I'm just curious why we're using `readonly` for props? I think I've seen it in other places too, but I'm pretty sure typescript makes props immutable (i.e. readonly) by default. 

Just to try it out, you can remove the `readonly` keyword, attempt to change the value using something like `props.civilTime = 'someOtherString'` and it will return a `TypeError`.

I'm trying to find some documentation that _guarantees_  props are readonly to confirm. Some starters:
https://reactjs.org/docs/components-and-props.html#props-are-read-only

image

@seanhealy @darryltec  Maybe you could also add some insight here?

packages/components/src/TimeFormatter/TimeFormatter.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export function TimeFormatter(props: TimeFormatterProps) {
+  const timeSplit = props.civilTime.split(":");
+  const hour = parseInt(timeSplit[0]);
+  const minute = parseInt(timeSplit[1]);
+
+  const amOrPm = getAmPmLabel(hour, props.useMilitaryTime);
+  const formattedHour = setCorrectHourFormat(hour, props.useMilitaryTime);
+
+  return `${formattedHour}:${minute} ${amOrPm}`;
```
JordanLos: @kkimble006 Do you think there's any value in error checking any of these values (e.g. hours not between 0-23 and minutes/seconds not between 0-59)? Or should we just trust that whoever is consuming this component does so responsibly?

packages/components/src/TimeFormatter/TimeFormatter.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+function getAmPmLabel(hour: number, useMilitaryTime: boolean) {
+  if (useMilitaryTime === true) {
+    return "";
+  }
+
+  return hour >= 12 ? "PM" : "AM";
+}
+
+function setCorrectHourFormat(hour: number, useMilitaryTime: boolean) {
```
JordanLos: @kkimble006 I noticed an edge case here: If you use `useMilitaryTime={false}` but set the hour to `0`, which is a valid hour in military time, it should return `12:xx AM`, but instead return `0:xx AM`, which is not a valid civilian time. 
image

packages/components/src/TimeFormatter/TimeFormatter.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { TimeFormatter } from ".";
+
+# TimeFormatter
+
+<ComponentStatus stage="pre" responsive="no" accessible="no" />
+
+In Jobber a TimeFormatter is used to …
```
JordanLos: @kkimble006 Did you want to leave a quick description of the component here?

packages/components/src/TimeFormatter/TimeFormatter.test.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import { cleanup } from "@testing-library/react";
+import { TimeFormatter } from ".";
+
+afterEach(cleanup);
+
+it("renders a TimeFormatter", () => {
```
JordanLos: @kkimble006 Time and dates often have edge cases and are prone to simple mistakes, which mean they could benefit from unit tests. What do you think about adding some simpler ones now? I could think of:
1) Military Time displays properly
2) Civilian time displays properly
3) `12` AM displays properly in civilian and military.

packages/components/src/TimeFormatter/TimeFormatter.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+## Simple Example
+
+<Playground>
+  <TimeFormatter civilTime={"11:30:00.000000000"} useMilitaryTime={false} />
```
seanhealy: I'm not sure that `MilitaryTime` is a broad name to use here.

seanhealy: Perhaps `useAnteMeridiem` or use `use12HourClock`? I feel like there is a more conventional name here I'm just not thinking of it…

packages/components/src/TimeFormatter/TimeFormatter.tsx
seanhealy: The readonly here is telling typescript that the property is readonly. This allows the compiler to find this error for us instead of relying on the codepath being executed and reported at runtime.

image

vs

image

It is nice to know that react is double checking this for us though too. :)

packages/components/src/TimeFormatter/TimeFormatter.tsx
seanhealy: I feel like since this component is accepting a CivilTime we should rely on it to ensure that it is a valid time…

seanhealy: We are probably also best using the browsers built in locale formatters here. They are going to deal with a lot more edge cases so we don't have to. :)

eg
```ts
(new Date('August 19, 1975 23:15:00 GMT+00:00')).toLocaleTimeString(undefined, {timeStyle: "short"})
```

packages/components/src/TimeFormatter/TimeFormatter.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Decide to show 24 hour vs.
+   * 12 hour display
+   */
+  readonly useMilitaryTime: boolean;
+}
+
+export function TimeFormatter(props: TimeFormatterProps) {
+  const timeSplit = props.civilTime.split(":");
```
seanhealy: civilTime here should be an instance of `CivilTime`. See:

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/InputTime/InputTime.tsx#L2

https://github.com/tc39/proposal-temporal/blob/master/objects.md#civiltime-

packages/components/src/TimeFormatter/TimeFormatter.tsx
kkimble006: Good Catch!

packages/components/src/TimeFormatter/TimeFormatter.test.tsx
kkimble006: We are not using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString to set the time, based on a CivilTime object. I don't think we need to test this.

packages/components/src/FormatTime/FormatTime.mdx
```diff
@@ -0,0 +1,45 @@
...
+  <FormatTime time={new CivilTime(18, 35)} use24HourClock={true} />
+  <br />
+  <FormatTime time={new CivilTime(2, 35)} use24HourClock={false} />
+  -
+  <FormatTime time={new CivilTime(18, 35)} use24HourClock={false} />
+</Playground>
+
+## FormatTime Properties
+
+<Props of={FormatTime} />
```
kkimble006: Just two comments, 
1.) This `Props of`, is not actually listing anything, this section is empty when I view it (I re-ran npm start) and refreshed. 
2.) I think the package-lock changes were from me, I am not sure if they are needed?? Should we remove them before merging?

seanhealy: 1. They are showing here: https://deploy-preview-87--jobber-atlantis.netlify.com/components/time-formatter

image

Sometimes when running in dev mode on your local machine you need to do an `npm run clean` to clear some caching. The Netlify branch previews are usually an easier way to test things though. 😄 

2. The lock file should be fine.

PR #88, opened 2019-08-12
docs: InputColor proposal

cwtedwards:
### Added

- Proposal for `InputColor` component, an input for colors.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


docs/proposals/InputColor/README.md
```diff
@@ -0,0 +1,64 @@
+# InputColor
+
+The `<InputColor>` component will ... _{Add a brief description of the
+component}_
+
+## Design Patterns
+
+\_{The goal of the `InputColor` component is to allow a color be chosen to
```
seanhealy: I'd think about this a bit more broadly.

Perhaps something like…

> An `InputColor` should allow a user to choose a color.
> 
> They can either pick a color from the color picker or enter a value in the input.

docs/proposals/InputColor/README.md
```diff
@@ -0,0 +1,64 @@
...
+reflect the brand of an SP
+
+An SP can either pick a color from the color pick box or enter in the hex value
+and both the color box and the hew value should reflect eachother }\_
+
+## Wireframe
+
+![mockup](https://user-images.githubusercontent.com/51097786/62891872-1b79d980-bd04-11e9-9c5e-a66955519674.png)
```
seanhealy: I'm not seeing this image for some reason…

docs/proposals/InputColor/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Interface
+
+```jsx
```
seanhealy: For the interface here consider less of the implementation side and more on what you will be providing it and how it will make use of that.

https://github.com/GetJobber/atlantis/pull/66/files

docs/proposals/InputColor/README.md
```diff
@@ -0,0 +1,64 @@
...
+}
+```
+
+## Props Table
+
+_{Provide a table in the following format of the component's public API}_
+
+| name           | type     | default          | description                                                                |
+| -------------- | -------- | ---------------- | -------------------------------------------------------------------------- |
+| `color`        | `string` | `#7db00e`        | This will be the hex color that will be displayed                          |
```
seanhealy: There are some conventions it's probably good to follow for the props. Perhaps look at [InputText](https://atlantis.frend.space/components/input-text)for an example.

You can see how in here what you have as color would be named `value`

docs/proposals/InputColor/README.md
```diff
@@ -0,0 +1,64 @@
...
+| `colorContext` | `string` | `Choose a color` | This will be used to give context on where/what the color will be used for |
+
+## Accessibility
+
+_{Should be able to navigate to it with a keyboard but outside of that i'm not
```
seanhealy: If you want to take a look around online to see about accessibility considerations that other color inputs have made?

https://polaris.shopify.com/components/forms/color-picker#navigation

docs/proposals/InputColor/README.md
cwtedwards: Hmm i'll try again let me know if this one works.

docs/proposals/InputColor/README.md
cwtedwards: after our talk yesterday I believe this is what you are looking for.

docs/proposals/InputColor/README.md
cwtedwards: value has been added as well as the on change

docs/proposals/InputColor/README.md
cwtedwards: I think contrast is something to consider for accessibility.

docs/proposals/InputColor/README.md
```diff
@@ -0,0 +1,66 @@
...
+```jsx
+render() {
+return (
+      <InputColor
+    placeholder="Button Color"
+    value="#aeab2e"
+    onChange ={onChange}
+    list="dataListId"
+  />
+  <datalist id={dataListID}>
```
seanhealy: I don't follow what the `list` prop and this `datalist` is here?

seanhealy:
Looking pretty good except I'm not sure what the datalist thing is? I think you may be ok just removing it?

docs/proposals/InputColor/README.md
cwtedwards: the list binds the datalist to the input. It's used to give the user some options of colors we decide as well as an option to pick their own from the picker(its being used on the component right now)

seanhealy: Is DataList another component?

Is it used internally in this one?

cwtedwards: I don't see it being another component at the moment. It is used with this component. Do you want me to expand on it in the proposal itself?

seanhealy: We shouldn't be referencing elements by ID in React like you would in jQuery style code.

If the data list is meant to be an autocomplete menu for the color picker then it should probably be internal to the component.

If that's the case we may want to build an `autocomplete` component and then use that in the color picker here…?

docs/proposals/InputColor/README.md
```diff
@@ -0,0 +1,66 @@
...
+| `onChange`    | `onChange(newValue: colorName): void` | none             | Callback that fires when the color is changed changed                                                                                                                                                                                                                                                                                                                |
+| `dataListId`  | `string?`                             | none             | This will allow us to link a color datalist                                                                                                                                                                                                                                                                                                                          |
```
seanhealy:
```suggestion
```

docs/proposals/InputColor/README.md
```diff
@@ -0,0 +1,66 @@
...
+| `dataList`    | `string[]?`                           | none             | This will allow us to pass in some default colors to be listed when the color picker is selected but also allow them to access the color wheel. We currently are using this for the color picker right now. the datalist will not be another component however. This is a standard HTML element(https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist). |
```
seanhealy:
```suggestion
| `colorSuggestions`    | `Color[]?`                           | none             | This will allow us to pass in some default colors to be listed when the color picker is selected but also allow them to access the color wheel. We currently are using this for the color picker right now. the datalist will not be another component however. This is a standard HTML element(https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist). |
```

We probably want to give the colors in this list names? Making an interface for them might help with that.

eg.

```ts
interface Color {
  name: string;
  value: string;
}

const red = {
  name: "Red",
  value: "#ff0000",
}
```

cwtedwards: Talked to the designer and I think we will just let the datalist vanish when we swap to the atlantis component since it doesn't work across all browsers. This should clear out the last of the uncertainty hopefully.

seanhealy:
I'm going to be syncing up with @JeffMarvin this morning. Will loop back to this after the results of that. 😄

JeffMarvin:
This proposal can be merged. It can always be iterated upon, but what is here is good.

PR #89, opened 2019-08-14
Expose the root of ie11 problems

darryltec:
## Changes

- put variables on `:root` so that ie11 polyfill can polyfill it
- overwrite css properties instead of custom variables for better support on older browsers

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Button/Button.css
```diff
@@ -143,25 +155,39 @@
...
     135deg,
-    var(--loadingColor) 0%,
-    var(--loadingColor) 25%,
-    var(--bgColor) 25%,
-    var(--bgColor) 50%,
-    var(--loadingColor) 50%,
-    var(--loadingColor) 75%,
-    var(--bgColor) 75%,
-    var(--bgColor) 100%
+    rgba(0, 0, 0, 10%) 0%,
```
seanhealy: Are you not able to use the vars here?

PR #98, opened 2019-08-23
Prevents deletion of AM/PM an numbers if data is controlled by extern…

kkimble006:
…al react component, not itself.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputTime/InputTime.tsx
+++ b/packages/components/src/InputTime/InputTime.tsx
@@ -40,7 +40,11 @@ export function InputTime({
   } else {
     return (
       <InputTimeSafari
-        defaultValue={defaultValue}
+        defaultValue={
+          value == undefined && defaultValue != undefined
+            ? defaultValue
+            : undefined
+        }
         value={value}
         onChange={onChange}
         {...params}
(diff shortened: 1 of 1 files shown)


packages/components/src/InputTime/InputTime.tsx
```diff
@@ -16,14 +16,23 @@ export function InputTime({
   ...params
 }: InputTimeProps) {
   const handleChange = (newValue: string) => {
+    if (newValue === "") {
+      return;
+    }
```
kkimble006: If there is nothing to change, we don't change.

packages/components/src/InputTime/InputTime.tsx
```diff
@@ -16,14 +16,23 @@ export function InputTime({
...
+
   if (supportsTime) {
     const fieldProps: FormFieldProps = {
       onChange: handleChange,
-      ...(defaultValue && { defaultValue: civilTimeToHTMLTime(defaultValue) }),
-      ...(value && { value: civilTimeToHTMLTime(value) }),
+      ...(value == undefined &&
+        defaultValue != undefined && {
+          defaultValue: civilTimeToHTMLTime(defaultValue),
+        }),
```
kkimble006: The other error we were seeing, was if value wasn't defined, ex: "", we were trying to go back to defaultValue, but this was undefined.

kkimble006: So now we only use defaultValue if value is NOT defined and defaultValue IS defined.

PR #110, opened 2019-09-03
feat: Support generic validation messages on inputs.

darryltec:
## Changes

- This allows us to have a `success` `error` `warn` `info` messages on inputs
- This also includes some sort of criteria on when should the message show up like

```
{
  shouldShow: value !== 'Jeff',
  message: 'Your name is supposed to be Jeff',
  status: 'error',
}
```

Sep-03-2019 15-59-25

### Added

- A separate input validation component to show error messages outside of the form field

image

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -100,7 +100,7 @@ export interface FormFieldProps {
   /**
    * **EXPERIMENTAL** This feature is still under development.
    */
-  readonly validation?: ValidationProp[];
+  readonly validations?: ValidationProp[];
 
   /**
    * Set the component to the given value.
@@ -157,7 +157,7 @@ export const FormField = React.forwardRef(
       size,
       type = "text",
       value,
-      validation,
+      validations,
     }: FormFieldProps,
     ref:
       | Ref<HTMLInputElement>
@@ -173,13 +173,21 @@ export const FormField = React.forwardRef(
       handleValidation();
     }, [value]);
 
+    let hasErrors = false;
+    if (validations) {
+      hasErrors =
+        validations.filter(
+          validation => validation.shouldShow && validation.status === "error",
+        ).length > 0;
+    }
+
     const wrapperClassNames = classnames(
       styles.wrapper,
       inline && styles.inline,
       size && styles[size],
       align && styles[align],
       errorMessage && styles.hasErrorMessage,
-      (invalid || errorMessage) && styles.invalid,
+      (invalid || errorMessage || hasErrors) && styles.invalid,
       disabled && styles.disabled,
       maxLength && styles.maxLength,
       {
@@ -196,7 +204,7 @@ export const FormField = React.forwardRef(
           <Text variation="error">{errorMessage}</Text>
         )}
 
-        {validation && <InputValidation messages={validation} />}
+        {validations && <InputValidation messages={validations} />}
 
         <Wrapper
           className={wrapperClassNames}
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -186,15 +186,7 @@ export const FormField = React.forwardRef(
       handleValidation();
     }, [value]);
 
-    let hasErrors = false;
-    if (validations) {
-      hasErrors = validations.some(validation => {
-        return (
-          (validation.shouldShow || validation.shouldShow === undefined) &&
-          validation.status === "error"
-        );
-      });
-    }
+    const hasErrors = hasErrorMessages(validations);
 
     const wrapperClassNames = classnames(
       styles.wrapper,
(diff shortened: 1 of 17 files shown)


packages/components/src/FormField/FormField.tsx
```diff
@@ -30,15 +31,15 @@ export interface FormFieldProps {
...
    *
    * Show an error message and highlight the the field red.
    */
   readonly errorMessage?: string;
 
   /**
    * Adjusts the form field to go inline with a content. This also silences the
-   * given `errorMessage` prop. You'd have to used the `onValidate` prop to
+   * given `validations` prop. You'd have to used the `onValidate` prop to
    * capture the message and render it somewhere else using the `Text` component.
```
seanhealy:
```suggestion
   * capture the message and render it somewhere else using the `InputValidation` component.
```

packages/components/src/FormField/FormField.tsx
```diff
@@ -118,13 +127,21 @@ export interface FormFieldProps {
   onBlur?(): void;
 
   /**
-   * **EXPERIMENTAL** This feature is still under development.
+   * **DEPRECATED** Use `onValidation` prop instead.
```
seanhealy: I'm sort of thinking we should follow up this PR / Release with another one pretty soon that removes this… What do you think?

packages/components/src/FormField/FormField.tsx
```diff
@@ -166,13 +185,24 @@ export const FormField = React.forwardRef(
       handleValidation();
     }, [value]);
 
+    let hasErrors = false;
+    if (validations) {
+      hasErrors =
+        validations.filter(validation => {
```
seanhealy: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

packages/components/src/InputNumber/InputNumber.mdx
```diff
@@ -110,49 +110,6 @@ Input number is used in forms that accept numbers as an answer.
   </Text>
 </Playground>
 
-## Using onValidate
-
-If you need to capture the error message and render it outside of the component.
```
seanhealy: I wonder if we should include a link here to the `InputText` validation docs?

packages/components/src/InputValidation/InputValidation.mdx
```diff
@@ -0,0 +1,56 @@
...
+# Input Validation
+
+<ComponentStatus stage="pre" responsive="yes" accessible="yes" />
+
+This component allows you to show the validation messages outside of the inputs.
+
+<Playground>
+  {() => {
+    const [value, setValue] = useState("Juan");
+    const [validationMessage, setValidationMessage] = useState(undefined);
```
seanhealy:
```suggestion
    const [validationMessages, setValidationMessages] = useState(undefined);
```

packages/components/src/InputValidation/InputValidation.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface InputValidationProps {
+  /**
+   * Array of validation messages
+   */
+  messages: ValidationProps[];
+}
+
+interface StatusMap {
+  [key: string]: ValidationStatus;
```
seanhealy: Perhaps something like…

```suggestion
  [key: keyof ValidationStatus]: ValidationStatus;
```

I don't think that is quite the right syntax.

packages/components/src/FormField/FormField.tsx
darryltec: @seanhealy Yap... once this is out i'll replace them too

packages/components/src/InputNumber/InputNumber.mdx
darryltec: You's right. I'll send them to `InputValidation` tho just because the example there is mostly focused on it

packages/components/src/InputValidation/InputValidation.tsx
darryltec: @seanhealy funny thing about this... I actually don't need this anymore because it maps directly to text variations now

darryltec: there also seems to be problems with what you suggested tho https://github.com/microsoft/TypeScript/issues/5683

packages/components/src/FormField/FormField.tsx
```diff
@@ -166,13 +186,23 @@ export const FormField = React.forwardRef(
       handleValidation();
     }, [value]);
 
+    let hasErrors = false;
```
seanhealy: Was just thinking we could make a little function that would return `hasErrors` this would let us do something like `const hasErrors = areThereMessages(validations)` here which would simplify reading this component and then isolate this logic into a method. Also, it's an excuse to use VSCode's fancy refactor tools!

packages/components/src/InputText/InputText.mdx
```diff
@@ -50,6 +50,45 @@ Use this to allow users to provide long answers.
   />
 </Playground>
 
+## Validation message
+
+You can add your own custom validation messages on a field. However, this
+shouldn't replace of your server-side validation.
```
seanhealy:
```suggestion
shouldn't replace server-side validation.
```

PR #119, opened 2019-09-06
feat: First pass at Input Group

darryltec:
InputGroup component to allow inputs to be laid out in tight grids.

### Added
- Input group y'alls - but don't be nesting verticals in horizontals.

image

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -11,10 +11,13 @@
   --field--padding-left: var(--space-base);
   --field--padding-right: var(--space-base);
   --field--height: var(--space-largest);
+
+  --field--wrapper-elevation: auto;
 }
 
 .wrapper {
   position: relative;
+  z-index: var(--field--wrapper-elevation);
   width: 100%;
   background-color: var(--field--background-color);
 }
@@ -66,6 +69,7 @@ select.formField {
 
 .formField:focus {
   --field--border-color: var(--color-outline);
+  --field--wrapper-elevation: var(--elevation-base);
 }
 
 .label {
@@ -128,6 +132,7 @@ select.formField {
 .invalid,
 .invalid .formField:focus {
   --field--border-color: var(--color-red);
+  --field--wrapper-elevation: var(--elevation-base);
 }
 
 .disabled {
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -13,6 +13,11 @@
   --field--height: var(--space-largest);
 
   --field--wrapper-elevation: auto;
+
+  --public-field--top-left-radius: var(--radius-base);
+  --public-field--top-right-radius: var(--radius-base);
+  --public-field--bottom-left-radius: var(--radius-base);
+  --public-field--bottom-right-radius: var(--radius-base);
 }
 
 .wrapper {
@@ -41,7 +46,10 @@
   width: inherit;
   height: var(--field--height);
   border: var(--field--border-width) solid var(--field--border-color);
-  border-radius: var(--radius-base);
+  border-top-right-radius: var(--public-field--top-right-radius);
+  border-bottom-right-radius: var(--public-field--bottom-right-radius);
+  border-bottom-left-radius: var(--public-field--bottom-left-radius);
+  border-top-left-radius: var(--public-field--top-left-radius);
   outline: none;
   color: var(--field--color);
   font-size: var(--base-unit);
@@ -178,7 +186,7 @@ select.formField {
   --field--padding-top: var(--space-larger);
 }
 
-.miniLabel .textareaLabel {
+.miniLabel.textareaLabel .label {
   background-color: var(--color-white);
 }
 
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -27,6 +27,11 @@
   background-color: var(--field--background-color);
 }
 
+.wrapper:focus-within {
+  --field--border-color: var(--color-outline);
+  --field--wrapper-elevation: var(--elevation-base);
+}
+
 .formField,
 .label {
   box-sizing: border-box;
@@ -73,11 +78,6 @@ textarea.formField {
 
 select.formField {
   padding-right: calc(var(--field--padding-right) + var(--space-large));
-}
-
-.formField:focus {
-  --field--border-color: var(--color-outline);
-  --field--wrapper-elevation: var(--elevation-base);
 }
 
 .label {
--- a/packages/components/src/InputGroup/InputGroup.css
+++ b/packages/components/src/InputGroup/InputGroup.css
@@ -12,7 +12,7 @@
   margin-right: -1px;
 }
 
-/* Border Radius Tweak  */
+/* Border corners are rounded on the outside edges of the group  */
 
 .inputGroup {
   --public-field--top-left-radius: 0;
--- a/packages/components/src/InputGroup/InputGroup.css
+++ b/packages/components/src/InputGroup/InputGroup.css
@@ -1,0 +1,61 @@
+.vertical > *:not(:last-child) {
+  margin-bottom: -1px !important;
+}
+
+.horizontal {
+  display: flex;
+  align-items: flex-end;
+}
+
+.horizontal > *:not(:last-child) {
+  margin-bottom: auto;
+  margin-right: -1px;
+}
+
+/* Border Radius Tweak  */
+
+.inputGroup input,
+.inputGroup textarea,
+.inputGroup select {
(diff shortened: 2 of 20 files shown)


packages/components/src/InputGroup/InputGroup.css
```diff
@@ -0,0 +1,61 @@
...
+  margin-bottom: auto;
+  margin-right: -1px;
+}
+
+/* Border Radius Tweak  */
+
+.inputGroup input,
+.inputGroup textarea,
+.inputGroup select {
+  border-radius: 0;
```
darryltec: Idea: form field CSS could have a `--shared-formField--border-radius` Custom property. Our something along those lines. So that it's communicated that this thing can be modified by other components

packages/components/src/FormField/FormField.css
```diff
@@ -11,12 +11,14 @@
   --field--padding-left: var(--space-base);
   --field--padding-right: var(--space-base);
   --field--height: var(--space-largest);
+
+  --field--wrapper-elevation: auto;
 }
 
 .wrapper {
   position: relative;
+  z-index: var(--field--wrapper-elevation);
```
seanhealy: Why do we need `z-index` here?

packages/components/src/InputGroup/InputGroup.css
seanhealy: Do we want this to be customised?

packages/components/src/InputGroup/InputGroup.css
```diff
@@ -0,0 +1,61 @@
...
+}
+
+.vertical > :first-child > input,
```
seanhealy: This may get more complicated if we support new inputs that aren't true simple html types. For example a color input… This is also something that is more exciting for Formfield as well currently.

I'm not sure that we should worry about it now but it is something we should start thinking about…

seanhealy:
The focus borders seem to sometimes stack oddly.

image

image

JeffMarvin:
@seanhealy Ready for a 90% review. I think we should change the boolean prop to something different. Maybe a "direction" prop with a default?

packages/components/src/InputGroup/InputGroup.test.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import { cleanup } from "@testing-library/react";
+import { InputGroup } from ".";
+
+afterEach(cleanup);
+
+it("renders a vertical InputGroup", () => {
+  const tree = renderer.create(<InputGroup>Test content</InputGroup>).toJSON();
```
seanhealy: We can probably get away with snapshot tests here for now. Perhaps a simple one in each direction then a more complicated combination?

packages/components/src/InputGroup/InputGroup.mdx
```diff
@@ -0,0 +1,47 @@
...
+Group a different kinds fields to show a visual relationship. If a number of
+form fields once submitted make up a group of content (ie. an address, or first
+and last name becoming a full name) they are great candidates for an input
+group.
+
+<Playground>
+  <InputGroup>
+    <InputText placeholder="Street 1" />
+    <InputText placeholder="Street 2" />
+    <InputGroup horizontal={true}>
```
seanhealy: I wonder if we should consider the case where someone puts a vertical group within a horizontal one?…

JeffMarvin: I've added some documentation and checking in the event that they are used that way.

packages/components/src/InputGroup/InputGroup.test.tsx
JeffMarvin: Done, also included an invalid nesting case.

packages/components/src/InputGroup/InputGroup.mdx
```diff
@@ -0,0 +1,55 @@
...
+and last name becoming a full name) they are great candidates for an input
+group.
+
+The default layout is for fields to flow in a vertical direction, one above the
+next. Use the `flowDirection` prop to specify when you want fields to flow in a
+horizontal direction.
+
+**Warning** - Nested `InputGroup` components need to be of
+`flowDirection="horizontal"` to render properly. If you don't follow this, the
+group will not render and en error will appear in the browser console.
```
seanhealy:
```suggestion
> **Note:** Nested `InputGroup` components need to be of
> `flowDirection="horizontal"`. This ensures that we avoid layout issues
> cased by nesting. Any non-horizontal nested groups will log an error to
> the console and not be rendered
```

seanhealy:
```suggestion
> **Note:** Nested `InputGroup` components need to be of
> `flowDirection="horizontal"`. This ensures that we avoid layout issues
> cased by nesting. Any non-horizontal nested groups will log an error to
> the console and not be rendered
```

Quote and note might be a better / more consistent look for this?

Also played with the wording a bit…

packages/components/src/InputGroup/InputGroup.mdx
```diff
@@ -0,0 +1,55 @@
...
+      <InputText placeholder="Province" />
+    </InputGroup>
+    <InputGroup flowDirection="horizontal">
+      <InputText placeholder="Zip code" />
+      <InputText placeholder="Country" />
+    </InputGroup>
+  </InputGroup>
+</Playground>
+
+<Playground>
```
seanhealy: Normally more specific examples would come after a `---` following the props table.

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Button/Button.mdx

packages/components/src/InputGroup/InputGroup.tsx
```diff
@@ -0,0 +1,35 @@
...
+  children,
+  flowDirection = "vertical",
+}: InputGroupProps) {
+  if (checkChildren(children)) return <></>;
+
+  const className = classnames(styles.inputGroup, styles[flowDirection]);
+  return <div className={className}>{children}</div>;
+}
+
+function checkChildren(childs: ReactElement | ReactElement[]): boolean {
```
seanhealy: Name suggestions?

```suggestion
function invalidGroupNesting(children: ReactElement | ReactElement[]): boolean {
```

packages/components/src/InputGroup/InputGroup.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return React.Children.toArray(childs).some(child => {
+    if (
+      child.type === InputGroup &&
+      child.props.flowDirection != "horizontal"
+    ) {
+      console.error(
```
JeffMarvin: @seanhealy By making this a console *error* won't we trigger JS error reporting (like rollbar)? That could be noisy.

packages/components/src/InputGroup/InputGroup.tsx
JeffMarvin: smh - meant to improve this. Good suggestion, thanks.

seanhealy:
@JeffMarvin Test 2

packages/components/src/InputGroup/InputGroup.tsx
seanhealy: Only on production. We probably also want to know about it if it's happening there.

packages/components/src/InputGroup/InputGroup.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+---
+
+## Nested Example
+
+The default layout is for fields to flow in a vertical direction, one above the
+next. Use the `flowDirection` prop to specify when you want fields to flow in a
+horizontal direction.
+
+> **Note** - Nested `InputGroup` components need to be of
```
seanhealy:
```suggestion
> **Note:** Nested `InputGroup` components need to be of
```

Just to be consistent with other notes I think we've been doing `Note:`?

packages/components/src/InputGroup/InputGroup.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+function isInvalidGroupNesting(childs: ReactElement | ReactElement[]): boolean {
+  return React.Children.toArray(childs).some(child => {
+    if (
+      child.type === InputGroup &&
+      child.props.flowDirection != "horizontal"
+    ) {
+      console.error(
+        "ERROR: InputGroup not rendered: nesting " +
+          "flowDirection = vertical columns not supported.",
```
seanhealy:
```suggestion
        "ERROR: InputGroup not rendered: nesting `flowDirection="vertical"` columns not supported.",
        "\n",
        "https://atlantis.getjobber.com/components/input-group#nested-example",
```

JeffMarvin:
@seanhealy Changes made, back to you!

PR #121, opened 2019-09-10
feat: BREAKING CHANGE: Basic validation

darryltec:
## Changes

- Adds in `required` prop for all fields that checks if the value is there or not
- Updates input number min max validation to use `validation` props insted of error messages. This would break the `onValidate` prop that is used in the client notification settings

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/FormField/FormField.tsx
```diff
@@ -142,180 +147,179 @@ export interface FormFieldProps {
    * Callback to get the the status and message when validating a field
    * @param messages
    */
-  onValidation?(messages: ValidationProps[]): void;
+  onValidation?(messages?: ValidationProps[]): void;
```
seanhealy: Why did you make `messages` optional?

packages/components/src/FormField/FormField.tsx
```diff
@@ -142,180 +147,179 @@ export interface FormFieldProps {
...
+    const message = errorMessage || "";
+    onValidate && onValidate(status, message);
+    /* END:DEPRECATED */
+
+    onValidation && onValidation(validations);
+  }, [value]);
+
+  if (required && validations) {
+    validations.push({
+      message: `${placeholder || "This"} is required`,
```
seanhealy: Wondering if we can takes some lessons from things like this.

In this case suggesting a path forwards instead of pointing out a problem.

https://uxmovement.com/forms/how-to-make-your-form-error-messages-more-reassuring/

packages/components/src/FormField/FormField.tsx
```diff
@@ -142,180 +147,179 @@ export interface FormFieldProps {
...
+) {
+  const [hasMiniLabel, setHasMiniLabel] = useState(
+    defaultValue || value ? true : false,
+  );
+  const identifier = uuid.v1();
+  const Wrapper = inline ? "span" : "div";
+
+  useEffect(() => {
+    /* DEPRECATED */
+    const status = errorMessage ? "fail" : "pass";
```
seanhealy: I'm being dumb but, I think I have a question about this.

packages/components/src/FormField/FormField.tsx
darryltec: Because sometimes there's none

packages/components/src/FormField/FormField.tsx
darryltec: I was actually just following an established jobber pattern http://sg-1.getjobber.com/components/shared/field-error/#writing-guidelines I'd pass it to the designers tho

packages/components/src/FormField/FormField.tsx
darryltec: What is the question?

darryltec:
Just realized something. The required validation fires even when initial value is empty. that's not a good experience

packages/components/src/FormField/FormField.tsx
seanhealy: I think we should just send an empty array in that case…

eddysims:
Closing as @darryltec's way is no longer "valid"

giphy

JK, we miss you @darryltec. We have just gone another route with Validation.

PR #124, opened 2019-09-12
feat: When you find the tip of the tool. Tooltip

darryltec:
### Added

- Tooltip component
- Only supports hover

image

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Tooltip/Tooltip.css
```diff
@@ -0,0 +1,65 @@
+:root {
+  --tooltip--border-color: var(--color-lightBlue);
+  --tooltip--arrow-size: var(--space-small);
+  --tooltip--offset: calc(-1 * calc(var(--space-smaller) + 1px));
+}
+
+.shadowActivator {
```
seanhealy: This is the best classname!

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+    ) {
+      const activator = shadowRef.current.nextElementSibling;
+      const tooltip = tooltipRef.current;
+      const activatorBounds = activator.getBoundingClientRect();
+      const xOffset =
+        activatorBounds.right -
+        activatorBounds.width / 2 -
+        tooltip.clientWidth / 2;
+
+      if (activatorBounds.top <= 100) {
```
seanhealy: Do you think we should make this check relative to the viewport?

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+  readonly children: ReactElement;
+  /**
+   * Tooltip text
+   */
+  readonly message: string;
+}
+
+export function Tooltip({ message, children }: TooltipProps) {
+  const [direction, setDirection] = useState("above");
+  const [showOnLoad, setShowOnLoad] = useState(false);
```
seanhealy: Should we take this out as it's no longer an option?

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,136 @@
...
+  useEffect(() => {
+    if (
+      tooltipRef.current &&
+      shadowRef.current &&
+      shadowRef.current.nextElementSibling
+    ) {
+      const activator = shadowRef.current.nextElementSibling;
+      const tooltip = tooltipRef.current;
+      const activatorBounds = activator.getBoundingClientRect();
+      const xOffset =
```
seanhealy: I feel like it might be nice to extract some of this positioning logic out into a function. Perhaps something like…

```ts
getPosition(direction: "above" | "below", activator: HTMLElement, tooltip: HTMLElement): {x: number, y: number};
```

packages/components/src/Tooltip/Tooltip.css
darryltec: good, that took a good few minutes to figure out haha

packages/components/src/Tooltip/Tooltip.tsx
darryltec: I don't think so. At least not in this case just because the tooltip just needs to avoid showing up outside of the viewport.. which means if the bounds goes negative, it's outside of the viewport

packages/components/src/Tooltip/Tooltip.tsx
darryltec: we should. what would i do without you sean

packages/components/src/Tooltip/Tooltip.tsx
darryltec: would be nice. lets see if i can get that working without some crazy react types

packages/components/src/Tooltip/Tooltip.tsx
darryltec: Oh, this was just wierdly named state. i need this, but i'll name it properly

PR #125, opened 2019-09-16
feat: Create base page component

darryltec:
## Changes

- Updated markdown to support basic usage

### Added

- Created a page component that would extend to have more than a heading and intro

image

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Markdown/Markdown.test.tsx
+++ b/packages/components/src/Markdown/Markdown.test.tsx
@@ -1,168 +1,71 @@
 import React from "react";
 import renderer from "react-test-renderer";
-import { cleanup } from "@testing-library/react";
 import { Markdown } from ".";
-
-afterEach(cleanup);
 
 it("renders Text Component", () => {
   const tree = renderer.create(<Markdown content="Paragraph" />).toJSON();
-  expect(tree).toMatchInlineSnapshot(`
-    <div
-      className="base"
-    >
-      <p
-        className="base regular base greyBlueDark"
-      >
-        Paragraph
-      </p>
-    </div>
-  `);
+  expect(tree).toMatchSnapshot();
 });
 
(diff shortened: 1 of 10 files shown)


packages/components/src/Markdown/Markdown.test.tsx
```diff
@@ -1,168 +1,71 @@
...
-      className="base"
-    >
-      <p
-        className="base regular base greyBlueDark"
-      >
-        Paragraph
-      </p>
-    </div>
-  `);
+  expect(tree).toMatchSnapshot();
```
seanhealy: I wonder if we should update the `plop` generator to not use inline snapshot…

packages/components/src/Markdown/Markdown.tsx
```diff
@@ -26,22 +32,37 @@ interface BaseRendererProps {
   children: ReactNode;
 }
 
-export function Markdown({ content, externalLink }: MarkdownProps) {
+export function Markdown({ content, externalLink, basicUsage }: MarkdownProps) {
+  const props = {
+    ...(basicUsage && {
+      disallowedTypes: ["paragraph", "heading"],
```
seanhealy: Should we allow `images` and `tables`, etc here?

https://www.markdownguide.org/cheat-sheet

packages/components/src/Markdown/Markdown.test.tsx
darryltec: Yah, I plan to update the default content too.. I'll add a task

packages/components/src/Markdown/Markdown.tsx
darryltec: nope!

PR #126, opened 2019-09-17
feat: Add feature switch component

darryltec:
### Added

- Add `FeatureSwitch` component to handle switching of features
image

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/FeatureSwitch/FeatureSwitch.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <Switch
+            value={enabled}
+            onChange={handleSwitch}
+            ariaLabel={description}
+          />
+          <AnimatePresence>
+            {hasSaveIndicator && savedIndicator && (
+              <motion.div
+                className={styles.savedIndicator}
+                initial={{ y: -4, opacity: 0 }}
```
darryltec: Gonna extract this out on its own `Animate` component https://github.com/GetJobber/atlantis/pull/126/commits/b586aa7ac8d8d6460980b9ec5b9a01f912be0c1f

agglasaur:
Does the feature content show up before toggling the feature on? Does it make sense to grey out that section if it is? Not sure where we're landing on the hide vs show for that kind of stuff (or if it should be the same as our rules around disabled actions)

darryltec:
> Does the feature content show up before toggling the feature on? 

Yep. Continuing with the "_not knowing what the setting would do until you turn it on_". It would be really nice to have a stricter rule on when to show the feature content our settings and not just a generic case-by-case basis.

> Does it make sense to grey out that section if it is?

Quite possibly. It's gonna be a little harder to do that in this component because you can pass in any children you want inside this component. `opacity` would do it, but yeh...

> Not sure where we're landing on the hide vs show for that kind of stuff (or if it should be the same as our rules around disabled actions)

Same. Sounds like you(@agglasaur ), me, and @ChristyCampos needs to have a lil chat?

agglasaur:
Maybe something we can discuss in design review? Back to the ol patterns
discussion

PR #146, opened 2019-10-08
fix(components): The time formatter actually works for GB now and README changes

MandyMeindersma:
### Added

- a line to the read me specifying what version of npm is needed

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Changed the way use24HourClock was used. It should more accurately hold its value now. 
- Had to manually add the locale to the "toLocaleTimeString" function

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ developers to quickly build beautiful and consistent interfaces for our users.
 
 ## Development
 
+###### Requires npm v5.7.1 or higher
+
 To install Atlantis locally for development:
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -21,7 +21,10 @@ developers to quickly build beautiful and consistent interfaces for our users.
 
 ## Development
 
-###### Requires npm v5.7.1 or higher
+##### Prerequisites
+
+- `node@10` or higher
+- `npm@6` or higher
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -21,6 +21,8 @@ developers to quickly build beautiful and consistent interfaces for our users.
 
 ## Development
 
+###### Requires npm v5.7.1 or higher
```
seanhealy: Really tiny nit-picky thing but after some thought I think this might be a bit nicer as:

```suggestion
### Prerequisites

- `node@10.x.x`
- `npm@6.x.x`
```

MandyMeindersma: I added a new commit similar to this! Let me know if it's okay.

Did you actually want to use node version 10 and npm 6 or were those an example? I assumed they were an example but I can go in and change again. Now worries. 

Thank you for all the tips.

seanhealy: Strictly speaking we can get away with what you added but I think I'd like to encourage node 10+ and npm 6+ just to help keep some low hanging edges safe. 😄

MandyMeindersma: okay! Updated now! Perhaps this works?

PR #148, opened 2019-10-17
feat: modal title becomes optional and hides header if undefined

JeffMarvin:
### Added

- Modal: `title` is now an optional prop, and will now hide the Header if the title is not provided 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Modal/Modal.tsx
+++ b/packages/components/src/Modal/Modal.tsx
@@ -19,6 +19,7 @@ interface ModalProps {
    * @default true
    */
   readonly dismissible?: boolean;
+  readonly showTitle?: boolean;
   readonly children: ReactNode;
   readonly primaryAction?: ButtonProps;
   readonly secondaryAction?: ButtonProps;
@@ -31,6 +32,7 @@ export function Modal({
   title,
   size,
   dismissible = true,
+  showTitle = true,
   children,
   primaryAction,
   secondaryAction,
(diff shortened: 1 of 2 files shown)


packages/components/src/Modal/Modal.tsx
```diff
@@ -69,12 +71,13 @@ export function Modal({
               stiffness: 300,
             }}
           >
-            <Header
-              title={title}
-              dismissible={dismissible}
-              onRequestClose={onRequestClose}
-            />
-
+            {showTitle && (
```
seanhealy: Just thinking it might be nice to do

```suggestion
            {title != undefined && (
```

here.

JeffMarvin: Let's do it!

JeffMarvin:
@seanhealy PTAL - took a new approach and added tests!

packages/components/src/Modal/Modal.tsx
```diff
@@ -12,13 +12,14 @@ interface ModalProps {
...
    */
-  readonly title: string;
+  readonly title?: string;
   readonly open?: boolean;
   readonly size?: keyof typeof sizes;
   /**
    * @default true
    */
   readonly dismissible?: boolean;
+  readonly showTitle?: boolean;
```
seanhealy: It looks like this is no longer needed.

JeffMarvin: Oh geez... yup, removed.

PR #150, opened 2019-10-17
docs: Change meridiam to period (am/pm)

pontiphex:
Meridiam (sic) was unclear about what it referred to. Period is not great either, but adding AM/PM helps.

---

Random photo of Atlantis


seanhealy:
Thanks!!

PR #158, opened 2019-10-25
feat(components): Adding debounce functionality for the Autocomplete component

victorfjobber:
## Changes

- Added a debounce import from lodash
- Changed `setOptions` to be wrapped in debounce.

### Added
- Debounced method for fetching results

### Changed
- How setOption works

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Autocomplete/Autocomplete.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.tsx
@@ -1,4 +1,5 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
+import { debounce } from "lodash";
 import styles from "./Autocomplete.css";
 import { Menu } from "./Menu";
 import { Option } from "./Option";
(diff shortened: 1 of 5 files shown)


packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -15,6 +16,11 @@ interface AutocompleteProps {
    */
   readonly value: Option | undefined;
 
+  /**
+   * Set Autocomplete value.
+   */
+  readonly debounceTime: number | undefined;
```
darryltec: I think we can not have this prop and start with a specific time for debounce until we get a specific use case that we actually need one. Unless you have a good reason as to why :D

also, the description is talking about a value

victorfjobber: whoops

victorfjobber: Do you think we can set is as optional for now or should we remove it at all? I was thinking that we might as well be able to customize this because if we have autocompletes with local data we might want it to not have a delay time. And for faster interactions, we might want to decrease the delay time. We can leave it as a default of 200ms maybe?

PR #165, opened 2019-10-30
feat: showClose prop added to Banner

codymoorhouse:
## Changes
Screen Shot 2019-10-30 at 12 20 55 PM

### Added
- Added new prop to hide the close button called `onShow` to the Banner Component
- Added associated test for the new prop

### Changed
The Banner Component

### Deprecated
n/a

### Removed
n/a

### Fixed
n/a

### Security
n/a

## Testing
Render component, use the onShow property

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Banner/Banner.mdx
+++ b/packages/components/src/Banner/Banner.mdx
@@ -54,3 +54,13 @@ import { Banner } from "@jobber/components/Banner";
 <Playground>
   <Banner type="error">Hulk smash has been unsuccessful</Banner>
 </Playground>
+
+---
+
+## Show Close
+
+<Playground>
+  <Banner type="notice" showClose={false}>
+    Hulk smash was super effective
+  </Banner>
+</Playground>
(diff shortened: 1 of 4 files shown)


seanhealy:
@codymoorhouse I just wanted to get a bit of clarity on when a non-dismissible banner would be used. 😄

codymoorhouse:
> @codymoorhouse I just wanted to get a bit of clarity on when a non-dismissible banner would be used. 😄

The vision I was thinking for a more permanent banner... something like (this way it would make it look nice as it is an informative message box, but not closable because that message should probably persist). It will still default to being closable

image

seanhealy:
A notification banner has a different semantic meaning from an info card like that. You might be better off using a simple `Card` with an accent colour perhaps?

codymoorhouse:
> A notification banner has a different semantic meaning from an info card like that. You might be better off using a simple `Card` with an accent colour perhaps?
> 
> I'd try that first but we can also talk more about this if you like. :) Just want to make sure we're using the patterns as predictably as possible. 😄

Oooh cool, I didn't even know that was a thing! I will close this and use a card with accent instead, thank you 😃

PR #170, opened 2019-11-07
feat: Extract Plop generators to shareable package

seanhealy:
### Added

- `@jobber/generators` to allow for the sharing of generators on other projects.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

You can test this by installing the pre-release in another project and running plop there.

`npm install --save-dev @jobber/generators@next`

---

Random photo of Atlantis


package.json
```diff
@@ -7,7 +7,7 @@
     "start": "docz dev",
     "start:withPrivateComponents": "PRIVATE_COMPONENTS=visible docz dev",
     "test": "jest",
-    "plop": "plop",
+    "plop": "plop -- --path='./'",
```
seanhealy: Double check this path.

.eslintignore
```diff
@@ -2,4 +2,4 @@ node_modules
 dist
 .rpt2_cache
 packages/design/colors.js
-plop/templates
+packages/generators/templates
```
JeffMarvin: A couple of thoughts on naming. 
Because people use the `npm run plop` command to invoke it, they will be familiar with that term. I assume that we're using `generators` because we may want to change the implementation (i.e. plop) in the future? Should we change the npm command name as well? Don't get me wrong, I like plop 💩, but if they are importing a `generator`, it seems misaligned to still run `plop`. 

Is "generators" too generic? Would "componentGenerator" be more clear? (Both here in the code and when importing into something like Jobber.)

.stylelintignore
```diff
@@ -1,3 +1,3 @@
-packages/design/foundation.css
 public/foundation.css
```
JeffMarvin: This is a tiny thing, but there is a chance to alphabetize this list while it is small...

package.json
JeffMarvin: I believe this is the correct path.

PR #175, opened 2019-11-20
feat(components): add onClick function to tabs

eddysims:
## Changes

Allows tabs to have an `onClick` function when changing tabs.

### Added

- `onClick` prop to `InternalTab`
- Runs new `onClick` function along side the `activateTab` function

### Changed

- Moves `activateTab` function from the `onClick` prop to its on prop on `InternalTab`

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Tabs/Tabs.tsx
+++ b/packages/components/src/Tabs/Tabs.tsx
@@ -97,7 +97,9 @@ export function InternalTab({
   label,
   selected,
   activateTab,
-  onClick = () => undefined,
+  onClick = () => {
+    return;
+  },
 }: InternalTabProps) {
   const className = classnames(styles.tab, { [styles.selected]: selected });
   const color = selected ? "green" : undefined;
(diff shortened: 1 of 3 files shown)


packages/components/src/Tabs/Tabs.tsx
```diff
@@ -87,15 +89,28 @@ export function Tab({ label }: TabProps) {
...
+  activateTab(): void;
+  onClick?(): void;
 }
 
-export function InternalTab({ label, selected, onClick }: InternalTabProps) {
+export function InternalTab({
+  label,
+  selected,
+  activateTab,
+  onClick = () => undefined,
```
seanhealy: Does this work here instead?

```suggestion
  onClick = () => return,
```

That way we are returning `void` instead of a value (even if that value is undefined).

eddysims: yep it works! Changes pushed

PR #177, opened 2019-12-02
feat: Spinner Component V1

dmcrorieGIT:
### Added

Added the first iteration of `Spinner` for review and discussion.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,17 @@
...
+import classnames from "classnames";
+import styles from "./Spinner.css";
+
+export interface SpinnerProps {
+  readonly small?: boolean;
+  readonly inline?: boolean;
+}
+
+export function Spinner({ small, inline }: SpinnerProps) {
+  const className = classnames(styles.spinner, {
```
cwtedwards: Maybe we can change the name on this to be a little more specific to the component maybe spinnerStyles ?

cwtedwards:
I also think it would be a good idea to look into some accessibility for this component.

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,86 @@
+.spinner {
+  position: relative;
+  pointer-events: none;
+  /* 1 */
```
seanhealy:
```suggestion
```

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,86 @@
...
+.spinner:after {
+  content: "";
+  display: block;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  z-index: 3;
+  width: 2.25rem;
+  height: 2.25rem;
+  -webkit-box-shadow: 0px 0px 0.125rem 0.1875rem rgba(1, 27, 37, 0.3);
```
seanhealy: We have Atlantis will automatically generate any necessary browser specific rules like this.

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,86 @@
...
+  opacity: 0.6;
+}
+
+.spinner:after {
+  content: "";
+  display: block;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  z-index: 3;
```
seanhealy: Do we need this `z-indez`?

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,86 @@
...
+  width: 2.25rem;
+  height: 2.25rem;
```
seanhealy: We should convert the sizes in here to our space units.

Some examples of this here:
https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Switch/Switch.css

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,86 @@
...
+/* -------------------------------------------------------------------------
+   Small Spinner
+   ------------------------------------------------------------------------- */
+.small:after {
+  top: 50%;
+  left: 50%;
+  width: 1rem;
+  height: 1rem;
+  -webkit-box-shadow: 0px 0px 1px 2px rgba(1, 27, 37, 0.4);
+  box-shadow: 0px 0px 1px 2px rgba(1, 27, 37, 0.4);
```
seanhealy: Can probably use our shadow presets as well.

https://github.com/GetJobber/atlantis/tree/master/packages/design/src

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import classnames from "classnames";
+import styles from "./Spinner.css";
+
+export interface SpinnerProps {
+  readonly small?: boolean;
```
seanhealy: What do you think of changing this to…

```suggestion
  readonly size?: "small" | "base";
```

Then below we can set size to be `"base"` by default.

packages/components/src/Spinner/Spinner.css
dmcrorieGIT: With presets:
Screen Shot 2019-12-17 at 8 10 03 AM

Custom:
Screen Shot 2019-12-17 at 8 10 20 AM

@seanhealy I see what your saying, this css looks like garbage just hardcoded the way they are, but the preset box-shadow doesn't really look all that great on this component... Perhaps I could add to the shadow css var? Then I can call it here like:  `var(--shadow-spinner);` or something?

packages/components/src/Spinner/Spinner.css
dmcrorieGIT: Yup you're totally right, this doesn't serve a purpose! 😄

packages/components/src/Spinner/Spinner.css
dmcrorieGIT: Totally! I removed the other webkit stuff too, didn't seem to be doing anything either! 😄

packages/components/src/Spinner/Spinner.tsx
dmcrorieGIT: yup, that would be a better name for sure!! will change this 😄

packages/components/src/Spinner/Spinner.css
dmcrorieGIT: Excellent, using space units now!

packages/components/src/Spinner/Spinner.tsx
dmcrorieGIT: Good callout, this is much better. Changed! 😄

packages/components/src/Spinner/Spinner.css
seanhealy: We could add a `--shadow-dark` or something along those lines.

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import classnames from "classnames";
+import styles from "./Spinner.css";
+
+export interface SpinnerProps {
+  readonly size?: "small" | "base";
```
seanhealy: We've been trying to add a little docs line explaining each the props as well as any defaults here.

eg.
```suggestion
  /**
   * Specifies the size of the spinner.
   *
   * @default "base"
   */
  readonly size?: "small" | "base";
```

packages/components/src/Spinner/__snapshots__/Spinner.test.tsx.snap
```diff
@@ -0,0 +1,19 @@
...
+
+exports[`renders the inline spinner 1`] = `
+<div
+  className="spinner inline"
+/>
+`;
+
+exports[`renders the small spinner 1`] = `
+<div
+  className="spinner"
```
seanhealy: Should this specify small?

dmcrorieGIT:
Screen Shot 2019-12-20 at 2 15 57 PM

@seanhealy For some reason the Props table isn't being generated 🤔

packages/components/src/Spinner/__snapshots__/Spinner.test.tsx.snap
dmcrorieGIT: yup, good catch!

packages/components/src/Spinner/Spinner.tsx
dmcrorieGIT: Excellent, I've done this with the rest of the props as well!

packages/components/src/Spinner/Spinner.css
dmcrorieGIT: Perfect, I put the custom css into variable `--shadow-dark` 😄

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,48 @@
...
+  inline,
+  ariaBusy = "false",
+  role = "none",
+}: SpinnerProps) {
+  const spinnerStyles = classnames(styles.spinner, {
+    [styles.small]: size === "small",
+    [styles.inline]: inline,
+  });
+
+  return <div className={spinnerStyles} role={role} aria-busy={ariaBusy}></div>;
```
dmcrorieGIT: hardcode this, don't use props

bdfranck:
@dmcrorieGIT How is this going? Is there any way that I can help out?

dmcrorieGIT:
> dmcrorieGIT How is this going? Is there any way that I can help out?

@bdfranck Honestly I just haven't been working on this branch due to a busy sprint; that being said, this guy is at 90% with just a few things to bring it to the finish line, I'll commit to having this guy done by the end of the week 😅

dmcrorieGIT:
@seanhealy The props don't seem to be showing up for some reason?
Screen Shot 2020-02-03 at 10 48 49 AM

Other than that though, this component is ready for another round of review 😄

chris-at-jobber:
I'd like this to include when and why to use each of the variations (small, large, inline). I can work with whomever is driving this - I've done some [thinking around spinners](https://docs.google.com/document/d/19pQzAml07Xa0m4GEW_eURhjIRPCAiK6aBW-nLWlH-ZM/edit#heading=h.2fd1ptborieg) before.

chris-at-jobber:
Question about the accessibility aspect of this - does the spinner announce itself as loading to users with assistive technology? I'm assuming that's what the `Busy` attribute is doing?

dmcrorieGIT:
@chris-at-jobber Yeah man, I'd love to work with you on this one!! 

The `Busy` informs the user that the page is busy doing something if they're using some form of accessibility software; I don't know a ton about how it works or even how to test it though 🤷‍♂

JeffMarvin:
@dmcrorieGIT When you have a chance, please merge master into this. That might fix the failing deploy preview.

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import classnames from "classnames";
+import styles from "./Spinner.css";
+
+export interface SpinnerProps {
```
JeffMarvin: Unless these props are intended to be used outside of this file/module, there's no need to export them. 
```suggestion
interface SpinnerProps {
```

packages/components/src/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,54 @@
...
+route: /components/spinner
+---
+
+import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { Spinner } from ".";
+
+# Spinner
+
+<ComponentStatus stage="pre" responsive="no" accessible="no" />
```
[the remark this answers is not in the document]
chris-at-jobber: I believe that to make this accessible, we need to add an `aria-role="progressBar" aria-live="polite"` to the `spinner` element. This [question from Stack Overflow](https://stackoverflow.com/questions/38704467/how-to-label-a-loading-animation-for-wai-aria) seems to be a way to do so, although we'll want to test this in use. I think the `polite` value makes sense given that a spinner could be present multiple times in a view.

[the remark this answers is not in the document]
chris-at-jobber: It looks like we're taking a different approach [here](https://github.com/GetJobber/atlantis/pull/177/files#diff-fe7f208b80034abd704ce5220c007cf3R26) but not necessarily the wrong one! We should test this in the browser with a screen-reader to see what the experience is like.

packages/components/src/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+Spinners are used to visualize an undetermined load time for a section of a page
+and prevent double clicking of a UI element by overlaying it while its action is
+processing.
```
chris-at-jobber: Some more detail to flesh out these docs... 

- The default spinner should be used to indicate loading of a view
- The `small` spinner should be used on individual elements of the interface (ie a calendar "pill", a button, a card's contents) or when inline with content (ie a message that says "Uploading Clients... {spinner}").

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,78 @@
+.spinner {
+  position: relative;
+  pointer-events: none;
+  opacity: 0.6;
+}
+
+.spinner:after {
```
eddysims: Is there a reason to use a pseudo `:after` here instead of adding the styles onto the `.spinner` class?

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,78 @@
...
+  animation-duration: 1s;
+  animation-iteration-count: infinite;
+  animation-name: animation-spin;
+  animation-timing-function: linear;
+  will-change: transform;
+}
+
+@-webkit-keyframes animation-spin {
+  0% {
+    -webkit-transform: rotate(0deg);
```
eddysims: Question for @seanhealy maybe, but shouldn't our postcss handle prefixes?

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,78 @@
...
+.spinner:after {
+  content: "";
+  display: block;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  width: var(--space-larger);
+  height: var(--space-larger);
+  box-shadow: var(--shadow-dark);
+  margin: -1.125rem 0 0 -1.125rem;
```
eddysims: These margins are using magic `rems` while width and height are using the variable `--space-larger`.

A more safe margin might be something like:

```css
margin-top: calc(var(--space-larger) / 2 * -1);
margin-left: calc(var(--space-larger) / 2 * -1);
```

This way if the var changes we are still safe.

packages/components/src/Spinner/Spinner.css
seanhealy: It should / does.

dmcrorieGIT: @eddysims could you briefly explain postcss and prefixes?? 🤔Shot in the dark, but I'm guessing that this first 0% keyframe is unnecessary?

dmcrorieGIT:
@JeffMarvin I rebased current master, but the props still don't seem to want to render sadly 😢

packages/components/src/Spinner/Spinner.css
dmcrorieGIT: you're right, it's not needed!

packages/components/src/Spinner/Spinner.css
dmcrorieGIT: Totally, that's way better! Fixed!

packages/components/src/Spinner/Spinner.tsx
dmcrorieGIT: Took away export!

packages/components/src/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,54 @@
...
+  <Spinner />
+</Playground>
+
+## Props
+
+<Props of={Spinner} />
+
+---
+
+<!--
```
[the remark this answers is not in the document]
dmcrorieGIT: Totally, unnecessary dead comment. Removed!

eddysims:
closing in favour of https://github.com/GetJobber/atlantis/pull/456

PR #192, opened 2020-01-23
fix(components): buttons set zero margin explicitly

JeffMarvin:
## Motivations

This zeros out the margin for the `Button` component so that Browser defaults (like Safari's) don't get used.

### Added

- 

### Changed

- For more consistent support across browsers, we're explicitly zeroing out the margin on `Button`. Safari was adding some padding by default, but NO WAY Safari.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

1. Load the Atlantis docs viewer for the `Button` component.
2. Using the browser tools, inspect the rendered value for the margin of the Buttons. Confirm it is zero.
3. Visually inspect the three loading buttons in the example to make sure they have no margin between them: `/components/button#loading`

Tested in:
 - [x] Safari current
 - [x] Chrome current
 - [x] Firefox current

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -16,6 +16,7 @@
   display: inline-flex;
   min-height: calc(var(--base-unit) * 2.25);
   box-sizing: border-box;
+  margin: 0;
   border: 1px solid transparent;
   border-radius: var(--radius-base);
   color: transparent;


eddysims:
@JeffMarvin is there any concern about where this component is used in Jobber? Will this cause issues with UI where there used to be a browsers default margin would have been used?

seanhealy:
As far as I am aware we aren't relying on this margin anywhere as it is unintended.

PR #200, opened 2020-02-14
docs(components): Drawer Component Proposal

bdfranck:
## Motivations

I wanted to propose a new Drawer component.

### Added

I added a proposal for the component.

### DIFF
--- a/docs/proposals/Drawer/README.md
+++ b/docs/proposals/Drawer/README.md
@@ -55,14 +55,16 @@ shouldn't overtake the main content in tab priority.
 ## Resources
 
 Work request drawer:
-https://github.com/GetJobber/Jobber/blob/master/app/assets/stylesheets/desktop/application.scss#L231
-https://github.com/GetJobber/Jobber/blob/master/app/assets/javascripts/desktop/utilities/drawer.js.coffee
+
+- https://github.com/GetJobber/Jobber/blob/master/app/assets/stylesheets/desktop/application.scss#L231
+- https://github.com/GetJobber/Jobber/blob/master/app/assets/javascripts/desktop/utilities/drawer.js.coffee
 
 Similar components in other design systems:
-https://material-ui.com/components/drawers/
-https://www.lightningdesignsystem.com/components/panels/
-https://ant.design/components/drawer/
-https://designsystem.quickbooks.com/component/drawers/
+
+- https://material-ui.com/components/drawers/
+- https://www.lightningdesignsystem.com/components/panels/
+- https://ant.design/components/drawer/
+- https://designsystem.quickbooks.com/component/drawers/
 
 The Jira ticket for this component:
 https://jobber.atlassian.net/browse/JOB-18723
--- a/docs/proposals/Drawer/README.md
+++ b/docs/proposals/Drawer/README.md
@@ -5,7 +5,8 @@ The `<Drawer>` component will provide a dismissible side panel.
 ## Design Patterns
 
 `<Drawer>` It will allow the user to view related content to their main task.
-Unlike a `<Modal>`, it will not obscure the main content from the user.
+Unlike a `<Modal>`, it will take full advantage of the available viewport
+height. This will make it ideal for long lists and other scrolling content.
 
 Some scenarios for a `<Drawer>` include the following:
 
@@ -50,3 +51,15 @@ https://www.figma.com/file/wKlxfdTZlrffPrPiWpEOEB/Insightful-Dashboard-Component
 
 The user should be able to navigate to the drawer by keyboard. The drawer
 shouldn't overtake the main content in tab priority.
+
+## Resources
+
+Work request drawer:
+https://github.com/GetJobber/Jobber/blob/master/app/assets/stylesheets/desktop/application.scss#L231
+https://github.com/GetJobber/Jobber/blob/master/app/assets/javascripts/desktop/utilities/drawer.js.coffee
+
+Similar components in other design systems:
+https://material-ui.com/components/drawers/
+https://www.lightningdesignsystem.com/components/panels/
+https://ant.design/components/drawer/
+https://designsystem.quickbooks.com/component/drawers/
--- a/docs/proposals/Drawer/README.md
+++ b/docs/proposals/Drawer/README.md
@@ -8,7 +8,7 @@ The `<Drawer>` component will allow users to view content that relates to their
 main task. It would appear on the right-hand side of the screen. The component
 would be toggled open and closed by a specified button or link. The component
 would fill the available vertical height of the parent container. It would allow
-the user to still view and interact with the main content. The user could
+the user to still view and interact with the main content. The user could configure the displayed content. The user could
 dismiss it once they are finished.
 
 Some scenarios for a `<Drawer>` include the following:
--- a/docs/proposals/Drawer/README.md
+++ b/docs/proposals/Drawer/README.md
@@ -24,7 +24,7 @@ There are also some outstanding questions:
 
 ## Wireframe
 
-https://www.figma.com/file/wKlxfdTZlrffPrPiWpEOEB/Insightful-Dashboard-Components?node-id=0%3A1
+<iframe style="border: none;" width="800" height="450" src="https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Ffile%2FwKlxfdTZlrffPrPiWpEOEB%2FInsightful-Dashboard-Components%3Fnode-id%3D0%253A1" allowfullscreen></iframe>
 
 ## Interface
 
--- a/docs/proposals/Drawer/README.md
+++ b/docs/proposals/Drawer/README.md
@@ -2,8 +2,6 @@
 
 The `<Drawer>` component will provide a dismissible right-hand side panel.
 
-<iframe style="border: none;" width="800" height="450" src="https://www.figma.com/embed?embed_host=share&url=https%3A%2F%2Fwww.figma.com%2Fproto%2FwKlxfdTZlrffPrPiWpEOEB%2FInsightful-Dashboard-Components%3Fnode-id%3D0%253A1%26viewport%3D645%252C478%252C0.5%26scaling%3Dmin-zoom" allowfullscreen></iframe>
-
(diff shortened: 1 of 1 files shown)


docs/proposals/Drawer/README.md
```diff
@@ -0,0 +1,70 @@
+# Drawer
+
+The `<Drawer>` component will provide a dismissible side panel.
+
+## Design Patterns
+
+`<Drawer>` It will allow the user to view related content to their main task.
+Unlike a `<Modal>`, it will take full advantage of the available viewport
+height. This will make it ideal for long lists and other scrolling content.
```
seanhealy: Other than the placement is there anything semantically different between a drawer and a modal?

Do you see users being able to open a drawer and then be able to continue interacting with the base page?

docs/proposals/Drawer/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Some scenarios for a `<Drawer>` include the following:
```
seanhealy: Do you see a situation where you could drill into more detail from a drawer?

One potential example:
image

I could also see drawer content being treated a bit like a [UINavigationController](https://developer.apple.com/documentation/uikit/uinavigationcontroller)?


If there will be navigation within a drawer should we consider how that might work here?

docs/proposals/Drawer/README.md
bdfranck: Yes, it would be excellent to have an "inline" version that displaces the content rather than overlays it. We're using the work overview one in our prototype dashboard right now.

image

docs/proposals/Drawer/README.md
```diff
@@ -0,0 +1,70 @@
...
+  </Content>
+</Drawer>
+```
+
+## Props Table
+
+| name      | type    | default | description                                             |
+| --------- | ------- | ------- | ------------------------------------------------------- |
+| title     | String  | -       | The title of the drawer                                 |
+| subtitle  | String  | -       | The subtitle of the drawer                              |
```
seanhealy: I can't quite put my finger on why but, `subtitle` feels a bit funny to me. Perhaps it's just how much this currently feels like a modal? Or, how it would work if navigation became part of a drawer…?

docs/proposals/Drawer/README.md
```diff
@@ -0,0 +1,70 @@
...
+| overlay   | String  | -       | If the drawer overlays the main content or displaces it |
```
seanhealy: I think that if `overlay` is true this very much becomes a modal. In those cases would a modal be a better fit?

docs/proposals/Drawer/README.md
```diff
@@ -0,0 +1,70 @@
...
+| open      | Boolean | -       | If the drawer is visible                                |
+| activator | Boolean | -       | The component that opens the drawer                     |
```
seanhealy: I don't think we'll need this prop.

docs/proposals/Drawer/README.md
bdfranck: Hmmm, maybe. But initially we just need it to display a list or other content. I can't imagine needing anything sophisticated like navigation.

docs/proposals/Drawer/README.md
bdfranck: Gotcha! 👍

docs/proposals/Drawer/README.md
jlelford: If we're counting the AF drawer as being part of this component it currently has similar functionality for its settings customization.

docs/proposals/Drawer/README.md
bdfranck: Hmmm, subtitle may be too prescriptive.

docs/proposals/Drawer/README.md
bdfranck: Hmmm, I was thinking that the activity feed could be changed from a modal to a drawer with an overlay but the behaviour would be similar. The biggest difference that I could see would be a sliding animation... which I haven't really thought about yet.

bdfranck: A close analogue would be the 'temporary' and 'persistent' drawers in Material Design: https://material-ui.com/components/drawers/.

docs/proposals/Drawer/README.md
bdfranck: Woah! I didn't know this existed within the panel.

image

docs/proposals/Drawer/README.md
devinserink: I wonder if this is a key difference between a 'side modal' and a 'side drawer' that would keep us from wanting to use it, even in overlay use cases. This is a key aspect of Modal: "It also prevent users from interacting with the rest of the application until a specific action is taken".   Seems like drawer use cases might be less flow-interrupting, and should probably support easy dismissal (escape key? click out? swipe away?), especially if they're overlaying other content.

docs/proposals/Drawer/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Accessibility
+
+The user should be able to navigate to the drawer by keyboard. The drawer
+shouldn't overtake the main content in tab priority.
```
chris-at-jobber: Should the drawer overtake the main content if there is an overlay?

docs/proposals/Drawer/README.md
chris-at-jobber: > Unlike a `<Modal>`, it will take full advantage of the available viewport height. This will make it ideal for long lists and other scrolling content.
This might be nitpicking, but could a `Modal` with a max-height and better handling of vertical overflow solve the same problem? Right now the behaviour of dialogs/modals in Jobber is that the entire view scrolls with the modal if "contentHeight" > "viewportHeight" which feels to me like the key difference being described here.

chris-at-jobber:
@bdfranck Would a fair summary of the proposed distinction between a `Drawer+overlay` and a `Modal` be that Modals would remain responsible for creation, editing, etc, while a drawer would only be responsible for displaying content?

docs/proposals/Drawer/README.md
bdfranck: Okay, I think there's too. much overlap between a modal and a drawer with an overlay. I'll update my proposal in a way that doesn't use the modal.

bdfranck:
@chris-at-jobber Does that simplify things? I think the modal and the drawer+overlay are too similar so I removed it.

docs/proposals/Drawer/README.md
```diff
@@ -0,0 +1,68 @@
...
+- Show a work request beside a new quote form
+- Show a converted quote beside an active job
+- Show a list of insight cards beside a dashboard
+
+There are also some outstanding questions:
+
+- Should the drawer be full-height of the viewport or just the parent container?
+- How should the drawer animate when it opens and closed?
+- Should we include an option for left-hand and right-hand drawers?
+- Do all drawers need to be dismissible? What about an always-open version?
```
chris-at-jobber: If a drawer is always open I _think_ there's a good argument that it isn't a `Drawer` (because it's just a static column in a layout)... maybe we could keep with the container metaphor and call that a `Basket` 😆

docs/proposals/Drawer/README.md
```diff
@@ -0,0 +1,68 @@
...
+Some scenarios for a `<Drawer>` include the following:
+
+- Show a list of notes beside a job detail view
+- Show a work request beside a new quote form
+- Show a converted quote beside an active job
+- Show a list of insight cards beside a dashboard
+
+There are also some outstanding questions:
+
+- Should the drawer be full-height of the viewport or just the parent container?
```
chris-at-jobber: If we take the overlay aspect out of this, parent seems sensible.

docs/proposals/Drawer/README.md
```diff
@@ -0,0 +1,68 @@
...
+- How should the drawer animate when it opens and closed?
+- Should we include an option for left-hand and right-hand drawers?
+- Do all drawers need to be dismissible? What about an always-open version?
+- Could the unscheduled visits accordion on the calendar turn into a drawer?
```
chris-at-jobber: I think so, that one was kind of shoe-horned into whatever I could make work when I built it and I'm not sure there's a great reason to keep it separate

chris-at-jobber:
@bdfranck I think this is much simpler now and also helps address some of your Unanswered Questions™

docs/proposals/Drawer/README.md
bdfranck: @chris-at-jobber Hmmm, do you think I need to incorporate this multi panel thing into the official proposal?

docs/proposals/Drawer/README.md
bdfranck: @chris-at-jobber It’s a drawer where something fell behind it and now it won’t close. 😉

bdfranck:
@chris-at-jobber Okay, I've updated the proposal with the answers to the questions. I've also included a link to a very simple prototype. What else should I provide?

docs/proposals/Drawer/README.md
chris-at-jobber: I think it's a valid use case, even in a non-overlay drawer - what comes to mind for me is giving the SP control over what kinds of insights they want in the insightful dashboard drawer. If there's no current plan to implement something like that, then maybe it's not necessary for the proposal, but if you think there's a strong likelihood that it would be valuable I would include it.

chris-at-jobber:
@bdfranck this looks good to me, I responded to your last question about the "settings/drill down" aspect - since I wrote that, I'm more convinced it should be included because it feels like a natural progression towards giving users more control, closer to the thing they're controlling. Something like "The drawer should have an option to give the user the ability to adjust content"

bdfranck:
@chris-at-jobber Cool, I've added the ability to configure content to the description.

bdfranck:
@chris-at-jobber What happens now? Is there anything else that I need to do?

chris-at-jobber:
@bdfranck The proposal looks good so the next step is building it 🏗 

### Who builds it?
Your team owns the development and Frends will be there along the way to review and offer assistance/consultation

### Anything else?
 @seanhealy has some thoughts around similarities to `Modal` so whomever on your team is involved in the building of this component should reach out when they're getting started

PR #204, opened 2020-02-27
feat(components): Add FormatDate component

naarok:
## Motivations

We need to format dates in Atlantis. Currently, only date format supported is:
MMM d, yyyy
ex. Feb 26, 2020

### Added

- FormatDate Component

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormatDate/FormatDate.tsx
+++ b/packages/components/src/FormatDate/FormatDate.tsx
@@ -1,0 +1,32 @@
+import React from "react";
+import { CivilDate } from "@std-proposal/temporal";
+
+interface FormatDateProps {
+  /**
+   * Date to be displayed.
+   */
+  readonly date: CivilDate;
+}
+
+export function FormatDate(date: FormatDateProps) {
+  return <>{strFormatDate(date.date)}</>;
+}
+
+function strFormatDate(date: CivilDate) {
+  const monthNames = [
+    "Jan",
+    "Feb",
+    "Mar",
+    "Apr",
+    "May",
+    "Jun",
+    "Jul",
+    "Aug",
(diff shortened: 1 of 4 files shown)


packages/components/src/FormatDate/FormatDate.tsx
```diff
@@ -0,0 +1,32 @@
...
+    "Jun",
+    "Jul",
+    "Aug",
+    "Sep",
+    "Oct",
+    "Nov",
+    "Dec",
+  ];
+
+  return monthNames[date.month - 1] + " " + date.day + ", " + date.year;
```
seanhealy: The updated Temporal Polyfill provides [toLocaleString](https://tc39.es/proposal-temporal/docs/date.html#toLocaleString) on a Temporal `Date`. We could consider updating to that. Another option would be to use the currently [built in Date formatter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) along the lines of what we do in `FormatTime`. Eg.

```suggestion
const formattingDate = new Date(
    date.year,
    date.month,
    date.day,
  )

  return formattingDate.toLocaleDateString(undefined, { year: 'numeric', month: 'short', day: 'numeric' });
```

I think it would be better to use one of these built in formatters since they're already there. 😄

naarok: Yup. How embarrassing. I was looking for that functionality, but my google foo failed. In my defence. searching for "javascript date format" in google. At least the first three hits don't give the right answer.

Will fix

PR #206, opened 2020-03-05
feat(components): Create ProgressBar

codymoorhouse:
## Motivations

As part of the the new enterprise on-boarding process we are creating, we needed a status bar to track which step in a wizard you are along the way. It seems like a good reusable component to add to Atlantis.

### Added

- A new progress component

### Changed

- Minor version bumps on some package-locks

## Testing

Using the netflify is a good approach. I created an example where you can click a button to see how it can be used.

---

Random photo of Atlantis


packages/components/src/ProgressBar/ProgressBar.css
```diff
@@ -0,0 +1,20 @@
...
+  height: var(--progress-bar--height);
+  border-radius: var(--progress-bar--borderRadius);
+  background: var(--progress-bar--backgroundColor);
+}
+
+.content {
+  height: var(--progress-bar--height);
+  border-radius: var(--progress-bar--borderRadius);
+  background-color: var(--progress-bar--barColor);
+  transition: width 0.75s ease-out;
```
eddysims: we have a `--timing-slow` css variable available at `:root`. Would we want to consider this here instead of a custom transition time?

packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * The total steps to use. For percentages you can set this to 100.
+   */
+  readonly totalSteps: number;
+}
+
+export function ProgressBar({ currentStep, totalSteps }: ProgressBarProps) {
+  const percentage = ((currentStep - 1) / (totalSteps - 1)) * 100;
+  const widthPercent = Math.min(100, Math.max(0, percentage));
+  const className = classnames(styles.container);
```
eddysims: Might want to consider a more clear name here. I think common practice would be something like `progressBarClassName`.

eddysims:
One thing I noticed is that our steps can go above and below the totalSteps. Probably want to prevent this.

image


image

codymoorhouse:
I updated the MDX as the buttons are external to the component to reflect how we can make sure we are not going below or over the max steps 😄

PR #210, opened 2020-03-13
feat(components): ListItem can have string as id

ibarreto:
## Motivations

We currently mostly use numeric IDs for objects in the Jobber database. We do have some objects that have UUIDs. There's also external objects (like the ones from Stripe) that have unique string ids. I wanted to add support of string IDs to this component

### Added

- The `ListItem` id can now be a string or number. Previously it was only number.
### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


ibarreto:
Noticed this in the preview. Should be `id | number` but it says `Reacttext`. Not sure if I'm missing something.
🔱_Atlantis

PR #216, opened 2020-03-16
chore: Upgrade docz - Elevation 

eddysims:
## Motivations

After the docz upgrade, the `Design` > `Elevation` was no longer working. This was due to React Fragment shorthand not being available in the `.mdx` file.

### Added

- 

### Changed

- Replaces fragment with `<React.Fragment>`

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/design/src/Elevations.mdx
+++ b/packages/design/src/Elevations.mdx
@@ -11,7 +11,7 @@ The following table lists all elements that has a z-index specified on it.
 export function Value({ of: type }) {
   const styles = getComputedStyle(document.documentElement);
   const getValue = styles.getPropertyValue(`--elevation-${type}`);
-  return <>{getValue}</>;
+  return <span>{getValue}</span>;
 }
 
 | Component        | Value                  |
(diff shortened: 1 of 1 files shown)


packages/design/src/Elevations.mdx
```diff
@@ -11,7 +11,7 @@ The following table lists all elements that has a z-index specified on it.
 export function Value({ of: type }) {
   const styles = getComputedStyle(document.documentElement);
   const getValue = styles.getPropertyValue(`--elevation-${type}`);
-  return <>{getValue}</>;
+  return <span>{getValue}</span>;
```
seanhealy: You likely tested this already but, are you able to use `React.Fragment` here?

I wonder if we could also just return `getValue`?

eddysims: I did not try the `<React.Fragment>` and it worked! Thanks for that

PR #220, opened 2020-03-19
chore: Keep our privates private once again.

seanhealy:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Private components are once again hidden away safely.

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/doczrc.js
+++ b/doczrc.js
@@ -51,10 +51,14 @@ const projectPlugin = () =>
   });
 
 /**
+ * Return an array of private components to hide from the user.
+ *
  * Atlantis contains some private components. These are indicated by an
  * empty `.private` file in the component's directory.
+ *
+ * @return {string[]} Private components.
  */
-function privateComponentReadmies() {
+function privateComponentReadmes() {
   // If specified at run time show documentation for privates components.
   if (process.env.PRIVATE_COMPONENTS === "visible") {
     return [];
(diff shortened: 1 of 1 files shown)


doczrc.js
```diff
@@ -50,15 +50,34 @@ const projectPlugin = () =>
     },
   });
 
+/**
+ * Atlantis contains some private components. These are indicated by an
+ * empty `.private` file in the component's directory.
+ */
 function privateComponentReadmies() {
```
JeffMarvin: Speeeling... Either this
```suggestion
function privateComponentReadmes() {
```
or something like
```suggestion
function privateComponentVisibility() {
```

doczrc.js
```diff
@@ -50,15 +50,34 @@ const projectPlugin = () =>
     },
   });
 
+/**
+ * Atlantis contains some private components. These are indicated by an
```
JeffMarvin: What does this function do? That should be the first part of the comment. The other notes are nice (and can be included) but don't explain the why of the function.

seanhealy:
@JeffMarvin As per npm convention, you can always [find commands in the `package.json`](https://github.com/GetJobber/atlantis/blob/master/package.json#L8) I don't know that it should be in the readme?

JeffMarvin:
> @JeffMarvin As per npm convention, you can always [find commands in the `package.json`](https://github.com/GetJobber/atlantis/blob/master/package.json#L8) I don't know that it should be in the readme?

Agreed, the package.json is the right place for it.

LGTM

PR #222, opened 2020-03-23
fix(components): export the colors type as a type for use

colefichter:
## Motivations
In the main jobber repo, we need to pass a color string all the way from rails into a react component, then apply it to an `<InlineLabel color="something">` tag. This change allows us to avoid type errors when passing the color in.

## Changes
Exporting a color type from InlineLabel component.

## Testing
It compiles! I will know it was successful if a client library can use the type successfully.

Note: Sean Healy was kind enough to walk me through the actual changes. They should be good to go!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2207,6 +2207,8 @@
         },
         "@jobber/formatters": {
           "version": "0.0.4",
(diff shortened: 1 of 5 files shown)


eddysims:
LGTM

PR #219, opened 2020-03-18
docs(design): include design guidelines for new icon creation

chris-at-jobber:
## Motivations

Every time someone creates a new icon, there's a question of "what are the guidelines to follow?"

## Changes

Adding guidelines

### Added

Guidelines for icon structure

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #223, opened 2020-03-24
docs(generators): Add design decisions to template Readme

eddysims:
## Motivations

It is important that when people contribute to Atlantis that they keep design decisions in mind. This PR will add Design Decisions section to the current template.

As I was testing, I found that the generator was going to the wrong directory, I have fixed to put the right directory in the `package.json` generate script.

## Changes

- Adds design decisions to template

### Added

- Design decisions to template

### Changed

- path of generated component

### Deprecated

- 

### Removed

- 

### Fixed

- path of generated component

### Security

- 

## Testing

- `npm run generate` a component

---

Random photo of Atlantis


eddysims:
@seanhealy I have added an `Example` component that will be in the netlify preview. We will need to remove it before this gets merged.

packages/generators/templates/component/{{name}}.{{mdx}}
```diff
@@ -28,6 +28,16 @@ import { {{ name }} } from "@jobber/components/{{ name }}";
 
 ---
 
+## Design Guidelines
+
+When contributing to the {{ name }} component, follow these guidelines:
+
```
JeffMarvin: I like the sections here. I wonder if we will get more quality information added (for both design guidelines and best practices) if we provide some examples, or even link to a component file/doc that does a good job of filling these in.

seanhealy:
@eddysims I think the example needs to be re-generated. 😄

JeffMarvin:
@eddysims Right now, the layout seems a bit hard to parse. The design guidelines and the best practices sections are interspersed among the other content.... 

image

packages/generators/templates/component/{{name}}.{{mdx}}
seanhealy: I wonder if @chris-at-jobber would have any thoughts along those lines. Even if the example was types of things to consider or something?

chris-at-jobber: To me this is a nice addition, it seems to be mirroring the placement from the [Icons PR](https://github.com/GetJobber/atlantis/pull/219) but outside of that context I think "design guidelines" is equivalent to usage guidelines a la [Button](https://atlantis.frend.space/components/button#work-actions). 

Icons are interesting in that they're such a primarily visual component, the structural notes there are  more structural guidelines than usage guidelines for the designers. 

---

# EXAMPLE

_Label | Label | Label_
[ import snippet thing ]

A paragraph section for the component description

[ Component Sandbox chunk ]

## Design Guidelines

This section would include things like Work Actions, Learning Actions, etc. 

## Props

Props stuff

---

I realize that this makes it a more design-centric way of documenting, but I feel like putting usage guidelines before technical guides could help teams better identify if a given component is right for their use case.

chris-at-jobber: Do you see the BEST PRACTICES section serving the role I'm talking about here?

chris-at-jobber:
@JeffMarvin following up on [this](https://github.com/GetJobber/atlantis/pull/223#issuecomment-604563761), I think we need to look at the "variations" of a component being grouped a bit more tightly (see "LOUD" in this Example, "WORK ACTIONS" on the [Button docs](https://atlantis.frend.space/components/button#work-actions), etc)

There are conceptual variations (ie Work, Learning, Destructive), visual variations (Icon, Size), and state variations (Loading) - with that lens do you think we could organize these Design Guidelines in that manner? I'd have to take a look at some of the other components to see if it makes sense, but I think some more aggressive hierarchy-ing would be helpful.

packages/generators/templates/component/{{name}}.{{mdx}}
eddysims: In regards to best practices, I really like what shopify does on their [components](https://polaris.shopify.com/components/forms/autocomplete#section-best-practices). They give information about how to use and *not use* each component.

Do you all think this would be a good place to keep items like this? In my mind "best practices" would continue to grow per component as consumers have questions about the component.

eddysims:
@chris-at-jobber @seanhealy @JeffMarvin 

I have updated the example to the following:

image

I feel that this is better than before as it allows for an optional section that can handle questions about design and consumption of a component. It can also provide the "best practice"  advice that I was hoping for in the previous version.

Can I get some thoughts on this?

JeffMarvin:
@chris-at-jobber [To your comment:](https://github.com/GetJobber/atlantis/pull/223#issuecomment-604563761),

>  I think we need to look at the "variations" of a component being grouped a bit more tightly (see "LOUD" in this Example, "WORK ACTIONS" on the [Button docs]
> I think some more aggressive hierarchy-ing would be helpful.

I agree. But we're just modifying the template in this PR.
I think we'd get more bang for our effort if we edited a few of the prominent components to have better content, than putting it in the generated one and hoping people get it.

JeffMarvin:
@eddysims I like the new combined sections. I think it would be best to be first, and then the code related stuff: Props section and other usage examples.

eddysims:
The layout is updated to show the design guidelines before the props. I now agree that I think this makes sense and flows quite nicely.

image

JeffMarvin:
@eddysims Ok this is good to go! Please pull the example component out and I will approve this when tests pass.

eddysims:
@JeffMarvin back to you, all checks have passed. After i get an approval ill merge this

PR #229, opened 2020-04-20
feat(components): Adds FormatEmail component

eddysims:
## Motivations

We would like a way to make sure that we are formatting email addresses the same way throughout Jobber. `<FormatEmail />` will take an email address and return a formatted email address.

## Changes

- Adds new FormatEmail component

### Added

- FormatEmail component

## Testing

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #231, opened 2020-04-21
docs(components): Avatar component proposal

eddysims:
## Motivations

We are going to need an Avatar component soon. This is a proposal on how to best create one.

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/proposals/Avatar/README.md
+++ b/docs/proposals/Avatar/README.md
@@ -21,13 +21,13 @@ avatars in Jobber, as well as the potential usage of colors within avatars.
 
 ## Props Table
 
-| name     | type   | default  | description                               |
-| -------- | ------ | -------- | ----------------------------------------- |
-| image    | String | -        | The avatar image to be use                |
-| name     | String | -        | A users name to be used for accessibility |
-| initials | String | -        | Initials to show if the image is null     |
-| color    | String | greyBlue | The border color that represents the user |
-| size     | String | small    | Size of the avatar - 'small'              | 'large' |
+| name     | type   | default  | description                                              |
+| -------- | ------ | -------- | -------------------------------------------------------- |
+| image    | String | -        | The avatar image to be use                               |
+| name     | String | -        | A users name to be used for accessibility                |
+| initials | String | -        | Initials to show if the image is null                    |
+| color    | String | greyBlue | The background and border color that represents the user |
+| size     | String | small    | Size of the avatar - 'small'                             | 'large' |
 
 ## Display
 
@@ -51,14 +51,7 @@ cause issues with color contrast.
 
 ## Size
 
-There are 2 options that I can think of for showing size, numeric and string.
-
-- Numeric: `size` prop takes a number and that is the `px` width/height of the
-  avatar
-  - Example: `<Avatar size={32} />`
-  - This would allow more flexebility in usage of the component, but less strict
-    and less consistancy
-- String: `size` prop would take a string that represents the size of the
+- `String`: `size` prop would take a string that represents the size of the
   component
   - Example: `<Avatar size="large" />`
   - This would box us into limited sizes as once the standard XS,S,M,L,XL are
--- a/docs/proposals/Avatar/README.md
+++ b/docs/proposals/Avatar/README.md
@@ -1,0 +1,78 @@
+# Avatar
+
+An `<Avatar />` component will be used to display an identifier (image or
+initials) for a user.
+
+Rebecca Li has done some
+[great research](https://jobber.atlassian.net/browse/JOB-19028) on the usage of
+avatars in Jobber, as well as the potential usage of colors within avatars.
+
+## Interface
+
+```
+<Avatar
+    image="ImageUrl"
+    name="Eddy Sims"
+    initials="ES"
+    color="#8f2d56" | color="var(--color-red)"
+    size="large" | size={32}
+/>
+```
+
+## Props Table
+
+| name     | type   | default  | description                               |
+| -------- | ------ | -------- | ----------------------------------------- |
+| image    | String | -        | The avatar image to be use                |
+| name     | String | -        | A users name to be used for accessibility |
+| initials | String | -        | Initials to show if the image is null     |
+| color    | String | greyBlue | The border color that represents the user |
+| size     | String | small    | Size of the avatar - 'small'              | 'large' |
+
+## Display
+
+An `<Avatar />` component would have a heiarchy of display options
+
+1. Display the image used in the `image` prop.
+2. If `image` is null, display the `initials` prop.
+   - Initials should fit within the avatar and scale with the size.
+   - If initials contain 3 letters, the font size should scale to proplerly fit
+     this
+3. If `image` and `initials` are null. Display a fallback icon
+   - We could use `person` from https://atlantis.getjobber.com/components/icon
+
+## Color
+
+Can be used to represent the border color of the avatar if an image is set, or
+the background color of the avatar if the image is not set.
+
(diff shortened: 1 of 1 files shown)


eddysims:
Circle is going to fail until https://github.com/GetJobber/atlantis/pull/229 is merged unfortunatley.

docs/proposals/Avatar/README.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+## Props Table
+
+| name     | type   | default  | description                               |
+| -------- | ------ | -------- | ----------------------------------------- |
+| image    | String | -        | The avatar image to be use                |
+| name     | String | -        | A users name to be used for accessibility |
+| initials | String | -        | Initials to show if the image is null     |
+| color    | String | greyBlue | The border color that represents the user |
```
JeffMarvin, line 29: We may want to be more specific than "string" here - is it a string that can be evaluated in CSS as a color?

docs/proposals/Avatar/README.md
```diff
@@ -0,0 +1,78 @@
+# Avatar
+
+An `<Avatar />` component will be used to display an identifier (image or
+initials) for a user.
+
+Rebecca Li has done some
```
JeffMarvin, line 6: Some parts of that proposal are being implemented here, and some aren't. That's fine. But, any thoughts on how to keep that design document relevant as we continue to develop the component? Maybe link it in the actual component Readme?

docs/proposals/Avatar/README.md
```diff
@@ -0,0 +1,78 @@
...
+[great research](https://jobber.atlassian.net/browse/JOB-19028) on the usage of
+avatars in Jobber, as well as the potential usage of colors within avatars.
+
+## Interface
+
+```
+<Avatar
+    image="ImageUrl"
```
JeffMarvin, line 14: Part of me feels like we should be specific with this and call the prop `imageURL` to be clear. Thoughts?

docs/proposals/Avatar/README.md
```diff
@@ -0,0 +1,78 @@
...
+the background color of the avatar if the image is not set.
+
+One thing to note is that for accessibility purposes, color selection could
+cause issues with color contrast.
+
+## Size
+
+There are 2 options that I can think of for showing size, numeric and string.
+
+- Numeric: `size` prop takes a number and that is the `px` width/height of the
```
JeffMarvin, line 56: This is interesting, but ultimately more work and probably not needed. I think the string size matches better what we will need for designs.

JeffMarvin:
@eddysims I added some comments with a couple things that could be addressed.

docs/proposals/Avatar/README.md
chris-at-jobber: As well, I would suggest that the user's avatar container should receive the `color` prop as a background colour if there is no photo present

docs/proposals/Avatar/README.md
chris-at-jobber: Yeah let's be more opinionated about this to start. If teams utilizing this component start to find that they can't use the component with some form of S,M,L, that's probably a good time to revisit

docs/proposals/Avatar/README.md
```diff
@@ -0,0 +1,78 @@
...
+3. If `image` and `initials` are null. Display a fallback icon
+   - We could use `person` from https://atlantis.getjobber.com/components/icon
+
+## Color
+
+Can be used to represent the border color of the avatar if an image is set, or
+the background color of the avatar if the image is not set.
+
+One thing to note is that for accessibility purposes, color selection could
+cause issues with color contrast.
```
chris-at-jobber, line 50: Currently calendar colours receive a class that adjusts text colour to account for background color, but this is all hard-coded (eg `bg_purple` sets the text to be white because we know the background is purple). Not saying we need to solve at this component level, but could we do [something automated](https://contrast-ratio.com/) when a calendar colour is set to calculate contrast and choose black or white text accordingly? Would also be an interesting tool to have for the color picker component.

docs/proposals/Avatar/README.md
chris-at-jobber: That would be a good start. Could we add a ticket to the Frends backlog capturing the other components (avatarGroup, teamAssignmentChecklist)?

I'm also fairly certain the Manage Team work will require the List variation of this component from the proposal
image

docs/proposals/Avatar/README.md
chris-at-jobber: Also of note, the border colour is white by default, with a greyBlue background

docs/proposals/Avatar/README.md
seanhealy: Or just `url` could probably be fine? :)

docs/proposals/Avatar/README.md
eddysims: This is something we are considering, but maybe not at the first pass of this component. We do however need something to make sure contrast is appropriate as soon as we allow custom colors.

docs/proposals/Avatar/README.md
eddysims: Sounds good! I will remove the numeric option

docs/proposals/Avatar/README.md
eddysims: Does `url` make it feel like it can be clicked? I feel `imageUrl` is more clear then just `url`

docs/proposals/Avatar/README.md
eddysims: With Atlantis being open source, @chris-at-jobber can we link to the design documents from the Readme? Im also wondering if I even should have put this in the proposal now.

@chris-at-jobber in regards to the `single line` and `2 line` displays, I don't think that these are the responsibility of the avatar. I would say that the avatar is only responsible for the display of the initials and/or image.

docs/proposals/Avatar/README.md
eddysims: @chris-at-jobber I have addressed this (in the component). But the default state is kind of weird, in that the background is greyBlue and the border is white. 

Any other state, the background/border would be the same color.

eddysims:
@JeffMarvin @seanhealy @chris-at-jobber all comments addressed. Back to you guys with a few specific questions for @chris-at-jobber.

docs/proposals/Avatar/README.md
seanhealy: It's my understanding that linking a figma file should grant only access to that file. Since it's relevant to this I _think_ it's fine. Perhaps something we should think about further for the future though.

docs/proposals/Avatar/README.md
seanhealy: No strong feelings. I think `href` would make it feel clickable to me. `url` just feels simpler. But, happy whichever.

seanhealy:
@eddysims Left a little comment. Nothing serious. I think I'm happy about this from my perspective. 😄

docs/proposals/Avatar/README.md
chris-at-jobber: `imageUrl` feels more intuitive to me, for what it's worth

docs/proposals/Avatar/README.md
chris-at-jobber: Let me look into some Figma permissions things - it might require a more involved account setup for engineers than current, but it should be feasible to say "Anyone at Jobber can view"

chris-at-jobber:
No further comments from me at this time, my only note was about our Figma account setup but don't let that block this work from moving ahead

PR #235, opened 2020-04-23
feat(components): Avatar component

eddysims:
## Motivations

We are going to need an Avatar component soon. This adds an avatar component to Atlantis

### Added

- Avatar Component

## Testing

---

Random photo of Atlantis


packages/components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,104 @@
...
+function deconstructCssCustomProperty(color: string) {
+  if (!color) {
+    return;
+  }
+
+  if (!color.toLowerCase().startsWith("var(")) {
+    return color;
+  }
+
+  const value = color
```
seanhealy, line 96: A little comment breaking this down might be helpful?

packages/components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,104 @@
...
+    .replace("var(", "")
+    .slice(0, -1)
+    .split(" ")
+    .join("");
+  return getComputedStyle(document.documentElement)
```
seanhealy, line 101: Just my OCD. 😉 

```suggestion

  return getComputedStyle(document.documentElement)
```

packages/components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return (
+    <div
+      className={className}
+      style={style}
+      role={imageUrl && "img"}
+      aria-label={name}
+    >
+      {initials && !imageUrl ? <Initials initials={initials} /> : ""}
+      {!initials && !imageUrl ? (
+        <Icon name="person" color={color && isDark(color) ? "white" : "blue"} />
```
seanhealy, line 55: Color prop here might be clearer if we pre-compute it into an `iconColor` variable perhaps? The combination of the `&&` and the ternary I'm finding a bit hard to follow.

packages/components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  return (
+    <div
+      className={className}
+      style={style}
+      role={imageUrl && "img"}
+      aria-label={name}
+    >
+      {initials && !imageUrl ? <Initials initials={initials} /> : ""}
```
seanhealy, line 53: Want to take a moment to look at this `imageUrl` and `initials` logic together? I have an idea that may simplify it but may be easier on a screen share?

packages/components/src/Avatar/Avatar.mdx
```diff
@@ -0,0 +1,94 @@
...
+    name="The Jobbler"
+    size="large"
+    initials="JBLR"
+    color="var(--color-indigo)"
+  />
+</Playground>
+
+## Design & Usage Guidelines
+
+When contributing to, or consuming the Avatar component consider the
```
seanhealy, line 34: I think it would be good to inline some of the usage guidelines from the file here. When is it appropriate to use the component. What is the intention of color. 

Some ideas here: https://polaris.shopify.com/components/images-and-icons/avatar

packages/components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,84 @@
...
+import classnames from "classnames";
+import styles from "./Avatar.css";
+import { isDark } from "./utilities";
+import { Icon } from "../Icon";
+
+type AvatarSize = "base" | "large";
+
+interface AvatarProps {
+  /**
+   * A path to the image that will be displayed
```
seanhealy, line 11:
```suggestion
   * A url for the image that will be displayed
```

packages/components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,84 @@
...
+   */
+  readonly imageUrl?: string;
+  /**
+   * A users name to be used for assistive technology
+   */
+  readonly name?: string;
+  readonly initials?: string;
+  readonly color?: string;
```
seanhealy, line 19: Wanna add a little comment for these two?

packages/components/src/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+interface AvatarProps {
+  /**
+   * A path to the image that will be displayed
+   */
+  readonly imageUrl?: string;
+  /**
+   * A users name to be used for assistive technology
+   */
+  readonly name?: string;
```
seanhealy, line 17: Wonder if we should make name or initials required?

eddysims: I thought about this originally. The reason I didn't is:

- for name, this is used for accessibility and only is used if there is an `imageUrl` set
- for initials, they are not required if you set an `imageUrl`

This is also the reason that I left the user icon as a fallback.

Thoughts?

packages/components/src/Avatar/Avatar.css
```diff
@@ -0,0 +1,44 @@
...
+  --avatar-size: 4.5rem;
+  --avatar-border-size: 4px;
+}
+
+.isDark {
+  color: var(--color-white);
+}
+
+.initials {
+  font-size: 0.5em;
```
seanhealy, line 30: I don't know if it's possible but, it might be nice to tie the fixed sizes in here to a preset var. If not we should probably scale them based on [`--base-unit`](https://github.com/GetJobber/atlantis/blob/master/packages/design/src/spacing.css#L2).

packages/components/src/Avatar/Avatar.css
```diff
@@ -0,0 +1,44 @@
...
+  color: var(--color-blue--dark);
+  font-size: var(--avatar-size);
+  -webkit-font-smoothing: antialiased;
+  background: no-repeat center center var(--color-greyBlue);
+  background-size: cover;
+}
+
+.large {
+  --avatar-size: 4.5rem;
+  --avatar-border-size: 4px;
```
seanhealy, line 22: https://github.com/GetJobber/atlantis/blob/master/packages/design/src/borders.css

packages/components/src/Avatar/Avatar.mdx
seanhealy: Like this!

Let's pull the figma link out for the moment. We should do some circling around on when is appropriate to link into other things. 😄

packages/components/src/Avatar/Avatar.tsx
seanhealy: I see what you're saying. We can do some type fanciness in here then to ensure that if we have a `url` then initials aren't required.

Eg.
https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Button/Button.tsx#L25-L43

Also happy to pair on this one if you wish. :)

PR #237, opened 2020-04-24
docs(components): Add design and usage guidelines to tooltip

chris-at-jobber:
## Motivation

Make it easier to know when and why to use tooltips

## Changes

There are design & usage guidelines now

### Added

There are design & usage guidelines now

### Changed

tweaked intro copy

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.mdx
+++ b/packages/components/src/Tooltip/Tooltip.mdx
@@ -14,7 +14,7 @@ import { Card } from "../Card";
 
 <ComponentStatus stage="rc" responsive="yes" accessible="yes" />
 
-Tooltips are extra information that briefly explain the function of a UI
+Tooltips provide in-context information that briefly explain the function of a UI
 element.
 
 `Tooltip` will apply a tooltip to a single wrapped Element.
@@ -31,6 +31,19 @@ import { Tooltip } from "@jobber/components/Tooltip";
     <Button label="Hover On Me Too" />
   </Tooltip>
 </Playground>
+
+## Design & Usage Guidelines
+
+When contributing to, or consuming the Tooltip component, consider the following:
+
+### Usage
+- A tooltip should be used to introduce a novel UI element or highlight a change to an existing interface.
+- A tooltip should contain information that is helpful, but not necessary for the user to successfully interact with the page.
+- A tooltip should not be a permanent fixture in the interface - if you need a message that remains in the view, consider putting that content inline. 
+
+### Content
+- Tooltip content should consist of only text and an optional "dismiss" CTA
+- Tooltip text content should not exceed 3 lines
 
 ## Props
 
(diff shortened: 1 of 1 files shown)


packages/components/src/Tooltip/Tooltip.mdx
```diff
@@ -32,6 +32,19 @@ import { Tooltip } from "@jobber/components/Tooltip";
...
 </Playground>
 
+## Design & Usage Guidelines
+
+When contributing to, or consuming the Tooltip component, consider the following:
+
+### Usage
+- A tooltip should be used to introduce a novel UI element or highlight a change to an existing interface.
+- A tooltip should contain information that is helpful, but not necessary for the user to successfully interact with the page.
+- A tooltip should not be a permanent fixture in the interface - if you need a message that remains in the view, consider putting that content inline.
```
agglasaur, line 42: This last one I'm unclear on. What do you mean a permanent fixture? A lot of tooltips are permanent in the interface

packages/components/src/Tooltip/Tooltip.mdx
```diff
@@ -32,6 +32,19 @@ import { Tooltip } from "@jobber/components/Tooltip";
...
+When contributing to, or consuming the Tooltip component, consider the following:
+
+### Usage
+- A tooltip should be used to introduce a novel UI element or highlight a change to an existing interface.
+- A tooltip should contain information that is helpful, but not necessary for the user to successfully interact with the page.
+- A tooltip should not be a permanent fixture in the interface - if you need a message that remains in the view, consider putting that content inline. 
+
+### Content
+- Tooltip content should consist of only text and an optional "dismiss" CTA
+- Tooltip text content should not exceed 3 lines
```
agglasaur, line 46: ty for specifying length, we've got some long bois out there 🙏

packages/components/src/Tooltip/Tooltip.mdx
chris-at-jobber: I think that's more my ideal state going forward and less a reflection of current state... If something is hidden in a tooltip and we're just using that to save space it's probably not a great UI solution?

packages/components/src/Tooltip/Tooltip.mdx
chris-at-jobber: l o o o n g b o i s

packages/components/src/Tooltip/Tooltip.mdx
agglasaur: I think there may always be supplementary information that are just not necessarily a problem of the UI (see: pricing grid and its million tooltips). Sure there may be other ways to handle that, but I do think tooltips have their place with permanence in a UI

chris-at-jobber: Yeah that's a fair point

There's also the idea of a tooltip as a bonus button label (which we don't do but could be kind of nice in the future) that systems like Material do pretty well

I think maybe the previous line about "helpful, but not necessary" covers what I was trying to say here

PR #238, opened 2020-04-24
docs(components): Update Switch design docs

chris-at-jobber:
## Motivations

Improve Switch design docs, especially since Radio Button is likely on the near horizon and the usage guidelines provide some disambiguation.

## Changes

Adding design guidelines, tweaking some others.

### Added

Design guidelines

### Changed

Content around Controlled and Disabled variants

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Switch/Switch.mdx
+++ b/packages/components/src/Switch/Switch.mdx
@@ -29,7 +29,17 @@ import { Switch } from "@jobber/components/Switch";
 
 ---
 
+## Design & Usage Guidelines
+
+The switch is a useful component for manipulating settings with binary (true/false, yes/no, on/off) conditions. For this reason its' label is limited to ON/OFF.
+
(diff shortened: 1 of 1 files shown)


packages/components/src/Switch/Switch.mdx
```diff
@@ -51,6 +61,8 @@ import { Switch } from "@jobber/components/Switch";
 
 ## Disabled
 
+A disabled switch cannot be operated by the user. If presenting a disabled "On" switch to the user, the context and ability for the user to enable the switch must be apparent to avoid creating a sense that the user has lost control of the interface.
```
agglasaur, line 64: I would just adjust the copy to be a bit less clinical sounding: instead of "the context and ability for the user to enable the switch must be apparent" can we just simply say "Provide clear description for the user on how to enable the switch."

PR #242, opened 2020-04-28
feat: add hook package with resize oberserver hook

JeffMarvin:
## Motivations

This create a new package in Atlantis `hooks` - to contain reusable and importable hooks for the Atlantis library.
It also adds a reusable hook `useResizeObserver` - based on [this spike code](https://github.com/GetJobber/atlantis/pull/228/files) 

### Added

- New `hooks` package.
- New `useResizeObserver` hook exported from the new `hooks` package.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

* Navigate to [/components/content](https://deploy-preview-242--jobber-atlantis.netlify.app/components/content) when running this branch.
* observe that the Content component background changes depending on how wide the component is.
* Navigate to new documentation pages:
  * https://deploy-preview-242--jobber-atlantis.netlify.app/hooks/use-resize-observer
  * https://deploy-preview-242--jobber-atlantis.netlify.app/hooks
---

Random photo of Atlantis


JeffMarvin:
This PR contains test code (modifications to the Content component) that should not ship.
It is marked as BLOCKED until that commit is removed.

packages/hooks/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "dist/*"
+  ],
+  "devDependencies": {
+    "typescript": "^3.8.3"
+  },
+  "dependencies": {
+    "resize-observer-polyfill": "^1.5.1"
+  },
+  "peerDependencies": {
+    "react": "^16.8.6"
```
eddysims, line 22: will we need `@types/react` as a peer as well to handle our typescript?

packages/hooks/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./useResizeObserver";
```
eddysims, line 1: Currently we have this at the root and exporting everything. I see that this is how we export in `@jobber/formatter` but not in `@jobber/components`. My assumption is that this is fine and that we will just export everything from the root. Just want to make sure that this is intentional and that we dont want to do something like:

```js
import { useResizeObserver } from "@jobber/hooks/useResizeObserver"
```

packages/hooks/src/useResizeObserver.ts
```diff
@@ -0,0 +1,41 @@
+import { useCallback, useLayoutEffect, useRef, useState } from "react";
+import ResizeObserver from "resize-observer-polyfill";
+
+export function useResizeObserver() {
+  const [entry, setEntry] = useState<ResizeObserverEntry>();
+  const [node, setNode] = useState() as any;
```
eddysims, line 6: I feel that this `any` type needs to be resolved. This is where I got hung up in the spike was trying to figure out the types for this.

packages/components/src/Content/Content.tsx
```diff
@@ -14,7 +15,20 @@ interface ContentProps {
 }
 
 export function Content({ children, spacing = "base" }: ContentProps) {
-  const className = classnames(styles.padded, spacings[spacing]);
+  const [ref, result] = useResizeObserver();
```
eddysims, line 18: I think for the documentation we should write this as 

```js
const [ref, {width, height}] =  useResizeObserver()
```

eddysims:
@JeffMarvin a few comments here. Back to you.

packages/hooks/package.json
JeffMarvin: It seemed ok without it...

packages/hooks/src/index.ts
JeffMarvin: Yeah, that's a fair question. @seanhealy Any advice on which of the two organizing structures we want here?

packages/components/src/Content/Content.tsx
JeffMarvin: Noted. Although, this is going to get pulled out of the PR when it ships.
Thoughts on where we should document the hook useage?

packages/hooks/src/useResizeObserver.ts
JeffMarvin: Ok, I'll take a look here.

packages/components/src/Content/Content.tsx
eddysims: I think that docz should have a page/section for hooks.

packages/hooks/src/useResizeObserver.ts
JeffMarvin: @seanhealy I could use your input on this question about `any` as well. I looked but don't know how to resolve it.

packages/components/src/Content/Content.tsx
JeffMarvin: Good call - I added the hooks package to the docz site, so that it pulls the Readme. I'll add this as an example in that Readme.

packages/hooks/src/index.ts
JeffMarvin: We could also ship this way, and iterate. Right now, making the switch wouldn't be hard.

JeffMarvin:
@eddysims When the latest changes finish building on Netlify, check the preview. I added more info around the Hooks in the docs menu (under hooks and under packages).

packages/hooks/src/index.ts
seanhealy: Yea, I'm not sure on this one either. I can see the upsides of both ways…

seanhealy: After talking with Jeff I think we're better off leaving this as is for now. the `hooks/foo` syntax requires a bit of fiddly build time configuration that I don't think we'll really benefit from doing right now.

packages/hooks/src/useResizeObserver.ts
JeffMarvin: @seanhealy and I went through this and he figured out a solution. That's been committed in the last few commits.

PR #244, opened 2020-04-28
feat(components): Add primaryAction and placement to Banner

eddysims:
## Motivations

We will soon need an action in the `Banner` component. This PR takes inspiration from the `Modal` component and adds a button to the `Banner` component.

We are also going to need a site-wide version of the `Banner`. This `Banner` should not be dismissible. This PR adds in the `placement` prop that will make the `Banner` dismissible.

### Added

- Adds `primaryAction` prop to `Banner` component
- Adds `placement` prop to `Banner` component

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Banner/Banner.mdx
```diff
@@ -20,7 +20,15 @@ import { Banner } from "@jobber/components/Banner";
...
-  <Banner type="success">You've successfully Hulked out</Banner>
+  <Banner
+    type="success"
+    primaryAction={{
+      label: "Poke Me",
+      onClick: () => alert("💥 SMASH 💥")
+    }}
+  >
+    You've successfully Hulked out
+  </Banner>
```
JeffMarvin, line 31: I wonder if it is worth being more explicit in the docs here about using this pattern, vs. putting buttons in as child components.

packages/components/src/Banner/Banner.tsx
```diff
@@ -4,18 +4,34 @@ import styles from "./Banner.css";
...
   readonly type: "notice" | "success" | "warning" | "error";
+  readonly primaryAction?: ButtonProps;
+  /**
+   * Determines the presentation style and some options for the banner. A site
+   * banner is not dismissable and appears at the top of the site layout. A
+   * page banner can be dismissed and appears inline at the top of the page.
+   *
+   * @default "page"
+   */
+  readonly placement?: "page" | "site";
```
JeffMarvin, line 20: I like this distinction, but this whole prop might be moot if BizOps just finished building a sitewide banner. We might want to leave this prop out for now...

packages/components/src/Banner/Banner.mdx
JeffMarvin: It might also be worth it to have a dedicated example section for our new case, instead of just adding to the default. You could talk in the new section about the button props and what that means (see the Modal component for examples).

packages/components/src/Banner/Banner.test.tsx
```diff
@@ -145,3 +41,65 @@ test("it should call the handler with a number value", () => {
...
+        label: "Hello World",
+        onClick: () => onClick(),
+      }}
+    >
+      Bruce
+    </Banner>,
+  );
+
+  fireEvent.click(getByText("Hello World"));
+  expect(onClick).toHaveBeenCalledTimes(1);
```
JeffMarvin, line 93: 👍

packages/components/src/Banner/Banner.tsx
```diff
@@ -25,20 +41,34 @@ export function Banner({ children, type, onDismiss }: BannerProps) {
     error: "red",
   };
 
+  if (primaryAction != undefined) {
+    primaryAction = Object.assign(
+      {
+        size: "small",
+        type: "tertiary",
+        variation: type === "notice" ? "learning" : "work",
```
JeffMarvin, line 49: What is this setting exactly?

packages/components/src/Banner/Banner.tsx
JeffMarvin: Yeah, the more I think about this, its a really good idea from design. But, we're not there yet in being able to fully implement it. I.e. just because we put this prop on this banner doesn't mean people start using it instead of building their own. We need to address the problem of multiple banners before we implement it here with a prop.

packages/components/src/Banner/Banner.tsx
eddysims: Added a description to this variable. Do you think that will suffice?

packages/components/src/Banner/Banner.mdx
```diff
@@ -54,3 +54,32 @@ import { Banner } from "@jobber/components/Banner";
...
+      label: "Poke Me",
+      onClick: () => alert("💥💥SMASH💥💥")
+    }}
+  >
+    You wouldn't like me when im angry.
+  </Banner>
+  <Banner
+    type="notice"
+    primaryAction={{
+      label: "Im always angry",
```
seanhealy, line 79:
```suggestion
      label: "I'm always angry",
```

packages/components/src/Banner/Banner.mdx
```diff
@@ -54,3 +54,32 @@ import { Banner } from "@jobber/components/Banner";
 <Playground>
   <Banner type="error">Hulk smash has been unsuccessful</Banner>
 </Playground>
+
+## Actions in Banners
+
+If you require an action in a `Banner`, use the `primaryActions` prop. The
+`primaryAction` accepts the same props as the `Button` component.
+
+It is discouraged to use a `<Button />` component when requiring an action
```
seanhealy, line 63: Don't quite follow this line?

packages/components/src/Banner/Banner.mdx
```diff
@@ -54,3 +54,32 @@ import { Banner } from "@jobber/components/Banner";
...
+
+<Playground>
+  <Banner
+    type="success"
+    primaryAction={{
+      label: "Poke Me",
+      onClick: () => alert("💥💥SMASH💥💥")
+    }}
+  >
+    You wouldn't like me when im angry.
```
seanhealy, line 74:
```suggestion
    You wouldn't like me when I'm angry.
```

packages/components/src/Banner/Banner.tsx
seanhealy: Might be better to add this as a documentation on the `type` prop so that you can see it in the docs viewer as well?

eddysims: I have added it to the documentation of `primaryAction` I don't think that type is the right place for it as it doesn't change anything about type.

PR #245, opened 2020-05-03
docs(components): Update card docs with design usage guidelines

chris-at-jobber:
## Motivations

Want to make the "why" behind a card more apparent

## Changes

Based on Jobber design team feedback, expanded upon card guidelines in foundational design terms.

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Card/Card.mdx
+++ b/packages/components/src/Card/Card.mdx
@@ -14,8 +14,8 @@ import { Content } from "../Content";
 
 <ComponentStatus stage="ready" responsive="yes" accessible="yes" />
 
-A card is used to group related information and tasks to make things easier to
-scan, and action upon
+A card is used to group related information and tasks so our users can scan 
+and prioritize information more easily.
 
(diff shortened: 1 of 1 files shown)


packages/components/src/Card/Card.mdx
```diff
@@ -40,6 +40,12 @@ import { Card } from "@jobber/components/Card";
   </Card>
 </Playground>
 
+## Design & Usage Guidelines
+
+The primary utility of a card is its' grouping of elements by proximity within a clear boundary, clearly marked by a header.
```
agglasaur, line 45: "grouping of elements by proximity within a clear boundary" sounds really technical and I don't think would help me make design decisions. What's the actual goal of a card? To group similar tasks so our users can scan and prioritize information more easily, or something like that?

chris-at-jobber: I tweaked the component description and this part, I think now it better gets at what I'm trying to say?

PR #247, opened 2020-05-03
docs(components): update definition/add guidelines to content

chris-at-jobber:
## Motivations

I've found Content to be a confusing component, as it's an invisible component that "auto-magically" changes the appearance of any children - trying to update the docs to better explain what it is and when to use it.

## Changes

Changed top description, added design guidelines

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Content/Content.mdx
+++ b/packages/components/src/Content/Content.mdx
@@ -52,7 +52,7 @@ import { Content } from "@jobber/components/Content";
 
 ## Design Usage & Guidelines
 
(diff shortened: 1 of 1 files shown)


packages/components/src/Content/Content.mdx
```diff
@@ -50,6 +50,10 @@ import { Content } from "@jobber/components/Content";
   </Content>
 </Playground>
 
+## Design Usage & Guidelines
+
+Content should be used whenever you have a container with mixed content that requires consistent spacing between the elements within said container.
```
agglasaur, line 55: I would just remove the "within said container" - it's implied

PR #248, opened 2020-05-04
feat(components): add action buttons to Page component 

JeffMarvin:
This adds primary and secondary action buttons and a menu as optional parts of the `<Page>` component. This will render them in the standard action button and action menu way. 

## Motivations
- This uses `ButtonProps` for flexibility and consistency across other places in the library.
- The CSS and markup are contained in the Page for now. In future work, we can bring it out into an ActionBar component.

### Added

- `<Page>` can now take a `primaryAction` (`ButtonProps` type), `secondaryAction`  (`ButtonProps` type) and an `actionMenu` prop (`SectionProps[]` type).

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Page/Page.css
+++ b/packages/components/src/Page/Page.css
@@ -16,21 +16,11 @@
   max-width: calc(var(--base-unit) * 64);
 }
 
-.actionGroup {
-  text-align: right;
-  vertical-align: top;
+.titleBar {
+  display: flex;
+  justify-content: space-between;
 }
 
 .actionGroup > :not(:last-child) {
   margin: 0 var(--space-small) 0 0;
-}
-
-.titleBar {
-  display: flex;
-  border-top: 1px solid red;
-}
-
-.titleBar > * {
-  flex: auto;
-  justify-content: space-between;
 }
--- a/packages/components/src/Page/Page.css
+++ b/packages/components/src/Page/Page.css
@@ -22,5 +22,6 @@
 }
 
 .actionGroup > :not(:last-child) {
-  margin: 0 var(--space-small) 0 0;
+  margin: 0;
+  margin-right: var(--space-small);
 }
--- a/packages/components/src/Page/Page.css
+++ b/packages/components/src/Page/Page.css
@@ -15,3 +15,22 @@
 .narrow {
   max-width: calc(var(--base-unit) * 64);
 }
+
+.actionGroup {
+  text-align: right;
+  vertical-align: top;
+}
+
+.actionGroup > :not(:last-child) {
+  margin: 0 var(--space-small) 0 0;
+}
+
+.titleBar {
+  display: flex;
+  border-top: 1px solid red;
+}
+
+.titleBar > * {
+  flex: auto;
+  justify-content: space-between;
+}
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -42,6 +42,7 @@ export interface PageProps {
 
   /**
    * Page title secondary action button settings.
+   *   Only shown if there is a primaryAction.
    */
   readonly secondaryAction?: ButtonProps;
 
@@ -64,6 +65,10 @@ export function Page({
 
   const showMenu = moreActionsMenu.length > 0;
   const showActionGroup = showMenu || primaryAction;
+
+  if (secondaryAction != undefined) {
+    secondaryAction = Object.assign({ type: "secondary" }, secondaryAction);
+  }
 
   return (
     <div className={pageStyles}>
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -6,7 +6,7 @@ import { Text } from "../Text";
 import { Content } from "../Content";
 import { Markdown } from "../Markdown";
 
-interface PageProps {
+export interface PageProps {
   readonly children: ReactNode | ReactNode[];
 
   /**
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -5,6 +5,8 @@ import { Heading } from "../Heading";
 import { Text } from "../Text";
 import { Content } from "../Content";
 import { Markdown } from "../Markdown";
+import { Button, ButtonProps } from "../Button";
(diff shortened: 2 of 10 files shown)


packages/components/src/Page/Page.tsx
```diff
@@ -32,21 +34,51 @@ interface PageProps {
...
 }
 
 export function Page({
   title,
   intro,
   children,
   width = "standard",
+  primaryAction,
+  secondaryAction,
+  moreActionsMenu,
```
seanhealy, line 61:
```suggestion
  moreActionsMenu = [],
```

packages/components/src/Page/Page.tsx
```diff
@@ -32,21 +34,51 @@ interface PageProps {
...
 }: PageProps) {
-  const className = classnames(styles.page, styles[width]);
+  const pageStyles = classnames(styles.page, styles[width]);
+
+  const showMenu = moreActionsMenu && moreActionsMenu.length > 0;
```
seanhealy, line 65:
```suggestion
  const showMenu = moreActionsMenu.length > 0;
```

seanhealy: Think these should fix your build error. :)

packages/components/src/Page/Page.tsx
JeffMarvin: Thanks, that makes so much sense. Such a clean way to handle that not being undefined. :)

packages/components/src/Page/Page.tsx
```diff
@@ -32,21 +34,51 @@ interface PageProps {
...
+
+  /**
+   * Page title primary action button settings.
+   */
+  readonly primaryAction?: ButtonProps;
+
+  /**
+   * Page title secondary action button settings.
+   */
+  readonly secondaryAction?: ButtonProps;
```
seanhealy, line 46: @chris-at-jobber Do you think we should provide a default button `type` for secondary?

packages/components/src/Page/Page.css
```diff
@@ -15,3 +15,12 @@
...
   max-width: calc(var(--base-unit) * 64);
 }
+
+.titleBar {
+  display: flex;
+  justify-content: space-between;
+}
+
+.actionGroup > :not(:last-child) {
+  margin: 0 var(--space-small) 0 0;
```
seanhealy, line 25: Might be easier to read if you specify…

```suggestion
.actionGroup > * {
  margin-left: var(--space-small);
```

Also, I _think_ this will allow you to avoid the `:last-child` too…

packages/components/src/Page/Page.tsx
```diff
@@ -32,21 +34,51 @@ interface PageProps {
...
 
   return (
-    <div className={className}>
+    <div className={pageStyles}>
       <Content>
         <Content spacing="large">
-          <Heading level={1}>{title}</Heading>
+          <div className={styles.titleBar}>
+            <Heading level={1}>{title}</Heading>
+            {showActionGroup && (
```
seanhealy, line 74: Do we still need this condition here?

packages/components/src/Page/Page.tsx
```diff
@@ -32,21 +34,51 @@ interface PageProps {
...
   width = "standard",
+  primaryAction,
+  secondaryAction,
+  moreActionsMenu = [],
 }: PageProps) {
-  const className = classnames(styles.page, styles[width]);
+  const pageStyles = classnames(styles.page, styles[width]);
+
+  const showMenu = moreActionsMenu.length > 0;
+  const showActionGroup = showMenu || primaryAction;
```
seanhealy, line 66:
```suggestion
  const showActionGroup = showMenu || primaryAction || secondaryAction;
```

packages/components/src/Page/Page.css
JeffMarvin: True, but then I am not setting the margin on the other sides... which seems less robust.

packages/components/src/Page/Page.tsx
JeffMarvin: Yup. We still could have a button or a menu.

JeffMarvin:
@seanhealy couple of answers. I agree we should see what @chris-at-jobber thinks about the type of the default button. But... that's also an easy follow-up PR to do.

packages/components/src/Page/Page.tsx
chris-at-jobber: I don't think I understand what you're asking, `secondary` is a` type` as far as I am aware

chris-at-jobber: If the question is "should the secondary button have a `type` of `secondary`?" then yes 😄

packages/components/src/Page/Page.tsx
chris-at-jobber: That makes sense to me

packages/components/src/Page/Page.tsx
seanhealy: Sorry, what I was meaning is that we're doing these checks lower down and if those are all false the CSS should collapse everything.

packages/components/src/Page/Page.tsx
seanhealy: As discussed would be good to add a note to the props to this effect.

packages/components/src/Page/Page.css
JeffMarvin: Discussed in person, going to use the verbose properties.

packages/components/src/Page/Page.tsx
JeffMarvin: Keeping it to avoid adding an empty div into the page.

JeffMarvin:
@seanhealy Back to you for a re-review.

eddysims:
LGTM 👍 

One thought about something we may want to tackle some time is the idea of a `<ButtonGroup />` component. 

We have this type of pattern in a few places in Atlantis:

```
{primaryAction && <Button {...primaryAction} />}
{secondaryAction && <Button {...secondaryAction} />}
```

Might be nice in the future to look at componentizing this.

PR #250, opened 2020-05-05
feat(components): LightBox

linds14sr20det:
## Motivations

WWR's work on adding images to line items requires a way to display the images. We wanted to use a lightbox to do so, and there are a plethora of good open source ligthboxes. We chose the frontend-collective's` react-image-lightbox` to do this. As such we provide a wrapper component to interface with this third party library.

## Changes

- See added.

### Added

- Added a wrapper component for frontend-collective's lightbox.

### Changed

-

### Deprecated

- 

### Removed

- 

### Fixed

- 
### Security

- 

## Testing

We rely on frontend-collective's unit tests for fine grained testing.
- Clicking the button opens a lightbox
- Clicking next and previous buttons rotate through images
- Captions and titles display as expected
- Closing the lightbox actually closes the lightbox

---

Random photo of Atlantis


packages/components/src/PasswordEditor/__snapshots__/PasswordEditor.test.tsx.snap
```diff
@@ -0,0 +1,42 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
seanhealy, line 1: Why is this in here?

packages/components/src/LightBox/LightBox.mdx
```diff
@@ -0,0 +1,74 @@
...
+          onRequestClose={() => isOpen(false)}
+        />
+      </>
+    );
+  }}
+</Playground>
+
+## Design & Usage Guidelines
+
+When contributing to, or consuming the LightBox component, consider the
```
seanhealy, line 58: This section is meant to be more usage oriented a good example of such would be the Best Practices / Content Guidelines here:

https://polaris.shopify.com/components/overlays/popover#navigation

packages/components/src/LightBox/LightBox.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+          title: title,
+          caption: caption,
+          url: "https://i.imgur.com/6Jcfgnp.jpg",
+        },
+      ]}
+      onRequestClose={handleClose}
+    />,
+  );
+
+  expect(queryByText(title)).toBeTruthy();
```
seanhealy, line 26:
```suggestion
  expect(queryByText(title)).toBeInstanceOf(HTMLElement);
```

seanhealy: Might not be a bad idea too to do a quick test that the `handleClose` gets called properly as well.

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from "react";
+import ExternalLightBox from "react-image-lightbox";
+
+// Library requires fetching it's CSS.
```
seanhealy, line 4: I may have written this but… 😉 

```suggestion
// Library requires fetching its CSS.
```

docs/proposals/LightBox/README.md
```diff
@@ -0,0 +1,72 @@
...
+| -------- | ------ | -------- | -------------------------------------------------------- |
+| image    | String | -        | The avatar image to be use                               |
+| name     | String | -        | A users name to be used for accessibility                |
+| initials | String | -        | Initials to show if the image is null                    |
+| color    | String | greyBlue | The background and border color that represents the user |
+| size     | String | small    | Size of the avatar - 'small'                             | 'large' |
+
+## Display
+
+An `<Avatar />` component would have a heiarchy of display options
```
seanhealy, line 35: Not sure if we need this file. It also mostly talks about the `Avatar`. 😄

linds14sr20det: Ah crap, forgot to delete this.

packages/components/src/LightBox/LightBox.test.tsx
linds14sr20det: Yeah this makes more sense. This is me still being a newb in this test library. I originally had `expect(queryByText(title)).toHaveText(title);` but that felt redundant. I then went through `modal.test.tsx` and this is how we do this test there FYI.

whereyoubinns:
@linds14sr20det when there is only one image it still displays the left and right arrows, can we make it accept either an array of images or just one image which doesn't show the arrows??

packages/components/src/LightBox/LightBox.css
```diff
@@ -0,0 +1,20 @@
...
+  content: "";
+  display: inline-block;
+  width: 100%;
+  height: 100%;
+  background-color: var(--color-greyBlue);
+  opacity: 0.8;
+}
+
+.wrapper:global(.ril__outer .ril__captionContent) {
+  color: rgb(193, 201, 206);
```
seanhealy, line 15: Can we use a color preset for these?

https://atlantis.getjobber.com/colors

packages/components/src/LightBox/LightBox.mdx
```diff
@@ -0,0 +1,73 @@
...
+          onRequestClose={() => isOpen(false)}
+        />
+      </>
+    );
+  }}
+</Playground>
+
+## Design & Usage Guidelines
+
+A lightbox should:
```
seanhealy, line 58: @chris-at-jobber Wouldn't mind you taking a quick look at the guides here if you have a moment? 😄

packages/components/src/LightBox/LightBox.test.tsx
```diff
@@ -0,0 +1,51 @@
+import React from "react";
+import { cleanup, render } from "@testing-library/react";
+import { LightBox } from ".";
+
+afterEach(cleanup);
```
seanhealy, line 5: A test in here confirming that the `onRequestClose` gets fired when the user clicks close would be good.

packages/components/src/LightBox/LightBox.mdx
```diff
@@ -0,0 +1,73 @@
...
+import { Button } from "../Button";
+import { LightBox } from ".";
+
+# Light Box
+
+<ComponentStatus stage="pre" responsive="no" accessible="no" />
+
+LightBox is a component designed to display an array of images to the user. A
+user clicks on a control and is shown a carousel style presentation of the
+defined images. This component is a wrapper of the frontend-collective's
```
chris-at-jobber, line 19: :"This component is a wrapper of..." sentence could be its' own line, it doesn't feel conceptually related to the other sentences in this paragraph. Should we link out to the frontend-collective as well?

packages/components/src/LightBox/LightBox.mdx
chris-at-jobber: I'd ideally like to see a bit more direction around the actual use case for it, this feels a bit more like the rationale for why it looks the way it does. @jlelford when you're back you can add any thoughts you might have, maybe something like 

"A lightbox should be used when viewing an image is the user's primary goal" so that we don't end up having the lightbox used for things like in-product upgrade prompts? A bit of disambiguation around when a lightbox vs a modal that contains an image, or something?

chris-at-jobber: To be clear, I also don't want to block this from going into Atlantis on doc changes

seanhealy:
Had to run to planning / retro will circle back on this as soon as I can.

linds14sr20det:
@seanhealy wanna take another look? I think this is good to go now unless you have any further comments.

packages/components/src/LightBox/LightBox.mdx
linds14sr20det: Made some changes here to better reflect your comments.

chris-at-jobber: Just got this from Josh, I think this encapsulates it pretty well:
>A lightbox's primary goal is to allow users to see a higher resolution view of images contained within the page they're viewing. The application of a lightbox facilitates smaller thumbnail images to be used in an interface without compromising the detail within.
The lightbox should only be used when displaying an image at a greater scale is the primary function of opening it. Additional content types would be better suited by using a dialogue or alternative method.

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -15,15 +15,21 @@ interface PresentedImage {
...
   readonly open: boolean;
   readonly images: PresentedImage[];
+  readonly imageIndex?: number;
   onRequestClose(): void;
 }
 
 export function LightBox({
   open = false,
   images,
+  imageIndex,
```
whereyoubinns, line 25:
```suggestion
  imageIndex = 0,
```
I think if you just set a default here you we can skip the conditional in 28-31 and just `const [currentImageIndex, setCurrentImageIndex] = useState(imageIndex);` on line 32

whereyoubinns, line 25:
```suggestion
  imageIndex = 0,
```
I think if you just set a default here you we can skip the conditional in 28-31 and just `const [currentImageIndex, setCurrentImageIndex] = useState(imageIndex);` on line 32

linds14sr20det: Oh yep, totally missed that!

packages/components/src/LightBox/LightBox.mdx
```diff
@@ -0,0 +1,95 @@
...
+<ComponentStatus stage="pre" responsive="no" accessible="no" />
+
+LightBox is a component designed to display an array of images to the user. A
+user clicks on a control and is shown a carousel style presentation of the
+defined images.
+
+This component is a wrapper of the frontend-collective's react-image-lightbox
+third party library.
+<a href="https://github.com/frontend-collective/react-image-lightbox">React
+Image Lightbox</a>
```
seanhealy, line 24: This comment might be better fit for inside the `Lightbox.tsx` file as the consumer of the component doesn't really need to know this. 😄

packages/components/src/LightBox/LightBox.mdx
seanhealy: This is still outstanding.

packages/components/src/LightBox/LightBox.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+Use this to specify which image in `images` to initialize the lightbox with.
+This is useful when you have a collection of thumbnails as you only need one
+collection of image urls, order doesn't matter.
+
+## onRequestClose
+
+This function must set open to false in order to close the lightbox. Note there
+is a 300ms easing animation on lightbox close that occurs before this function
+is called.
```
seanhealy, line 95: These comments should be moved in with the props in the `tsx` file. Then they will properly show up in the props table.

eg: https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Avatar/Avatar.tsx#L9-L31

image

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+interface LightBoxProps {
+  readonly open: boolean;
+  readonly images: PresentedImage[];
+  readonly imageIndex?: number;
+  onRequestClose(): void;
+}
+
+export function LightBox({
+  open = false,
```
seanhealy, line 23: `open` is required so we don't really need a default here.

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -0,0 +1,61 @@
...
+}: LightBoxProps) {
+  const [currentImageIndex, setCurrentImageIndex] = useState(imageIndex);
+  let nextSrc = undefined;
+  let prevSrc = undefined;
+  if (images.length > 1) {
+    nextSrc = images[(currentImageIndex + 1) % images.length].url;
+    prevSrc =
+      images[(currentImageIndex + images.length - 1) % images.length].url;
+  }
+
```
seanhealy, line 36: We should probably have a `useEffect` in here to watch `imageIndex` incase it's externally updated so we can update the internal `currentImageIndex`.

Roughly…

```tsx
useEffect(() => {
  setCurrentImageIndex(imageIndex);
}, [imageIndex]);
```

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -0,0 +1,61 @@
...
+          nextSrc={nextSrc}
+          prevSrc={prevSrc}
+          imageTitle={images[currentImageIndex].title}
+          imageCaption={images[currentImageIndex].caption}
+          onCloseRequest={onRequestClose}
+          onMovePrevRequest={() =>
+            setCurrentImageIndex(
+              (currentImageIndex + images.length - 1) % images.length,
+            )
+          }
```
seanhealy, line 53: It's generally not great to declare handlers inline like this. Would probably be better to declare a `function handleMovePrevious()` after the return value and then pass it in here instead.

```suggestion
          onMovePrevRequest={handleMovePrevious}
```

When you inline handlers causes them to be alloced and destroyed with every component render which can have a performance hit.

packages/components/src/LightBox/LightBox.mdx
linds14sr20det: I disagree, I think having it here is useful for devs and designers. There is a lot of out of the box functionality that react-image-lightbox offers that may not be apparent otherwise. If a team is aware that a feature is potentially just a prop away vs. writing that whole feature we could save some work/headaches.

PR #251, opened 2020-05-06
feat(components): ProgressBar can have a size

whereyoubinns:
## Motivations

The progress bar was only one size that was too large when using it in smaller spaces. 

### Added

ProgressBar can have a small size, which is half the size of the default ProgressBar. 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/ProgressBar/ProgressBar.css
+++ b/packages/components/src/ProgressBar/ProgressBar.css
@@ -1,19 +1,17 @@
 :root {
-  --progress-bar--borderRadius: 1rem;
   --progress-bar--backgroundColor: var(--color-grey--lightest);
   --progress-bar--barColor: var(--color-lime);
 }
 
-.container {
+.container,
+.content {
   width: 100%;
   height: var(--progress-bar--height);
-  border-radius: var(--progress-bar--borderRadius);
+  border-radius: calc(var(--progress-bar--height) / 2);
   background: var(--progress-bar--backgroundColor);
 }
 
 .content {
-  height: var(--progress-bar--height);
-  border-radius: var(--progress-bar--borderRadius);
   background-color: var(--progress-bar--barColor);
   transition: width var(--timing-slower) ease-out;
 }
--- a/packages/components/src/ProgressBar/ProgressBar.css
+++ b/packages/components/src/ProgressBar/ProgressBar.css
@@ -1,5 +1,6 @@
 :root {
   --progress-bar--borderRadius: 1rem;
+  --progress-bar--height: 1rem;
   --progress-bar--backgroundColor: var(--color-grey--lightest);
   --progress-bar--barColor: var(--color-lime);
 }
--- a/packages/components/src/ProgressBar/ProgressBar.css
+++ b/packages/components/src/ProgressBar/ProgressBar.css
@@ -1,6 +1,5 @@
 :root {
   --progress-bar--borderRadius: 1rem;
-  --progress-bar--height: 1rem;
   --progress-bar--backgroundColor: var(--color-grey--lightest);
   --progress-bar--barColor: var(--color-lime);
 }
--- a/packages/components/src/ProgressBar/ProgressBar.tsx
+++ b/packages/components/src/ProgressBar/ProgressBar.tsx
@@ -15,7 +15,8 @@ interface ProgressBarProps {
   readonly totalSteps: number;
 
   /**
-   * Changes the size to small
+   * Set the size of the progress bar
+   * @default base
    */
   readonly size?: keyof typeof sizes;
 }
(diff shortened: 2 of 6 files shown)


packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -12,12 +13,24 @@ interface ProgressBarProps {
...
+   * Changes the size to small
+   */
+  readonly size?: keyof typeof sizes;
 }
 
-export function ProgressBar({ currentStep, totalSteps }: ProgressBarProps) {
+export function ProgressBar({
+  currentStep,
+  totalSteps,
+  size,
```
JeffMarvin, line 26: size 
```suggestion
  size = "base",
```

packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -12,12 +13,24 @@ interface ProgressBarProps {
    * The total steps to use. For percentages you can set this to 100.
    */
   readonly totalSteps: number;
+
+  /**
+   * Changes the size to small
+   */
+  readonly size?: keyof typeof sizes;
```
JeffMarvin, line 20: In other places in Atlantis, we've just used strings. You could do:
```suggestion
  readonly size?: "small" | "base";
```

packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -12,12 +13,24 @@ interface ProgressBarProps {
    * The total steps to use. For percentages you can set this to 100.
    */
   readonly totalSteps: number;
+
+  /**
+   * Changes the size to small
```
JeffMarvin, line 18: Suggestion:
```suggestion
   * Set the size of the progress bar (optional)
```

packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -12,12 +13,24 @@ interface ProgressBarProps {
...
+  currentStep,
+  totalSteps,
+  size,
+}: ProgressBarProps) {
   const percentage = ((currentStep - 1) / (totalSteps - 1)) * 100;
   const widthPercent = Math.min(100, Math.max(0, percentage));
-  const progressBarContainerClassName = classnames(styles.container);
+  const progressBarContainerClassName = classnames(
+    styles.container,
+    size && sizes[size],
```
JeffMarvin, line 32: You'd have to adjust this, depending on what other suggestions from the PR you take. :)

packages/components/src/ProgressBar/Sizes.css
```diff
@@ -0,0 +1,3 @@
+.small {
+  --progress-bar--height: 0.5rem;
```
seanhealy, line 2: Would be better to use the spacing units here.

While you're in here do you want to double check that the "base" size is also using the spacing units? 😄

packages/components/src/ProgressBar/ProgressBar.tsx
seanhealy: What @whereyoubinns has here is actually probably better for this case. It is saying that size can be any of the "classes" exported by `sizes.css`. So, if we just had the base size there we're golden!

packages/components/src/ProgressBar/ProgressBar.tsx
seanhealy: `(optional)` will be inferred by the docs viewer.

whereyoubinns:
Made some changes based on both your feedback! I kept the enumerable sizes on seans suggestion and updated some of the wording. 

I also realized our use case actually needs a progress bar with the "smaller" size, so I updated the spacing accordingly. 

@JeffMarvin / @seanhealy could you do another review please :)

packages/components/src/ProgressBar/Sizes.css
```diff
@@ -0,0 +1,7 @@
+.smaller {
```
seanhealy, line 1: Do we anticipating having a `small` size in the future that would be between `base` and `smaller`? If not this might be better as just `small` since this is what a user of the component will need to specify when they are using it and specifying a size.

cc. @chris-at-jobber 

```suggestion
.small {
```

packages/components/src/ProgressBar/ProgressBar.css
```diff
@@ -1,6 +1,5 @@
 :root {
   --progress-bar--borderRadius: 1rem;
```
seanhealy, line 2: We should update this to be a percentage I think so that the radius scales properly with the height.

whereyoubinns:
@seanhealy updated both of those :)

PR #255, opened 2020-05-11
fix: Build css types on save

eddysims:
## Motivations

When you are working in Atlantis and save a `css` file, we would like the types to generate. This used to happen, and then all of the sudden it didn't anymore. 

This PR makes that happen again!

It also contains an idea to build the types on `npm start`. The idea behind this is that you get your css types built when you start Atlantis.

### Added

- Adds prestart command to package.json

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Builds css types on save

### Security

- 

## Testing

- `npm start`, see that types are generated.
- add a new css class to a css file and save it. See that the type is added to the `.css.d.ts` file\\
---

Random photo of Atlantis

### DIFF
--- a/doczrc.js
+++ b/doczrc.js
@@ -9,7 +9,7 @@ const projectPlugin = () =>
       const config = getConfig();
 
       /**
-       * Generate css types on save.
+       * Generate css types on `.css` file save.
        */
       config.module.rules.push({
         enforce: "pre",
--- a/doczrc.js
+++ b/doczrc.js
@@ -9,6 +9,16 @@ const projectPlugin = () =>
       const config = getConfig();
 
       /**
+       * Generate css types on save.
+       */
+      config.module.rules.push({
+        enforce: "pre",
+        test: /\.css$/,
+        exclude: /node_modules/,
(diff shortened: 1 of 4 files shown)


doczrc.js
```diff
@@ -8,6 +8,16 @@ const projectPlugin = () =>
     onCreateWebpackConfig: ({ stage, rules, actions, loaders, getConfig }) => {
       const config = getConfig();
 
+      /**
+       * Generate css types on save.
```
JeffMarvin, line 12:
```suggestion
       * Generate css types on `.css` file save.
```

package.json
```diff
@@ -4,6 +4,7 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
+    "prestart": "cd packages/components && npm run build:cssTypes",
```
JeffMarvin, line 7: 🤯 I had no idea there were so many of these! https://docs.npmjs.com/misc/scripts

JeffMarvin: I need to learn more `npm` 😆

PR #256, opened 2020-05-15
refactor(design): Move icons logic to @jobber/design

hosseintalebi:
## Motivations

Ability to use Atlantis icons on React Native (jobber-mobile)

### Added

- `getIcon` utility function in `@jobber/design`

### Changed

- Most of the logic regarding getting the correct icon has moved from `components/icon` to `design` 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Icons should work the same as before on Atlantis
- We should be able to use icons from @jobber/design in React Native

---

image


packages/design/src/icons/getIcon.ts
```diff
@@ -0,0 +1,113 @@
...
+import { iconMap } from "./iconMap";
+
+export type IconNames =
+  | keyof typeof iconMap.icons
+  | "longArrowUp"
+  | "longArrowDown"
+  | "remove"
+  | "thumbsDown"
+  | "truck";
+export type IconColorNames =
```
hosseintalebi, line 14: The reason we don't use `keyof typeof colors` here is that we use a global type declaration for all css files on jobber-mobile which prevents us from getting any meaningful types from css files.

packages/design/src/icons/getIcon.ts
```diff
@@ -0,0 +1,113 @@
+import classnames from "classnames";
```
JeffMarvin, line 1: This whole file has a bunch of different mappings in it. 
Consider looking for a way to define these as pure JS data structures (JSON?) or config file formats (yaml? lol barf). Even moving them to be defined that way and imported would help their portability.

packages/components/src/Button/Button.test.tsx
```diff
@@ -163,12 +163,12 @@ it("renders a Button with an icon", () => {
       type="button"
     >
       <svg
-        className="icon base"
+        className="ZDlJU6tECg7ouG-b27gya _3ctOeOMP7zLhU37n25xoZC"
```
JeffMarvin, line 166: Are these hashes deterministic? This seems like it could be brittle...

.eslintignore
```diff
@@ -2,4 +2,5 @@ node_modules
 dist
 .rpt2_cache
 packages/design/colors.js
+packages/design/icons
```
JeffMarvin, line 5: Just to make sure I understand, I'll ask: why are we ignoring this here?

packages/components/src/Button/Button.test.tsx
JeffMarvin: I wonder if there's a way to turn off the classname encoding for the `test` environment, just to avoid this mess/confusion...

hosseintalebi: In the [rollup setup](https://github.com/GetJobber/atlantis/blob/34420a8ac694fd719fe9a7a2dd1653689ef9e901/packages/design/rollup.config.js#L15) for the `design` package, we use `[hash:base64]` for scope names which helps us create unique names that won't collide but they also look ugly. My understanding is that the generated class names stay they same, so they won't be brittle.

We could use a different function other than `[hash:base64]` to generate the class names but it won't be similar to how we generate class names for the [`components` package](https://github.com/GetJobber/atlantis/blob/367969a369682fa1a8a44dd3c4b719f50410415c/packages/components/rollup.config.js#L16).

Since `design` is an npm dependency of the components package, I don't think we can turn off the classname encoding in this case. Those classnames are generated during the build time of the package, which means when we npm install that dependency, they're already there.

@seanhealy do you have any suggestions?

packages/design/src/icons/getIcon.ts
hosseintalebi: I tried to change as little as possible in the original files, so most of what you see here is basically just filed moved from the `components/Icon` to `design`.

For the `classnames` mapping, that's the way it's done everywhere in Atlantis. I believe extracting them requires a conversation to make sure we're consistent with the rest of packages.

The other mapping is `mapToCorrectIcon`, which is a tiny one that we can create a hash map for it. I don't think moving a small map like that to a different file would give us much benefit. Maybe we can make it an inline hash map.

Let me know what you think.

.eslintignore
hosseintalebi: `packages/design/src/icons` is where the actual code lives but `packages/design/icons` is for the generated css files for icons after build. This folder is also git [ignored](https://github.com/GetJobber/atlantis/blob/34420a8ac694fd719fe9a7a2dd1653689ef9e901/packages/design/.gitignore#L3).

packages/design/src/icons/getIcon.ts
JeffMarvin: Ok, no worries. When/if we need to, we can bring these definitions out.

packages/components/src/Button/Button.test.tsx
JeffMarvin: Ok, if these are deterministic, that takes my concern out.

PR #259, opened 2020-05-21
feat(components): Add routing Link to Button

eddysims:
## Motivations

In an SPA we need are able to do client side routing. We want to be able to use the `Button` component as a link for our router.

This PR adds a `to` prop that will allow the `Button` component be used as a `Link` component for `react-router`.

### Added

- `to` prop to Button
- Capability to use Button in client side routing

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
     "@types/lodash": "^4.14.149",
     "@types/react": "^16.9.9",
     "@types/react-dom": "^16.9.2",
-    "@types/react-router-dom": "^5.1.5",
     "css-loader": "^3.4.2",
     "docz": "^2.3.0-alpha.13",
     "docz-core": "^2.3.0-alpha.12",
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
     "react-helmet-async": "^1.0.4",
+    "react-router-dom": "^5.2.0",
     "style-loader": "^0.23.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -78,6 +78,7 @@
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
     "react-helmet-async": "^1.0.4",
+    "react-router-dom": "^5.2.0",
```
seanhealy, line 81: Do we need this here so that it works in the docs viewer?

packages/components/src/Button/Button.css
```diff
@@ -116,7 +116,9 @@
 .disabled,
 .disabled:hover {
   border-color: var(--button--color-disabled);
+  user-select: none;
```
seanhealy, line 119: 🥳

packages/components/src/Button/Button.test.tsx
```diff
@@ -1,266 +1,267 @@
...
-                  type="button"
-                >
-                  <span
-                    className="base extraBold small uppercase greyBlue"
-                  >
-                    Submit
-                  </span>
-                </button>
-        `);
+                    <button
```
seanhealy, line 105: I kind of want to switch all of these to `toMatchSnapshot` what do you think?

package.json
eddysims: yes it is

PR #262, opened 2020-05-25
feat(components): Add disabled prop to feature switch

dhaberst:
## Motivations

We should be able to pass a disabled prop through FeatureSwitch so that the switch can be disabled!

### Added

- optional `disabled` prop in `FeatureSwitch` 
### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

--- Add the disabled prop to the FeatureSwitch and watch the magic happen 🎉 

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FeatureSwitch/FeatureSwitch.tsx
+++ b/packages/components/src/FeatureSwitch/FeatureSwitch.tsx
@@ -20,6 +20,13 @@ interface FeatureSwitchProps {
   readonly description: string;
 
   /**
+   * Determines if the switch is disabled
+   *
+   * @default false
+   */
+  readonly disabled?: boolean;
+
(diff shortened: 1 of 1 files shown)


packages/components/src/FeatureSwitch/FeatureSwitch.tsx
```diff
@@ -63,6 +70,7 @@ interface FeatureSwitchProps {
 export function FeatureSwitch({
   children,
   description,
+  disabled,
```
eddysims, line 73:
```suggestion
  disabled = false,
```

eddysims:
@dhaberst just one small suggestion to keep with the standards on the component

PR #265, opened 2020-05-26
feat(components): Adjust table styling

chris-at-jobber:
## Motivations

We should bake in common presentation elements to the `table` component so the teams implementing don't need to resolve styling every time

### Added

- Border to the tables

### Changed

- Padding on cells

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Table/Cells/Cell.css
+++ b/packages/components/src/Table/Cells/Cell.css
@@ -1,4 +1,6 @@
 .cell {
+  padding: var(--horizontal-inset) calc(var(--horizontal-inset) / 2)
+    var(--horizontal-inset) calc(var(--horizontal-inset) / 2);
   text-align: left;
 }
 
@@ -21,6 +23,7 @@
 .cell:first-child {
   padding-left: var(--horizontal-inset);
 }
+
 .cell:last-child {
   padding-right: var(--horizontal-inset);
 }
--- a/packages/components/src/Table/Cells/Cell.css
+++ b/packages/components/src/Table/Cells/Cell.css
@@ -1,6 +1,5 @@
 .cell {
-  padding: var(--horizontal-inset) calc(var(--horizontal-inset) / 2)
-    var(--horizontal-inset) calc(var(--horizontal-inset) / 2);
+  padding: var(--horizontal-inset) calc(var(--horizontal-inset) / 2);
   text-align: left;
 }
 
--- a/packages/components/src/Table/Table.css
+++ b/packages/components/src/Table/Table.css
@@ -4,9 +4,9 @@
   --horizontal-inset: var(--space-base);
   width: 100%;
 
-  border: solid 1px var(--border-color);
+  box-shadow: 0 0 0 1px var(--border-color);
   border-collapse: collapse;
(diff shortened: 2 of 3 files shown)


packages/components/src/Table/Cells/Cell.css
```diff
@@ -1,4 +1,6 @@
 .cell {
+  padding: var(--horizontal-inset) calc(var(--horizontal-inset) / 2)
```
chris-at-jobber, line 2: Maintains even spacing between cells (in the event of a numeric and non-numeric cell being in sequence, the headers and data would bump up against each other)

Other way of handling this could be simply setting one end to have 0 and the other to have full, instead of dividing by 2?

packages/components/src/Table/Table.css
```diff
@@ -1,10 +1,12 @@
...
   --border-color: var(--color-grey--lighter);
   --header-border-color: var(--color-blue);
-  --horizontal-inset: 0.5rem;
-
+  --horizontal-inset: var(--space-base);
   width: 100%;
+
+  border: solid 1px var(--border-color);
   border-collapse: collapse;
+  border-radius: var(--space-smallest);
```
chris-at-jobber, line 9: I can't get the radius working 👻 - I think something to do with `border-collapse`?

packages/components/src/Table/Table.css
```diff
@@ -1,10 +1,12 @@
 .table {
   --border-color: var(--color-grey--lighter);
   --header-border-color: var(--color-blue);
-  --horizontal-inset: 0.5rem;
-
+  --horizontal-inset: var(--space-base);
   width: 100%;
+
```
eddysims, line 6: We are using box shadow here as a pseudo border. When we use `border-collapse: collapse` it takes away the ability to add a border radius. 

Using box shadow allows us to have the best of both worlds

eddysims:
@chris-at-jobber can you take one more look to make sure that the QA still looks good?

chris-at-jobber:
@eddysims love it, great solution! Should we get @JeffMarvin or @seanhealy to give an unbiased 3rd party review?

chris-at-jobber:
(Also tested and it works)

packages/components/src/Table/Table.css
```diff
@@ -1,10 +1,12 @@
...
+  box-shadow: 0 0 0 1px var(--border-color);
```
seanhealy, line 7:
```suggestion
  box-shadow: 0 0 0 var(--border-base) var(--border-color);
```

https://github.com/GetJobber/atlantis/blob/master/packages/design/src/borders.css#L2

seanhealy:
This looks good. Only thought is that before we push this out we should confirm that the new styles look ok in the places that it is currently used in the main app as we may need a bit of adjustment.

chris-at-jobber:
@seanhealy how do we track those down?

eddysims:
> This looks good. Only thought is that before we push this out we should confirm that the new styles look ok in the places that it is currently used in the main app as we may need a bit of adjustment.

I have confirmed with all use cases that this looks okay. Going to merge this in.

PR #266, opened 2020-05-26
feat(components): remove quoteCopy icon and add system colour to userUnassigned

chris-at-jobber:
## Motivations

- We previously removed the quoteCopy icon from Jobber because we introduced the `copy` icon so this completes the step to deprecate that icon

- The `userUnassigned` icon should be red wherever it is used, this makes that a systemic change

### Changed

- userUnassigned is red now

### Removed

- quoteCopy ✌️ 

Random photo of Atlantis


kdbanman:
ayyyyyy

In the year #266, King Odaenathus of Palmyra invaded Persia to conquer the capital city, Ctesiphon.  He did not succeed!  

Better luck on the third try, Odaenathus.

✌️

PR #269, opened 2020-05-28
feat(components): Button as Icon only

eddysims:
## Motivations

Currently, if an icon is used as a stand-alone thing in a button, you must add an empty string as as a label. This also causes some bonkers padding on the button.

This PR will allow you to use a `<Button />` with just an `icon` and fix the padding.
 
## Changes

Allow you to use a `<Button />` with just an `icon`.

### Added

- Aria label to the buttons

### Changed

- ButtonProps to allow for an icon or label to be required
- Padding on icon if only an icon is present

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Button/Button.tsx
```diff
@@ -7,6 +7,7 @@ import { Typography } from "../Typography";
 import { Icon, IconNames } from "../Icon";
 
 interface ButtonFoundationProps {
+  readonly ariaLabel?: string;
```
seanhealy, line 10: Might be good to describe that `ariaLabel` overrides `label` for readers if it is present?

eddysims: @seanhealy addressed

PR #271, opened 2020-06-01
docs(components): fix card page hierarchy

chris-at-jobber:
## Motivations

You should only have a single h1 per page

### Changed

Bump usage down to h2, bump its' children down to h3

Random photo of Atlantis


agglasaur:
how the heck do i approve

PR #274, opened 2020-06-04
fix(components): update button padding on icon button

chris-at-jobber:
## Changes

Reduced left and right padding on icon button so that these buttons have a square shape rather than a slightly-wider-than-tall rectangular shape

### Changed

- reduced left and right padding on `hasIcon` button that does not also have a label

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -141,8 +141,8 @@
 
 /* Icon */
 
-.hasIcon {
-  width: calc(var(--base-unit)*2.25);
+.onlyIcon {
+  width: calc(var(--base-unit) * 2.25);
   padding: 0;
 }
 
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -147,6 +147,7 @@
 }
 
 .hasIconAndLabel {
+  width: auto;
   padding: 0 var(--space-base);
 }
 
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -142,7 +142,8 @@
 /* Icon */
 
 .hasIcon {
-  padding: 0 var(--space-small);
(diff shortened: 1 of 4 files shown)


packages/components/src/Button/Button.css
```diff
@@ -140,7 +140,14 @@
 }
 
 /* Icon */
+
+.hasIcon {
+  width: 2.25rem;
```
seanhealy, line 145: Can we make this based off of one of our spacing units?

seanhealy, line 145: Can we make this based off of one of our spacing units?

packages/components/src/Button/Button.tsx
```diff
@@ -95,6 +95,7 @@ export function Button(props: ButtonProps) {
   } = props;
 
   const buttonClassNames = classnames(styles.button, styles[size], {
+    [styles.hasIcon]: icon,
```
JeffMarvin, line 98:
```suggestion
    [styles.onlyIcon]: icon && !label,
```

packages/components/src/Button/Button.css.d.ts
```diff
@@ -10,6 +10,7 @@ export const disabled: string;
 export const small: string;
 export const base: string;
 export const large: string;
+export const hasIcon: string;
```
JeffMarvin, line 13:
```suggestion
export const onlyIcon: string;
```

packages/components/src/Button/Button.css
```diff
@@ -140,7 +140,14 @@
 }
 
 /* Icon */
+
+.hasIcon {
```
JeffMarvin, line 144:
```suggestion
.onlyIcon {
```

JeffMarvin:
Talked to @chris-at-jobber and verified the approach I suggest. Approving and merging.

PR #280, opened 2020-06-11
feat: export foundation as a js file with the corresponding types

Afroholic78:
## Motivations
We created a foundation.js file as well as its type to allow us to export it into react native.
Keeping the CSS modules was causing us to run into issues where we had to make work arounds with external dependencies. Importing it like this should fix this situation

## Changes
Changed the npm build method to include the new npm builds.

### Added
Added a `buildfoundation.js` to create the js file using postcss and added a subsequent `tsc` call to create the type file.

## Testing
Run `npm run:build` and see these new files available for import

---

Random photo of Atlantis


packages/design/jobberStyle.js
```diff
@@ -0,0 +1,81 @@
...
+// eslint-disable-next-line import/no-internal-modules
+const customPropertiesObject = require("./foundation.js");
+
+const customProperties = customPropertiesObject.customProperties;
+const allKeys = Object.keys(customProperties);
+const resolvedCssVars = allKeys.reduce((acc, key) => {
+  const newKey = key.replace("--", "");
+  acc[newKey] = jobberStyle(key);
+  return acc;
+}, {});
```
seanhealy, line 12: It might be nice to extract this block into a function that returns `resolvedCssVars`. it would just group it up nicely and give it a nice name.

packages/design/jobberStyle.js
```diff
@@ -0,0 +1,81 @@
...
+
+const jsonContent =
+  "export const JobberStyle = " + JSON.stringify(resolvedCssVars, undefined, 2);
+
+fs.writeFile("./foundation.js", jsonContent, "utf8", function(err) {
```
seanhealy, line 17: It looks here like we are editing `foundation.js` in place. I wonder if it would be nicer to have a raw file that is generated from postcssCustomProperties in a temp location then using that to generate a fresh `foundation.js` that way we don't have two scripts generating the same file.

packages/design/jobberStyle.js
```diff
@@ -0,0 +1,81 @@
...
+  if (err) {
+    console.log("An error occured while writing JSON object to File.");
+    return console.log(err);
+  }
+  console.log("JSON file has been saved.");
+});
+
+function jobberStyle(styling) {
+  const varRegExp = /var\((.*)\)/;
+  const calcRegExp = /calc\((.*)\)/;
```
seanhealy, line 27: Can we use a post css plugin to handle calc for us ahead of this script?

packages/design/jobberStyle.js
```diff
@@ -0,0 +1,81 @@
...
+fs.writeFile("./foundation.js", jsonContent, "utf8", function(err) {
+  if (err) {
+    console.log("An error occured while writing JSON object to File.");
+    return console.log(err);
+  }
+  console.log("JSON file has been saved.");
+});
+
+function jobberStyle(styling) {
+  const varRegExp = /var\((.*)\)/;
```
seanhealy, line 26: Perhaps we could have a regex object that we store all these in as a constant at the module level so we don't need to duplicate them?

```
const regex = {
  var: /var\((.*)\)/,
  ...
}
```

packages/design/jobberStyle.js
```diff
@@ -0,0 +1,81 @@
...
+  const calcRegExp = /calc\((.*)\)/;
+  const removeAllNonNumerals = /[^0-9.+\-/*]/gi;
+  const styleValue = customProperties[styling];
+  const varRegexResult = varRegExp.exec(styleValue);
```
seanhealy, line 30: This is a really unhelpful comment but, I'm finding this method a bit hard to parse. I wonder if it would be more readable if broken into smaller more functional bits?

packages/design/jobberStyle.js
Afroholic78: We could but the raw file wouldn't be needed for anything else.
Would it be better to have a raw file and at the end of the processing have something to delete the raw file?

packages/design/jobberStyle.js
Afroholic78: the postcss plugin wasn't handling calcs that have css vars.
`'--space-smallest': 'calc(var(--base-unit)/8)',` doesn't get evaluated and stays as is
if the expression was `'--space-smallest': 'calc(8/8)',` it gets resolved to `1`

packages/design/package.json
```diff
@@ -8,16 +8,22 @@
...
     "build:rollup": "rollup --config",
     "build:css": "postcss src/foundation.css --output foundation.css --use postcss-import",
     "build:css:icons": "postcss src/icons/*.css --dir icons/ --use postcss-import",
     "build:ts": "tsc --project .",
     "build:cssTypes": "tcm src/icons",
     "build:colors": "node buildColors.js",
+    "build:foundation": "node buildFoundation.js && node jobberStyle.js && npm run build:foundationTypes && npm run build:addTypeLinterExceptions && npm run build:addStyleLinterExceptions && npm run build:removeTempFiles",
+    "build:foundationTypes": "tsc --declaration --emitDeclarationOnly --allowJs foundation.js --outDir .",
+    "build:addTypeLinterExceptions": "(printf '/* tslint:disable */\n/* eslint-disable */\n/* This file is automatically generated and should not be edited. */\n'; cat foundation.d.ts) > foundation.d.ts.temp",
+    "build:addStyleLinterExceptions": "(printf '/* tslint:disable */\n/* eslint-disable */\n/* This file is automatically generated and should not be edited. */\n'; cat foundation.js) > foundation.js.temp",
```
seanhealy, line 25: `StyleLint` is a tool we actually use for CSS linting to avoid confusion with it wonder if `addFoundationLinterExceptions` might make more sense here?

packages/design/package.json
```diff
@@ -8,16 +8,22 @@
...
+    "build:removeTempFiles": "rm -f src/foundation.js && mv foundation.d.ts.temp foundation.d.ts && mv foundation.js.temp foundation.js",
```
seanhealy, line 26: I wonder if it would be nicer to make a `tmp` directory that we can just git ignore then we can skip this explicit cleanup step?

packages/design/jobberStyle.js
```diff
@@ -0,0 +1,93 @@
...
+  if (err) {
+    console.log("An error occured while writing JSON object to File.");
+    return console.log(err);
+  }
+  console.log("JSON file has been saved.");
+});
+
+//jobberStyle recursively resolves css vars
+//varRegexResult returns --base-unit from var(--base-unit)
+//calcRegexResult returns var(--base-unit) / 16 from calc(var(--base-unit) / 16)
```
seanhealy, line 30: Might be clearer to reference these inline in the method near where they are relevant.

packages/design/jobberStyle.js
```diff
@@ -0,0 +1,93 @@
...
+
+fs.writeFile("./foundation.js", jsonContent, "utf8", function(err) {
+  if (err) {
+    console.log("An error occured while writing JSON object to File.");
+    return console.log(err);
+  }
+  console.log("JSON file has been saved.");
+});
+
+//jobberStyle recursively resolves css vars
```
seanhealy, line 28: Perhaps something like this just to give a better idea to someone new what's going on here and why? Also, realize that this example isn't quite right but, hopefully it's not too far off.

```suggestion
/**
 * Recursively resolve css custom properties.
 *
 * eg:
 * ```
 * jobberStyle(`
 *   "--base": "5",
 *   "--foo": "calc(var(--base) * 2),
 * `);
 * ```
 * =>
 * ```
 * {
 *   "--base": "5",
 *   "--foo": "10",
 * }
 * ```
 */
```

packages/design/jobberStyle.js
```diff
@@ -0,0 +1,93 @@
...
+  finalExpression,
+}) {
+  const cssVariable = cssVariableRegexResult[1];
+  const realValue = jobberStyle(cssVariable);
+  const unresolvedCssVariable = group;
+  return finalExpression.replace(unresolvedCssVariable, realValue);
+}
+
+function isSpacingValue(value) {
+  return !!String(value).match(/(^\d+(px|%)$)|(^\d+$)/);
```
seanhealy, line 83: Might not be a bad idea for this to support rem or other CSS length units.

https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#Lengths

Afroholic78: Right, I think I saw that we're trying to move to `rem`

packages/design/package.json
Afroholic78: The previous steps can't write to files in place so we write to tmp files and then make the tmp file the real ones (See the `mv` parts). So we would still need most of it. It would remove the `rm -f` part, I can do it if you think this would benefit from that but the `mv` steps are still required to happen

PR #281, opened 2020-06-11
feat(components): Make Page actions responsive

eddysims:
## Motivations

When we shove the `Page` into small places such as a mobile device, the actions get squished. This PR helps to let the actions breathe in smaller places

## Changes

Makes `Page` actions responsive

### Added

- Responsive css to `Page` component

### Changed

- Makes the `Button` in `Menu` full width

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

In preview build, resize the `Page` component that uses actions.

---

Random photo of Atlantis


packages/components/src/Menu/Menu.tsx
```diff
@@ -54,7 +54,14 @@ export function Menu({ activator, items }: MenuProps) {
   }, [visible]);
 
   if (!activator) {
-    activator = <Button label="More Actions" icon="more" type="secondary" />;
+    activator = (
+      <Button
+        fullWidth={true}
```
eddysims, line 59: I have looked at this in and because its wrapped in an `inline-block` div this should be okay to do. Open to discuss.

packages/components/src/Page/Page.tsx
```diff
@@ -62,25 +63,60 @@ export function Page({
...
 }: PageProps) {
   const pageStyles = classnames(styles.page, styles[width]);
+  const [titleBarRef, { width: titleBarWidth }] = useResizeObserver<
+    HTMLDivElement
+  >();
+
+  const titleBarClasses = classnames(styles.titleBar, {
+    [styles.small]: titleBarWidth && titleBarWidth > 265,
+    [styles.medium]: titleBarWidth && titleBarWidth > 500,
+    [styles.large]: titleBarWidth && titleBarWidth > 750,
```
seanhealy, line 73: I think we should add, to the followup, a task to formalize these breakpoints.

packages/components/src/Page/Page.css
```diff
@@ -16,12 +16,57 @@
...
+  flex: 1;
+}
+
+.actionButton + .actionButton {
+  margin-left: var(--space-small);
+}
+
+.actionButton > div {
+  width: 100%;
+}
```
seanhealy, line 58: This feels a little bit funny to me. What are we doing with it?

packages/components/src/Menu/Menu.tsx
seanhealy: The other option would be to pass in the activator… I'm kind of torn. What you have works and I think is cleaner but, it also feels a bit odd…

eddysims: Have discussed with Sean, he's okay with this.

packages/components/src/Page/Page.tsx
eddysims: I have added a task to do this

packages/components/src/Page/Page.css
eddysims: I have added a task for this as well.

eddysims:
@seanhealy I have added the comments in a task to handle the future changes on this one. Back to you.

PR #282, opened 2020-06-12
feat(components): menu CSS updates for mobile responsive behaviour

chris-at-jobber:
## Motivations

The Atlantis `menu` should have responsive parity with the existing `dropdown-menu` in Jobber Online

## Changes

- not using resizeObserver because this behaviour is specific to browser window size
- converts menu into an action-sheet-esque drawer on small screens
- Updates test framework

### Added

- media queries and associated stylinh

### Changed

- menu has a dark overlay and is fixed to the bottom of the viewport on browser windows smaller than 640px wide

Random photo of Atlantis


chris-at-jobber:
Things I'm stuck on...

• The old implementation of this menu used CSS classes to swap states, which made smooth transitions between show and hide states easy. Because this implementation actually removes the menu from the DOM we'll need to use animations, and I'm not sure if we have a library for that?

• The pointer arrow that visually connects the menu to the button on medium+ screens is being a bit fussy with overflow states so I've left that out for now, but we should solve for it

seanhealy:
@chris-at-jobber We have Framer Motion for animations in Atlantis at the moment.

seanhealy:
@chris-at-jobber We have Framer Motion for animations in Atlantis at the moment.

You can see an example of it in a few places. Tooltip may be the most relevant to this?

seanhealy:
@chris-at-jobber Wanna take a look at the build here and see what you think?

packages/components/src/Menu/Menu.css
```diff
@@ -10,28 +10,18 @@
...
-    left: 0;
-    max-height: 72vh;
-    padding-bottom: env(safe-area-inset-bottom);
-    border-radius: var(--radius-base) var(--radius-base) 0 0;
-    overflow-y: scroll;
-    transition: transform 300ms ease;
-    -webkit-overflow-scrolling: touch;
-  }
-
-  .menu:last-child {
```
chris-at-jobber: We still need this to give the last menu item proper spacing:
image

packages/components/src/Menu/Menu.css
```diff
@@ -10,28 +10,18 @@
...
-    padding-bottom: var(--space-small);
-  }
+  -webkit-overflow-scrolling: touch;
 }
 
 @media only screen and (min-width: 640px), print {
```
chris-at-jobber, line 27: I think we need to set an explicit right value now on these medium+ screens to get the menus aligning to the right of the button as they were prior

packages/components/src/Menu/Menu.css
chris-at-jobber: (I'll push this up)

packages/components/src/Menu/Menu.css
chris-at-jobber: (also pushed up)

packages/components/src/Menu/Menu.tsx
```diff
@@ -13,6 +13,11 @@ import { Button } from "../Button";
 import { Typography } from "../Typography";
 import { Icon, IconNames } from "../Icon";
 
+const variation = {
+  startOrStop: () => ({ opacity: 0, y: window.innerWidth > 640 ? 10 : 50 }),
```
chris-at-jobber, line 17: Is this where the actual animation is being defined?

seanhealy: @chris-at-jobber  It is.

chris-at-jobber:
@eddysims @JeffMarvin Sean wrote this latest commit from my machine, so could one of you kindly give this a 👀 🙏

packages/components/src/Menu/Menu.css
eddysims: the `left: auto` should handle that I believe

packages/components/src/Menu/Menu.css
```diff
@@ -8,27 +8,53 @@
...
+  bottom: 0;
+  left: 0;
   z-index: var(--elevation-menu);
-  width: calc(var(--base-unit) * 12.5);
-  max-height: 300px;
-  box-shadow: var(--shadow-base);
+  width: 100%;
+  max-height: 72vh;
   padding: var(--space-small);
+  padding-bottom: env(safe-area-inset-bottom);
```
eddysims, line 19: what is `env()`? I think this might be a `scss` function that got sucked over from sg1

packages/components/src/Menu/Menu.css
```diff
@@ -8,27 +8,53 @@
...
-.below {
-  top: 100%;
-  margin-top: var(--space-smallest);
+@media only screen and (min-width: 640px), print {
+  .menu {
+    position: absolute;
+    left: auto;
+    z-index: var(--elevation-menu);
+    width: calc(var(--base-unit) * 12.5);
+    max-height: 300px;
```
eddysims, line 39: should we use a `calc(var(--base-unit) * 18.75)` for this number here?

packages/components/src/Menu/Menu.css
```diff
@@ -8,27 +8,53 @@
...
+  .menu:last-child {
+    padding-bottom: var(--space-small);
+  }
 }
 
-.below {
-  top: 100%;
-  margin-top: var(--space-smallest);
+@media only screen and (min-width: 640px), print {
+  .menu {
```
eddysims, line 34: a few attributes here are double declared from the small version and can be removed.

```
z-index
border
background-color
```

packages/components/src/Menu/Menu.tsx
```diff
@@ -7,11 +7,25 @@ import React, {
...
 import { Button } from "../Button";
 import { Typography } from "../Typography";
 import { Icon, IconNames } from "../Icon";
 
+const variation = {
+  overlayStartStop: { opacity: 0 },
+  startOrStop: (position: string) => {
+    let y = 10;
+
+    if (position === "below") y *= -1;
```
eddysims, line 21: why this vs `y = -10`? I feel this is more confusing

packages/components/src/Menu/Menu.css
seanhealy: https://developer.mozilla.org/en-US/docs/Web/CSS/env

packages/components/src/Menu/Menu.tsx
seanhealy: I did this so it reverses the amount so they don't get out of sync.

packages/components/src/Menu/Menu.css
seanhealy: Or, even a rounder number would be nice. I wonder if we could *20 it? @chris-at-jobber

packages/components/src/Menu/Menu.css
eddysims: interesting. never heard of this

chris-at-jobber: [env()](https://developer.mozilla.org/en-US/docs/Web/CSS/env) is a real thing! 😆  This is for the notch / no button iPhones/other devices that move towards that pattern of screen-based home interactions.

packages/components/src/Menu/Menu.css
chris-at-jobber: Yep, added that in below 👍

packages/components/src/Menu/Menu.css
chris-at-jobber: On second look at the dropdowns in Jobber right now we shouldn't even be setting a max-height on this unless we have a good plan to revisit the entire menu pattern

seanhealy: I wonder if max hight should be more something like `90vh` so it scrolls but only if it's REALLY BIG?

packages/components/src/Menu/Menu.tsx
chris-at-jobber: If we set the `10` as a variable(?) and then did a more explicit multiplication of the variable could that solve the out-of-sync issue while making it more apparent what's happening?

chris-at-jobber: Jesse made that comment once about magic numbers for animations in a coffeescript thing I did once

packages/components/src/Menu/Menu.css
chris-at-jobber: ✅

packages/components/src/Menu/Menu.tsx
seanhealy: I think that's basically what we're doing here? Starting `y` at `10` then making it negative if the position is below.

packages/components/src/Menu/Menu.css
chris-at-jobber: Probably the nicest way to solve for this is "let the menu grow from (the nearest point to the button) to (16px from the top or bottom of the viewport) and then scroll if there's overflow", which I think I could hack out in jQuery 😆 ...

seanhealy: Oh, I was thinking of the responsive version with the above. I agree with you for the non-responsive tough.

chris-at-jobber:
@eddysims @seanhealy I would like to swap these breakpoint variables in place of all of the manual breakpoints in this file but I'm not sure how the syntax for these actually works...

```
@custom-media --handhelds (width < 640px);
@custom-media --medium-screens (width >= 640px);
@custom-media --wide-screens (width >= 1024px);
```

seanhealy:
@chris-at-jobberhttps://github.com/postcss/postcss-custom-media

seanhealy:
@chris-at-jobber I think this is the syntax that you're looking for.

https://github.com/postcss/postcss-custom-media

You can also see it in: `packages/components/src/Typography/css/Typography.css`

seanhealy:
Talked to Chris, we'll followup with the responsive breakpoints as part of the other task related to that.

packages/components/src/Menu/Menu.css
chris-at-jobber: Yeah max-height of 72vh on the small one is intentional to ensure there's always adequate tap-to-close room at the top of the view

chris-at-jobber: (Client More Actions is probably the best example) 
image

PR #294, opened 2020-06-25
feat(eslint): Add resolution patch support for eslint 7

jcw-:
## Motivations

Tried to use eslint 7 in a downstream project but the monkey patch for plugin resolution is eslint 6 specific.

## Changes

Switched to an [npm package](https://www.npmjs.com/package/@rushstack/eslint-patch) that wraps this patch and supports both eslint 6 and 7.

### Added

- Support for eslint 7

---

Random photo of Atlantis


seanhealy:
@jcw- It looks like ESLint got bumped in the lock file for some reason which is breaking ci. Otherwise this looks great!

jcw-:
@seanhealy whoops, yeah i was playing with that - force-pushed, should be fixed now

PR #295, opened 2020-06-25
build: Responsive CSS breakpoints

eddysims:
## Motivations

We have custom media queries in Atlantis, however they were not working in Atlantis.

This PR helps make the responsive breakpoints (and all other postcss) work in Atlantis as well as Jobber.

## Changes

Adds postcss to Atlantis.

### Added

- Responsive Breakpoints page.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- postcss working in atlantis

### Security

- 

## Testing

---

Random photo of Atlantis


eddysims:
@seanhealy you will notice I moved the webpack work from `.doczrc` to `gatsby-config.js`. This is the proper way to handle this in Docz 2 I believe

gatsby-node.js
```diff
@@ -0,0 +1,101 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
seanhealy, line 1: Did this not work when built as a plugin or did you extract it for other reasons?

packages/design/src/ResponsiveBreakpoints.mdx
```diff
@@ -0,0 +1,48 @@
...
+menu: Design
+route: /breakpoints
+---
+
+import { Button } from "@jobber/components/Button";
+
+# Responsive Breakpoints
+
+To make most components responsive, you will want to use the
+[useResizeObserver hook](/hooks/use-resize-observer) so that the component can
```
seanhealy, line 12: I wonder if we should make this a bit more prominent some how? As it is an important point that is quite different from how we are used to doing responsiveness at Jobber.

gatsby-node.js
eddysims: This is the proper way to handle modifying the webpack stuff in Docz 2 I believe. Im open to moving it back in, but I do like the idea of keeping the `.doczrc` a little bit cleaner.

packages/design/src/ResponsiveBreakpoints.mdx
eddysims: @seanhealy I had thought of this in another PR, but pulled it in here. I updated the atlantis menu to bring design and hooks up to the top of the menu to make them more prominent. I think that will help users get their eyes on it. 

Thoughts?

eddysims:
@seanhealy back to you

PR #296, opened 2020-06-25
feat(components): add FormatRelativeDateTime

naarok:
## Motivations

We needed a way to display date times in a relate manor. E.g. 4 minutes ago, 1 year ago
### Added

- FormatRelativeDateTime

### Changed

-none

### Deprecated

- none

### Removed

- none

### Fixed

- none

### Security

- none

## Testing

Use the new component

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
+++ b/packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
@@ -1,7 +1,7 @@
 ---
 name: FormatRelativeDateTime
 menu: Components
-route: /components/datetime-formatter
+route: /components/format-relative-date-time
 ---
 
 import { Playground, Props } from "docz";
--- a/packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
+++ b/packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
@@ -6,7 +6,7 @@ route: /components/format-relative-date-time
 
 import { Playground, Props } from "docz";
 import { ComponentStatus } from "@jobber/docx";
-import { CivilDate } from "@std-proposal/temporal";
+import { CivilDateTime } from "@std-proposal/temporal";
 import { FormatRelativeDateTime } from ".";
 
 # FormatRelativeDateTime
@@ -14,10 +14,18 @@ import { FormatRelativeDateTime } from ".";
 <ComponentStatus stage="pre" responsive="yes" accessible="yes" />
 
 A FormatRelateDateTime displays the date and time using relative wording or
-approximate values.
+approximate values. Note. This component only represents current date and dates
+in the past. No future dating.
 
-Rules: If the message was received less than an hour ago the format is x minutes
-ago
+## Simple Example
+
+```ts
+import { FormatRelativeDateTime } from "@jobber/components/FormatRelativeDateTime";
+```
+
+## Rules
+
+If the message was received less than an hour ago the format is x minutes ago
 
 If the message was received today, but more than an hour ago, the format becomes
 the time it was received
@@ -31,14 +39,8 @@ previous, the format is Month Date
 If the message was received exactly one year ago or more than one year ago, the
 format is Month Day, Year
 
-## Simple Example
-
-```ts
-import { FormatRelativeDateTime } from "@jobber/components/FormatRelativeDateTime";
-```
-
 <Playground>
-  <FormatRelativeDateTime date={new CivilDate(2020, 2, 26)} />
+  <FormatRelativeDateTime date={new CivilDateTime(2020, 2, 26, 14, 29, 39)} />
 </Playground>
 
 ## Props
--- a/packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
+++ b/packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
@@ -1,0 +1,31 @@
+---
+name: FormatRelativeDateTime
+menu: Components
+route: /components/datetime-formatter
+---
+
+import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { CivilDate } from "@std-proposal/temporal";
+import { FormatRelativeDateTime } from ".";
+
+# FormatRelativeDateTime
+
+<ComponentStatus stage="pre" responsive="yes" accessible="yes" />
+
+In Jobber a FormatDate is used to ensure that the date is displayed in the
+expected format. No text styling is applied, this simply formats the date.
+
+## Simple Example
+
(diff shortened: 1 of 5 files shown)


eddysims:
looks like `new CivilDate` is returning one day off.

image

eddysims:
`1 minutes ago` should be `1 minute ago`.

image

eddysims:
We probably want to consider some sort of handling for future dates?

image

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return "a year or more";
+}
+
+function strFormatDate(date: Date, options: { [key: string]: string }) {
+  return date.toLocaleDateString(undefined, options);
+}
+
+function showMinutes(minutes: number) {
+  if (minutes < 1) {
```
eddysims, line 63:
```suggestion
  if (minutes <= 1) {
```

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.tsx
```diff
@@ -0,0 +1,68 @@
...
+) {
+  const now = Date.now() / 1000; //seconds
+  const inputDate = new Date(civilDateTime.date.toJSON());
+  const delta = now - inputDate.getTime() / 1000; //seconds;
+
+  switch (relativeTimeRange(delta)) {
+    case "less than an hour":
+      return showMinutes(Math.round(delta / 60));
+    case "less then a day":
+      return <>{inputDate.toLocaleTimeString()}</>;
```
eddysims, line 22: because you are not returning any elements, you can remove all the fragments (`<></>`) and just do:

```
return inputDate.toLocaleTimeString()
```

eddysims: Same will go for all return statements

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.tsx
```diff
@@ -0,0 +1,68 @@
...
+  if (delta < 24) return "less then a day";
+
+  delta = delta / 24;
+  if (delta < 7) return "less than a week";
+
+  delta = delta / 365;
+  if (delta < 1) return "less than a year";
+
+  return "a year or more";
+}
```
eddysims, line 56: Not a blocker, but I'm not a huge fan of this function returning a string to compare to. although I'm not sure sure of a better solution.

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
```diff
@@ -0,0 +1,46 @@
...
+the time it was received
+
+If the message was received this within the last six days, the format is Sun,
+Mon, Tue, Wed, Thu, Fri, Sat
+
+If the message was received between six days ago and yesterday’s date one year
+previous, the format is Month Date
+
+If the message was received exactly one year ago or more than one year ago, the
+format is Month Day, Year
```
eddysims, line 32: I think this whole rules content should have its own section below the `Simple Example`. 

```
## Rules

...
```

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
```diff
@@ -0,0 +1,46 @@
+---
+name: FormatRelativeDateTime
+menu: Components
+route: /components/format-relative-date-time
+---
+
+import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { CivilDate } from "@std-proposal/temporal";
```
eddysims, line 9: I think using `CivilDateTime` is a better example, as it allows for usage with hours and min.

eddysims:
@naarok a few comments here. Back to you! Let me know if you have questions.

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.tsx
naarok: Yeah. Sean and I worked on this a bunch. It was the best we could come up with. On the plus side, it is nice and "self documenting"

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
naarok: Good point.

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.tsx
naarok: Thanks. Stupid of me.

naarok:
Re. Future dates. Out of scope for this PR. If we want to handle them, a new set of rules need to be defined and a new story created.

naarok:
Re. Date being of using CivilDate. Couple things. Example should have been CivilDateTime like you recommended. With that, boundary conditions of "2020, 02, 26, 0, 0, 0" and "2020, 02, 26, 23, 59, 59" work as expected. The problem is that the function doesn't actually properly know how to handle CivilDate. The signature clearly requires a CivilDateTime. I'm not sure if we should/can we protect against wrong use of this.

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.tsx
naarok: If I drop the <> and </>, then I get compile errors. In particular, I could find no way to get the following to compile without the <> and </>
`<>{strFormatDate(inputDate, { month: "short", day: "numeric" })}</>`

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.mdx
naarok: Done

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.tsx
eddysims: If you take out the `<>` and the `{}` and just return the function:

```
return strFormatDate(inputDate, { month: "short", day: "numeric" });
```

This should work. Would be happy to jump on a chat to get working as well.

eddysims:
@naarok just one comment in regards to the fragments.

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.tsx
naarok: @eddysims Got it now, I was keeping the extra { } surrounding the whole thing.

PR #297, opened 2020-06-30
feat(components): Convert px value for base spacing to REM

chris-at-jobber:
## Motivations

@kkimble006 made a good callout about our use of px vs rem for accessibility concerns (mostly related to display of type). I looked back at the [PR where we created spacing](https://github.com/GetJobber/atlantis/pull/51) and there's no record of using px for a particular reason, so I'm switching to rems.

### Changed

- We use rem instead of px for our base unit now

image


eddysims:
@chris-at-jobber my only question is are we okay with the docz page still showing the px value? Should we have something on this page talking about rems vs pixels?

image

chris-at-jobber:
Oh wow I forgot this was still open

chris-at-jobber:
If we can print the `rem` value to the docs, I would love that; otherwise I'm happy to add documentation that makes it clear we're actually setting the value using `rem`

chris-at-jobber:
We get the same accessibility benefits of REM from our CSS variables, so this is not required.

PR #303, opened 2020-07-08
  feat(components): InputText Automatic Resize

tyler-santos:
## Motivations
We need to be able to have a textarea that would start as 1 line tall, and grow to a set amount, and then become scrollable if more room was needed beyond that set maximum. 

The current implementation almost allowed us to do this, however it wouldn't allow us to set a minimum number of rows.

## Changes

Instead of just rows as a single number being passed into the InputText component, now we have the ability to pass in an associative array which has a min and max key, corresponding to the minimum and maximum number of rows that would like to be shown.

### Added

handleChange and calculateRows functions which actively monitor and calculates how many rows should be shown for a multiline text area **with** a set maximum and minimum number of rows.

### Changed

The only change in existing functionality is the removal of the min-height for text area so that we could have a minimum value less than 3. Instead we set the default rows value to be 3 for multiline, which can be overrided with any value. Essentially it behaves the exact same as before but now you can have less than 3 rows in the textarea for use cases such as ours.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Added a jest test to ensure that the calculations were correct.

**QA**
From the input text page in Atlantis, try using the following code snippet in one of the testing areas.
`<InputText defaultValue="Rocinante" multiline rows={{min:1, max:6}} />`

Play with the min and max sizes to see different behaviours. 
Test to make sure that the behaviour of Rows as a single number still works and multiline still works as a standalone with a default number of visible rows of 3.

---

Random photo of Atlantis

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 
 **/tsconfig.tsbuildinfo
 .DS_Store
+.idea
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -72,7 +72,6 @@
 
 textarea.formField {
   height: auto;
-  min-height: calc(var(--field--height) * 2);
   resize: vertical;
 }
 
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -154,7 +154,7 @@ select.formField {
 
 .miniLabel:not(.small) .formField {
   --field--padding-top: calc(var(--space-base) + var(--space-smaller));
-  --field--padding-bottom: var(--space-smaller);
+  --field--padding-bottom: var(--space-small);
 }
 
 .miniLabel .label {
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -172,7 +172,7 @@ export const FormField = React.forwardRef(
       onValidation,
       placeholder,
       readonly,
-      rows = 3,
+      rows,
       size,
       type = "text",
       value,
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -172,7 +172,7 @@ export const FormField = React.forwardRef(
       onValidation,
       placeholder,
       readonly,
-      rows,
+      rows = 3,
       size,
       type = "text",
       value,
--- a/packages/components/src/InputText/InputText.tsx
+++ b/packages/components/src/InputText/InputText.tsx
@@ -100,12 +100,11 @@ function InputTextInternal(
   }
 
   function getLineHeight(textArea: HTMLTextAreaElement) {
-    const { fontSize, lineHeight } = window.getComputedStyle(textArea);
-    const lh =
-      lineHeight === "normal"
-        ? parseFloat(fontSize) * 1.2
-        : parseFloat(lineHeight);
-    return lh;
+    const lineHeight = window
+      .getComputedStyle(textArea)
+      .getPropertyValue("line-height");
+
+    return parseInt(lineHeight, 10);
   }
 
   function insertText(text: string) {
--- a/packages/components/src/InputText/InputText.tsx
+++ b/packages/components/src/InputText/InputText.tsx
@@ -1,10 +1,4 @@
-import React, {
-  Ref,
-  createRef,
-  forwardRef,
-  useImperativeHandle,
-  useState,
-} from "react";
+import React, { Ref, createRef, forwardRef, useImperativeHandle } from "react";
 import { XOR } from "ts-xor";
 import { FormField, FormFieldProps } from "../FormField";
 
@@ -64,7 +58,7 @@ function InputTextInternal(
     },
   }));
 
-  const [currentRows, setRows] = useState(initializeRows());
+  const lines = typeof props.rows === "object" ? props.rows.min : props.rows;
 
   return (
     <FormField
@@ -72,46 +66,47 @@ function InputTextInternal(
       type={props.multiline ? "textarea" : "text"}
       ref={inputRef}
       onChange={handleChange}
-      rows={currentRows}
+      rows={lines}
     />
   );
 
-  function initializeRows() {
-    if (typeof props.rows === "object") {
-      return props.rows.min;
-    }
-    return props.rows;
-  }
-
   function handleChange(newValue: string) {
     props.onChange && props.onChange(newValue);
-    if (
-      props.rows &&
-      typeof props.rows !== "number" &&
-      inputRef &&
-      inputRef.current instanceof HTMLTextAreaElement
-    ) {
-      const calculatedRows = getCalculatedRows(inputRef.current);
+    if (inputRef && inputRef.current instanceof HTMLTextAreaElement) {
+      resize();
+    }
 
-      if (
-        calculatedRows >= props.rows.min &&
-        calculatedRows <= props.rows.max
-      ) {
-        setRows(calculatedRows);
+    function resize() {
+      if (inputRef.current) {
+        inputRef.current.style.height = "auto";
+        inputRef.current.style.height = textAreaHeight() + "px";
       }
     }
+    function textAreaHeight() {
+      if (
+        typeof props.rows === "object" &&
+        inputRef.current &&
+        typeof getLineHeight() === "number"
+      ) {
+        const maxHeight = props.rows.max * getLineHeight();
+        return Math.min(inputRef.current.scrollHeight, maxHeight);
+      }
+      return undefined;
+    }
+    function getLineHeight() {
+      if (inputRef.current) {
+        const { fontSize, lineHeight } = window.getComputedStyle(
+          inputRef.current,
+        );
+        const lh =
+          lineHeight === "normal"
+            ? parseFloat(fontSize) * 1.2
+            : parseFloat(lineHeight);
+        return lh;
+      }
+      return undefined;
+    }
   }
-
-  function getCalculatedRows(currentInputRef: HTMLElement) {
-    const currentRef = currentInputRef as HTMLTextAreaElement;
-    const textareaLineHeight = parseInt(
-      window.getComputedStyle(currentRef).getPropertyValue("line-height"),
-      10,
-    );
-    const textareaScrollHeight = currentRef.scrollHeight;
-    return calculateRows(textareaScrollHeight, textareaLineHeight);
-  }
-
   function insertText(text: string) {
     const input = inputRef.current;
     if (input) {
@@ -119,10 +114,6 @@ function InputTextInternal(
       props.onChange && props.onChange(input.value);
     }
   }
-}
-
-export function calculateRows(scrollHeight: number, lineHeight: number) {
-  return Math.floor(scrollHeight / lineHeight) - 1;
 }
 
 export const InputText = forwardRef(InputTextInternal);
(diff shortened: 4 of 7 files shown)


packages/components/src/InputText/InputText.tsx
```diff
@@ -1,7 +1,18 @@
...
+  Ref,
+  createRef,
+  forwardRef,
+  useImperativeHandle,
+  useState,
+} from "react";
 import { XOR } from "ts-xor";
 import { FormField, FormFieldProps } from "../FormField";
 
+interface MultiRows {
```
JeffMarvin, line 11: `RowRange` might be more accurate and clear for what this is doing.

packages/components/src/InputText/InputText.tsx
```diff
@@ -50,14 +61,54 @@ function InputTextInternal(
...
     />
   );
 
+  function initializeRows() {
+    if (!props.rows) return;
+    if (typeof props.rows === "number") {
+      return props.rows;
+    } else {
+      return props.rows.min;
+    }
```
seanhealy, line 82: I _think_ you might be able to simplify this with:

```suggestion
    if (typeof props.rows === "object") {
      return props.rows.min;
    }
    return props.rows;

```

packages/components/src/InputText/InputText.tsx
```diff
@@ -50,14 +61,54 @@ function InputTextInternal(
...
+  }
+
+  function handleChange(newValue: string) {
+    props.onChange && props.onChange(newValue);
+    if (!props.rows) return;
+    if (
+      typeof props.rows !== "number" &&
+      inputRef &&
+      inputRef.current instanceof HTMLTextAreaElement
+    ) {
```
seanhealy, line 92: Perhaps?

```suggestion
    if (
      props.rows &&
      typeof props.rows !== "number" &&
      inputRef &&
      inputRef.current instanceof HTMLTextAreaElement
    ) {
```

packages/components/src/InputText/InputText.tsx
```diff
@@ -27,7 +38,7 @@ interface MultilineProps extends BaseProps {
   /**
    * Specifies the visible height of a long answer form field.
```
JeffMarvin, line 39: Probably should update the comment to include the new option for this prop here.

packages/components/src/InputText/InputText.tsx
```diff
@@ -50,14 +61,54 @@ function InputTextInternal(
...
+      const calculatedRows = calculateRows(
+        textareaScrollHeight,
+        textareaLineHeight,
+      );
+      if (
+        calculatedRows >= props.rows.min &&
+        calculatedRows <= props.rows.max
+      ) {
+        setRows(calculatedRows);
+      }
```
JeffMarvin, line 108: This whole block could be a separate function. If you define it on line 111 (i.e. within the component scope) then it will have access to all of the scope that it currently does, so it shouldn't be too complicated.

JeffMarvin: Basically, you could pass in `inputRef` and return `calculatedRows` which you would then use in `setRows(calculatedRows)`.
Happy to pair on that change if you want!

JeffMarvin:
@tyler-santos Does this work? I am trying to get the rows to show up on some of these examples but I can't consistently get the right behaviour.  Can you post a quick QA plan for one way to show that this is working? 
https://deploy-preview-303--jobber-atlantis.netlify.app/components/input-text

JeffMarvin:
@tyler-santos Oh, also, I think we should add an example block in the relevant Readme file for the component that shows how this new prop is used.

tyler-santos:
> @tyler-santos Does this work? I am trying to get the rows to show up on some of these examples but I can't consistently get the right behaviour. Can you post a quick QA plan for one way to show that this is working?
> https://deploy-preview-303--jobber-atlantis.netlify.app/components/input-text

Yes it does work! Here is a little gif showing the behaviour. I will put in QA steps. 
Screen Recording 2020-07-10 at 2 19 42 PM

packages/components/src/InputText/InputText.tsx
tyler-santos: I think this would work as well but linter complains if I do it this way.

packages/components/src/InputText/InputText.tsx
tyler-santos: Good call thanks!

packages/components/src/InputText/InputText.tsx
tyler-santos: Chatted about this separately and got it resolved :)

seanhealy:
@tyler-santos Just gave it a test it looks like it grows upon adding text / lines but isn’t shrinking if that text is removed.

tyler-santos:
@GetJobber/frends I have requested a review on this, the old specs still need to be removed and new ones added but it would be good to get some feedback on the new implementation. Thank you very much!

crpahl:
Tyler is out today so I'll be paying attention to any requested changes and addressing them for now 😄

.gitignore
```diff
@@ -6,3 +6,4 @@ node_modules
 
 **/tsconfig.tsbuildinfo
 .DS_Store
+.idea
```
crpahl, line 9: I added this because I'm a dirty RubyMine user and it kept picking up changes to this RubyMine file.

packages/components/src/FormField/FormField.tsx
```diff
@@ -172,7 +172,7 @@ export const FormField = React.forwardRef(
       onValidation,
       placeholder,
       readonly,
-      rows,
+      rows = 3,
```
crpahl, line 175: Hmm is this change still needed?

I'm handling the default in `InputText.tsx`

```
  function getRowRange(): RowRange {
    if (props.rows === undefined) {
      return { min: 3, max: 3 };
    } else if (typeof props.rows === "object") {
      return { min: props.rows.min, max: props.rows.max };
    } else {
      return { min: props.rows, max: props.rows };
    }
  }
```

packages/components/src/InputText/InputText.tsx
```diff
@@ -52,12 +61,53 @@ function InputTextInternal(
...
+      return { min: 3, max: 3 };
+    } else if (typeof props.rows === "object") {
+      return { min: props.rows.min, max: props.rows.max };
+    } else {
+      return { min: props.rows, max: props.rows };
+    }
+  }
+
+  function resize(textArea: HTMLTextAreaElement) {
+    if (rowRange.min === rowRange.max) return;
```
crpahl, line 91: Return if no resizing is needed. This will handle the case where the values of rows is:

```
rows = {3}
```

or something like:

```
rows = {{min: 3, max: 3}}
```

packages/components/src/InputText/InputText.tsx
```diff
@@ -52,12 +61,53 @@ function InputTextInternal(
...
+  function textAreaHeight(textArea: HTMLTextAreaElement) {
+    const maxHeight = rowRange.max * getLineHeight(textArea);
+    return Math.min(textArea.scrollHeight, maxHeight);
+  }
+
+  function getLineHeight(textArea: HTMLTextAreaElement) {
+    const { fontSize, lineHeight } = window.getComputedStyle(textArea);
+    const lh =
+      lineHeight === "normal"
+        ? parseFloat(fontSize) * 1.2
```
crpahl, line 106: @eddysims

I got this from Eddy's example, where does the magic `1.2` number come from? Should this be a dynamically retrieved value from somewhere?

packages/components/src/FormField/FormField.css
```diff
@@ -72,7 +72,6 @@
 
 textarea.formField {
   height: auto;
-  min-height: calc(var(--field--height) * 2);
```
crpahl, line 75: I'm guessing this still needs to be removed? Sorry I wasn't around when this change was made.

crpahl:
I'm getting this weird resize behaviour and I'm not sure why. It's very slight...

resize-weirdness

crpahl:
Not sure if this is something I should have fixed now, but I noticed there's no padding at the bottom as the scroll area starts to expand. It looks like this is an existing problem on master too?

image

crpahl:
@seanhealy Ready for a follow-up review! I added a few comments and in-line comments for things I wasn't sure about.

.gitignore
seanhealy: 😂

packages/components/src/FormField/FormField.css
eddysims: Im pretty sure it does, in the case you want, `min-height` should be covered by the `rows` attribute

packages/components/src/InputText/InputText.tsx
```diff
@@ -25,9 +30,12 @@ interface MultilineProps extends BaseProps {
   readonly multiline: true;
 
   /**
-   * Specifies the visible height of a long answer form field.
+   * Specifies the visible height of a long answer form field. Can be in the
+   * form of a single number to set a static height, or an associative array
+   * with a min and max keys indicating the minimum number of visible rows, and
+   * the maximum number of visible rows.
```
eddysims, line 36: lets change `associative array` to `object`.

packages/components/src/InputText/InputText.tsx
eddysims: yeah, this isnt the best way... we can do something like

```
const lineHeight = window
        .getComputedStyle(inputRef.current)
        .getPropertyValue("line-height");
```

*note* I grabbed this from a slack message, so it might not be exactly correct, but it should be similar to what you want.

eddysims:
@crpahl back to you with some changes. 

One other note. because we are not taking padding/border into consideration on `resize()` we see a small shrink in the textarea on first keystroke when we have something like `rows={{min: 3, max: 4}}`. This is because we remove the bottom padding from the size calculation.

crpahl:
@eddysims In regards to this: https://github.com/GetJobber/atlantis/pull/303#issuecomment-661245148

If you look at the gif I posted here: https://github.com/GetJobber/atlantis/pull/303#issuecomment-659699978

The height change is also happening when `resize()` isn't getting called. So something else is causing the resize and I'm not sure what it is... I know I'm missing something obvious, but I'm not sure what it is.

packages/components/src/InputText/InputText.tsx
crpahl: Ah it's the line height. Makes sense now 👍

crpahl:
@eddysims I've addressed your changes. The weird shrinking no longer happens for this new change. Resizing still happens here though https://github.com/GetJobber/atlantis/pull/303#issuecomment-659699978 (the case where no range is given).

This seems to happen because `min-height: calc(var(--field--height) * 2);` was removed from the CSS and this seemed to be preventing the shrinking in that case.

eddysims:
@crpahl did some digging and it looks like the vertical padding is just not even when the input changes from normal to `miniLabel`.

With no text in the `textarea` we get a total of `28px` in vertical padding

image


With text it changes to `24px` in total

image

One way I have found to fix this is to change https://github.com/GetJobber/atlantis/blob/74166e0dd3bda7324c534709e67823241499a2ff/packages/components/src/FormField/FormField.css#L157 to `--field--padding-bottom: var(--space-small);`. I _think_ that will work without changing the other inputs

crpahl:
@eddysims Thanks, that seemed to fix it!

crpahl:
@eddysims Back to you :) Let me know if anything else needs updating!

PR #321, opened 2020-07-20
feat(components): Adds an onEnter handler for Form Fields

eddysims:
## Motivations

We need to be able to call a function when we hit the enter key on Input Fields.

## Changes

Adds an `onEnter` prop to FormFields that will all a call back for when the `Enter` key is hit.

### Added

- `onEnter` callback to FormFields

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.test.tsx
+++ b/packages/components/src/FormField/FormField.test.tsx
@@ -7,30 +7,31 @@ afterEach(cleanup);
 it("renders correctly with no props", () => {
   const tree = renderer.create(<FormField />).toJSON();
   expect(tree).toMatchInlineSnapshot(`
-        <div
-          className="wrapper"
-          style={
-            Object {
-              "--formField-maxLength": undefined,
-            }
-          }
-        >
-          <label
-            className="label"
-            htmlFor="123e4567-e89b-12d3-a456-426655440001"
-          >
-             
-          </label>
-          <input
-            className="formField"
-            id="123e4567-e89b-12d3-a456-426655440001"
-            onBlur={[Function]}
-            onChange={[Function]}
-            onFocus={[Function]}
-            type="text"
-          />
-        </div>
-    `);
+    <div
+      className="wrapper"
+      style={
+        Object {
+          "--formField-maxLength": undefined,
+        }
+      }
+    >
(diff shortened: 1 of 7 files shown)


packages/components/src/FormField/FormField.test.tsx
```diff
@@ -256,3 +262,30 @@ test("it should call the validation handler with a fail status when there's an e
 
   expect(validationHandler).toHaveBeenCalledWith("fail", "Nope!");
 });
+
+test("it should handle when the enter key is pressed", () => {
```
seanhealy, line 266: Might be good to test the `ctrl` & `shift` cases as well.

seanhealy, line 266: Might be good to test the `ctrl` & `shift` cases as well.

packages/components/src/FormField/FormField.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-statements */
```
seanhealy, line 1: Can we avoid having this?

eddysims: no :( not without some refactors I don't think. the `handleKeyDown` method causes it.

eddysims:
@seanhealy comments addressed and tests added. Back to you!

packages/components/src/FormField/FormField.tsx
seanhealy: Could you add a comment about it perhaps then? Also, perhaps use a disable next line version instead of the whole file?

eddysims:
@seanhealy added in a comment around max statements

kkimble006:
@eddysims and @seanhealy Just FYI, I am sure this is obvious this would happen, but in our PR when using this code, we also need to update a number of snapshots in Jobber (https://github.com/GetJobber/Jobber/pull/20810/files#diff-6049cc0be78a6dad21f1c0870b097fa5R29)

kkimble006:
@eddysims I accidentally clicked the "resolve conflict" button 🤦

PR #324, opened 2020-07-28
feat(components): modify autocomplete to show section headings, descriptions and details

linds14sr20det:
## Motivations

The autocomplete component needs the ability to group results under categories, and also provide better descriptions of the options that are returned to the user. In certain situations the checkmarks are redundant, as the form element and highlighting coupled with the search narrowing makes it clear what has been selected.

### Added

- The options array passed to an autocomplete can contain either a flat array of `options` as per the original functionality, or a nested structure of `GroupOptions`. Each group option's label becomes a header for the list. A group option contains an array of options which becomes the sub items within that list.
- A user can also add details and/or a description. These items will show up below the options label or beside the description.

### Changed

- The existing arrow up and arrow down functions will now skip headings.
- Browser autocomplete is off by default

## Testing

Most of these changes are visual, but specific testing should be centred around the arrow up/down functionality, the change handlers and getOptions functions. These should not regress. Aside from this all combinations of visual elements should work.

---

Random photo of Atlantis


JeffMarvin:
@linds14sr20det It's probably fine to leave, but for the future, we don't include any JIRA specific stuff like ticket ID's here, as this is an open source project. (I've also had to break myself of the finely trained habit of naming branches this way, lol.)

packages/components/src/Autocomplete/Autocomplete.mdx
```diff
@@ -154,3 +154,58 @@ using either promises or `await` in `getOptions`.
...
+You can specify an array of sections which contains options objects.
+
+<Playground>
+  {() => {
+    const options = [
+      { label: "Ships", heading: true },
+      {
+        value: 1,
+        label: "Sulaco",
+        description: "They mostly come at night, mostly.",
```
JeffMarvin, line 169: :chef_kissing_hand:

packages/components/src/Autocomplete/Autocomplete.mdx
```diff
@@ -154,3 +154,58 @@ using either promises or `await` in `getOptions`.
...
 </Playground>
+
+## Autocomplete with section headings, description and extra details
+
+You can specify an array of sections which contains options objects.
+
+<Playground>
+  {() => {
+    const options = [
+      { label: "Ships", heading: true },
```
JeffMarvin, line 165: The array format you have isn't very descriptive of the relationship between the heading and the sublist members. E.g. we would end up having to process the array to look for headings, and we don't know easily which members of the list. We might want a more direct structure, so that we can manage the sub-lists individually. (Maybe at one point we want to hide or disable some in the list based on heading?) 

I had originally thought of the data in this format, which captures more of the membership of the headings and the sublist options, and gives us some metadata in the heading level if we need it (eg. `order`).
```
options = [
  {
    label: "Ships",
    order: 0,
    options: [
      { value, label, description, details...}, 
      { value, label, description, details...}, 
    ]
  }, {
    label: "Planets",
    order: 1,
    options: [
      { value, label, description, details...}, 
      { value, label, description, details...}, 
    ]
  }
]
```

packages/components/src/Autocomplete/Autocomplete.mdx
```diff
@@ -154,3 +154,58 @@ using either promises or `await` in `getOptions`.
...
+      { value: 9, label: "Bespin" },
+      { value: 10, label: "Tatooine" },
+      { value: 11, label: "Enterprise" },
+      { value: 12, label: "Ariel" }
+    ];
+    const [value, setValue] = useState();
+    return (
+      <Autocomplete
+        value={value}
+        sections={true}
```
JeffMarvin, line 194: Hopefully we can find a structure for the data passed in (I guess through `getOptions`) - that means the component can automatically detect the difference. Then we can avoid having to specify it through a prop, and just have it be based on the data structure. Thoughts?

JeffMarvin: Oh... does sections actually get used? I think you might not even be using it...

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -52,6 +67,10 @@ export function Menu({
             </div>
             <div className={styles.text}>
               <Text>{option.label}</Text>
+              {option.details !== undefined && <span>{option.details}</span>}
```
JeffMarvin, line 70: I like this. I think that also means that lists without headings can also use these. If so, please add a quick example that shows this option for a normal list.

JeffMarvin: We might also want to think about the different layout cases and make sure we render a nice layout for each possible combination. If not, the other option is to only render `description` when we also have `details` or something like that.

JeffMarvin:
@linds14sr20det Your approach generally looks good - mostly had comments around the data structure. Happy to chat on a call if you have any questions.

JeffMarvin:
As per our discussion (and the commit history, haha) I'm cool to leave this structure as is for now until we get it out in the wild a bit more.

packages/components/src/Autocomplete/Autocomplete.mdx
linds14sr20det: Just recapping our IRL discussion here for everyone else's benefit @JeffMarvin.

I went down the path above and it doesn't quite make sense to do the above yet. While the data structure looks a lot nicer the drawbacks are:
- rendering becomes an O(n^2) operation (probably not a problem with limited numbers of headings)
- arrow key traversal of the drop down is much more complex
- backwards compatibility is harder as inferring the type of data structure being passed in isn't ideal. Having to pass an empty heading for lists without headings feels weird.

We are gonna keep an eye on this and see how it's use evolves over time and potentially move to the above data structure if future needs warrant it.

packages/components/src/Autocomplete/Autocomplete.mdx
linds14sr20det: Nope this is cruft! Sorry, gone now.

linds14sr20det:
@whereyoubinns @JeffMarvin this should be good for review now! Please and thanks!

packages/components/src/Autocomplete/Menu.tsx
linds14sr20det: Yeah that sounds, good, adding this all in to the mdx.

packages/components/src/Autocomplete/Autocomplete.css
```diff
@@ -17,6 +17,14 @@
   background-color: var(--color-white);
 }
 
+.options.medium {
+  max-height: 300px;
```
linds14sr20det, line 21: I really don't like this but unfortunately this is how we define our line item picker height in jobber online. This is a concession to maintain visual parity.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -16,11 +16,31 @@ interface AutocompleteProps {
...
+
+  /**
+   * Turn off checkmarks on selected result elements.
+   */
+  readonly checkmarks?: boolean;
+
+  /**
+   * Turn on browser autocomplete for the input.
+   */
+  readonly browserAutocomplete?: "off" | "";
```
whereyoubinns, line 42: even after adding this I still see the browser autocomplete in the dropdown 
image

packages/components/src/Autocomplete/__snapshots__/Autocomplete.test.tsx.snap
```diff
@@ -1,5 +1,123 @@
...
+    }
+  >
+    <label
+      className="label"
+      htmlFor="123e4567-e89b-12d3-a456-426655440009"
+    >
+      placeholder_name
+    </label>
+    <input
+      autoComplete="off"
```
whereyoubinns, line 22: is the autocomplete flag not working cause this is camel case?

linds14sr20det: If I change it to not be camel case I get type errors. 
Screen Shot 2020-07-29 at 3 42 34 PM

packages/components/src/Autocomplete/Autocomplete.css
JeffMarvin: I would add your comment as a CSS comment here. We'll wonder why later...

JeffMarvin:
@linds14sr20det I'll try to review this later today. Sorry, I missed the ping earlier.
Have you tested this in all supported browsers? Also, based on you mentioning keyboard support, I assume you're already testing for accessibility.

packages/components/src/Autocomplete/__snapshots__/Autocomplete.test.tsx.snap
linds14sr20det: We resolved this, some craziness with chrome's usage of the off and false keywords.

whereyoubinns:
this is looking good to me. There are no changes that need to be made for backwards compatibility. Tested across Edge, FF, Chrome, opera, safari. We're not supporting IE 11 right?

JeffMarvin:
@whereyoubinns We are currently supporting IE11. It only has to work, not look good.

packages/components/src/Autocomplete/Autocomplete.tsx
JeffMarvin: Why are these "off" and blank string? Seems like a weird way to describe a boolean...

packages/components/src/Autocomplete/Autocomplete.css
```diff
@@ -50,18 +60,37 @@
   outline-color: var(--color-outline);
 }
 
+.option.separator:not(:last-child) {
+  border-bottom: 1px solid rgb(225, 225, 225);
```
JeffMarvin, line 64: These should be variables from the palette, not hardcoded...

packages/components/src/Autocomplete/Autocomplete.css
```diff
@@ -50,18 +60,37 @@
...
+}
+
 .options:hover .option.active:not(:hover) {
   background-color: transparent;
   outline-color: transparent;
 }
 
+.autocomplete .heading:not(:first-child) {
+  padding-top: 0.5em;
```
JeffMarvin, line 73: Same thing here... vertical rhythms should use the CSS variables for spacing.

packages/components/src/Autocomplete/Autocomplete.css
```diff
@@ -17,6 +17,16 @@
...
 }
 
+/*Unfortunately this is how we define our line item picker height in jobber online.
+This is a concession to maintain visual parity.*/
+.options.medium {
+  max-height: 300px;
+}
+
+.options h5 {
+  border-bottom: 1px solid black;
```
seanhealy, line 27: we have border presets as well

seanhealy: https://github.com/GetJobber/atlantis/blob/master/packages/design/src/borders.css

seanhealy:
Few other little things here if that's ok?

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -16,11 +16,31 @@ interface AutocompleteProps {
...
+
   /**
    * Hint text that goes above the value once the form is filled out.
    */
   readonly placeholder: string;
 
+  /**
+   * Turn on separators between result elements.
+   */
+  readonly separators?: boolean;
```
seanhealy, line 32: What are these?

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -16,11 +16,31 @@ interface AutocompleteProps {
...
+
+  /**
+   * Turn off checkmarks on selected result elements.
+   */
+  readonly checkmarks?: boolean;
+
+  /**
+   * Turn on browser autocomplete for the input.
+   */
+  readonly browserAutocomplete?: string;
```
seanhealy, line 42: I don't think this should ever be on?

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -14,19 +15,26 @@ enum IndexChange {
 interface MenuProps {
   readonly visible: boolean;
   readonly options: Option[];
+  readonly menuSize: string | undefined;
```
seanhealy, line 18: I think `size` would be more conventional base on the props on our other components.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -16,11 +16,31 @@ interface AutocompleteProps {
...
 
+  /**
+   * Turn on separators between result elements.
+   */
+  readonly separators?: boolean;
+
+  /**
+   * Turn off checkmarks on selected result elements.
+   */
+  readonly checkmarks?: boolean;
```
seanhealy, line 37: What is the reason to remove these?

packages/components/src/Autocomplete/Autocomplete.mdx
seanhealy: Nesting here would make this function more like the Menu component and its headers. I think it might be worth a bit more discussion here.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -16,11 +16,31 @@ interface AutocompleteProps {
    */
   readonly value: Option | undefined;
 
+  /**
+   * Hint text that goes above the value once the form is filled out.
+   */
+  readonly menuSize?: "medium";
```
seanhealy, line 22: `"Base"` would be better here to follow the standard size for other Atlantis components.

Do we only accept one size here?

packages/components/src/Autocomplete/Autocomplete.tsx
whereyoubinns: image
They're the borders between rows

packages/components/src/Autocomplete/Autocomplete.tsx
whereyoubinns: can we just remove the option to turn it on altogether?

packages/components/src/Autocomplete/Autocomplete.tsx
whereyoubinns: for autocomplete being used as a search it doesn't fit our designs. 
ex. i search for a specific product the only result in the list is the item that fits the search, and when the drop down opens again the only results would be the search results 
The case where you select something from the list and then open the entire list again doesn't apply to the use case so the checkmarks don't make sense
[image](https://user-images.githubusercontent.com/49168885/88966964-6456d500-d26a-11ea-85c5-0196bbb665ac.png)

packages/components/src/Autocomplete/Autocomplete.tsx
whereyoubinns: yep only one size. Instead should we make this `fixedSize` or something?

seanhealy: if there is only ever one size can we remove the option altogether?

packages/components/src/Autocomplete/Autocomplete.tsx
seanhealy: If you clear the value (I assume you're doing this) after selecting the item from the menu the checkmark should also be removed?

packages/components/src/Autocomplete/Autocomplete.tsx
seanhealy: If there's no reason to enable it the I think we don't need it as a prop.

packages/components/src/Autocomplete/Autocomplete.tsx
seanhealy: What semantic difference is there between autocompletes with separators and ones without? What decides if this should be on or not?

packages/components/src/Autocomplete/Autocomplete.tsx
seanhealy: We'll loop in with Josh / Rebeca on Tuesday.

packages/components/src/Autocomplete/Autocomplete.mdx
seanhealy: We'll followup with Steve on this on Tuesday as well.

packages/components/src/Autocomplete/Autocomplete.tsx
seanhealy: We'll have separators show up implicitly if there are details or description.

packages/components/src/Autocomplete/Autocomplete.tsx
linds14sr20det: It's actually not a boolean as per the spec. Autocomplete can be se to a string that will match autocomplete types. It's a pretty wacky spec. https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

whereyoubinns:
After typing and one result is in the drop down, hitting enter autofills the input with the header instead of the row
drop down ex

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -21,6 +22,16 @@ interface AutocompleteProps {
...
 
+  /**
+   * Turn off checkmarks on selected result elements.
+   */
+  readonly checkmarks?: boolean;
+
+  /**
+   * Hint adjusts the interface to either have small or large spacing.
+   */
+  readonly inputSize?: FormFieldProps["size"];
```
seanhealy, line 33: `size` would be more conventional here if it's controlling the size of the input.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -21,6 +22,16 @@ interface AutocompleteProps {
    */
   readonly placeholder: string;
 
+  /**
+   * Turn off checkmarks on selected result elements.
+   */
+  readonly checkmarks?: boolean;
+
+  /**
+   * Hint adjusts the interface to either have small or large spacing.
```
seanhealy, line 31: What is this prop doing?

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -32,15 +41,31 @@ interface AutocompleteProps {
    * is retuned from this method to the user as available options.
    * @param newInputText
    */
-  getOptions(newInputText: string): Option[] | Promise<Option[]>;
+  getOptions(
+    newInputText: string,
+  ): OptionCollection | Promise<OptionCollection>;
+
+  /**
+   * Optional additional blur behaviour (clicking away from the input text or
```
seanhealy, line 49: We don't need to indicate that this is optional in test here as the `?` on the param shows that. The docs viewer will use that to indicate that this is optional.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -32,15 +41,31 @@ interface AutocompleteProps {
...
+   * hitting escape)
+   */
+  onBlur?(): void;
+
+  /**
+   * Optional additional focus behaviour (clicking on the input text)
```
seanhealy, line 55: We don't need to indicate that this is optional in test here as the `?` on the param shows that. The docs viewer will use that to indicate that this is optional.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -21,6 +25,11 @@ interface AutocompleteProps {
    */
   readonly placeholder: string;
 
+  /**
+   * Adjusts the input text box to either have small or large height.
+   */
+  readonly size?: FormFieldProps["size"];
```
seanhealy, line 31: Since this is optional we should specify what it defaults to. Probably `base`?

packages/components/src/FormField/FormField.tsx
```diff
@@ -12,6 +12,11 @@ export interface FormFieldProps {
    */
   readonly align?: "center" | "right";
 
+  /**
+   * Determines if browser form autocomplete is enabled.
+   */
+  readonly autocomplete?: string;
```
seanhealy, line 18: I think a `boolean` would make more sense here.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -77,12 +104,22 @@ export function Autocomplete({
...
     if (newText) {
-      debouncedSetOptions(await getOptions(newText));
+      debouncedSetOptions(mapResultsToOptions(await getOptions(newText)));
     } else {
-      setOptions(initialOptions);
+      setOptions(mapResultsToOptions(initialOptions));
     }
   }
 
+  function mapResultsToOptions(items: AnyOption[]) {
```
seanhealy, line 113:
```suggestion
  function mapToOptions(items: AnyOption[]) {
```

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -91,6 +128,7 @@ export function Autocomplete({
 
   function handleInputChange(newText: string) {
     updateInput(newText);
+    onChange({ label: newText });
```
seanhealy, line 131: I don't think we want to call this here as it will fire constantly as they type regardless of a valid option being chosen.

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -38,7 +44,17 @@ export function Menu({
       {options.map((option, index) => {
         const optionClass = classnames(styles.option, {
           [styles.active]: index === highlightedIndex,
+          [styles.separator]: addSeparators,
         });
+        if (isGroup(option)) {
+          return (
+            <div key={option.label} className={styles.heading}>
+              <Typography element="h5" fontWeight="bold">
```
seanhealy, line 52: Can we use a [`Heading`](https://atlantis.getjobber.com/packages-components-src-heading-heading) here instead?

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -51,7 +67,19 @@ export function Menu({
               )}
             </div>
             <div className={styles.text}>
-              <Text>{option.label}</Text>
+              <div className={styles.label}>
+                <Text>{option.label}</Text>
+                {option.description !== undefined && (
+                  <Typography textColor="greyBlueDark">
```
seanhealy, line 73: Why can't this be `Text`? Everything looks to be the same colour.

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -66,22 +94,32 @@ export function Menu({
   function setupKeyListeners() {
     useOnKeyDown("ArrowDown", (event: KeyboardEvent) => {
       if (!visible) return;
-
       event.preventDefault();
+
+      const requestedIndex = options[highlightedIndex + IndexChange.Next];
```
seanhealy, line 99: Can we collapse the common functionality between `ArrowDown` and `ArrowUp` into a single method that these can then call. there is a decent amount of duplication between the two.

packages/components/src/Autocomplete/Autocomplete.tsx
seanhealy: This method can also be pulled up to the module level since it doesn't depend on anything within the component.

seanhealy:
I noticed when testing that if there are section headers we do not select the first valid option as you type like we do if there are not headers.

packages/components/src/Autocomplete/Autocomplete.css
```diff
@@ -7,16 +7,24 @@
...
   z-index: var(--elevation-menu);
   width: 100%;
+  /* Unfortunately this is how we define our line item picker height in jobber online.
+  This is a concession to maintain visual parity. */
+  max-height: 300px;
   box-shadow: var(--shadow-base);
   box-sizing: border-box;
   margin: 0;
-  padding: var(--space-small);
+  padding: 0;
```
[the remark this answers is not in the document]
linds14sr20det: Screen Shot 2020-08-18 at 8 52 43 AM
vs.
Screen Shot 2020-08-18 at 8 52 26 AM

I think we want the full width separators which padding 0 provides.

packages/components/src/Autocomplete/Autocomplete.tsx
linds14sr20det: Just as a heads up we don't specify input text's default size.

packages/components/src/FormField/FormField.tsx
linds14sr20det: This one caught me out too. Autocomplete as used by browsers isn't a boolean. You can actually specify types to autocomplete on. https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete
The value section is extensive. If we used a boolean we would have to convert it to a string of password to turn it off for all browsers. I think leaving it as a string is better.

packages/components/src/Autocomplete/Autocomplete.tsx
seanhealy: Ah, that's one of the old ones then. 😢 Don't worry about it here then.

PR #331, opened 2020-08-12
fix(components): Make Toast variation optional

eddysims:
## Motivations

The `Toast` should default the variation prop, which means it should be optional. Success is also the most common type of `Toast` so that should be the default.

## Changes

Makes the `variation` prop of toast optional and default to `success`

### Added

- 

### Changed

- variation default is now `success`

### Deprecated

- 

### Removed

- 

### Fixed

- makes variation prop optional

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Toast/Toast.test.tsx
+++ b/packages/components/src/Toast/Toast.test.tsx
@@ -16,6 +16,11 @@ afterEach(() => {
   document.body.innerHTML = ``;
 });
 
+const successMessage =
+  "Successful Message that should last the full 5 seconds, it just needs to be 50 characters long";
+const infoMessage = "Bland Toast";
+const errorMessage = "Errorful should last inbetween min-max";
+
 it("creates the placeholder div on showToast call", () => {
   const { getByText } = render(<MockToast />);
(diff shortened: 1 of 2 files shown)


packages/components/src/Toast/Toast.test.tsx
```diff
@@ -29,19 +29,23 @@ it("creates the placeholder div on showToast call", () => {
...
+it("renders a Slice of Toast when the 'showToast' method is called", () => {
   const { getByText, getByTestId } = render(<MockToast />);
 
-  fireEvent.click(getByText("No Variation"));
-  expect(getByText("Bland Toast")).toBeInstanceOf(HTMLSpanElement);
-  expect(getByTestId("knot")).toBeInstanceOf(SVGElement);
+  fireEvent.click(getByText("Success"));
+  expect(
+    getByText(
+      "Successful Message that should last the full 5 seconds, it just needs to be 50 charachters long",
```
JeffMarvin, line 38:
```suggestion
      "Successful Message that should last the full 5 seconds, it just needs to be 50 characters long",
```

eddysims:
@JeffMarvin I had to push some test fixes, could you take another look

PR #332, opened 2020-08-13
feat(components): new FormatFile component

JeffMarvin:
### Added

- New component to represent a file through various states (uploaded, uploading) and to support a delete operation.
  - Supports uploading and uploaded state. (Not error state yet.)
  - Uploading state does not currently allow a "cancel" operation.

### Changed

- Button can now support the combination of the following props together: `type="tertiary"` and `variation="destructive"`

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/package.json
```diff
@@ -32,6 +32,7 @@
     "@use-it/event-listener": "^0.1.3",
     "classnames": "^2.2.6",
     "color": "^3.1.2",
+    "filesize": "^6.1.0",
```
JeffMarvin, line 35: This is used to give us human readable sizes based on the file size provided in bytes.

packages/components/src/FormatFile/FormatFile.mdx
```diff
@@ -0,0 +1,60 @@
...
+        key: "abc",
+        name: "myballisbigandroundIamrollingitontheground.png",
+        type: "image/png",
+        size: 213402324,
+        progress: 0.3,
+        src: () =>
+          Promise.resolve(
+            "https://res-1.cloudinary.com/crunchbase-production/image/upload/c_thumb,h_256,w_256,f_auto,g_faces,z_0.7,q_auto:eco/ywedismk5vf5uqc2nngg"
+          )
+      }}
```
seanhealy, line 42: We could pull this out of the playground here and document it separately perhaps? Just to keep this example simpler? If you think this makes sense I can do a quick share and show you how.

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -0,0 +1,121 @@
+import React, { useState } from "react";
+import filesize from "filesize";
+import { IconNames } from "@jobber/design";
+import styles from "./FormatFile.css";
+import { Button } from "../Button";
+import { Icon } from "../Icon";
+import { Typography } from "../Typography";
+import { ProgressBar } from "../ProgressBar";
+
+interface FileUpload {
```
seanhealy, line 10: I wonder if we should have a home for types like this that are component agnostic… 🤔

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -0,0 +1,121 @@
...
+    switch (mimeType) {
+      case "application/pdf":
+        return "pdf";
+      case "application/vnd.ms-excel":
+        return "excel";
+      default:
+        return "file";
+    }
+  }
+}
```
seanhealy, line 121: These two can probably be pulled up to the module level to keep the component simpler.

```suggestion
}

function getHumanReadableFileSize(sizeInBytes: number): string {
  return filesize(sizeInBytes);
}

function getIconNameFromType(mimeType: string): IconNames {
  if (mimeType.startsWith("image/")) return "camera";
  if (mimeType.startsWith("video/")) return "video";

  switch (mimeType) {
    case "application/pdf":
      return "pdf";
    case "application/vnd.ms-excel":
      return "excel";
    default:
      return "file";
  }
}
```

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -0,0 +1,121 @@
...
+            <ProgressBar
+              size="small"
+              currentStep={file.progress * 100}
+              totalSteps={100}
+            />
+          </div>
+        )}
+      </div>
+      <div className={styles.contentBlock}>
+        <Typography element="span" fontWeight="semiBold">
```
seanhealy, line 83: Adding a new weight for this isn't bad but, I think we should confirm that our existing weights can't work here as we don't use semi bold anywhere else?

JeffMarvin: I was surprised that we didn't have anything for this weight... the closest size we had  is clearly bold. I'll run this by Rebecca though.

packages/components/src/FormatFile/FormatFile.mdx
JeffMarvin: Which part? I think its actually useful to have this structure laid out here in the example doc.

packages/components/src/FormatFile/FormatFile.tsx
JeffMarvin: For now, I pulled it out into its own file. We can decide where it should live... although a central place makes sense.

packages/components/src/FormatFile/FormatFile.mdx
seanhealy: Why don't we talk in the morning? I think it may be easier?

packages/components/src/FormatFile/FormatFile.mdx
```diff
@@ -0,0 +1,57 @@
...
+        type: "application/pdf",
+        size: 2134023,
+        progress: 1
+      }}
+      onDelete={() => alert("🧨BALEETED!💥")}
+    />
+  </Content>
+</Playground>
+
+## Props
```
seanhealy, line 55: @JeffMarvin / @rebecca-li Any Design & Usage Guidelines you think would be good to add in here?

https://github.com/GetJobber/atlantis/blob/master/packages/generators/templates/component/%7B%7Bname%7D%7D.%7B%7Bmdx%7D%7D#L30

packages/components/src/FormatFile/FormatFile.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    type: "application/pdf",
+    size: 1022,
+    progress: 1,
+  };
+  const deleteHandler = jest.fn();
+  const { getByLabelText } = render(
+    <FormatFile onDelete={deleteHandler} file={testFile} />,
+  );
+
+  act(() => {
```
seanhealy, line 46: Were you getting a warning without the `act` here? I'm pretty sure the updated `fireEvent` should handle the `act` internally.

packages/components/src/FormatFile/FormatFile.mdx
rebecca-li: - FormatFile components should take up the full width of the parent container (1 or 2 files per row)

packages/components/src/FormatFile/FormatFile.tsx
JeffMarvin: I ended up taking it out, it didn't change the styling enough to justify.

packages/components/src/FormatFile/FormatFile.test.tsx
JeffMarvin: Fixed this with `waitFor` and `done`.

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -0,0 +1,91 @@
...
+            <ProgressBar
+              size="small"
+              currentStep={file.progress * 100}
+              totalSteps={100}
+            />
+          </div>
+        )}
+      </div>
+      <div className={styles.contentBlock}>
+        <Typography element="span">{file.name}</Typography>
```
seanhealy, line 55: Does `Text` work here?

PR #344, opened 2020-08-24
feat(components): Disable RadioOption component

eddysims:
## Motivations

When using a `RadioGroup` component there may be options that you do not want the user to be able to select. This PR will allow for disabling `RadioOption`

## Changes

Allows for disabling options in `RadioGroup`

### Added

- `disabled` prop to the `RadioOption` component
- adds focus styling to the `RadioOption` component

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/RadioGroup/RadioGroup.css
+++ b/packages/components/src/RadioGroup/RadioGroup.css
@@ -1,12 +1,12 @@
+.radioGroup {
+  display: inline-flex;
+  flex-direction: column;
+}
+
 .input {
   /* Hide checkbox on UI but not screen readers and still allow focus state */
   position: absolute;
-  width: 1px;
-  height: 1px;
-  overflow: hidden;
-  clip: rect(0 0 0 0);
-  clip-path: inset(100%);
-  white-space: nowrap;
+  left: -999em;
 }
 
 .label {
@@ -28,13 +28,20 @@
   transition: all 0.2s;
 }
 
+.input:focus + .label:before {
+  box-shadow: 0 0 1px 0.125rem var(--color-outline);
+}
+
 .input:checked + .label::before {
   border-color: var(--color-green--dark);
   background-color: var(--color-green);
-  transition: all 0.2s;
 }
 
-.radioGroup {
-  display: inline-flex;
-  flex-direction: column;
+.input[disabled] + .label {
+  color: var(--color-blue--lighter);
+}
+
+.input[disabled] + .label::before {
+  border-color: var(--color-grey--light);
+  background-color: var(--color-grey--lighter);
 }
--- a/packages/components/src/RadioGroup/RadioGroup.css
+++ b/packages/components/src/RadioGroup/RadioGroup.css
@@ -6,7 +6,7 @@
 .input {
   /* Hide checkbox on UI but not screen readers and still allow focus state */
   position: absolute;
-  left: -999em;
+  left: -999vw;
 }
 
 .label {
--- a/packages/components/src/RadioGroup/RadioGroup.mdx
+++ b/packages/components/src/RadioGroup/RadioGroup.mdx
@@ -6,8 +6,9 @@ route: /components/radio-group
 
 import { Playground, Props } from "docz";
 import { ComponentStatus } from "@jobber/docx";
-import { RadioGroup } from ".";
-import { RadioOption } from ".";
+import { RadioGroup, RadioOption } from ".";
+import { Content } from "@jobber/components/Content";
+import { Checkbox } from "@jobber/components/Checkbox";
 import { useState } from "react";
 
 # Radio Group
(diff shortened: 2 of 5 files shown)


packages/components/src/RadioGroup/RadioGroup.css
```diff
@@ -1,12 +1,12 @@
...
 .input {
   /* Hide checkbox on UI but not screen readers and still allow focus state */
   position: absolute;
-  width: 1px;
-  height: 1px;
-  overflow: hidden;
-  clip: rect(0 0 0 0);
-  clip-path: inset(100%);
-  white-space: nowrap;
+  left: -999em;
```
seanhealy, line 9: Is there any chance this does any odd scrolling. Also, not problem with what you have here but I wonder if `vw` would be safer for those people with REALLY REALLY REALLY wide displays? :p

packages/components/src/RadioGroup/RadioGroup.mdx
```diff
@@ -14,7 +14,8 @@ import { useState } from "react";
 
 <ComponentStatus stage="ready" responsive="yes" accessible="yes" />
 
-RadioGroup will allow users to choose from one of a list of provided options.
+RadioGroup will allow users to choose a single selection from one of a list of
```
JeffMarvin, line 17: Should we rename this component `InputRadio` or something? Might be good to get ahead of that before this is used in too many places... Just a thought...

packages/components/src/RadioGroup/RadioGroup.mdx
```diff
@@ -33,6 +34,30 @@ import { RadioGroup } from "@jobber/components/RadioGroup";
...
+
+## RadioOption Props
+
+<Props of={RadioOption} />
+
+## Disabled Options
+
+`RadioOptions` can be disabled by passing in the disabled prop. Disabling an
+option will prevent a user from being able to select the option.
+
```
JeffMarvin, line 49: I wonder if it is worth mentioning that this can be set dynamically, not just when the form loads....

seanhealy: It would be quite unusual for a prop to not allow this. I'm not sure it's necessary…?

packages/components/src/RadioGroup/RadioGroup.mdx
seanhealy: I like this idea. I think it's only used in one other place in Jobber right now.
I think lets make a followup ticket for that though perhaps? I guess it would be quick to do though…?

eddysims: I also though of this... @seanhealy whats your opinion?

packages/components/src/RadioGroup/RadioGroup.css
eddysims: Moving left off screen doesn't cause scroll. I'll make the `vw` change for those users with displays greater than 15984px lol.

packages/components/src/RadioGroup/RadioGroup.mdx
eddysims: I will make the playground a little better for this.

packages/components/src/RadioGroup/RadioGroup.css
seanhealy: Not only do they need a window that wide but the checkbox would have to be _right_ on the right edge of the window. But, I think that we'll get a nice thank you letter from a wide screen user thanking you for this at some point. 😂

packages/components/src/RadioGroup/RadioGroup.mdx
seanhealy: Would you be OK doing the rename in a followup PR? As the breaking change will be a _bit_ slower to bring in?

eddysims:
@seanhealy @JeffMarvin Comments addressed. Back to you!

eddysims:
@rebecca-li do you want to quickly take a look at the disabled and focus states of the RadioGroup?

rebecca-li:
Disabled
- remove the 'pointing' cursor from showing up on the disabled option
- wondering if we have a standard color we use for text of disabled elements? In some parts of Jobber we use `Grey Blue Dark` with lowered opacity, in our Figma library we use `Grey Blue Lighter` 

Maybe @chris-at-jobber will know more?

packages/components/src/RadioGroup/RadioGroup.css
```diff
@@ -28,13 +28,20 @@
...
   border-color: var(--color-green--dark);
   background-color: var(--color-green);
-  transition: all 0.2s;
 }
 
-.radioGroup {
-  display: inline-flex;
-  flex-direction: column;
+.input[disabled] + .label {
+  color: var(--color-blue--lighter);
```
chris-at-jobber, line 41: @rebecca-li I think to your comment, we should standardize this and rather than choosing from our full palette on every disabled element, introduce a `--color-disabled` that uses one of our colours. In [this proposal](https://www.figma.com/file/FLvj5SVGr0t40ErSf5hq4m/Design-System-Contribution-Colour-at-Jobber?node-id=30%3A581) I suggest a colour that's fairly close to `var--blue-lighter` for inactive states, so for this case (and going forwards) we could introduce the concept of an inactive colour to Atlantis similar to how we have an `error` colour at the moment.

chris-at-jobber: To my knowledge, we'd want to check Atlantis' `Checkbox`, `Button`, `Switch`, and possibly `Text` to see if there would be modifications needed there. I think in SG-1 we usually take that `60% or 70% of another grey` approach

PR #347, opened 2020-08-27
docs(components): Update Text Proposal

eddysims:
## Motivations

We need to make some modifications to the `Text` component. This is a proposal for how to most effectively make those changes.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/proposals/Text/Text_Mods.md
+++ b/docs/proposals/Text/Text_Mods.md
@@ -1,0 +1,51 @@
+# Modify `Text` to include new props
+
+Currently the `Text` component has one prop: `variation`. This prop is used to
+handle the style (size and color) of the text that is rendered. The issue that
+we are finding is that there is currently no way to combine styles.
+
+**Example:**
+
+We want to have small feedback text, and base sized feedback text. This
+currently isn't possible with the current `Text` component.
+
+## Current Props
+
+| prop      | type   | options                                                           |
+| --------- | ------ | ----------------------------------------------------------------- |
+| variation | string | "success", "error", "default", "subdued", "intro", "warn", "info" |
+
+## Proposed Solutions
+
+I think that there are 2 ways to handle this.
+
+### Variation takes an array
+
+The variation prop could be changed to take an array of strings. This would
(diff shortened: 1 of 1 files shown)


docs/proposals/Text/Text_Mods.md
```diff
@@ -0,0 +1,51 @@
...
+In this case the new props could look like this:
+
+| prop      | type   | options                                       |
+| --------- | ------ | --------------------------------------------- |
+| size      | string | "small", "base", "large"                      |
+| variation | string | "success", "error", "subdued", "warn", "info" |
+
+## Opinion
+
+In my opinion, the `Multiple Props` is the way to go. I think that the breaking
```
JeffMarvin, line 49: @eddysims Agreed. I wonder if we should choose something other than variation and then just deprecate that prop? We could map the existing variations over to use the new props....

eddysims: @JeffMarvin I actually dont think that variation is that bad. We currently use it in Button and Toast. thoughts?

JeffMarvin: Yeah... I just went through and looked at those Atlantis components, and I think `variation` is what we want. 
What's the scope of the breaking change?

docs/proposals/Text/Text_Mods.md
```diff
@@ -0,0 +1,51 @@
...
+styling with the `error` styles and then over riding them with the `info`
+styles.
+
+### Multiple Props
+
+Text could take 2 props instead of one. `size` and `variation`. This would allow
+for us to do things like `<Text size="small" variation="info">`.
+
+My concern with this is that it may result as a breaking change. The reason
+being is that the `intro` variation would need to be moved into the `size` prop.
```
seanhealy, line 38: We could keep `default` and `intro` in as variations and deprecate them for now. Then we can clean them up with a breaking change later perhaps?

PR #352, opened 2020-09-09
feat(components): BREAKING CHANGE: FormField Validation

eddysims:
## Motivations

We will soon need to be able to have a form component available. For this form component we will also need to know the state of the form that we are using.

We want to be able to validate forms and fields more easily. To do this we are introducing `react-hook-form` to Atlantis. 

This PR will allow us to validate form fields using a `validations` prop.

## Changes

This PR introduces 5 new things.

1. a `Form` component
2. a `useFormState` hook that can be used to get the state of a form.
3. a `submit` prop on `Button` that allows it to submit a form.
4. a refactor of the `validations` prop on `FormField` component
5. re-styling of `InputValidation` component

### Added

- `Form` component
- `useFormState` hook
- a `submit` prop on `Button` component to render as `type="submit"`
- react-hook-forms
- validations prop to `InputNumber`

### Changed

- Some eslint and gatsby node rules to recognize the `@jobber/hooks` package
- validations prop on `FormField` components to a new object shape
- moved `InputValidation` below the `FormField` component
- `FormField` from `forwardRef` to having an `inputRef` prop

### Deprecated

- 

### Removed

- `errorMessage` prop from `FormField` component
- `onValidate()` prop from `FormField` component

### Fixed

- Small styling change around `FormField` validation to keep it associated with the field when used in `Content`

### Security

- 

## Testing

---

Random photo of Atlantis


eddysims:
@rebecca-li if you want to take a look

packages/components/src/FormField/FormField.tsx
```diff
@@ -146,145 +138,150 @@ export interface FormFieldProps {
...
   /**
-   * **DEPRECATED** Use `onValidation` prop instead.
-   *
    * Callback to get the the status and message when validating a field
-   * @param status
-   * @param message
+   * @param messages
    */
-  onValidate?(status: "pass" | "fail", message: string): void;
+  onValidation?(message: string | undefined): void;
```
seanhealy, line 144: Is `message` undefined if the input is valid?

packages/components/src/FormField/FormField.tsx
```diff
@@ -146,145 +138,150 @@ export interface FormFieldProps {
...
+    register,
+    /**
+     * We currently only use `errors` from formState, but there are lots of
+     * other values that come with it. https://react-hook-form.com/api#formState
+     */
+    formState: { errors },
+  } =
+    useFormContext() != undefined
+      ? useFormContext()
+      : useForm({ mode: "onBlur" });
```
seanhealy, line 197: What do you think about `onTouched` here instead? I _think / hope_ this would mean that it is `onBlur` the first time then `onChange` for subsequent changes?

packages/components/src/FormField/FormField.tsx
```diff
@@ -294,67 +291,69 @@ export const FormField = React.forwardRef(
               onFocus={handleFocus}
               onKeyDown={event => handleKeyDown(event)}
               onBlur={handleBlur}
-              ref={ref as Ref<HTMLInputElement>}
+              ref={e => {
+                if (inputRef) {
+                  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+                  // @ts-ignore
```
seanhealy, line 297: Now that we're on `inputRef` do we still need the exception here?

packages/components/src/FormField/FormField.tsx
```diff
@@ -146,145 +138,150 @@ export interface FormFieldProps {
...
+      case "select":
+        return <select {...fieldProps}>{children}</select>;
+      case "textarea":
+        return (
+          <textarea
+            rows={rows}
+            onFocus={handleFocus}
+            onKeyDown={event => handleKeyDown(event)}
+            onBlur={handleBlur}
+            ref={inputRef as Ref<HTMLTextAreaElement>}
```
seanhealy, line 278: I wonder if we should allow for validations on textareas as well?

seanhealy: If it's tricky we could do that as a followup?

packages/components/src/InputValidation/InputValidation.mdx
```diff
@@ -24,27 +24,23 @@ import { InputValidation } from "@jobber/components/InputValidation";
...
           My name is
           <InputText
-            validations={[
-              {
-                shouldShow: value !== "Jeff",
-                message: "Have you considered a better name, like Jeff?",
-                status: "error"
+            validations={{
+              pattern: {
+                value: /Jeff/g,
```
seanhealy, line 36: Should we have the `g` in here? I think we want to match just a single `Jeff` no?

packages/components/src/InputValidation/InputValidation.tsx
```diff
@@ -3,34 +3,15 @@ import { AnimatePresence, motion } from "framer-motion";
...
-  shouldShow?: boolean;
-}
-
 interface InputValidationProps {
   /**
-   * Array of validation messages
+   * Validation message to be displayed
    */
-  messages: ValidationProps[];
+  message: string;
```
seanhealy, line 10: If `onValidation` can call with `undefined` I wonder if this should also accept `undefined`?

seanhealy (changes requested):
Liking it so far! This came together quite nicely. Few little thoughts / ideas your way.

packages/components/src/FormField/FormField.tsx
eddysims: I think this was in here when I was learning. Not needed

packages/components/src/InputValidation/InputValidation.tsx
eddysims: This should always accept a string. The other undefined was not needed.

rebecca-li:
A few notes:

- Since invalid states should always include an error message, should we show them in these examples?
Invalid
Invalid

- In the example for inputNumber, should we change the placeholder/mini-label to Number (or something else)? I'm afraid this will give people the idea that the mini-label can be used for helper text or error messages.

- Have we considered having the error message disappear once corrected; without having to tab away? Once a field has an error, can validation checks be made after each keystroke, and the error messaging going away as soon as the input becomes valid?
Have you considered a better name, like Jeff
VALIDATION MESSAGE

- In the [form validation proposa](https://www.figma.com/file/EOYVyc9P30AzVue3XELCKi/Design-System-Contribution-Form-Validation?node-id=30%3A581)l, we had error text appearing below the field. Not sure if this PR is to address the proposal or if that's happening in another ticket.
￼

eddysims:
@rebecca-li @seanhealy 

> Since invalid states should always include an error message, should we show them in these examples?

If this is the case, i _think_ the invalid prop can be removed altogether. Can for sure confirm that an Input can not be invalid without a message?

> In the example for inputNumber, should we change the placeholder/mini-label to Number (or something else)? I'm afraid this will give people the idea that the mini-label can be used for helper text or error messages.

Done 👍 

> Have we considered having the error message disappear once corrected; without having to tab away? Once a field has an error, can validation checks be made after each keystroke, and the error messaging going away as soon as the input becomes valid?

This was addressed by one of @seanhealy's requests as well

> In the form validation proposal, we had error text appearing below the field. Not sure if this PR is to address the proposal or if that's happening in another ticket.

Done 👍

eddysims:
@rebecca-li @seanhealy now this is ready for re-review https://deploy-preview-352--jobber-atlantis.netlify.app/

rebecca-li:
- Error before getting a chance to type something in:
My name is

- More space between the icon and the text:
Screen Shot 2020-09-14 at 2 21 08 PM

- If we want to include the icon, I think a smaller variation would look cleaner. Example from proposal:
Screen Shot 2020-09-14 at 2 19 02 PM

We can go without the icon until a smaller version is available

packages/components/src/FormField/FormField.tsx
```diff
@@ -146,145 +137,158 @@ export interface FormFieldProps {
...
+      case "textarea":
+        return (
+          <textarea
+            rows={rows}
+            onFocus={handleFocus}
+            onKeyDown={event => handleKeyDown(event)}
+            onBlur={handleBlur}
+            ref={e => {
+              /**
+               * @typescript-eslint/ban-ts-ignore is ignored as `.current` is readonly
```
seanhealy, line 278: Would this be required if inputRef was a `MutableRefObject`? Or does that not work for other reasons?

seanhealy (changes requested):
Thought about the exception here happy to chat if you would like. :)

packages/components/src/FormField/FormField.tsx
eddysims: I tried `MutableRefObject` and it causes errors in other files. Example is in `InputText` we get errors for things like `.blur() is not assignable`.

eddysims:
@rebecca-li @seanhealy ready for re-review

rebecca-li:
LGTM!

packages/components/src/FormField/FormField.tsx
```diff
@@ -146,145 +138,156 @@ export interface FormFieldProps {
   onBlur?(): void;
 
   /**
-   * **DEPRECATED** Use `onValidation` prop instead.
-   *
    * Callback to get the the status and message when validating a field
-   * @param status
-   * @param message
+   * @param messages
```
seanhealy, line 142:
```suggestion
   * @param message
```

seanhealy (approved):
One REALLY tiny comment typo. I'll approve but you can still fix it if you like then, feel free to keep approval.

PR #355, opened 2020-09-14
feat(components): modify autocomplete to scroll list up and down

dmcrorieGIT:
## Motivations

Autocomplete highlighting behaviour must be changed as highlighted option falls out of view when scrolling too far down or too far up

### Changed

- When list item falls out of view on menu list, the list will scroll to keep that item in view
autocomplete-scrolling

## Testing

- [ ] change the `options` in the mdx file for `Autocomplete` to include a bunch of options with labels that start with the same letter
- [ ] In the autocomplete playground, type in that first letter that most of your options begin with
- [ ] Start scrolling down with the arrow keys
- [ ] When you get to the bottom of the list and keep going, ensure that the list scrolls so that the highlighted item never falls out of view
- [ ] Ensure the same for scrolling up

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Autocomplete/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu.tsx
@@ -29,7 +29,7 @@ export function Menu({
   const optionMenuClass = classnames(styles.options, {
     [styles.visible]: visible,
   });
-  const menuDiv = useRef<HTMLDivElement>();
+  const menuDiv = useRef() as React.MutableRefObject<HTMLDivElement>;
 
   const detectSeparatorCondition = (option: Option) =>
     option.description || option.details;
@@ -45,12 +45,7 @@ export function Menu({
   useEffect(() => setHighlightedIndex(initialHighlight), [options]);
 
   return (
-    <div
-      className={optionMenuClass}
-      ref={element => {
-        menuDiv.current = element as HTMLDivElement;
-      }}
-    >
+    <div className={optionMenuClass} ref={menuDiv}>
       {options.map((option, index) => {
         const optionClass = classnames(styles.option, {
           [styles.active]: index === highlightedIndex,
--- a/packages/components/src/Autocomplete/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu.tsx
@@ -116,17 +116,16 @@ export function Menu({
       direction: "up" | "down",
     ) {
       const itemDiv = menuDivElement.querySelector(
-        `button.${classnames(styles.option)}:nth-child(${highlightedIndex +
-          1})`,
+        `button.${styles.option}:nth-child(${highlightedIndex + 1})`,
       ) as HTMLButtonElement;
-      if (!itemDiv) {
-        return;
-      }
+      if (!itemDiv) return;
       const menuTop = menuDivElement.getBoundingClientRect().top;
-      const itemTop = itemDiv.getBoundingClientRect().top;
-      const itemHeight = itemDiv.getBoundingClientRect().height;
-      const itemTrueBottom =
-        itemDiv.getBoundingClientRect().bottom + itemHeight;
+      const {
+        top: itemTop,
+        height: itemHeight,
+        bottom: itemBottom,
+      } = itemDiv.getBoundingClientRect();
+      const itemTrueBottom = itemBottom + itemHeight;
       const menuBottom = menuDivElement.getBoundingClientRect().bottom;
       if (direction == "up" && itemTop - itemHeight < menuTop) {
         menuDivElement.scrollTop -= itemHeight;
--- a/packages/components/src/Autocomplete/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu.tsx
@@ -51,10 +51,8 @@ export function Menu({
       className={optionMenuClass}
       style={{ scrollBehavior: "smooth" }}
       ref={ref => {
-        if (!menuDiv.current) {
-          const stuff = ref as HTMLDivElement;
-          menuDiv.current = stuff;
-        }
+        const newMenuDiv = ref as HTMLDivElement;
+        menuDiv.current = newMenuDiv;
         return menuDiv.current;
       }}
     >
@@ -79,10 +77,8 @@ export function Menu({
             <div
               className={styles.icon}
               ref={ref => {
-                if (!menuItemRefs.current[index]) {
-                  const newMenuRef = ref as HTMLDivElement;
-                  menuItemRefs.current[index] = newMenuRef;
-                }
+                const newMenuRef = ref as HTMLDivElement;
+                menuItemRefs.current[index] = newMenuRef;
                 return menuItemRefs.current[index];
               }}
             >
--- a/packages/components/src/Autocomplete/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu.tsx
@@ -120,10 +120,10 @@ export function Menu({
       menuDivElement: HTMLDivElement,
       direction: "up" | "down",
     ) {
-      const itemDiv = menuDivElement.querySelectorAll(
+      const itemDiv = menuDivElement.querySelector(
         `button.${classnames(styles.option)}:nth-child(${highlightedIndex +
           1})`,
-      )[0] as HTMLButtonElement;
+      ) as HTMLButtonElement;
       if (!itemDiv) {
         return;
       }
--- a/packages/components/src/Autocomplete/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu.tsx
@@ -49,6 +49,7 @@ export function Menu({
   return (
     <div
       className={optionMenuClass}
+      style={{ scrollBehavior: "smooth" }}
       ref={ref => {
         if (!menuDiv.current) {
           const stuff = ref as HTMLDivElement;
--- a/packages/components/src/Autocomplete/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu.tsx
@@ -49,7 +49,6 @@ export function Menu({
   return (
     <div
       className={optionMenuClass}
-      style={{ scrollBehavior: "smooth" }}
       ref={element => {
         setRef(element, menuDiv);
       }}
--- a/packages/components/src/Autocomplete/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu.tsx
@@ -118,19 +118,20 @@ export function Menu({
 
     function scrollMenuIfItemNotInView(
       menuDivElement: HTMLDivElement,
-      direction: string,
+      direction: "up" | "down",
     ) {
-      const tempDiv = menuDivElement.getElementsByClassName(
-        classnames(styles.option),
-      )[highlightedIndex] as HTMLButtonElement;
-      if (!tempDiv) {
+      const itemDiv = menuDivElement.querySelectorAll(
+        `button.${classnames(styles.option)}:nth-child(${highlightedIndex +
+          1})`,
+      )[0] as HTMLButtonElement;
+      if (!itemDiv) {
         return;
       }
       const menuTop = menuDivElement.getBoundingClientRect().top;
-      const itemTop = tempDiv.getBoundingClientRect().top;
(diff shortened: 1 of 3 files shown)


packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -44,7 +47,17 @@ export function Menu({
   useEffect(() => setHighlightedIndex(initialHighlight), [options]);
 
   return (
-    <div className={optionMenuClass}>
+    <div
+      className={optionMenuClass}
+      style={{ scrollBehavior: "smooth" }}
```
willcodeforcoffee, line 52: Can you add this to the `optionMenuClass` instead of adding a style attribute?

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -44,7 +47,17 @@ export function Menu({
...
+      ref={ref => {
+        if (!menuDiv.current) {
+          const stuff = ref as HTMLDivElement;
+          menuDiv.current = stuff;
+        }
+        return menuDiv.current;
+      }}
```
willcodeforcoffee, line 59: I don't understand why you need to build the ref this way? Can'y you just set `ref={menuDiv}`?

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -63,7 +76,16 @@ export function Menu({
...
-            <div className={styles.icon}>
+            <div
+              className={styles.icon}
+              ref={ref => {
+                if (!menuItemRefs.current[index]) {
+                  const newMenuRef = ref as HTMLDivElement;
+                  menuItemRefs.current[index] = newMenuRef;
+                }
+                return menuItemRefs.current[index];
+              }}
```
willcodeforcoffee, line 87: This looks really similar to the above code, too. Maybe it should be refactored into a method?

willcodeforcoffee (changes requested):

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: @willcodeforcoffee It works, but typescript doesn't seem to like that

Screen Shot 2020-09-14 at 1 46 07 PM

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Good idea, refactored into it's own method!

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Yup, added the style into the css file!

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -29,6 +30,8 @@ export function Menu({
   const optionMenuClass = classnames(styles.options, {
     [styles.visible]: visible,
   });
+  const menuItemRefs = useRef<HTMLDivElement[]>([]);
```
seanhealy, line 33: I don't think we need this ref. We can use `queryBySelector` getting the n-th child in the scrolling function to accomplish the same thing.

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -19,6 +19,7 @@ interface MenuProps {
   onOptionSelect(chosenOption: Option): void;
 }
 
+// eslint-disable-next-line max-statements
```
seanhealy, line 22: How many statements are we at? We should avoid doing this if we can. I think we have a few opportunities to simplify the logic in here.

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -87,6 +100,24 @@ export function Menu({
     </div>
   );
 
+  function setRef(
```
seanhealy, line 103: I'm not following why we need this method?

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -99,15 +130,53 @@ export function Menu({
...
       }
+      if (menuDiv.current && menuItemRefs.current[highlightedIndex]) {
+        scrollDownIfTooLow(
+          menuDiv.current,
+          menuItemRefs.current[highlightedIndex],
+        );
+      }
     });
 
+    function scrollDownIfTooLow(
```
seanhealy, line 141: I think we should be able to generalize this into a single method with `scrollUpIfTooHigh`.

seanhealy (changes requested):
Hey, got some quick comments in to get y'all unblocked. Hope they make sense. Sorry they are kind of right to the point. Happy to talk about any of them or look at them together as well.

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Totally, DRY'd up these methods into one!

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: We were set the refs in a DRY way, but now that we only need the ref for the menu, we don't need this anymore! Killed this method ⚔️

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Yup, totally! Simplified logic and removed max statement es-lint disable

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Good idea!! Now using query by classname instead of setting refs 😄

dmcrorieGIT:
@seanhealy Thanks for the feedback!! I've addressed all comments, could you take another look? 😄

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -130,50 +101,41 @@ export function Menu({
...
+      if (menuDiv.current) {
+        scrollMenuIfItemNotInView(menuDiv.current, "up");
       }
     });
 
-    function scrollUpIfTooHigh(
+    function scrollMenuIfItemNotInView(
       menuDivElement: HTMLDivElement,
-      itemDivElement: HTMLDivElement,
+      direction: string,
```
seanhealy, line 121:
```suggestion
      direction: "up" | "down",
```

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -130,50 +101,41 @@ export function Menu({
...
     ) {
-      const tempDiv = itemDivElement.parentElement as HTMLDivElement;
+      const tempDiv = menuDivElement.getElementsByClassName(
```
seanhealy, line 123: I _think_ you may be better doing something like:

```ts
const option = menuDivElement.querySelector(`${CLASSNAME}:nth-child(${highlightedIndex})`)
```

This will push all the querying into the browser so it will be more efficient about it and we don't need an array here.

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -50,7 +48,7 @@ export function Menu({
     <div
       className={optionMenuClass}
       ref={element => {
-        setRef(element, menuDiv);
+        menuDiv.current = element as HTMLDivElement;
```
seanhealy, line 51: Are you not able to just do?

```
ref={menuDiv}
```

seanhealy (changes requested):
Looking good. Few other thoughts for you here. 😄

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: It doesn't look like I can. When I try, it gives me an error since I'm using a `useRef`, which initializes with `undefined`. If I use `createRef` instead, this works fine here, but since we're using keydown listeners the state of the ref is always `undefined`, and so the desired behaviour is lost 🤔

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Yup, it makes sense to constrict the type on this parameter! Done 😎

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Yup, that's a better way of query selecting; I'll do that!

dmcrorieGIT:
@seanhealy Thanks for the second round of feedback!! I addressed your comments. Could we do another round?? 😅 

another-thor

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Just to clarify, it technically "works" using `ref={menuDiv}`, but the typescript linter blows up about it 😅 

Screen Shot 2020-09-15 at 1 43 01 PM

seanhealy: I _think_ if you refine your ref as follows it should clear this up.

```ts
const menuDiv = useRef() as MutableRefObject<HTMLDivElement>;
```

I suspect what is happening is typescript is unhappy that you aren't handing an undefined case that in practical terms won't come up.

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -99,15 +101,45 @@ export function Menu({
...
+        scrollMenuIfItemNotInView(menuDiv.current, "up");
+      }
     });
 
+    function scrollMenuIfItemNotInView(
+      menuDivElement: HTMLDivElement,
+      direction: "up" | "down",
+    ) {
+      const itemDiv = menuDivElement.querySelector(
+        `button.${classnames(styles.option)}:nth-child(${highlightedIndex +
```
seanhealy, line 124: Should just be able to do:

```suggestion
        `button.${styles.option}:nth-child(${highlightedIndex +
```

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -99,15 +101,45 @@ export function Menu({
...
+          1})`,
+      ) as HTMLButtonElement;
+      if (!itemDiv) {
+        return;
+      }
```
seanhealy, line 129: Just personal preference for a guard statement.

```suggestion
      if (!itemDiv) return;
```

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -99,15 +101,45 @@ export function Menu({
...
+      const menuTop = menuDivElement.getBoundingClientRect().top;
+      const itemTop = itemDiv.getBoundingClientRect().top;
+      const itemHeight = itemDiv.getBoundingClientRect().height;
```
seanhealy, line 127:
```suggestion
      const {top: itemTop, height: itemHeight} = itemDiv.getBoundingClientRect();
```

seanhealy: I'm so so on this suggestion. It's _slightly_ better as it's not re-calling `itemDiv.getBoundingClientRect` for each variable but it may not be as readable

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Done!

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Oh nice! took out `classnames`

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: I like this better than getting the bounding rect for each element, so I'm going to use this change 😎

packages/components/src/Autocomplete/Menu.tsx
dmcrorieGIT: Oh yeah, that worked!! 🤩

dmcrorieGIT:
@seanhealy More changes addressed!

seanhealy (approved):

PR #356, opened 2020-09-14
 fix(components): fix a border issue on card accent

bdfranck:
## Motivations

The current card component has a minor visual issue with the accent. Because the accent and the standard card border are using the same border it creates a visual artifact where the accent angles into the standard border.

I tried exploring the below ideas:
- [x] See if there is a [CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/border) for managing border corners. (There isn't aside from border radius.)
- [x] Try using the CSS `outline` property for the accent. (The outline property can't be used for a single edge)
- [x] Try using the CSS `box-shadow` property. (I can't figure out how to use it for just one edge)
- [x] Try using a `::before` pseudo element. (Works!)

### Fixed

The accent border angles into the side border (3x zoom image).

image

I'm using a `::before` pseudo selector to put the border on top of the card and have a clean straight line.

image

### DIFF
--- a/packages/components/src/Card/Card.css
+++ b/packages/components/src/Card/Card.css
@@ -33,7 +33,15 @@
 }
 
 .accent {
+  border-top-width: 0;
+}
+
+.accent::before {
+  content: " ";
+  display: block;
+  margin: 0 calc(-1 * var(--border-base));
   border-top: 0.5rem solid var(--card--accent-color);
+  border-radius: var(--radius-base) var(--radius-base) 0 0;
 }
 
 .clickable {


bdfranck:
I've tested this in the following:
- [ ] Edge (Latest Mac)
- [ ] Edge (Latest Windows)
- [ ] Chrome (Latest Mac)
- [ ] Firefox (Latest Mac)
- [ ] Safari (Latest Mac)

rebecca-li:
LGTM!

bdfranck:
Hmmm, looks like I needed to request a formal review. 🤔

bdfranck:
I'll just post the request in the Friends of Frends channel.

PR #357, opened 2020-09-16
feat(components): Form component

eddysims:
## Motivations

We will soon need to be able to have a form component available. For this form component we will also need to know the state of the form that we are using.

## Changes

This PR introduces 3 new things.

1. a `Form` component
2. a `useFormState` hook that can be used to get the state of a form.
3. a `submit` prop on `Button` that allows it to submit a form.

### Added

- `Form` component
- `useFormState` hook
- a `submit` prop on `Button` component to render as `type="submit"`

### Changed

- Some eslint and gatsby node rules to recognize the `@jobber/hooks` package

### Deprecated

- 

### Removed

- 

### Fixed

- Small styling change around `FormField` validation to keep it associated with the field when used in `Content`

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Button/Button.tsx
```diff
@@ -106,15 +109,17 @@ export function Button(props: ButtonProps) {
     [styles.loading]: loading,
   });
 
+  const buttonType: ButtonType =
+    useFormContext() == undefined ? "button" : "submit";
```
seanhealy, line 113: Are we sure that we want to make all buttons in forms submit?

packages/components/src/Form/Form.mdx
```diff
@@ -0,0 +1,97 @@
...
+import { Form } from "@jobber/components/Form";
+import { Content } from "@jobber/components/Content";
+import { InputText } from "@jobber/components/InputText";
+import { Button } from "@jobber/components/Button";
+
+# Form
+
+<ComponentStatus stage="pre" responsive="no" accessible="no" />
+
+Write a description about what Form component is trying to achieve.
```
seanhealy, line 19: Write a description about what Form component is trying to achieve. 😉

packages/components/src/Form/Form.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const submitHandler = jest.fn();
+  const { getByText, getByLabelText } = render(
+    <MockForm onSubmit={submitHandler} />,
+  );
+
+  const input = getByLabelText("test form");
+  fireEvent.change(input, { target: { value: "hello" } });
+  fireEvent.click(getByText("submit"));
+
+  waitFor(() => {
```
seanhealy, line 18: does `await waitFor` work here instead of the `done()`?

packages/components/src/FormField/FormField.tsx
```diff
@@ -233,7 +233,7 @@ export function FormField({
   );
 
   return (
-    <>
+    <div>
```
seanhealy, line 236: Does this need to be a div?

packages/hooks/src/useFormState/useFormState.mdx
```diff
@@ -0,0 +1,70 @@
...
+import { Divider } from "@jobber/components/Divider";
+
+# UseFormState
+
+`useFormState` is an extremly simple hook that wraps the
+[useState()](https://reactjs.org/docs/hooks-state.html) hook and suplies some
+default values.
+
+`useFormState` **must** only by used when using a [`<Form />`](/components/form)
+component. Otherwise, it is useless.
```
seanhealy, line 22: Perhaps?

```suggestion
`useFormState` should **only** by used when using a [`<Form />`](/components/form)
component.
```

packages/components/src/Form/Form.tsx
```diff
@@ -0,0 +1,36 @@
...
+  } = methods;
+
+  useEffect(() => onStateChange && onStateChange({ isDirty, isValid }), [
+    isDirty,
+    isValid,
+  ]);
+
+  return (
+    <FormProvider {...methods}>
+      <form onSubmit={handleSubmit(submitHandler)}>{children}</form>
```
seanhealy, line 29: Related to the topic of pressing return in an input in a form how does that behave in this case?

seanhealy (changes requested):
Looking really nice! Few questions / thoughts for you. Happy to chat if you'd like.

packages/components/src/Form/Form.test.tsx
eddysims: we do need the `done()` here, but it can be taken out of the `waitFor`. I'm not sure if it makes a difference or not?

packages/components/src/FormField/FormField.tsx
eddysims: This is done so that when using something that has `display: flex` (such as `Content`) the validation message and the field are treated as one element. Without this, the validation message will get equal vertical spacing as 2 fields.

eddysims:
@seanhealy back to you!

packages/components/src/Button/Button.test.tsx
```diff
@@ -125,6 +125,18 @@ it("renders a Link as a Button for routing", () => {
   expect(tree).toMatchSnapshot();
 });
 
+it("renders button type='button' if not in a form", () => {
```
seanhealy, line 128: Probably no longer need this test?

packages/components/src/Button/Button.tsx
```diff
@@ -66,11 +68,19 @@ interface CancelActionProps extends ButtonFoundationProps {
   readonly type?: "secondary" | "tertiary";
 }
 
+interface SubmitButtonProps
+  extends Omit<ButtonFoundationProps, "external" | "onClick"> {
```
seanhealy, line 72: Probably want to also omit `to`, `variation` and `type` I think?

seanhealy (changes requested):
Two little thoughts then I think we're there!

seanhealy (approved):

PR #359, opened 2020-09-22
feat(components): Add className param to LightBox

larymc:
## Motivations
 - Chat panel is picking up the click which closes the LightBox, which means one click will close both Light.  Adding `click_ignore` would escape the click.

### Added

 - Added new optional param `className` in LightBox.

## Testing

 - Tested in playground.  Checked class added in `className` showed up in debugger.
---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/LightBox/LightBox.tsx
+++ b/packages/components/src/LightBox/LightBox.tsx
@@ -30,6 +30,10 @@ interface LightBoxProps {
    */
   readonly imageIndex?: number;
   /**
+   * Insert class to the component.
+   */
+  readonly className?: string;
+  /**
    * This function must set open to false in order to close the lightbox. Note there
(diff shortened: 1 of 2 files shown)


larymc:
Putting up this PR according to the discussion in [this thread](https://getjobber.slack.com/archives/CHV3VNPDE/p1600468962005300).  Thanks!

eddysims:
hey Larry, unfortunately this is something that we don't want to allow in Atlantis. The reason being is that a change like this starts to expose "escape hatches" into Atlantis components where devs can potentially change the way a component works. 

To keep Atlantis and its use in Jobber as consistent as possible, we want to keep components as specific as possible.

Would be happy to chat about the reason that you need this, and maybe try to find a better solution.

larymc:
Found alternate solution.  Closing this PR.

PR #373, opened 2020-10-02
fix(components): Remove input name if no validation is set

eddysims:
## Changes

Removes the generated `name` from inputs if no `name` and no `validations` are set.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- Removes the generated `name` from inputs if no `name` and no `validations` are set.

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/FormField/FormField.tsx
```diff
@@ -200,7 +200,7 @@ export function FormField({
     shouldShowMiniLabel(defaultValue, value),
   );
   const [identifier] = useState(uuid.v1());
-  name = name || `generatedName--${identifier}`;
+  name = !name && validations ? `generatedName--${identifier}` : name;
```
seanhealy, line 203: This ternary works but, it's a bit to parse how would you feel about an if statement instead? As `name` is already `name` if it's set.

seanhealy (changes requested):

packages/components/src/FormField/FormField.tsx
eddysims: 👍

seanhealy (approved):

PR #377, opened 2020-10-14
docs(components): Grid component

bdfranck:
## Motivations

This PR adds a `<Grid>` proposal.

### DIFF
--- a/docs/proposals/Grid/README.md
+++ b/docs/proposals/Grid/README.md
@@ -17,7 +17,7 @@ I can't think of any accessibility considerations at this time.
 ## Responsiveness & Mobile
 
 There's a future opportunity to change the number of rows for different
-breakpoints.
+breakpoints. I don't see a need to add it for the first iteration.
 
 ## Interface
 
@@ -29,8 +29,6 @@ breakpoints.
 ```
 
 ## Props Table
-
-_{Provide a table in the following format of the component's public API}_
 
 | name     | type             | default       | description                      |
 | -------- | ---------------- | ------------- | -------------------------------- |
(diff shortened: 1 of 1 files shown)


chris-at-jobber:
@bdfranck This looks like a great first crack! For the purposes of alignment with @ChristyCampos 's work on a future layout vision, I'd suggest we rename this to `Layout` as your initial component will probably serve as the first building block of a larger system - @GetJobber/mcweb might have more thoughts on technical implementation. One thing we'll probably need to consider for IE11 support is a graceful degradation/progressive enhancement approach, possibly using flexbox?

eddysims:
@chris-at-jobber @bdfranck I like the idea of this. Tagging @seanhealy and @JeffMarvin for discussion as well.

Are we open to using CSS grid and grid template areas for the implementation? I have an idea for this that I **think** would make the usage nice, the problem would be IE support.

My very rough thought of usage would be something like this:

```ts
const template = `
head head head
sidebar body body
footer footer footer
` 

<Grid template={template} spacing={"small" | "base" | "large"}>
    <GridItem area={head}>
        ...head content
    </GridItem>
    <GridItem area={sidebar}>
        ...sidebar content
    </GridItem>
    <GridItem area={body}>
        ...body content
    </GridItem>
    <GridItem area={footer}>
        ...footer content
    </GridItem>
</Grid>
```

With the option to change the `template` prop to an array of breakpoints with new templates. It would be a change in thinking from flexbox, but would allow for some more flexible layouts.


https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-areas
https://css-tricks.com/snippets/css/complete-guide-grid/
https://caniuse.com/css-grid

chris-at-jobber:
@eddysims I think we need to start looking at `Grid` implementations that can fall back to a simpler flex solution - whether that's a single column, or just some less sophisticated reflow, this feels to me like a great use case for "start with functional, enhance along with the user's system capabilities"

chris-at-jobber:
@bdfranck Given the state of the drawer work I think we can close this one off and re-open if it's needed again

Issue #384, opened 2020-10-16
Adjust padding on buttons with icons relative to size


seanhealy:
Related Components:
- https://atlantis.getjobber.com/components/button

Currently our buttons have different padding based on what size prop is being used. (Small 8px, Base 12px, Large 16px).

However, one an icon prop is introduced, the padding defaults to 16px regardless of size. The padding should stay the same according to its size.

Small button with 8px padding
![image](https://user-images.githubusercontent.com/131327/96310145-fc3fa180-0fc3-11eb-8e7a-f7553856fde2.png)

Padding changes to 16px when icon is introduced
![image](https://user-images.githubusercontent.com/131327/96310161-0366af80-0fc4-11eb-9930-6f2dbe9b16c1.png)

Base button with 12px padding
![image](https://user-images.githubusercontent.com/131327/96310173-095c9080-0fc4-11eb-8664-20f8d3d44dd1.png)

Padding changes to 16px when an icon is introduced
**INCLUDE IMAGE**
![image](https://user-images.githubusercontent.com/131327/96310182-0eb9db00-0fc4-11eb-8574-4d3dba15412a.png)

0length:
@seanhealy I'd like to work on this issue.

chris-at-jobber:
Issue resolved, thanks @0length !

PR #388, opened 2020-10-17
fix(components): update button padding on has icon and label button

0length:
## Motivations

I contributed by searching for issues related to hacktoberfest, and I found this. This afternoon I started analyzing the code and seeing what was causing this issue as well as working out a solution. 

With this pull request, it will always be the right solution for `this` issue ( #384 ).

## Changes
Delete Button padding property on `.hasIconAndLabel` class name selector so the padding property active is from size based selector.

**Thankyou**

### DIFF
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -155,7 +155,7 @@
 }
 
 .onlyIcon.large {
-  width: calc(var(--space-large) * 2.25);
+  width: calc(var(--space-large) * 2);
 }
 
 .hasIconAndLabel {
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -146,6 +146,18 @@
   padding: 0;
 }
 
+.onlyIcon.small {
+  width: calc(var(--space-small) * 3);
+}
+
+.onlyIcon.base {
+  width: calc(var(--space-base) * 2.25);
+}
+
+.onlyIcon.base {
+  width: calc(var(--space-large) * 2.25);
+}
+
 .hasIconAndLabel {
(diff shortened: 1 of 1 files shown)


eddysims:
@chris-at-jobber can you look at this one? Some visual changes to button here. I have listed what I believe are all the variations of the button.

### Base Button

image

### Large Button

image

### Small Button

image

### Base Icon Button

image

### Large Icon Button

image

### Small Icon Button

image

### Base Button w/ Icon

image

### Large Button w/ Icon

image

### Small Button w/ Icon

image

chris-at-jobber:
I think my only concerns are with the small and large icon-only buttons - for our purposes I don't think we actively use either option anywhere, but I would expect that they maintain the 1:1 ratio of the `base` icon-only button. 

Basically:
image

packages/components/src/Button/Button.css
```diff
@@ -146,9 +146,20 @@
...
+.onlyIcon.small {
+  width: calc(var(--space-small) * 3);
+}
+
+.onlyIcon.base {
+  width: calc(var(--space-base) * 2.25);
+}
+
+.onlyIcon.large {
+  width: calc(var(--space-large) * 2.25);
```
chris-at-jobber, line 158: If the multiplier here is set to `2` instead of `2.25` then we get the 48x48 (1:1) ratio that we're looking for - otherwise this looks great!

0length: oh yeah.. done

chris-at-jobber:
@0length looks great, thanks! @eddysims I'm with approving this one

chris-at-jobber (approved):
👍

Issue #379, opened 2020-10-16
Autocomplete should not show empty section headers


seanhealy:
Related Components:
- https://atlantis.getjobber.com/components/autocomplete#section-headings

If there are no options in a section hide the section header.

Example:

![image](https://user-images.githubusercontent.com/131327/96309239-27c18c80-0fc2-11eb-9719-33fb95f4930a.png)

merelj:
I'd like to work on this one if you don't mind

PR #390, opened 2020-10-17
feat(components): Autocomplete now filters out headers with no options

merelj:
## Motivations

Solves issue https://github.com/GetJobber/atlantis/issues/379

## Changes

Added options filtering to updateSearch function (not sure if it should also run on the initial array -- should it?)

### Changed

Autocomplete.updateSearch

### Fixed

fixed typo in a comment

## Testing

type in "Be" in filter at /components/autocomplete#section-headings of docz

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Autocomplete/Autocomplete.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.tsx
@@ -52,7 +52,7 @@ interface AutocompleteProps {
 
   /**
    * Called as the user types in the input. The autocomplete will display what
-   * is retuned from this method to the user as available options.
+   * is returned from this method to the user as available options.
    * @param newInputText
    */
   getOptions(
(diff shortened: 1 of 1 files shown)


packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -135,8 +135,11 @@ export function Autocomplete({
   }
 
   async function updateSearch() {
-    const opts = await getOptions(inputText);
-    setOptions(mapToOptions(opts));
+    const opts: AnyOption[] = await getOptions(inputText);
+    const filteredOpts = opts.filter((o: AnyOption) =>
```
eddysims, line 139: A preference here would be to use a variable name that is more descriptive. `o`, while works, doesn't tell us what we are looking at. A name like `option` may be better in this case

eddysims (changes requested):

packages/components/src/Autocomplete/Autocomplete.tsx
merelj: Makes sense 👍 
Changed the naming a bit, hope it's better now.

eddysims (approved):

Issue #385, opened 2020-10-16
Modal should receive focus when opened


seanhealy:
Related Components:
- https://atlantis.getjobber.com/components/modal

When a modal is opened, you still have to tab through all the content of the page to get to it. Tab and Shift + Tab should not focus outside of the modal. When a dialog opens, focus moves to an element inside the dialog. 

https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=111%2C122%2C131%2C132%2C133%2C141%2C143%2C144%2C146%2C148%2C1410%2C1411%2C1412%2C1413%2C211%2C212%2C223%2C224%2C225%2C226%2C241%2C242%2C243#focus-order

voodoods:
I'd like to work on that.

chris-at-jobber:
Issue resolved, thanks @voodoods !

PR #391, opened 2020-10-18
fix(components): Make modal focus when opened

voodoods:
## Motivations

Improve component behaviour and contribute to hacktoberfest by providing a fix for https://github.com/GetJobber/atlantis/issues/385

### Fixed

- Fix focus on modal content on mount

### Security

- fixed React vulnerability by making conditional for `catchKeyboardEvent` hook part of the hook itself 

This follows the [Only Call Hooks on Top Level](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level) rule as stated by the React docs.

### Testing

- added unit test for checking focus on open modal
- manual test: open modal in playground (`yarn start` --> http://localhost:3333/components/modal) and press tab key

### Misc

- Please don't forget to add "hacktoberfest" and / or "hacktoberfest-accepted" labels to this pull request to enable [verification](https://hacktoberfest.digitalocean.com/faq). Thank you :)

### DIFF
--- a/packages/components/src/ConfirmationModal/tests/__snapshots__/snapshot-ConfirmationModal.test.tsx.snap
+++ b/packages/components/src/ConfirmationModal/tests/__snapshots__/snapshot-ConfirmationModal.test.tsx.snap
@@ -3,6 +3,7 @@
 exports[`renders a simple ConfirmationModal 1`] = `
 <div
   className="container"
+  tabIndex={0}
 >
   <div
     className="overlay"
--- a/packages/components/src/Modal/Modal.test.tsx
+++ b/packages/components/src/Modal/Modal.test.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { cleanup, fireEvent, render } from "@testing-library/react";
 import { Modal } from ".";
+import styles from "./Modal.css";
 
 afterEach(cleanup);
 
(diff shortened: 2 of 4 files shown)


eddysims:
@voodoods Looking good so far! A couple things:

1. `Escape` no longer closes the `Modal`
2. I believe that when the `Modal` is open we do not want the focus to be able to leave the modal. @seanhealy or @chris-at-jobber can either of you confirm this?

chris-at-jobber:
@eddysims For #2, that would be desirable and if we can solve for it in this PR that's fine, but I feel like it may fall outside of the scope of the initial improvement. One suggested solution I have seen is [the `inert` polyfill](https://github.com/GoogleChrome/inert-polyfill). We have a similar issue with [Lightbox](https://atlantis.getjobber.com/components/light-box) so I might suggest that we handle this separately.

@voodoods I've tested this with a screenreader as well and it has the desired effect of reading out the title, close button, content and actions. Other than Eddy's request to maintain the `Esc` behaviour this looks good to me!

voodoods:
@eddysims @chris-at-jobber Thanks for your feedback!

- [x] fixed `Escape` keyboard event
- [x] moved new hook to separate function call so to improve readability a little more

Hope I didn't miss anything else. Please have another look.

chris-at-jobber:
Behaviour looks good to me 👍 

@eddysims I'll leave review on the code to you

eddysims (approved):

PR #398, opened 2020-10-26
chore: Forces node version on circle

eddysims:
## Motivations

CircleCI was setting the node versions to `15`. This wasn't so good. This PR will force the node version to `10`

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ orbs:
 references:
   container_config: &container_config
     docker:
-      - image: circleci/node
+      - image: circleci/node:10
     working_directory: ~/atlantis
 
   restore_repo: &restore_repo


.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ orbs:
 references:
   container_config: &container_config
     docker:
-      - image: circleci/node
+      - image: circleci/node:10
```
JeffMarvin, line 12: I'm curious, is there any value in specifying the full semver version we want?
https://hub.docker.com/r/circleci/node/tags?page=1&ordering=-name

eddysims: I don't think so. It also looks like the other versions have not been maintained for the last 2 years. My assumption is that `10` is the best practice.

Atlantis works (afaik) in all node 10 versions.

JeffMarvin:
@eddysims I've confirmed that the correct docker image for Node 10 is being used to set up all of the CircleCI runs on this PR.  I left a note with a question, but it would be an iterative improvement on this, and given the blocking nature of this PR, I am approving and merging this in the mean time.

JeffMarvin (approved):

PR #401, opened 2020-10-27
feat(components): Add numberOfLines prop to Text component

Pavithrak:
## Motivations
This PR introduces the `numberOfLines` prop to the Text component, which mimics the [same property](https://reactnative.dev/docs/0.9/text#numberoflines) in react native Text compoent.

## Changes

* Made the Text component accept `numberOfLines` optional prop
* This is passed down to the Typography component, which conditionally applies the class and `webkit-line-clamp` style.

### Added

`numberOfLines` prop to Text component


packages/components/src/Typography/Typography.tsx
```diff
@@ -51,18 +53,26 @@ export function Typography({
...
   );
 
+  const truncateLines: CSSProperties = {};
+  if (shouldTruncateText) {
+    truncateLines.WebkitLineClamp = numberOfLines;
+  }
+
   return (
-    <Tag id={id} className={className}>
+    <Tag id={id} className={className} style={truncateLines}>
```
Pavithrak, line 75: Sorry, I am not sure how to conditionally add a prop. It is because of this that lot of snapshot tests are updated.

Pavithrak:
I would like to add some tests to the Typography component if the approach is okay.

packages/components/src/Text/Text.tsx
```diff
@@ -2,6 +2,7 @@ import React, { ReactNode } from "react";
 import { Typography, TypographyOptions } from "../Typography";
 
 interface TextProps {
+  numberOfLines?: number;
```
seanhealy, line 5: @eddysims / @chris-at-jobber What do you think of for a name for this prop. I'm sort of leaning towards `maxLines`. As it reflects that this will be the maximum number of lines to be shown before it truncates?

Pavithrak: Oh, I thought we wanted to make the api close to what React native has for Text component. https://reactnative.dev/docs/0.9/text#numberoflines

packages/components/src/Typography/css/Truncate.css
```diff
@@ -0,0 +1,6 @@
+.textTruncate {
+  display: -webkit-box !important;
```
eddysims, line 2: does this `!important` need to be here? If so can we maybe throw in a comment as to why?

packages/components/src/Typography/Typography.tsx
```diff
@@ -51,18 +53,27 @@ export function Typography({
...
     textColor && textColors[textColor],
     emphasisType && emphasis[emphasisType],
+    shouldTruncateText && truncate.textTruncate,
   );
 
+  let truncateLines: CSSProperties | undefined;
+  if (shouldTruncateText) {
+    truncateLines = {};
+    truncateLines.WebkitLineClamp = numberOfLines;
+  }
```
eddysims, line 73: I wonder if this makes more sense to look something like this

```
const truncateLines: CSSProperties | undefined = {};
  if (shouldTruncateText) {
    truncateLines.WebkitLineClamp = numberOfLines;
  }
```

Since we are sending the var to the style attribute no matter what.

packages/components/src/Text/Text.tsx
```diff
@@ -28,5 +33,10 @@ export function Text({ variation = "default", children }: TextProps) {
     info: { textColor: "lightBlue", size: "base" },
   };
 
-  return <Typography {...variationMap[variation]}>{children}</Typography>;
+  const selectedVariation = variationMap[variation];
+  if (numberOfLines) {
+    selectedVariation.numberOfLines = numberOfLines;
+  }
```
eddysims, line 39: another small clean up. Is it more readable to do something like

```
const selectedVariation = {
    ...variationMap[variation],
    numberOfLines: numberOfLines || undefined,
  };
```

packages/components/src/Text/Text.mdx
```diff
@@ -69,3 +69,20 @@ only be used right after a page title.
...
+
+Used to truncate large text with `...` based on line wrapping, such that the
+total number of lines does not exceed this number.
+
+<Playground>
+  <Text numberOfLines="3">
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum nec pulvinar nunc. Suspendisse nec eros pretium, rutrum purus sit
+    amet, finibus ante. Donec pretium condimentum scelerisque. Duis eu ligula nec metus suscipit feugiat.
+
+    Etiam sapien sapien, mattis eu tincidunt quis, pretium sed metus. Maecenas quis dolor lacinia libero rhoncus fringilla. Cras mi ante,
```
eddysims, line 83: This makes it appear as though we can have paragraphs within the `Text` component, which I don't think is the case. Can we remove the empty line?

eddysims (changes requested):
A few comments here. Im also torn on the name of the prop. But I understand it being the same as React Nativ

eddysims:
@Pavithrak back to you with a few comments

packages/components/src/Text/Text.tsx
seanhealy: I think if it was specifying the number of lines regardless of content I'd agree. @hosseintalebi it's my understanding that we'll be wrapping a lot of the react native components as well to provide more opinionated / sensible defaults on them?

If that's the case I think I'd still prefer `maxLines` or something more indicative of the behaviour here. 😄

packages/components/src/Typography/Typography.tsx
Pavithrak: That changed almost every single snapshot file. So this way if we don't numberOfLines then style is not added to the Tag and the snapshots don't fail

packages/components/src/Text/Text.mdx
Pavithrak: 👍  I will remove the paragraph. But we are actually planning to use the Text component to show the notes preview text in activity feed. This preview text is supposed to preserve the paragraph and space formatting that was entered by SPs. Should the Text component not be used in that case?

packages/components/src/Typography/css/Truncate.css
Pavithrak: [This](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp) doc says that for webkit-line-clamp to work the display should be set to `webkit-box`. So thought of enforcing that display is always set to this value.

Pavithrak: If that makes sense, I can add that as a comment there.

packages/components/src/Text/Text.mdx
Pavithrak: @eddysims

packages/components/src/Text/Text.tsx
Pavithrak: So something like `maxLines = single | small | large ` same as [here](https://github.com/GetJobber/jobber-mobile/blob/192507714699035d7a6ff83d1e10f5524268e1e5/src/components/JobberText/JobberText.tsx#L34) ?

Pavithrak: @seanhealy

packages/components/src/Text/Text.mdx
eddysims: @Pavithrak I would think that if the text needs to preserve the paragraph structure that we would want to use `Content` and then wrap each paragraph in `Content`

```ts
<Content>
   <Text>
      ...
   </Text>
   <Text>
      ...
   </Text>
</Content>
```

Or alternatively the `Markdown` component will preserve paragraphs as well.

packages/components/src/Typography/css/Truncate.css
eddysims: sounds good!

eddysims:
@Pavithrak I think just the comment from Sean and tests and everything else looks pretty good to me.

ZakaryH:
the first 4-5 files changed in here seem unrelated. was that the IDE automatically reformatting things?

packages/components/src/Text/Text.tsx
ZakaryH: +1 for some presets for the lengths rather than an open ended `number` prop

packages/components/src/Typography/Typography.tsx
```diff
@@ -51,18 +53,27 @@ export function Typography({
...
+
   return (
-    <Tag id={id} className={className}>
+    <Tag id={id} className={className} style={truncateLines}>
```
ZakaryH, line 76: debating whether or not I think it would be better to have a handful of css classes defined

eg.

```
.truncate--small {}

.truncate--large {}
```

which lets us stick to only `className` and avoid adding an inline style

vs this approach that simply takes the number of lines we want to limit it to

packages/components/src/Text/Text.tsx
```diff
@@ -30,7 +32,11 @@ export function Text({
   };
 
   return (
-    <Typography textColor={textColors[variation] as TextColor} size={size}>
+    <Typography
+      textColor={textColors[variation] as TextColor}
+      size={size}
+      numberOfLines={numberOfLines || undefined}
```
ZakaryH, line 38: do we need to explicitly set this fallback here?

if we do need it, an alternative approach might be to set this as a default of `undefined` up on line 23

packages/components/src/Text/Text.tsx
eddysims: i think something like

```
maxLines: 'single' | 'small' | 'base' | 'large' | 'larger'
```

Or a subset of those would make sense.

ZakaryH:
### Removing the 'base' value

with my last commit I removed the concept of `base` from the presets and replaced it with "medium" to match our mobile API preset increments. I also used a different mechanism for handling when the `maxLines` prop is not needed

it's important that invocations of `Text` which don't care about `maxLines` never have to concern themselves with the prop. internally, whether we are doing that via a default value when one is not provided eg.

```
export function Text({
  variation = "default",
  size = "base",
  children,
  maxLines = "base",
...
```

or if we intercept that `undefined` case in some other way like I have in the last commit, I don't think it matters too much aside from perhaps a consistency angle if that's how we do it everywhere else

what I dislike about the `base` approach is that I'm not convinced a `base` concept makes sense for `maxLines` since it's an optional/situational prop. most invocations of this Text component don't have a `base`, at least not in the same way a `size` prop has a base value.

also I'm not a fan of supporting multiple ways to get the same result. if we have that default value of `base` and then map that to `undefined` we could now either omit the prop or say `maxLines = "base"` and we'd get the same thing. no truncation. I don't see any advantage to supporting multiple ways and feel it only leaves room for inconsistency.


------


### Failing Autocomplete test

so this was passing on CI but failing for me locally when I attempted to commit. I tried updating the snapshot but now it fails on CI and passes locally :|

it's failing because the snapshot is looking for a uuid generated value that is 1 off (44 vs 43). we're mocking that `uuid`

```
let count = 0;
const finalSection = 426655440000;

function fakeUUID() {
  count += 1;
  return `123e4567-e89b-12d3-a456-${finalSection + count}`;
}
```

and it seems to be keeping track of count, so presumably the CI and my local environment are calling the `uuid` a different amount of times during the tests. I haven't been able to pin down where/why

@eddysims if you have a moment I wouldn't mind a hand with this, as well as a final review since this test failure is (probably) not related to the actual changes functionality of this PR so all that should be good to be looked at. thanks!

packages/components/src/Text/Text.tsx
```diff
@@ -2,6 +2,8 @@ import React, { PropsWithChildren } from "react";
 import { Typography, TypographyOptions } from "../Typography";
 
 interface TextProps {
+  maxLines?: "single" | "small" | "medium" | "large" | "larger";
```
eddysims, line 5: I read your comments about changing `base` to `medium` just waiting for some other feedback on this as `base` is a pretty common pattern. Will get back to you on this asap.

eddysims (changes requested):
@ZakaryH just 2 small comments and one that I am just waiting for some feedback on.

packages/components/src/Text/Text.tsx
seanhealy: What is the default for this prop if it is unset?

eddysims: @ZakaryH we have discussed this more internally and here are our thoughts:

1. We should try not to have an optional prop without a default. In this case, `maxLines` should be optional with a default of something like `unlimited` which will have no line clamp.

2. For the time being, let's go with a consistent `base` for the name of the option. We are going to discuss if `base` makes sense for this and other components going forward, as it may be a little bit confusing.

ZakaryH:
@eddysims pushed up those changes! making a corresponding mobile PR to update `medium` -> `base` there too

eddysims (approved):

PR #406, opened 2020-10-28
docs(components): Proposal for Gallery component

Pavithrak:
## Motivations
This is a proposal to create a new component that would show list of Thumbnail images that gets displayed in a LightBox on click.


docs/proposals/ThumbnailGallery/README.md
```diff
@@ -0,0 +1,61 @@
+# Component Name
+
+The `ThumbnailGallery` component will accept a list of attachments for images
+and display them as thumbnails. This would also have an option to show the
+images in a LightBox component on click of those images.
+
+## Accessibility
+
+Assistive technology should be able to see the images and know the image name
+based on the `name` property of each attachment.
```
chris-at-jobber, line 10: additional considerations:

- How will users navigate the gallery with a keyboard
- How will "count of images left" blocks announce themselves to screen-readers

@chriscarline any other thoughts? [Checklist](https://jobber.atlassian.net/wiki/spaces/DES/pages/631604349/Accessibility+Checklist) for reference

chriscarline: Navigation between the thumbnails can be with `Tab` and `Shift+tab`(for backwards navigation). Order would be left to right.

I'm not as familiar with the standards for naming elements for screen readers but something like `[# of extra attachments] more attachments, button` if the reader is able to know the number of attachments. If not, maybe something like `More attachments, button` would work.

docs/proposals/ThumbnailGallery/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+```
+<ThumbnailGallery
+    attachments={[{
+            name: "filename.png",
+            url: "https://url-to-that-image"
+            type: "image/png"
+            size: 12345
+`   }]}
+    openGalleryOnClick=true|false
```
eddysims, line 31: I don't mind the idea of this. What do you think about a different approach, just to have another idea.

Instead of having the lightbox as part of this component, we would instead have an `onClick` prop that would return the image that was clicked. This way the lightbox can still be used outside of the component, but the consumer of the component can also do other things.

As i typed that I also think that a combination of the 2 would also be nice, something like

```
function handleClick() {
  hasGallery && openGalleryLightbox() // opens the gallery
  onClick && onClick({...image}) // returns the image to the consumber.
}

File: docs/proposals/ThumbnailGallery/README.md [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+            }}/>);
+})}
+```
+
+## Props Table
+
+| name               | type    | default | description                                                                     |
+| ------------------ | ------- | ------- | ------------------------------------------------------------------------------- |
+| attachments        | Array   | -       | Accepts a list of details about the images                                      |
+| openGalleryOnClick | Boolean | true    | A flag to determine if images should be shown in a LightBox or not when clicked |
```
eddysims, line 54: what do you think of the name `hasGallery`. I feel `openGalleryOnClick` is a little bit to descriptive.

docs/proposals/ThumbnailGallery/README.md
Pavithrak: I was under the assumption that showing the thumbnail view and opening images in LightBox is a pattern that we want to follow in more than few places in jobber-online. Without the LightBox, I feel it might be simpler to loop over attachments and use FormatFile in the places that we need it.

Pavithrak: I am not too familiar working with Atlantis. So does a proposal reflect how the complete component should look like or should it be around the first iteration of the component. For example, we were discussing if we should do the `show more` part of this component later and not for the first iteration. @chriscarline

Pavithrak: @eddysims I am not too familiar working with Atlantis. So does a proposal reflect how the complete component should look like or should it be around the first iteration of the component. For example, we were discussing if we should do the `show more` part of this component later and not for the first iteration. @chriscarline

docs/proposals/ThumbnailGallery/README.md
Pavithrak: yeah, that works. I will update the name.

docs/proposals/ThumbnailGallery/README.md
eddysims: The proposal is more for you to think through the component before you start working on the component. All of these comments are just considerations for you to think about before you start building the component, so that when the time comes we have thought of (hopefully) all the questions.

docs/proposals/ThumbnailGallery/README.md
```diff
@@ -0,0 +1,60 @@
...
+## Wireframe
+
+[mockup](mockup.png)
+
+## Interface
+
+```
+<ThumbnailGallery
+    attachments={[{
+            name: "filename.png",
```
seanhealy, line 26: We have a `FileUpload` interface in `packages/components/src/InputFile/InputFile.tsx` that's probably good to use for these.

docs/proposals/ThumbnailGallery/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Wireframe
+
+[mockup](mockup.png)
+
+## Interface
+
+```
+<ThumbnailGallery
+    attachments={[{
```
seanhealy, line 25: Perhaps `images` since these should always be images?

docs/proposals/ThumbnailGallery/README.md
```diff
@@ -0,0 +1,60 @@
...
+`   }]}
+    openGalleryOnClick=true|false
+/>
+```
+
+Internally ThumbnailGallery would look like this along with some style details
+
+```
+{attachemnts && attachments.map(attachment => {
+    return (<FormatFile file={{
```
seanhealy, line 39: I think we may want to use a `FormatImage` instead of a `FormatFile` here instead. `FormatImage` unfortunately doesn't yet exist. I know @JeffMarvin is looking at doing this for Hacktoberfest. I'm not sure where he's at with that. In lue of that though we could make a quick internal version specifically for the `Gallery`…

docs/proposals/ThumbnailGallery/README.md
```diff
@@ -0,0 +1,60 @@
+# Component Name
+
+The `ThumbnailGallery` component will accept a list of attachments for images
```
seanhealy, line 3: I think we can just call this `Gallery`.

docs/proposals/ThumbnailGallery/README.md
eddysims: The more I think about it, do we need this prop? Would there ever be a case where we wouldn't want this to happen? Lets just ❌  this prop

seanhealy (changes requested):
Few comments back for you. 😄

docs/proposals/ThumbnailGallery/README.md
Pavithrak: I was looking at it for the other PR, and it looks like `FileUpload` has the progress property which in this case would always be 1. So thought of having another interface specific for Gallery that does not require progress key but would be passing it as 1 when it is sending it to FormatFile.

docs/proposals/ThumbnailGallery/README.md
Pavithrak: Oh that's something I am not sure, assuming the case where we want to just show just an icon for pdf and video documents. But the light box would just show the images.

Pavithrak:
@seanhealy I think I have addressed some feedback and I have two questions regarding [FileUpload interface](https://github.com/GetJobber/atlantis/pull/406#discussion_r514882435) and the [types of attachments](https://github.com/GetJobber/atlantis/pull/406#discussion_r514883343) to be allowed.

docs/proposals/ThumbnailGallery/README.md
JeffMarvin: I'll address this in the review body - I think we can find a way to support both types.

docs/proposals/ThumbnailGallery/README.md
JeffMarvin: > Just confirmed with @chriscarline that, for the use case in activity feed, we do need to show the icons for pdfs and videos in thumbnail view. But LightBox would exclude non image attachments.

I think this might lead to confusing behaviour (the sequence of images in the lightbox wouldn't match the thumbnails, and that would be worse if not all of the thumbnails are shown).

@chriscarline & @chris-at-jobber - would some Figma work help here? It feels like the full requirements of the component aren't completely captured in the text... (see other discussion in this PR for examples.)

docs/proposals/ThumbnailGallery/README.md
JeffMarvin: We are currently exploring something that could combine both an uploaded fie/image or a URL if that is known. Part of the implementation will try to work with our current types to create something flexible and clear.

docs/proposals/ThumbnailGallery/README.md
JeffMarvin: @Pavithrak To the question of whether the `show more` part should be considered here, my answer is yes! 

The reason we want to do a proposal is to think about the component from a number of perspectives and use cases. And, to consider how this fits into what other components we have, or will have soon. A few examples that can really change or determine a component's behaviour could be: 
- whether this gallery will ever include files, or will just be images (this decision will have pro's and con's for both)
- that it should allow "truncation" of its contents (the "show more") -- i.e. you want a way to control how many items the gallery displays.... and what it does when they click on the "more" image.
- anything else we might possible see this kind of component being used for in the medium term. 

Not that these are hard, just that they take a bit of consideration and communication back and forth to make sure the component does what you need, but in a way that builds on what we already have. 

Ideally, these conversations start with designers when they are first approaching the design of the UI. And, it's usually with designers that these wide ranging considerations make sense (should it include other file types? does that limit what we want to do with image galleries? how will it work on mobile, and on phone web browsers, etc.).

We're still learning how to engage with teams on this stuff, so I will follow up for more feedback in person about this process! :)

JeffMarvin (changes requested):
@chriscarline & @chris-at-jobber - would some Figma work help here? It feels like the full requirements of the component aren't completely captured.

At its core, a Gallery is a collection of thumbnails that calls an externally provided callback function any time a member is clicked. In the first use case, the callback would launch the lightbox. (This next bit gets technical: A callback is the solution we've arrived at here from the PR discussion. And I think it is a good one. It means that a <Lightbox> or other functionality can be added outside of the gallery component and not need to be coupled with the component itself. The PR would need to change, and I'll add a suggestion of where that would be.)

But, we have two other use cases lurking in the discussion:
1. We need a small thumbnail gallery that includes images and other file types (according to the mockup attached https://github.com/GetJobber/atlantis/pull/406/files#diff-d2d77723ab99aba52b51c9c5cba5e4c38546ac0cd8aa2ffc1d9fc7bbb9f21f1d).
 - this also allows for truncation, somehow it can only show a subset of the items, and can also show more or react when clicked? (That behaviour is not specified anywhere I could see.)

2. We need a larger thumbnail gallery for cases where we manage a set of images for upload. (In the Figma for FileUpload) These are always just images (I think?).

Is there a way to have these both be fulfilled by a `<Gallery>`? I think so.

Even if we can't answer all of those questions precisely, I think they should be considered before we move forward with this proposal.

docs/proposals/Gallery/README.md
```diff
@@ -0,0 +1,58 @@
+# Component Name
+
+The `Gallery` component will accept a list of attachments for images and display
```
JeffMarvin, line 3: We mostly refer to images in this proposal, but it isn't clear if we also mean other files. We should be clear. (But, probably through some discussion first!)

docs/proposals/Gallery/README.md
```diff
@@ -0,0 +1,58 @@
...
+LightBox component on click of those images.
+
+## Accessibility
+
+Assistive technology should be able to see the images and know the image name
+based on the `name` property of each attachment.
+
+## Responsiveness & Mobile
+
+This component show maximum number of images possible in a line and show the
```
JeffMarvin, line 14: I have some questions about what is being described here. What does "maximum number possible" mean?
Do we want the component to always go full width in the container it appears in, and then truncate at one line of thumbnails? What sizes of items do we expect to put in the gallery? (I think some designer considerations might help here.)

docs/proposals/Gallery/README.md
```diff
@@ -0,0 +1,58 @@
...
+                progress: 1
+            }}/>);
+})}
+```
+
+## Props Table
+
+| name        | type  | default | description                                |
+| ----------- | ----- | ------- | ------------------------------------------ |
+| attachments | Array | -       | Accepts a list of details about the images |
```
JeffMarvin, line 52: We will want to provide a callback function to the component to be called whenever an item is clicked on. We may want more than one callback function if we want to include `delete` and `showMore` - but we should probably discuss in person what we want before determining the best structure for the callbacks.

docs/proposals/Gallery/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Future possible props
+
+For future iteration, we could make this component accept a `wrap` boolean prop
```
JeffMarvin, line 56: This is good to consider, we should include it in some of the current design discussions.

docs/proposals/ThumbnailGallery/README.md
Pavithrak: Discussed about this with @chriscarline. We feel the ideal approach is that we would want to see a pdf preview in LightBox as well but since that is not currently supported by LightBox (I am not really sure, does it?) we are okay to not display pdfs in LightBox and maybe onclick of a pdf will open them in a new tab just like how it behaves in notes section.

Pavithrak:
> @chriscarline & @chris-at-jobber - would some Figma work help here? It feels like the full requirements of the component aren't completely captured.
> 
> At its core, a Gallery is a collection of thumbnails that calls an externally provided callback function any time a member is clicked. In the first use case, the callback would launch the lightbox. (This next bit gets technical: A callback is the solution we've arrived at here from the PR discussion. And I think it is a good one. It means that a or other functionality can be added outside of the gallery component and not need to be coupled with the component itself. The PR would need to change, and I'll add a suggestion of where that would be.)
> 
> But, we have two other use cases lurking in the discussion:
> 
> 1. We need a small thumbnail gallery that includes images and other file types (according to the mockup attached https://github.com/GetJobber/atlantis/pull/406/files#diff-d2d77723ab99aba52b51c9c5cba5e4c38546ac0cd8aa2ffc1d9fc7bbb9f21f1d).
> 
> * this also allows for truncation, somehow it can only show a subset of the items, and can also show more or react when clicked? (That behaviour is not specified anywhere I could see.)
> 
> 1. We need a larger thumbnail gallery for cases where we manage a set of images for upload. (In the Figma for FileUpload) These are always just images (I think?).
> 
> Is there a way to have these both be fulfilled by a `<Gallery>`? I think so.
> 
> Even if we can't answer all of those questions precisely, I think they should be considered before we move forward with this proposal.

1. Yeah, we would like to see all the files attached to the notes. Just like how the notes section shows in quote/job/invoices. It would be great if LightBox can show a preview of all the files but it is still acceptable if we can show only images and open other files in new tab.

2. This should show as much thumbnails as the width of its container permits and display the `Show more` icon for the rest. Onclick of the `show more` should once again open the gallery view with all the images/files.

3. I am not sure if I can answer for image uploads since that's something we did not consider for our case.

JeffMarvin:
I'll follow up in person as well. 

My hope with my last comment on the review was that some designers could chat, maybe work out some rough requirements and maybe a Figma. Did @chris-at-jobber and @chriscarline have a chance to discuss and cover some of the questions I raised?

> This should show as much thumbnails as the width of its container permits and display the Show more icon for the rest. Onclick of the show more should once again open the gallery view with all the images/files.
- As mentioned, though, this is easy to say in theory but there may be some practical considerations. I was hoping that the designers could talk about this and maybe offer some Figma-based designs around how this component should behave in different useage contexts, breakpoints, etc.

> I am not sure if I can answer for image uploads since that's something we did not consider for our case.
But... that's why I want the designers to talk. The idea is that we try to make things in Atlantis that work for more than a single use case, so we can reuse them in different places. Since we know we need this for the file upload, that's what I was hoping the designers would be able to collaborate on.

ZakaryH:
A few last thoughts questions I'm not 100% on

- with the `+4` functionality, this document says it will happen on a single line with a `wrap` for multiple lines as a future prop whereas our figma doc and presumably desired outcome would be that behavior right now

- more implementation oriented: should we be passing in full size images and building a thumbnail of the size we want, or should we ask that appropriately sized thumbnails are the input to this? would there be a tolerance range on that, scaling down an image looks fine for a range of sizes. stretching can look not so good

- configurable size, or a static 56px? our design mockups assume a static size since our current needs all have the same thumbnail dimensions

- configurable callback on every item type? or would we want to be opinionated here and automatically open supported image types in lightbox. is it even practical to pass in a lightbox launching callback, I'd assume there's need for a component - so that might rule out that option for images anyways

chriscarline:
The +4 is the ideal but not needed for now.

For now static size as we don't yet have other use cases that warrant multiple sizes. Again, it's another thing to consider for the future.

docs/proposals/Gallery/README.md
```diff
@@ -0,0 +1,74 @@
...
+The `Gallery` component will accept an array of files and display them as
+thumbnails for images, other file types will instead display a generic icon
+corresponding to the file type. This would also have an option to show the
+images in a LightBox component on click of those images.
+
+## Accessibility
+
+We will support navigation through the items with tab and shift-tab. Since we
+cannot guarantee the `name` value passed in will be readable or useful, instead
+the alt value should reference the total number of files in the gallery and
```
JeffMarvin, line 12: `alt value` of what? I'm not certain what you are referring to.

docs/proposals/Gallery/README.md
```diff
@@ -0,0 +1,74 @@
...
+which file the user is currently focused on.
+
+eg. `alt=[1 of 4] attached files`
+
+Since the non-images will not be an actual thumbnail, rather some markup, we can
+use the `role=img` and `alt=...` attributes
+
+Triggering the onclick equivalent action should also be accessible via "space"
+and "return".
```
JeffMarvin, line 21: Not sure what is being referred to in this line. Do you want this when a certain part of the component is focussed?  (Which part?) What under focus would trigger the click? (Space and return are probably too specific. I think we want whatever the browser keyboard action for a click would be.)

JeffMarvin (changes requested):
Ok, some thoughts. Let's sync up to chat about these @ZakaryH

joelwarrington-jobber:
@chris-at-jobber - do you think this proposal is worth implementing? I don't mind taking it on and getting something kickstarted

chris-at-jobber:
@joelwarrington-jobber I think so - can we book some time next week to chat about this in more depth? I've got a lot of context that's probably missing at this point

joelwarrington-jobber:
Closing and continuing work in https://github.com/GetJobber/atlantis/pull/713

PR #407, opened 2020-10-29
feat(components): Thumbnail gallery implementation

Pavithrak:
## Motivations
We wanted to create a new component that would show list of thumbnail images and open them in a LightBox on click.

## Changes
* Added new component
* Introduced `showMetaData` prop to FormatFile

### Added
This new component will take list of attachments and them as thumbnails and open the lightbox when an image is clicked.

### Changed

This component uses the FormatFile internally which used to show the file name and size along with the thumbnail. For the current use case to work, I introduced `showMetaData` prop to FormatFile. This is true by default and hides the file name and size when false.

---


packages/components/src/ThumbnailGallery/ThumbnailGallery.css
```diff
@@ -0,0 +1,9 @@
+.thumbnail {
+  display: inline-block;
+  margin-right: 0.5rem;
```
ZakaryH, line 3: is there a spacing var that we can use here instead?

packages/components/src/ThumbnailGallery/ThumbnailGallery.css
```diff
@@ -0,0 +1,9 @@
...
+}
+
+.thumbnailContainer {
+  max-height: calc(var(--file-card--height) * 2.1);
```
ZakaryH, line 7: what is the `2.1` value from/for?

packages/components/src/ThumbnailGallery/ThumbnailGallery.css
```diff
@@ -0,0 +1,9 @@
+.thumbnail {
+  display: inline-block;
```
ZakaryH, line 2: I haven't played around with this yet, but I wonder if a `display: flex` on the parent might suit our needs better

I'll update once I've actually tried it out

packages/components/src/ThumbnailGallery/ThumbnailGallery.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [open, setOpen] = useState(false);
+  const [imageIndex, setImageIndex] = useState(0);
+
+  return (
+    <div className={styles.thumbnailContainer}>
+      {attachments &&
+        attachments.map((attachment, index) => {
+          return (
+            <div
+              key={`attachment-${index}`}
```
ZakaryH, line 50: I know there's warnings from React about using the `index` as the key value. I also know there's cases where it's acceptable to do. I can't say for sure which is applicable here, but something to keep in mind

packages/components/src/ThumbnailGallery/ThumbnailGallery.tsx
```diff
@@ -0,0 +1,93 @@
...
+interface ThumbnailGalleryProps {
+  /**
+   * Takes list of details about the files to display as thumbnails.
+   */
+  readonly attachments: [Attachment];
+
+  /**
+   * Flag to show images in a LightBox on click
+   */
+  readonly openGalleryOnClick?: boolean;
```
ZakaryH, line 37: do you think there's a case for not wanting to open lightbox on click but instead do something else entirely?

packages/components/src/ThumbnailGallery/ThumbnailGallery.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * The size of the file in bytes.
+   */
+  readonly size: number;
+
+  /**
+   * The data url of the file.
+   */
+  url: string;
```
ZakaryH, line 25: are we modifying this, or could it also be a `readonly`?

packages/components/src/ThumbnailGallery/ThumbnailGallery.css
Pavithrak: @ZakaryH  This was just to make sure we only show two lines. But I think that's going away now?

packages/components/src/ThumbnailGallery/ThumbnailGallery.tsx
```diff
@@ -0,0 +1,93 @@
+import React, { useState } from "react";
+import styles from "./ThumbnailGallery.css";
+import { FormatFile } from "../FormatFile";
+import { LightBox } from "../LightBox";
+
+interface Attachment {
```
seanhealy, line 6: We have a `FileUpload` interface in `packages/components/src/InputFile/InputFile.tsx` that's probably good to use here.

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -18,9 +18,14 @@ interface FormatFileProps {
...
    */
   onDelete?(): void;
+  showMetaData?: boolean;
 }
 
-export function FormatFile({ file, onDelete }: FormatFileProps) {
+export function FormatFile({
+  file,
+  onDelete,
+  showMetaData = true,
```
seanhealy, line 27: What is `showMetaData` doing here?

seanhealy:
I think the proposal for this component is still being looked at? Just want to make sure we're not getting ahead of ourselves if there are changes that come from the proposal?

https://github.com/GetJobber/atlantis/pull/406

Pavithrak:
> I think the proposal for this component is still being looked at? Just want to make sure we're not getting ahead of ourselves if there are changes that come from the proposal?
> 
> #406

Oh yeah, sorry, didn't mean to send this for review, just something I wanted to try while the proposal is getting reviewed.

seanhealy:
Ah, sounds great! 😄

packages/components/src/ThumbnailGallery/ThumbnailGallery.tsx
Pavithrak: I was initially using that but it looks like FileUpload has the progress property which in this case would always be 1. So thought of having another interface specific for Gallery that does not require progress key but would be passing it as 1 when it is sending it to FormatFile.

packages/components/src/FormatFile/FormatFile.tsx
Pavithrak: When false, it would hide the name and size information that this components shows by default.

chris-at-jobber:
Closing - this was a proof of concept branch happening alongside [Gallery proposal](https://github.com/GetJobber/atlantis/pull/406)

PR #410, opened 2020-10-31
feat(components): add lastIndex argument to onRequestClose function prop of LightBox 

Pavithrak:
## Motivations

Passed the last index of the image the user viewed to the onRequestClose function. 
This could be useful if LightBox is used for some onboarding functionality and if we wanted to build some analytics around how much further the user is engaged in the content.

## Changes
Passed the `currentImageIndex` state value to the `onRequestClose` function when LightBox is closed.

### DIFF
--- a/packages/components/src/LightBox/LightBox.mdx
+++ b/packages/components/src/LightBox/LightBox.mdx
@@ -70,3 +70,42 @@ suited by using a dialogue or alternative method.
 ## Props
 
 <Props of={LightBox} />
+
+### onRequestClose
+
+This function receives the index of the last image the user viewed before
+closing the LightBox component. This might we useful if we need to build some
+analytics on top of this component.
+
+<Playground>
+  {() => {
+    const [open, isOpen] = useState(false);
+    return (
+      <>
+        <Button label="😈 Click me!" onClick={() => isOpen(true)} />
+        <LightBox
+          open={open}
+          images={[
+            {
+              title: "😃 Happy Face",
+              url: "https://i.imgur.com/6Jcfgnp.jpg",
+              caption: "This is a happy face"
+            },
+            {
+              title: "😮",
+              url: "https://i.imgur.com/OsLxYkT.jpg",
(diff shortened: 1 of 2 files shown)


packages/components/src/LightBox/LightBox.mdx
```diff
@@ -70,3 +70,45 @@ suited by using a dialogue or alternative method.
 ## Props
 
 <Props of={LightBox} />
+
+### onRequestClose
+
+This function receives the index of the last image the user viewed before
+closing the LightBox component with the key lastPosition. This might we useful
+if we need to build some analytics on top of this component.
```
JeffMarvin, line 78:
```suggestion
if we want to return information about the final state when closed (e.g. for analytics).
```

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -33,8 +37,9 @@ interface LightBoxProps {
    * This function must set open to false in order to close the lightbox. Note there
    * is a 300ms easing animation on lightbox close that occurs before this function
    * is called.
+   * This function receives the last image position viewed in the LightBox as an argument.
```
JeffMarvin, line 40: This comment should be updated to reflect the new options object.

packages/components/src/LightBox/LightBox.mdx
```diff
@@ -70,3 +70,46 @@ suited by using a dialogue or alternative method.
...
+    const [open, isOpen] = useState(false);
+    return (
+      <>
+        <Button label="😈 Click me!" onClick={() => isOpen(true)} />
+        <LightBox
+          open={open}
+          images={[
+            {
+              title: "😃 Happy Face",
+              url: "https://i.imgur.com/6Jcfgnp.jpg",
```
JeffMarvin, line 92: Are these intended to be working URLs? I am not seeing them load when I run this.

JeffMarvin (approved):
Looks good, thanks!

PR #412, opened 2020-11-05
fix(components): Only focus the modal once when it opens

eddysims:
## Motivations

There was a fix to `Modal` that focused the modal when it was opened. This however caused a bug if there was an `InputText` in the `Modal`. The `Modal` would re-render on input change and re-focus the `Modal` container instead of keeping the focus in the `InputText`

## Changes

Makes the modal only focus on initial open.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- `Modal` re-focusing on itself on re-renders.

### Security

- 

## Testing

---

Random photo of Atlantis


chris-at-jobber:
image

Unrelated issue, but it looks like `InputText` is inheriting font-family from the user-agent stylesheet?

seanhealy (approved):

PR #415, opened 2020-11-09
fix(components): Sets AutoComplete to blank if no matches

eddysims:
## Motivations

When the autocomplete is blurred with `!allowFreeForm` and does not match a result, the `onChange` was correctly being called with undefined, however the input value was not reflecting this change.

This left a confusing user experience as the data returned and the data showed was different.

## Changes

Clears the `input` on `Autocomplete` if `!allowFreeForm` and input does not match a result on blur.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Clears the `input` on `Autocomplete` if `!allowFreeForm` and input does not match a result on blur to that the data and visuals match

### Security

- 

## Testing

- Navigation to `Autocomplete` page on preview.
- add `allowFreeForm={false}` to the first `Playground` example.
- type `Nos` into the input and see that there is one result.
- tab out, value in the input should be cleared

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Autocomplete/Autocomplete.test.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.test.tsx
@@ -162,3 +162,112 @@ test("it should call the handler skipping headings when an option is selected", 
 
   expect(changeHandler).toHaveBeenCalledWith(headingOptions[1].options[0]);
 });
+
+it("should remove the menu when blurred", async () => {
+  const changeHandler = jest.fn();
+  const { getByRole, getByText, queryByText } = render(
+    <Autocomplete
+      value={undefined}
+      onChange={changeHandler}
+      initialOptions={options}
+      getOptions={returnOptions(options)}
+      placeholder="placeholder_name"
+    />,
+  );
+
+  const input = getByRole("textbox");
+
+  input.focus();
+
+  await waitFor(() => {
+    expect(getByText("option_0")).toBeInstanceOf(HTMLParagraphElement);
+  });
+
+  input.blur();
+
+  await waitFor(() => {
+    expect(queryByText("option_0")).toBeNull();
+  });
+});
+
+it("should call onBlur callback when blurred", async () => {
+  const blurHandler = jest.fn();
+  const { getByRole } = render(
+    <Autocomplete
+      value={undefined}
+      onChange={jest.fn()}
+      initialOptions={options}
+      getOptions={returnOptions(options)}
+      placeholder="placeholder_name"
+      onBlur={blurHandler}
+    />,
+  );
+
+  const input = getByRole("textbox");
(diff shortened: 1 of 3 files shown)


JeffMarvin:
👀

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -118,12 +118,14 @@ export function Autocomplete({
...
       />
-      <Menu
-        visible={menuVisible}
-        options={options}
-        selectedOption={value}
-        onOptionSelect={handleMenuChange}
-      />
+      {menuVisible && (
+        <Menu
+          visible={true}
```
JeffMarvin, line 123: I don't disagree with this change. Are we doing this to make sure Menu isn't getting rendered when it is invisible?

packages/components/src/Autocomplete/Autocomplete.test.tsx
```diff
@@ -162,3 +162,112 @@ test("it should call the handler skipping headings when an option is selected",
...
+  });
+
+  input.blur();
+
+  await waitFor(() => {
+    expect(changeHandler).toHaveBeenCalledWith(undefined);
+  });
+});
+
+it("sets the input value to blank if allowFreeForm is false and not matched", async () => {
```
JeffMarvin, line 245: Just for my understanding... is it irrelevant to test "sets input to blank when allowFreeForm is true and not matched"?

packages/components/src/Autocomplete/Autocomplete.test.tsx
```diff
@@ -162,3 +162,112 @@ test("it should call the handler skipping headings when an option is selected",
 
   expect(changeHandler).toHaveBeenCalledWith(headingOptions[1].options[0]);
 });
+
```
JeffMarvin, line 165: Nice! Tests are so good for us to invest in. They are docs and also examples for other devs to look at when they write their own tests.

JeffMarvin (approved):
Solid stuff! I'm approving this because I don't think any of my questions would block this.

packages/components/src/Autocomplete/Autocomplete.test.tsx
eddysims: that use case wouldn't be true, with `allowFreeForm` to be true the input can equal a value that is not set in options.

My mindset here was to test everything that I touched, there is probably room for a test around `allowFreeform={true}` for some time in the future

packages/components/src/Autocomplete/Autocomplete.tsx
eddysims: yep, this basically takes the visibility from css into react

PR #419, opened 2020-11-17
feat(components): Add size prop to Text

eddysims:
## Motivations

The `Text` component needed an improvement that allowed it to have size and variations. This PR will add the `size` prop.

## Changes

Adds a `size` prop on `Text` component. 

### Added

- `size` prop to the `Text` component

### Changed

- 

### Deprecated

- 

### Removed

- `intro` from `variations` of `Text`

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Text/Text.mdx
+++ b/packages/components/src/Text/Text.mdx
@@ -29,7 +29,9 @@ import { Text } from "@jobber/components/Text";
 
 ---
 
-## Default
+## Variations
+
+### Default
 
 Use this as the body text
 
@@ -37,7 +39,7 @@ Use this as the body text
   <Text>Ask the information you need upfront from clients and new leads</Text>
 </Playground>
 
-## Subdued
+### Subdued
 
 De-emphasize a piece of text that is less important to the user
 
@@ -45,7 +47,7 @@ De-emphasize a piece of text that is less important to the user
   <Text variation="subdued">Job note linked to related invoice</Text>
 </Playground>
 
-## Feedback message
+### Feedback message
 
 <Playground>
   <Text variation="success">Invoice sent</Text>
@@ -61,7 +63,7 @@ De-emphasize a piece of text that is less important to the user
 
 ## Sizes
 
-## Intro
+### Intro
 
 `Text` can be used with different sizes.
 
--- a/packages/components/src/Text/Text.mdx
+++ b/packages/components/src/Text/Text.mdx
@@ -74,8 +74,9 @@ about the page title and can only be used right after a page title.
   </Text>
 </Playground>
 
-The `small` size is intended to be used as helper text on things like
-checkboxes.
+The `small` size is intended to be used as small, helper content on things like
+checkboxes, radio buttons, inputs or any other UI elements that need help
+provided.
 
 <Playground>
   <Text size="small">What is this, a font for 🐜s?</Text>
--- a/packages/components/src/Text/Text.mdx
+++ b/packages/components/src/Text/Text.mdx
@@ -45,17 +45,6 @@ De-emphasize a piece of text that is less important to the user
   <Text variation="subdued">Job note linked to related invoice</Text>
 </Playground>
 
-## Intro
-
-Used as an intro text that gives more information about the page title. This can
-only be used right after a page title.
-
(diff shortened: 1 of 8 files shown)


packages/components/src/Text/Text.mdx
```diff
@@ -69,3 +58,25 @@ only be used right after a page title.
...
+about the page title and can only be used right after a page title.
+
+<Playground>
+  <Text size="large">
+    Attract new customers by asking happy clients for a Facebook recommendation
+  </Text>
+</Playground>
+
+The `small` size is intended to be used as helper text on things like
+checkboxes.
```
JeffMarvin, line 78: suggest expanding this a bit, as I expect this to be a big styling component. 
`checkboxes.` -> `checkboxes and other form inputs, or smaller context help for other UI or content.`

packages/components/src/Text/Text.tsx
```diff
@@ -1,32 +1,37 @@
...
-}
 
-export interface VariationMap {
-  [variation: string]: TypographyOptions;
+  readonly size?: "small" | "base" | "large";
 }
 
-export function Text({ variation = "default", children }: TextProps) {
-  const variationMap: VariationMap = {
-    default: { textColor: "greyBlueDark", size: "base" },
```
JeffMarvin, line 22: Since `size` is no longer being defined for each variation, should we talk more in the Docz readme for this component about the interactions between size and variation? Or, provide some different examples that include some of these previous defaults?

JeffMarvin (commented):
In testing, found this:
When the prop exists but doesn't have a proper value, it renders the text in the large size:
image
image

packages/components/src/Text/Text.tsx
eddysims: The only previous example that changed was `intro`. I added in a section about size that I think covers this. Do you have ideas for other variations to cover?

packages/components/src/Text/Text.mdx
eddysims: This is where Typescript will help I think. The default size of `Typography` is not `small`. So any invalid prop there would render the default size of typography. 

In real world usage of the examples above (outside of Docz) typescript would catch this and complain about an invalid prop value.

image

image

JeffMarvin: Yes, you are totally right! Ok, this is fine. Its weird (can we fix it anyway?) but its not standard behavior.

packages/components/src/Text/Text.tsx
JeffMarvin: No, actually, I think that covers it.

packages/components/src/Text/Text.mdx
JeffMarvin: Talked a bit more, and it is not only "nonstandard" it is invalid TypeScript. So, no need to write code that replaces that!

chris-at-jobber:
Design decisions look good!

JeffMarvin (approved):

packages/components/src/Text/Text.mdx
```diff
@@ -69,3 +58,26 @@ only be used right after a page title.
     Drag to rearrange the order that fields show up in Jobber
   </Text>
 </Playground>
+
+## Sizes
+
+## Intro
```
chris-at-jobber, line 64: Let's bump these size subheaders down to an h3 - I think given the changes we want to go

### Large
(example)
### Base
(example)
### Small
(example)

chris-at-jobber:
Just some small formatting notes on the design guidelines 👍

JeffMarvin:
@chris-at-jobber I pass the "approval torch" over to you. Feel free to add the right approved label when this looks good.

eddysims:
@chris-at-jobber done

chris-at-jobber:
I moved to put `Base` first in the sizes guideline since that's the default and as such will likely have the broadest amount of usage.

eddysims:
@chris-at-jobber build works again.

chris-at-jobber:
I'm good with this if you are!

chris-at-jobber (approved):

PR #420, opened 2020-11-18
feat(components): Add label and description to RadioOption

eddysims:
## Motivations

RadioGroup needs to be able to handle a description. The current use case does not allow that to be done very nicely. This PR adds a `label` and `description` prop to `RadioOption` that will allow for a nicer way to add labels and descriptions

### Added

- `label` prop to `RadioOption`
- `description` prop to `RadioOption`
- Additional styling to `RadioOption`

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/RadioGroup/RadioGroup.css
+++ b/packages/components/src/RadioGroup/RadioGroup.css
@@ -12,6 +12,7 @@
 .label {
   display: inline-flex;
   align-items: center;
+  cursor: pointer;
 }
 
 .input + .label::before {
@@ -20,12 +21,11 @@
   width: var(--space-base);
   height: var(--space-base);
   box-sizing: border-box;
-  margin: 0 var(--space-small);
+  margin: 0 var(--space-small) 0 0;
   border: var(--border-thick) solid var(--color-grey--light);
   border-radius: 100%;
   background-color: var(--color-white);
-  cursor: pointer;
-  transition: all 0.2s;
+  transition: all var(--timing-base);
 }
 
 .input:focus + .label:before {
@@ -39,9 +39,15 @@
 
 .input[disabled] + .label {
   color: var(--color-blue--lighter);
+  cursor: not-allowed;
 }
 
 .input[disabled] + .label::before {
   border-color: var(--color-grey--light);
   background-color: var(--color-grey--lighter);
+}
+
+.description {
+  margin-bottom: var(--space-smallest);
+  padding-left: var(--space-large);
 }
--- a/packages/components/src/RadioGroup/RadioGroup.css
+++ b/packages/components/src/RadioGroup/RadioGroup.css
@@ -48,6 +48,6 @@
 }
 
 .description {
-  margin-bottom: var(--space-smallest);
+  margin-bottom: var(--space-small);
   padding-left: var(--space-large);
 }
--- a/packages/components/src/RadioGroup/RadioGroup.mdx
+++ b/packages/components/src/RadioGroup/RadioGroup.mdx
@@ -8,9 +8,10 @@ import { Playground, Props } from "docz";
 import { ComponentStatus } from "@jobber/docx";
 import { RadioGroup, RadioOption } from ".";
 import { Content } from "@jobber/components/Content";
+import { Text } from "@jobber/components/Text";
 import { Checkbox } from "@jobber/components/Checkbox";
 import { Divider } from "@jobber/components/Divider";
-import { Icon } from "@jobber/components/Icon";
+import { Avatar } from "@jobber/components/Avatar";
 import { useState } from "react";
 
 # Radio Group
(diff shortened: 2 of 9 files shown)


eddysims:
@chris-at-jobber Can you please look at this one? It changes some things about `RadioOption`

- Adds a description.
- Changes the cursor to be pointer when over a label (same functionality as `Checkbox`)
- Changes the cursor to be not-allowed when over a disabled label (same functionality as `Checkbox`)

chris-at-jobber:
Looks good visually, although I have two concerns:

1. there's no connection between the description and the radio option to a screen-reader (it only reads out the radio option label, the description is skipped entirely)
2. the new "radiooptionwithchildren" does not pass any information to a screen-reader beyond the n/n radio count (eg "radio button, 1 of 5")

chris-at-jobber:
> aria-describedby (property)§
> Identifies the element (or elements) that describes the object. See related aria-labelledby.
> 
> The aria-labelledby attribute is similar to the aria-describedby in that both reference other elements to calculate a text alternative, but a label should be concise, where a description is intended to provide more verbose information.
> 
> The element or elements referenced by the aria-describedby comprise the entire description. Include ID references to multiple elements if necessary, or enclose a set of elements (e.g., paragraphs) with the element referenced by the ID.

packages/components/src/RadioGroup/RadioGroup.css
```diff
@@ -39,9 +39,15 @@
...
+  cursor: not-allowed;
 }
 
 .input[disabled] + .label::before {
   border-color: var(--color-grey--light);
   background-color: var(--color-grey--lighter);
 }
+
+.description {
+  margin-bottom: var(--space-small);
```
chris-at-jobber, line 51: This looks good, can we do the same to `Checkbox` description?

image

image

eddysims:
> Looks good visually, although I have two concerns:
> 
> 1. there's no connection between the description and the radio option to a screen-reader (it only reads out the radio option label, the description is skipped entirely). I think we can use a `described-by` attribute here?
> 2. the new "radiooptionwithchildren" does not pass any information to a screen-reader beyond the n/n radio count (eg "radio button, 1 of 5")

@chris-at-jobber I have added in the `describedby` to the description.

I have also found that the `children` version does pass info in as long as there is text/aria info to read. In the example here, there is nothing to pass because of the icons.

If you were to do something like `<RadioOption>Apple</RadioOption>` the screen reader would pick that up.

chris-at-jobber:
Hmm, I don't love that it gives the user the ability to pass in `children` that don't read anything out, but I think that's a function of `Icon` being the example here. I'm hard-pressed to think of use case where `children` is used, though, so maybe this is a concern for the end implementer? 

It would be nice to require an aria-label or something as a required identifier, even if it's not visible, but with the free-form aspect of `children` that could result in some strange cases of multiple labels?

packages/components/src/RadioGroup/RadioGroup.test.tsx
```diff
@@ -53,28 +40,76 @@ test("it should be able to disable options", () => {
...
-        <RadioOption value="one">One</RadioOption>
-        <RadioOption value="two">Two</RadioOption>
+      <RadioGroup onChange={handleSecondValue} value={valueTwo}>
+        <RadioOption value="one" label="One" />
+        <RadioOption value="two" label="Two" />
       </RadioGroup>
     </>
   );
+
+  function handleFirstChange(val: string) {
```
JeffMarvin, line 106: This and the function below use two words `change` and `value` when I think they are referring to the same thing. Consider making the use the same word.

packages/components/src/RadioGroup/RadioOption.tsx
```diff
@@ -0,0 +1,85 @@
...
+   */
+  description?: string;
+}
+
+type RadioOptionProps = XOR<
+  RadioOptionsWithDescription,
+  PropsWithChildren<BaseRadioOptions>
+>;
+
+export function RadioOption({ children }: PropsWithChildren<RadioOptionProps>) {
```
JeffMarvin, line 25: This isn't required to fix... but this function component definition gets a bit lost among all the typescript. Thoughts on a way we could make this a bit easier to read by moving it or highlighting it with a comment?

packages/components/src/RadioGroup/RadioGroup.mdx
```diff
@@ -75,3 +76,55 @@ option will prevent a user from being able to select the option.
...
+    );
+  }}
+</Playground>
+
+## Render RadioOption with Children
+
+If the `RadioGroup` is being used to select something other then text, you can
+render the `RadioOptions` with `children`.
+
+**Note:** When using `children`, `label` and `description` are not allowed.
```
JeffMarvin, line 114: It might help to explain why they aren't allowed here. Also might be good to include a bit more explanation about how screen readers will use this label and description, and how each should be written (short and longer, respectively).

JeffMarvin (changes requested):
Code looks good, just had a few small comments/suggestions.

Have you done any smoke checks with this in Jobber yet?

packages/components/src/RadioGroup/RadioOption.tsx
eddysims: I have added a comment around this one.

eddysims:
@JeffMarvin @chris-at-jobber comments addressed and ready for re-review

packages/components/src/RadioGroup/RadioGroup.mdx
```diff
@@ -114,18 +115,33 @@ render the `RadioOptions` with `children`.
...
-          </RadioOption>
-        ))}
-      </RadioGroup>
+      <>
+        <Content>
+          <Text>Select a user</Text>
+          <RadioGroup onChange={setPhoto} value={company}>
+            {users.map(user => (
+              <RadioOption value={user.name}>
+                <Avatar initials={user.initials} name={user.name} />
```
JeffMarvin, line 139: This is a great idea (providing a better example)!

packages/components/src/RadioGroup/RadioOption.tsx
```diff
@@ -22,6 +22,10 @@ type RadioOptionProps = XOR<
   PropsWithChildren<BaseRadioOptions>
 >;
 
+/**
```
JeffMarvin, line 28: Nice, thanks!

JeffMarvin:
Looks good to me. 
@chris-at-jobber If you like what you see here, smash that APPROVE button.

packages/components/src/RadioGroup/RadioGroup.mdx
```diff
@@ -111,7 +111,8 @@ A `description` can be used to further describe the label of the `RadioOption`.
 If the `RadioGroup` is being used to select something other then text, you can
 render the `RadioOptions` with `children`.
 
-**Note:** When using `children`, `label` and `description` are not allowed.
+**Note:** When using `children`, `label` and `description` are not allowed due
+to the face that they would collide with the `children` prop.
```
chris-at-jobber, line 116: TYPO: fact** that they would collide

chris-at-jobber (approved):
Minor docs typo but I'm good approving with that fixed

JeffMarvin (approved):

PR #426, opened 2020-11-27
feat(components): Reduce width of drawer for medium breakpoint

bdfranck:
## Motivations

This PR improved the appearance of the drawer on smaller screens by introducting a medium breakoint. It also fixes some layout issues with the drawerGrid component.

## Changes

The drawer width is too large for screens between 640px and 1024px wide

image

Now the drawer is smaller for this screen size.

image

### DIFF
--- a/packages/components/src/Drawer/Drawer.css
+++ b/packages/components/src/Drawer/Drawer.css
@@ -18,6 +18,10 @@
   @media (--medium-screens), print {
     position: relative;
     z-index: auto;
+    width: 280px;
+  }
+
+  @media (--wide-screens) {
     width: 420px;
   }
 }
--- a/packages/components/src/Drawer/Drawer.css
+++ b/packages/components/src/Drawer/Drawer.css
@@ -1,4 +1,5 @@
 :root {
+  --drawer-width: calc(var(--space-base) * 17.5);
   --drawer--background-color: var(--color-white);
   --drawer--header-background-color: var(--color-grey--lightest);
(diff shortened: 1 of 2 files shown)


packages/components/src/Drawer/Drawer.css
```diff
@@ -18,6 +18,10 @@
   @media (--medium-screens), print {
     position: relative;
     z-index: auto;
+    width: 280px;
```
eddysims, line 21: Can we move these to css custom properties and use a calc based off of the base size?

Example:

```
:root {
   --drawer-width: calc(var(--space-base) * 17.5);
}
```

then in `@media (--wide-screens)` redeclare it as `--drawer-width: calc(var(--space-base) * 26.25);`

eddysims (changes requested):

eddysims:
@bdfranck one change requested. Also just double checking that a media query is the right way to go.

We also have https://atlantis.getjobber.com/hooks/use-resize-observer which allows for responsiveness to live on the component instead of the browser.

packages/components/src/Drawer/Drawer.css
bdfranck: ooooo can do!

bdfranck:
@eddysims Thanks for the review! I've addressed your changes. I think a media query works for now but we can consider the `useResizeObserver` for future versions.

eddysims (approved):

PR #427, opened 2020-11-30
fix(components): Change `Table` borders from box-shadow to pseudo element

eddysims:
## Motivations

The `Table` was using `box-shadow` as borders due to the fact that `<table>` can not have a `border-radius` applied to it. While this looks fine when a `Table` is by its self. It looks a little bonkers when paired with something like a `Card`.

## Changes

Change the `Table` borders from `box-shadow` to a `pseudo` element.

### Added

- 

### Changed

- Change the `Table` borders from `box-shadow` to a `pseudo` element.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Table/Table.css
+++ b/packages/components/src/Table/Table.css
@@ -2,11 +2,26 @@
   --border-color: var(--color-grey--lighter);
   --header-border-color: var(--color-blue);
   --horizontal-inset: var(--space-base);
+  position: relative;
   width: 100%;
-
-  box-shadow: 0 0 0 var(--border-base) var(--border-color);
   border-collapse: collapse;
+}
+
+.table::after {
+  content: "";
+  position: absolute;
+  top: 0;
+  left: 0;
+  z-index: 1;
+  width: calc(100% - 1px);
+  height: 100%;
+  border: solid 1px var(--border-color);
   border-radius: var(--radius-base);
+}
+
+.table * {
+  position: relative;
+  z-index: 2;
 }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Table/Table.css
```diff
@@ -2,13 +2,28 @@
...
+
+.table::after {
+  content: "";
+  position: absolute;
+  top: 0;
+  left: 0;
+  z-index: 1;
+  width: calc(100% - 1px);
+  height: 100%;
+  border: solid 1px var(--border-color);
```
chris-at-jobber, line 18: Can we do `solid var(--border-base) var(--border-color)`?

chris-at-jobber:
@eddysims left a quick comment, I can take a more thorough look and do some testing tomorrow!

packages/components/src/Table/Table.css
chris-at-jobber: Also if the 1px on line 16 is also based on desired border width, this would apply there as well

eddysims: changed it to css custom properties

chris-at-jobber:
Seeing a lil' baby gap here: 
image

eddysims:
> Seeing a lil' baby gap here:
> image

image

@chris-at-jobber Fixed!

chris-at-jobber (approved):
Looks good to me!

PR #429, opened 2020-12-01
feat(components): add keyboard to inputText

whereyoubinns:
## Motivations
Allow a numeric keyboard to be default for inputText on mobile web

### Added
- Allow FormField to accept a `keyboard` prop, default to "text"

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/InputNumber/InputNumber.tsx
```diff
@@ -10,7 +10,7 @@ import { FormField, FormFieldProps } from "../FormField";
 interface InputNumberProps
   extends Pick<
     FormFieldProps,
-    Exclude<keyof FormFieldProps, "type" | "children" | "rows">
+    Exclude<keyof FormFieldProps, "type" | "children" | "rows" | "keyboard">
```
eddysims, line 13: Leaving this out of `InputNumber` will make `InputNumber` show the `text` mode. May want to consider using `decimal` in this case.

https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/#decimal

image

chris-at-jobber:
Looks good from design review ✅  

Should have a code review going this afternoon!

packages/components/src/InputText/InputText.test.tsx
```diff
@@ -108,6 +109,39 @@ it("renders a textarea with 4 rows", () => {
...
+        inputMode="numeric"
+        onBlur={[Function]}
+        onChange={[Function]}
+        onFocus={[Function]}
+        onKeyDown={[Function]}
+        type="text"
+      />
+    </div>
+  `);
+});
```
eddysims, line 143: Can we write a non-snapshot test for this one?

Something along the lines of:

```
const input = getByLabelText("");
expect(input.attr('inputMode', 'numeric))
```

^^ That is super rough, but I think more clear than a snapshot.

eddysims (changes requested):
Whenever we are working in `FormField` its a little more high risk. I wonder if a better option for this would be to exclude it from the input unless its set, instead of setting the default to `text`.

This would prevent all these snapshots from being re-created. As well would just not add the `inputMode` unless the prop was used.

packages/components/src/InputText/InputText.test.tsx
eddysims: Sorry also, maybe this test should be included in `FormField` instead of `InputText` as thats where the prop comes from?

eddysims:
@whereyoubinns a few comments here

packages/components/src/FormField/FormField.tsx
```diff
@@ -296,6 +302,7 @@ export function FormField({
               type={type}
               maxLength={maxLength}
               max={max}
+              inputMode={keyboard}
```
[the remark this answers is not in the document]
whereyoubinns: not to text area. I couldn't think of a use case where you'd want a numeric input over multiple lines

packages/components/src/InputNumber/InputNumber.tsx
whereyoubinns: only including this attribute when keyboard prop is specified makes inputNumber a default numerical keyboard again based on `type`

packages/components/src/InputText/InputText.test.tsx
whereyoubinns: yep good call

whereyoubinns:
@eddysims Can you take another look? Only including the attribute now if the prop is set

packages/components/src/FormField/FormField.tsx
```diff
@@ -262,6 +268,7 @@ export function FormField({
       readOnly: readonly,
       onChange: handleChange,
       value: value,
+      ...(keyboard && { inputMode: keyboard }),
```
eddysims, line 271: I believe that this can just be `inputMode: keyboard`.

If a prop is `undefined` it wont be added to the DOM.

eddysims (changes requested):
@whereyoubinns looking much better. One small final thing and I think its good to go

packages/components/src/FormField/FormField.tsx
whereyoubinns: oh cool, didn't know that

eddysims (approved):

PR #430, opened 2020-12-02
feat(components): convert `progressBar` to semantic HTML and count step 1

chris-at-jobber:
## Motivations

Making some improvements to progressBar that @jlelford noticed while working with Atlantis! Also sneaks in an accessibility improvement I've been wanting to make for some time, using `<progress>` instead of divs so that users of assistive tech can know what;s going on!

Also, it makes sense that on step 1 of N, you see your "currentStep" visualized!

### Changed

- divs are now `<progress>`, and associated CSS changes
- this cuts out some JS calculations we were doing
- Step 1 is visualized as a step!

## References

https://css-tricks.com/html5-progress-element/
---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/ProgressBar/ProgressBar.css
+++ b/packages/components/src/ProgressBar/ProgressBar.css
@@ -1,17 +1,25 @@
 :root {
   --progress-bar--backgroundColor: var(--color-grey--lightest);
-  --progress-bar--barColor: var(--color-lime);
+  --progress-bar--barColor: var(--color-green);
+  --progress-bar--height: var(--space-base);
 }
 
-.container,
-.content {
+progress[value] {
   width: 100%;
-  height: var(--progress-bar--height);
-  border-radius: calc(var(--progress-bar--height) / 2);
-  background: var(--progress-bar--backgroundColor);
+  /* Reset the default appearance */
+  border: none;
+  -webkit-appearance: none;
+  appearance: none;
 }
 
-.content {
+progress[value]::-webkit-progress-bar {
+  box-shadow: inset 0px 2px 4px rgba(0, 0, 0, 0.05);
+  border-radius: calc(var(--progress-bar--height) / 2);
+  background-color: var(--progress-bar--backgroundColor);
+}
+
+progress[value]::-webkit-progress-value {
+  border-radius: calc(var(--progress-bar--height) / 2);
   background-color: var(--progress-bar--barColor);
-  transition: width var(--timing-slower) ease-out;
+  transition: all var(--timing-slow) ease-out;
 }
--- a/packages/components/src/ProgressBar/ProgressBar.css
+++ b/packages/components/src/ProgressBar/ProgressBar.css
@@ -6,19 +6,27 @@
 
 progress[value] {
   width: 100%;
-  /* Reset the default appearance */
   border: none;
-  -webkit-appearance: none;
   appearance: none;
 }
 
+progress[value],
 progress[value]::-webkit-progress-bar {
   box-shadow: inset 0px 2px 4px rgba(0, 0, 0, 0.05);
   border-radius: calc(var(--progress-bar--height) / 2);
+}
+
+progress[value]::-webkit-progress-bar {
   background-color: var(--progress-bar--backgroundColor);
 }
 
 progress[value]::-webkit-progress-value {
+  border-radius: calc(var(--progress-bar--height) / 2);
+  background-color: var(--progress-bar--barColor);
+  transition: all var(--timing-slow) ease-out;
+}
+
+progress[value]::-moz-progress-bar {
   border-radius: calc(var(--progress-bar--height) / 2);
   background-color: var(--progress-bar--barColor);
   transition: all var(--timing-slow) ease-out;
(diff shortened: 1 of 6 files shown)


chris-at-jobber:
TODO: get sizes working, wrestle with Typescript things?

packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -26,20 +26,12 @@ export function ProgressBar({
   totalSteps,
   size = "base",
 }: ProgressBarProps) {
-  const percentage = ((currentStep - 1) / (totalSteps - 1)) * 100;
-  const widthPercent = Math.min(100, Math.max(0, percentage));
-  const progressBarContainerClassName = classnames(
-    styles.container,
-    sizes[size],
-  );
+  const percentage = (currentStep / totalSteps) * 100;
```
chris-at-jobber, line 29: I think I could actually remove this, and change the accessibility text to `{currentStep} of {totalSteps} complete`

packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -26,20 +26,12 @@ export function ProgressBar({
...
   const progressBarContentClassName = classnames(styles.content);
```
chris-at-jobber, line 30: I am very sure I don't need this, or `styles.container` above, but when I take this one out the progress element loses all of its styling... which I don't understand because I don't even have a classname called `.content` in either the CSS or the rendered HTML anymore?

image

chris-at-jobber: I do still need the `styles[size]` to allow the "small" prop to work

packages/components/src/ProgressBar/ProgressBar.css
```diff
@@ -1,17 +1,25 @@
...
+progress[value] {
   width: 100%;
-  height: var(--progress-bar--height);
+  /* Reset the default appearance */
+  border: none;
+  -webkit-appearance: none;
+  appearance: none;
+}
+
+progress[value]::-webkit-progress-bar {
```
chris-at-jobber, line 15: This uses the webkit prefix, there's also a -moz- prefix... do I need to add these prefixes or should autoprefixer bail me out?

chris-at-jobber: Figured it out, autoprefixer won't bail me out

- MOZ and WEBKIT actually use _different pseudo classes_ so trying to combine them into one just doesn't work

chris-at-jobber: They two pseudo classes for webkit don't have the same styling concerns as the one element for moz

packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -26,20 +26,11 @@ export function ProgressBar({
...
 
   return (
-    <div className={progressBarContainerClassName}>
-      <div
-        className={progressBarContentClassName}
-        style={{ width: `${widthPercent}%` }}
-      />
-    </div>
+    <progress max={totalSteps} value={currentStep}>
+      {currentStep} of {totalSteps}
```
chris-at-jobber, line 33: Works for both steps and loading

ie "248 of 512"

Maybe a fancier way could be adding a prop for type of progress? Steps vs filesize, and units?

But I'm thinking the context of whatever you did before this action would be enough here

seanhealy:
This is probably out of scope but, I would LOVE if we could change these props to just accept a number between `0` and `1` instead of everything else we have going on in here. I think it would simplify the usage a lot. Perhaps also allow `undefined` to display an indeterminate state?

packages/components/src/ProgressBar/ProgressBar.tsx
chris-at-jobber: Conversation: will not remove, because this actually allows the accessibility text to more accurately reflect the information the ProgressBar conveys

chris-at-jobber:
@seanhealy I need to generate new snapshots (how???) but otherwise I think this is ready for another 👀

chris-at-jobber:
Tested appearance in:

- [ ] Chrome
- [ ] Safari
- [ ] Firefox (the bar doesn't transition as smoothly from step to step, but that's a trade off I am comfortable with)
- [ ] Opera
- [ ] Edge

whereyoubinns:
@chris-at-jobber very nice! 

Just a callout that some usages i've seen directly rely on the progress step to check for certain conditions as well as setting the step directly to 2 to see that first step.

Since previously both step 0 and step 1 rendered an empty bar and with this change step 1 now is visual this might require some changes wherever this is implemented

chris-at-jobber:
Good call, thanks @whereyoubinns !

seanhealy:
Looks good. Only comment I'd have is to change the example under `Example with State` so that the decrement button mins out at 0.

```<button onClick={() => setStep(Math.max(0, step - 1))}>```

chris-at-jobber:
@seanhealy PTAL 🙏

seanhealy (approved):

PR #431, opened 2020-12-02
feat(components): Add dismissible prop to banner

larymc:
## Motivations
 - We do not want the banner reminding SP to setup 2FA phone number to be dismissible. Adding a prop to make non-dismissible banner available.

### Added
- dismissible prop in banner component.
- `dismissible` is true (default)
image

- `dismissible` is false
image

## Testing
 - Tested with playground in Atlantis.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Banner/Banner.test.tsx
+++ b/packages/components/src/Banner/Banner.test.tsx
@@ -29,6 +29,17 @@ it("renders a warning banner", () => {
   expect(tree).toMatchSnapshot();
 });
 
+it("renders without close button", () => {
+  const tree = renderer
+    .create(
+      <Banner type="warning" dismissible={false}>
+        Warn
+      </Banner>,
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
 test("it should call the handler with a number value", () => {
   const changeHandler = jest.fn();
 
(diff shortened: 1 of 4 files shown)


packages/components/src/Banner/Banner.test.tsx
```diff
@@ -29,6 +29,17 @@ it("renders a warning banner", () => {
...
+it("renders without close button", () => {
+  const tree = renderer
+    .create(
+      <Banner type="warning" dismissible={false}>
+        Warn
+      </Banner>,
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
+});
```
eddysims, line 41: Can we achieve this test without a snapshot? would something like below be a more suitable option?

```
const { queryByLabelText } = render(<Banner dismissible={false}>Foo</Banner>
expect(queryByLabelText("Close")).toBeNull()
```

eddysims (changes requested):

eddysims:
@larymc just one small comment here.

packages/components/src/Banner/Banner.test.tsx
larymc: @eddysims Updated the test and also removed the snapshot.  Thanks!

eddysims (approved):

PR #432, opened 2020-12-02
feat(hooks): Add useOnKeyDown hook

linds14sr20det:
## Motivations

This PR adds a onKeyDown event hook. This will allow us to implement some UX niceties, for example automatically setting focus to a button when a user completes some input action.

This PR is almost entirely the work of @0length. This feature was built as a Hacktoberfest project. I've simply cleaned up his code and cleaned up the usage examples. All credit goes to him. The original PR description was:
```
### Overview 
This is refer to https://github.com/GetJobber/atlantis/pull/394#discussion_r509685744
### Changed

Create new hook called `useOnKeyDown` to create on keydown event listener handler by specific one or multiple key in the same function.
**Thanks**
```

### Added

- Generic onKeyDown hook.

## Testing

 - Ensure that the mdx file examples function as expected.

---

Random photo of Atlantis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3142,9 +3142,9 @@
       "version": "file:packages/components",
       "dev": true,
       "requires": {
-        "@jobber/design": "^0.9.1",
+        "@jobber/design": "^0.6.1",
         "@jobber/formatters": "^0.0.4",
-        "@jobber/hooks": "^1.0.0",
+        "@jobber/hooks": "^0.2.1",
         "@popperjs/core": "^2.0.6",
         "@std-proposal/temporal": "0.0.1",
         "@types/color": "^3.0.1",
@@ -3152,14 +3152,10 @@
         "@types/react-router-dom": "^5.1.5",
         "@types/uuid": "^3.4.5",
         "@use-it/event-listener": "^0.1.3",
-        "axios": "^0.19.2",
         "classnames": "^2.2.6",
         "color": "^3.1.2",
-        "filesize": "^6.1.0",
         "framer-motion": "^1.10.3",
         "lodash": "^4.17.15",
-        "react-dropzone": "^11.0.2",
-        "react-hook-form": "^6.7.2",
         "react-image-lightbox": "^5.1.1",
         "react-markdown": "^4.1.0",
         "react-router-dom": "^5.2.0",
@@ -4577,12 +4573,6 @@
           "version": "1.1.2",
           "resolved": "https://registry.npmjs.org/fastparse/-/fastparse-1.1.2.tgz",
           "integrity": "sha512-483XLLxTVIwWK3QTrMGRqUfUpoOs/0hbQrl2oz4J0pAcm3A3bu84wxTFqGqkJzewCLdME38xJLJAxBABfQT8sQ=="
-        },
-        "filesize": {
-          "version": "6.1.0",
-          "resolved": "https://registry.npmjs.org/filesize/-/filesize-6.1.0.tgz",
-          "integrity": "sha512-LpCHtPQ3sFx67z+uh2HnSyWSLLu5Jxo21795uRDuar/EOuYWXib5EmPaGIBuSnRqH2IODiKA2k5re/K9OnN/Yg==",
-          "dev": true
         },
         "fill-range": {
           "version": "4.0.0",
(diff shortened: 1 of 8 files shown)


packages/hooks/src/useOnKeyDown/useOnKeyDown.ts
```diff
@@ -0,0 +1,51 @@
...
+      keys.some(item => {
+        if (typeof item === "string") return keyboardEvent.key === item;
+        return Object.keys(item).every(
+          index => keyboardEvent[index] === item[index],
+        );
+      })
+    ) {
+      callback(event);
+    }
+    if (
```
seanhealy, line 43:
```suggestion

    if (
```

packages/hooks/src/useOnKeyDown/useOnKeyDown.ts
```diff
@@ -0,0 +1,51 @@
...
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  //@ts-ignore
+  useEventListener<KeyboardEvent>("keydown", handler);
+
+  function handler(event: globalThis.KeyboardEvent) {
+    const keyboardEvent = (event as unknown) as VerboseKeyComparator;
+    if (typeof keys === "string" && keyboardEvent.key === keys) {
+      callback(event);
+    }
+    if (
```
seanhealy, line 32:
```suggestion

    if (
```

packages/hooks/src/useOnKeyDown/useOnKeyDown.ts
```diff
@@ -0,0 +1,51 @@
...
+  // Pending: https://github.com/donavon/use-event-listener/pull/12
+  // The types in useEventListener mistakenly require a SyntheticEvent for the passed generic.
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  //@ts-ignore
+  useEventListener<KeyboardEvent>("keydown", handler);
+
+  function handler(event: globalThis.KeyboardEvent) {
+    const keyboardEvent = (event as unknown) as VerboseKeyComparator;
+    if (typeof keys === "string" && keyboardEvent.key === keys) {
+      callback(event);
```
seanhealy, line 30: I wonder if we might want to explicitly `return` after the `callback` in these conditions. While we shouldn't be able to have two of them be true it might not be a bad safety net?

packages/hooks/src/useOnKeyDown/useOnKeyDown.ts
```diff
@@ -0,0 +1,51 @@
...
+type KeyComparator = XOR<VerboseKeyComparator, SimpleKeyComparator>;
+
+export function useOnKeyDown(
+  callback: (event: globalThis.KeyboardEvent) => void,
+  keys: KeyComparator[] | KeyComparator,
+) {
+  // Pending: https://github.com/donavon/use-event-listener/pull/12
+  // The types in useEventListener mistakenly require a SyntheticEvent for the passed generic.
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  //@ts-ignore
```
seanhealy, line 24: Looks like the underlying issue requiring the ignore here has been fixed upstream wanna check if we can get rid of this?

packages/hooks/src/useOnKeyDown/useOnKeyDown.mdx
```diff
@@ -0,0 +1,75 @@
...
+    useOnKeyDown(
+      e => {
+        setListText("You pressed '" + e.key + "'");
+      },
+      ["Shift", "Enter"]
+    );
+    return (
+      <Card title="Shift or Enter Example">
+        <Content>
+          <pre>{listText}</pre>
```
seanhealy, line 43:
```suggestion
        <Content>
        Press shift or enter.
          <pre>{listText}</pre>
```

Thought we could keep the instructions around here. Or, we could also use the card title?

seanhealy (changes requested):
Few quick little thoughts but think we're close here.

package-lock.json
```diff
@@ -9720,42 +9730,105 @@
       }
     },
     "@testing-library/dom": {
-      "version": "7.16.2",
-      "resolved": "https://registry.npmjs.org/@testing-library/dom/-/dom-7.16.2.tgz",
-      "integrity": "sha512-4fT5l5L+5gfNhUZVCg0wnSszbRJ7A1ZHEz32v7OzH3mcY5lUsK++brI3IB2L9F5zO4kSDc2TRGEVa8v2hgl9vA==",
+      "version": "7.28.1",
```
seanhealy, line 9733: Since we didn't have a change to the corresponding `package.json` file we can probably revert these `package-lock.json` changes as I'm guessing they are a remnant of the package tomfoolery we were going through yesterday.

seanhealy (changes requested):
@linds14sr20det One quick change then I think we're good!

seanhealy (approved):

PR #433, opened 2020-12-02
feat: add required column to component proposal template

hosseintalebi:
## Motivations

We noticed in the review process of component proposals that we couldn't figure out if a property was required or not. 

## Changes

Added a `required` column to the prop table in the proposal template.

### Added

Added a `required` column to the prop table in the proposal template.

image

### DIFF
--- a/docs/proposals/TEMPLATE.md
+++ b/docs/proposals/TEMPLATE.md
@@ -34,6 +34,6 @@ it from another react component? This should consist primarily of code blocks}_
 
 _{Provide a table in the following format of the component's public API}_
 
-| name | type | required | default | description |
-| ---- | ---- | -------- | ------- | ----------- |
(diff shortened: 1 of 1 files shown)


docs/proposals/TEMPLATE.md
```diff
@@ -34,6 +34,6 @@ it from another react component? This should consist primarily of code blocks}_
 
 _{Provide a table in the following format of the component's public API}_
 
-| name | type | default | description |
-| ---- | ---- | ------- | ----------- |
-| ...  | ...  | ...     | ...         |
+| name | type | required | default | description |
```
JeffMarvin, line 37: Consider `required?` (with the question mark) to hint at the boolean nature/expectation....

seanhealy (approved):

PR #435, opened 2020-12-05
feat(components): Allow text to align center in banner

larymc:
## Motivations
 - We would like to use Atlantis banner component in global banner in Jobber online.  Global banner's text is center aligned so we add the ability to align text center in Atlantis banner.

### Added

 - `align` prop in banner component.
 - `align` is left (default)
image

 - `align` is center
image

## Testing
 - Tested in Atlantis playground.

---

Random photo of Atlantis


packages/components/src/Banner/Banner.test.tsx
```diff
@@ -29,6 +29,17 @@ it("renders a warning banner", () => {
...
 
+it("renders centered text", () => {
+  const tree = renderer
+    .create(
+      <Banner type="warning" align="center">
+        Center
+      </Banner>,
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
```
larymc, line 40: Is it okay to use snapshot or shall we check if classname `centerText` exists?

chris-at-jobber:
Looking at the proposal for [Account Banner](https://www.figma.com/file/AZFSe2T7FID7d5DP52NTkX/Design-System-Contribution-Account-Banner?node-id=30%3A581) that @jlelford put together, the text is left aligned like our other banners. Is there 

a) a pressing need to center the text on a global `Banner`?

or

b) a case where we just center the text in `Banner`?

I'm not sure it's helpful from a component standpoint to add presentation distinctions that don't have a meaning behind them. My main concern with this is that we're not tying the presentation change to the use case, and teams having to make the decision on whether or not to center text becomes a case-by-case basis.

larymc:
a) I don't think there is a pressing need for center the text on global banner, just that the existing global banner is center so I was thinking to keep it as is.

b) Yes, you are right, we can center the text in the parent tag.

Thanks for your comment @chris-at-jobber , I'll close this PR.

PR #443, opened 2020-12-18
feat(components)!: Visual and layout updates to banner

chris-at-jobber:
## Motivations

Had some good talks with @jlelford and @larymc about some new directions for `Banner` that will make it very nice for everyone! We'll remove the Banner's responsibility for how it is laid out within a view, and make some other changes to allow for it to be used more flexibly.

Also, as is my wont, some accessibility improvements are making their way into Banner.

### Added

- new role of "status" for assistive technologies. **_I still need to get it working to update to "alert" when the banner is of type "error"!_**
- new Atlantis variable of `--shadow-outline` for use cases where we're not using CSS `outline` to add a focus state

### Changed

- remove border-top for all instances
- remove border-radius on bottom
- add hover and focus states for close button
- add shadow

## Testing

Test it [here](https://5ff4bbfda7776e00071ac0b1--jobber-atlantis.netlify.app/components/banner)

Banner should:

- [ ] take up 100% of available width
- [ ] have a box-shadow
- [ ] have no border-radii or top border

When Banner has a primary action:

...and Banner is wider than 480px

- [ ] text should take up as much space as needed up until button, then wrap to a new line

...and Banner is less than 480px

- [ ] text should take up as much space as needed, pushing button to a new line if necessary
- [ ] if text wraps, there should be even spacing between content, action, and edges of Banner

When Banner is dismissible:

- [ ] closeButton should always be pinned to the top right of Banner

---

Random photo of Atlantis


packages/components/src/Banner/Banner.tsx
```diff
@@ -53,22 +66,31 @@ export function Banner({
...
-              <Icon name="cross" color={iconColors[type]} />
-            </button>
+            <div className={styles.dismissWrapper}>
+              <button
+                className={styles.closeButton}
+                onClick={handleClose}
+                aria-label="Close"
+              >
+                <Icon name="cross" color={iconColors[type]} />
+              </button>
```
eddysims, line 92: We have a component `ButtonDismiss` that uses a `Button` component. Not sure if this is something we want to use here, but just double checking.

packages/components/src/Banner/Banner.css
```diff
@@ -1,27 +1,63 @@
...
+  height: var(--space-large);
+  padding-left: var(--flash--inset);
+  align-self: stretch;
 }
 
 .closeButton {
-  width: var(--space-larger);
-  margin-top: calc(var(--space-smallest) - var(--flash--inset));
-  margin-right: calc(var(--space-smallest) - var(--flash--inset));
+  position: absolute;
```
eddysims, line 58: I feel this should be on the `dismissWrapper` instead of the button. That way we aren't leaving an empty `div` in the DOM flow

packages/components/src/Banner/Banner.css
```diff
@@ -1,27 +1,63 @@
...
-  padding: var(--flash--inset);
-  border-top: var(--border-thick) solid var(--color-grey--light);
-  border-bottom-right-radius: var(--space-smaller);
-  border-bottom-left-radius: var(--space-smaller);
+  box-shadow: var(--shadow-base);
+  padding: var(--flash--inset--small) var(--flash--inset);
   color: var(--color-black);
   background-color: var(--color-grey--lightest);
-  align-items: center;
+  align-items: start;
```
eddysims, line 12: This give a warning in `docz`. Let's change to `flex-start`

image

eddysims (changes requested):
@chris-at-jobber a few comments for the first pass here. Looking good!

packages/components/src/Banner/Banner.tsx
chris-at-jobber: I tried and it's a bit more opinionated (solid grey fill on hover/focus). I think we could make some changes to `ButtonDismiss`, but I think that could be a separate PR?

chris-at-jobber:
@eddysims changes addressed!

packages/components/src/Banner/Banner.css
```diff
@@ -30,6 +63,16 @@
   flex: 0;
 }
 
+/* .medium .closeButton {
+  top: var(--flash--inset);
+} */
```
eddysims, line 68: oops here

packages/components/src/Banner/Banner.css
```diff
@@ -30,6 +63,16 @@
...
+
+.closeButton:hover {
+  background: rgba(0, 0, 0, 0.025);
+}
+
 .closeButton:focus {
+  box-shadow: 0 0 1px 0.125rem var(--color-outline);
```
eddysims, line 75: Is there a reason that we didnt use the `--shadow-base` here?

eddysims (changes requested):
just 2 small things @chris-at-jobber

packages/components/src/Banner/Banner.css
chris-at-jobber: 🤦

packages/components/src/Banner/Banner.css
chris-at-jobber: `--shadow-base` is for things that are meant to have a "shadow" appearance to indicate depth/elevation (toast, menu, etc), where this is simulating an outline. We don't use actual outline because our current style across the product has more of a "glow" than default outline options.

chris-at-jobber: Do you think it's worth consolidating this style in an `--shadow-outline` variable though??

eddysims: if you think that we will use this pattern again, then yes i think so.

eddysims:
@chris-at-jobber I think just the css var and then it looks good

chris-at-jobber:
@eddysims variable created!

eddysims (approved):

PR #447, opened 2021-01-07
fix(components): allow Toast to expand on small screens

chris-at-jobber:
## Motivations

Toast gets compacted quite a bit on small screens, even when there is sufficient room for it to breathe!
image

## Changes

Allows the Toast to take up the full width of the screen, minus equivalent spacing on left, right and bottom (uses --space-large on all 3)

### Changed

- Some CSS changes to allow Toast to expand

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Toast/Toast.css
+++ b/packages/components/src/Toast/Toast.css
@@ -5,20 +5,25 @@
 
 .container {
   display: flex;
+  position: fixed;
+  right: 0;
+  bottom: var(--space-large);
+  left: 0;
+  z-index: var(--elevation-toast);
+  width: 100%;
+  box-sizing: border-box;
   flex-direction: column;
   align-items: center;
-  position: fixed;
-  bottom: var(--space-large);
-  left: 50%;
-  z-index: var(--elevation-toast);
+  padding: 0 var(--space-large);
   color: var(--color-greyBlue--dark);
   font-size: var(--typography--fontSize-large);
-  transform: translateX(-50%);
+  pointer-events: none;
 }
 
 .toast {
   box-shadow: var(--shadow-base);
   overflow: hidden;
+  pointer-events: all;
 }
 
 .toast + .toast {


packages/components/src/Toast/Toast.css
```diff
@@ -5,20 +5,25 @@
...
   z-index: var(--elevation-toast);
+  width: 100%;
+  box-sizing: border-box;
+  flex-direction: column;
+  align-items: center;
+  padding: 0 var(--space-large);
   color: var(--color-greyBlue--dark);
   font-size: var(--typography--fontSize-large);
-  transform: translateX(-50%);
+  pointer-events: none;
```
chris-at-jobber, line 20: This is probably the spiciest thing I'm doing in this PR... it simplifies the flex layout and positioning, and lets us kick the transform out though

packages/components/src/Toast/Toast.css
```diff
@@ -5,20 +5,25 @@
...
   display: flex;
-  flex-direction: column;
-  align-items: center;
   position: fixed;
+  right: 0;
   bottom: var(--space-large);
-  left: 50%;
+  left: 0;
   z-index: var(--elevation-toast);
+  width: 100%;
```
chris-at-jobber, line 13: You might think "should we cap this width on larger screens?" and I was thinking maybe, but the guidelines around `Toast` recommend short messages so I think it's fine to not cap this until we run into an issue

JeffMarvin: ^ The content of these toast messages could get created/pumped in from something that a developer isn't paying much attention to, or the content changes later and nobody notices that the toast got out of control. 
How much work would it be to have the Toast be a bit more "defensive" or robust in this case? It might be worth it to include now.

packages/components/src/Toast/Toast.css
JeffMarvin: Yeah, does this effectively remove the hit area from the whole toast, and just make it the dismiss button? I think I am still ok with that because we are self-dismissing these. Any accessibility concerns here?

JeffMarvin (commented):
QA looks good (tested all modern browsers, small and large screen sizes).

Just one question about a11y and the smaller hit area.

packages/components/src/Toast/Toast.css
chris-at-jobber: This allows for click-through _behind_ the transparent container for `Toast` - then on the Toast itself I'm adding back the pointer-events so the user can select text, dismiss, etc. So the actual hit areas aren't changing, just making sure that if there was some element the user is trying to interact with that's behind the container, the user can still do so!

JeffMarvin (approved):

PR #449, opened 2021-01-11
build: @jobber/docz-tools package

eddysims:
## Motivations

We want to start using `docz` in other projects. When this happens, we want to be able to have the same theme that we have in Atlantis in other areas. This PR creates a new package for `@jobber/docz-tools` that will allow us to install the theme in other installs of docz.

We also want to make our `docz` theme more accessible and build it more with Atlantis components. There will be quite a few places in this new theme that will now use Atlantis components.

## Changes

Creates a new `@jobber/docz-tools` package that contains the `docz` theme for Atlantis.

### Added

- new `@jobber/docz-tools` package.

### Changed

- Makes theme use more Atlantis components in its build

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- pull down branch and run `npm ci`
- run `npm run clean && npm start` and see that the theme is the new theme
- run `npm run clean && npm run build && npm run serve` and see that the built code is using the new theme.

---

Random photo of Atlantis


packages/docz-tools/src/gatsby-theme-docz/components/Navigation/LevelThreeLink.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { levelThreeLink } from "./styles";
+
+interface LevelThreeLinkProps {
+  readonly title: string;
+  readonly slug: string;
+}
+
+export function LevelThreeLink({ title, slug }: LevelThreeLinkProps) {
+  const link = `#${slug}`;
+  const currentHash = getCurrentHash();
```
seanhealy, line 14: If they're using react router it may have a hook for this?

packages/docz-tools/src/gatsby-theme-docz/components/Navigation/styles.ts
```diff
@@ -0,0 +1,76 @@
+export const navigation = {
+  pt: "base",
```
seanhealy, line 2: We talked about expanding these shorthands to their more verbose form. I think that would still make this more readable.

packages/docz-tools/src/gatsby-theme-docz/components/Playground/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+            </Box>
+          </Box>
+
+          {showLiveError && <LiveError sx={styles.error} />}
+        </LiveProvider>
+      </Resizable>
+    </Box>
+  );
+
+  function transformCode() {
```
seanhealy, line 64: Curious why we need this? :)

packages/docz-tools/src/gatsby-theme-docz/components/Props/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  if (defaultValue.value === "''") {
+    return "[Empty string]";
+  }
+
+  if (propType && propType.name === "string") {
+    return defaultValue.value.replace(/'/g, '"');
+  }
+
+  // @ts-expect-error
```
seanhealy, line 96: What is this about again?

packages/docz-tools/src/gatsby-theme-docz/components/index.ts
```diff
@@ -0,0 +1,29 @@
+// @ts-expect-error
```
seanhealy, line 1: Want to look at these together?

packages/docz-tools/src/gatsby-theme-docz/theme/global.ts
```diff
@@ -0,0 +1,5 @@
+export const global = {
```
seanhealy, line 1: Is this an override?

packages/docz-tools/src/gatsby-theme-docz/theme/space.ts
```diff
@@ -0,0 +1,11 @@
+export const space = {
```
seanhealy, line 1: I kind of wonder if we should make a followup to have these generated.

seanhealy (changes requested):
Few thoughts back your way.

doczrc.js
```diff
@@ -43,112 +12,59 @@ export default {
     "Atlantis is a design system for Jobber. The primary objective for Atlantis is to provide a system of reusable components to help developers to quickly build beautiful and consistent interfaces for our users.",
   typescript: true,
   port: 3333,
+  host: "0.0.0.0",
```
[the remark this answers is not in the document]
eddysims: we don't **need** to, but what host allows is for your local docz to be available on your network. This means developing a component can be done on a mobile device. Would have been very handy when we were doing things like the keyboards for `InputText`.

ensureChangelogHeaders.js
```diff
@@ -29,7 +29,6 @@ glob.sync(path.join("packages/*/CHANGELOG.md")).forEach(file => {
 name: ${startCase(directoryName)}
 menu: Changelog
 ---
-
```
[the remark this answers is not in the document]
eddysims: nope, that was most likely an accident.

packages/docz-tools/gatsby-node.js
```diff
@@ -0,0 +1,64 @@
...
+      const areCssRules =
+        rule.oneOf && rule.oneOf.some(r => r.test.test("style.css"));
+
+      return !areCssRules;
+    }),
+    libCssRule,
+    cssRule,
+  ];
+
+  if (JOBBER_ALIAS) {
```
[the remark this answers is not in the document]
eddysims: Yes, this is actually for when we use this outside of `Atlantis`. What i was finding is that the alias needs to be different in atlantis then in a new docz project. The other option would be to copy this out to the root level `gatsby-node` and do another alias there that would override this one.

May be confusing to read. Would be happy to chat about this one, as there are a couple options.

packages/docz-tools/index.js
```diff
@@ -0,0 +1 @@
+// the Jobbler says shush
```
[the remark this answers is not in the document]
eddysims: Somewhere along the way I found that a gatsby plugin (which docz-tools essentially becomes), needs a `main` in the `package.json`. I beleive i was finding that it was causing an error pointing it into the `src` dir. 

I think that we will also at some point want to export components out of this package. This `index.js` would let us do that.

packages/docz-tools/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "dependencies": {
+    "@emotion/react": "^11.1.4",
+    "@jobber/components": "^2.22.1",
+    "@jobber/design": "^0.9.2",
+    "deepmerge": "^4.2.2",
+    "ts-xor": "^1.0.8"
+  },
+  "devDependencies": {},
+  "peerDependencies": {
+    "docz": "^2.3.1",
```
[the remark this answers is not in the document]
eddysims: This is another weird one. I actually want to change this to `*`.

The problem here is that when these are included as `dependencies`, there is some wildly confusing errors that come out due to mis-matched versions of `react`/`react-dom`/`docz`/`theme-ui`. My thinking is that everything that is in `peerDependencies` comes with `docz` out of the box, so they should always be there.

doczrc.js
```diff
@@ -43,112 +12,59 @@ export default {
...
   menu: ["Atlantis", "Patterns", "Components", "Hooks", "Design"],
-  files: "{README.md,CONTRIBUTING.md,**/*.mdx,packages/*/CHANGELOG.md}",
+  files: [
+    "./docs/**/*.{md,mdx}",
+    ...packages.map(pckg => `./packages/${pckg}/src/**/*.{md,mdx}`),
```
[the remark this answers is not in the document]
eddysims: `package` is a reserved keyword.

packages/docz-tools/gatsby-node.js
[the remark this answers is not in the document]
eddysims: I have renamed the var to `USE_ATLANTIS_ALIASES`

packages/docz-tools/src/gatsby-theme-docz/components/Navigation/LevelThreeLink.tsx
eddysims: This is copied from docz

packages/docz-tools/src/gatsby-theme-docz/components/Playground/index.tsx
eddysims: Also copied from docz

packages/docz-tools/src/gatsby-theme-docz/components/Actions/index.tsx
```diff
@@ -0,0 +1,56 @@
+/** @jsx jsx */
+import { Box, jsx } from "theme-ui";
+import { useConfig, useCurrentDoc } from "docz";
+import { Button } from "@jobber/components/Button";
+import { Tooltip } from "@jobber/components/Tooltip";
+import * as styles from "./styles";
```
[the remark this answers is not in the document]
eddysims: we cant, as there is no default export from the `./styles`.

packages/docz-tools/src/gatsby-theme-docz/components/Actions/styles.ts
```diff
@@ -0,0 +1,8 @@
+export const actions = {
+  position: "fixed",
+  top: "large",
```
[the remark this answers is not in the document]
eddysims: yes, however I have just tested and we can just use css modules, which will be nicer I think. Im going to add a backlog item to change some of these components.

packages/docz-tools/src/gatsby-theme-docz/theme/space.ts
eddysims: I think we can look at actually getting off of theme ui as a follow up task. Might make more sense for us.

packages/docz-tools/src/gatsby-theme-docz/theme/global.ts
eddysims: This is removed

packages/docz-tools/src/gatsby-theme-docz/components/Navigation/styles.ts
eddysims: Do you think we should look at that, or handle this in a backlog item to remove theme-ui

packages/docz-tools/src/gatsby-theme-docz/components/index.ts
eddysims: what are your thoughts on a follow up for this?

packages/docz-tools/src/gatsby-theme-docz/components/Actions/styles.ts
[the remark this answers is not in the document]
eddysims: Yes, this refers to `theme.space.large` in this case.

eddysims:
@seanhealy I think I have addressed everything here!

packages/docz-tools/src/gatsby-theme-docz/components/index.ts
seanhealy: I'm fine with that if you wanna make a task for us to do so.

seanhealy (changes requested):
Two last comments and I _think_ we're good.

eddysims:
@seanhealy comments addressed. Back to you!

packages/docz-tools/gatsby-node.js
```diff
@@ -0,0 +1,65 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-env node */
+const path = require("path");
+
+const { USE_ATLANTIS_ALIASES } = process.env;
+const useAtlantisAliases = USE_ATLANTIS_ALIASES == "true";
```
seanhealy, line 6: Did you use `==` for a reason here?

```suggestion
const useAtlantisAliases = USE_ATLANTIS_ALIASES === "true";
```

seanhealy (changes requested):
One last real little thought and I think we're good!

seanhealy (approved):

PR #452, opened 2021-01-15
feat(components): add iOS autocomplete support for OTPs to form fields

linds14sr20det:
## Motivations

The input text component may be used to handle OTP's. As such to support safari and iMessage OTP integration we are adding `"one-time-code"` to the props for form fields.

## Changes

Adding `"one-time-code"` to the props for form fields.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/FormField/FormField.tsx
```diff
@@ -22,7 +22,7 @@ export interface FormFieldProps {
   /**
    * Determines if browser form autocomplete is enabled.
```
seanhealy, line 23: Want to add something here along the lines of `"one-time-code"` is considered experimental and should not be used without consultation?

packages/components/src/FormField/FormField.tsx
```diff
@@ -22,7 +22,7 @@ export interface FormFieldProps {
...
    */
-  readonly autocomplete?: boolean;
+  readonly autocomplete?: boolean | "one-time-code";
```
seanhealy, line 25: I don't think that this will actually work.

https://github.com/GetJobber/atlantis/pull/452/files#diff-7664eb8fb7178dac2dae4e5ccb337b655a800ab7dade9a2ec9501c0d428c78f7R221

seanhealy (changes requested):

seanhealy (approved):

PR #453, opened 2021-01-18
fix(components): change line-height on large font size

ellaestigoy:
## Motivations

The readability of the large text is not great, so adjusting the line height will help.

## Changes

CSS change on large font size `line-height` to large

### Changed

- Change line-height from `-tightest` to `-large`

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Typography/css/FontSizes.css
+++ b/packages/components/src/Typography/css/FontSizes.css
@@ -20,7 +20,7 @@
 
 .large {
   font-size: var(--typography--fontSize-large);
-  line-height: var(--typography--lineHeight-tightest);
+  line-height: var(--typography--lineHeight-large);
 }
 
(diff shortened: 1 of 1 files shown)


chris-at-jobber:
Tested:

- [ ] New line-height looks good when used in Text
- [ ] New line-height looks good when used in Heading
- [ ] New line-height looks good when used in Modal, Card, Drawer headers

chris-at-jobber:
image
image

A consideration I'm going to take for future is to consider splitting out LineHeights into their own concern, instead of coupling them with fonts-size! This looks OK but is a side effect of how tying the two together has some knock-on effects that make testing a bit tougher.

chris-at-jobber (approved):
Looks good! Note for future me added 😄

PR #455, opened 2021-01-19
fix: "not-allowed" cursor style should work for Disabled Button

skwidz:
## Motivations

Disabled buttons should have a `not-allowed` cursor on mouse-hover

### Fixed
- removed a `pointer-events: none` styling from Disabled Button, fixing a broken cursor definition 
!-- for any bug fixes -->

## Testing
Checkout the branch, and go to the Button's component page

Hovering over the `Disabled` Buttons should now show the `not-available`  cursor (crossed-out circle) 

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -137,7 +137,6 @@ a.tertiary:focus {
   border-color: var(--button--color-disabled);
   user-select: none;
   background-color: var(--button--color-disabled);
-  pointer-events: none;
   cursor: not-allowed;
 }
 


chris-at-jobber:
There'll need to be a bit of CSS to catch the versions of `Button` that have a `url` prop:
image

Right now, because that's rendering an `a` element with more specific styling, we need to make sure those elements aren't picking up a hover state when disabled.

@eddysims do you have any suggestions here?

eddysims:
ugh sorry @skwidz i keep forgetting to finish my comment. 

As chris mentioned, there is going to be some work with `pointer-events` for the `disabled` button with a `url` or `to`. 


Some options that come to mind would be to check for the disabled state before applying the `url` or `to` to the button.  Or (less ideal) display a different tag like a `span` if there is a disabled state.

chris-at-jobber:
Closing off as I don't think there's a great urgency on this one - we can re-open if needed

PR #456, opened 2021-01-22
feat(components): Spinner

whereyoubinns:
## Motivations

Continues the work of https://github.com/GetJobber/atlantis/pull/177 to add a spinner component to indicate loading states

### Added

Spinner component

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Spinner/Spinner.css
+++ b/packages/components/src/Spinner/Spinner.css
@@ -1,0 +1,44 @@
+.spinner {
+  width: var(--space-larger);
+  height: var(--space-larger);
+  box-shadow: var(--shadow-dark);
+  border: solid var(--color-blue--light);
+  border-top-color: var(--color-white);
+  border-width: calc(var(--space-small) * 0.75);
+  border-radius: 50%;
+  vertical-align: middle;
+  background-color: var(--color-blue--lightest);
+  animation-duration: 1s;
+  animation-iteration-count: infinite;
+  animation-name: animation-spin;
+  animation-timing-function: linear;
+  will-change: transform;
+}
+
+@keyframes animation-spin {
+  0% {
+    transform: rotate(0deg);
+  }
+
+  100% {
+    transform: rotate(359deg);
+  }
+}
+
+/* -------------------------------------------------------------------------
+   Small Spinner
+   ------------------------------------------------------------------------- */
+
+.small {
+  width: var(--space-base);
+  height: var(--space-base);
+  border-width: var(--space-smaller);
+}
+
+/* -------------------------------------------------------------------------
+   Inline Spinner
+   ------------------------------------------------------------------------- */
+
+.inline {
+  display: inline-block;
+}
--- a/packages/components/src/Spinner/Spinner.mdx
+++ b/packages/components/src/Spinner/Spinner.mdx
@@ -1,0 +1,53 @@
+---
+name: Spinner
+menu: Components
+route: /components/spinner
+---
+
+import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { Spinner } from ".";
+import { Text } from "../Text";
+
+# Spinner
+
+<ComponentStatus stage="pre" responsive="no" accessible="yes" />
+
+Spinners are used to visualize an undetermined load time for a section of a page
+and prevent double clicking of a UI element by overlaying it while its action is
+processing.
+
+The default spinner should be used to indicate loading of a view
+
+```ts
+import { Spinner } from "@jobber/components/Spinner";
+```
+
+<Playground>
+  <Spinner />
+</Playground>
+
+## Props
+
+<Props of={Spinner} />
+
+---
+
+## Small
+
+The `small` spinner should be used on individual element of an interface (ie.
+button or card contents) or when inline with content
+
+<Playground>
+  <Spinner size="small" />
+</Playground>
+
+## Inline
+
+Use `inline` to render the spinner inline with content.
+
+<Playground>
+  <Text>
+    Uploading... <Spinner size="small" inline={true} />
+  </Text>
+</Playground>
--- a/packages/components/src/Spinner/Spinner.tsx
+++ b/packages/components/src/Spinner/Spinner.tsx
@@ -1,0 +1,34 @@
+import React from "react";
+import classnames from "classnames";
+import styles from "./Spinner.css";
+
+interface SpinnerProps {
+  /**
+   * Specifies the size of the spinner
+   *
+   * @default "base"
+   */
+  readonly size?: "small" | "base";
+
+  /**
+   * Spinner becomes an inline element when true,
+   * otherwise it is a block element
+   */
+  readonly inline?: boolean;
+}
+
+export function Spinner({ size = "base", inline }: SpinnerProps) {
+  const spinnerStyles = classnames(styles.spinner, {
+    [styles.small]: size === "small",
+    [styles.inline]: inline,
+  });
+
+  return (
+    <div
+      className={spinnerStyles}
+      role="alert"
+      aria-busy={true}
+      aria-label="loading"
+    />
+  );
+}
--- a/packages/design/src/shadows.css
+++ b/packages/design/src/shadows.css
@@ -2,4 +2,5 @@
   --shadow-base: 0px 1px 4px 0px rgba(0, 0, 0, 0.0980392),
     0px 4px 12px 0px rgba(0, 0, 0, 0.0470588);
   --shadow-focus: 0px 0px 4px 2px var(--color-outline);
+  --shadow-dark: 0px 0px 4px 2px var(--color-blue--lightest);
 }
--- a/packages/components/src/Spinner/Spinner.css.d.ts
+++ b/packages/components/src/Spinner/Spinner.css.d.ts
@@ -1,0 +1,3 @@
+export const spinner: string;
+export const small: string;
+export const inline: string;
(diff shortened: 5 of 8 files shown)


packages/components/src/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,53 @@
...
+# Spinner
+
+<ComponentStatus stage="pre" responsive="no" accessible="yes" />
+
+Spinners are used to visualize an undetermined load time for a section of a page
+and prevent double clicking of a UI element by overlaying it while its action is
+processing.
+
+The default spinner should be used to indicate loading of a view
+
```
whereyoubinns, line 21: @chris-at-jobber I updated the copy based on your comments left on #177, let me know if you have more thoughts here

packages/components/src/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,53 @@
...
+</Playground>
+
+## Inline
+
+Use `inline` to render the spinner inline with content.
+
+<Playground>
+  <Text>
+    Uploading... <Spinner size="small" inline={true} />
+  </Text>
```
whereyoubinns, line 52: not totally sure how to make this look more inline, the spinner dips below the text here, help with the css would be appreciated!
image

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,34 @@
...
+    [styles.small]: size === "small",
+    [styles.inline]: inline,
+  });
+
+  return (
+    <div
+      className={spinnerStyles}
+      role="alert"
+      aria-busy={true}
+      aria-label="loading"
```
whereyoubinns, line 31: @chris-at-jobber again thoughts on accessibility? based on #177, and the current roles in the custom spinner in JO.
I tried to test this with the mac voiceover but boy oh boy

packages/components/src/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,53 @@
...
+---
+
+import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { Spinner } from ".";
+import { Text } from "../Text";
+
+# Spinner
+
+<ComponentStatus stage="pre" responsive="no" accessible="yes" />
```
whereyoubinns, line 14: potentially `stage="ready"' based on the outcome of this review 😆

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  100% {
+    transform: rotate(359deg);
+  }
+}
+
+/* -------------------------------------------------------------------------
+   Small Spinner
+   ------------------------------------------------------------------------- */
```
eddysims, line 30: I think that this component is small enough and the css is specific enough that we don't need the comments explaining them.

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,44 @@
+.spinner {
+  width: var(--space-larger);
+  height: var(--space-larger);
+  box-shadow: var(--shadow-dark);
+  border: solid var(--color-blue--light);
+  border-top-color: var(--color-white);
+  border-width: calc(var(--space-small) * 0.75);
+  border-radius: 50%;
```
eddysims, line 8: we have a `--radius-circle` css property that would work here.

packages/components/src/Spinner/Spinner.css
```diff
@@ -0,0 +1,44 @@
...
+  vertical-align: middle;
```
eddysims, line 9: will this attr be needed on all spinners? Or just the inline ones?

packages/components/src/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,53 @@
+---
+name: Spinner
+menu: Components
+route: /components/spinner
+---
+
+import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { Spinner } from ".";
+import { Text } from "../Text";
```
eddysims, line 10: you can import components in `mdx` files as `import { Text } from "@jobber/components/Text"`

packages/components/src/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+# Spinner
+
+<ComponentStatus stage="pre" responsive="no" accessible="yes" />
+
+Spinners are used to visualize an undetermined load time for a section of a page
+and prevent double clicking of a UI element by overlaying it while its action is
+processing.
```
eddysims, line 18: I don't feel that it is the responsibility for the spinner to prevent the double clicking. A Spinner is just the UI element.

packages/components/src/Spinner/Spinner.mdx
eddysims: I think this is fine. @chris-at-jobber thoughts?

eddysims (changes requested):

eddysims:
@whereyoubinns some comments here. I don't think that there is anything to big in them.

packages/components/src/Spinner/Spinner.tsx
chris-at-jobber: I've tested this with VoiceOver and it seems to work fairly well, one suggestion:

Currently it reads out "Empty alert: loading" because the screenreader expects content inside of an alert. Ideally it reads out "alert: loading" which I think we could achieve with something like the example [here](https://codepen.io/JasonAment/pen/abOJmdo) but, also, I think communicates the context decently as-is.

As a side note, if you need a hand testing with VoiceOver ping the accessibility guild channel! Lots of folks in there who can help you get started, it can be _a lot_ at first.

packages/design/src/shadows.css
```diff
@@ -2,4 +2,5 @@
   --shadow-base: 0px 1px 4px 0px rgba(0, 0, 0, 0.0980392),
     0px 4px 12px 0px rgba(0, 0, 0, 0.0470588);
   --shadow-focus: 0px 0px 4px 2px var(--color-outline);
+  --shadow-dark: 0px 0px 4px 2px var(--color-blue--lightest);
```
chris-at-jobber, line 5: @whereyoubinns 2 comments on this shadow:

- let's move this into the Spinner.css file for now. If we need to use it in other cases it can be brought up here, but for now I think the spinner is probably a special case
- we don't have a nice systemic way of doing this with CSS variables, but can we use the RGBA values from `--color-blue` to make this shadow ~40% opacity?  `rgba(1, 41, 57, 0.4)` 
- It will give us a similar visual appearance on a white background, but act as a more "shadowy" shadow if it's overlaying content

packages/components/src/Spinner/Spinner.mdx
chris-at-jobber: It does look a bit low there - at first poke in the inspector, using a `margin-top: -4px` (or equivalent with our spacing units) aligns it a bit more cleanly. @eddysims I know that's kind of gross CSS-wise but unless there's a cleaner way, I'm good with it:

image

This gives it a nice vertical alignment on all size variants of `Text`.

chris-at-jobber: Maybe a follow-up improvement we can make, but I think it would be nice to encapsulate the Text part of this into the inline version of Spinner so that teams don't need to figure out...

- order: (spinner) (text) VS (text) (spinner)
- the right versions (inline/small/base) of spinner

To get this approach

chris-at-jobber:
Just the one CSS change-required from me! Opportunity to possibly optimize the accessibility a bit but I think it's doing a decent job for screenreaders as-is.

packages/components/src/Spinner/Spinner.mdx
eddysims: a nicer css would be `margin-top: calc(var(--space-smaller * -1)`. Would still allow for the css property to be used

packages/components/src/Spinner/Spinner.mdx
chris-at-jobber: @eddysims this is based off of the SG-1 description, where by default the Spinner element overlays all content it wraps

chris-at-jobber: ^ but I think to that point, this version of Spinner does not appear to wrap children in the same way, so it is extra true that it can be removed from the description!

packages/components/src/Spinner/Spinner.mdx
```diff
@@ -0,0 +1,53 @@
...
+The default spinner should be used to indicate loading of a view
+
+```ts
+import { Spinner } from "@jobber/components/Spinner";
+```
+
+<Playground>
+  <Spinner />
+</Playground>
+
```
chris-at-jobber, line 29: I'd like to add this section, it's part of our new docs template but this particular component PR is based on some older work so it's not captured here:

## Design & Usage Guidelines

Use `Spinner` to indicate loading of content where the actual amount of loading time or progress is unknown.

### Related Components

If the amount of progress or loading time is known, use [ProgressBar](https://atlantis.getjobber.com/components/progress-bar) to give the user better context around the process.

packages/components/src/Spinner/Spinner.mdx
chris-at-jobber: @whereyoubinns Thanks! I'm going to bring this down into the Design & Usage Guidelines with some minor tweaks to better reflect this version of Spinner.

whereyoubinns:
@eddysims @chris-at-jobber thanks for the review :) I've updated based on your comments

eddysims:
@chris-at-jobber code looks good to me. If the docs/design look good to you im ✅

chris-at-jobber (approved):
Looks good! Thanks @whereyoubinns 🎉

eddysims (approved):

PR #457, opened 2021-01-25
feat(components): Add subtitle to Page component

ellaestigoy:
## Motivations
We need a subheader to supplement another heading 

### After
image

### Added
* Added subheader under the title

---

Random photo of Atlantis


packages/components/src/Page/Page.tsx
```diff
@@ -101,7 +108,16 @@ export function Page({
       <Content>
         <Content spacing="large">
           <div className={titleBarClasses} ref={titleBarRef}>
-            <Heading level={1}>{title}</Heading>
+            <div>
+              <Heading level={1}>{title}</Heading>
+              <div className={styles.subheader}>
```
chris-at-jobber, line 113: @ellaestigoy we should have a conditional check here so that we only render this div and its' children if there is content for the subheader

packages/components/src/Page/Page.tsx
```diff
@@ -23,6 +24,11 @@ export interface PageProps {
    */
   readonly title: string;
 
+  /**
+   * Subheader of the page.
+   */
+  readonly subheader: string;
```
chris-at-jobber, line 30: Can we rename this to subtitle? It's more specific

packages/components/src/Page/Page.css
```diff
@@ -21,6 +21,10 @@
   margin-bottom: var(--space-base);
 }
 
+.titleBar .subheader {
+  margin-bottom: var(--space-base);
```
chris-at-jobber, line 25: This is just a visual thing, but I don't think we need the extra spacing here

### With
image

### Without
image

chris-at-jobber: I think it feels better with a more even rhythm between

title
subtitle
intro
content

Rather than the larger gap in between them

packages/components/src/Page/Page.tsx
chris-at-jobber: If we remove the margin-bottom per my comment above, I think this wrapper div can be removed altogether and we can conditionally render the `<Text>` element itself

ellaestigoy: @chris-at-jobber with small screen, i think we need spacing in there especially if there's an action group

packages/components/src/Page/Page.css
```diff
@@ -65,6 +61,10 @@
   width: 100%;
 }
 
+.titleBar.small .subtitle {
```
chris-at-jobber, line 64: Works on small
image

but we also have a "smaller" that isn't classnamed because it's the "default":
image

I _think_ this should get it working for all sizes:

```
.titleBar .subtitle {
  margin-bottom: 0;
}

.titleBar.medium .subtitle,
.titleBar.large .subtitle {
  margin-bottom: large;
}
```

chris-at-jobber: (There's a larger discussion about how we've named the size breakdowns, but that's on me as a larger fix because I was the last one messing with these sizes 😅 )

packages/components/src/Page/Page.mdx
```diff
@@ -89,3 +89,17 @@ of footer, it should exist after Page.
     </Content>
   </Page>
 </Playground>
+
+---
+
+## Page with Subtitle
```
chris-at-jobber, line 95: Can we add a description of when/why you should use a subtitle?

chris-at-jobber:
@ellaestigoy few remaining little things! Looking good.

packages/components/src/Page/Page.mdx
```diff
@@ -15,8 +15,8 @@ import { Text } from "../Text";
 <ComponentStatus stage="pre" responsive="yes" accessible="yes" />
```
chris-at-jobber, line 15: While we're in here, let's update this stage to "ready"

packages/components/src/Page/Page.css
```diff
@@ -78,3 +78,16 @@
 .titleBar.small .actionButton + .actionButton {
   margin-left: var(--space-small);
 }
+
+.titleBar .subtitle {
```
chris-at-jobber, line 82: Right now this is giving us no margin at the smallest size, and too much space on the medium-up sizes:

I think part of the issue is the [parent `<Content>` element ](https://github.com/GetJobber/atlantis/pull/457/files#diff-0a99e41ee8928670c3b01e35510e5adbe2e64d0d51dab50d7eb059e602eccc20R109) being set to large.

If we remove the "large" on that element, I think we can actually get rid of all these checks and just go

```
.titleBar .subtitle {
  margin-bottom: var(--space-base);
}
```

And get pretty even spacing the whole way down (I just gathered screenshots and pasted them in Figma...):
image

chris-at-jobber:
@ellaestigoy one last change to simplify the spacing and we should be good

chris-at-jobber (approved):
🎉

PR #460, opened 2021-01-28
feat(components): spinner loading state on input text

linds14sr20det:
## Motivations

This feature adds the ability to incorporate a spinner into the text input field. This would be used when a component is making a call to validate the contents of the field or doing some other long running task with the contents of the field.

## Changes

The spinner component now is used in the form field generic component.
### Added

- a boolean `showSpinner` has been added to the form field component, however it will only work with text input components.
### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

The mdx has a section that includes the spinner. 
---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.test.tsx
+++ b/packages/components/src/FormField/FormField.test.tsx
@@ -371,9 +371,8 @@ test("it should render the spinner when loading is true", () => {
     <FormField placeholder="foo" type="text" loading={true} />,
   );
   const spinner = getByLabelText("loading");
-  const classNames = spinner.getAttribute("class");
 
-  expect(classNames).toContain("spinner small");
+  expect(spinner).toBeInstanceOf(HTMLElement);
 });
 
 it("it should set the autocomplete value with one-time-code", () => {
--- a/packages/components/src/FormField/FormField.test.tsx
+++ b/packages/components/src/FormField/FormField.test.tsx
@@ -366,6 +366,16 @@ test("it should set the inputMode when the keyboard prop is set", () => {
   expect(name).toContain(keyboardMode);
 });
 
+test("it should render the spinner when loading is true", () => {
+  const { getByLabelText } = render(
+    <FormField placeholder="foo" type="text" loading={true} />,
+  );
+  const spinner = getByLabelText("loading");
+  const classNames = spinner.getAttribute("class");
+
+  expect(classNames).toContain("spinner small");
+});
+
 it("it should set the autocomplete value with one-time-code", () => {
   const { getByLabelText } = render(
     <FormField placeholder="foo" autocomplete={"one-time-code"} />,
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -260,12 +260,12 @@ export function FormField({
       </label>
       {fieldElement()}
       {type === "text" && loading && (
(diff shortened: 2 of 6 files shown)


chris-at-jobber:
Following the [guiding principles](https://atlantis.getjobber.com/CONTRIBUTING.md#guiding-principles) for Atlantis, I'd like to use a semantic name for this property.

Can we use a more meaningful `loading` similar to how we handle a similar feedback state on [Button](https://atlantis.getjobber.com/components/button#loading)?

chris-at-jobber:
I'd also be interested to see if there's an approach that allows us to make this change on `InputText` rather than `FormField`, if `InputText` is the only case for using this. It's a bit confusing that components like Select and Input Number now offer a `showSpinner` prop that doesn't do anything. 

This is a technical question which I don't have the answer to, but would make it more understandable on the other components if we can have this property only be present for the component that it's needed for.

chris-at-jobber (dismissed):
Property name should be more meaning-driven than presentation-driven; my other comment about where this change is applied is more open to discussion!

linds14sr20det:
@chris-at-jobber Agreed on the rename from showSpinner to loading. That's done and pushed up.

As for your second comment I'd like to put this on inputText only, but that isn't possible due to the way inputText is built. It renders a type of form field, and form field would need to be refactored to yield child content. We would also have to consider how yielding child content would affect all other formfield components. Basically it'd be a much bigger re-write than I'm willing to take on right now.

linds14sr20det:
@seanhealy @chris-at-jobber this should be good for review pending the design proposal. Unsure why the autocomplete snapshot wanted to be updated.

chris-at-jobber:
@linds14sr20det that Autocomplete snapshot is something we're trying to debug, you can kick it out and tests should get back to passing.

@seanhealy I think I have enough context from previous discussions about this with @JeffMarvin and @jlelford ; I'm not going to be available for reviews tomorrow, so consider this design approval on my end.

packages/components/src/FormField/FormField.test.tsx
```diff
@@ -366,6 +366,16 @@ test("it should set the inputMode when the keyboard prop is set", () => {
...
 });
 
+test("it should render the spinner when loading is true", () => {
+  const { getByLabelText } = render(
+    <FormField placeholder="foo" type="text" loading={true} />,
+  );
+  const spinner = getByLabelText("loading");
+  const classNames = spinner.getAttribute("class");
+
+  expect(classNames).toContain("spinner small");
```
seanhealy, line 376: I think we can just check chat the result of `getByLabelText("loading")` is an instance of HTMLElement as the rest of this is checking concerns that are internal to the Spinner component.

packages/components/src/FormField/FormField.tsx
```diff
@@ -252,6 +259,11 @@ export function FormField({
         {placeholder || " "}
       </label>
       {fieldElement()}
+      {type === "text" && loading && (
+        <span className={styles.icon}>
```
seanhealy, line 263: Are you ok changing this `icon` classname to `postfix` or something since that's what it's more shifted into here?

seanhealy (changes requested):
Two little thoughts for you.

seanhealy (approved):

PR #466, opened 2021-02-04
fix(components): exports date formatting as a string

Afroholic78:
## Motivations
We needed to export formatDate as a string so the resulting date can be passed as an attribute to other components.
Allowing the returned formatted date to be a string offers more flexibility on the places it can be used.

## Changes
Exported the strFormateDate (the method used by FormatDate).

## Testing
I'm now able to import `strFormatDate` and use it.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormatDate/FormatDate.mdx
+++ b/packages/components/src/FormatDate/FormatDate.mdx
@@ -1,9 +1,12 @@
 ---
-
-name: FormatDate menu: Components route: /components/date-formatter ---import {
-Playground, Props } from "docz"; import { ComponentStatus } from "@jobber/docx";
-import { CivilDate } from "@std-proposal/temporal"; import { FormatDate,
-strFormatDate } from "."; import { Text } from "../Text";
+name: FormatDate
+menu: Components
+route: /components/date-formatter
+---import { Playground, Props } from "docz";
+import { ComponentStatus } from "@jobber/docx";
+import { CivilDate } from "@std-proposal/temporal";
+import { FormatDate, strFormatDate } from ".";
+import { Text } from "../Text";
 
 # FormatDate
 
--- a/packages/components/src/FormatDate/FormatDate.mdx
+++ b/packages/components/src/FormatDate/FormatDate.mdx
@@ -1,13 +1,14 @@
----
-name: FormatDate
-menu: Components
-route: /components/date-formatter
+## import {Text} from "../Text";
+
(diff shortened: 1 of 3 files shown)


Afroholic78:
Additionally to this PR, I wonder if we should be allowing a date formatting to passed in ?

Afroholic78:
@seanhealy we had discussed this change a while back, you mind giving it a review. Let me know if you have thoughts on https://github.com/GetJobber/atlantis/pull/466#issuecomment-773627650

eddysims:
Hey, @Afroholic78 the code for this looks good. Now that the function is available through `@jobber/components`, I think that we should add some documentation into the `FormatDate` mdx files about it.

Afroholic78:
@eddysims sounds good, 
I was wondering is there a reason we wouldn't want to use `moment` under the hood to allow the passing of specific date formats (instead of using the Date implementation)

Afroholic78:
@eddysims Made the changes requested

packages/components/src/FormatDate/FormatDate.mdx
```diff
@@ -1,13 +1,14 @@
----
-name: FormatDate
-menu: Components
-route: /components/date-formatter
+## import {Text} from "../Text";
+
+name: FormatDate menu: Components route: /components/date-formatter
+
```
eddysims, line 4: Something happened here with the `frontmatter`

image

packages/components/src/FormatDate/FormatDate.mdx
```diff
@@ -26,6 +27,16 @@ import { FormatDate } from "@jobber/components/FormatDate";
   <FormatDate date={new CivilDate(2020, 2, 26)} />
 </Playground>
 
+## Getting a string version
```
eddysims, line 30: Can we add a description here of why someone would need a string version?

eddysims (changes requested):

Afroholic78:
@eddysims made the changes required
What about `date-fns`? https://www.npmjs.com/package/date-fns
I can drop it for now if it's a bigger conversation than is warranted for this PR. 
But because we need a specific format the current formatting solution won't really hold up. 😬

eddysims (approved):

eddysims:
@Afroholic78 yeah I think let's take that discussion out of this PR as adding new libraries should involve some discussion and investigation.

PR #474, opened 2021-02-24
fix(components): Refix the vertical alignment firefix on Firefox for FormFields

TJKruitbosch:
## Motivations

In a previous PR (#113), the style `vertical-align: -moz-middle-with-baseline;` was added to the `FormField` base component in order to fix some very unaligned inline form fields on Firefox. Since that time, the alignment of these fields on Firefox has slipped downward, leading to some strangely flowing input fields:

Screen Shot 2021-02-24 at 9 42 39 AM

I've changed around the vertical align styles for inline `FormFields` to align with the baseline once more:

Screen Shot 2021-02-24 at 3 07 09 PM

It works with the other `FormField`-using Input components (`InputNumber`, `InputPassword`, `InputTime`, `Select`):

Screen Shot 2021-02-24 at 3 09 42 PM

~It even stays really close to existing behaviour for multiline (i.e. `<textarea>`) `InputText`:~
The outcome of mixing `inline` with `multiline` on `InputText` has changed, but the use case is not supported.

Screen Shot 2021-02-24 at 3 12 54 PM
Screen Shot 2021-02-26 at 3 04 00 PM

And best of all, one less vendor-specific CSS value.

Screen Shot 2021-02-24 at 3 29 01 PM

### Deprecated

- Mixing the `inline` prop with `multiline` in `InputField` no longer vertically centers the resulting `<textarea>`. This combination of properties is not supported and may be removed in a future release. 

### Fixed

- `FormField`-inheriting components (e.g. `InputText`) now properly aligned with the text baseline on Firefox

## Testing

Use the `inline={true}` prop on the elements `InputText` (with and without `multiline={true}`), `InputNumber`, `InputPassword`, `InputTime`, and `Select`. On Firefox, the text within the form field should match baselines with the text outside the form field. On other browsers, this was already the case prior to this change and it continues to be after.

Tested on Firefox 85, Chrome 88, and (simulated) IOS 13 Safari.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -99,7 +99,16 @@ select.formField {
 .inline {
   display: inline-block;
   width: auto;
-  vertical-align: -moz-middle-with-baseline;
+  vertical-align: baseline;
+}
+
+.inline > input,
+.inline > select {
+  display: inline-block;
+}
+
+.inline.inlineTextArea {
+  vertical-align: middle;
 }
 
 /* Sizes */
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -99,16 +99,11 @@ select.formField {
 .inline {
   display: inline-block;
   width: auto;
-  vertical-align: baseline;
 }
 
 .inline > input,
 .inline > select {
   display: inline-block;
-}
-
-.inline.inlineTextArea {
-  vertical-align: middle;
 }
 
 /* Sizes */
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -232,6 +232,7 @@ export function FormField({
   const wrapperClassNames = classnames(
     styles.wrapper,
     inline && styles.inline,
+    inline && type === "textarea" && styles.inlineTextArea,
     size && styles[size],
     align && styles[align],
     disabled && styles.disabled,
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -232,7 +232,6 @@ export function FormField({
   const wrapperClassNames = classnames(
     styles.wrapper,
     inline && styles.inline,
-    inline && type === "textarea" && styles.inlineTextArea,
     size && styles[size],
     align && styles[align],
     disabled && styles.disabled,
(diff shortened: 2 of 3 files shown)


packages/components/src/FormField/FormField.css
```diff
@@ -99,7 +99,16 @@ select.formField {
...
+  vertical-align: baseline;
+}
+
+.inline > input,
+.inline > select {
+  display: inline-block;
+}
+
+.inline.inlineTextArea {
+  vertical-align: middle;
```
seanhealy, line 111: I don't think we need to support inline TextAreas. We should perhaps in a followup officially remove support for them. I really can't think of a case where we would want this behaviour. 🤔

@chris-at-jobber Any thoughts on this part?

packages/components/src/FormField/FormField.tsx
```diff
@@ -232,6 +232,7 @@ export function FormField({
   const wrapperClassNames = classnames(
     styles.wrapper,
     inline && styles.inline,
+    inline && type === "textarea" && styles.inlineTextArea,
```
seanhealy, line 235: If we drop text area support this should no longer be needed.

seanhealy (dismissed):
Looks great! Thanks so much, just one thought about supporting text area and inline.

packages/components/src/FormField/FormField.css
```diff
@@ -99,7 +99,16 @@ select.formField {
 .inline {
   display: inline-block;
   width: auto;
-  vertical-align: -moz-middle-with-baseline;
```
eddysims, line 102: I took a quick look at Firefox and I was able to just remove this line altogether and got the same result. Just wondering if we need the `vertical-align: baseline` or if we can just remove the `vertical-align` altogether?

TJKruitbosch: `baseline` is the initial value for `vertical-align`, and it's not an inherited property, so that could be removed. I like explicitly setting it – even if it's the default – if it's important enough to the goal of the class (the next person to see this code can see that explicitly matching the baseline is the intent), but I suppose that does make for extra CSS bloat.

packages/components/src/FormField/FormField.css
chris-at-jobber: I'm good with dropping support for inline TextArea 👍

TJKruitbosch: Added a new commit that removed this extra styling

packages/components/src/FormField/FormField.tsx
TJKruitbosch: Removed

packages/components/src/FormField/FormField.css
TJKruitbosch: Removed the unnecessary `vertical-align` statement along with another cleanup

eddysims (approved):

PR #479, opened 2021-03-03
docs: Add some documentation about validations

eddysims:
## Motivations

We do not currently have very good documentation about `validations`. This PR Adds some documentation about validations

### Added

- Documentation about validation
- 
### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Form/Form.mdx
+++ b/packages/components/src/Form/Form.mdx
@@ -21,6 +21,9 @@ import { Button } from "@jobber/components/Button";
 The `Form` component is a wrapper component that handles the submission and
 validation of forms.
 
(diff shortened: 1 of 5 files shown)


gregb-jobber (commented):
This looks good. Just knowing that it's there and that we're using React Hook Forms is way better than wondering what the `validations` value does. The other thing that was confusing was that there's all these types for the `validations` prop, but you don't actually have to go get those types you just need an object or a function. So that might be something small you can add.

seanhealy (approved):

PR #489, opened 2021-03-19
feat(docz-tools): Using Page for Atlantis

eddysims:
## Motivations

It would be cool if Atlantis used more Atlantis to build Atlantis. This is a proof of concept for that. I have moved over `🔱 Atlantis` and `Patterns > FeatureSwitch`

**A few notes**

- `Page` requires a title, which means that we need some way to supply that. Currently, I am using the `name` from the `mdx` file. This also meant removing the first `h1` from the `mdx` file.

- The `Actions` component has been moved to page actions of `Page`. This means they will no longer stick to the top. Which I think is fine, but would love to get a designer's eye on. 

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,9 +1,7 @@
 ---
-name: Atlantis
+name: 🔱 Atlantis
 route: /
 ---
-
-# 🔱 Atlantis
 
 [![CircleCI](https://circleci.com/gh/GetJobber/atlantis/tree/master.svg?style=svg&circle-token=3f1b0343273ef589350516e23713e81c8c3ac094)](https://circleci.com/gh/GetJobber/atlantis/tree/master)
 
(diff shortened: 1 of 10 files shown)


chris-at-jobber:
Looking pretty solid off the bat! I think it's totally fine that the actions aren't sticky.

A few things I noticed...

### Github actions

Not sure what's going on with the icon/text alignment in this Menu
image
Would expect it to be more like this:
image
And I think we could use the Menu subheading like
**View on Github**
- Atlantis Repository
- {{ComponentName}}

### Page alignment

The alignment of text and sandboxes gets out of whack when the `Page` gets below ~1200px wide:
image

### Open question

Could we use the `intro` prop of the `Page` to present a larger introductory paragraph without too much re-work? Not a requirement, I just think it would be a nice flex 😂

eddysims:
Closing for now. If we want this lets add a backlog item @chris-at-jobber

PR #492, opened 2021-03-23
docs(components): Popover Component Proposal

rjdestigter:
## Motivations

https://jobber.atlassian.net/browse/JOB-29707

### Added

Component proposal

---

tongue pops are Allysa's secret

### DIFF
--- a/docs/proposals/Popover/README.md
+++ b/docs/proposals/Popover/README.md
@@ -39,6 +39,12 @@ option based on responsive information.
 const AttachPdfToQuote = (props { checked: boolean}) => {
 	const checkboxRef = useRef()
 	const dontTellMeAnyMore = () => { ... }
+    const [toggled, setToggled] = useState(true)
+
+    const onDismiss = () => {
+        setToggled(false)
+        dontTellMeAnymore()
+    }
 
 	return (
 		<>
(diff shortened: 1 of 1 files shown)


docs/proposals/Popover/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+
+## Props Table
+
+### Popover
+
+| name        | type                                                                           | required? | default   | description                                                                                                                                                                                      |
+| ----------- | ------------------------------------------------------------------------------ | --------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| attachTo    | `HTMLElement \| React.Ref<HTMLElement \| null> \| React.ReactNode `            | ✓         |           | Element the popover will attach to and point at.                                                                                                                                                 |
```
seanhealy, line 98: The ref is interesting here but I'm wondering if we need it out of the gate. 🤔

docs/proposals/Popover/README.md
```diff
@@ -0,0 +1,131 @@
...
+| activateOn  | `"click" \| "focus"`                                                           |           | `"click"` | When `attachTo` is React node it it will be cloned and the popover will be toggled _on click_ by default.                                                                                        |
+| dismissable | `boolean`                                                                      |           | `true`    | Whether or not the popover is dismissable.                                                                                                                                                       |
+| toggled     | `boolean`                                                                      |           |           | Control popover viability rather than letting the component take care of it.                                                                                                                     |
```
seanhealy, line 101: Think that we can simplify this a lot if we look at how modal works with `open` and `onRequestClose`. I'm thinking that especially with this being a pretty foundational component we can keep it pretty barebones then add other triggers into more specific evolutions of this.

docs/proposals/Popover/README.md
```diff
@@ -0,0 +1,131 @@
...
+| children    | `React.ReactNode \| (renderProps: { dismiss: () => void }) => React.ReactNode` | ✓         |           | Pop-over content. A render function may also be provided and will provide you with a callback to dismiss the popover and potentially other stateful information or access to popper related APIs |
+| arrow       | `PopperJS.Placement`                                                           |           | `"auto"`  |
```
seanhealy, line 103: Do we have a case yet where we need to specify the arrow direction? Wondering if we can just set this to auto for now and think about it more in the future if we have a case that really needs it.

docs/proposals/Popover/README.md
```diff
@@ -0,0 +1,131 @@
...
+| dismissable | `boolean`                                                                      |           | `true`    | Whether or not the popover is dismissable.                                                                                                                                                       |
+| toggled     | `boolean`                                                                      |           |           | Control popover viability rather than letting the component take care of it.                                                                                                                     |
+| children    | `React.ReactNode \| (renderProps: { dismiss: () => void }) => React.ReactNode` | ✓         |           | Pop-over content. A render function may also be provided and will provide you with a callback to dismiss the popover and potentially other stateful information or access to popper related APIs |
```
seanhealy, line 102: Simplifying the `open` might also let us keep the children simpler as well.

docs/proposals/Popover/README.md
```diff
@@ -0,0 +1,131 @@
...
+        <PopoverBase
+            attachTo={<button>No scrubs</button>}
+            activateOn={'focus'}
+        >
+            <marquee>Party like it's 1999</marquee>
+        <PopoverBase>
+    )
+}
+
+```
```
seanhealy, line 90: Really like how you've looked a couple different usage examples here!!! :D

docs/proposals/Popover/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+    onDismiss?: () => void;
+} & (
+    { attachTo: HTMLElement | React.Ref<HTMLElement | null>; } |
+    { attachTo: React.ReactNode; activateOn?: "click" | "focus" }
+)
+```
+
+> !Important: When providing a React node to be cloned for `props.attachTo` the
+> element must be able to provide a reference to the DOM.
```
seanhealy, line 126: 👌

seanhealy (changes requested):
Few thoughts here. Most of them are along the lines of if we need complexity in this component itself or if it should perhaps be part of a more specific evolution. Very happy to chat so we can keep it moving along. 😄

docs/proposals/Popover/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Props Table
+
+### Popover
+
+| name           | type                                            | required? | default   | description                                                                                               |
+| -------------- | ----------------------------------------------- | --------- | --------- | --------------------------------------------------------------------------------------------------------- |
+| attachTo       | `HTMLElement \| React.Ref<HTMLElement \| null>` | ✓         |           | Element the popover will attach to and point at.                                                          |
+| activateOn     | `"click" \| "focus"`                            |           | `"click"` | When `attachTo` is React node it it will be cloned and the popover will be toggled _on click_ by default. |
+| dismissable    | `boolean`                                       |           | `true`    | Whether or not the popover is dismissable.                                                                |
```
seanhealy, line 85: Were you still planning on taking these out? I had though after we talked we figured we could just keep the high level popover really simple and control its visibility with `open` rather than trying to manage internal state.

docs/proposals/Popover/README.md
```diff
@@ -0,0 +1,103 @@
...
+}
+```
+
+## Props Table
+
+### Popover
+
+| name           | type                                            | required? | default   | description                                                                                               |
+| -------------- | ----------------------------------------------- | --------- | --------- | --------------------------------------------------------------------------------------------------------- |
+| attachTo       | `HTMLElement \| React.Ref<HTMLElement \| null>` | ✓         |           | Element the popover will attach to and point at.                                                          |
```
seanhealy, line 83: I'm pondering the name of this prop a bit again. Part of me would like to keep it the same as menu's activator just for consistency's sake but I do get how it isn't quite the same… There is value It think it just using the same words though so people can better "assume" what props will be called. 🤔

seanhealy (changes requested):
Two quick thoughts back your way.

docs/proposals/Popover/README.md
rjdestigter: @seanhealy Yes, I forgot to adjust the props table. `activeOn` is out the door. I think we're keeping `dismissable` to control whether we want the dismiss X button or not?

docs/proposals/Popover/README.md
rjdestigter: I get what you are thinking. However I do think there are going to be use cases where "activator" just doesn't apply. E.g. the case we are building this for is one of them. The popover is merely pointing at the "checkbox" and directing the users attention to it. In that sense you could argue the checkbox is "activating" but that does feel weird to me.

chris-at-jobber:
Popover is being implemented in [this PR](https://github.com/GetJobber/atlantis/pull/494)!

PR #493, opened 2021-03-26
feat(components): add 'disabled' props to Tab component

kingstonfung:
## Motivations

- There is a need for disabling tabs switching, let's say when the user has engaged partly into a workflow that is wrapped with `<Tabs>`.
- Or when the user is awaiting for confirmation with contents from a specific `<Tab>`, we want to disallow tab switching until their committed action has reached a conclusion (such as transmitting / sending form information)

### Added

- `disabled` prop is added to `<Tab>` and `<InternalTab>`
- `.tab.disabled` css styling added, following similar approach with button disabled stylings

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

1. Simply load up this PR and head over to `/components/tabs`
1. Enter the disabled value (eg: `disabled={true}`)
1. Observe the disabled UI and blocked interactions

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Tabs/Tabs.css
+++ b/packages/components/src/Tabs/Tabs.css
@@ -61,6 +61,17 @@
   background-color: var(--color-yellow--lightest);
 }
 
+.tab.disabled {
+  border-color: var(--button--color-disabled);
+  background-color: var(--button--color-disabled);
+  cursor: not-allowed;
+}
+
+.tab.disabled:hover,
+.tab.disabled:focus {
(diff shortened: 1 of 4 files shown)


dhaberst:
Tested and code looks good to me!

kingstonfung:
This is no longer needed, plus we need to do deeper dives into more UX considerations and accessibility. So I'll be closing out this PR for now. Thanks!

PR #494, opened 2021-03-29
feat(components): Add new popover component

robbyjobber:
## Motivations
- We want to create an informational popover/tooltip component
- [jira](https://jobber.atlassian.net/secure/RapidBoard.jspa?rapidView=158&modal=detail&selectedIssue=JOB-29558)
- [figma](https://www.figma.com/file/i2CF0Qou8QIzhs8umPvnnB/Design-System-Contribution-%5BPopover%5D?node-id=30%3A581)
image

### Added

- new popover/tooltip component that we can attach to a DOM element

### Changed
- we allowed refs to be applied to Atlantis `Checkbox` in order for our tooltip to properly latch onto the checkbox element

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Checkbox/Checkbox.tsx
+++ b/packages/components/src/Checkbox/Checkbox.tsx
@@ -52,69 +52,64 @@ interface CheckboxProps {
   onChange?(newValue: boolean): void;
 }
 
-export const Checkbox = React.forwardRef<HTMLDivElement, CheckboxProps>(
-  (
-    {
-      checked,
-      defaultChecked,
-      disabled,
-      label,
-      name,
-      value,
-      indeterminate = false,
-      description,
-      onChange,
-    },
-    ref,
-  ) => {
-    const wrapperClassName = classnames(
-      styles.wrapper,
-      disabled && styles.disabled,
-    );
-    const inputClassName = classnames(styles.input, {
-      [styles.indeterminate]: indeterminate,
-    });
-    const iconName = indeterminate ? "minus2" : "checkmark";
+export function Checkbox({
+  checked,
+  defaultChecked,
+  disabled,
+  label,
+  name,
+  value,
+  indeterminate = false,
+  description,
+  onChange,
+}: CheckboxProps) {
+  const wrapperClassName = classnames(
+    styles.wrapper,
+    disabled && styles.disabled,
+  );
+  const inputClassName = classnames(styles.input, {
+    [styles.indeterminate]: indeterminate,
+  });
+  const iconName = indeterminate ? "minus2" : "checkmark";
 
-    return (
-      <div className={styles.checkbox} ref={ref}>
-        <label className={wrapperClassName}>
-          <span className={styles.checkHolder}>
-            <input
-              type="checkbox"
-              checked={checked}
-              defaultChecked={defaultChecked}
-              className={inputClassName}
-              aria-label={label}
-              onChange={handleChange}
-              value={value}
-              name={name}
-              disabled={disabled}
-            />
-            <span className={styles.checkBox}>
-              <Icon name={iconName} size="small" color="white" />
-            </span>
+  return (
+    <div className={styles.checkbox}>
+      <label className={wrapperClassName}>
+        <span className={styles.checkHolder}>
+          <input
+            type="checkbox"
+            checked={checked}
+            defaultChecked={defaultChecked}
+            className={inputClassName}
+            aria-label={label}
+            onChange={handleChange}
+            value={value}
+            name={name}
+            disabled={disabled}
+          />
+          <span className={styles.checkBox}>
+            <Icon name={iconName} size="small" color="white" />
           </span>
+        </span>
 
-          {label != undefined && (
-            <span className={styles.label}>
-              <Text>{label}</Text>
-            </span>
-          )}
-        </label>
-        {description && (
-          <div className={styles.description}>
-            <Text variation="subdued" size="small">
-              {description}
-            </Text>
-          </div>
+        {label != undefined && (
+          <span className={styles.label}>
+            <Text>{label}</Text>
+          </span>
         )}
-      </div>
-    );
+      </label>
+      {description && (
+        <div className={styles.description}>
+          <Text variation="subdued" size="small">
+            {description}
+          </Text>
+        </div>
+      )}
+    </div>
+  );
 
-    function handleChange(event: ChangeEvent<HTMLInputElement>) {
-      const newChecked = event.currentTarget.checked;
-      onChange && onChange(newChecked);
-    }
-  },
-);
+  function handleChange(event: ChangeEvent<HTMLInputElement>) {
+    const newChecked = event.currentTarget.checked;
+    onChange && onChange(newChecked);
+  }
+}
--- a/packages/components/src/Checkbox/Checkbox.tsx
+++ b/packages/components/src/Checkbox/Checkbox.tsx
@@ -52,64 +52,69 @@ interface CheckboxProps {
   onChange?(newValue: boolean): void;
 }
 
-export function Checkbox({
-  checked,
-  defaultChecked,
-  disabled,
-  label,
-  name,
-  value,
-  indeterminate = false,
-  description,
-  onChange,
-}: CheckboxProps) {
-  const wrapperClassName = classnames(
-    styles.wrapper,
-    disabled && styles.disabled,
-  );
-  const inputClassName = classnames(styles.input, {
-    [styles.indeterminate]: indeterminate,
-  });
-  const iconName = indeterminate ? "minus2" : "checkmark";
+export const Checkbox = React.forwardRef<HTMLDivElement, CheckboxProps>(
+  (
+    {
+      checked,
+      defaultChecked,
+      disabled,
+      label,
+      name,
+      value,
+      indeterminate = false,
+      description,
+      onChange,
+    },
+    ref,
+  ) => {
+    const wrapperClassName = classnames(
+      styles.wrapper,
+      disabled && styles.disabled,
+    );
+    const inputClassName = classnames(styles.input, {
+      [styles.indeterminate]: indeterminate,
+    });
+    const iconName = indeterminate ? "minus2" : "checkmark";
 
-  return (
-    <div className={styles.checkbox}>
-      <label className={wrapperClassName}>
-        <span className={styles.checkHolder}>
-          <input
-            type="checkbox"
-            checked={checked}
-            defaultChecked={defaultChecked}
-            className={inputClassName}
-            aria-label={label}
-            onChange={handleChange}
-            value={value}
-            name={name}
-            disabled={disabled}
-          />
-          <span className={styles.checkBox}>
-            <Icon name={iconName} size="small" color="white" />
+    return (
+      <div className={styles.checkbox} ref={ref}>
+        <label className={wrapperClassName}>
+          <span className={styles.checkHolder}>
+            <input
+              type="checkbox"
+              checked={checked}
+              defaultChecked={defaultChecked}
+              className={inputClassName}
+              aria-label={label}
+              onChange={handleChange}
+              value={value}
+              name={name}
+              disabled={disabled}
+            />
+            <span className={styles.checkBox}>
+              <Icon name={iconName} size="small" color="white" />
+            </span>
           </span>
-        </span>
 
-        {label != undefined && (
-          <span className={styles.label}>
-            <Text>{label}</Text>
-          </span>
+          {label != undefined && (
+            <span className={styles.label}>
+              <Text>{label}</Text>
+            </span>
+          )}
+        </label>
+        {description && (
+          <div className={styles.description}>
+            <Text variation="subdued" size="small">
+              {description}
+            </Text>
+          </div>
         )}
-      </label>
-      {description && (
-        <div className={styles.description}>
-          <Text variation="subdued" size="small">
-            {description}
-          </Text>
-        </div>
-      )}
-    </div>
-  );
+      </div>
+    );
 
-  function handleChange(event: ChangeEvent<HTMLInputElement>) {
-    const newChecked = event.currentTarget.checked;
-    onChange && onChange(newChecked);
-  }
-}
+    function handleChange(event: ChangeEvent<HTMLInputElement>) {
+      const newChecked = event.currentTarget.checked;
+      onChange && onChange(newChecked);
+    }
+  },
+);
--- a/packages/components/src/Popover/Popover.css
+++ b/packages/components/src/Popover/Popover.css
@@ -2,7 +2,7 @@
   --card--background-color: var(--color-white);
   --card--base-padding: var(--space-base);
   --popover--padding: var(--space-base);
-  --popover--width: calc(var(--base-unit) * 37.5);
+  --popover--width: calc(var(--base-unit) * 21.875);
   --dismiss--button-padding: calc(var(--base-unit) / 4);
 }
 
@@ -23,25 +23,13 @@
 
 .popover {
   z-index: 1;
-  width: 350px;
+  width: 100%;
   max-width: var(--popover--width);
-  border: 1px solid rgb(203, 203, 203);
-  border-radius: 2px;
-  color: rgb(51, 51, 51);
-  font-size: 13px;
-  font-weight: bold;
-  line-height: 1.5em;
-  text-align: left;
-  background: rgb(250, 250, 250);
-  filter: drop-shadow(0px 0px 6px rgba(181, 181, 181, 0.4));
-}
-.popover > .background {
-  height: 100px;
-  background-image: url("https://images.pexels.com/photos/1563356/pexels-photo-1563356.jpeg?auto=compress&cs=tinysrgb&dpr=1&w=500");
-}
-.popover > p {
-  margin: 0;
-  padding: 8px;
+  border: 1px solid var(--color-grey--light);
+  border-radius: var(--radius-base);
+  font-size: var(--base-unit);
+  line-height: normal;
+  background: var(--color-white);
 }
 
 .popover > * {
@@ -51,8 +39,8 @@
 .arrow,
 .arrow::before {
   position: absolute;
-  width: 15px;
-  height: 15px;
+  width: var(--base-unit);
+  height: var(--base-unit);
   background: inherit;
 }
 
@@ -63,7 +51,7 @@
 .arrow::before {
   content: "";
   visibility: visible;
-  border: 1px solid rgb(203, 203, 203);
+  border: 1px solid var(--color-grey--light);
   transform: rotate(45deg);
   transform-origin: center;
   clip-path: polygon(-4px -4px, 15px 0, 0 15px);
--- a/packages/components/src/Popover/Popover.css
+++ b/packages/components/src/Popover/Popover.css
@@ -15,16 +15,16 @@
 }
 
 .dismissButton {
-  position: absolute;
-  top: 0;
-  right: 0;
+  float: right;
   padding: var(--dismiss--button-padding);
+  text-align: right;
 }
 
 .popover {
-  z-index: 1;
+  z-index: var(--elevation-base);
   width: 100%;
   max-width: var(--popover--width);
+  box-shadow: var(--shadow-base);
   border: var(--border-base) solid var(--color-grey--light);
   border-radius: var(--radius-base);
   font-size: var(--base-unit);
--- a/packages/components/src/Popover/Popover.css
+++ b/packages/components/src/Popover/Popover.css
@@ -1,17 +1,28 @@
 :root {
   --card--background-color: var(--color-white);
   --card--base-padding: var(--space-base);
+  --popover--padding: var(--space-base);
   --popover--width: calc(var(--base-unit) * 37.5);
+  --dismiss--button-padding: calc(var(--base-unit) / 4);
 }
 
 .header {
   display: flex;
+  position: "relative";
+  padding: var(--popover--padding);
   justify-content: space-between;
   align-items: center;
-  padding: var(--modal--padding);
+}
+
+.dismissButton {
+  position: absolute;
+  top: 0;
+  right: 0;
+  padding: var(--dismiss--button-padding);
 }
 
 .popover {
+  z-index: 1;
   width: 350px;
   max-width: var(--popover--width);
   border: 1px solid rgb(203, 203, 203);
@@ -23,7 +34,6 @@
   text-align: left;
   background: rgb(250, 250, 250);
   filter: drop-shadow(0px 0px 6px rgba(181, 181, 181, 0.4));
-  z-index:1;
 }
 .popover > .background {
   height: 100px;
--- a/packages/components/src/Popover/Popover.css
+++ b/packages/components/src/Popover/Popover.css
@@ -25,7 +25,7 @@
   z-index: 1;
   width: 100%;
   max-width: var(--popover--width);
-  border: 1px solid var(--color-grey--light);
+  border: var(--border-base) solid var(--color-grey--light);
   border-radius: var(--radius-base);
   font-size: var(--base-unit);
   line-height: normal;
@@ -47,7 +47,7 @@
 .arrow::before {
   content: "";
   visibility: visible;
-  border: 1px solid var(--color-grey--light);
+  border: var(--border-base) solid var(--color-grey--light);
   transform: rotate(45deg);
   transform-origin: center;
   clip-path: polygon(-4px -4px, 15px 0, 0 15px);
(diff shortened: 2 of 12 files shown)


packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -52,64 +52,69 @@ interface CheckboxProps {
...
+      defaultChecked,
+      disabled,
+      label,
+      name,
+      value,
+      indeterminate = false,
+      description,
+      onChange,
+    },
+    ref,
```
robbyjobber, line 68: We are using `React.forwardRef` here in order to allow our checkbox to be used as a reference for our popover component

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -52,64 +52,69 @@ interface CheckboxProps {
...
-            aria-label={label}
-            onChange={handleChange}
-            value={value}
-            name={name}
-            disabled={disabled}
-          />
-          <span className={styles.checkBox}>
-            <Icon name={iconName} size="small" color="white" />
+    return (
+      <div className={styles.checkbox} ref={ref}>
```
robbyjobber, line 80: The only other major change to the component is `ref={ref}`.  the rest are just spacing/tabbing.

packages/components/src/Popover/Popover.css
```diff
@@ -0,0 +1,95 @@
...
+  position: absolute;
+  top: 0;
+  right: 0;
+  padding: var(--dismiss--button-padding);
+}
+
+.popover {
+  z-index: 1;
+  width: 350px;
+  max-width: var(--popover--width);
```
rjdestigter, line 27: What is the max width for?

packages/components/src/Popover/Popover.css
```diff
@@ -0,0 +1,95 @@
...
+  border: 1px solid rgb(203, 203, 203);
```
rjdestigter, line 28: Should border color / size use the same variables as card if available?

packages/components/src/Popover/Popover.css
```diff
@@ -0,0 +1,95 @@
...
+  border-radius: 2px;
+  color: rgb(51, 51, 51);
+  font-size: 13px;
+  font-weight: bold;
+  line-height: 1.5em;
+  text-align: left;
+  background: rgb(250, 250, 250);
+  filter: drop-shadow(0px 0px 6px rgba(181, 181, 181, 0.4));
+}
+.popover > .background {
```
rjdestigter, line 38: Copy Pasta No Need

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+        onRequestClose={handleClose}
+      >
+        {content}
+      </PopoverTestComponent>,
+    );
+  });
+
+  const { getByText, queryByTestId, container } = rendered;
+
+  expect(getByText(title)).toBeTruthy();
```
rjdestigter, line 49: I think, rather than using `getBy..` we can use `queryBy..` since that will return `null` if not present and subsequently not be "truthy". (Rather than `getBy..` which will throw an error)

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+      </PopoverTestComponent>,
+    );
+  });
+
+  const { queryByTestId } = rendered;
+
+  expect(queryByTestId("popover")).toBeNull();
+});
+
+it("shouldn't render a popover if open is false", async () => {
```
rjdestigter, line 90: Is this a duplicate test?

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+      title="Test Popover"
+      open={false}
+    >
+      Test Content
+    </PopoverTestComponent>,
+  );
+
+  const popover = container.querySelector("popover");
+
+  await waitFor(() => {
```
rjdestigter, line 103: `querySelector` returns immediately so the usage of `waitFor` here is not necessary. (The value of `popover` will never change)

packages/components/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Callback executed when the user wants to close/dismiss the popover
+   */
+  readonly onRequestClose?: () => void;
+
+  /**
+   * Describes the preferred placement of the popper.  The first element in the placement array will be taken
+   * as the preferred location with the rest of the array as fallback.
+   */
+  readonly preferredPlacement?: "top" | "bottom" | "left" | "right" | "auto";
```
rjdestigter, line 39: Since `auto` is the permanent fallback, should it still be a prop option?

packages/components/src/Popover/Popover.test.tsx
robbyjobber: yes 😞 . Removing it!

packages/components/src/Popover/Popover.tsx
robbyjobber: wouldn't typescript complain if we didn't have `auto` as a type when we use it as a fallback?

packages/components/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,127 @@
...
+  preferredPlacement = "auto",
+  title,
+}: PopoverProps) {
+  const [popperElement, setPopperElement] = useState<HTMLElement | null>(null);
+  const [arrowElement, setArrowElement] = useState<HTMLElement | null>(null);
+  const { styles: popperStyles, attributes } = usePopper(
+    attachTo instanceof Element ? attachTo : attachTo.current,
+    popperElement,
+    {
+      modifiers: buildModifiers(arrowElement),
```
rjdestigter, line 62: The `usePopper` internals does do some recomputation (https://github.com/popperjs/react-popper/blob/master/src/usePopper.js#L113) on each render if the list of modifiers has changed. (Which in our case will be so for each render). I am not really concerned about any performance issues but this is something to keep in mind if we do run into that in the future.

packages/components/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,127 @@
...
+}
+
+export function Popover({
+  onRequestClose,
+  children,
+  dismissible = true,
+  attachTo,
+  open,
+  preferredPlacement = "auto",
+  title,
```
rjdestigter, line 54: I'm not sure if _Popover_ should concern itself with _title_. If a popover needs a title then that can just be part of it's children, no?

packages/components/src/Popover/Popover.css
robbyjobber: updated it to total `350px`

```
  --popover--width: calc(var(--base-unit) * 21.875);
```

packages/components/src/Popover/Popover.tsx
robbyjobber: yep makes sense.  As discussed we'll let consumers handle creating the title in the content

packages/components/src/Popover/Popover.css
```diff
@@ -0,0 +1,83 @@
...
+  top: 0;
+  right: 0;
+  padding: var(--dismiss--button-padding);
+}
+
+.popover {
+  z-index: 1;
+  width: 100%;
+  max-width: var(--popover--width);
+  border: 1px solid var(--color-grey--light);
```
seanhealy, line 28: We have some border custom properties here we can use. Need to make them more obvious through docs.

https://github.com/GetJobber/atlantis/blob/master/packages/design/src/borders.css

packages/components/src/Popover/Popover.css
```diff
@@ -0,0 +1,83 @@
...
+  content: "";
+  visibility: visible;
+  border: 1px solid var(--color-grey--light);
+  transform: rotate(45deg);
+  transform-origin: center;
+  clip-path: polygon(-4px -4px, 15px 0, 0 15px);
+}
+
+.popover[data-popper-placement^="top"] > .arrow {
+  bottom: -5px;
```
seanhealy, line 61: Sort of pondering if we should make these sizes `calc`ed of of our sizing ones. Not sure if it would be worth it but it would be really nice to have them all tied together. 🤔

packages/components/src/Popover/Popover.mdx
```diff
@@ -0,0 +1,212 @@
...
+
+<ComponentStatus stage="pre" responsive="no" accessible="no" />
+
+Write a description about what Popover component is trying to achieve.
+
+```ts
+import { Popover } from "@jobber/components/Popover";
+```
+
+<Playground>
```
seanhealy, line 27: I'm sort of thinking that it might be nicer to keep the first example here a bit more simple. Perhaps just something as simple as a button that shows a really simple popover?

Then we can get into more complex examples of open by default, etc lower down if we think we need to. 😄

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { Popover } from ".";
+import { PopoverProps } from "./Popover";
+
+afterEach(cleanup);
+
+let rendered: RenderResult;
+const content = "Test Content";
+
+const PopoverTestComponent = (props: PopoverProps) => {
+  const checkboxRef = useRef();
```
seanhealy, line 18: Perhaps?

```suggestion
  const divRef = useRef();
```

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  );
+};
+
+it("should render a Popover with the content and dismiss button", async () => {
+  const handleClose = jest.fn();
+
+  await act(async () => {
+    rendered = render(
+      <PopoverTestComponent
+        attachTo={undefined}
```
seanhealy, line 35: What does it do if you pass `undefined` here? Is this something it should support?

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+        onRequestClose={handleClose}
+      >
+        {content}
+      </PopoverTestComponent>,
+    );
+  });
+
+  const { queryByText, queryByTestId, container } = rendered;
+
+  expect(queryByText(content)).toBeTruthy();
```
seanhealy, line 46: Might be nicer to test that we're actually getting an HTML element here.

```suggestion
  expect(queryByText(content)).toBeInstanceOf(HTMLElement);
```

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  expect(queryByTestId("popover")).not.toBeNull();
+
+  fireEvent.click(container.querySelector("[aria-label='Close modal']"));
```
seanhealy, line 49: Would be nicer to query for the text `Close Modal` rather than doing a css selector here.

packages/components/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Whether or not the popover is dismissable via the header close button.
+   */
+  readonly dismissible?: boolean;
+
+  /**
+   * Control popover viability.
+   */
+  readonly open?: boolean;
```
seanhealy, line 27: I think `open` might be required. If not how do we handle it being unspecified?

packages/components/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,109 @@
+/* eslint-disable no-null/no-null */
```
seanhealy, line 1: Why is this off? I don't think we should have to disable it.

packages/components/src/Popover/Popover.tsx
seanhealy: It is nice if the default is also available as an option so I think it's good to include here. 

Would also be nice to tag the default here. Eg.
https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Avatar/Avatar.tsx#L27-L29

packages/components/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,109 @@
...
+  );
+
+  const popperClassName = classnames(classes.popover);
+  const arrowClassName = classnames(classes.arrow);
+  const dismissButton = classnames(classes.dismissButton);
+
+  return (
+    open && (
+      <div
+        data-testid="popover"
```
seanhealy, line 67: It would be nice to have or tests check for content or something else that is visible to the user. If we do this then we shouldn't need a `data-testid` here.

Point 2 here: https://testing-library.com/docs/guiding-principles 😄

packages/components/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const { styles: popperStyles, attributes } = usePopper(
+    attachTo instanceof Element ? attachTo : attachTo.current,
+    popperElement,
+    {
+      modifiers: buildModifiers(arrowElement),
+      placement: preferredPlacement,
+    },
+  );
+
+  const popperClassName = classnames(classes.popover);
```
seanhealy, line 60: If were' not doing any conditions then we might not need `classnames` at all. All it is doing here is passing the string through with no changes.

seanhealy (changes requested):
Looking great! Have a few comments for you. Happy to chat about any of them. 😁

packages/components/src/Popover/Popover.test.tsx
robbyjobber: I was having some difficulty creating my HOC `PopoverTestComponent`.  Hooks can only be created in the component and I built the HOC to accept PopoverProps which unfortunately requires `attachTo`. So I just pass undefined there. 

```
const PopoverTestComponent = (props: PopoverProps) => {
  const checkboxRef = useRef();
  return (
    <>
      <div ref={checkboxRef}></div>
      <Popover {...props} attachTo={checkboxRef}>
        {props.children}
      </Popover>
    </>
  );
};
```

Would there be another way?

packages/components/src/Popover/Popover.css
robbyjobber: Yea I was thinking about that too.  I don't think it's worth it right now since the arrow is so specific to popover.  Maybe if other components will use it in the future? Though I find that unlikely.

packages/components/src/Popover/Popover.mdx
robbyjobber: sounds good! 👍

packages/components/src/Popover/Popover.test.tsx
robbyjobber: But it's a dismiss button with an `X` icon so can't query by text unfortunately.

packages/components/src/Popover/Popover.tsx
robbyjobber: this lint asks us to use undefined instead of null.  We get this strange issue when trying to use undefined with ref

image

Essentially we're fighting with React's `ref type`.

```
type Ref<T> = { bivarianceHack(instance: T | null): void }["bivarianceHack"] | RefObject<T> | null;
```

the `bivarianceHack(instance: T | null): void }["bivarianceHack"]` is type we're trying to satisfy by passing `setPopperElement` from the `const [popperElement, setPopperElement] = useState<HTMLElement | null>(null);` declaration

If we use undefined, the react typing would complain so hence why we need this disable lint sadly 😢

packages/components/src/Popover/Popover.test.tsx
seanhealy: If you query by label it should find the aria label as well.

packages/components/src/Popover/Popover.css
```diff
@@ -0,0 +1,83 @@
...
+  top: 0;
+  right: 0;
+  padding: var(--dismiss--button-padding);
+}
+
+.popover {
+  z-index: 1;
+  width: 100%;
+  max-width: var(--popover--width);
+  border: var(--border-base) solid var(--color-grey--light);
```
chris-at-jobber, line 28: If this PR lands first, reminder for me to come in and update this border-color to match what I'm defining in[ this PR](https://github.com/GetJobber/atlantis/pull/500)

seanhealy: It's a race! 🏁

packages/components/src/Popover/Popover.css
```diff
@@ -0,0 +1,83 @@
...
+
+.dismissButton {
+  position: absolute;
+  top: 0;
+  right: 0;
+  padding: var(--dismiss--button-padding);
+}
+
+.popover {
+  z-index: 1;
```
chris-at-jobber, line 25: Can we use one of our elevation variables here? This feels like it should have the same elevation as `menu`:

```
  --elevation-menu: 6;
```

Although you could also use `elevation-base` as `1` seems to be working here for you.

packages/components/src/Popover/Popover.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+### Usage
+
+- A popover should be used to introduce a novel UI element or highlight a change
+  to an existing interface.
+- A popover should contain information that is helpful, but not necessary for
+  the user to successfully interact with the page.
+- Once a popover has been shown, it should not show up again once the user has
+  interacted with it.
+- **You must use a `useRef` hook to create a reference point for the popover to
```
chris-at-jobber, line 71: I would consider this a technical note - it's getting buried here under the design guidelines. Can we explain this on the `attachTo` prop itself, where it's more relevant?

packages/components/src/Popover/Popover.mdx
```diff
@@ -0,0 +1,195 @@
...
+  latch to.**
+
+### Content
+
+- Popover text content should not exceed 3 lines and can consist of an optional
```
chris-at-jobber, line 76: Dismiss should not be optional

packages/components/src/Popover/Popover.mdx
```diff
@@ -0,0 +1,195 @@
...
+          preferredPlacement="bottom"
+        >
+          <Content>Bottom Popover</Content>
+        </Popover>
+      </div>
+    );
+  }}
+</Playground>
+
+## Dismissible is false
```
chris-at-jobber, line 169: Dismissible should never be false:

> should contain information that is helpful, but not necessary for the user to successfully interact with the page.

The user should always have a means to remove a Popover, as it actively blocks some portion of the interface and isn't required to use the page.

chris-at-jobber (changes requested):
@robbyjobber a few notes on the code review, and some design changes (cc @chriscarline ) :

image

- Content should not bleed under the dismiss button
- Popover should have a shadow, per the Figma; let's use `var(--shadow-base)` for this case
- The user should always be able to dismiss a Popover

packages/components/src/Popover/Popover.mdx
```diff
@@ -0,0 +1,195 @@
...
+-->
+
+## Design & Usage Guidelines
+
+When contributing to, or consuming the Popover component, consider the
+following:
+
+### Usage
+
+- A popover should be used to introduce a novel UI element or highlight a change
```
chris-at-jobber, line 65: I realized this is just the content for tooltip moved over - can we reflect the guidelines from the Figma?

> The popover component displays floating informative and actionable content in relation to a target. The popover can contain content, media, or other components. 
> 
> Some scenarios for popover include the following:
> 
> · Show available related actions
> · First time experience
> · To call out product features or changes

packages/components/src/Popover/Popover.mdx
robbyjobber: that's a fair point.  I'll remove the optional `dismissible` prop

packages/components/src/Popover/Popover.mdx
robbyjobber: sounds good!

chris-at-jobber:
> @chris-at-jobber
> 
> > * Content should not bleed under the dismiss button
> 
> I'm not sure how to go about this. If I don't allow bleeding then we won't be able to achieve this design
> image
> 
> on the other hand I find it odd also that the consumers have to make sure that their content doesn't go under the dismiss button.
> 
> Any recommendations on approach?

I think a future iteration of Popover can have a `media` slot (or something similar), and if `media` is defined we can conditionally apply a CSS class on the dismiss button to allow the button to float overtop like that.

rjdestigter:
CSS float has the potential to solve this problem https://codepen.io/chautelly/pen/JjErJxd

chris-at-jobber:
@rjdestigter It's been so long, I forgot about float! 😂  Great call

robbyjobber:
Same! Pushing my changes up. Thanks everyone!

chris-at-jobber (dismissed):
Shadow and button/content overlap update look good!

image

Let's get rid of this section now that it's not needed:
image

I did a quick pass with accessibility in mind. Looking at the specifications from the [Figma](https://www.figma.com/file/i2CF0Qou8QIzhs8umPvnnB/Design-System-Contribution-Popover?node-id=30%3A586), can we make sure that...

- Focus is set on the Popover when it is opened
- Focus is returned to the attached element when the Popover is closed

As well, can we make sure that there's a `described-by` relationship created, as well as a `has-popup` on the attached element?

We should also provide a role to this element so assistive technology can provide users with some context.

I believe [`dialog` is an appropriate role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) here, and looking at some other libraries that's also how they're handling this. Can we update the role, and also the aria-label prop on `ButtonDismiss` to "Close dialog"?

robbyjobber:
> Shadow and button/content overlap update look good!
> 
> image
> 
> Let's get rid of this section now that it's not needed:
> image
> 
> I did a quick pass with accessibility in mind. Looking at the specifications from the [Figma](https://www.figma.com/file/i2CF0Qou8QIzhs8umPvnnB/Design-System-Contribution-Popover?node-id=30%3A586), can we make sure that...
> 
> * Focus is set on the Popover when it is opened
> * Focus is returned to the attached element when the Popover is closed

@chris-at-jobber and @seanhealy I think we'll have to get in a call to clarify on the approach.  I have a couple ideas how we can make this work.

> As well, can we make sure that there's a `described-by` relationship created, as well as a `has-popup` on the attached element?
I think the describe-by relationship will need to be applied to the popover `children` content.  This part is dynamic to the content being passed to popover so there isn't a way to apply this descriptor sensibly without using the `children` content.

The same for the `has-popup`.  The consumer will have to ensure that when they create the HTML Element they want to attach the popover to.  

That is if you mean we should have these aria-descriptors in the mdx playground examples 🤔 

> We should also provide a role to this element so assistive technology can provide users with some context.
> 
> I believe [`dialog` is an appropriate role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role) here, and looking at some other libraries that's also how they're handling this. Can we update the role, and also the aria-label prop on `ButtonDismiss` to "Close dialog"?

👍

rjdestigter:
Yeah, I thnk we're mixing things up here @chris-at-jobber. The popover is not always a popup or dialog. In our case, when using the popover to identify a changed or new feature, the popover is just a regular part of the the DOM hierarchy and rendered right after the element it is pointing to. It does not appear after some user interaction.

chris-at-jobber:
Had a discussion around the accessibility considerations and I think this is good to go for now from a design/a11y perspective. 

@seanhealy , back to you if you'd like to take one more pass on the code!

packages/components/src/Popover/Popover.mdx
```diff
@@ -0,0 +1,172 @@
...
+    const leftRef = useRef();
+    const rightRef = useRef();
+    const bottomRef = useRef();
+    const [showPopover, setShowPopover] = useState(true);
+    const styles = {
+      display: "flex",
+      flexDirection: "column",
+      alignItems: "flex-start",
+      width: "200px",
+    };
```
seanhealy, line 122: What do we need `styles` for here? Would be nice if we can avoid it so we're aren't providing this as an example to people on how to style components. :) 

Perhaps we could use full width buttons and wrap them all in `<Content>`?

packages/components/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,97 @@
...
+  readonly open: boolean;
+
+  /**
+   * Callback executed when the user wants to close/dismiss the popover
+   */
+  readonly onRequestClose?: () => void;
+
+  /**
+   * Describes the preferred placement of the popper.  The first element in the placement array will be taken
+   * as the preferred location with the rest of the array as fallback.
```
seanhealy, line 30: This no longer takes an array.

seanhealy (changes requested):
Two really little questions then I think we're good to go!

packages/components/src/Popover/Popover.mdx
robbyjobber: ah good question @seanhealy.  full width reference elements causes an unfavourable visual behaviour with popperjs
image

I've done some HTML manipulation and achieved this result
image

seanhealy (approved):

PR #500, opened 2021-04-05
docs(design): Docs With Borders (write docs for Borders)

chris-at-jobber:
## Motivations

Sean said this on a PR:
image

And I thought
image

## Changes

Add docs for our Borders foundation elements

### Added

- docs for borders
- reference in that to Divider

---

Random photo of Atlantis

### DIFF
--- a/packages/design/src/Borders.mdx
+++ b/packages/design/src/Borders.mdx
@@ -1,7 +1,7 @@
 ---
 name: Dividing Content
 menu: Design
-route: /borders-and-dividers
+route: /borders
 ---
 
 # Borders and Dividers
@@ -12,8 +12,14 @@ platforms: borders and dividers.
 
 ## Borders
 
-To add a border to an element, use the border variables in the component's
+### Border Widths
+
+To add a border to an element, use the border-width variables in the component's
 styling to ensure consistent border thickness throughout Jobber.
+
+Border style should be `solid` unless a specific use case calls for some other
+format, such as indicating a dropzone on the
+[`InputFile` component](https://atlantis.getjobber.com/components/input-file).
 
 export function Example({ of: size }) {
   const style = {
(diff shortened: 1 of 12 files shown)


packages/design/src/Borders.mdx
```diff
@@ -0,0 +1,64 @@
...
+The `base` border width should be used for most use cases (see
+[`Card`](https://atlantis.getjobber.com/components/card) and
+[`List` items](https://atlantis.getjobber.com/components/list) for in-context
+examples). You may want to consider thicker borders when the border conveys a
+closing of already-divided sections, such as a
+[`Table` footer](https://atlantis.getjobber.com/components/table).
+
+### Border Colors
+
+Borders should use `var(--color-grey--lighter)` for most use cases.
```
eddysims, line 50: thoughts on adding a `--color-border` to make this a little easier.

packages/design/src/Borders.mdx
```diff
@@ -0,0 +1,64 @@
+---
+name: Dividing Content
```
eddysims, line 2: I understand that this name is probably more appropriate as we are also talking about `Divider` here. However it did take me a few seconds to figure out where to go to look for border information.

packages/design/src/Borders.mdx
```diff
@@ -0,0 +1,64 @@
...
+### Border Widths
+
+To add a border to an element, use the border-width variables in the component's
+styling to ensure consistent border thickness throughout Jobber.
+
+Border style should be `solid` unless a specific use case calls for some other
+format, such as indicating a dropzone on the
+[`InputFile` component](https://atlantis.getjobber.com/components/input-file).
+
+export function Example({ of: size }) {
```
eddysims, line 24: can we name this something more descriptive.

eddysims (changes requested):

eddysims:
@chris-at-jobber a few thoughts here.

packages/design/src/Borders.mdx
seanhealy: Variations of this component is actually used in a few different design guides. I wonder if we can generalize them. 🤔

packages/design/src/Borders.mdx
seanhealy: That would explain why I couldn't find it in the docs. 😂

packages/design/src/Borders.mdx
chris-at-jobber: I think I'll leave this as is for now, and we can see if there's a good standard... I think if I leave this as `Example` and we come back to standardizing, it'll be easier to flag at that point?

PR #504, opened 2021-04-08
fix: resolve render issues when using controlled component inside form component 

joelwarrington-jobber:
## Motivations

When using a controlled `<FormInput />` such as `<InputText />` renders are inconsistent, especially when first entering a value. What would happen is that when a user first focuses on an `<InputText />` (or any input that is based off of `<FormInput />`) and types in a value, the first key press is ignored and subsequent keypress trigger the specified `onChange` callback.

After some investigation into this issue, I have noticed that the underlying package `react-hook-form` used for `<Form />` and `<FormInput />` sets the value directly to the input which is not expected as there are some optimizations used to minimize renders (Proxy).

Thus to ensure renders occur when the provided `value` prop changes, you should use `watch` and `setValue` passed by the `useFromContext` hook. To read more about this, I'd check-out [watch docs](https://react-hook-form.com/v5/api/#watch), [setValue docs](https://react-hook-form.com/v5/api/#setValue), and the [custom register tab of controlled component usage docs](https://react-hook-form.com/advanced-usage#ControlledmixedwithUncontrolledComponents)

One requirement of the change implemented here - any controlled input REQUIRES a `name` prop to work correctly. I'm not sure what the best way to document this in the types so I'm open to suggestions. Because controlled inputs require `name`, it is required for all `<Autocomplete />` components.

### Fixed

- fixed inconsistent renders when using controlled input inside form component

## Testing

To make your life easier, I've created two branches to test my changes, to view without the changes pull branch [test-input-changes-without-optimized-rendering](https://github.com/GetJobber/atlantis/tree/test-input-changes-without-optimized-rendering).

Navigate to `/components/form` and type in the inputs in the first example.

image

To view with the changes pull branch [test-input-changes-with-optimized-rendering](https://github.com/GetJobber/atlantis/tree/test-input-changes-with-optimized-rendering). Do the same as before (navigate to `Form.mdx`) and you will see the same example, which works as expected.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -303,7 +303,7 @@ export function FormField({
        */
 
       if (value !== undefined && name !== undefined) {
-        if (isControlled) {
+        if (isControlled && watch(name) !== value) {
           setValue(name, value.toString());
         }
       }
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -312,11 +312,6 @@ export function FormField({
         name !== undefined
       ) {
         if (isControlled) {
-          console.log("setting value in useEffect, ", {
-            value,
-            controlledValue,
-            name,
-          });
           setValue(name, value.toString());
         }
       }
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -8,7 +8,12 @@ import React, {
 } from "react";
 import classnames from "classnames";
 import uuid from "uuid";
-import { ValidationRules, useForm, useFormContext } from "react-hook-form";
+import {
+  ValidationRules,
+  useForm,
+  useFormContext,
+  useWatch,
+} from "react-hook-form";
 import styles from "./FormField.css";
 import { Icon } from "../Icon";
 import { InputValidation } from "../InputValidation";
@@ -202,6 +207,7 @@ export function FormField({
   const {
     register,
     setValue,
+    control,
     /**
      * We currently only use `errors` from formState, but there are lots of
      * other values that come with it. https://react-hook-form.com/api#formState
@@ -280,6 +286,13 @@ export function FormField({
       [styles.select]: type === "select",
     });
 
+    // const controlledValue = watch(name || "", );
+    const controlledValue = useWatch({
+      control,
+      name: name || "",
+      defaultValue: defaultValue || value?.toString() || "",
+    });
+
     const fieldProps = {
       id: identifier,
       className: fieldClasses,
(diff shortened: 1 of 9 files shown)


packages/components/src/FormField/FormField.tsx
```diff
@@ -281,11 +283,32 @@ export function FormField({
...
-      value: value,
+      value:
+        typeof name !== "undefined" && typeof value !== "undefined"
+          ? watch(name)
+          : undefined,
       inputMode: keyboard,
-      ...(defaultValue && { defaultValue: defaultValue }),
+      ...(defaultValue
+        ? { defaultValue: defaultValue }
+        : value && { defaultValue: value?.toString() }),
```
eddysims, line 293: This is forcing a `defaultValue` and removing the `value` on all inputs. Is that intended?

eddysims (changes requested):
@joelwarrington-jobber a few questions here for a first pass.

packages/components/src/FormField/FormField.tsx
joelwarrington-jobber: I have removed this and updated the `watch` function to use the `defaultValue` as a default value if specified or `value`. This fixes issues with first render not displaying values on the `<Select/>` component.

packages/components/src/FormField/FormField.tsx
```diff
@@ -281,11 +283,32 @@ export function FormField({
       disabled: disabled,
       readOnly: readonly,
       onChange: handleChange,
-      value: value,
+      value:
+        typeof name !== "undefined" && typeof value !== "undefined"
```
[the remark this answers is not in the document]
joelwarrington-jobber: added `isControlled` condition which will also set an autogenerated name if set to `true`

eddysims:
@joelwarrington-jobber I took another look at this. It looks like when a field is `controlled` without having `validations` set it is re-rendering multiple times per keypress. 

rerenders

joelwarrington-jobber:
> @joelwarrington-jobber I took another look at this. It looks like when a field is `controlled` without having `validations` set it is re-rendering multiple times per keypress.
> 
> rerenders

I don't entirely have the capacity to investigate further at the moment, but there certainly is something weird going on. I am assuming that it is likely coming from some of the `react-hook-forms` wizardry. 

Does the multiple renders block this from being able to be approvable?

joelwarrington-jobber:
> @joelwarrington-jobber I took another look at this. It looks like when a field is `controlled` without having `validations` set it is re-rendering multiple times per keypress.
> 
> rerenders

Reviewed the issue with multiple renders. It looks like when using `watch` or the `useWatch` hook, you run into this issue. In reality, we don't need to watch the value to be setting it as a prop, as `setValue` will handle that for us. The only instance where this is not applicable is for `Select`'s which for some reason don't get set. Using `value=value` is necessary in that case.

eddysims (approved):

PR #508, opened 2021-04-13
feat(components): Add new Disclosure component

kingstonfung:
## Motivations

- Contributing a reusable disclosure component that lives within Atlantis' design system
- [Accepted Proposal](https://github.com/GetJobber/atlantis/tree/master/docs/proposals/Disclosure)
- [Figma](https://www.figma.com/file/NSFGqkxgycP8JbEwmaDW8d/Design-System-Contribution-%5BDisclosure%5D?node-id=26%3A2)

Screen Shot 2021-04-13 at 8 22 57 AM
Screen Shot 2021-04-13 at 8 22 26 AM

## Changes

- `<Disclosure>` will become available after merging in this PR

### Added

- Added `/Disclosure` folder, containing the component and related css and test files

## Testing

1. Simply pull down this branch, and start up the service
2. Head over to http://localhost:3333/components/disclosure and observe the newly added component

---

oppa gangnam style


packages/components/src/Disclosure/Disclosure.css
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.summary:hover {
+  background: var(--color-yellow--lightest);
+}
+
+.summary:focus,
+.summary:active {
+  border: none;
+  outline-color: var(--color-yellow);
```
chris-at-jobber, line 35: ~can we use `var(--color-focus)` for this?~

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.details[open] > .content {
+  -webkit-animation: openAnimation var(--timing-base);
+  animation: openAnimation var(--timing-base);
+}
+
+@keyframes openAnimation {
+  from {
+    margin-top: -20px;
```
chris-at-jobber, line 57: Let's use one of our `space` [variables](https://atlantis.getjobber.com/space) here

packages/components/src/Disclosure/Disclosure.css
chris-at-jobber: We use `var(--shadow-focus)` rather than `outline` to convey :focus state where possible:
image
image

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -0,0 +1,64 @@
...
+  cursor: pointer;
+  outline-color: none;
+}
+
+.summary::-webkit-details-marker {
+  display: none;
+}
+
+.summary:hover {
+  background: var(--color-yellow--lightest);
```
chris-at-jobber, line 29: Should also include `.summary:focus` here

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -0,0 +1,64 @@
...
+conter, which result in an odd square rendering glitch when
+displaying the background color together.
+*/
+.arrowIconWrapper > svg > path {
+  transform-origin: 50% 50% 0;
+  transition: transform var(--timing-base);
+}
+
+.flippedVertical > svg > path {
+  transform: rotate(180deg);
```
chris-at-jobber, line 13: Can we do a `rotateX` here so the arrow flips, rather than spins?

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.summary {
+  display: flex;
+  border: none;
+  list-style: none;
+  cursor: pointer;
+  outline-color: none;
```
chris-at-jobber, line 21: Can we add `transition: background var(--timing-quick), box-shadow(--var-timing-base)` so we get  a smoother transition on the hover/focus state changes?

packages/components/src/Disclosure/Disclosure.css
chris-at-jobber: And can we reverse the animation when `Disclosure` closes?

chris-at-jobber (changes requested):
Looking good so far! Couple of CSS variable notes, and a few design/QA things:

Something's up with the layout on Safari:
image

Spacing and alignment is a bit off on the title element compared to the mockup:
image

Over to @seanhealy for any additional thoughts!

packages/components/src/Disclosure/Disclosure.css
kingstonfung: good call, and flip works well and I don't need to drill down into the `> svg > path`

kingstonfung:
@chris-at-jobber : 704e9c9 should take care most of the css changes, thanks for those tips! 

I'm going to take a deeper dive into the "rewinding" animation as mentioned... will see if there's any workarounds without letting the browser snapping close that `<details>` element.

kingstonfung:
@chris-at-jobber : After taking a deeper dive into the closing animation, it seems like it is not fully supported yet... 

[From MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#events):
> Note: Unfortunately, at this time there's no built-in way to animate the transition between open and closed.

The only one 'hackaround' example I found is to fully disable the browser to pickup the closing initiation, and then animate the close manually, **and then time it** with a delay and take away the "open" attribute to say it is closed:

https://codepen.io/sh0ji/pen/zbgxGL?editors=0110 (JS line 101)

So I don't think it is very viable a this point, but let me know what you think. Thanks!

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -0,0 +1,69 @@
+.arrowIconWrapper {
+  transform-origin: 50% 50% 0;
+  transition: transform var(--timing-base);
```
chris-at-jobber, line 3: Seeing this in practice I think `timing-quick` would be better for the arrow, I'm finding my eye gets pulled to it a bit too much 👁️

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -0,0 +1,69 @@
...
+}
+
+.details[open] > .content {
+  -webkit-animation: openAnimation var(--timing-base);
+  animation: openAnimation var(--timing-base);
+}
+
+@keyframes openAnimation {
+  0% {
+    margin-top: calc(var(--space-largest) * -1);
```
chris-at-jobber, line 59: Played around with this a bit on local and I think if we use `small` here, we can actually drop the 50% keyframe and it feels a bit smoother

chris-at-jobber (changes requested):
Something I've just encountered and I either missed on the first pass, or it's new: clicking on the details is closing the Disclosure 😬 
disclosure_wrongclick

kingstonfung:
That click bug was definitely my bad, I was using `onClick` which captures the click event from _everywhere_ instead of `onToggle`.

Now the tricky part: I'm getting hit by `onToggle` not being supported by TS, but it works / lints / builds fine locally 😢 

```
Property 'onToggle' does not exist on type
'DetailedHTMLProps<DetailsHTMLAttributes<HTMLElement>, HTMLElement>'.
```

This one I definitely need help to troubleshoot

packages/components/src/Autocomplete/__snapshots__/Autocomplete.test.tsx.snap
```diff
@@ -80,14 +80,14 @@ exports[`renders correctly when invalid 1`] = `
   >
     <label
       className="label"
-      htmlFor="123e4567-e89b-12d3-a456-426655440043"
+      htmlFor="123e4567-e89b-12d3-a456-426655440044"
```
chris-at-jobber, line 83: @kingstonfung this is a flakey snapshot that _might_ be contributing to your issues. Try kicking these changes out of your PR and see if that helps?

packages/components/src/Disclosure/Disclosure.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+## Responsiveness
+
+- This component should handle both click and touch, as well as keyboard inputs
+- This component should fill the width of its container
+- If the component is less than `375px` wide, the title will wrap and should not
+  be truncated
+- Text increases/decreases when user zooms in/out
+
+## Mockup
```
chris-at-jobber, line 77: We can remove the entire mockup section now that we have a working-in-code version 😃

packages/components/src/Disclosure/Disclosure.mdx
```diff
@@ -0,0 +1,114 @@
...
+# Disclosure
+
+```ts
+import { Disclosure } from "@jobber/components/Disclosure";
+```
+
+<Playground>
+  <Disclosure title="Open to see more content organa palpatine lobot jawa kamino.">
+    <Content>
+      <p>
```
chris-at-jobber, line 21: If we can import `Text` it would be nice to show this example built with all Atlantis!

kingstonfung:
I added `--space-small` amount of padding at the top + bottom of the content wrapper, otherwise it appears too close to the `<summary>` bar:

Screen Shot 2021-04-14 at 6 32 29 PM

vs now:
Screen Shot 2021-04-14 at 6 32 12 PM

But let me know if that is unnecessary (or any other amounts we should tweak with)

kingstonfung:
Review comments & changes are all caught up now, thanks again!

The tricky part is commit # b35549c where I needed to bump our `@types/react` to `16.9.43` ... this one will need a closer look, please let me know how to run further tests (or other processes) to ensure it can be properly adopted into Atlantis.

Thanks!

chris-at-jobber:
Looks good to design! Will get some 🧑‍💻  eyes on this for you.

chris-at-jobber (approved):
Design approved 👍

packages/components/src/Disclosure/Disclosure.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+it("renders a Disclosure that is opened by default", () => {
+  const newProps = {
+    ...sampleProps,
+    defaultOpen: true,
+  };
```
seanhealy, line 28: Not a big deal at all but I think I like the way you are splatting in your props in the `should not attempt to call the handler when it is not set` test a bit better.

packages/components/src/Disclosure/Disclosure.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+it("should call the handler when toggling the Disclosure component", async () => {
+  const toggleHandler = jest.fn();
+  render(
+    <Disclosure onRequestToggle={toggleHandler} {...sampleProps}>
+      <span>Cotton candy tootsie roll lemon drops tiramisu cake tart.</span>
+    </Disclosure>,
+  );
+
+  const summaryElement = document.querySelector("summary");
```
seanhealy, line 60: Would be better to `queryByLabelText` or `queryByText` here rather than looking for a `summary` element as it better reflects that the user will interact with and couples us less to the implementation.

https://testing-library.com/docs/queries/about

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export function Disclosure({
+  children,
+  title,
+  open,
+  defaultOpen = false,
+  onRequestToggle,
+}: DisclosureProps) {
+  const [isOpen, setOpen] = useState(open || defaultOpen);
```
seanhealy, line 45:
```suggestion
  const [isOpen, setOpen] = useState(open ?? defaultOpen);
```

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  const summaryClassName = classnames(styles.summary);
+  const detailsClassName = classnames(styles.details);
+  const contentClassName = classnames(styles.content);
+  const summaryWrapClassName = classnames(styles.summaryWrap);
```
seanhealy, line 50: If we don't have any conditions on these we can just reference the `styles.foo` directly in the jsx.

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,78 @@
...
+   * For use when the component is being used as an Uncontrolled Component.
+   * @default false
+   */
+  readonly defaultOpen?: boolean;
+
+  /**
+   * This function would toggle the open state.
+   * For use when the component is being used as a Controlled Component.
+   */
+  onRequestToggle?(isOpen: boolean): void;
```
seanhealy, line 35: I think it might be good to change the verbiage on the parameter here as it's really requesting this new state.

Perhaps, `requestOpen`, `shouldOpen`?

I was also wondering if two methods a `onRequestOpen` and `onRequestClose` would be better but, I don't think I like that if we can get a good name for the parameter.

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { ReactNode, useState } from "react";
+import classnames from "classnames";
+import styles from "./Disclosure.css";
+import { Heading } from "../Heading";
+import { Icon } from "../Icon";
+
+export interface DisclosureProps {
```
seanhealy, line 7: Do we need to export `DisclosureProps` here?

seanhealy (changes requested):
Few thoughts back your way. Looking great though. Thanks so much!

packages/components/src/Disclosure/Disclosure.tsx
kingstonfung: This one is an odd one... I had to export it otherwise CI was giving me troubles:
https://app.circleci.com/pipelines/github/GetJobber/atlantis/3123/workflows/4fb09704-029b-4cbd-9df2-5819d32a3ad4/jobs/13958

(however local builds & lint worked fine)

packages/components/src/Disclosure/Disclosure.test.tsx
kingstonfung: ah... I must have gotten myself into a weird state during WIP for this particular test.

I fall back to use `querySelector` because I wans't able to target the `<summary>` element (since it'll just pickup the `<h4>` heading)

But retrying that seems to work now so I'll switch back to use `queryByText`

kingstonfung:
@seanhealy : Please take a look at febf0ae and let me know what you think

I've also addressed a bug where the component would lose its mind and be stuck in a loop of open-close if it's set to be opened by default.

Also responded back to your question about exporting `DisclosureProps` at https://github.com/GetJobber/atlantis/pull/508#discussion_r614343797

Thanks!

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,84 @@
...
+   * For use when the component is being used as an Uncontrolled Component.
+   * @default false
+   */
+  readonly defaultOpen?: boolean;
+
+  /**
+   * This function would toggle the open state.
+   * For use when the component is being used as a Controlled Component.
+   */
+  requestOpen?(shouldOpen: boolean): void;
```
seanhealy, line 35: Sorry, my previous comment was confusing. I was thinking about the argument not the function name. I think keeping the function named `onRequestToggle` probably makes the most sense. :)

packages/components/src/Disclosure/Disclosure.tsx
seanhealy: If you take it out of the exports in index.ts as well does it work?

seanhealy (changes requested):
Two last little thoughts and I think we're good here! Thanks so much. 😄

seanhealy (approved):

PR #510, opened 2021-04-14
docs: remove ComponentStatus and clean up some "related components" and "usage" verbiage

chris-at-jobber:
## Motivations

We've talked before about the unreliable nature of our "Component Status" badges in Atlantis, including that we should...

- [better define them](https://jobber.atlassian.net/browse/JOB-27069)
- improve them

Until such time, we should remove them as they lead to more questions than answers!

**Edit:** I thought about this some more, and given that we believe we _can_ deliver value with this information but need to better define and double-check the statuses of the existing components in Atlantis, I have changed this to render the component status  but not _display_ it, so that we do not need to manually re-add the status components. 

## Changes

Prevent component status badges from displaying in Atlantis.

### Removed

- The component status component is no longer visible

## Testing

Look at the docs for the components; you should not see these badges:
image

---

Random photo of Atlantis

### DIFF
--- a/packages/docx/src/ComponentStatus/ComponentStatus.tsx
+++ b/packages/docx/src/ComponentStatus/ComponentStatus.tsx
@@ -76,7 +76,7 @@ export function ComponentStatus({
   };
 
   return (
-    <div>
+    <div style={{ display: "none" }}>
       <StatusShield {...stageData} />
       <StatusShield {...responsiveData} />
       <StatusShield {...accessibleData} />
--- a/packages/components/src/Autocomplete/Autocomplete.mdx
+++ b/packages/components/src/Autocomplete/Autocomplete.mdx
@@ -6,15 +6,12 @@ showDirectoryLink: true
 ---
 
 import { Playground, Props } from "docz";
-import { ComponentStatus } from "@jobber/docx";
 import { Button } from "../Button";
 import { Autocomplete } from ".";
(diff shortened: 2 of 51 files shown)


packages/docx/src/ComponentStatus/ComponentStatus.tsx
```diff
@@ -76,7 +76,7 @@ export function ComponentStatus({
   };
 
   return (
-    <div>
+    <div style={{ display: "none" }}>
```
eddysims, line 79: I'm wondering if instead of returning `display: none`. We should just return a [fragment](https://reactjs.org/docs/fragments.html) `<></>` for now so that we are totally removing this element from the DOM.

Thoughts?

eddysims (commented):
@chris-at-jobber just one question. If you disagree I'm good with what you have here.

packages/docx/src/ComponentStatus/ComponentStatus.tsx
chris-at-jobber: Oh that make sense! Didn't know that was a thing 😄

chris-at-jobber: @eddysims after our offline discussion about the larger changes this requires, going to leave as-is

eddysims (approved):

PR #513, opened 2021-04-21
fix(components): update popover return type and examples

robbyjobber:
## Motivations
- upon using popover, I was receiving invalid jsx element
image
- this is due to popover returning false if `open=false` and the popover component otherwise

### Changed

- added a ternary operator to return either the component or a jsx fragment depending on the `open` condition.

### Removed

- removed `dismissible` from the mdx examples

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Popover/Popover.tsx
+++ b/packages/components/src/Popover/Popover.tsx
@@ -50,26 +50,26 @@ export function Popover({
     },
   );
 
-  return (
-    open && (
-      <div
-        role="dialog"
-        ref={setPopperElement}
-        style={popperStyles.popper}
-        className={classes.popover}
-        {...attributes.popper}
-      >
-        <div className={classes.dismissButton}>
-          <ButtonDismiss onClick={onRequestClose} ariaLabel="Close dialog" />
-        </div>
-        {children}
-        <div
-          ref={setArrowElement}
-          className={classes.arrow}
-          style={popperStyles.arrow}
-        />
+  return open ? (
+    <div
+      role="dialog"
+      ref={setPopperElement}
+      style={popperStyles.popper}
+      className={classes.popover}
+      {...attributes.popper}
+    >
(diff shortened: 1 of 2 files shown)


packages/components/src/Popover/Popover.tsx
```diff
@@ -50,26 +50,26 @@ export function Popover({
...
+        <ButtonDismiss onClick={onRequestClose} ariaLabel="Close dialog" />
       </div>
-    )
+      {children}
+      <div
+        ref={setArrowElement}
+        className={classes.arrow}
+        style={popperStyles.arrow}
+      />
+    </div>
```
rjdestigter, line 70: You should be able to just return `null` or `undefined`

packages/components/src/Popover/Popover.tsx
```diff
@@ -50,26 +50,26 @@ export function Popover({
...
+  ) : (
+    <></>
```
eddysims, line 72: @robbyjobber I believe what we would normally do here is something like this.

```tsx
return (
  <>
    {open && (
      ...
    )}
  </>
)
```

This way, we are always returning _at least_ a fragment, and we don't end up with a weird ternary in our component.

eddysims (changes requested):

packages/components/src/Popover/Popover.tsx
robbyjobber: unfortunately can't return null cuz of the lint and undefined causes us to run into the same error as `false | JSXElement` 😭

packages/components/src/Popover/Popover.tsx
robbyjobber: oh makes sense!

seanhealy (dismissed):

eddysims (approved):

robbyjobber:
Thank you!

arnold

PR #519, opened 2021-04-28
docs(design): More detail on semantic color usage

chris-at-jobber:
## Motivations

In adding semantic colors to Atlantis, we should help people know how to use them.

Content references [this Figma.](https://www.figma.com/file/PNo7p9G0OxygCOjgm4rpngav/Product-Colors?node-id=450%3A124)

While writing the docs, I had one of those "as you write it, you feel like the system is bad if you need to write too much" moments. As such, there are also some adjustments to the colors.css design package as well, to clarify the intent of the light and dark shades of the status colors and eliminate redundancy with extraneous "surface status" colors. 

### Added

- Added guidelines for how to use semantic colors

### Changed

- Renamed some of the "Status" variables and removed the `surface--{status}` colors

## Testing

Go to the colors page in Atlantis and take a look!

---

### DIFF
--- a/packages/design/src/Colors.mdx
+++ b/packages/design/src/Colors.mdx
@@ -68,7 +68,9 @@ Use to signify that an interactive element is disabled.
 
 ##### Disabled--Secondary
 
-Use when a disabled element needs more than one shade.
+Use when a disabled element needs more than one color to be readable in a
+disabled state; for example, a button's background and label colors must be
+different.
 
 #### Focus
 
--- a/packages/design/src/Colors.mdx
+++ b/packages/design/src/Colors.mdx
@@ -33,7 +33,7 @@ readability.
 
 #### Heading
 
-Headings have a bold, dark color to cement their hierarchy.
+Headings have a bold, high-contrast color to cement their hierarchy.
 
 #### Text
 
@@ -110,8 +110,8 @@ Surfaces are the background-colors of almost every element in Jobber.
 
 #### Surface
 
-Most elements in Jobber have a white surface; if clickable, they have a light
-hover color.
+Most elements in Jobber have a light surface to indicate their nearness to the
+user; if clickable, they have a hover color.
 
 #### Surface--Background
 
--- a/packages/design/src/Colors.mdx
+++ b/packages/design/src/Colors.mdx
@@ -18,138 +18,141 @@ Atlantis is built with a
 overtop of a base palette. Use the semantic color values whenever possible to
 ensure you are using colors for their intended purpose.
 
-## Typography
-
-Typographic elements in Jobber should use consistent colors to ensure readability.
-
-### Heading
-
-Headings have a bold, dark color to cement their hierarchy.
-
-### Text
-
-A slightly softer color is used for body text for greater readability.
-
-#### Text--Supporting
-
-Text that is relevant but less important can be lower-contrast to suggest its’ reduced importance.
-
-## Interactive
-
-Use these colors in buttons and form  controls to communicate the presence and meaning of interaction.
-
-### Interactive
-
-The default color used for interactive elements.
-
-### Destructive
-
-Use to signfify that an interaction will destroy something in the users’ account or workflow.
-
-### Cancel
-
-Use to signify that an interaction will cancel or exit a workflow
-
-### Disabled
-
-Use to signify that an interactive element is disabled.
-
-#### Disabled--Secondary
-
-Use when a disabled element needs more than one shade.
-
-### Focus
-
-Use to indicate that an element has been focused.
-
-## Status
-
-Use these colors in labels, icons, filters, alerts, and other elements where color can add meaning to the state of the system or an item in the system.
-
-### Critical
-
-Action required; user must see this status to be unblocked.
-
-### Warning
-
-Action _may_ be required as a consequence of current state.
-
-### Success
-
-No action required; an action has completed successfully.
-
-### Informative
-
-No action required; but helpful to know about.
-
-### Inactive
-
-No action required; not part of an active workflow.
-
-## Surfaces
-
-Surfaces are the background-colors of almost every element in Jobber.
-
-### Surface
-
-Most elements in Jobber have a white surface; if clickable, they have a light hover color.
-
-### Surface--Background
-
-A slightly darker surface gives a receded appearance relative to main surfaces.
-
-### Surface--Reverse
-
-When a strong constrast is needed with the rest of the application, a reversed surface can be used.
-
-#### Status Surfaces
-
-Elements that convey status have surface colors to help signify their meaning.
-
-##### Surface--Error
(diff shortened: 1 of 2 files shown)


packages/design/src/colors.css
```diff
@@ -46,10 +46,6 @@
   --color-surface--hover: var(--color-yellow--lightest);
   --color-surface--background: var(--color-grey--lightest);
   --color-surface--reverse: var(--color-greyBlue--dark);
-  --color-surface--success: var(--color-green--lighter);
```
chris-at-jobber, line 49: These surface-status values are now wrapped up in the status colors, as they should be

packages/design/src/colors.css
```diff
@@ -17,24 +17,24 @@
...
-  --color-informative--light: var(--color-lightBlue--lightest);
-  --color-informative--dark: var(--color-lightBlue--dark);
+  --color-informative--surface: var(--color-lightBlue--lightest);
+  --color-informative--onSurface: var(--color-lightBlue--dark);
 
   --color-inactive: var(--color-blue--light);
-  --color-inactive--light: var(--color-blue--lightest);
-  --color-inactive--dark: var(--color-blue--dark);
+  --color-inactive--surface: var(--color-blue--lightest);
+  --color-inactive--onSurface: var(--color-blue--dark);
```
chris-at-jobber, line 37: Renaming these status variants to be more clear about their intended purpose

packages/design/src/Colors.mdx
```diff
@@ -24,4 +24,135 @@ ensure you are using colors for their intended purpose.
   alongside color to ensure your content can be understood by everyone.
 </Banner>
 
+## Usage Guidelines
```
chris-at-jobber, line 27: Ideally there's a nice way to bring the swatches for a section together with the usage guidelines, but until then I think this works. There's something poetic about semantic guidelines being detached from the visuals they describe anyways, right? 😉

chris-at-jobber: Using the h2 as an anchor point makes this reasonable to jump between, for now.

image

packages/design/src/Colors.mdx
```diff
@@ -24,4 +24,135 @@ ensure you are using colors for their intended purpose.
...
+
+### Typography
+
+Typographic elements in Jobber should use consistent colors to ensure
+readability.
+
+#### Heading
+
+Headings have a bold, dark color to cement their hierarchy.
```
eddysims, line 36: Should we use the word `dark` here? If we eventually theme Atlantis, then `dark` is no longer the right word.

Maybe `high contrast` or something like that?

packages/design/src/Colors.mdx
```diff
@@ -24,4 +24,135 @@ ensure you are using colors for their intended purpose.
...
+reduced importance.
+
+### Interactive
+
+Use these colors in buttons and form controls to communicate the presence and
+meaning of interaction.
+
+#### Interactive
+
+The default color used for interactive elements.
```
eddysims, line 54: There is double `Interactive` headings. Is that on purpose?

packages/design/src/Colors.mdx
```diff
@@ -24,4 +24,135 @@ ensure you are using colors for their intended purpose.
...
+
+Use to signify that an interaction will cancel or exit a workflow
+
+#### Disabled
+
+Use to signify that an interactive element is disabled.
+
+##### Disabled--Secondary
+
+Use when a disabled element needs more than one shade.
```
eddysims, line 71: This seems strange to me. Why would this be a use case?

packages/design/src/Colors.mdx
```diff
@@ -24,4 +24,135 @@ ensure you are using colors for their intended purpose.
...
+
+No action required; not part of an active workflow.
+
+### Surfaces
+
+Surfaces are the background-colors of almost every element in Jobber.
+
+#### Surface
+
+Most elements in Jobber have a white surface; if clickable, they have a light
```
eddysims, line 113: again here, not sure if we want to be color specific?

packages/design/src/Colors.mdx
```diff
@@ -24,4 +24,135 @@ ensure you are using colors for their intended purpose.
...
+The primary color associated with our brand, from website to ads to product; AKA
+“Jobber Green”.
+
+#### Brand--Highlight
+
+Use to highlight an element in a way that aligns with our website. Use with
+caution, it’s _bright!_
+
+<br />
+<br />
```
eddysims, line 154: Why these 2 `br`s here?

packages/design/src/colors.css
eddysims: So if i understand these correct, If i had a component with a background of `--color-success--surface` anything on top of that would use `--color-success--onSurface`

eddysims (changes requested):
@chris-at-jobber a few thoughts in here for you.

packages/design/src/Colors.mdx
chris-at-jobber: 👍

packages/design/src/Colors.mdx
chris-at-jobber: 👍

packages/design/src/Colors.mdx
chris-at-jobber: Yep, Interactive is the category and this interactive color is the default color for interactive elements

packages/design/src/Colors.mdx
chris-at-jobber: We don't have spacing in our typography components (because typically they're used in a web app and not a text document) so this gives some spacing to segment the two sections a bit more

packages/design/src/colors.css
chris-at-jobber: Yep!

packages/design/src/Colors.mdx
chris-at-jobber: Elements like checkbox, radio, button, all have multiple tones:

image
image
image

packages/design/src/colors.css
chris-at-jobber: This is not entirely reflective of all current elements that could use this approach (ie Banner), but holds true for elements like InlineLabel, and I think we can look at making Banner, Toast, etc use this approach more consistently.

packages/design/src/Colors.mdx
chris-at-jobber: Changed this to "light" b/c even in dark mode, the foreground surfaces, will be lighter than backgrounds.

packages/design/src/Colors.mdx
chris-at-jobber: Actually I just had an idea to make this more clear, though I still think we need the color. Will update shortly.

eddysims (approved):

PR #522, opened 2021-05-05
feat(components): use semantic colors in Button

chris-at-jobber:
## Motivations

### Changed

- a bunch of CSS in button; mostly replacing internal custom CSS properties from that component to leverage the Atlantis system properties

### Removed

- A bunch of Typescript that kind of scares me!

## Testing

Go to the component page and test focus, hover states, etc, try adding icons to different variations and sizes

---

Random photo of Atlantis


packages/components/src/Button/Button.tsx
```diff
@@ -155,64 +154,11 @@ export function Button(props: ButtonProps) {
...
-        textColor={getColor(variation, type, disabled)}
-      >
-        {label}
-      </Typography>
+      {icon && <Icon name={icon} size={size} />}
+      <span className={styles.buttonLabel}>{label}</span>
     </>
   );
 }
-
```
chris-at-jobber, line 187: I'm open to thoughts on whether removing all of this is better or not from a technical perspective; what I want to get us away from is using that `brightness(80%)` filter on hover and focus states, and I think these functions only passing in one color are why we used that filter.

chris-at-jobber: Also my thinking around using `buttonLabel` and CSS instead of `Typography`

chris-at-jobber:
One note: I'm no longer clear on _how_ disabled actually works but it _seems_ fine?

packages/components/src/Button/Button.css
```diff
@@ -24,121 +13,112 @@
...
-}
-
-.work:hover,
-a.work:hover,
-.work:focus,
-a.work:focus {
-  border-color: var(--button--color-work--hover);
-  background-color: var(--button--color-work--hover);
+  --button--color-variation: var(--color-interactive);
+  --button--color-variation--hover: var(--color-interactive--hover);
```
chris-at-jobber, line 61: These are redundant with the button's base CSS, but without a `.work` in here the tests blew up because the typing from css.d.ts wouldn't have a `work` string.

packages/components/src/Button/Button.css
```diff
@@ -24,121 +13,112 @@
...
 }
 
-.secondary:hover,
-.secondary:focus,
-a.secondary:hover,
-a.secondary:focus {
-  background-color: var(--button--color-secondary);
-  filter: brightness(80%);
+.secondary {
+  border-color: var(--button--color-variation);
```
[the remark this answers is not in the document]
chris-at-jobber: I was doing some thinking about Button outside of the semantic color scope of things, and maybe there's some work we could do in a follow-up where we actually just set a `default` Button that looks like the current `work - secondary`, and remove those props. Basically, set that button as the default, and then you can _modify_ it as needed to have more or less emphasis (primary/tertiary, would probably change to something like primary/subtle) and change the variation (learning, destructive, cancel) as needed.

packages/components/src/Button/Button.css
```diff
@@ -162,6 +182,52 @@ a.tertiary:focus {
 
 /* Icon */
 
+.button svg path {
```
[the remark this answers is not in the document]
chris-at-jobber: moved all of this into the color/fill/background-color theming stuff, so I think as we discussed this should be OK to no longer comment out?

eddysims (approved):

PR #523, opened 2021-05-05
feat(components): use semantic colors in Switch

chris-at-jobber:
## Motivations

### Changed

- a bunch of CSS in Switch; mostly replacing internal custom CSS properties from that component to leverage the Atlantis system properties

### Removed

- Old component-specific properties where relevant

## Testing

Go to the component page and test focus, disabled, etc

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Switch/Switch.css
+++ b/packages/components/src/Switch/Switch.css
@@ -1,5 +1,5 @@
 :root {
-  --switch--pipSize: calc(var(--space-base) * 1.875); /* equivalent to 30px */
+  --switch--pipSize: calc(var(--space-base) * 1.875);
   --switch--labelWidth: calc(var(--switch--pipSize) * 1.3);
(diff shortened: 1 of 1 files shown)


packages/components/src/Switch/Switch.css
```diff
@@ -1,5 +1,5 @@
 :root {
-  --switch--pipSize: 30px;
+  --switch--pipSize: calc(var(--space-base) * 1.875); /* equivalent to 30px */
```
eddysims, line 2: This comment isn't necessary I don't think. 

The danger with a comment like this is if `--space-base` were to change, this would be incorrect.

eddysims (changes requested):
@chris-at-jobber one small comment here

chris-at-jobber:
@eddysims done!

eddysims (approved):

PR #528, opened 2021-05-05
feat(components): use semantic colors in FormField and Inputs

chris-at-jobber:

I changed all of these components in one PR as the changes are _extremely_ similar, in many cases fairly entwined with each other, and I think easier to review in this format.

### Changed

- some CSS properties in each component; mostly replacing descriptive CSS properties from that component to leverage the Atlantis system semantic properties
- note that in FormField we are actually leveraging a component-specific custom property that in turn leverages our semantic colors, so I actually did _not_ remove that one!

### Removed

- Old component-specific or descriptive properties where relevant

## Testing

Go to the component pages, test out focus, hover, error, as relevant ❤️ 

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -4,7 +4,7 @@
   --field--placeholder-color: var(--color-greyBlue--light);
   --field--border-width: var(--border-base);
   --field--border-color: var(--color-border);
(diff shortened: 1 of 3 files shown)


chris-at-jobber:
@eddysims fixed up, includes some snapshots for InputAvatar which uses `button` version of InputFile

PR #548, opened 2021-05-15
feat(components): use aside and ButtonDismiss

chris-at-jobber:
## Motivations

Noticed some opportunities for cleanup while doing semantic color swap, coming back to it!

## Changes

Use `aside` instead of `div` for the Drawer parent container element. From [the spec:](https://www.w3.org/TR/html52/sections.html#the-aside-element)

> The aside element represents a section of a page that consists of content that is tangentially related to the content of the parenting sectioning content, and which could be considered separate from that content. Such sections are often represented as sidebars in printed typography.
> 
> The element can be used for typographical effects like pull quotes or sidebars, for advertising, for groups of nav elements, and for other content that is considered separate from the main content of the nearest ancestor sectioning content.
> 
> Assistive Technology may convey the semantics of the aside to users. This information can provide a hint to users as to the type of content. For example the role of the element, which in this case is "complementary", can be announced by screen reader software when a user navigates to an aside element. User Agents may also provide methods to navigate to aside elements.

From our docs, Drawer is for "sidebars that allow users to view supplementary content." This feels like a good opportunity to make this component more semantically accurate in construction.

Also, use ButtonDismiss to close the drawer, rather than a custom styled `button`.

### Changed

- div to aside on parent
- custom button for ButtonDismiss 

### Removed

- No more import of Icon into the custom button

---

Random photo of Atlantis


packages/components/src/Drawer/Drawer.tsx
```diff
@@ -29,14 +29,14 @@ export function Drawer({
   const drawerClassNames = classnames(styles.container, open && styles.open);
   return (
     <>
-      <div className={drawerClassNames} data-testid="drawer-container">
+      <aside className={drawerClassNames} data-testid="drawer-container">
```
eddysims, line 32: @chris-at-jobber my concern here is that `aside` is considered a landmark. Having multiple same landmarks (multiple drawers) on a page will throw an accessibility issue in [AXE](https://www.deque.com/axe/). 

The way around this is to assign an `aria-label` to the `aside`. 

Thoughts on this?

chris-at-jobber: Ooh good catch. I think we can assume/hope that you'd only have one Drawer per page, but otherwise I wonder if we could use the value entered as the "title" prop as an aria-label as a failsafe?

eddysims: Yeah I think thats a good idea. do you want to make that change and I will approve this.

chris-at-jobber: @eddysims done!

eddysims (approved):

PR #550, opened 2021-05-21
feat(components): Add ability to open Page intro links in a new tab 

irmalcol:
## Motivations
I was adding links to Jobber's help docs from a settings page (see the change [here](https://github.com/GetJobber/Jobber/pull/25164/files#diff-a1ec27a51bee40e7da882fbf2063446086ae0652ee3b1d3e8d89c82303cdbf9aR54)), and I couldn't get links open in a new tab when they were in the `intro` of the `Page` component.

### Added

- Add `externalIntroLinks` prop (boolean) to the `Page` component
  - Typescript error if `externalIntroLinks` is used without also specifyinig an `intro` prop
  - Defaults to `false` so this is a backwards-compatible change

## Testing

- Go to the Atlantis Page component docs locally ([localhost link](http://localhost:3333/components/page)), scroll to the bottom example with the Help Center link in the intro
- (control) verify the Help Center link opens in the same tab
- Add a new `externalIntroLinks={true}` and verify Help Center link opens in a new tab
- Change to `externalIntroLinks={false}` and verify Help Center link opens in the same tab

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -90,6 +90,7 @@ export type PageProps = PagePropsWithIntro | PagePropsNoIntro;
 export function Page({
   title,
   intro,
+  externalIntroLinks,
   subtitle,
   children,
   width = "standard",
@@ -169,7 +170,11 @@ export function Page({
           </div>
           {intro && (
             <Text size="large">
-              <Markdown content={intro} basicUsage={true} />
+              <Markdown
+                content={intro}
+                basicUsage={true}
+                externalLink={externalIntroLinks}
+              />
             </Text>
           )}
         </Content>
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -75,8 +75,8 @@ interface PagePropsWithIntro extends BasePageProps {
 interface PagePropsNoIntro extends BasePageProps {
   // No descriptions are necessary here, since they are already covered
   // inside of the PagePropsWithIntro interface.
-  readonly intro: undefined;
-  readonly externalIntroLinks: never;
+  readonly intro?: undefined;
+  readonly externalIntroLinks?: never;
 }
 
 /**
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -54,15 +54,36 @@ interface BasePageProps {
 }
 
 interface PagePropsWithIntro extends BasePageProps {
+  /**
+   * Content of the page. This supports basic markdown node types
+   * such as `_italic_`, `**bold**`, and `[link name](url)`.
+   *
+   * The `externalIntroLinks` prop can only be used if the
+   * `intro` prop is specified.
+   */
   readonly intro: string;
+
+  /**
+   * Causes any markdown links in the `intro` prop to open in a new
+   * tab, i.e. with `target="_blank"`.
+   *
+   * Defaults to `false`.
+   */
   readonly externalIntroLinks?: boolean;
 }
 
 interface PagePropsNoIntro extends BasePageProps {
+  // No descriptions are necessary here, since they are already covered
+  // inside of the PagePropsWithIntro interface.
   readonly intro: undefined;
   readonly externalIntroLinks: never;
 }
 
+/**
+ * We have a discriminated union here to ensure usage
+ * of externalIntroLinks is only valid if intro is
+ * also specified.
+ */
 export type PageProps = PagePropsWithIntro | PagePropsNoIntro;
 
 // eslint-disable-next-line max-statements
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -10,14 +10,8 @@ import { Button, ButtonProps } from "../Button";
 import { Menu, SectionProps } from "../Menu";
 import { Emphasis } from "../Emphasis";
 
-export interface PageProps {
+interface BasePageProps {
   readonly children: ReactNode | ReactNode[];
-
-  /**
-   * Content of the page. This supports basic markdown node types such as
-   * `_italic_`, `**bold**`, and `[link name](url)`
-   */
-  readonly intro?: string;
 
   /**
    * Title of the page.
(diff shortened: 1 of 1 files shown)


irmalcol:
I'm still planning on writing a Jest test to make sure the `target="_blank"` in included on rendered links with the new prop, but I wanted to get this up for some early eyes.

irmalcol:
@eddysims I decided to use a discriminated union (good reading [here](https://www.benmvp.com/blog/conditional-react-props-typescript/) and [here](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions)) instead of the `XOR` library to achieve the conditional validity based on the existence of the `intro` prop. I did it this way so we can avoid adding another 3rd-party dependency to the `Page` component. I'm curious what you think, since the `PageProps` are a bit more difficult to read now (which I don't like).

irmalcol:
Added Jest tests in https://github.com/GetJobber/atlantis/pull/550/commits/9338edd8202f85c1276981df5a0d35fba6890ef7.

packages/components/src/Page/Page.tsx
```diff
@@ -59,10 +53,44 @@ export interface PageProps {
...
+  readonly intro?: undefined;
+  readonly externalIntroLinks?: never;
+}
+
+/**
+ * We have a discriminated union here to ensure usage
+ * of externalIntroLinks is only valid if intro is
+ * also specified.
+ */
+export type PageProps = PagePropsWithIntro | PagePropsNoIntro;
```
eddysims, line 87: I still think the `XOR` is the right choice here.

As noted in the [docs](https://www.npmjs.com/package/ts-xor)

> Typescript's union operator (|) allows combining two types A and B, into a superset type C which can contain all the members of both A and B.

> But sometimes the requirements dictate that we combine two types with mutually exclusive members. So take the members A.a and B.b. Given type C = A | B then we want to impose the restriction that we can set either C.a or C.b but never both AND at least one of the two!

Using a union forces us to redefine `intro` and `externalIntroLinks` as you have done here.

If we are to use `XOR` we don't need to redefine this, because we are saying that `PageProps` will either be `BasePageProps` or `PagePropsWithIntro` but never a combination of both.

```ts
export type PageProps = XOR<BasePageProps, PagePropsWithIntro>
```

eddysims (changes requested):

packages/components/src/Page/Page.tsx
eddysims: To add on to this in regards to not wanting add a 3rd party package. 

This is a valid point to worry about, however we already use this package in lots of places throughout Atlantis, so it wouldn't add anything to the payload.

Also to note, that this is only a typing library. When we compile the Atlantis components down to javascript, the output of this file is 0.

irmalcol: You make good points in favour of using the `XOR` library, and I'm not necessarily opposed to using it, but I’m also not sure we’re on the same page in terms of how my code works with the discriminated union, and I think it would be good to sync up before we make a decision.

The behaviour offered by the `XOR` library is the exact behaviour I've implemented with the discriminated union. The "discriminated" part of the union is that both parts of the union can never be true at the same time (logical XOR) by virtue of how `PagePropsWithIntro` and `PagePropsNoIntro` are written.

I'll attempt to explain:

`PagePropsWithIntro` has `intro:  string`, and `PagePropsNoIntro` has `intro?: undefined`. `intro` can never be both a `string` and `undefined` at the same time, so `intro` acts as the discriminating member of the union.
- If `intro` is specified as a string, then we can _only_ use the `PagePropsWithIntro` part of the union, so we have the option to also specify the link behaviour because of `externalIntroLinks?: boolean`.
- If we don't specify `intro` (i.e. it's undefined), then we can _only_ use the `PagePropsNoIntro` part of the union, so we _are not allowed_ to set link behaviour because of `externalIntroLinks?: never`.

I don't think this behaviour is clear at first glance (otherwise we wouldn't be having this conversation), and that's not a good thing. I was hoping my code comments would suffice to clarify how this works. Does this explanation make sense? If so, do you still want to go with the `XOR` library? I'm totally OK to do that. It's still nice to know we have a pathway out of a mess if something ever happens to the 3rd-party lib.

eddysims: Yep, it makes sense what you are doing here, and if we didn't already have a pattern to do so I think you made the right choice. Ideally this would be done easier in Typescript. For the time being, let's use `XOR` here so that we keep with the current standards currently set in Atlantis.

You mentioned this

> I don't think this behaviour is clear at first glance

This is really the reason that we choose to use `XOR` currently in Atlantis. It makes doing this quite a bit more readable (especially when we get to more detailed types).

irmalcol: That works for me! Glad we're on the same page now. I'll make the change. Thanks for your 👀 on this @eddysims!

irmalcol:
I've updated this PR with what I _think_ are all the required changes, BUT I pulled in `master`, since it's been a few days since I last worked on this, and now I'm seeing some weird behaviour. I made a Loom to show the weirdness that you can see here: https://www.loom.com/share/006a0c30d7354f2ea2980c95a6dab25a.

@eddysims (a) does the code look good to you now, and (b) any idea what's going on here with the weird behaviour? I can try some Git-foo to try these changes again without merging master if that will help!

packages/components/src/Page/Page.tsx
```diff
@@ -59,10 +58,35 @@ export interface PageProps {
...
+   * Defaults to `false`.
+   */
+  readonly externalIntroLinks?: boolean;
+}
+
+interface PagePropsNoIntro extends PageFoundationProps {
+  readonly intro?: undefined;
+  readonly externalIntroLinks?: never;
+}
+export type PageProps = XOR<PagePropsWithIntro, PagePropsNoIntro>;
```
eddysims, line 83: This `PagePropsNoIntro` shouldn't be needed. We should be able to leave PageFoundationProps the way it was and just do something like:

```tsx
interface PageWithIntroProps extends PageFoundationProps {
  readonly intro: string;
  readonly externalIntroLinks?: boolean;
}

export type PageProps = XOR<PageFoundationProps, PageWithIntroProps>;
```

In this case we say that `PageFoundationProps` has an optional `intro` prop.

With the `XOR` we are saying that if the optional `externalIntroLinks` is used, the type will be `PageWithIntroProps` which will mean that `intro` is required.

eddysims (changes requested):

chris-at-jobber:
@irmalcol were you still thinking of moving this one ahead?

PR #561, opened 2021-06-10
feat(hooks): add useCollectionQuery and useIsMounted [JOB-32207]

jlapstra:
## Motivations

useCollectionQuery and useIsMounted are both hooks that live within two repositories: Jobber Online and Jobber Mobile.
Moving them to Atlantis will allow for easier edits to the hooks, and easier maintenance of the tests.

Communiverse has already reviewed all of this in pieces as it was landing. We're just looking for final feedback from the McWeb team.

There is a known BUG within useCollectionQuery caused by Apollo. It has been fixed within their release candidate and a workaround solution PR will be up soon. The link to the issue can be found here: https://github.com/apollographql/apollo-client/issues/7491#issuecomment-767985363

### Added

- useIsMounted hook added, along with tests and documentation
- useCollectionQuery hook added, along with tests and documentation
- Adds a `Configuration` object for the hooks library to configure things like Error Reporters

### Changed

- updates to package.json to allow for Apollo Client to be used

### Deprecated

- NA

### Removed

- NA

### Fixed

- NA

### Security

- NA

## Testing

http://localhost.test:3333/hooks/use-collection-query

4 items in the list
Fetch More loads 4 more
[ BUG IN USECOLLECTIONQUERY -- FIX PR coming soon ] Refresh will return all fetchedMore items. 

http://localhost.test:3333/hooks/use-is-mounted

Clicking the appropriate buttons will unmount or mount the component. This is a dependancy of useCollecitonQuery

---

Random photo of Atlantis


packages/hooks/src/useCollectionQuery/mdxUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+  loadingIndicator: {
+    flex: 1,
+    paddingBottom: 100,
+  },
+  actions: {
+    paddingTop: 20,
+  },
+};
+
+export const propsList = [
```
chris-at-jobber, line 84: Can we use the `<Props>` instead of this `List` for consistency with the rest of Atlantis' docs? Example from Button.mdx...

image

packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
```diff
@@ -0,0 +1,343 @@
...
+    refresh,
+    loadingRefresh,
+    nextPage,
+    loadingNextPage,
+    loadingInitialContent,
+  };
+}
+
+/**
+ * The following method uses an `output` variable, and indirectly modifies it through the `outputCollection` variable.
```
chris-at-jobber, line 215: Does this chunk of comments matter to the end users of this hook? If so I'd add it to the mdx, as we try to surface these kinds of trade-offs and considerations where possible ([example](https://atlantis.getjobber.com/components/disclosure#notes))

packages/hooks/src/useCollectionQuery/mdxUtils.ts
jlapstra: @chris-at-jobber I tried and TLDR; sounds like it is not supported for hooks as of yet. I would have LOVED to use this.

packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
jlapstra: @chris-at-jobber  These comments are more about maintenance in the future and explaining decisions made. They do not matter to the end users of this hook :)

packages/hooks/package.json
```diff
@@ -13,8 +13,19 @@
     "dist/*"
   ],
   "devDependencies": {
+    "@apollo/client": "^3.3.16",
```
hosseintalebi, line 16: I think this should be added to `dependencies`. The same for `"graphql"`, and `"graphql-tag"`.

packages/hooks/src/configure.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const Configuration: ConfigurationOptions = {
+  errorNotifier(message, error) {
+    console.error(message, error);
+  },
+};
+
+export function configure(options: ConfigurationOptions) {
+  Object.assign(Configuration, options);
+}
```
hosseintalebi, line 15: It doesn't look like we are using this anywhere.

packages/hooks/src/configure.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  });
+
+  it("should set the errorNotifier to the passed in configuration option", () => {
+    const myNotifier = jest.fn();
+    configure({
+      errorNotifier: myNotifier,
+    });
+
+    expect(Configuration.errorNotifier).toBe(myNotifier);
+  });
```
hosseintalebi, line 18: since `configure` has not been used anywhere, I wonder if we need this test.

packages/hooks/package.json
andrewjdmccann: @hosseintalebi It's already peer dependency, not an actual dependency of the library. Unless I'm misunderstanding peer dependencies.

packages/hooks/src/configure.ts
andrewjdmccann: @hosseintalebi It's not just a test utility. It's currently used in tests, but in both Jobber Online and Jobber-mobile we will need to use this to configure the `errorNotifier` to correctly notify the roll bar instances.

packages/hooks/src/configure.test.ts
andrewjdmccann: See this comment. https://github.com/GetJobber/atlantis/pull/561#discussion_r657291605

packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
```diff
@@ -0,0 +1,343 @@
...
+type GetNodeByPath<TSubscription> = (
+  data: TSubscription | undefined,
+) => Node | undefined;
+
+export function useCollectionQuery<TQuery, TSubscription = undefined>({
+  query,
+  queryOptions,
+  getCollectionByPath,
+  subscription,
+}: UseCollectionQueryArguments<
```
hosseintalebi, line 100: Apollo `useQuery` signature is:
```tsx
function useQuery(query: DocumentNode , options?: QueryHookOptions<TData, TVariables>)
```

Just to be consistent, should we move `query` out of the object?
```suggestion
export function useCollectionQuery<TQuery, TSubscription = undefined>(query: DocumentNode, {
  queryOptions,
  getCollectionByPath,
  subscription,
}: UseCollectionQueryArguments<
```

packages/hooks/src/configure.ts
andrewjdmccann: Each environment that uses hooks will configure an errorNotifier that notifies that environments error reporting system.

jlapstra: Hi @hosseintalebi ! `Change : Adds a Configuration object for the hooks library to configure things like Error Reporters` This Configuration object will be used by Jobber Online and Jobber Mobile to specify how to log errors and other configurations needed in future :) @andrewjdmccann Please correct me if I am incorrect!

packages/hooks/src/useCollectionQuery/__mocks__/queries.ts
```diff
@@ -0,0 +1,72 @@
+import gql from "graphql-tag";
```
seanhealy, line 1: This can be imported from Apollo.

```suggestion
import { gql } from "@apollo/client";
```

packages/hooks/src/useCollectionQuery/mdxUtils.ts
```diff
@@ -0,0 +1,148 @@
+import { ApolloClient, InMemoryCache } from "@apollo/client";
+import gql from "graphql-tag";
```
seanhealy, line 2:
```suggestion
import { ApolloClient, InMemoryCache, gql } from "@apollo/client";
```

packages/hooks/src/useCollectionQuery/useCollectionQuery.test.tsx
```diff
@@ -0,0 +1,345 @@
...
+  SubscriptionQueryType,
+  buildListRequestMock,
+  buildListRequestMockForNextPage,
+  buildSubscriptionRequestMock,
+  listQueryResponseMock,
+  setListQueryMockHasNextPage,
+  subscriptionQueryMock,
+  wait,
+  wrapper,
+} from "./__mocks__";
```
seanhealy, line 16: This doesn't quite fit mocks. Which are normally used to mock modules.
https://jestjs.io/docs/manual-mocks

It's totally cool to have utility methods like you have here but, in that case I don't think they fit with `__mocks__`

packages/hooks/src/useIsMounted/useIsMounted.mdx
```diff
@@ -0,0 +1,68 @@
...
+import { useEffect, useState } from "react";
+import { Button } from "@jobber/components/Button";
+import { useIsMounted } from "@jobber/hooks";
+
+# UseIsMounted
+
+`useIsMounted` is a generalized hook that upon initial loading triggers a
+`useLayoutEffect` that sets the `isMounted.current` to true when the component
+mounts. When the component unmounts, `useLayoutEffect` calls a cleanup function
+that sets `isMounted.current` to false.
```
seanhealy, line 17: I don't know that we need to dive into the implementation here so much as what is the effect of this hook and how it would be used.

packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
```diff
@@ -0,0 +1,343 @@
...
+  ]);
+
+  const nextPage = useCallback(() => {
+    if (loadingInitialContent || loadingRefresh || loadingNextPage) {
+      return;
+    }
+
+    const pageInfo = getCollectionByPath(data)?.pageInfo;
+
+    if (!pageInfo || !pageInfo.hasNextPage) {
```
seanhealy, line 146: Would `!pageInfo?.hasNextPage` work here?

packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
```diff
@@ -0,0 +1,343 @@
...
+    setLoadingNextPage(true);
+
+    fetchMore({
+      variables: {
+        cursor: pageInfo.endCursor,
+      },
+      updateQuery: (prev, { fetchMoreResult }) =>
+        fetchMoreUpdateQueryHandler(prev, fetchMoreResult, getCollectionByPath),
+    })
+      // tslint:disable-next-line:no-unsafe-any
```
seanhealy, line 159: No tslint needed here! 😄

```suggestion
```

packages/hooks/src/configure.ts
```diff
@@ -0,0 +1,15 @@
+export type ErrorNotifierFunction = (message: string, error: unknown) => void;
+
+export interface ConfigurationOptions {
```
seanhealy, line 3: Does this need to be exported?

packages/hooks/src/configure.ts
```diff
@@ -0,0 +1,15 @@
+export type ErrorNotifierFunction = (message: string, error: unknown) => void;
```
seanhealy, line 1: I'm thinking a bit about this configure piece here and how it fits into the greater systems in Atlantis. No direct comments now but more a heads up that I may have a followup or questions here. 😄

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
```diff
@@ -0,0 +1,143 @@
...
+  mapDataToListItem,
+  styles,
+  propsList,
+  subscriptionPropsList,
+} from "./mdxUtils";
+
+# UseCollectionQuery
+
+`useCollectionQuery` is a generalized hook that that fetches a collection of
+data. For use with any lists populated by graphQL collections.
```
seanhealy, line 25: Just thinking that it does more than just fetch. 😄 

```suggestion
`useCollectionQuery` is a generalized hook that that aids in the interaction of data collections. For use with any lists populated by graphQL collections.
```

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
```diff
@@ -0,0 +1,143 @@
...
+</Playground>
+
+## Subscription props
+
+<Playground>
+  {() => {
+    return <List items={subscriptionPropsList} />;
+  }}
+</Playground>
+
```
seanhealy, line 142: I think it would also be nice to talk about `useCollectionQuery`s return value to document what they provide / do.

```
{
    data,
    error,
    refresh,
    nextPage,
    loadingRefresh,
    loadingNextPage,
    loadingInitialContent,
}
```

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
```diff
@@ -0,0 +1,143 @@
...
+          </div>
+          <div style={styles.loadingIndicator}>
+            <h4>Refreshing Indicator</h4>
+            {loadingRefresh && <Spinner />}
+          </div>
+          <div style={styles.loadingIndicator}>
+            <h4>Fetching More Indicator</h4>
+            {loadingNextPage && <Spinner />}
+          </div>
+        </div>
```
seanhealy, line 97: It's a bit of an abuse of intention but, it might be easier to just use a `Button` component with the `disabled` prop here rather than a bunch of custom styling. Just thinking if we can strip out some of the boiler plate it can better keep the example focused on the workings of the hook itself.

@chris-at-jobber thoughts?

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
```diff
@@ -0,0 +1,143 @@
...
+        <Divider />
+        <div>
+          {items && (
+            <>
+              <h4>Star Wars Planets</h4>
```
seanhealy, line 102: https://atlantis.getjobber.com/packages-components-src-heading-heading

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
```diff
@@ -0,0 +1,143 @@
...
+            {loadingRefresh && <Spinner />}
+          </div>
+          <div style={styles.loadingIndicator}>
+            <h4>Fetching More Indicator</h4>
+            {loadingNextPage && <Spinner />}
+          </div>
+        </div>
+        <Divider />
+        <div>
+          {items && (
```
seanhealy, line 100: Might be easier to use a `List` here.

https://atlantis.getjobber.com/components/list

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
```diff
@@ -0,0 +1,143 @@
...
+  }}
+</Playground>
+
+## Props
+
+<Playground>
+  {() => {
+    return <List items={propsList} />;
+  }}
+</Playground>
```
seanhealy, line 133: You should be able to just do?

```suggestion
<List items={propsList} />
```

packages/hooks/src/useCollectionQuery/mdxUtils.ts
seanhealy: Since this isn't currently able to pull from the actual typings I think a markdown table in the MDX might be easier than what we have here. At least in that case the docs are closer to the file showing the docs.

https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables

packages/hooks/src/useCollectionQuery/mdxUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+        id: edge.node.id,
+        icon: "starFill",
+        iconColor: "green",
+        content: edge.node.name,
+      };
+    });
+    return items;
+  }
+  return undefined;
+}
```
seanhealy, line 69: I think if we use the List component we can probably get rid of this and just do a simple map inline. This is hiding away interactions that I think would be good for the consumer of the component to see and, map could handle pretty easily within the playground.

seanhealy (changes requested):
This is looking really great! Thanks so much for everything here. I do have some pondering to do around the Rollbar and Config thing there but in the mean time have some thoughts back your way. If you have any questions or thoughts on them lemme know happy to chat! 😃

packages/hooks/src/useCollectionQuery/useCollectionQuery.test.tsx
andrewjdmccann: Would you prefer a `__test__` directory next to the `mock` directory? Happy to do that, just figured it wans't worth the extra overhead.

packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
andrewjdmccann: @seanhealy: 
It certainly could and since you're maintianing this repo if you feel strongly I'll change it, but I think the above form is more readable and easier to reason around.

The reason is that `pageInfo` not existing and `hasNextPage` are two different states that *could* have two  different code paths. They happen to share a code path right now but that doesn't have to be the case.

`!pageInfo?.hasNextPage` leaves it up to the reader to decide if the auther of the code explicitly chose that `!pageInfo` should return `nil` or if it was an accident because `pageInfo` was optionally `null` and to get `hasNextPage` to work the auther just added `?`

packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
andrewjdmccann: So now that I think about it a little more, does it make sense to make the first argument just the query? Unlike `useQuery` we're doing a *alot* more than just `query`ing. We're querying, fetching more, subscribing to more, and in the future `search`ing. As a result would it make more sense to make sure each argument is labeled when it's used because the set of features that this will do is quite a bit larger than `useQuery`?

andrewjdmccann: @hosseintalebi Thoughts?

andrewjdmccann:
@seanhealy Have you had a chance to think any more about the `Configuration` side of this? It's definitely the most impactful part of this work and we'd want to get it right (because it sets a precedent for all of the hooks package, and for the components package as well).

packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
andrewjdmccann: Thoughts?

packages/hooks/src/useCollectionQuery/useCollectionQuery.test.tsx
andrewjdmccann: I'll go with calling it `__test_utils__` for now unless I hear back about a better name.

packages/hooks/src/configure.ts
andrewjdmccann: If it's an input type I like it being exported mostly for the sake of people that maybe implement it on a previous line rather than inline.

```
  const myOptions: ConfigurationOptions = {
    //...
  };
  
  Configuration.configure(myOptions);
```

andrewjdmccann: If the type was only used internally I wouldn't export it.

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
jlapstra: @seanhealy Sorry -- I am unclear on what you mean. I am using a List here.

packages/hooks/src/useCollectionQuery/mdxUtils.ts
jlapstra: Hello @seanhealy! Could you clarify "At least in that case the docs are closer to the file showing the docs."

Personally I really like the list. I believe it is closer to the generated List of props. Within their props list, the descriptions are hidden and need to be tapped to open. That is why I included them as list caption (smaller text, less focus). 

The most important information is listed in a bold font for each item. I was also able to add a link to queryOptions if folks would like to find more information they can click on that list item and be directed to more info.

I am also hoping that at some point they will build out Props for hooks, and then this all goes away.

@chris-at-jobber Do you have thoughts here?

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
jlapstra: @seanhealy This is an interesting idea. I am curious what @chris-at-jobber thinks. Would these buttons not be functional other then being disabled / enabled?

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
jlapstra: Great idea! @seanhealy

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
chris-at-jobber: I am not sure I understand what is being suggested, or why 😅  What would the buttons replace? Why would they be disabled?

jlapstra: @chris-at-jobber My interpretation is that there would be three buttons that did nothing when clicked. These buttons would be disabled when not in a loading state, and enabled if they were in a loading state. The code becomes much simpler, and does not need styling. I actually really like this change! I have recorded a demo for you :) 


https://user-images.githubusercontent.com/6353164/123437099-52c29e00-d59d-11eb-8773-d040ffa8aa40.mp4

packages/hooks/src/useCollectionQuery/mdxUtils.ts
jlapstra: We can totally do that. Did you know the playground does not like the `?.` syntax? However, I can just ensure edges exist before the mapping.

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
chris-at-jobber: Hmmm

I don't think I understand what those buttons are actually communicating, and they flicker in and out of "active" to quickly to be of any real help. I thought the previous approach of "header + spinner" was being used to visualize that different loading paths are hit in the code for different types of loading, but if that's not what we're trying to do, maybe just remove the whole section? It seems like the list content updating when you trigger the actions is good enough.

packages/hooks/src/useCollectionQuery/useCollectionQuery.mdx
seanhealy: This is just me not reading properly.

seanhealy (approved):

PR #562, opened 2021-06-10
feat(hooks): add useCollectionQuery and useIsMounted [JOB-32207]

jlapstra:
## Motivations

Both Jobber Online and Jobber Mobile use useCollectionQuery for 2WSMS. The version in Jobber Online, however did not have tests as it was implemented in Jobber Online during a big bug fix. It was always the intention to move useCollectionQuery to Atlantis, no time like now.

## Changes

To make useCollectionQuery work, we needed to also move over useIsMounted.
Tests have been written.

### Added

Both hooks can now be updated from a shared component in Atlantis

### Changed

The functionality should not change.

### Deprecated

NA

### Removed

NA

### Fixed

NA

### Security

NA

## Testing

Tests should work. Integrations with Jobber Online and Mobile should be tested completely.

---

Random photo of Atlantis

### DIFF
--- a/packages/hooks/src/useCollectionQuery/__mocks__/queries.ts
+++ b/packages/hooks/src/useCollectionQuery/__mocks__/queries.ts
@@ -1,0 +1,72 @@
+import gql from "graphql-tag";
+
+export const LIST_QUERY = gql`
+  query ConversationMessages($cursor: string) {
+    conversation(id: "MQ==") {
+      smsMessages(first: 1, after: $cursor) {
+        edges {
+          node {
+            __typename
+            id
+          }
+        }
+        nodes {
+          __typename
+          id
+        }
+        pageInfo {
+          endCursor
+          hasNextPage
+        }
+        totalCount
+      }
+    }
+  }
+`;
+
+export interface ListQueryType {
+  __typename?: "Query";
+  conversation?: {
+    __typename?: "Conversation";
+    smsMessages: {
+      edges: Array<{
+        __typename?: "SMSMessageConnection";
+        node: {
+          __typename?: "SMSMessage";
+          id: string;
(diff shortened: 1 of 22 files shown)


packages/hooks/src/useIsMounted/useIsMounted.ts
```diff
@@ -0,0 +1,30 @@
+import { useLayoutEffect, useRef } from "react";
+
+/**
+ * Why does this work?
```
jlapstra, line 4: Should this go in an MDX file? @andrewjdmccann

andrewjdmccann: It should but I think it should also be in this file.

packages/hooks/src/useCollectionQuery/__mocks__/queries.ts
```diff
@@ -0,0 +1,72 @@
+import gql from "graphql-tag";
+
+export const LIST_QUERY = gql`
```
jlapstra, line 3: is this actually the `list query` or the `conversation query`? @andrewjdmccann

jlapstra:
@andrewjdmccann This looks good! I do believe it is the `conversation query / conversation query type` that we want here, instead of the `list query / list query type`.

packages/hooks/src/useCollectionQuery/__mocks__/queries.ts
andrewjdmccann: This was a very intentional change. While this particular implementation is a Conversation query, the "interface" is that all we really need is for it to return a list. We don't care about anything else.

jlapstra:
LGTM 👍

PR #563, opened 2021-06-11
fix: Text / Typography don't always respect maxLines [JOB-32400]

joelwarrington-jobber:
## Motivations

In certain cases - it appears that the `maxLines` prop on `<Text/>` or `<Typography/>` was ignored.
See following from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp):
```
The -webkit-line-clamp CSS property allows limiting of the contents of a block container to the specified number of lines.

It only works in combination with the display property set to -webkit-box or -webkit-inline-box and the -webkit-box-orient property set to vertical.
```

The `-webkit-box-orient: "Vertical"` was missing, and has been added.

### Fixed

- Fixed the `maxLines` prop on Text & Typography to be slightly more consistent with popular browsers.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Typography/Typography.test.tsx
+++ b/packages/components/src/Typography/Typography.test.tsx
@@ -205,6 +205,7 @@ it("should add textTruncate class when numberOfLines property is passed", () => 
       className="base regular textTruncate"
       style={
         Object {
+          "WebkitBoxOrient": "vertical",
           "WebkitLineClamp": 3,
         }
       }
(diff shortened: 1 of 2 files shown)


joelwarrington-jobber:
Oddly enough - we do have this property in the `truncate.css` file, but am not getting that property in the imported package.

Using version `^2.35.1` and seeing this as the CSS for truncate: 
```css
._3RPdE_X9LPA4hEoBHeEyKm {
   display: -webkit-box !important;
   overflow: hidden;
}
```

eddysims (approved):

PR #578, opened 2021-06-22
feat(components): Refactor FormField to fix some bugs

eddysims:
## Motivations

We were finding issues with the  `InputText` and mixing controlled and uncontrolled inputs. Since the `InputText` uses the `FormField` component, we have decided to do a refactor of the `FormField` component.

For this change, we are forcing all inputs to be controlled inputs through the use of the `Controller` component provided by `react-hook-form`. 

### Added

- 

### Changed

- Refactored the `FormField` component

### Deprecated

- 

### Removed

- 

### Fixed

- Allows controlled and uncontrolled inputs to be used together.

### Security

- 

## Testing

---

Random photo of Atlantis


joelwarrington-jobber (commented):
This refactor is 🌶️ 🎆 !

From a quick glance - didn't have any concerns and it looks soo much better! Really appreciate taking this on.

chris-at-jobber:
We have a ["controlled" section on InputText docs](https://atlantis.getjobber.com/components/input-text/#controlled), does that need to be updated based on any of this work?

eddysims:
> We have a ["controlled" section on InputText docs](https://atlantis.getjobber.com/components/input-text/#controlled), does that need to be updated based on any of this work?

@chris-at-jobber no i don't think so. The usage stays the same. We still allow consumers to use the input as controlled or uncontrolled, however we do all the work internally to make it a controlled component.

packages/components/src/FormField/FormField.tsx
```diff
@@ -1,426 +1,216 @@
...
-
-  const error = name && errors[name] && errors[name].message;
+  /**
+   * Generate a name if one is not supplied, this is the name
+   * that will be used for react-hook-form and not neccessarily
+   * attached to the DOM
+   */
+  const [controlledName] = useState(
+    name ? name : `generatedName--${identifier}`,
+  );
```
hosseintalebi, line 62: Can we move this to a hook? Here is how we have done it on the mobile side:
https://github.com/GetJobber/jobber-mobile/blob/c0f73a4957d898a3c2671b9d93faf158320b90e7/src/hooks/useFormController/useFormController.ts

packages/components/src/FormField/FormFieldTypes.ts
```diff
@@ -0,0 +1,163 @@
...
+   * Show an error message above the field. This also
+   * highlights the the field red if an error message shows up.
+   */
+  readonly validations?: ValidationRules;
+
+  inputRef?: RefObject<
+    HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement
+  >;
+
+  actionsRef?: RefObject<FieldActionsRef>;
```
seanhealy, line 162: Thinking we should probably `Exclude<>` this from the props on `InputText` and the others. Or, perhaps it would be better to switch the `Exclude<>` there to a whitelist so we don't get surprise visitors like this showing up.

packages/components/src/FormField/FormField.tsx
```diff
@@ -1,426 +1,216 @@
...
+
+          function setAutocomplete(autocompleteSetting: boolean | string) {
+            if (autocompleteSetting === "one-time-code") {
+              return "one-time-code";
+            }
+
+            return autocompleteSetting ? undefined : "autocomplete-off";
+          }
+        }}
+      />
```
hosseintalebi, line 209: The jsx here is long and it's kind of using the renderProps technique that was kind of replaced by hooks. Feel free to take a look at here and see how we do it on the mobile side:
https://github.com/GetJobber/jobber-mobile/pull/952/files

packages/components/src/FormField/FormField.tsx
```diff
@@ -1,426 +1,216 @@
...
-
-  inputRef?: RefObject<
-    HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement
-  >;
-}
-
-/**
- * Max statements is disabled for this component due the event handlers.
- */
-// eslint-disable-next-line max-statements
```
seanhealy, line 178: 🎉🎉🎉🎉

packages/components/src/FormField/FormField.tsx
```diff
@@ -1,426 +1,216 @@
...
-    size && styles[size],
-    align && styles[align],
-    disabled && styles.disabled,
-    maxLength && styles.maxLength,
-    {
-      [styles.miniLabel]:
-        (hasMiniLabel || type === "time" || type === "select") && placeholder,
-      [styles.invalid]: invalid || error,
+  useImperativeHandle(actionsRef, () => ({
+    setValue: newValue => {
```
seanhealy, line 76: Now that we have the `useEffect` above this I'm pondering if we still need this. I feel like the answer is we do but I'm not 100% sure… 🤔

packages/components/src/FormField/FormField.tsx
```diff
@@ -1,426 +1,216 @@
...
+                      type={type}
+                      maxLength={maxLength}
+                      max={max}
+                      min={min}
+                      ref={inputRef as MutableRefObject<HTMLInputElement>}
+                      {...textFieldProps}
+                    />
+                    {loading && <FormSpinner />}
+                  </>
+                );
```
seanhealy, line 155: Part of me feels like splitting the internals of this render callback off as a separate component would be nice. Thoughts? It does lean on a lot of scoped things from above but it would simplify this a decent amount too.

packages/components/src/InputNumber/InputNumber.test.tsx
```diff
@@ -6,7 +6,7 @@ import { InputNumber, InputNumberRef } from ".";
 afterEach(cleanup);
 
 it("renders an input type number", () => {
-  const tree = renderer.create(<InputNumber value="123" />).toJSON();
+  const tree = renderer.create(<InputNumber value={123} />).toJSON();
```
seanhealy, line 9: This should have always required a number but, did it get more strict in practice as the result of these changes?

packages/components/src/InputNumber/InputNumber.test.tsx
```diff
@@ -31,41 +29,7 @@ it("renders an input type number", () => {
         onFocus={[Function]}
         onKeyDown={[Function]}
         type="number"
-        value="123"
-      />
-    </div>
-  `);
-});
-
-it("renders an error", () => {
```
seanhealy, line 40: Why remove this? Just because it's testing testing a bit beyond this component's concerns?

packages/components/src/InputTime/InputTime.test.tsx
```diff
@@ -20,13 +20,10 @@ it("renders a InputTime", () => {
       <label
         className="label"
         htmlFor="123e4567-e89b-12d3-a456-426655440001"
-      >
-         
-      </label>
+      />
```
seanhealy, line 23: Any idea why the format of all these `label`s changed? I don't think it's a problem just wondering. 😊

packages/components/src/InputTime/InputTime.test.tsx
```diff
@@ -54,18 +51,16 @@ it("renders an initial time when given 'defaultValue'", () => {
...
       <input
         className="formField"
-        defaultValue="11:23"
         id="123e4567-e89b-12d3-a456-426655440002"
         onBlur={[Function]}
         onChange={[Function]}
         onFocus={[Function]}
         onKeyDown={[Function]}
         type="time"
+        value="11:23"
```
seanhealy, line 63: This changed from `defaultValue` to `value` I think that's expected but just wanted to confirm since it's buried in a snapshot here.

seanhealy (changes requested):
Few thoughts and questions for you.

I'm also thinking about the rollout of these. On paper it should be pretty safe but, given the size of the change and how often inputs are used I wonder if we should do some deliberate spot checking when pulling it into other projects.

packages/components/src/InputNumber/InputNumber.test.tsx
eddysims: This was removed as it was testing the old way that errors were presented. This snapshot didn't include an error at all. 

I was going to rewrite it, however there were 2 more useful tests further down that tested the error states

packages/components/src/InputNumber/InputNumber.test.tsx
eddysims: This has been like this for a long time. I have noticed it before, but never had a reason to fix it.

packages/components/src/InputTime/InputTime.test.tsx
eddysims: Yep, my understanding is that is what we want as defaultValue will essentially just transform into the value now.

packages/components/src/InputTime/InputTime.test.tsx
eddysims: I also wondered about this. Im not sure why this changed... I saw this and didn't think it was a big deal. I can investigate more if you would like?

packages/components/src/FormField/FormField.tsx
eddysims: This is something i struggled on.

When I pulled this out into its own component I was having issues with the typing of it. The reason being is that we found some issues with the types around the `Controller` component. 

I was unable to type the `{...rest}` spread when pulling this into its own component. 

I did try to make it a little cleaner by bringing the `Spinner` and `Icon` into their respected return statements and thought that helped a bit.

packages/components/src/FormField/FormField.tsx
eddysims: I think that we do as this still fixes the insertion issue where the cursor goes to the end of the input.

packages/components/src/FormField/FormFieldTypes.ts
eddysims: I have added this to the Exclude on all of the Input components. I have also added a backlog item to switch this from an `Exclude` to a `Pick`. I think that is a little 👻 to do in this PR. I would prefer to not change the way that all of the types work for each component at once.

packages/components/src/FormField/FormField.tsx
eddysims: @hosseintalebi we do require this `identifier` to be the same in the `FormLabel` and the input. do you think that it would still be better to do something like:

```ts
const [identifier] = useState(uuid.v1());
const [controlledName] = useControlledName(name, identifier);

// hook

function useControlledName(name: string | undefined, identifier: string) {
   return name ? name : `generatedName--${identifier}`
}
```

packages/components/src/FormField/FormField.tsx
eddysims: @hosseintalebi yeah, I agree. I tried using the hook at first and struggled with it. To be honest I don't really remember why. 

One thing to note is that we found some typing issues on `v6` of react hook forms that could have been where I got stuck.

Do you think it is worth investigating using the hook?

packages/components/src/InputTime/InputTime.test.tsx
seanhealy: I wouldn't worry about it was more just curious. 😂

packages/components/src/FormField/FormField.tsx
seanhealy: Yea, I can see that. Let's leave it as is for now.

packages/components/src/FormField/FormField.tsx
seanhealy: 👍

packages/components/src/FormField/FormField.tsx
hosseintalebi: @eddysims We have made some progress on the mobile side to only use the hooks. Let's get back to it after we have the mobile side ready.

seanhealy (approved):
This looks great! So much nicer than what was there before.

PR #580, opened 2021-06-22
feat(components): Implement Loader component

joelwarrington-jobber:
## Motivations
- Implemented the `<Loader />` component to address determinate and non-determinate loading and provide accessibility features for assistive technology when content is changing!
- The Loader component is intended to overlap loading content if there is a skeleton which can be displayed

### Added

- Added the Loader component to address and display determinate and non-determinate loading of sub-components

---

Random photo of Atlantis


chris-at-jobber:
@joelwarrington-jobber Thanks for putting this up! A few early thoughts...

I wonder if this is more of an extension of [Spinner](https://atlantis.getjobber.com/components/spinner) than a new component? The [SG-1 version of Spinner](http://localhost:4001/components/shared/spinner-loader/) works the way your proposal describes by default.

Currently loading overlays (in Jobber, using SG-1) use a white overlay to "dim" the element, rather than the modal-style overlay used in this proposal:
image

The progress variant looks like what we do in FormatFile (which could also probably use the "white overlay" approach) - I think rather than "progress" could we call this "determinate" and "indeterminate"? Then it's more about what _type_ of loading you're doing rather than what the UI element for each is.

joelwarrington-jobber:
> @joelwarrington-jobber Thanks for putting this up! A few early thoughts...
> 
> 
> 
> I wonder if this is more of an extension of [Spinner](https://atlantis.getjobber.com/components/spinner) than a new component? The [SG-1 version of Spinner](http://localhost:4001/components/shared/spinner-loader/) works the way your proposal describes by default.
> 
> 
> 
> Currently loading overlays (in Jobber, using SG-1) use a white overlay to "dim" the element, rather than the modal-style overlay used in this proposal:
> 
> image
> 
> 
> 
> The progress variant looks like what we do in FormatFile (which could also probably use the "white overlay" approach) - I think rather than "progress" could we call this "determinate" and "indeterminate"? Then it's more about what _type_ of loading you're doing rather than what the UI element for each is.

I agree with the determinate/intermediate approach. 

Do you think it's worth continuing this proposal approach and cleaning up the styling?

I'm open to working on fleshing out the component if there's value in it. I'm mostly wanting this for DevCenter work (where SG-1 doesn't apply) but think this could be used elsewhere as well.

joelwarrington-jobber:
or would it be best to apply this concept to our spinner component and drop the determinate loading portion (or do that portion in the progress bar component)

eddysims:
@chris-at-jobber @joelwarrington-jobber I had chatted about this a long time ago with someone. Im not sure who, but I think that this is the correct approach. The reason is that unlike SG-1, we are unable to add a `loading` class to components. 

I think the idea would be that the `Loading` component would use the spinner when `isLoading` (or whatever the prop is called) is true. 

I think that this could use more description in the mdx to explain what the component is doing and how it will be used.

chris-at-jobber:
OK I'm good with that. I just checked the web app and we're only using `<Spinner>` in one place, and it uses the `inline` prop which I think doesn't break this.

There's also an instance of someone making a one-off `Spinner` that behaves exactly how this proposed `Loading` component would work (wraps a child element and throws an overlay with a spinner while loading), so we could clean that up as we roll this out.

Does something like this make sense?

### name
Loading

### props
| name | type | description | default |
| inline | boolean | removes overlay and has the component render in the flow of the layout | false |
| determinate | boolean | shows spinner when false, progress bar when true | false |
| currentValue | number | current value, only when determinate = true | 0? |
| maxValue | number | max value, only when determinate = true | 100 |

joelwarrington-jobber:
The only question/clarification I would have is for the inline prop, how is this intended to be used?

Other than that, agree with the comments.

chris-at-jobber:
https://atlantis.getjobber.com/components/spinner#inline for example!

joelwarrington-jobber:
@chris-at-jobber / @eddysims - updated the component and added some examples.

@chris-at-jobber would love to have a chat around design considerations with this before getting too deep into the testing/documentation side of things. Once we have that chat - I can update the documentation in `.mdx` to add thoughts in.

chris-at-jobber:
@joelwarrington-jobber the color updates have made their way into `/design` 🌈

joelwarrington-jobber:
@chris-at-jobber can you suggest a semantic PR title?

joelwarrington-jobber:
I'm not sure if for a proposal a full PR description is needed?

chris-at-jobber:
@joelwarrington-jobber I'll update the title! We've got a semantic PR [title generator](https://atlantis.getjobber.com/pull-request-name-generator) that's linked in the PR description template for future. I'd consider a component proposal a `feat(components)` change.

chris-at-jobber:
For a proposal I think the Motivations and Added will usually be the key ones - you can delete the other points out if they aren't relevant. (In some instances you might be updating an existing component, in which case you could keep the "changed" section.)

packages/components/src/Loader/Loader.css
```diff
@@ -0,0 +1,21 @@
+.loader {
+  position: relative;
+  padding: var(--space-base);
```
chris-at-jobber, line 3: I don't think we want this padding here. It's collapsing the item it overlays which is going to cause layout jumping when the loading state changes:

image
image

packages/components/src/Loader/Loader.tsx
```diff
@@ -0,0 +1,87 @@
...
+    [styles.overlay]: !inline,
+    [styles.determinate]: determinate,
+  });
+
+  return (
+    <div ref={ref as RefObject<HTMLDivElement>}>
+      {loading ? (
+        <div aria-busy="true" aria-live="polite" className={classname}>
+          {children}
+          <div>
```
chris-at-jobber, line 73: Can we give this div a classname so that we don't need to do the nested selectors in the CSS? :last-child type selectors are quite brittle

joelwarrington-jobber:
> @joelwarrington-jobber I'll update the title! We've got a semantic PR [title generator](https://atlantis.getjobber.com/pull-request-name-generator) that's linked in the PR description template for future. I'd consider a component proposal a `feat(components)` change.

Ahhh - I've used that for other PRs but wasn't clear to use it for a proposal!

joelwarrington-jobber:
Going to be updating the description!

packages/components/src/Loader/Loader.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+.overlay > div:last-child {
+  display: flex;
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background-color: var(--color-overlay--dimmed);
```
chris-at-jobber, line 18: overlay--dimmed is not working because I used the wrong hyphenation syntax in the token names. Fix [here](https://github.com/GetJobber/atlantis/pull/599)

chris-at-jobber (changes requested):
Few small things (one of them is my bad, link to a fix included)

packages/components/src/Loader/Loader.css
joelwarrington-jobber: merging in the fix doesn't resolve it, because I think components uses a generated css file, going to wait until that fix is in :)

packages/components/src/Loader/Loader.tsx
joelwarrington-jobber: Added class name in ffa84d8

packages/components/src/Loader/Loader.css
joelwarrington-jobber: Updated the padding here: ffa84d8

it was added for the progress bar, so it doesn't touch both sides, but have moved it so that it's margin in the loader indicator component

packages/components/src/Loader/Loader.tsx
```diff
@@ -0,0 +1,87 @@
...
+   * Removes the overlay and has the component render in the flow of the layout
+   * @default false
+   */
+  readonly inline?: boolean;
+
+  /**
+   * Shows a spinner when false, progress bar when true.
+   * @default false
+   */
+  determinate?: boolean;
```
chris-at-jobber, line 20: The props table is showing these differently than I'd expect:

image

Is it related to the interfaces being defined below?

chris-at-jobber:
While `Loader` is wrapping an element, the contents of that element should not be accessible to a screen-reader as we're assuming we are awaiting content (the same reason we use the overlay to prevent cursor interactions). I'm able to enter the loading content area and read it, and the Loading alert element is only announced to me after I progress through the content.

Expected: user cannot interact with content inside of `Loader` while `loading` is true.

image

chris-at-jobber (changes requested):
@joelwarrington-jobber looking good! Some accessibility feedback, and a question about the props table that @eddysims is probably better equipped to review.

packages/components/src/Loader/Loader.tsx
joelwarrington-jobber: It looks like the Props of component doesn't deal well with union types. @eddysims what do you think is the best way to communicate the props, as I feel as though the union type works well in code editors but not well with this static prop table

joelwarrington-jobber:
Requested review from McWeb as Chris is on vacation

packages/components/src/Loader/Loader.mdx
```diff
@@ -0,0 +1,210 @@
...
+  What type of content does this component present (documents, text, images, other components)?
+  What is the recommended, or expected, amount of content? What is the components’ behaviour when
+  the provided content diverges from what is expected (image aspect ratio, amount of text, video
+  vs audio, filesize)? Are there any important references to the [Product Vocabulary](https://atlantis.getjobber.com/guides/product-vocabulary)
+  that should be made regarding this component? What happens when content is unavailable or unreliable
+  (poor connection, empty states, flaky GPS data?)
+-->
+
+<!--
+## Accessibility
```
chris-at-jobber, line 163:
```suggestion
## Accessibility

The Loading component uses the `aria-busy=”true” aria-live=”polite”` HTML attributes to convey to assistive technology that the region is loading.

Assistive technology, keyboards, and mouse cursors should not be able to interact with the content inside of Loading while it is awaiting new data/content. This is achieved using `aria-hidden=”true”` on the wrapper element to remove child elements from the accessibility tree while loading. Screen-readers should only receive the information that the area is “busy” as outlined above, while the dimmed interface and loading indicators signify the same status visually.
```

chris-at-jobber: (and I guess un-comment it out so it shows up!)

chris-at-jobber (approved):
This looks good to me!

@eddysims / @deepvishwas / @adambobadam would one of you mind doing a code review? 🙏

eddysims (changes requested):
@joelwarrington-jobber a few changes here. Feel free to reach out if any of them are confusing or dont make sense.

joelwarrington-jobber:
@eddysims I merged master into this branch to resolve conflicts and started getting significantly more lint errors which weren't caused by the work in this PR, do you know what's caused that?

hosseintalebi:
@chris-at-jobber @eddysims @joelwarrington-jobber I think the name `Loader` can be confusing here since it's the same name we use for the `.loader.tsx` files for the loader components of the view. What do you think if we call this `Sprinner`?

chris-at-jobber:
@hosseintalebi we already have a [Spinner](https://atlantis.getjobber.com/components/spinner) in Atlantis (it's actually used by the Loader), and this component also has a version that does not display a Spinner. Would `LoadingWrapper` perhaps be more clear? Part of the goal here was to name it more by what it's doing rather than the specific UI elements it's rendering.

joelwarrington-jobber:
@hosseintalebi I will defer naming to @chris-at-jobber, but if you have any suggestions other than `<Spinner />` would like to hear it!

joelwarrington-jobber:
going to have to also spend the time to merge master in and fix lint issues, which are not introduced by this PR but seem to have started after the first merge... time to investigate 🔬

joelwarrington-jobber:
Work in this PR will be added in separate PRs which tackle each component individually (Spinner, ProgressBar)

See #649

PR #581, opened 2021-06-22
docs: Add skipnav to `Layout`

chris-at-jobber:
## Motivations

It's a Bad Time™ to use your keyboard to make your way through Atlantis, if you have to tab through all of the sidenav items. As a general accessibility practice, it's important to allow keyboard users to skip main navigation in favour of the main content.

This adds a skipnav link in keeping with [WebAIM guidance.](https://webaim.org/techniques/skipnav/)

## Changes

Added a skipnav link to the Layout module.

### Added

- There's now a link that jumps the user to the main content of a page in Atlantis
- If you are on a less-than-wide screen, the link will also close the menu
- if you are on a wide+ screen, the link does not close the menu (because the menu should not ever close on wide screens)

### Changed

- The `content` div in Layout now has an id of `mainContent` to act as a target for the skipnav link

## Testing

Use the `tab` key on pageload in Atlantis - the skipnav link should appear before any other elements receive focus (other than the `close menu` button on smaller screens)

On wide screens, the skipnav link should be full-width to ensure that it is immediately noticeable.

On screens where the menu has an overlay, the skipnav link should be constrained to the sidenav "logo" area. 

---

Random photo of Atlantis

### DIFF
--- a/packages/docz-tools/src/gatsby-theme-docz/components/Layout/Layout.module.css
+++ b/packages/docz-tools/src/gatsby-theme-docz/components/Layout/Layout.module.css
@@ -73,6 +73,7 @@
   transition: all 0.2s ease;
 
   @media (--wide-screens) {
+    width: calc(100% - var(--table-of-contents-width) - var(--space-base));
     box-shadow: var(--shadow-focus);
   }
 }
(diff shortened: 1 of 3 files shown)


chris-at-jobber:
### Screenshots
image
image
image

eddysims:
@chris-at-jobber if you are on an extra wide screen, there is a `z-index`ing issue.


image

chris-at-jobber:
@eddysims good catch! Adjusted width to account for Table of Contents column when that's visible

eddysims (approved):

PR #589, opened 2021-07-06
feat(hooks): Update useCollectionQuery to handle search terms [32867]

tyler-santos:
## Motivations

- We want to be able to use useCollectionQuery with search and a subscription, if someone is searching for something, we don't want new subscription data to come in and mess up those results. This fixes that problem. 

### Changed

- Changed useCollectionQuery so that it does not pull in new data when a user is searching for content

## Testing
- Specs were written to specifically test this situation.

---

Random photo of Atlantis

### DIFF
--- a/packages/hooks/src/useCollectionQuery/useCollectionQuery.test.tsx
+++ b/packages/hooks/src/useCollectionQuery/useCollectionQuery.test.tsx
@@ -382,7 +382,6 @@ describe("#subscribeToMore", () => {
       // Wait for subscription
       await act(() => wait(200));
 
-      console.log(result?.current);
       expect(
         result?.current?.data?.conversation?.smsMessages?.edges?.length,
       ).toBe(1);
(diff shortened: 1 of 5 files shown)


jlapstra:
@tyler-santos @andrewjdmccann Do you think we should modify the mdx at all to mention search terms? Do we need to mention how subscriptions will work with a variable search_term?

tyler-santos:
> @tyler-santos @andrewjdmccann Do you think we should modify the mdx at all to mention search terms? Do we need to mention how subscriptions will work with a variable search_term?

I'm not sure if we need to mention anything, I think it is something we want to "just work" and not actually call out. However that is just my 2 :cents

andrewjdmccann:
I’m leaning to not include it for now as well

packages/hooks/src/useCollectionQuery/useCollectionQuery.test.tsx
```diff
@@ -342,4 +362,30 @@ describe("#subscribeToMore", () => {
...
+        },
+      );
+
+      // Wait for initial load
+      await act(waitForNextUpdate);
+
+      // Wait for subscription
+      await act(() => wait(200));
+
+      console.log(result?.current);
```
andrewjdmccann, line 385: Accidental console.log left in

tyler-santos: thank you!

andrewjdmccann (approved):

PR #590, opened 2021-07-09
feat(design): add active surface and dimmed overlay colors

chris-at-jobber:
## Motivations

Reflecting some more common surface and overlay patterns that I picked up on when rolling semantic colours out across Jobber. Also updates the docs to have color swatches associated with the usage guidelines, albeit in a rough fashion that could potentially be componentized in the future.

## Changes

Some new colours! Some new swatches in the docs!

### Added

- color-surface--active extends the surface--hover pattern
- color-overlay--dimmed extends the overlay pattern, but for "dimming" temporarily-non-interactive (but not disabled) UI

### Changed

- color-white now consumes an rgb value a la color-greyBlue for use in color-overlay
- components using surface--active now use them
- color docs have HTML-based color swatches
- split out accessibility in docs as its' own section, vs as a Banner

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/List/ListItem.css
+++ b/packages/components/src/List/ListItem.css
@@ -48,7 +48,7 @@
 /* States */
 
 .isActive {
-  background-color: var(--color-green--lightest);
+  background-color: var(--color-surface--active);
 }
 
 /* TODO: Move truncate in typography */
--- a/packages/design/src/Colors.mdx
+++ b/packages/design/src/Colors.mdx
@@ -86,7 +86,7 @@ The default color used for interactive elements.
   background-color:var(--color-destructive)"
 ></div>
 
-Use to signfify that an interaction will destroy something in the users’ account
+Use to signify that an interaction will destroy something in the users’ account
 or workflow.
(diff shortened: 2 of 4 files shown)


chris-at-jobber:
Review from @lorborb :

> 'Signify' under 'Destructive' is misspelled
Wonder if, in a v2 it might be helpful to state what kinds of text colour should be paired with each interactive colour to achieve 4.5:1, I could see this being ambiguous when it comes to cancelled/disabled interactions
'contrast' misspelled under 'Surface--Reverse'
Should 'Overlay' include an opacity property?
Border includes the text 'Most of our borders should use this color for...... maybe we could state a hex code or something to make sure it's precise or maybe add a link to the 'Swatch list' OR (if we want to keep specifics out of it) maybe just describe as 'Use border when ———— for subtle definition.

chris-at-jobber:
@lorborb updated the content, let me know how this looks 👀

lorborb (approved):
much thorough, yes, good.

PR #597, opened 2021-07-23
feat(components): add breadcrumb to `Page`

chris-at-jobber:
## Motivations

A few one-off instances of "back button" breadcrumbs have been used alongside `Page` - a good indicator that it's time to incorporate this mechanism into Page.

## Changes

Adds a breadcrumb to Page that will jump up one level in the application's navigation directory.

### Added

- New Button at the top of Page
- `breadcrumb` property that sets the breadcrumb label and determines if it exists

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -20,6 +20,12 @@ export interface PageProps {
   readonly intro?: string;
 
   /**
+   * Label for the "back" button when Page is nested.
+   * The breadcrumb button follows a relative URL to jump up one level.
+   */
+  readonly breadcrumb?: string;
+
+  /**
    * Title of the page.
    */
   readonly title: string;
@@ -61,6 +67,7 @@ export interface PageProps {
 
 // eslint-disable-next-line max-statements
 export function Page({
+  breadcrumb,
   title,
   intro,
   subtitle,
@@ -107,6 +114,15 @@ export function Page({
     <div className={pageStyles}>
       <Content>
         <Content>
+          {breadcrumb && (
+            <Button
+              icon="backArrow"
+              url="../"
+              label={breadcrumb}
+              type="tertiary"
+              size="small"
+            ></Button>
+          )}
           <div className={titleBarClasses} ref={titleBarRef}>
             <div>
               <Heading level={1}>{title}</Heading>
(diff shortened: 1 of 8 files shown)


packages/components/src/Page/Page.tsx
```diff
@@ -107,6 +114,15 @@ export function Page({
     <div className={pageStyles}>
       <Content>
         <Content>
+          {breadcrumb && (
+            <Button
+              icon="backArrow"
+              url="../"
```
joelwarrington-jobber, line 120: For added flexibility, would we be able to have 2 additional properties of a breadcrumb?


```typescript
interface breadcrumb {
  label: string;
  /* Default = "../" */
  url?: string;
  onClick?(): void;
}
```

If the `onClick` property is passed, the url is ignored

packages/components/src/Page/Page.tsx
```diff
@@ -107,6 +114,15 @@ export function Page({
...
+              label={breadcrumb}
+              type="tertiary"
+              size="small"
+            ></Button>
+          )}
```
joelwarrington-jobber, line 125: if multi-level breadcrumbs were to be a possibility, we could make it so that the breadcrumb type is an array, if the length is 1, use the button style and then some other type of style for multiple?

```typescript
/**
   * Label for the "back" button when Page is nested.
   * The breadcrumb button follows a relative URL to jump up one level.
   */
  readonly breadcrumbs?: Breadcrumb[];
```

```typescript
interface Breadcrumb {
  label: string;
  /* Default = "../" */
  url?: string;
  onClick?(): void;
}
```

joelwarrington-jobber (commented):
Overall looking good! Added comments which are suggestions and not necessarily must be addressed IMO

tophernuts:
👌👀👌👀👌👀👌👀👌👀 good shit go౦ԁ sHit👌 thats ✔ some good👌👌shit right👌👌there👌👌👌 right✔there ✔✔if i do ƽaү so my self 💯 i say so 💯 thats what im talking about right there right there (chorus: ʳᶦᵍʰᵗ ᵗʰᵉʳᵉ) mMMLGTMMᎷМ💯 👌👌 👌НO0ОଠOOOOOОଠଠOoooᵒᵒᵒᵒᵒᵒᵒᵒᵒ👌 👌👌 👌 💯 👌 👀 👀 👀 👌👌Good shit

chris-at-jobber:
Needs more discussion

joelwarrington-jobber:
@chris-at-jobber, can we create an issue to capture that discussion? Having the breadcrumb functionality would be great, and be better to have a defined pattern instead of using buttons

PR #600, opened 2021-08-03
feat(components): Update react-hook-form and FormField component

eddysims:
## Motivations

There were some issues with the `FormField` component and `ref`s. This PR will update the `react-hook-form` component which gives us access to the `useController` hook from `react-hook-form`. 

## Changes

Updates the `react-hook-form` to the latest version and updates the `FormField` component to use the new `useController` hook.

### Added

- `useFormController` hook
- focus on error field on `Form` submission

### Changed

- updated the `react-hook-form` package
- move `FormField` to use the `useFormController` hook

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -36,7 +36,7 @@
     "framer-motion": "^1.10.3",
     "lodash": "^4.17.20",
     "react-dropzone": "^11.0.2",
-    "react-hook-form": "^6.7.2",
+    "react-hook-form": "^7.12.1",
     "react-image-lightbox": "^5.1.1",
     "react-markdown": "^4.1.0",
     "react-popper": "^2.2.5",
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -3,217 +3,193 @@ import React, {
   FocusEvent,
   KeyboardEvent,
   MutableRefObject,
+  PropsWithChildren,
   useEffect,
-  useImperativeHandle,
   useState,
 } from "react";
 import classnames from "classnames";
-import uuid from "uuid";
-import { Controller, useForm, useFormContext } from "react-hook-form";
 import styles from "./FormField.css";
 import { FormFieldProps } from "./FormFieldTypes";
 import { FormLabel } from "./FormLabel";
 import { FieldWrapper } from "./FieldWrapper";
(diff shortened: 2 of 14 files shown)


packages/components/package.json
```diff
@@ -36,7 +36,7 @@
     "framer-motion": "^1.10.3",
     "lodash": "^4.17.20",
     "react-dropzone": "^11.0.2",
-    "react-hook-form": "^6.7.2",
+    "react-hook-form": "^7.12.1",
```
hosseintalebi, line 39: 🎉

packages/components/src/FormField/useFormController.ts
```diff
@@ -0,0 +1,46 @@
...
+    control,
+    rules: validations,
+    defaultValue: value || "",
+  });
+
+  return { error: errors[fieldName], field };
+}
+
+function useControlName(name: string | undefined) {
+  const [identifier] = useState(uuid.v1());
```
lmorgan29, line 44: the `uuid` method is the only difference here between this version and the mobile one, is that correct?

packages/components/src/FormField/FormField.tsx
```diff
@@ -3,219 +3,195 @@ import React, {
...
-   * Generate a name if one is not supplied, this is the name
-   * that will be used for react-hook-form and not neccessarily
-   * attached to the DOM
-   */
-  const [controlledName] = useState(
-    name ? name : `generatedName--${identifier}`,
-  );
+  const { error, field } = useFormController({
+    name,
+    value: value || defaultValue,
```
lmorgan29, line 53: any reason to use `||` over `??` ? If for some reason you wanted a value of 0/false/etc, the default would be used instead.

packages/components/src/FormField/useFormController.ts
eddysims: yep, it is a direct copy other then the `uuid` package.

packages/components/src/FormField/FormField.tsx
eddysims: This has been changed.

chris-at-jobber:
@eddysims this one can be closed, yeah?

PR #605, opened 2021-08-10
feat(components): InputField prefix-postfix

deepvishwas:
## Motivations
Formfield component enhancement request for - prefix and postfix

## Changes
- Ability to add prefix text and icon
- Ability to add postfix text and icon

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/FormField/Affix.tsx
```diff
@@ -0,0 +1,76 @@
+import React from "react";
+import { IconNames } from "@jobber/design";
+import styles from "./FormField.css";
+import { FormFieldProps } from "./FormFieldTypes";
+import { Icon } from "../Icon";
+
+export function Affix(props: FormFieldProps) {
```
eddysims, line 7: we tend to want to de-structure the props here to make the component/props more readable.

packages/components/src/FormField/Affix.tsx
```diff
@@ -0,0 +1,76 @@
...
+        {props.postfixIcon && getPostfixIcon(props.postfixIcon)}
+      </div>
+    );
+  }
+
+  return (
+    <>
+      {isPrefix() ? prefix() : false}
+      {isPostfix() ? postfix() : false}
+    </>
```
eddysims, line 74: In React, we want to return our markup. I know that this is how `FormField` returns, but that is not best practice. All of the functions and conditionals should be in here.

This should also, ideally be 2 components. One for `Prefix` and one for `Postfix`.

packages/components/src/FormField/FormField.tsx
```diff
@@ -89,15 +94,31 @@ export function FormField(props: FormFieldProps) {
...
+            {
+              [styles.withDoublePrefix]: prefixIcon && prefixText,
+            },
+
+            {
+              [styles.withPostfix]: !postfixIcon != !postfixText,
+            },
+            {
+              [styles.withDoublePostfix]: postfixIcon && postfixText,
+            },
```
eddysims, line 120: This can be one object, there is no need to create a new object for each check.

```ts
{ 
  [styles.select]: type === "select",
  [styles.withPrefix]: !prefixIcon != !prefixText,
  [styles.withDoublePrefix]: prefixIcon && prefixText,
  [styles.withPostfix]: !postfixIcon != !postfixText,
  [styles.withDoublePostfix]: postfixIcon && postfixText,
}
```

packages/components/src/FormField/FormField.tsx
```diff
@@ -89,15 +94,31 @@ export function FormField(props: FormFieldProps) {
         name={controlledName}
         rules={{ ...validations }}
         defaultValue={value ?? defaultValue ?? ""}
+        // eslint-disable-next-line max-statements
```
eddysims, line 97: Why is this needed? This is something we really want to avoid doing. If this is absolutely needed, we want to add a description of why.

packages/components/src/FormField/FormField.tsx
```diff
@@ -141,6 +162,7 @@ export function FormField(props: FormFieldProps) {
               default:
                 return (
                   <>
+                    <Affix {...props} />
```
eddysims, line 165: I think that you could have this component accept only a single prop

```
{prefix && <Prefix prefix={prefix} />}
```

packages/components/src/FormField/FormField.tsx
```diff
@@ -151,6 +173,7 @@ export function FormField(props: FormFieldProps) {
                       ref={inputRef as MutableRefObject<HTMLInputElement>}
                     />
                     {loading && <FormSpinner />}
+                    <Affix {...props} />
```
eddysims, line 176: Same as above

packages/components/src/FormField/FormFieldTypes.ts
```diff
@@ -115,6 +116,26 @@ export interface FormFieldProps {
...
+  /**
+   * Set the component prefix to the given prefix text.
+   */
+  readonly postfixText?: string | number;
+
+  /**
+   * Set the component to the given value.
+   */
+  readonly postfixIcon?: IconNames;
+
```
eddysims, line 138: Ideally these are only 2 props.

We want props to be as predictable and simple to the consumer as possible. If I am a consumer of the component, i would not expect to have to add `prefixIcon`, `prefixText`, `prefixAction` to have a full prefix on my component.
 
```ts
type Affix = {
  icon?: IconName
  label?: string
  action?: () => void
}

prefix?: Affix
postfix?: Affix
```

eddysims (changes requested):
Why was the `.private` file removed? This is needed to keep the component private from the rest of Atlantis. This shouldn't be removed.

packages/components/src/FormField/FormField.tsx
deepvishwas: 👍🏻

chris-at-jobber:
Closing as this work is being finalized in https://github.com/GetJobber/atlantis/pull/613

PR #615, opened 2021-08-23
fix(design): temp fix the colors until a good fix is available

eddysims:
## Motivations

the colors are currently breaking when using `rgba` with a `var`. This PR will revert the `rgba` with a `var` colors for the time being, until a better fix is worked on.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/design/src/colors.css
+++ b/packages/design/src/colors.css
@@ -51,8 +51,8 @@
   --color-border: var(--color-grey--lighter);
   --color-border--section: var(--color-blue);
 
-  --color-overlay: rgba(var(--color-greyBlue--rgb), 0.8);
-  --color-overlay--dimmed: rgba(var(--color-white--rgb), 0.6);
+  --color-overlay: rgba(101, 120, 132, 0.8);
+  --color-overlay--dimmed: rgba(255, 255, 255, 0.6);
 
   /* Brand */
   --color-brand: var(--color-green);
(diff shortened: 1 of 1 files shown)


packages/design/src/colors.css
```diff
@@ -110,8 +110,10 @@
   --color-grey--lightest: rgb(244, 244, 244);
   --color-grey--dark: rgb(118, 118, 118);
 
-  --color-greyBlue--rgb: 101, 120, 132;
-  --color-greyBlue: rgb(var(--color-greyBlue--rgb));
+  /* TODO: Reintroduce RGB once the issue with JobberStyle is fixed */
+  /* --color-greyBlue--rgb: 101, 120, 132; */
+  /* --color-greyBlue: rgb(var(--color-greyBlue--rgb)); */
```
adambobadam, line 115: Just a minor thing: wondering what the protocol is for comments like this in the code base since Git has this history :)

chris-at-jobber:
## QA

- [x] Works in Atlantis
- [x] Works in mobile app
- [x] Works in web app

The one odd note was the Chrome devtools displaying the `overlay--dimmed` value as an hsla value in Atlantis:
image

But the computed color is the same...
image

And the output in consuming apps is as expected.

packages/design/src/colors.css
chris-at-jobber: Our plan is to revisit this this sprint for cleanup after we've got the ReactNative consumption issue unblocked, so I think it's OK to leave this kind of note for ourselves.

chris-at-jobber (approved):

PR #620, opened 2021-08-25
fix(components): Show tooltip on focus and let the screen readers announce the message [JOB-26479]

darryltec:
## Motivations

There's a huge deficiency from tooltips where screen readers doesn't announce the tooltip message and the tooltip doesn't show when exclusively using the keyboard to trigger the tooltip target 

### Added

- Show/hide the tooltip on focus/blur
- Announce tooltip message on screen readers

### Changed

- Made the tooltip transition out

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


chris-at-jobber:
Tested in-app ✅ 

- [x] adds focus to otherwise non-focusable child elements
- [x] Tooltip content is read to screen-reader as expected

packages/components/src/Tooltip/Tooltip.test.tsx
```diff
@@ -13,8 +13,11 @@ test("tooltip shouldn't show up", () => {
       <div>{content}</div>
     </Tooltip>,
   );
-  expect(queryByText(message)).toBeNull();
-  expect(getByText(content)).toBeTruthy();
+
+  await waitFor(() => {
+    expect(queryByText(message)).not.toBeInTheDocument();
+  });
+  expect(getByText(content)).toBeInTheDocument();
```
joelwarrington-jobber, line 20: You shouldn't need to await `queryByText` as it doesn't require async, see [React Testing Library Cheatsheet](https://testing-library.com/docs/react-testing-library/cheatsheet/)

packages/components/src/Tooltip/Tooltip.test.tsx
```diff
@@ -31,6 +34,59 @@ test("tooltip should show up on hover", () => {
...
+
+  const { getByTestId, getByText } = render(
+    <Tooltip message={message}>
+      <div data-testid={contentID}>{content}</div>
+    </Tooltip>,
+  );
+
+  act(() => {
+    getByTestId(contentID).dispatchEvent(new Event("focus"));
+  });
```
joelwarrington-jobber, line 54: I don't think we should need to use `act` directly and can use `waitFor` instead

packages/components/src/Tooltip/Tooltip.test.tsx
```diff
@@ -31,6 +34,59 @@ test("tooltip should show up on hover", () => {
   act(() => {
     getByTestId(contentID).dispatchEvent(new MouseEvent("mouseenter"));
   });
-  expect(getByText(message)).toBeTruthy();
-  expect(getByText(content)).toBeTruthy();
+  expect(getByText(message)).toBeInTheDocument();
+  expect(getByText(content)).toBeInTheDocument();
+});
+
+test("tooltip should show up on focus", () => {
```
joelwarrington-jobber, line 41: personal preference, should try to use `it` followed by a string that works like this:

```suggestion
it("should show up on focus", () => {
```

joelwarrington-jobber: this isn't a hard stop but might be useful to start a consistent approach

joelwarrington-jobber (changes requested):
overall implementation is good to go, just cleaning up the tests

packages/components/src/Tooltip/Tooltip.test.tsx
darryltec: I'm actually waiting for the `.not.toBeInTheDocument();` bit as that one takes a moment to appear

packages/components/src/Tooltip/Tooltip.test.tsx
darryltec: ah same here. don't know why I missed that. haha. I'll update it

packages/components/src/Tooltip/Tooltip.test.tsx
darryltec: actually wait. the way the waitFor works is it would retry for a bit and it it still fails, then it will fail. 

> waitFor may run the callback a number of times until the timeout is reached

packages/components/src/Tooltip/Tooltip.test.tsx
darryltec: I'll do you one better. I used the `userEvent` instead and not use `act`anymore

packages/components/src/Tooltip/Tooltip.test.tsx
joelwarrington-jobber: you shouldn't need to do that though, look at this example: https://testing-library.com/docs/guide-disappearance/#nottobeinthedocument

darryltec: @joelwarrington-jobber in most cases, yup. until you deal with things that needs to transition out in which case there's a bit of delay before that element truly disappears

joelwarrington-jobber (approved):
LGTM

darryltec:
Master merged cleanly. Should be safe to merge in

PR #621, opened 2021-08-26
feat(components): Add visibility toggle for password input

joelzwarrington:
## Motivations

In some cases before submitting a signup or login form, users might want to see if their password is what they typed, having a toggle to see their password allows them to do that. This feature has been added in a way that a developer can decide if a password input should have the visibility toggle as it is by default not enabled.

### Added

- Added a toggle action to make passwords visible using a suffix for password input

## Testing

- Set `hasVisibility` prop to true on a password, and click on the suffix action to change between showing and hiding the password 
---

Random photo of Atlantis


chris-at-jobber:
@joelwarrington-jobber Found something fun!

When I have a password manager (PWM for the rest of this post), the browser extension adds the little icon to the password input. 
- at larger sizes, this is fine because the PWM's icon appears to respect the spacing of our action
- the PWM's icon's spacing _scales_ so it gets progressively closer to, and then passes under, our action, as the inputs get smaller (illustrated at a few sizes so you can see the progression)

image

I know each PWM has their own icon, so this may actually change from LastPass to OnePass to etc etc, and so I'm not sure how much we actually _control_ it (it looks like it's being injected into the `input` as an inline style):
image

Not necessarily a blocking comment, as again I'm not sure how much control we have over an unknown number of password manager extensions... but something to consider?

joelwarrington-jobber:
@chris-at-jobber I found that adding a `margin-right` of `8px` to the input itself would position the LastPass icon correctly. Do you think this would be the right approach? Should we even take this approach?

https://user-images.githubusercontent.com/80279872/134242428-69657766-4835-47cb-aee0-a8efa29d7303.mov

chris-at-jobber:
I like the approach! This is a prime use case for the suffix action. CSS-wise I'd want to get @eddysims to weigh in on the margin approach as he just spent a ton of time in the weeds of `FormField`, but it seems reasonable at first glance.

eddysims:
@joelwarrington-jobber @chris-at-jobber 

The code in here looks good to me, pending the merge conflict.

My thoughts on the `margin-right: 8px` solution:

- I worry that with this we are solving a **very** specific problem. Are we sure that this will work for all password managers? Or is this a solution for the Chrome extension for LastPass? I use a different password manager and it appears very slightly different.

image

- Do we want to give that margin to everyone without a password manager? Doing this to a user that does not have a password manager means that they will have 8px extra spacing on the right hand side of the input all the time. Unless we have a way to check if they have a password manager

image

---

After typing this up. My final thought is: **If we are okay with the above, add it in here. If we are not, make it a separate PR**

joelwarrington-jobber:
@chris-at-jobber @eddysims if you two are okay with it, I think we should solve that problem in a different PR/more sophisticated way. I've added an issue to keep track of that and maybe discuss what needs to be done before any work here: https://github.com/GetJobber/atlantis/issues/687

joelwarrington-jobber:
@eddysims mind giving this a review with the comment above in mind?

PR #624, opened 2021-08-29
feat(components): add subtle button [JOB-34785]

chris-at-jobber:
## Motivations

We're seeing a growing need to have actions that blend in with their surroundings more, so we should define it rather than working around it!

### Added

- New "variation" of button called "subtle"
- New color "subtle" to interactive colors

### Deprecated

- Sets up button to easily replace instances of cancel with subtle, before a full removal

---

Random photo of Atlantis


eddysims (commented):
@chris-at-jobber the code for this looks good to me. I have a couple questions in regards to the typing.

1. Would we want a `subtle` button to only be an icon? Do we want it to allow for a label?
2. Do you think this is better as a `variation` rather than a `type`?

chris-at-jobber:
@eddysims
- I think we can allow it to be a label as well
- Maybe? Let me poll the designers who I was showing this to the other day and get back to you

chris-at-jobber:
Let's go with "type"

joelwarrington-jobber (changes requested):
In the description of the subtle action - it describes the button as transparent - which is not really the case.

Can you clarify if the subtle actions should be transparent? also how do you think we would handle contrast between the icon outlines and the background

image

chris-at-jobber:
@joelwarrington-jobber should definitely be transparent, good catch! Will push that up.

As far as contrast, this has been built with an assumption that we're using Button on some variation of our `surface` colors but that's a good point about potentially having a version that works against our `surface--reverse` colour. I'm not sure if there's a great way to do that programmatically?

chris-at-jobber:
Button is now transparent - does receive a fill on hover/focus 🚁

darryltec:
@chris-at-jobber 

1. Are the text and icon supposed to have different colors? The icon is pure black but the text is something else.  image
2. Cancel have no outline now?  image

joelwarrington-jobber:
Transparency works!

https://user-images.githubusercontent.com/80279872/132419139-5056f39b-aa6a-4e7e-bc05-6464951f9c4c.mov

However - as you can see from the video above - maybe we can try to address the contrast? I don't know if there's a way to automatically detect the contract and update the color within the component but that'd be my first suggestion.

If not technically possible - maybe we can add a color prop for the icon/label?

chris-at-jobber:
@joelwarrington-jobber I _really_ want to avoid us setting actual color props on buttons. We've seen in the past that it gets way out of control as it's a really descriptive approach that doesn't scale well (ie in the old days stacking utility classes to make a `button--green button--ghost button--large` that becomes really difficult to update). I'd almost prefer to put design documentation that "these subtle actions should only be used overtop of `surface` or `surface--background` surfaces" and then see anyone needs a lighter version introduce it then, rather than add more props now?

joelwarrington-jobber:
> @joelwarrington-jobber I _really_ want to avoid us setting actual color props on buttons. We've seen in the past that it gets way out of control as it's a really descriptive approach that doesn't scale well (ie in the old days stacking utility classes to make a `button--green button--ghost button--large` that becomes really difficult to update). I'd almost prefer to put design documentation that "these subtle actions should only be used overtop of `surface` or `surface--background` surfaces" and then see anyone needs a lighter version introduce it then, rather than add more props now?

@chris-at-jobber, I'm happy with your suggestion and agree, adding a colour prop would be very cumbersome! We could always consider using a specific css colour variable that we actively expose as something people could change? But that's something to be explored in the future.

Will hop out of reviewing this as I believe @darryltec might take that on, but if another review is needed, simply request one and I'll jump in :)

chris-at-jobber:
Thanks @joelwarrington-jobber !

darryltec:
@chris-at-jobber so variations are not supposed to work, eh? 
image

chris-at-jobber:
Yep, any button at "subtle" will look the same 

Designers were asking if we should make it so you can only use it with `work` and `cancel` for example... we could do that if we wanted to right?

darryltec:
@chris-at-jobber Totally. I think we should remove the variation when it's a subtle type tho since it won't do anything. Gonna try it on my machine first. One momo.

darryltec:
Talked offline. `subtle` is gonna be a variation now (pending design discussion) as it is pretty much the tertiary cancel with out the extra weight of the word "cancel".

packages/components/src/Button/Button.css
```diff
@@ -73,39 +73,85 @@ a.button:active {
...
+  --button--color-variation: var(--color-interactive--subtle);
+  --button--color-variation--hover: var(--color-interactive--subtle--hover);
 }
 
 /* Type */
 
-.primary:not(.cancel),
-.primary:not(.cancel):hover,
-.primary:not(.cancel):focus {
+.primary:not(.cancel, .subtle),
```
chris-at-jobber, line 84: I have opted to used the selector list argument here because

1. reduced number of total lines of CSS to read
2. doing it in separate lines appeared to cancel out whichever one was listed second

packages/design/src/colors.css
```diff
@@ -8,6 +8,8 @@
   --color-destructive--hover: var(--color-red--dark);
   --color-cancel: var(--color-greyBlue);
   --color-cancel--hover: var(--color-greyBlue--dark);
```
chris-at-jobber, line 10: Originally planned to remove cancel color in this PR, preserving to prevent a breaking change in mobile

packages/design/src/Colors.mdx
```diff
@@ -77,19 +77,22 @@ currently tied our interactive color to our brand color._
...
+icon action in main navigation, or buttons to dismiss or cancel a workflow.
+
 #### Destructive
 
 <Swatch colors={[`--color-destructive`, "--color-destructive--hover"]} />
 
 Use to signify that an interaction will destroy something in the users’ account
 or workflow.
 
-#### Cancel
```
chris-at-jobber, line 87: Removing docs on "cancel" color

chris-at-jobber:
@darryltec Please, take a look 🙏

chris-at-jobber:
QA: If you want to quickly verify that I didn't break `cancel` as part of the soft-deprecation plan, add this snippet to the sandbox:

```
    <Button label="Cancel" type="primary" variation="cancel" />
    <Button label="Cancel" type="secondary" variation="cancel" />
    <Button label="Cancel" type="tertiary" variation="cancel" />

    <Button variation="cancel" type="tertiary" icon="search" aria-label="search" />
    <Button variation="cancel" type="tertiary" icon="cog" aria-label="settings" />
    <Button variation="cancel" type="tertiary" icon="help" aria-label="help" />
    <Button variation="cancel" type="tertiary" icon="more" aria-label="more" />
```

packages/design/src/colors.css
darryltec: Speaking of... we have to update the mobile app now too, eh?

chris-at-jobber: Yep, but @agglasaur is working on some changes to Button so there's going to be work done in there anyways 😅

darryltec: I am cool with that answer

darryltec (approved):
Tis' looks good to me

chris-at-jobber:
🙏  just waiting for specs and then will merge in

PR #625, opened 2021-08-30
fix(components): Make onclick card operable by keyboard [JOB-30838]

darryltec:
## Motivations

- `onClick` cards receives focus but it's inoperable with keyboard keys. 
  - Specifically, `Enter` and `Spacebar` which you normally get on a button
- `href` cards stays the same with it only being triggered by `enter` as per how it normally works
- TIL: 
  - `spacebar` and `enter/return` key only works for `button` tag
  - only `enter/return` works for an `a` tag
  - even if you add a tabindex and an onclick on a div, `spacebar` and `enter/return` key won't work so you have to trigger it manually

### Added

- Added an internal `CardClickable` component to isolate the codes that's only used when a card is clickable
  - This includes a `ref` which is used to trigger the click event that gets passed in to the `onClick` prop instead of the keyboard event. 
  - The `ref` is also a big reason why it has to be on its own component since the `ref` would have to either be an `a` tag or a `div` which is a mess when it comes to typescript.
- Prevent scroll down after a spacebar tap when focused on a clickable card

### Changed

- Cards with an `onClick` prop can be triggered by an `Enter` or `Spacebar` key tap while focused on said card.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Card/CardClickable.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const ENTER_KEY = "Enter";
+const SPACEBAR_KEY = " ";
+
+/**
+ * This is only intended to be used in the Card component.
+ * Please use `<Card onClick={onClick} />` component instead.
+ */
+export function CardClickable({
```
darryltec, line 16: Couldn't find an example for a nested components naming so I went with the whole `<ParentChild />` naming style since this component should only be used by the `Card`.

packages/components/src/Card/CardClickable.tsx
```diff
@@ -0,0 +1,49 @@
...
+      role="button"
+      tabIndex={0}
+    >
+      {children}
+    </div>
+  );
+
+  function handleKeyup(event: React.KeyboardEvent<HTMLDivElement>) {
+    if (event.key === ENTER_KEY || event.key === SPACEBAR_KEY) {
+      cardRef.current?.click();
```
darryltec, line 40: Went with manually triggering a click instead of passing in the event since the `KeyboardEvent` is different from a `MouseEvent`. Also, because typescript.

packages/components/src/Card/CardClickable.tsx
deepvishwas: 👍🏻  Not sure what Atlantis convention is but if it was up to Shopify this would have been called ClickCards. Weird I know. I like plain old [Domain]|[Context]|Component|[Type] approach

darryltec: We kind of have one but not officially(?). 

image

I agree tho, `ClickableCard` actually makes sense. I just went with it being under the `Card` umbrella.

chris-at-jobber:
Looks good to design 👍

packages/components/src/Card/CardClickable.tsx
deepvishwas: 💯

packages/components/src/Card/CardClickable.tsx
deepvishwas: I wonder if card content could ever have an anchor tag or a button? 
If they do, those actions won't be triggered.

e.g. Jobber link will not work 
image

packages/components/src/Card/CardClickable.tsx
```diff
@@ -0,0 +1,50 @@
...
+}: ClickableCardProps) {
+  const cardRef = createRef<HTMLDivElement>();
+
+  return (
+    <div
+      ref={cardRef}
+      data-testid="clickable-card"
+      className={className}
+      onClick={onClick}
+      onKeyUp={handleKeyup}
```
deepvishwas, line 29: Could we combine key-up and key-down handlers into one?

packages/components/src/Card/tests/CardClickable.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+test("it should trigger the click event after enter press", () => {
+  const clickHandler = jest.fn();
+
+  const { getByTestId } = render(
+    <CardClickable className="boink" onClick={clickHandler}>
+      <p>Press enter</p>
+    </CardClickable>,
+  );
+
```
deepvishwas, line 13: Could we add like fireEvent.keyUp(getByTestId("clickable-card"), { key: "Backspace" });
expect(clickHandler).toNotHaveBeenCalledTimes();

just to make sure it is not being called on any other key press.

packages/components/src/Card/tests/CardClickable.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+test("it should trigger the click event after spacebar press", () => {
+  const clickHandler = jest.fn();
+
+  const { getByTestId } = render(
+    <CardClickable className="boink" onClick={clickHandler}>
+      <p>Press spacebar</p>
+    </CardClickable>,
+  );
+
+  fireEvent.keyUp(getByTestId("clickable-card"), { key: " " });
```
deepvishwas, line 27: I think, same as above

packages/components/src/Card/CardClickable.tsx
darryltec: Not with the context of firing events through space and enter since the `button`, which i'm mimicking here, fires on keyup

packages/components/src/Card/CardClickable.tsx
darryltec: I doubt we'd do that but it could happen. I could also make sure that the keyup listeners only fire when the focused element is actually the card and not things inside it. good catch!

deepvishwas: Awesome Cheers 🍻

packages/components/src/Card/CardClickable.tsx
deepvishwas: That makes sense

packages/components/src/Card/CardClickable.tsx
darryltec: So, new learnings today. The native enter press on a link also fires a click which bubbles up to the parent. So we can't stop it. haha

darryltec: So, it will not fire the enter/space keyup's anymore unless you're focused on the card itself. it firing the callback when focused on the link and pressing enter is a native thing that i think we should not remove.

packages/components/src/Card/tests/CardClickable.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { getByTestId } = render(
+    <CardClickable className="boink" onClick={clickHandler}>
+      <p>Press spacebar</p>
+    </CardClickable>,
+  );
+
+  getByTestId("clickable-card").blur();
+  fireEvent.keyUp(getByTestId("clickable-card"), { key: " " });
+  expect(clickHandler).toHaveBeenCalledTimes(0);
+});
```
joelwarrington-jobber, line 65: as a potential fast-follow, could I suggest the following structure, not required though

```
describe("when focused", () => {
  describe("when using url", () => {
    it("doesn't navigate using spacebar", () => {});
    it("navigates using enter", () => {});
  });
  describe("when using onClick", () => {
    it("should trigger with enter or spacebar", () => {});
  });
});

describe("when not focused", () => {
  it("shouldnt trigger with enter or spacebar", () => {});
});
```

joelwarrington-jobber (approved):
LGTM

- Card with `url` acts as an Anchor with `Enter`
- Card with `onClick` acts as a Button, with both `Enter` and `Space` working to trigger

https://user-images.githubusercontent.com/80279872/134086872-1eb3e25a-578b-4487-be4e-eebc9bdd69d3.mov

chris-at-jobber:
@darryltec if you can ✂️  cut a pre-release of this, I can do a QA pass in Jobber

darryltec:
@chris-at-jobber 
```
 - @jobber/components@2.43.4-pre.10+9606e17
 - @jobber/docz-tools@1.6.29-pre.10+9606e17
```

packages/components/src/Card/tests/CardClickable.test.tsx
darryltec: Ah, was debating if i should do that or not. haha. I'll go do it

darryltec:
Blocked. Pending confirmation with Growth on the onboarding Cards with checkbox acting weird.

@joelwarrington-jobber Updated the test so if you can reapprove, that would be great!

joelwarrington-jobber (approved):
👍  LGTM

chris-at-jobber:
Blocked note: new `onClick` card is incompatible with an in-app implementation that has built keyboard functionality around Card. Will explore a resolution.

darryltec:
Looked in to the issue of the clickable card on the onboarding modals. It's using a completely custom component that is not related to card so this should be safe to be merged. Tho some tests does need updating so will hold off on merging until the chips is out.

chris-at-jobber:
@darryltec we good now?(pending the issue in #711 )

darryltec:
Yup. we should be good to go

PR #626, opened 2021-08-30
feat(components): Allow inline labels to be dismissible

joelwarrington-jobber:
## Motivations

`InlineLabel`'s seem like usage would be for tags to display status or select categories and feels appropriate to be able to dismiss selections. I would particularly be interested in having dismissible available so that a multi-select (either select or autocomplete) can use these to display selected values.

image

### Added

- Added ability to configure a dismiss action for inline labels

## Testing

- See markdown document for `<InlineLabel />` and configure an `onDimiss` function which will show the dismissible button

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InlineLabel/InlineLabel.tsx
+++ b/packages/components/src/InlineLabel/InlineLabel.tsx
@@ -1,7 +1,9 @@
 import React, { ReactNode } from "react";
+import { XOR } from "ts-xor";
 import classnames from "classnames";
 import styles from "./InlineLabel.css";
 import { Typography } from "../Typography";
+import { ButtonDismiss } from "../ButtonDismiss";
 
 export type InlineLabelColors =
   | "greyBlue"
@@ -19,7 +21,7 @@ export type InlineLabelColors =
   | "lightBlue"
   | "indigo";
 
-interface InlineLabelProps {
+interface BaseInlineLabelProps {
   /**
    * The size of the label
    * @default base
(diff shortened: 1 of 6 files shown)


packages/components/src/InlineLabel/InlineLabel.tsx
```diff
@@ -55,6 +67,15 @@ export function InlineLabel({
       <Typography element="span" size={sizeMapper[size]} textCase="uppercase">
         {children}
       </Typography>
+      {dismissable && (
+        <ButtonDismiss
+          onClick={() => {
+            onDismiss && onDismiss();
+          }}
+          ariaLabel={dismissAriaLabel || ""}
```
darryltec, line 75: you could put that empty string as a default so you won't have to null check it.

image

Also, thoughts on putting "Remove" as the default?

packages/components/src/InlineLabel/InlineLabel.tsx
```diff
@@ -33,6 +35,13 @@ interface InlineLabelProps {
   children: ReactNode;
 }
 
+interface DismissableInlineLabelProps extends BaseInlineLabelProps {
+  dismissAriaLabel: string;
+  onDismiss(): void;
+}
+
+type InlineLabelProps = XOR<BaseInlineLabelProps, DismissableInlineLabelProps>;
+
```
darryltec, line 44: you can actually avoid the whole XOR thing by adding 1 prop with the `DismissableInlineLabelProps` as the type

For example

```tsx
interface InlineLabelProps {
  // label props here
  ...
  dismissAction?: DismissableInlineLabelProps;
}

interface DismissableInlineLabelProps {
  ariaLabel: string;
  onDismiss(): void;
}
```

and using it as 

```jsx
<InlineLabel dismissAction={{ ariaLabel: "", onDismiss: () => {} }} >
  Thing
</InlineLabel>
```

bonus point for not having to say `dismissAriaLabel`

darryltec (commented):
Was told about the pending design discussion and coordination but imma still leave this comment right here

joelwarrington-jobber:
@darryltec thanks for the review! I was unaware internal teams were looking to build out a chip component which would handle the exact behaviour in this PR. Going to close this PR out and link the chip component when that PR is started

joelwarrington-jobber:
See work to implement chip component instead: https://github.com/GetJobber/atlantis/pull/629

PR #629, opened 2021-08-31
feat(components): Implement chip component

joelwarrington-jobber:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


joelwarrington-jobber:
https://user-images.githubusercontent.com/80279872/131458182-839727e9-d0d1-44df-9b02-a5f88c7b5eae.mov

packages/components/src/Chip/Chip.mdx
```diff
@@ -0,0 +1,42 @@
...
+<Playground>
+  <Chip
+    label="Darth Vader"
+    dismissAction={{
+      ariaLabel: "Remove",
+      onDismiss: () => alert("Luke, I am your father"),
+    }}
+  />
+</Playground>
+
```
joelwarrington-jobber, line 26: Would like some feedback on what to document here as I'm not sure what considerations were made when it was designed

packages/components/src/ButtonDismiss/ButtonDismiss.tsx
```diff
@@ -16,6 +21,7 @@ export function ButtonDismiss({ onClick, ariaLabel }: ButtonDismissProps) {
         onClick={onClick}
         type="tertiary"
         variation="cancel"
+        size={size == "small" ? "small" : "base"}
```
darryltec, line 24: This should do the trick since you can only pass in either `small` or `base`. If the person using this doesn't pass in a size, it'll fallback to `base` according to line `14`

```suggestion
        size={size}
```

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,32 @@
...
+  border: var(--chip-border-size) solid var(--chip-border-color);
+  border-radius: var(--chip-border-radius);
+  align-items: center;
+  gap: var(--chip-item-spacing);
+}
+
+.chip:hover {
+  box-shadow: var(--chip-shadow-hover);
+  transition: all var(--timing-base) ease-in-out;
+}
```
darryltec, line 26: Kinda odd for it to have a hover effect but not be clickable. @chris-at-jobber was the chip itself meant to be clickable?

packages/components/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { ButtonDismiss } from "../ButtonDismiss";
+
+interface DismissibleAction {
+  ariaLabel: string;
+  onDismiss(): void;
+}
+
+interface ChipProps {
+  label: string;
+  avatar?: Exclude<AvatarProps, "size">;
```
darryltec, line 14: Just a thought. I wonder if there's a better way to deal with this exclude thing. Mainly concerned about how it shows up in the props table

image

packages/components/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import styles from "./Chip.css";
+import { Avatar, AvatarProps } from "../Avatar";
+import { Text } from "../Text";
+import { ButtonDismiss } from "../ButtonDismiss";
+
+interface DismissibleAction {
```
darryltec, line 7: Hmmm maybe `DismissActionProp`? Consistent with how we name props. I.e., `ChipProps`

```suggestion
interface DismissActionProp {
```

packages/components/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,30 @@
...
+export function Chip({
+  label,
+  avatar,
+  dismissAction: { ariaLabel, onDismiss },
+}: ChipProps) {
+  return (
+    <span className={styles.chip}>
+      {avatar && <Avatar {...avatar} size="small" />}
+      <Text>{label}</Text>
+      <ButtonDismiss ariaLabel={ariaLabel} onClick={onDismiss} size="small" />
```
darryltec, line 27: Extra convenience thing. Since the label is a `String`, we can pretty much guess what the aria label would be. I.e., `"Remove" + label` = `"Remove Darryl Tec"`. Then at that point, the aria label can be optional or, dare I say, not exists?.

darryltec (changes requested):

packages/components/src/Chip/Chip.mdx
darryltec: @chris-at-jobber could probably help or point you to the right person/place

packages/components/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface DismissibleAction {
+  ariaLabel: string;
+  onDismiss(): void;
+}
+
+interface ChipProps {
+  label: string;
+  avatar?: Exclude<AvatarProps, "size">;
+  dismissAction: DismissibleAction;
```
darryltec, line 15: Had a little sudden realization here. All of your tests and examples mentions the aria label as "Remove"

image
image

Should this prop be called remove instead of dismiss?

darryltec:
Might need to balance that padding out 

Chip___🔱_Atlantis
Chip___🔱_Atlantis

cc: @chris-at-jobber

packages/components/src/ButtonDismiss/ButtonDismiss.tsx
```diff
@@ -5,9 +5,14 @@ import { Button } from "../Button";
...
 }
 
-export function ButtonDismiss({ onClick, ariaLabel }: ButtonDismissProps) {
+export function ButtonDismiss({
+  onClick,
+  ariaLabel,
+  size = "base",
+}: ButtonDismissProps) {
   return (
     <div className={styles.closeButtonWrapper}>
```
darryltec, line 17: This is not you, but i'm making a note of it here

We might have to add a `display:flex` on that `closeButtonWrapper` class as it is housing an `inline-*` element. which adds some super nice white space/odd spacing

height of the button is 25.63px

image

which makes the height of the chip 31.63

image

Adding the `display: flex` (there might be a different way) evens those babies out

image
image

Which is, I'm going to assume, the height that we want(??) and wouldn't mess with some vertical alignment things.

And most importantly, fix this alignment issue as well

Before
image

After
Chip___🔱_Atlantis

packages/components/src/Chip/Chip.css
chris-at-jobber: Chip should be clickable and focusable - it's a pretty versatile component. @joelwarrington-jobber can you look at the Chip in the Jobber codebase as a reference?

packages/components/src/Chip/Chip.tsx
joelwarrington-jobber: Interfaces which are used for React components should use the `Props` suffix, in this case `DimissibleAction` is a sub-interface so I don't think it needs to have the `Props` suffix.

packages/components/src/Chip/Chip.tsx
joelwarrington-jobber: While I agree it should always likely be `Remove + label`, I believe allowing whoever to specify the string works best, especially considering localization and what not.

packages/components/src/Chip/Chip.tsx
darryltec: Ah that's fair!

packages/components/src/Chip/Chip.tsx
joelwarrington-jobber: The `ariaLabel` specified in the tests haven't been selected explicitly. However I do believe the prop name could be changed, I like using `onDismiss` because it is consistent with the `ButtonDismiss` component, however always welcome to change if a different name is better

packages/components/src/ButtonDismiss/ButtonDismiss.tsx
joelwarrington-jobber: Would we be able to move this discussion to an issue regarding to the dismiss button?

packages/components/src/Chip/Chip.tsx
joelwarrington-jobber: I find the `Props` react component to display the table doesn't like conditional interfaces or the custom `Exclude, Pick, etc.` interface utilities. How to resolve that I'm not sure since these types work best when writing actual TypeScript, any thoughts @eddysims ?

packages/components/src/Chip/Chip.tsx
darryltec: Ah! Also realized that `ButtonDismiss` is a private component. In that case, I'd follow the components that already uses it. Which is the popover and modal that both calls `onRequestClose` so something along the lines of `onRequest...`. `onRequestDismiss`?

Anyway, my point about your example aria label is going with the more natural word that someone consuming this would use. The fact that you used "Remove" instead of "Dismiss" as an example of how would someone use this kinda suggests that this should probably be "remove". I don't have a strong opinion on it tho, just thinking about the predictability of it when the non-atlantis friends are consuming it since they wouldn't know that `ButtonDismiss` exists.

packages/components/src/Chip/Chip.tsx
darryltec: Good point!! Forgot about the localization thing

packages/components/src/Chip/Chip.css
joelwarrington-jobber: The Figma file I was referencing didn't have a lot of the considerations written out which has led to me assuming functionality that is not directly visible. Have started investigations into our internal Chip component and see what changes can be made from there

joelwarrington-jobber:
@darryltec re: padding on the left and right sides of the Chip, they're the same, however the icon is a button which doesn't have a border which is visible without hovering or focusing.

@chris-at-jobber let me know if you would however want to increase padding on either side to make them appear more balanced

packages/components/src/ButtonDismiss/ButtonDismiss.tsx
darryltec: Yup

packages/components/src/Chip/Chip.tsx
darryltec: It's no biggie tho. Alternatively, you can allow people to pass in the entire `Avatar` component and in the code we always set the size to small. Regardless of wether someone sets it to base. Don't really know if that's better. haha.

But yeah, just concerned about the people reading that and not understanding what even is supposed to be there.

joelwarrington-jobber:
adding blocked label - as am waiting on event being available for buttons which is in  https://github.com/GetJobber/atlantis/pull/631

joelwarrington-jobber:
Set to 90% but I believe it's a bit less than that at the moment, as need to:
- flesh out documentation
- document types/props
- confirm prop/component usage
- confirm color/spacing design

https://user-images.githubusercontent.com/80279872/131581809-3ee3d7ce-10e5-42d6-98ec-e36a50f22c29.mov

joelwarrington-jobber:
@vanikadhamija has requested to take on this PR and take it across the finish line 😄

darryltec:
I will now be taking this on. Thanks @joelwarrington-jobber @vanikadhamija @igor-m-jobber 👌🏼

chris-at-jobber:
Closing per Darryl's note above!

PR #631, opened 2021-08-31
feat(components): Add `event` argument to onClick events

joelwarrington-jobber:
## Motivations

In some cases it is useful to have access to the mouse event when building more complex components that nest multiple components. In some cases it's needed to `event.stopPropagation()` or `event.preventDefault()`.

### Added

- Added `event` argument for all existing onClick events

## Testing

- Use any of the components which have `onClick` events and log it when clicked!

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -27,7 +27,7 @@ interface ButtonFoundationProps {
   readonly label?: string;
   readonly loading?: boolean;
   readonly size?: "small" | "base" | "large";
-  onClick?(): void;
+  onClick?(event: React.MouseEvent<HTMLElement>): void;
 }
 
 interface ButtonIconProps extends ButtonFoundationProps {
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -27,7 +27,9 @@ interface ButtonFoundationProps {
   readonly label?: string;
   readonly loading?: boolean;
   readonly size?: "small" | "base" | "large";
-  onClick?(event: React.MouseEvent<HTMLElement>): void;
+  onClick?(
+    event: React.MouseEvent<HTMLAnchorElement | HTMLButtonElement>,
+  ): void;
 }
 
 interface ButtonIconProps extends ButtonFoundationProps {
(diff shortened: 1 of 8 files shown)


darryltec:
Looks good to me. Did some prerelease too in case there were surprises. Install it from here `npm install @jobber/components@2.43.5-pre.2+c71c4d2`. Looks like _no surprises_ but wouldn't hurt for someone else to check also.

@eddysims do you have any reservations on surfacing events on every click-y bits?

packages/components/src/Button/Button.tsx
```diff
@@ -27,7 +27,7 @@ interface ButtonFoundationProps {
   readonly label?: string;
   readonly loading?: boolean;
   readonly size?: "small" | "base" | "large";
-  onClick?(): void;
+  onClick?(event: React.MouseEvent<HTMLElement>): void;
```
eddysims, line 30: I'm just going to add this here, but applies to all of the work. Is there a reason you chose to pass the event as a generic `HTMLElement` instead of being more specific.

```tsx
onClick?(event: React.MouseEvent<HTMLButtonElement | HTMLAnchorElement>): void
```

eddysims (changes requested):
@joelwarrington-jobber just one question here.

packages/components/src/Button/Button.tsx
joelwarrington-jobber: @eddysims I followed the example in the card component, however I'm okay with making the change to make it more specific. 

Do you think anything is gained by making it more specific, and is the added work to maintain that if we change components worth it?

darryltec: Was looking in to this yesterday and it looks like we gain a little bit more data from the event (and by that I mean typescript would allow us). I.e. the `HTMLButtonElement` gives you `disabled` to check if the button is disabled or not. An `a` tag to grab the `href`.

I'm okay with keeping it generic and making it specific when someone does need to as we're starting with the generic type anyway.

joelwarrington-jobber: Sounds good - I also chose HTMLElement as everything would be a child of this type, but I can explicitly change the onClick to be specific for all the cases. 🔜

eddysims (approved):

PR #633, opened 2021-09-03
feat(components): Add destructive  option for confirmation modal

deepvishwas:
## Motivations
Currently, the confirmation modal only accepts a primary “work” action, which breaks this pattern.

we have determined the pattern for this is to show a secondary destructive button, which triggers a confirmation modal with a primary destructive action

## Changes
Adding actionType work | Destructive to the confirmation Button

### Added
Adding actionType work | Destructive to the confirmation Button

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -56,15 +56,9 @@ interface ButtonLinkProps extends ButtonFoundationProps {
 }
 
 interface BaseActionProps extends ButtonFoundationProps {
-  readonly variation?: "work" | "learning" | "subtle";
+  readonly variation?: "work" | "learning" | "subtle" | "destructive";
   readonly type?: "primary" | "secondary" | "tertiary";
 }
-
-interface DestructiveActionProps extends ButtonFoundationProps {
-  readonly variation: "destructive";
-  readonly type?: "primary" | "secondary" | "tertiary";
-}
-
 interface CancelActionProps extends ButtonFoundationProps {
   readonly variation: "cancel";
   readonly type?: "secondary" | "tertiary";
@@ -72,7 +66,6 @@ interface CancelActionProps extends ButtonFoundationProps {
 
 interface SubmitActionProps
   extends Omit<ButtonFoundationProps, "external" | "onClick"> {
-  readonly variation?: "work";
   readonly type?: "primary";
   readonly submit: boolean;
 }
@@ -87,7 +80,7 @@ interface SubmitButtonProps
 
 export type ButtonProps = XOR<
   BaseActionProps,
-  XOR<DestructiveActionProps, XOR<CancelActionProps, SubmitActionProps>>
+  XOR<CancelActionProps, SubmitActionProps>
 > &
   XOR<SubmitButtonProps, XOR<ButtonLinkProps, ButtonAnchorProps>> &
   XOR<ButtonIconProps, ButtonLabelProps>;
--- a/packages/components/src/ConfirmationModal/ConfirmationModal.mdx
+++ b/packages/components/src/ConfirmationModal/ConfirmationModal.mdx
@@ -109,6 +109,38 @@ that presents a confirm modal.
   }}
 </Playground>
 
+## Destructive
+
+If the user is confirming an action that will destroy, delete, or remove
+something, use a destructive ConfirmationModal to visually reinforce the
+potential consequences.
+
+<Playground>
+  {() => {
+    const [open, setOpen] = useState(false);
+    return (
+      <>
+        <Button
+          label="Delete Bob"
+          variation="destructive"
+          type="secondary"
+          onClick={() => setOpen(true)}
+        />
+        <ConfirmationModal
+          title="Delete Bob?"
+          message={`Deleting Bob will remove their data from your account for good.`}
+          confirmLabel="Delete Bob"
+          open={open}
+          onConfirm={() => alert("Bob has been deleted")}
+          onCancel={() => alert("Bob will not be deleted")}
+          onRequestClose={() => setOpen(false)}
+          variation="destructive"
+        />
+      </>
+    );
+  }}
+</Playground>
+
 ## Related components
 
 - To present non-blocking, contextual, text-only content, use a
--- a/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
+++ b/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
@@ -22,7 +22,6 @@ interface ConfirmationModalState {
 interface ConfirmOrCancelAction {
   type: "confirm" | "cancel";
 }
-
 interface DisplayAction {
   type: "display";
   title?: string;
@@ -108,6 +107,11 @@ interface BaseConfirmationModalProps {
   readonly cancelLabel?: string;
 
   /**
+   * Type (Work or destructive) for confirm button.
+   */
+  readonly variation?: "work" | "destructive";
+
+  /**
    * Callback for when the confirm button is pressed.
    */
   onConfirm?(): void;
@@ -148,6 +152,7 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
     onConfirm,
     onCancel,
     onRequestClose,
+    variation = "work",
   }: ConfirmationModalProps,
   ref: Ref<ConfirmationModalRef>,
 ) {
@@ -206,6 +211,7 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
       dismissible={false}
       primaryAction={{
         label: state.confirmLabel,
+        variation: variation === "destructive" ? "destructive" : "work",
         onClick: () => {
           dispatch({ type: "confirm" });
           onRequestClose && onRequestClose();
--- a/packages/components/src/ConfirmationModal/tests/interaction-ConfirmationModal.test.tsx
+++ b/packages/components/src/ConfirmationModal/tests/interaction-ConfirmationModal.test.tsx
@@ -134,3 +134,43 @@ test("message should handle markdown", () => {
 
   expect(getByText("hello")).toBeInstanceOf(HTMLHeadingElement);
 });
+
+test("should have work type button by default", () => {
+  const { getByText } = render(
+    <ConfirmationModal
+      title="Should we?"
+      message="Do something…"
+      open={true}
+      confirmLabel="okay"
+    />,
+  );
+  expect(getByText("okay").parentElement).toHaveClass("work");
+});
+
+test("destructive type should have destructive confirmation button", () => {
+  const { getByText } = render(
+    <ConfirmationModal
+      title="Should we?"
+      message="Do something…"
+      open={true}
+      confirmLabel="Pull the plug"
+      variation="destructive"
+    />,
+  );
+
+  expect(getByText("Pull the plug").parentElement).toHaveClass("destructive");
+});
+
+test("Work should have work styled confirmation button", () => {
+  const { getByText } = render(
+    <ConfirmationModal
+      title="Should we?"
+      message="Do something…"
+      open={true}
+      confirmLabel="Ok"
+      variation="work"
+    />,
+  );
+
+  expect(getByText("Ok").parentElement).toHaveClass("work");
+});


packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -107,6 +111,11 @@ interface BaseConfirmationModalProps {
    */
   readonly cancelLabel?: string;
 
+  /**
+   * Action Type (Work or destructive) for confirm button.
+   */
+  readonly actionType?: ActionType;
```
eddysims, line 117: I think that the name `actionType` might be a bit to specific, i think something more semantic like just `type` would be more appropriate. 

Reason being is that if we were to sometime in the future change the color of the text to red, we could do that by saying `type="destructive"`. Whereas `actionType="destructive"` would only describe the button.

eddysims: Another note here is that you have a type issue. You are saying that `ActionType` is a type of `type` and then `actionType` is a type of `ActionType`. To use this prop we would need to do

```
actionType={{type: 'destructive'}}
```

I think it makes more sense to just have `readonly type?: "work" | "descructive"`

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -148,6 +157,7 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
     onConfirm,
     onCancel,
     onRequestClose,
+    actionType,
```
eddysims, line 160: can we set a default here? We _usually_ want a default value for most optional props.

```tsx
...
onRequestClose
actionType = 'work'
...
```

eddysims (changes requested):

eddysims:
Left some comments, also can we get a test for this

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
chris-at-jobber: I think I'd default to `work`, if only to avoid this being a breaking change

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
deepvishwas: 👍🏻

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
```diff
@@ -40,6 +40,7 @@ import {
           onConfirm={() => alert("✅")}
```
chris-at-jobber, line 40: I'd like to change this sandbox example up a bit:

```
() => {
  const [open, setOpen] = useState(false)
  return (
    <>
      <Button type="secondary" variation="destructive" label="Delete Bob" onClick={() => setOpen(true)} />
      <ConfirmationModal
        title="Delete Bob?"
        message={`Deleting Bob will remove their data from your account for good.`}
        confirmLabel="Delete Bob"
        open={open}
        onConfirm={() => alert('Bob has been deleted')}
        onCancel={() => alert('Bob will not be deleted')}
        onRequestClose={() => setOpen(false)}
        type="destructive"
      />
    </>
  )
}
```

As I think it's a clearer demonstration of the use case this way

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -107,6 +106,11 @@ interface BaseConfirmationModalProps {
    */
   readonly cancelLabel?: string;
 
+  /**
+   * Action Type (Work or destructive) for confirm button.
+   */
+  readonly type?: "work" | "destructive" | "learning" | undefined;
```
chris-at-jobber, line 112: I wouldn't expect there to be a learning variation of the ConfirmationModal.. I'm not sure if we explicitly need to say you _can't_ use `learning`, but I would _prefer_ it if we can restrict it to just work and destructive for now.

chris-at-jobber:
Destructive version looks good! Would like to see the sandbox example updated a bit and remove the `learning` variation. 
image

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
deepvishwas: the type is passed down to the variation of button type, to keep it simple I kept the learning option.  🤔

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
deepvishwas: Sure thing !

packages/components/src/Button/Button.tsx
```diff
@@ -54,7 +54,7 @@ interface ButtonLinkProps extends ButtonFoundationProps {
 }
 
 interface BaseActionProps extends ButtonFoundationProps {
-  readonly variation?: "work" | "learning";
+  readonly variation?: "work" | "learning" | "destructive";
```
eddysims, line 57: This shouldn't require changes to `Button`. There are currently restrictions on Button that make us not want this option available here.

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
eddysims: I think we would only want the 2 options available. If learning is not available on here, that means it can not be passed down (which is what we want.

also, by adding the `?` in `type?:` you are saying that this prop is optional. Means that you do not need to also allow for `undefined`

eddysims (changes requested):

packages/components/src/Button/Button.tsx
deepvishwas: Fair point but without adding this I see no easy way 
image

deepvishwas: Modal's primaryAction is of type ButtonProp which allows variation to be either 'work | learning' or 'destructive'

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
```diff
@@ -26,6 +26,8 @@ import {
 } from "@jobber/components/ConfirmationModal";
 ```
 
+Default type of confirmation Modal
```
eddysims, line 29: We don't need to say `default` as it is the first one on the page. It is assumed that it is the default.

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
```diff
@@ -40,6 +42,7 @@ import {
           onConfirm={() => alert("✅")}
           onCancel={() => alert("🙅‍♂️")}
           onRequestClose={() => setOpen(false)}
+          type="work"
```
eddysims, line 45: This can be left off as it is supplied with a default value of `work`

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
```diff
@@ -109,6 +112,32 @@ that presents a confirm modal.
   }}
 </Playground>
 
+## Destructive
+
+Sometimes when the action is critical or destructive in nature (e.g. Deleting
+user or account), you may want to use `destructive type` modal.
```
eddysims, line 118: I don't think that we need to give an example here. (deleting user of account)

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -107,6 +107,14 @@ interface BaseConfirmationModalProps {
    */
   readonly cancelLabel?: string;
 
+  /**
+   * Type (Work or destructive) for confirm button.
+   */
+  readonly type?: Exclude<
+    BaseActionProps["variation"] | DestructiveActionProps["variation"],
+    "learning"
+  >;
```
eddysims, line 116: the more i think about this, I don't think we actually want to couple this to the button types. The reason being is the same as my comment around the naming `actionType`. We may in the future want the `type` to represent something other than just the button variation. I think we should actaully just go with `"work" | "destructive"` here.

I was also thinking that this may make more sense to be named `variation` instead of `type`.

Thoughts?

eddysims (changes requested):

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
```diff
@@ -109,6 +112,32 @@ that presents a confirm modal.
...
+
+<Playground>
+  {() => {
+    const [open, setOpen] = useState(false);
+    return (
+      <>
+        <Button label="Open" onClick={() => setOpen(true)} />
```
chris-at-jobber, line 125:
```suggestion
        <Button label="Delete Bob" variation="destructive" type="secondary" onClick={() => setOpen(true)} />
```

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
chris-at-jobber: Variation maps more closely to the way we describe the work and destructive options of Button, seems reasonable

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
chris-at-jobber:
```suggestion
If the user is confirming an action that is will destroy, delete, or remove something, use a destructive ConfirmationModal to visually reinforce the potential consequences.
```

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
deepvishwas: I would like to keep it simple too ("work" | "destructive") but this variation value is assigned to the primary action of modal (which is button variation), if we could find a way around it that would be awesome.

packages/components/src/Button/Button.mdx
```diff
@@ -188,11 +188,36 @@ such as a Modal header.
         <Button label="Cancel" type="tertiary" variation="subtle" />
       </Content>
     </Content>
-    <div style={{ backgroundColor:"var(--color-surface--background)", padding:"var(--space-base)" }}>
```
deepvishwas, line 191: Please ignore this file it's autoformatted

packages/components/src/InputEmail/InputEmail.tsx
```diff
@@ -21,7 +21,8 @@ export function InputEmail(props: InputEmailProps) {
   );
 
   function checkForValidEmail(value: string) {
-    const emailRegex = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+    const emailRegex =
```
deepvishwas, line 24: Please ignore this file it's autoformatted

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
```diff
@@ -26,6 +26,8 @@ import {
 } from "@jobber/components/ConfirmationModal";
 ```
 
+Confirmation Modal
```
chris-at-jobber, line 29:
```suggestion
```

chris-at-jobber: (don't need this line, page title is sufficient)

packages/components/src/Button/Button.tsx
```diff
@@ -56,11 +56,11 @@ interface ButtonLinkProps extends ButtonFoundationProps {
 }
 
 interface BaseActionProps extends ButtonFoundationProps {
-  readonly variation?: "work" | "learning" | "subtle";
+  readonly variation?: "work" | "learning" | "subtle" | "destructive";
```
chris-at-jobber, line 59: Out of curiousity, why do we need to change Button? We still have `DestructiveActionProps` below

chris-at-jobber:
It looks like we [got rid of learning as a variation for this component](https://github.com/GetJobber/atlantis/pull/633/files#diff-84c051289a927ab62741ce345385b6eec95d1d66d4c93b977eca97bfc06f94edR112) (which is good), but I'm still seeing the ability to create a `learning` confirmationModal:
image

packages/components/src/Button/Button.tsx
deepvishwas: The destructive variation that we created will be assigned to the variation of the primary action prop of Modal. Which is a weird combo of Button props, that can take either BaseActionProps or Destructive action props but not both.  

image

deepvishwas:
> It looks like we [got rid of learning as a variation for this component](https://github.com/GetJobber/atlantis/pull/633/files#diff-84c051289a927ab62741ce345385b6eec95d1d66d4c93b977eca97bfc06f94edR112) (which is good), but I'm still seeing the ability to create a `learning` confirmationModal:
> image

I think typescript will take care of this while development 
image

MDX playground allows everything e.g. on Atlantis main repo if you try to assign a 'type' prop to the modal which does not exist. You still get the work type modal.
image

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
deepvishwas: removed

chris-at-jobber:
Is it concerning that it works, though? I can set a nonsense variation and it defaults to `work` which makes sense:
image

But when I set it to `learning`, which shouldn't work, it changes the primaryAction to a learning button:
image

packages/components/src/ConfirmationModal/ConfirmationModal.mdx
```diff
@@ -109,6 +109,38 @@ that presents a confirm modal.
   }}
 </Playground>
 
+## Destructive
+
+If the user is confirming an action that is will destroy, delete, or remove
```
adambobadam, line 114: Very minor/nitpicky: Can remove "is" from sentence.  So, `If the user is confirming an action that will destroy, delete, or remove`

deepvishwas: I had that doubt too, checked with Grammarly and it gave me okay :P

deepvishwas:
@eddysims  @chris-at-jobber  - updated the branch. There could be a possible miss or two while resolving conflicts  😬

chris-at-jobber (dismissed):
`variation="learning"` should not work when used for ConfirmationModal

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -206,6 +211,7 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
       dismissible={false}
       primaryAction={{
         label: state.confirmLabel,
+        variation: variation,
```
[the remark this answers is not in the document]
deepvishwas: > Still seeing that I can successfully set `variation="learning"`:
> image

Same as here https://github.com/GetJobber/atlantis/pull/633#issuecomment-920965531

While development Typescript will take care of this. mdx does not follow the rules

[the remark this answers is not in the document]
deepvishwas: nonetheless 
adding the check

chris-at-jobber:
Nice! When I set `learning` it defaults to `work` 👌 

image

eddysims (approved):

PR #634, opened 2021-09-03
fix(components): Trap focus on the open modal [JOB-26477]

darryltec:
## Motivations

- When a modal is open, it should keep the tabbing between elements within the modal as per Accessibility Guidelines.

### Added

- Added a hook that traps the focus within the target element and can be reuseable if we need to.
- When a modal is open, 
  - It adds an `aria-hidden="true"` on the body
  - It adds an `aria-hidden="false"` on the modal itself

### Changed

- Use hook for the esc key listener

### Deprecated

- 

### Removed

- 

### Fixed

- Pressing shift+tab on a modal goes to the last focusable element within the modal instead of going outside of it
- Pressing tab when focused on the last focusable element goes to the first focusable element (which is the modal container) instead of going outside of it

### Security

- 

## Testing

- Open a modal
- press tab or shift+tab repeatedly
- the focus shouldn't go outside of the modal

---

Random photo of Atlantis


darryltec:
reintroduced the issue with modal removing focus on the textfield. don't review yet

seanhealy:
Ok, I'll wait. 🍻

darryltec:
we good @seanhealy. review away!!!!!! (miss you 😢 )

chris-at-jobber:
Tested with Voiceover:

- [x] Sets focus as expected
- [x] Prevents focus from leaving the Modal

There's the outstanding issue of focus not returning to the trigger element, but that is existing and I believe we have discussed tackling that in a separate PR, correct?

darryltec:
> There's the outstanding issue of focus not returning to the trigger element, but that is existing and I believe we have discussed tackling that in a separate PR, correct?

yerp!

chris-at-jobber (dismissed):
Design approved ✅

darryltec:
> There's the outstanding issue of focus not returning to the trigger element, but that is existing and I believe we have discussed tackling that in a separate PR, correct?

PR here: https://github.com/GetJobber/atlantis/pull/638

joelwarrington-jobber (approved):
LGTM

https://user-images.githubusercontent.com/80279872/132926614-75a845bd-6847-4c68-bfc2-1f8ce7ba4b60.mov

https://user-images.githubusercontent.com/80279872/132928308-9fa6d095-06df-4aeb-816a-b4a4af46d4f9.mov

## Fast-Follow PR candidates
- Bringing the `useFocusTrap` hook into our hook library

darryltec:
> Bringing the useFocusTrap hook into our hook library

I'd keep it where it is.
1. Prevent unwanted uses for it. Using this outside of the modal doesn't make sense yet without that context, you're pretty much limiting the accessibility of things. 
2. Chris and I can't think of any other place where we have to trap focus other than a modal. We've ruled out popover and menu.

Putting this on the hook library sounds like an "In case we need it" scenario. It's a nice sentiment, but also a double edge sword. If you can think of a feature that would use this properly in the nearest future, let me know. Otherwise, I'd keep it within the modal.

PR #638, opened 2021-09-07
fix(components): Create a hook to focus back on modal and popover activator

darryltec:
## Motivations

Focus back to the element that opened the modal/popover when they close to Improve the accessibility of it for keyboard users.

### Added

- A hook called `useRefocusOnActivator` that puts the focus back on the thing that opened the modal/popover when they gets closed
- Added documentation and test for said hook

### Changed

- Implemented `useRefocusOnActivator` hook on Modal and Popover

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Open a modal or popover
- Close modal or popover via keyboard tabbing/shortcuts or clicking close
- It should bring back the focus to the one that opened it

## Limitations
- Only applies to element that can be focused
- The activator has to open the modal/popover in order for it to be able to focus back
  - This won't work if you programmatically open a dialog by any other means as that would mean it wasn't opened by a user intent.

---

Random photo of Atlantis

### DIFF
--- a/packages/hooks/src/useRefocusOnActivator/useRefocusOnActivator.test.tsx
+++ b/packages/hooks/src/useRefocusOnActivator/useRefocusOnActivator.test.tsx
@@ -1,36 +1,87 @@
-import React, { useState } from "react";
+import React, { ReactNode, useState } from "react";
 import { fireEvent, render } from "@testing-library/react";
 import { useRefocusOnActivator } from ".";
 
-const activatorId = "activator";
-const closerId = "closer";
+function waitForNextTick(callback: () => void) {
+  // Need to wait for the DOM to finish it's thing
+  // https://github.com/testing-library/jest-dom/issues/53
+  // Based on this finding [https://github.com/testing-library/jest-dom/issues/53#issuecomment-421671377]
+  // and [https://stackoverflow.com/a/11299876] the focus always goes back to
+  // the < body > before focusing on the correct element. In turn, there's a
+  // delay and you have to wait for a tick to get the right element.
+  setTimeout(callback, 20);
+}
 
-it("return the focus on the activator", () => {
-  const { getByTestId } = render(<TestComponent />);
+it("should return the focus on the activator", () => {
+  const activatorId = "activator";
+  const closerId = "closer";
+  const { getByTestId } = render(
+    <TestComponent activatorId={activatorId} closerId={closerId} />,
+  );
   fireEvent.click(getByTestId(activatorId));
   fireEvent.click(getByTestId(closerId));
 
-  // Run on next tick
-  setTimeout(() => {
+  waitForNextTick(() => {
     expect(getByTestId(activatorId)).toHaveFocus();
-  }, 0);
+  });
 });
 
-function TestComponent() {
+describe("Nested components with return focus hook", () => {
+  it("should return the focus on the correct activator", () => {
+    const activatorId = "activator";
+    const closerId = "closer";
+    const childActivatorId = "activator-child";
+    const childCloserId = "closer-child";
+    const { getByTestId } = render(
+      <TestComponent activatorId={activatorId} closerId={closerId}>
+        <TestComponent
+          activatorId={childActivatorId}
+          closerId={childCloserId}
+        />
+      </TestComponent>,
+    );
+    fireEvent.click(getByTestId(activatorId));
+    fireEvent.click(getByTestId(childActivatorId));
+    expect(getByTestId(childCloserId)).toBeInTheDocument();
+
+    fireEvent.click(getByTestId(childCloserId));
+    waitForNextTick(() => {
+      expect(getByTestId(childActivatorId)).toHaveFocus();
+    });
+
+    fireEvent.click(getByTestId(closerId));
+    waitForNextTick(() => {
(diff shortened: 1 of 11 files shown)


chris-at-jobber:
Seeing some interesting behaviour...

- go to page with component that uses this hook (ie Modal or Popover)
- open and close component the first time
- focus does not reset
- open and close component a second, third, fourth, fifth etc...
- focus _does_ reset

chris-at-jobber (changes requested):
See comment (does not work the first time you attempt it following a fresh pageload, works all subsequent times)

packages/hooks/src/useRefocusOnActivator/useRefocusOnActivator.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+function waitForNextTick(callback: () => void) {
+  // Need to wait for the DOM to finish it's thing
+  // https://github.com/testing-library/jest-dom/issues/53
+  // Based on this finding [https://github.com/testing-library/jest-dom/issues/53#issuecomment-421671377]
+  // and [https://stackoverflow.com/a/11299876] the focus always goes back to
+  // the < body > before focusing on the correct element. In turn, there's a
+  // delay and you have to wait for a tick to get the right element.
+  setTimeout(callback, 20);
+}
```
darryltec: Looking for alternatives for this. If you have a suggestion, let me know!

packages/hooks/src/useRefocusOnActivator/useRefocusOnActivator.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useRefocusOnActivator } from ".";
+
+function waitForNextTick(callback: () => void) {
+  // https://github.com/testing-library/jest-dom/issues/53
+  // Based on this finding [https://github.com/testing-library/jest-dom/issues/53#issuecomment-421671377]
+  // and [https://stackoverflow.com/a/11299876] the focus always goes back to
+  // the <body> before focusing on the correct element. In turn, there's a
+  // delay and you have to wait for a tick to get the right element.
+  setTimeout(callback, 20);
+}
```
darryltec, line 12: Looking for alternatives for this. If you have an idea, let me know!

adambobadam:
Hey @darryltec!  I am noticing that when I hide the card there and return the focus to the button it shifts the page to the left and back, I wonder why this is?

packages/hooks/src/useRefocusOnActivator/useRefocusOnActivator.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  return (
+    <>
+      <button data-testid={activatorId} onClick={() => setOpen(true)}>
+        Click me!
+      </button>
+      {open && (
+        <div>
+          Haloo!
+          <button onClick={() => setOpen(false)} data-testid={closerId}>
```
adambobadam, line 80: Neat, seems pretty clever!

packages/hooks/src/useRefocusOnActivator/useRefocusOnActivator.ts
```diff
@@ -0,0 +1,26 @@
...
+  useEffect(() => {
+    let activator: Element | null | undefined;
+
+    if (active && !activator) {
+      activator = document.activeElement;
+    }
+
+    return () => {
+      if (active) {
+        if (activator instanceof HTMLElement) {
```
adambobadam, line 19: What other types could `activator` be here?

Also, just a minor/nitpicky thing:  could two nested ifs make an `and` here maybe? 🤔

darryltec:
> Hey @darryltec! I am noticing in the netlify preview that when I hide the card there and return the focus to the button it shifts the page to the left and back, I wonder why this is?

Gonna assume it's the scrollbar showing up and then disappearing.

packages/hooks/src/useRefocusOnActivator/useRefocusOnActivator.ts
darryltec: hmm.. kinda torn between that tho. The `document.activeElement` is of an `Element` type which doesn't have the focus function. Instead of casting it as `as HTMLElement` I went with the safer way of making sure it is actually an `HTMLElement`. 

But yeah if it does return something that is not an `HTMLElement` the activator won't get wiped. To answer your question of what other types are out there, i don't know, hence it's separate haha.

adambobadam: Okay!  That makes sense I think!

adambobadam:
> > Hey @darryltec! I am noticing in the netlify preview that when I hide the card there and return the focus to the button it shifts the page to the left and back, I wonder why this is?
> 
> Gonna assume it's the scrollbar showing up and then disappearing.

Hmm ya, I wonder if it's the card opening up there then

adambobadam (approved):

adambobadam:
Just approving assuming Chris' changes are addressed :)

chris-at-jobber:
I did a QA pass and it looks like the issue I found is fixed! Nice work @darryltec

PR #639, opened 2021-09-08
build: Prevent manual releases

eddysims:
## Motivations

In Atlantis 🔱, we want to ensure that there are never any direct version bumps to any of the @jobber packages. The reason being is that if a version is manually bumped, lerna will have a struggle to try and figure out what the next proper version to release is.

### Added

- a new script to check for manual releases

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,22 @@ jobs:
           paths:
             - ~/.npm
 
+  check_for_manual_release:
+    <<: *container_config
+    steps:
+      - *restore_repo
+      - *sum_lockfiles
+      - *restore_npm_cache
+      - *npm_install
+      - run:
+          name: Check for manual releases
+          command: |
+            if git log --max-count=1 --pretty=%s | grep -q "chore(release): Publish"; then
+              echo "Skipping run on release commit."
+            else
+              npm run preventManualRelease
+            fi
+
   lint_javascript:
     <<: *container_config
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -187,18 +187,21 @@ workflows:
       - install_dependencies:
           requires:
             - checkout_code
+      - check_for_manual_release:
+          requires:
+            - install_dependencies
       - lint_javascript:
           requires:
-            - install_dependencies
+            - check_for_manual_release
       - lint_css:
           requires:
-            - install_dependencies
+            - check_for_manual_release
       - test:
           requires:
-            - install_dependencies
+            - check_for_manual_release
       - build_docs:
           requires:
-            - install_dependencies
+            - check_for_manual_release
       - release:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,6 +82,7 @@ jobs:
       - run:
           name: Check for manual releases
           command: |
+            set -e
             if git log --max-count=1 --pretty=%s | grep -q "chore(release): Publish"; then
               echo "Skipping run on release commit."
             else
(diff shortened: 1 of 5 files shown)


scripts/preventManualRelease.js
```diff
@@ -0,0 +1,133 @@
...
+          process.exit(1);
+        }
+      })
+      .catch(error => {
+        if (error.response.data.message) {
+          console.log(chalk.red(error.response.data.message));
+          process.exit(1);
+        } else {
+          console.log(chalk.red("Something went wrong"));
+          process.exit(1);
```
adambobadam, line 83: Minor/nitpicky:  Could `process.exit(1)` be moved outside of the if/else since you do it regardless of the condition?

.circleci/config.yml
```diff
@@ -72,6 +72,23 @@ jobs:
...
+      - *restore_npm_cache
+      - *npm_install
+      - run:
+          name: Check for manual releases
+          command: |
+            set -e
+            if git log --max-count=1 --pretty=%s | grep -q "chore(release): Publish"; then
+              echo "Skipping run on release commit."
+            else
+              npm run preventManualRelease
```
adambobadam, line 89: Trying to parse, but brain does not compute 😭, what is this command doing here?

scripts/preventManualRelease.js
```diff
@@ -0,0 +1,133 @@
...
+        const versions = Object.keys(time);
+
+        if (versions.includes(version)) {
+          console.log(
+            chalk.green(`✅ ${name}@${version} is currently in npm registry`),
+          );
+        } else {
+          console.log(
+            chalk.red(
+              `${name}@${version} was not found in the npm registry. Please check that you have not committed a new version to your branch`,
```
adambobadam, line 71: Neat! So this should be able to handle earlier versions too hey?

adambobadam:
Okay, when I ran `npm ci` locally lerna said it was successful after bumping the version manually, but when I pushed that version to github lerna freaked out.  I haven't been able to see those errors you are printing out yet.  How would I go about seeing them?

scripts/preventManualRelease.js
eddysims: This will check that the version is currently on the npm registry. We do not want to allow anyone to commit a version that is not already on the npm registry.

.circleci/config.yml
eddysims: This is saying that if the commit message is "chore(release): Publish" then don't run this script. The reason for this is we want the publish step (which is handled by lerna) to be able to commit the new versions of each package into master.

eddysims:
> Okay, when I ran `npm ci` locally lerna said it was successful after bumping the version manually, but when I pushed that incorrect version to github to my own branch (see here: #640) lerna correctly freaked out! I haven't been able to see those errors you are displaying yet. How would I go about seeing them? 🙂

Hey @adambobadam to test it you can look at the CI job on https://github.com/GetJobber/atlantis/pull/639/commits/9dd1fa0fac6695026b848d7db3841e0de75ea51f or you can test it locally by going into `./src/components/package.json`, then change the version to something like `5.1.10` and then run `npm run preventManualRelease`

.circleci/config.yml
adambobadam: oooooh I see, probably why the script wasn't running for me when I committed to my branch!

scripts/preventManualRelease.js
adambobadam: Makes sense!!!

.circleci/config.yml
adambobadam: Okay, that makes sense!

adambobadam:
LGTM!  

Manually tested the script and then tested it on my branch as well, seems to work as expected 👍

adambobadam (approved):

PR #643, opened 2021-09-10
fix(components): properly style disabled primary buttons

chris-at-jobber:
## Motivations

Some of the `:not` specificity introduced in adding subtle buttons is over-riding the styling of primary disabled buttons.
image

## Changes

Include disabled buttons in the specificity around primary buttons to ensure correct styling is applied

### Fixed

- Primary buttons look as expected when disabled: 
image

---

Random photo of Atlantis


darryltec (approved):
Can confirm. looks disabled. There's probably a way to avoid the whole comma thing, but let's leave that to the time where fully delete the cancel from existing

PR #644, opened 2021-09-10
fix(hooks): Temporarily delete flakey hook test

darryltec:
## Motivations

Added a hook to refocus back to the element that was opened. As I was writing a test, I encountered a dilemma where it takes a few moments for the element i need to be focused so I went with a silly solution that backfired. Deleting the test until further notice.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- useRefocusOnActivator test was removed for its flakey-ness

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


adambobadam:
@darryltec and I were investigating if we skip flakey tests in Atlantis.  But we don't seem to be doing any skips on tests in Atlantis and only in Jobber.

adambobadam (approved):

PR #648, opened 2021-09-13
fix(components): remove aria-hidden prop on trap focus

darryltec:
## Motivations

Upon adding the modal to trap the focus (#634) I've added an `aria-hidden=true` on the body and `false` on the modal. I saw an example of that when dealing with tooltips and assumed that was one of the way to deal with hiding things behind the modal. Trapping still works tho.

So the `getByRole` (`findBy...`, `queryBy...`) [respects the `aria-hidden`](https://testing-library.com/docs/queries/byrole/#hidden), thus breaking all modal tests in Jobber Online from getting/finding the things inside the modal as it was aria-hidden through the body. It was just the test that was broken. UI wise, it all be working: trapping and screen readers.

Lesson: Confirm how aria-hidden works and pre-release test atlantis in jobber online (clicking around and running tests)

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- The use of aria-hidden prop. Needs more exploration on that front.

### Fixed

- 

### Security

- 

## Testing

- Install the prerelease in jobber `npm i @jobber/components@2.46.7-pre.0+276c588`
- Run test in Jobber online
- There should be no broken test
- Still in Jobber online
- Open any Atlantis modal
- Tab around - The focus should stay within the dialog

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Modal/useFocusTrap.test.tsx
+++ b/packages/components/src/Modal/useFocusTrap.test.tsx
@@ -37,20 +37,6 @@ it("should not trap the tabbing and focus on the first child node", () => {
   expect(getByTestId(targetId).previousElementSibling).toHaveFocus();
 });
 
-it("should have the proper aria-hidden on body and ref target when active trapping focus", () => {
-  const { getByTestId } = render(<TestComponent />);
-  expect(document.body.hasAttribute("aria-hidden")).toBeTruthy();
-  expect(document.body.getAttribute("aria-hidden")).toBe("true");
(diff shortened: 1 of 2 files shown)


chris-at-jobber:
QA  on behaviour the web app looks good, I will defer testing the tests to the professionals.

### Focus the Modal trigger
image

### Modal opens with focus set to the Modal (first `tab` takes me to the dismiss button)
image

### When I close the Modal, focus returns to the trigger
image

eddysims (approved):

PR #649, opened 2021-09-14
feat(components): Ability to wrap content with Spinner

joelwarrington-jobber:
## Motivations

See https://github.com/GetJobber/atlantis/pull/580

### Added

- Added ability to specify child content for a Spinner
- Added ability to conditionally display Spinner

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Spinner/BaseSpinner.tsx
+++ b/packages/components/src/Spinner/BaseSpinner.tsx
@@ -1,0 +1,34 @@
+import React from "react";
+import classnames from "classnames";
+import styles from "./Spinner.css";
+
+export interface BaseSpinnerProps {
+  /**
+   * Specifies the size of the spinner
+   *
+   * @default "base"
+   */
+  readonly size?: "small" | "base";
+
+  /**
+   * Spinner becomes an inline element when true,
+   * otherwise it is a block element
+   */
+  readonly inline?: boolean;
+}
+
+export function BaseSpinner({ size = "base", inline }: BaseSpinnerProps) {
+  const spinnerStyles = classnames(styles.spinner, {
+    [styles.small]: size === "small",
+    [styles.inline]: inline,
+  });
+
+  return (
+    <div
+      className={spinnerStyles}
+      role="alert"
+      aria-busy={true}
+      aria-label="loading"
+    />
+  );
+}
--- a/packages/components/src/Spinner/Spinner.tsx
+++ b/packages/components/src/Spinner/Spinner.tsx
@@ -1,34 +1,45 @@
-import React from "react";
+import React, { ReactNode } from "react";
 import classnames from "classnames";
 import styles from "./Spinner.css";
+import { BaseSpinner, BaseSpinnerProps } from "./BaseSpinner";
 
-interface SpinnerProps {
+interface SpinnerProps extends BaseSpinnerProps {
   /**
    * Specifies the size of the spinner
    *
-   * @default "base"
+   * @default true
    */
-  readonly size?: "small" | "base";
+  readonly loading?: boolean;
 
   /**
-   * Spinner becomes an inline element when true,
-   * otherwise it is a block element
+   * The content to wrap with the loading spinner
    */
-  readonly inline?: boolean;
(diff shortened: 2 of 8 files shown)


chris-at-jobber:
@joelwarrington-jobber I really liked this smart [sizing functionality](https://github.com/GetJobber/atlantis/pull/580/files?file-filters%5B%5D=.css&file-filters%5B%5D=.snap&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-2d706a804fe321d6217cf9202f3d2002630fb467add394f7fdd658664a579ab0R42) from the Loading PR, can we bring that back?

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -1,34 +1,45 @@
-import React from "react";
+import React, { ReactNode } from "react";
 import classnames from "classnames";
 import styles from "./Spinner.css";
+import { BaseSpinner, BaseSpinnerProps } from "./BaseSpinner";
 
-interface SpinnerProps {
+interface SpinnerProps extends BaseSpinnerProps {
   /**
    * Specifies the size of the spinner
```
chris-at-jobber, line 8: This comment line should move with the size prop

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -1,34 +1,45 @@
...
    *
-   * @default "base"
+   * @default true
    */
-  readonly size?: "small" | "base";
+  readonly loading?: boolean;
```
chris-at-jobber, line 12: I get what this prop is doing, but I would get some feedback from other engineers on this name. I might be overthinking it because we were just talking about `Loaders` before this, but getting some end-user thoughts would be nice. 

As an aside, it doesn't look like this has a default but I believe this should default to `false` as I'm wondering if we could get into a state where this component ends up "perma-loading" even once we have content, if the end-user dev doesn't set an explicit "false" when loading is not true (which is hopefully _most_ of the time_). Or is this fine, because it's an optional prop so it's only coming into play when an engineer is explicitly using it?

(No specific "change requested" here, just some ideas that might be worth discussing)

packages/components/src/Spinner/BaseSpinner.tsx
```diff
@@ -0,0 +1,34 @@
...
+   *
+   * @default "base"
+   */
+  readonly size?: "small" | "base";
+
+  /**
+   * Spinner becomes an inline element when true,
+   * otherwise it is a block element
+   */
+  readonly inline?: boolean;
```
chris-at-jobber, line 17: Thinking about this prop and I'm wondering if we could remove this as a property and make it dependent on whether or not `children` are present in the <Spinner>? (This might be a terrible idea, just wanted to discuss it)

chris-at-jobber (changes requested):
Only specifically requested change is bringing back the smart resize based on container size from the old PR; rest of my comments are discussion points that an engineer is definitely better qualified than myself to answer 😄

joelwarrington-jobber:
@chris-at-jobber I think a coffee & chat would be best to review your comments and get thoughts, to which I can add follow-up stuff here

darryltec:
I'd like to be in that conversation. I would like to understand what this is for and the problem we're trying to solve. If there's a proposal, that would be even better as I am not a fan of having to wrap everything in page spinner. Have we looked at how other design systems have done this?

joelwarrington-jobber:
@darryltec the original intent is to communicate loading while being able to display static content, potentially used with something like a skeleton/placeholder.

The only example the loading spinner wrapping content appears to be from semantic UI.
- https://semantic-ui.com/elements/loader.html

However now that I'm thinking about it, maybe it could be better to define strategy for loading placeholder content. Such as https://semantic-ui.com/elements/placeholder.html

joelwarrington-jobber:
Per our discussion, going to close this.

- Wrapping content with a loader seems to a be a bad practice we don't want to push going forward
- Instead, we may want to instead promote usage of placeholder/skeleton content such as what could be implemented by #660

PR #651, opened 2021-09-15
refactor(components): BREAKING CHANGE: remove cancel from button and uses of button [JOB-37117]

chris-at-jobber:
## Motivations

We have removed all instances of `cancel` in uses of Atlantis button; this removes the cancel variation from Button itself 

### Removed

- cancel, replaced with subtle

---

Random photo of Atlantis


joelwarrington-jobber:
@chris-at-jobber maybe this should be marked as a breaking change, so it triggers a major version bump

joelwarrington-jobber (changes requested):
@chris-at-jobber can you merge latest master in to resolve conflicts?

Afterwards, should be okay to review, and likely approve!

chris-at-jobber:
@joelwarrington-jobber if you have some time this is ready for a look!

packages/components/src/Button/Button.mdx
```diff
@@ -182,10 +182,10 @@ elements such as a Modal header.
         <Button label="Cancel" type="primary" variation="subtle" />
       </Content>
       <Content>
-        <Button label="Cancel" type="secondary" variation="subtle" />
+        <Button label="Dismiss" type="secondary" variation="subtle" />
```
chris-at-jobber, line 185: Showing a broader range of examples

packages/components/src/Button/Button.tsx
```diff
@@ -80,7 +81,7 @@ interface SubmitButtonProps
 
 export type ButtonProps = XOR<
   BaseActionProps,
-  XOR<CancelActionProps, SubmitActionProps>
+  XOR<DestructiveActionProps, SubmitActionProps>
```
chris-at-jobber, line 84: I think this is appropriate use of XOR?

joelwarrington-jobber (approved):
LGTM

`cancel` variation no longer works and is replaced by `subtle`
image
image

ConfirmationModal working as expected:
image


Have also confirmed locally that the Typing of Button is accurate to what we want.
- Error when trying to use `cancel`
- `type` restrictions on Submit type button

adambobadam:
LGTM, just wondering why `cancel` is beeing yeeted in the first place.  I.e., are there any instances where using `subtle` instead of `cancel` wouldn't make sense? Or where `cancel` might still be needed?

chris-at-jobber:
@adambobadam more details on the initial work [here!](https://jobber.atlassian.net/browse/JOB-34785)

adambobadam:
> @adambobadam more details on the initial work [here!](https://jobber.atlassian.net/browse/JOB-34785)

Thanks for that! 🙏

chris-at-jobber:
Also chatted offline w/ @adambobadam 

image

PR #668, opened 2021-09-20
feat(design): add Typography letter spacing to @jobber/design [JOB-29564]

hosseintalebi:
## Motivations

To allow mobile to pull in letter spacing (originally it was intended for moving line height and other typography values such that it can be used in mobile, but there were complications with the media type in React Native; more info [here](https://jobber.atlassian.net/browse/JOB-36053))

## Changes

- Add letter spacing values to @jobber/design

Here are the prereleases to test:
 - @jobber/components@2.46.10-pre.7+633f965
 - @jobber/design@0.17.1-pre.28+633f965
 - @jobber/docz-tools@1.7.13-pre.7+633f965

Random photo of Atlantis


packages/design/src/Typography.mdx
```diff
@@ -0,0 +1,104 @@
...
+
+import { Playground } from "docz";
+import { Heading } from "@jobber/components/Heading";
+import { Text } from "@jobber/components/Text";
+import { Emphasis } from "@jobber/components/Emphasis";
+import { Content } from "@jobber/components/Content";
+import { Button } from "@jobber/components/Button";
+import { Checkbox } from "@jobber/components/Checkbox";
+
+# Typography
```
chris-at-jobber, line 15: @hosseintalebi I don't want to get into the specific values/tokens throughout this doc, as I think that's more of a concern for the typography component(s); my thinking is more to describe the main components consumers will use, and the rationale of how we arrived at the values those components are built with.

chris-at-jobber:
Getting some outside designer eyes on my words

packages/design/src/colors.css
```diff
@@ -170,5 +170,5 @@
   --color-indigo--dark: rgb(55, 69, 132);
 
   --color-white--rgb: 255, 255, 255;
-  --color-white: rgba(var(--color-white--rgb));
+  --color-white: rgb(var(--color-white--rgb));
```
chris-at-jobber, line 173: This spooks me because I have no idea how this was working before... I assume if RGBA doesn't get an argument for `a` it falls back to 100%?

chris-at-jobber:
Design comments from @paco-mtz-design :

> Under Font Families > Base, you describe that Source Sans Pro "is used for body copy, most headings, data display, form inputs, labels, and more!". Would it be overkill here to also show a visual example of data display and form inputs? (I kind of missed seeing SSP in action in all those different cases).

> Under Line Height, I got a clear understanding of what leading means, but as a user I might wonder if this is a setting I need to learn how to use? I.e. could you clarify better that this is already baked in the settings for Heading and Text and there's no need to define it explicitly? (at least if that's the case!)

packages/design/src/letterSpacing.css
```diff
@@ -0,0 +1,4 @@
+:root {
+  --typography--letterSpacing-base: 0;
```
lmorgan29, line 2: I just realized this likely belongs in `typography.css` 🤔

packages/design/src/Typography.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+## Line height
+
+Line height, or
+[leading](https://www.shillingtoneducation.com/blog/leading-typography/), is the
+space between subsequent lines of type. Our `Heading` and `Text` components
+handle their line-heights internally, so you don't need to worry specific
+implementation values as long as you're working with those components.
+
+Our base line-height of `1.25` is optimized for reading multiple lines of
```
jlelford, line 123: If possible, would it make sense to inherit the line height appropriate for the size of the font being used? My assumption is while we have options, there is a preferred choice per type size or font?

Just trying to think of ways to avoid minor variation if that choice is presented.

chris-at-jobber: `Heading` and `Text` will never vary in the line-heights that they present for each "level" (for both web and mobile) so I wanted to stay away from specific details here... maybe talking less about the values would be helpful? In Atlantis we actually intentionally decoupled line-height from font-size because of the reasons described later in this paragraph (at one point setting a font-size of `large` would give you a `large` line-height, which gets less useful as you get into larger font-sizes). I'm thinking I could get rid of this whole paragraph and update the previous one to say

> Line height, or [leading](https://www.shillingtoneducation.com/blog/leading-typography/), is the space between subsequent lines of type. The line-heights in the `Heading` and `Text` components are optimized for each variation's unique combination of font-size, weight, and family, so you don't need to adjust their line-height manually when you use them.

chris-at-jobber: Then the subsequent paragraph could read

> At smaller sizes, a more generous line-height relative to the font-size is used to allow the user to more easily parse multiple lines of small text. As font-size increases into "display" sizes such as a level 1 `Heading`, the ratio of line-height to font-size decreases as the larger font-size is more readable and requires less line-to-line breathing room.

jlelford: Yeah, I think that would be 👌 . The previous version almost promoted the customization of it, being clear that it's not required to adjust using this component is solid.

packages/design/src/Typography.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+Our base font-size is based on our [Spacing](/space) scale for a consistent
+sizing rhythm across Jobber. At its' default zoom level, the equivalent is
+`16px`. While this is higher than in some systems where `14` may be more common,
+Source Sans Pro is optically a bit smaller, particularly at the x-height, so we
+start with `16` and work our way up and down from there.
+
+## Line-height
+
+Line-height, or
```
chris-at-jobber, line 117: @jlelford fixed for ya 😺

adambobadam (approved):
LGTM!  Also, based on the other documentation under design (e.g., spacing), should we expose the variables on typography?  I.e., https://github.com/GetJobber/atlantis/blob/d607c8faf2a688f34e34843000e9313195be5388/packages/components/src/Typography/css/Typography.css#L1-L21

chris-at-jobber:
@adambobadam I would prefer [to not do that](https://github.com/GetJobber/atlantis/pull/668#discussion_r712561871)

PR #681, opened 2021-09-24
docs: use inverse examples for remove and delete

chris-at-jobber:
## Motivations

Had a discussion with designers where the example of using the inverse of create was a useful framing for when to use delete, vs remove.

### Added

- A few lines about delete and remove and their relation to create and add.

---

Random photo of Atlantis


jlelford (approved):
LGTM

PR #704, opened 2021-10-01
feat(components): Add keyboard shortcut to close menu on Escape

### LINKED ISSUE #696: Menu: Esc key should close the menu
The `Esc` key does not work as advertised (VoiceOver, for example, understands that Menu is a menu and reads “press the Esc key to close”, but we do not offer this functionality). This means that to close the menu, the user must ensure they have tabbed to the button and then hit return/space to close.

joelwarrington-jobber:
## Motivations

Closes #696

- The escape key does not worked as advertised with the assigned `menu` and `menuitem` roles
- Pressing the escape key should close the menu

### Added

- Added keyboard shortcut listener for `Escape`

### Changed

- Updated and refactored Menu testing structure
- Updated how Page is mocking our hooks so that only the hooks we want to mock will be mocked, and other functions will work as expected.

## Testing

1. Go to the `<Menu />` component page
   1. Using mouse
      - Can click on the menu to open
      - Can click on the menu to close if open
      - Can click on anything other than menu and menu items to close
   2. Using keyboard
      - Can hit `enter` to open the menu
      - Can hit `enter` to close the menu when open and button is focused
      - Can hit `escape` to close the menu when it's open (no matter what has focus)
2. Perform the same tests using the `<Page />` component when it has more actions    

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Menu/Menu.test.tsx
+++ b/packages/components/src/Menu/Menu.test.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import renderer from "react-test-renderer";
 import { cleanup, fireEvent, render, waitFor } from "@testing-library/react";
 import { Menu } from ".";
 import { Button } from "../Button";
@@ -7,9 +6,9 @@ import { Button } from "../Button";
 afterEach(cleanup);
 jest.mock("uuid");
 
-it("renders a Menu", () => {
-  const tree = renderer
-    .create(
+describe("Menu", () => {
+  it("renders", () => {
+    const { container } = render(
       <Menu
         items={[
           {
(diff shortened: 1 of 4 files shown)


chris-at-jobber (commented):
Keyboard and mouse interactions work as expected 👌

packages/components/src/Menu/Menu.test.tsx
```diff
@@ -33,74 +32,126 @@ it("renders a Menu", () => {
...
+        />,
+      );
 
-  fireEvent.click(getByRole("menuitem"));
-  expect(clickHandler).toHaveBeenCalledTimes(1);
+      fireEvent.click(getByRole("button"));
+      fireEvent.keyDown(container, {
+        key: "Escape",
+        code: "Escape",
+        ctrlKey: true,
```
darryltec, line 107: interesting. y u testing with the `ctrlKey` pressed?

packages/components/src/Menu/Menu.tsx
```diff
@@ -178,6 +181,20 @@ export function Menu({ activator, items }: MenuProps) {
...
+    const { key } = event;
+    if (!visible) return;
+
+    event.preventDefault();
+    event.stopPropagation();
+    switch (key) {
+      case "Escape": {
+        hide();
+        break;
+      }
```
darryltec, line 195: interesting choice. y not

``` js
key === "Escape" && hide();
```

or something easier to scan 🤔

chris-at-jobber:
> Looks good to me. I'll let you decide what to do with my 2 comments.
> 
> Also, @chris-at-jobber we cool with this happening? If someone does put a menu in the modal, pressing escape when the menu is open closes both.
> 
> Oct-05-2021 10-50-59
> 
> _Also, this is another one of those, I have no idea how to fix it, just making us aware that it be a thing_

Hmmmm

I certainly don't love it, do you think we could figure it out in a separate PR? To my knowledge we aren't doing this anywhere yet with Atlantis components, but we will need to solve for it at some point.

packages/components/src/Menu/Menu.test.tsx
joelwarrington-jobber: I had copied the fire event from another test and didn't notice the ctrl key 🤭 

removed it in 99e07af as it's not needed for the test

joelwarrington-jobber:
@chris-at-jobber & @darryltec I'd suggest a follow-up PR/issue for the modal + menu interaction, we'd be able to check what has focus, and stop propagation.

joelwarrington-jobber:
@darryltec could I get a re-review/approval for the last two commits after tests?

packages/components/src/Page/Page.test.tsx
```diff
@@ -6,10 +6,12 @@ import { SectionProps } from "../Menu";
 
 jest.mock("@jobber/hooks", () => {
   return {
+    ...(jest.requireActual("@jobber/hooks") as {}),
```
darryltec, line 9: Will it be better to mock what you need rather than all of it?

joelwarrington-jobber: @darryltec, when mocking the entire `@jobber/hooks` module, if you don't require actual before what we actually want to mock, other hooks are no longer defined.

In this case, we're wanting to test `useOnKeydown` hook, and don't actually need to mock it, since it's quite easy to test, whereas `useResizeObserver` is hard to test and should be mocked.

In this case, we pull in everything that's defined, and override what should be mocked below.

darryltec (approved):
Further discussion
- Adding a menu on a modal and pressing esc when the menu is open would close both component

PR #705, opened 2021-10-01
feat(components): Support dismissible type on `<Chips>`

darryltec:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


chris-at-jobber:
Some notes on testing with VO:

- reads out the characters in Avatar, not needed when the user's name is present (ie "DT Darryl Tec")
- wonder if perhaps using a [datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) is more appropriate than a listbox in terms of properly communicating to the user what's going on
- unclear after I add/remove person from the list of selected users that they have been added/removed successfully

chris-at-jobber:
Bit of a quirk in the post-selection experience when...

- you're scrolled down the list so the first result is no longer in view
- you make a selection

Selection resets to the first result, _but_ the scroll does not reset so it's unclear that you actually have the first user selected. Upon any keypress scroll resets.
 
https://user-images.githubusercontent.com/39704901/150209409-83f94c39-a32d-4126-a86d-50194414162f.mov

packages/components/src/Chips/Chips.mdx
```diff
@@ -17,6 +18,14 @@ import { Text } from "@jobber/components/Text";
 
 # Chips
 
+<!-- DELETE ME: For testing purposes, delete before putting up for review -->
+
+### Add/dismiss selection playgound
```
rebecca-li, line 23: I can see a case where we may have a button to add another chip to the group. For example, could be used for adding tags to a page. 
Screen Shot 2022-01-19 at 5 22 06 PM


I can also think of cases (e.g. compose message, send email/sms modal) where we don't have a button, but instead just show the input area where they would be typing in chips to add. Here is an example from Jobber Online:

Screen Shot 2022-01-19 at 5 20 23 PM

rebecca-li: With the send email/sms modal, adding a comma in between strings is also used to create a new chip. Makes sense as functionality if the input allows for any value

rebecca-li: Screen Shot 2022-01-19 at 5 29 24 PM

If you enter a term that has matches, wondering if it makes more sense to have the "Add da" option listed first instead of last. If I wanted to input a custom value, typing in the value and hitting enter would select the first match on the list rather than the custom value I entered.

PR #706, opened 2021-10-01
feat(components): Autofocus on first Menu item

### LINKED ISSUE #697: Menu: First item should receive focus when opened
The first menu item does not receive focus when opened (user must tab to the first menu item).

Bhavin789:
## Motivations

Closes #697 

This change is better for user experience and accessibility as users can start selecting menu items using tabs. Otherwise, the tab indexing was traveled along the whole page which is a very long process 

### Added

- Added auto-focus of the first item of Menu 

### Changed

- Focuses the first item when menu is opened

## Testing

1. Run the local dev environment and search for the Menu component
2. Click on the `More Actions` in the examples.
3. Verify that the first item of the Menu is focused.
4. Also verify that pressing tab should move to the next item.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Menu/Menu.test.tsx
+++ b/packages/components/src/Menu/Menu.test.tsx
@@ -1,6 +1,12 @@
 import React from "react";
 import renderer from "react-test-renderer";
-import { cleanup, fireEvent, render, waitFor } from "@testing-library/react";
+import {
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
 import { Menu } from ".";
 import { Button } from "../Button";
 
@@ -95,4 +101,31 @@ test("it should passthrough an activator's click action", () => {
 
   fireEvent.click(getByRole("button"));
   expect(clickHandler).toHaveBeenCalledTimes(1);
+});
+
+test("it should focus first action item from the menu when clicked", () => {
+  const { getByRole } = render(
+    <Menu
+      activator={<Button label="Menu" />}
+      items={[
+        {
+          header: "Send as...",
+          actions: [
+            {
+              label: "Text Message",
+              icon: "sms",
+            },
+            {
+              label: "Email",
+              icon: "email",
+            },
+          ],
+        },
+      ]}
+    />,
+  );
+
+  fireEvent.click(getByRole("button"));
+  const firstMenuItem = screen.getAllByRole("menuitem")[0];
+  expect(firstMenuItem).toHaveFocus();
 });
--- a/packages/components/src/Menu/Menu.test.tsx
+++ b/packages/components/src/Menu/Menu.test.tsx
@@ -103,7 +103,7 @@ test("it should passthrough an activator's click action", () => {
   expect(clickHandler).toHaveBeenCalledTimes(1);
 });
 
-test("it should focus first action item from the menu when clicked", () => {
+it("should focus first action item from the menu when activated", () => {
   const { getByRole } = render(
     <Menu
       activator={<Button label="Menu" />}
--- a/packages/components/src/Menu/Menu.tsx
+++ b/packages/components/src/Menu/Menu.tsx
@@ -151,10 +151,10 @@ export function Menu({ activator, items }: MenuProps) {
                 <div key={key} className={styles.section}>
                   {item.header && <SectionHeader text={item.header} />}
 
-                  {item.actions.map(action => (
+                  {item.actions.map((action, index) => (
                     <Action
                       key={action.label}
-                      shouldFocus={key === 0}
+                      shouldFocus={key === 0 && index === 0}
                       {...action}
                     />
                   ))}
@@ -169,8 +169,6 @@ export function Menu({ activator, items }: MenuProps) {
 
   function toggle(callbackPassthrough?: (event?: MouseEvent) => void) {
     return (event: MouseEvent) => {
-      console.log("Hello");
-
       setVisible(!visible);
       callbackPassthrough && callbackPassthrough(event);
     };
--- a/packages/components/src/Menu/Menu.tsx
+++ b/packages/components/src/Menu/Menu.tsx
@@ -2,7 +2,9 @@ import React, {
   MouseEvent,
   ReactElement,
   createRef,
+  useEffect,
   useLayoutEffect,
+  useRef,
   useState,
 } from "react";
 import uuid from "uuid";
@@ -150,7 +152,11 @@ export function Menu({ activator, items }: MenuProps) {
                   {item.header && <SectionHeader text={item.header} />}
 
                   {item.actions.map(action => (
-                    <Action key={action.label} {...action} />
+                    <Action
+                      key={action.label}
+                      shouldFocus={key === 0}
+                      {...action}
+                    />
                   ))}
                 </div>
               ))}
@@ -163,6 +169,8 @@ export function Menu({ activator, items }: MenuProps) {
 
   function toggle(callbackPassthrough?: (event?: MouseEvent) => void) {
     return (event: MouseEvent) => {
+      console.log("Hello");
+
       setVisible(!visible);
       callbackPassthrough && callbackPassthrough(event);
     };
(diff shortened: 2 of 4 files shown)


packages/components/src/Menu/Menu.tsx
```diff
@@ -208,15 +214,30 @@ export interface ActionProps {
...
+
+  /**
+   * Focuses explicitly if set to true
+   */
+  shouldFocus?: boolean;
 }
 
-function Action({ label, icon, onClick }: ActionProps) {
+function Action({ label, icon, onClick, shouldFocus = false }: ActionProps) {
+  // eslint-disable-next-line no-null/no-null
```
Bhavin789, line 225: This is added because a `ref` is not allowed `undefined`. This, we need to specify it some value and the best I could think of was a `null`.

Please do suggest if something else can be used here. Thanks!

joelwarrington-jobber:
@Bhavin789 I've updated the description of the PR so that it is automatically linked to the issue 😄

Bhavin789:
> @Bhavin789 I've updated the description of the PR so that it is automatically linked to the issue 😄

Absolutely, thanks @joelwarrington-jobber 😃!

packages/components/src/Menu/Menu.tsx
```diff
@@ -208,15 +214,30 @@ export interface ActionProps {
...
+  const actionButtonRef = useRef<HTMLButtonElement | null>(null);
```
joelwarrington-jobber, line 226: Let's be consistent with how we get around the ESLint rule for `null` using type casting:
```suggestion
  const actionButtonRef = useRef() as RefObject<HTMLButtonElement>;
```

packages/components/src/Menu/Menu.tsx
```diff
@@ -208,15 +214,30 @@ export interface ActionProps {
    * Callback when an action gets clicked
    */
   onClick?(event: React.MouseEvent<HTMLButtonElement>): void;
+
+  /**
+   * Focuses explicitly if set to true
```
joelwarrington-jobber, line 219:
```suggestion
   * Focus on the action when rendered
```

packages/components/src/Menu/Menu.test.tsx
```diff
@@ -96,3 +102,30 @@ test("it should passthrough an activator's click action", () => {
   fireEvent.click(getByRole("button"));
   expect(clickHandler).toHaveBeenCalledTimes(1);
 });
+
+test("it should focus first action item from the menu when clicked", () => {
```
joelwarrington-jobber, line 106:
```suggestion
it("should focus first action item from the menu when activated", () => {
```

packages/components/src/Menu/Menu.tsx
```diff
@@ -208,15 +214,30 @@ export interface ActionProps {
...
-function Action({ label, icon, onClick }: ActionProps) {
+function Action({ label, icon, onClick, shouldFocus = false }: ActionProps) {
+  // eslint-disable-next-line no-null/no-null
+  const actionButtonRef = useRef<HTMLButtonElement | null>(null);
+
+  useEffect(() => {
+    if (actionButtonRef.current && shouldFocus) {
+      actionButtonRef.current.focus();
+    }
+  }, [shouldFocus]);
```
joelwarrington-jobber, line 232: @chris-at-jobber I hope you might know better, but I noticed in the mui menu component that the first item is not focused when using the mouse, but focused when using keyboard. In this implementation, it's focusing on both onClick and when using keyboard. Should that be the case?

joelwarrington-jobber (changes requested):
A few small changes in language, and also a confirmation on implementation with mouse onClick vs keyboard

packages/components/src/Menu/Menu.tsx
joelwarrington-jobber: This is how we do it right now in other components, so should be good to match. This might not be the best approach long term but we shouldn't tackle that here 😃

joelwarrington-jobber:
One other change that should be addressed... we should exclude the `shouldFocus` from the menu props as we don't want people to use it and should only be used internally to the menu component. Let me know if you'd like help with that.

packages/components/src/Menu/Menu.tsx
Bhavin789: Definitely! makes sense for the consistency. TS can sometimes really go against the wind and tell you NOT to use JS/React the way they are designed to be used 😄. Will update. Thanks 🙂!

packages/components/src/Menu/Menu.tsx
Bhavin789: Sure, will update this as well. Thanks!

packages/components/src/Menu/Menu.tsx
Bhavin789: I am actually working on a11y this close for the first time. Just going by the video. Somehow, I am not able to replicate the interaction with just tab 😕.
Is it possible for you to tell me how you interacted? I am using tab after a click but it is closing the menu for me. Thanks!

Bhavin789:
> One other change that should be addressed... we should exclude the `shouldFocus` from the menu props as we don't want people to use it and should only be used internally to the menu component. Let me know if you'd like help with that.

@joelwarrington-jobber As far as I could understand, its used in the `Action` component and thus, the `Menu` Props remain unaffected. Please let me know if I am missing something here. Thanks
 
image

packages/components/src/Menu/Menu.tsx
Bhavin789: > @chris-at-jobber I hope you might know better, but I noticed in the mui menu component that the first item is not focused when using the mouse, but focused when using keyboard. 

I am sorry, this wasn't meant for me but for @chris-at-jobber . My bad.

chris-at-jobber: @joelwarrington-jobber that's a great catch! The first item should only receive focus on the keyboard interaction; the mouseclick should retain focus on the clicked trigger element. Here's a video showing both interactions:

https://user-images.githubusercontent.com/39704901/135682590-8e1e2886-c7bf-4906-9b2d-213024e9a03b.mov
 
Note that on the first interaction (keyboard) we have the desired interaction. I move focus to the menu trigger, action the menu "trigger" button with `space`/`return`, focus moves to the first menu item, and I can hit `return` or `space` to action the menu item.

On the second interaction (mouse click) I'm able to hit `return` to action the first menu option, but I should not be able to. See this [demo from w3 org](https://www.w3.org/WAI/tutorials/menus/application-menus-code/) on mouse-click events for reference.

Tested in Menu's usage within the [Page component.](https://deploy-preview-706--jobber-atlantis.netlify.app/components/page#page-with-actions)

joelwarrington-jobber:
@Bhavin789 as you can see from the screenshot below, I'm able to define `shouldFocus` when this should be an internal detail:

image

Ideal case is the following:

image

You should be able to edit the `SectionProps`, so that the actions omits that type:

```ts
export interface SectionProps {
  /**
   * Defines the section header to further explain the group of actions.
   */
  header?: string;

  /**
   * List of actions.
   */
  actions: Omit<ActionProps, "shouldFocus">[];
  // ^ this was previously: actions: ActionProps[];
  // which made `shouldFocus` available
}
```

Bhavin789:
> You should be able to edit the SectionProps, so that the actions omits that type:

Oh yes, we can omit this prop from the type itself. Thanks for the suggestion @joelwarrington-jobber!

packages/components/src/Menu/Menu.tsx
Bhavin789: Thanks for the details here @chris-at-jobber. I am still trying to understand the mouse interaction. From the `w3` org, what I could see is I am able to hit return and see `100%` when I 
1. Click on `View` 
2. Press `return` 

image

And similar is what I can see from the current implementation as well. 

In the `w3`, if you:
1. Click on `View` 
2. Move the cursor away, it comes with a focussed white background and I can click return to alert.

image


In both cases - mouse and keyboard - hitting `return` is showing the alert with focus. Any idea on what I might be missing? Apologies for too many follow-ups.

chris-at-jobber: You know what, you're right! The visual presentation of their demo made it less apparent that the focus was in fact set, but the same behaviour is there. No changes required on this topic 👍

packages/components/src/Menu/Menu.tsx
```diff
@@ -208,15 +215,30 @@ export interface ActionProps {
...
+
+  /**
+   * Focus on the action when rendered
+   */
+  shouldFocus?: boolean;
 }
 
-function Action({ label, icon, onClick }: ActionProps) {
+function Action({ label, icon, onClick, shouldFocus = false }: ActionProps) {
+  // eslint-disable-next-line no-null/no-null
```
joelwarrington-jobber, line 226: Can you please remove this linking comment?
```suggestion
```

joelwarrington-jobber (changes requested):

packages/components/src/Menu/Menu.tsx
Bhavin789: Oh, I forgot. Let me quickly do it 👍. Thanks

joelwarrington-jobber (approved):
LGTM 👍 

First element in Menu will receive focus when activated with either click on keyboard.

https://user-images.githubusercontent.com/80279872/135871036-960c5a44-75f7-4763-ab36-7af590a8c8c4.mov

PR #710, opened 2021-10-02
feat(components): Add onClick to Avatar and refactor tests

joelwarrington-jobber:
## Motivations

It would be nice to have the capability to have avatar as an activator for the Menu component, similarly to how it's used within GitHub.
image

### Added

- Added onClick to the avatar component

### Changed

- Refactored the Avatar testing

## Testing

- Go to tooltip documentation and browse to the `usage with onclick` example, and click on it!
- Also supported keyboard shortcut enter, as it's a button
- Also can add onClick to any type of avatar (image or initial)

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Avatar/Avatar.css
+++ b/packages/components/src/Avatar/Avatar.css
@@ -5,18 +5,20 @@
 
 .avatar {
   display: flex;
-  align-items: center;
-  justify-content: center;
   width: var(--avatar-size);
   height: var(--avatar-size);
   min-width: var(--avatar-size);
   min-height: var(--avatar-size);
+  box-shadow: var(--shadow-base);
+  border-color: transparent;
   border-radius: var(--radius-circle);
   color: var(--color-heading);
   font-size: var(--avatar-size);
   -webkit-font-smoothing: antialiased;
   background: no-repeat center center var(--color-greyBlue);
   background-size: cover;
+  align-items: center;
+  justify-content: center;
 }
 
 .avatar:focus {
@@ -52,4 +54,13 @@
   box-shadow: 0 0 0 var(--border-thick) var(--color-white) inset;
   border-style: solid;
   border-width: var(--avatar-border-size);
+}
+
+.clickable {
+  cursor: pointer;
+}
+
+.clickable:focus {
+  box-shadow: var(--shadow-focus);
+  outline: none;
 }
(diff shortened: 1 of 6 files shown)


eddysims:
@chris-at-jobber what is your opinion on this? To me this feels weird to have an `onClick` on `Avatar`. I would say in the case mentioned in the description (activator for `Menu`) we would want to create a separate component for that use case.

darryltec:
I agree with @eddysims on this one. It's weird for that avatar to have an onclick. Most use cases that i've seen where an avatar is clickable is it's always wrapped in another component/element that accepts the click. Meaning, it's always separate from the _simple_ use case of an avatar.

Can you think of another use case aside from the menu activator. Because this opens this avatar component up to 

> "If you can click it to go to another page, you should be able to pass in a route on it too. So we need a `to` prop."

And that bloats the responsibility of the avatar component. If it's just the menu activator, I think it would be best to determine if the activator is a button component or something else entirely. If it's something else that doesn't accept an onClick then the menu would have to work around that internally. Or limit it to only a button since the menu passes in custom aria's that only **our** button component accepts.

Before we continue with all of that goodness, let's ask the question first of do we need this avatar activator in Jobber? It is nice. It's also might be another thing we won't need in the near future.

chris-at-jobber:
Agreed with the above! I think even looking at the Github examples it appears to be wrapped in other components, while Avatar itself is [more limited](https://primer.style/react/Avatar#avatar) in its' responsibilities.

joelwarrington-jobber:
In the process of getting some design feedback around the usage of something like this with @agglasaur and will update here.

If we're looking to go forward with something like this, a component wrapping avatar could make sense. 👍

joelwarrington-jobber:
Still want to go forward with adding onClick to the avatar so that it can be used as a menu activator, going to continue using this PR and revert changed, as it keeps discussion in one place.

darryltec:
@joelwarrington-jobber let's talk about this tomorrow. I'd like to understand the reasoning behind moving forward with adding an onClick on the avatar. If it's still to be an activator for the menu, there's a much better way to handle it that supercharges the menu component.

joelwarrington-jobber:
> @joelwarrington-jobber let's talk about this tomorrow. I'd like to understand the reasoning behind moving forward with adding an onClick on the avatar. If it's still to be an activator for the menu, there's a much better way to handle it that supercharges the menu component.

@darryltec it is to be a menu activator. Can you elaborate on how it'd be better to apply changes to the menu component?

PR #714, opened 2021-10-04
feat(components): Add description to base FormField component

joelwarrington-jobber:
## Motivations

FormField's do not have a way to display longer information/description outside of an error.

In this PR I've added a description, which acts similarly to the description found in checkbox
 
checkbox example

input text example

### Added

- Added description to FormField component (and inputs that inherit)

### Changed

- Refactored/reorganized FormField tests using `Describe` so it's easier to follow

## Testing

- Add `description` to any input which is inheriting FormField props, such as `InputText`, `InputPassword`, `Select`
- Use with `Multiline` inputs

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormFieldDescription.tsx
+++ b/packages/components/src/FormField/FormFieldDescription.tsx
@@ -1,0 +1,15 @@
+import React from "react";
+import { Text } from "../Text";
+
+interface FormFieldDescriptionProps {
+  description: string;
+}
+export function FormFieldDescription({
+  description,
+}: FormFieldDescriptionProps) {
+  return (
+    <Text variation="subdued" size="small">
+      {description}
+    </Text>
+  );
+}
--- a/packages/components/src/FormField/FormField.test.tsx
+++ b/packages/components/src/FormField/FormField.test.tsx
@@ -50,6 +50,15 @@ describe("FormField", () => {
     });
   });
 
+  describe("with a description", () => {
+    const label = "This is a hint!";
+
+    it("renders", () => {
+      const { getByText } = render(<FormField description={label} />);
+      expect(getByText(label)).toBeInTheDocument();
+    });
+  });
+
   describe("with a controlled value", () => {
     it("should set the value", () => {
       const value = "Look, some words!";
--- a/packages/components/src/FormField/FormField.test.tsx
+++ b/packages/components/src/FormField/FormField.test.tsx
@@ -4,178 +4,347 @@ import { cleanup, fireEvent, render, waitFor } from "@testing-library/react";
 import { FormField } from ".";
 
 afterEach(cleanup);
-it("renders correctly with no props", () => {
-  const tree = renderer.create(<FormField />).toJSON();
-  expect(tree).toMatchInlineSnapshot(`
-    <div
-      className="wrapper"
-      style={
-        Object {
-          "--formField-maxLength": undefined,
-        }
-      }
-    >
-      <div
-        className="inputWrapper"
-      >
-        <label
-          className="label"
-          htmlFor="123e4567-e89b-12d3-a456-426655440001"
-        />
-        <input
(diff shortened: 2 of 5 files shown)


chris-at-jobber:
@joelwarrington-jobber can we make sure that this description has a `described-by` relationship with the input? 

Also I was wondering if there's consideration for how this would work if an input also has validation enabled (does the error text replace the description, or stack beneath it?)

adambobadam:
Hey @chris-at-jobber This is what the description field and the validation looks like when they are both present.

image

adambobadam:
Hey @joelwarrington-jobber would splitting this PR up into two different ones make sense?

There seems to be a big change to the FormField tests and it might be good to see all the changes separately and make sure just the description is passing for now.  It would separate the concerns into two, and would be a bit easier to manage.

What do you think?

packages/components/src/FormField/FormFieldDescription.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { Text } from "../Text";
+
+interface FormFieldDescriptionProps {
+  description: string;
```
adambobadam, line 5: Should this be readonly?

adambobadam:
@chris-at-jobber Should there be a bit of space under the field?

space joel

joelwarrington-jobber:
> @joelwarrington-jobber can we make sure that this description has a `described-by` relationship with the input?
> 
> Also I was wondering if there's consideration for how this would work if an input also has validation enabled (does the error text replace the description, or stack beneath it?)



> @chris-at-jobber Should there be a bit of space under the field?
> 
> space joel

Going to differ design considerations to @agglasaur as it's needed for some work we're doing

packages/components/src/FormField/FormFieldDescription.tsx
joelwarrington-jobber: Sure, doesn't change much as it's only used internally, but to be consistent with our other type declarations, i've updated it in 43f730f

adambobadam: woot woot!

agglasaur:
@joelwarrington-jobber In most cases I think it makes sense for the description to appear with the error field stacked (looking at our usage of similar help text in Jobber Online) - it will be awkward with the password field as the help text is essentially the error as well - im not sure if its worth being able to "replace" it in that instance, or we...slightly reword it so its slightly less awkward?

Screen Shot 2021-10-05 at 9 09 34 AM

Otherwise, yes please add the spacing and the `described-by`

chris-at-jobber:
One visual note - can we either adjust the sizing of the error text to be `small`, or the description to use `base`? I would prefer `base` for both.

agglasaur:
@chris-at-jobber the checkbox and radio button description field use `small` by comparison - do you think those should be `base`?

chris-at-jobber:
Seeing an unexpected inconsistency in the spacing when InputText is used in isolation, vs in a Form:

### Alone:
image

### In a Form:
image

It looks like the culprit is the `Content` wrapper, which uses a `>` selector. When I remove it from the Form example, the description spacing is correct, but the Form collapses in on itself:
image

I know we have some component-specific reactions to `Content` in places like Card, Tabs, and Modal - is this another place where we need that, or do we need a wrapper div around the entire `InputText + description + error` container for this to work?

chris-at-jobber:
@joelwarrington-jobber design looks good, tested with/without description and with/without error validation present. Spacing looks as expected in all cases.

@darryltec could we get a pre-release up once `npm` is back online so this can be tested in-product? Joel's team is building out a feature using description at the moment and we'll want to put this through some larger QA as it's touching `FormField`.

darryltec:
Here you go! Happy testing.

```
npm install @jobber/components@2.53.3-pre.9+291bfa7
```

PR #719, opened 2021-10-05
feat(components): add ability to specify external url on card

### LINKED ISSUE #717: Card: Add ability to specify `external` similarly to button on card
Card has the ability to specify a URL but it doesn't give you the option to open it in a new tab if it's an external URL such as the button

shreyavi:
## Motivations
Fixes #717 

## Changes
Cards have the ability to have url links that open in a new tab.

## Testing

To see fixed version, pull down PR run the project and go to cards section in components and click on Clickable with External URL card

---

Random photo of Atlantis


joelwarrington-jobber (changes requested):
Looks good, however could you add a test case to check if external is set on a card, that the property is correctly set on the card anchor tag?

shreyavi:
@joelwarrington-jobber Added test and snapshot

packages/components/src/Card/tests/Card.test.tsx
```diff
@@ -109,3 +109,14 @@ test("it should should be clickable if it's clickable", () => {
...
+it("renders an external link card", () => {
+  const tree = renderer
+    .create(
+      <Card url="https://frend.space" external={true}>
+        <p>This is a link card.</p>
+      </Card>,
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
+});
```
joelwarrington-jobber, line 122: Could you test by getting the attribute specifically? You could use React testing library to get the anchor tag, and then getting the target attribute and confirming it's "_target"? 

Snapshots can work, but might be best to test specifically for the attribute.

joelwarrington-jobber (changes requested):

packages/components/src/Card/tests/Card.test.tsx
shreyavi: Done, let me know if it looks good

packages/components/src/Card/tests/Card.test.tsx
```diff
@@ -109,3 +109,23 @@ test("it should should be clickable if it's clickable", () => {
...
+
```
joelwarrington-jobber, line 123:
```suggestion
```

packages/components/src/Card/tests/Card.test.tsx
```diff
@@ -109,3 +109,23 @@ test("it should should be clickable if it's clickable", () => {
...
+it("renders an external link card with target attribute", () => {
+  render(
+    <Card url="https://frend.space" external={true}>
+      <p>This is a link card.</p>
+    </Card>,
+  );
+  expect(screen.getByRole("link")).toHaveAttribute("target", "_blank");
```
joelwarrington-jobber, line 130:
```suggestion
  const { getByRole } = render(
    <Card url="https://frend.space" external={true}>
      <p>This is a link card.</p>
    </Card>,
  );
  expect(getByRole("link")).toHaveAttribute("target", "_blank");
```

joelwarrington-jobber (changes requested):
Last few changes and should be good to go!

packages/components/src/Card/tests/Card.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import renderer from "react-test-renderer";
-import { fireEvent, render } from "@testing-library/react";
+import { fireEvent, render, screen } from "@testing-library/react";
```
joelwarrington-jobber, line 3:
```suggestion
import { fireEvent, render } from "@testing-library/react";
```

joelwarrington-jobber, line 3:
```suggestion
import { fireEvent, render } from "@testing-library/react";
```

packages/components/src/Card/Card.tsx
```diff
@@ -44,6 +46,9 @@ export function Card({
     accent && colors[accent],
   );
 
+  const tagProps = {
+    ...(external && { target: "_blank" }),
+  };
```
darryltec, line 51: You can actually write this on the element itself on line `78`. Saves you a const and also tagProps only applies to the `a` tag so that also makes it straightforward

```jsx
<a className={className} target={external ? "_blank" : "_self"} href={url}>

Comment (darryltec) [reply packages/components/src/Card/Card.tsx:51]: Which exposes this issue. 
image

Comment (joelwarrington-jobber) [reply packages/components/src/Card/Card.tsx:51]: @darryltec ahh, thanks for catching that! 

@shreyavi, in that case could you add the `rel` attribute as mentioned in the screenshot above if the target is `_blank`

Comment (shreyavi) [reply packages/components/src/Card/Card.tsx:51]: @darryltec in that case what do I put in else for rel here       
`<a className={className} target={external ? "_blank" : "_self"} rel={external ? "noopener noreferrer"} href={url}>
`
Any ideas?

Comment (shreyavi) [reply packages/components/src/Card/Card.tsx:51]: tag makes the most sense imo, https://www.w3schools.com/tags/att_a_rel.asp

File: packages/components/src/Card/tests/Card.test.tsx (hunk 2 above)
Comment (shreyavi) [reply packages/components/src/Card/tests/Card.test.tsx]: Resolved

File: packages/components/src/Card/tests/Card.test.tsx (hunk 3 above)
Comment (shreyavi) [reply packages/components/src/Card/tests/Card.test.tsx]: Resolved

File: packages/components/src/Card/Card.tsx [hunk 6]
```diff
@@ -70,7 +72,12 @@ export function Card({
     );
   } else if (url) {
     return (
-      <a className={className} href={url}>
+      <a
+        className={className}
+        target={external ? "_blank" : "_self"}
+        rel={external ? "noopener noreferrer" : "tag"}
+        href={url}
+      >
```
joelwarrington-jobber, line 80:
```suggestion
      <a
        className={className}
        href={url}
        {...(external && { target: "_blank", rel: "noopener noreferrer" })}
      >
```

Could I suggest this? You shouldn't need to worry about the anchor when it's not external, and this way we apply the `target` and `rel` at the same time.

joelwarrington-jobber (changes requested):

packages/components/src/Card/Card.tsx
shreyavi: Sure sounds good

shreyavi: @joelwarrington-jobber So we don't want to pass target=_self in case of external=false as suggested by @darryltec earlier

packages/components/src/Card/tests/Card.test.tsx
joelwarrington-jobber: @shreyavi screen wasn't removed, you shouldn't need to get screen from the package itself, and should instead destructure the render function

shreyavi: I am so sorry about that, just removed it

joelwarrington-jobber (approved):
LGTM

Thank you @shreyavi for your contribution 🥇  💯 

- [X] `rel` and `target` get set correctly
image

- [X] Keyboard and clicking on card opens in new tab as expected

https://user-images.githubusercontent.com/80279872/136084672-6bd60cc5-5508-4e90-9755-2367e4bfad42.mov

- [X] Default is internal and doesn't add `rel` and `target` attributes
image

Issue #690, opened 2021-09-27
RadioGroup is missing `role` and `aria-labelledby` attributes


eddysims:
The `RadioGroup` has a div that wraps all of the `RadioOptions` within. This div should have a role of `radiogroup` so that assistive technology can better identify it  

The `RadioGroup` should have an `aria-label`, `aria-labelledby` relationship, or a `label`, so that the group makes sense.

HeyKos:
I'd be happy to take this on this week if you want to assign me to it.

HeyKos:
@joelwarrington-jobber I am having trouble running the `npm ci` command locally. 
I am using macOS Big Sur version `11.3.1` 
I am running code `v14.18.0` and npm `7.24.1`.  I've attached the full debug log. The seemingly pertinent error is:

```
npm ERR! code 1
npm ERR! path /Users/mikekoser/Documents/Projects/atlantis/packages/components
npm ERR! command failed
npm ERR! command sh -c npm run clean; npm run build
npm ERR! > @jobber/components@2.51.0 clean
npm ERR! > rm -rf dist/* tsconfig.tsbuildinfo
npm ERR!
npm ERR!
npm ERR! > @jobber/components@2.51.0 build
npm ERR! > npm run build:cssTypes && npm run build:rollup
npm ERR!
npm ERR!
npm ERR! > @jobber/components@2.51.0 build:cssTypes
npm ERR! > tcm src
npm ERR!
npm ERR! Wrote /Users/mikekoser/Documents/Projects/atlantis/packages/components/src/Autocomplete/Autocomplete.css.d.ts
npm ERR! Wrote /Users/mikekoser/Documents/Projects/atlantis/packages/components/src/ButtonDismiss/ButtonDismiss.css.d.ts
npm ERR! Wrote /Users/mikekoser/Documents/Projects/atlantis/packages/components/src/Banner/notificationTypes.css.d.ts
npm ERR! Wrote /Users/mikekoser/Documents/Projects/atlantis/packages/components/src/Button/Button.css.d.ts
npm ERR! Wrote /Users/mikekoser/Documents/Projects/atlantis/packages/components/src/Card/colors.css.d.ts
npm ERR! Wrote /Users/mikekoser/Documents/Projects/atlantis/packages/components/src/Card/Card.css.d.ts
npm ERR! Wrote /Users/mikekoser/Documents/Projects/atlantis/packages/components/src/Content/Content.css.d.ts
[...]

PR #723, opened 2021-10-06
feat(components): RadioGroup - Added aria-labeledby attribute

HeyKos:
## Motivations

The RadioGroup has a div that wraps all of the RadioOptions within. This div has a role of radiogroup so that assistive technology can better identify it

The RadioGroup should have an aria-label, aria-labelledby relationship, or a label, so that the group makes sense.

Closes: #690 

### Changed

Added a new required prop to the `RadioGroup.tsx` component to represent an aria label for the radio group. The label is rendered as a `<span>`, but is visually hidden. The `<div>` with the `radiogroup` `role` references the id of the new element with the `aria-labeledby` attribute. This allows screen readers to better identify the purpose of a radio group.

## Testing

A new Jest test was added to ensure that the new label `<span>` is rendered. The snapshot was also updated to make sure that the snapshot test reflects the new markup.

### DIFF
--- a/packages/components/src/RadioGroup/RadioGroup.tsx
+++ b/packages/components/src/RadioGroup/RadioGroup.tsx
@@ -19,6 +19,11 @@ interface RadioGroupProps {
   onChange(newValue: string | number): void;
 
   /**
+   * Defines the label that describes the radio group.
+   */
+  readonly label: string;
+
+  /**
    * The name of the radio group, that links the radio options back up
    * to the group.
    *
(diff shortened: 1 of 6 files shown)


packages/components/src/RadioGroup/RadioGroup.tsx
```diff
@@ -30,22 +35,34 @@ interface RadioGroupProps {
...
-          {...option.props}
-        >
-          {option.props.children}
-        </InternalRadioOption>
-      ))}
-    </div>
+    <>
+      <span id={labelId} className={styles.groupLabel}>
+        {label}
+      </span>
```
joelwarrington-jobber, line 48: @chris-at-jobber i know you had created the issue for this PR, I'm curious if the fieldset and legend might be better suited for this use case as it appears more semantically correct?

HeyKos: @joelwarrington-jobber I've seen the same approach taken for other component libraries as well. I'd be down for swapping out the markup. Let me know what y'all decide.

PR #738, opened 2021-10-15
fix(components): don't show tooltip when it's not truncated [JOB-36864]

darryltec:
## Motivations

We add a tooltip when the chip starts truncating

image

But there's no easy way to know if a text is truncating or not so we determine it by the height of the content compared to the parent. If the height of the content is bigger then we can assume that it is truncated.

That logic fails on firefox and mobile safari as it adds a little bit more height to the content (.5 px to be exact) thus, triggering the tooltip to exist.

### How did you fix it

- Calculate the height of the content and the parent
- Since we know that the truncation happens by clamping the text to only show 1 line, if the content height is twice the size of the parent, then it is truncating
### Known limitations
- If you resize the screen and the chip starts truncating, the tooltip won't show as it is triggered via render rather than resize
  - Left it as is with the assumption that it'll rarely happen
  - Also happens vice versa

### Added

- 

### Changed

- the logic on determining if a tooltip shows up or not

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- On firefox, ensure the tooltip doesn't show on any example
- On mobile safari, ensure the tooltip doesn't show on any example

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Typography/css/Typography.css
+++ b/packages/components/src/Typography/css/Typography.css
@@ -32,4 +32,7 @@
   margin: 0;
   padding: 0;
   font-family: var(--typography--fontFamily-normal);
+  /* Wrap really long words or URL's */
+  overflow-wrap: break-word;
+  word-wrap: break-word;
 }


chris-at-jobber:
Tested on-device on Chrome (iPhone X)

✅  Does not show Tooltip on non-truncated Chip

❌  Does not show Tooltip on truncated Chip

image

darryltec:
@chris-at-jobber 

> Although I'm wondering if the issue is that I changed the label length after the page loaded, so the issue you mentioned with re-rendering is the cause?

Could be.

> Although, if you notice - the truncation is not happening so much as it's just overflowing the container (no ellipses).

yah, you added a really long word that technically doesn't wrap so it's breaking the trucation CSS too as that relies the text to go on multiple lines to truncate it. Which also breaks the tooltip logic haha. should be ez to deal with tho

darryltec:
@jackson-zhipeng-chang prerelease here if you want to try it in your branch 
```
npm install @jobber/components@2.58.1-pre.2+57571a1
```

I'll try it on my side too

darryltec:
tooltip on tips doesn't show on mobile

jackson-zhipeng-chang:
LGTM

PR #762, opened 2021-10-29
fix(components): Fix form field inconsistent spacing when used under Form and InputGroup JOB-37646

darryltec:
## Motivations

The `<Input* />` components aren't contained in a single tag. In some cases, those are okay but the input's also renders an error and now description below the field, thus opening those elements to be updated by anything that encapsulates it such as `<Form />` and `<InputGroup />`.

### Standalone input 

image

### In a `<Form />`

image

### In an `<InputGroup />`

image

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

#### Standalone input 

image

#### In a `<Form />`

image

#### In an `<InputGroup />`

image

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormFieldWrapper.tsx
+++ b/packages/components/src/FormField/FormFieldWrapper.tsx
@@ -1,4 +1,5 @@
 import React, {
+  Fragment,
   PropsWithChildren,
   RefObject,
   useEffect,
@@ -11,6 +12,7 @@ import styles from "./FormField.css";
 import { AffixIcon, AffixLabel } from "./FormFieldAffix";
 import { FormFieldDescription } from "./FormFieldDescription";
 import { InputValidation } from "../InputValidation";
+import { Content } from "../Content";
 
 interface FormFieldWrapperProps extends FormFieldProps {
   error: string;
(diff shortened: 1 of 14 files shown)


packages/components/src/FormField/FormFieldWrapper.tsx
```diff
@@ -100,14 +104,18 @@ export function FormFieldWrapper({
...
+      {!inline && (description || error) && (
+        <div>
+          {description && (
+            <FormFieldDescription
+              id={descriptionIdentifier}
+              description={description}
+            />
+          )}
+          {error && <InputValidation message={error} />}
+        </div>
```
robbyjobber, line 116: I like this so much better

robbyjobber:
@darryltec Do we want to address the spacing on this?  Seems like the error/description is getting cut off 

image
 
image

robbyjobber (changes requested):
Codewise LGTM. looks like some CSS changes needed tho 📓

darryltec:
> Do we want to address the spacing on this? Seems like the error/description is getting cut off

@robbyjobber on the inputgroup? No, that one is a lot harder

robbyjobber (approved):
> @robbyjobber on the inputgroup? No, that one is a lot harder
ah I see.  In that case LGTM then

PR #764, opened 2021-10-29
InputCurrency Proposal

TJKruitbosch:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputCurrency/InputCurrency.mdx
+++ b/packages/components/src/InputCurrency/InputCurrency.mdx
@@ -1,0 +1,156 @@
+---
+name: InputCurrency
+menu: Components
+route: /components/input-currency
+showDirectoryLink: true
+---
+
+import { Playground, Props } from "docz";
+import { useState } from "react";
+import { ComponentStatus } from "@jobber/docx";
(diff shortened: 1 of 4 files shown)


chris-at-jobber:
Initial questions/comments/thoughts!

Not sure if this a product of WIP, but is there a reason to only allow the input to align right and center?
image

I can understand totalling interfaces benefiting from a right-align but most cases I've seen this the value defaults to left for proximity to the currency prefix.

Would also suggest that a currency prefix should be included by default so we're rendering this consistently across the product.

This input is fairly opinionated about its' empty state. As soon as I `blur` the Input it resets to `0.00`.
image
image

Do you think that's something we need to enforce? If we pass a `defaultValue` that would have the same effect as a starting point right?

And finally, did you look at any libraries/browser APIs for localized currency formatting? I know there's a comment that that's not in the scope of this, but I believe there are some ways we could get this for relatively "cheap".

That's all for now!

chris-at-jobber:
Closing as this has gone a tad stale!

PR #766, opened 2021-11-01
docs(components): Put InputValidation below the Input where it belongs

chris-at-jobber:
## Motivations

Our pattern for validating Inputs is to render the error message below the inputs. See the [validation section for InputText](https://atlantis.getjobber.com/components/input-text/#validation-message).

We do this for consistency, but also to allow screen-reader users a better change to catch errors in the flow of navigating the form via the `tab` key.

image

However, most people looking for docs on InputValidation, go to [InputValidation](https://atlantis.getjobber.com/components/input-validation) where we show them this:

image

This is confusing for designers who then want to abstract the "error message goes above" pattern to other use cases:

> I'm wondering if there should be similar error text above a checkbox like we have for InputValidation

## Changes

Moves the InputValidation component in the sandbox to render below the Input, following our expected pattern.

### Fixed

Validation message shows beneath the Input
image

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputValidation/InputValidation.mdx
+++ b/packages/components/src/InputValidation/InputValidation.mdx
@@ -31,7 +31,6 @@ import { InputValidation } from "@jobber/components/InputValidation";
     const [validationMessages, setValidationMessages] = useState(undefined);
     return (
       <div>
-        {validationMessages && <InputValidation message={validationMessages} />}
         <Text>
           My name is
           <InputText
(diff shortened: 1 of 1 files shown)


joelwarrington-jobber (approved):
LGTM 👍 

It would be nice to follow-up with more examples of when to use InputValidation and when it might be used, but that's for another day!

adambobadam (approved):

chris-at-jobber:
[These reviews](https://c.tenor.com/7hJpFtl63HQAAAAM/sanic-the-hedgehob.gif) 

Thanks Joel + Adam 🙏

PR #767, opened 2021-11-01
feat(components): Add variation to FormField suffix

joelzwarrington:
## Motivations

Certain suffix actions could warrant a different type of action, and using variation can convey this information, such as with destructive, subtle or learning variations.

### Added

- `variation` prop added to FormField suffix when `onClick` present

## Testing

See `InputText` documentation in the `Prefix/Suffix` section 

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormFieldAffix.tsx
+++ b/packages/components/src/FormField/FormFieldAffix.tsx
@@ -8,12 +8,12 @@ import { Icon } from "../Icon";
 
 interface AffixLabelProps extends Affix {
   labelRef: RefObject<HTMLDivElement>;
-  variation?: "prefix" | "suffix";
+  affixVariation?: "prefix" | "suffix";
 }
 
 export function AffixLabel({
   label,
-  variation = "prefix",
+  affixVariation: variation = "prefix",
   labelRef,
 }: AffixLabelProps) {
   const affixLabelClass = classnames(styles.affixLabel, {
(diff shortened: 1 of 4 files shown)


joelwarrington-jobber:
@chris-at-jobber do you think exposing the variation is an acceptable pattern we want to expose?

image

chris-at-jobber:
@joelwarrington-jobber the initial plan for this was to fast-follow with a change to the suffixAction itself (not as a prop) that would use the subtle button (subtle button was in development at the same time, and the follow, as you can see, was not fast 😅 ) - I think if we did something like default to subtle and expose the other variations, as long as the button is always `tertiary`, that would be a decent end state. 

I can't say specifically that we do or do not ever want to use things like destructive, my only real concern would be adding additional decision points for a designer like "should I use a destructive variation when clearing a search vs a subtle variation". 

I'd also probably exclude `learning` from the variations as those use cases are rarely something that I think would get baked into an interface, they're typically found in prompts that can be dismissed.

chris-at-jobber:
@joelzwarrington I got some feedback from the rest of the design team and we've landed on "change the suffix to always be `subtle` when it has an `onClick`".

Did you want to update this PR to reflect that, or would you prefer to close this one and I can get up a PR to make that smaller change?

joelzwarrington:
This PR is superseded by #784

PR #771, opened 2021-11-03
proposal(components): DatePicker Proposal

eddysims:
## Motivations

We will be working on a `DatePicker` component in the near future. It is a good idea to do a quick proposal to make sure that the component will meet the needs of others before we build it.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,64 @@
...
+ * Disabling no-internal-modules here because we need
+ * to reach into the package to get the css file.
+ */
+// eslint-disable-next-line import/no-internal-modules
+import "react-datepicker/dist/react-datepicker.css";
+
+interface DatePickerReturnedDates {
+  readonly formatted: string;
+  readonly raw: Date;
+}
```
eddysims, line 17: @joelwarrington-jobber This is what i was thinking for the return value, this way we have a "readable" date and a raw date that we can do things like store in DBs

joelwarrington-jobber: I like it!

joelwarrington-jobber (commented):
overall, I'm really loving this approach!

I'm wondering if we want to adopt the InputText pattern from their documentation:


https://user-images.githubusercontent.com/80279872/140197840-84177aec-e2bb-4cad-9e9c-d7c29b560bcd.mov

packages/components/src/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+interface DatePickerInlineProps extends DatePickerBaseProps {
+  /**
+   * Display the Datepicker inline instead of in a modal style
+   */
+  readonly inline: boolean;
+}
+
+type DatePickerProps = XOR<DatePickerInlineProps, DatePickerModalProps>;
```
adambobadam, line 40: Is there an advantage to extending and creating a new inline interface with this XOR here vs just having `inline` be optional in the base props?  Just curious :)

packages/components/src/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [startDate, setStartDate] = useState(new Date());
+
+  return (
+    <ReactDatePicker
+      selected={startDate}
+      onChange={handleChange}
+      inline={inline}
+      customInput={
+        activator ? (
+          cloneElement(activator)
```
adambobadam, line 52: What is the reason for this line here?  🤔

packages/components/src/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Disabling no-internal-modules here because we need
+ * to reach into the package to get the css file.
+ */
+// eslint-disable-next-line import/no-internal-modules
+import "react-datepicker/dist/react-datepicker.css";
+
+interface DatePickerReturnedDates {
+  readonly formatted: string;
+  readonly raw: Date;
```
adambobadam, line 16: Is there a reason why `raw` isn't optional?  Just wondering!

packages/components/src/Datepicker/Datepicker.tsx
chris-at-jobber: In the proposal it says...

>  The reason is that we do not want the consumer to use inline and activator props at the same time.

adambobadam: Ah okay, cool!

packages/components/src/Datepicker/Datepicker.tsx
eddysims: For this one, this will be the returned value from the `onChange` prop. We would use it like this.

```tsx
function Component() {
 return <Datepicker onChage={handleChange}
 
 function handleChange(val: DatePickerReturnedDates) {
   const { formatted, raw } = val;
 }
}

**Comment by eddysims** [comment]
> overall, I'm really loving this approach!
> 
> I'm wondering if we want to adopt the InputText pattern from their documentation:
> 
>  Screen.Recording.2021-11-03.at.3.47.42.PM.mov

This will be a follow up. There will be 2 components, `DatePicker` and `InputDate` which will use the `DatePicker`

File: packages/components/src/Datepicker/Datepicker.tsx (hunk 4 above)
Comment (adambobadam) [reply packages/components/src/Datepicker/Datepicker.tsx:16]: Ah okay gotcha!  Was just wondering (more for my own knowledge) why it wasn't like:
```  readonly raw?: Date;
```

eddysims: It is becuase this is always returned fro, the `onChange`. You can see in the `handleChange` we do this.

```tsx
function handleChange(val: Date) {
    ...
    onChange && onChange({ raw: val, formatted: strFormatDate(val) });
}
```

Here we **always** return the `raw` and the `formatted` values. So when we set the interface as:

```tsx
interface DatePickerReturnedDates {
  readonly formatted: string;
  readonly raw: Date;
}
```

We are letting the consumer know that these will always be returned from the `onChange` function. If we were to do something like:


```tsx
interface DatePickerReturnedDates {
  readonly formatted: string;
  readonly raw?: Date;
}
```

We would be saying "You will always get the `formatted` date, but you might not always get the `raw` date" which would be untrue.

adambobadam: Okay, that makes sense!  Thanks for that :)

adambobadam (approved):
lgtm

chris-at-jobber:
Closing as this proposal will be implemented in https://github.com/GetJobber/atlantis/pull/774

PR #774, opened 2021-11-08
feat(components): DatePicker component

eddysims:
## Motivations

We are needing to add a `DatePicker` component to 🔱 Atlantis. This PR adds the `DatePicker`.

**Some Notes**

- This component was proposed in https://github.com/GetJobber/atlantis/pull/771
- To create the custom header component within the `DatePicker`, we needed to patch the `@types/react-datepicker` package and export the header props. There is a PR up to make this change in the real package. Once that is merged, we will no longer need this patch applied. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56999
- We needed a proper way to use the `Poppins` font within the Header of the `DatePicker`. To do this, we needed to add a new prop of `fontFamily` to the `Typography` component. 

### Added

- `DatePicker` component
- `fontFamily` prop to the `Typography` component.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Typography/css/FontFamilies.css
```diff
@@ -0,0 +1,7 @@
+.normal {
```
chris-at-jobber, line 1: In mobile we call this `base` iirc

chris-at-jobber:
Datepicker title should be title-case:
image

[Figma](https://www.figma.com/file/kiLocaMeG193MbyhvDiHWc/Design-System-Contribution-%5B-InputDate-%5D?node-id=332%3A5526) reference:
image

chris-at-jobber:
Day-of-week labels are rendering `#000`, should be `color-text`:
image

chris-at-jobber:
When inline, no shadow is fine, but should use `shadow-base` for the popover-style datepicker:
image

chris-at-jobber:
Going to push up the docs work shortly, this is looking really good!

chris-at-jobber:
Note on the activator:

- when I click the activator to close, it closes and then immediately re-opens the Datepicker rather than closing it
- can there be a prop exposed to configure the activator Button, or can we set it to not be a primary action at least? I don't think the datepicker should commonly be a primary action

chris-at-jobber:
One last thing I noticed... it appears that the "popover" version is still causing layout reflow, albeit subtle:

image

Popover version should not cause any layout reflow when opened.

packages/components/src/DatePicker/DatePicker.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * This gets the first result from the `changeHandler` mock
+   * now that it has been called
+   */
+  const [[changeHandlerResult]] = changeHandler.mock.calls;
+
+  expect(changeHandlerResult).toMatchObject({
+    raw: expect.any(Date),
+    formatted: "Apr 15, 2021",
+  });
```
adambobadam, line 52: Not to self - can this just be replaced with:

```
  expect(changeHandler).toHaveBeenCalledWith({
    raw: expect.any(Date),
    formatted: "Apr 15, 2021",
  });
```

adambobadam:
@chris-at-jobber
>    can there be a prop exposed to configure the activator Button, or can we set it to not be a primary action at least? I don't think the datepicker should commonly be a primary action

There's already an activator prop, by default there's a primary action, what would you like us to use instead?

> when I click the activator to close, it closes and then immediately re-opens the Datepicker rather than closing it

This is default behaviour by the datepicker component and would take some looking into.  Is this a blocking concern or can we fast follow?

chris-at-jobber:
> There's already an activator prop, by default there's a primary action, what would you like us to use instead?

Let’s default to a `<Button type="tertiary" />`

> This is default behaviour by the datepicker component and would take some looking into. Is this a blocking concern or can we fast follow?

We could split this into a secondary PR. It feels pretty broken so I wouldn't consider this component "done" until that's addressed. Either there should be no initial close animation, or clicking the activator should properly close the datepicker.

chris-at-jobber:
Did review w/ @adambobadam, follow-up tickets have been created for:

- refocus activator after `esc`
- datepicker blinks in and out of existence when clicking the activator

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export function DatePicker({ onChange, activator, inline }: DatePickerProps) {
+  const [startDate, setStartDate] = useState(new Date());
+
+  const datePickerClassNames = classnames(styles.datePicker, {
+    [styles.inline]: inline,
+  });
+
+  return (
+    <div>
```
darryltec, line 49: Might be best if this is inline-block

darryltec: also maybe a `<Fragment />` if `inline` is true.

```
import React, { Fragment, ReactElement, useState } from "react";

const Tag = inline ? "Fragment" : "div";

return <Tag>{...}</Tag>
```

Or look at how the element is being laid out in the browser and make em clearer and _intentional_, it seems inconsistent on what it would do. Right now it's
- Taking full width when there's an activator (even tho the button won't take the full width, the parent does)
- Shrink to it's smallest size when it's inline (same deal as above ^ but since we wrapped it in the div, the parent is taking full width)

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const selectedDateClass = ".react-datepicker__day--selected";
+
+    const selectedDate = document.querySelector(
+      selectedDateClass,
+    ) as HTMLDivElement;
+
+    if (selectedDate) {
+      selectedDate.focus();
+    }
+  }
```
darryltec, line 90: It seems to be focusing on the wrong calendar haha

https://user-images.githubusercontent.com/15986172/141370829-41243689-1fc7-40de-91f1-256e129598c6.mp4

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  function focusSelectedDate() {
+    const selectedDateClass = ".react-datepicker__day--selected";
+
+    const selectedDate = document.querySelector(
+      selectedDateClass,
+    ) as HTMLDivElement;
+
+    if (selectedDate) {
```
darryltec, line 87: Probably safer to check it this way instead of casting

```suggestion
    );

    if (selectedDate instanceof HTMLDivElement) {
```

darryltec (changes requested):
If you lay this out with other elements like so

```
<div>
  <DatePicker />
  <Button />
</div>
```

Should the button go beside the datepicker or below?

image

Asking because, by default, it's `inline-block` but also has a width of `100%` which would act like a `block` element

image

---

It seems that the `raw` time is frozen on when you loaded the page. Doesn't matter if you change the date, the time stays on when you loaded the page.

image

Is that intentional? Are the stakeholders okay with that? Do we know why they need the time on a datepicker?

adambobadam:
@darryltec

> Did you guys discuss around changing this to look if the button exists instead of rendering a snapshot?

Yeah we did at some point!  Eddy and I did there.  He argued that it should remain a snapshot (if my memory is correct here) because:

1. It is predictable and not going to be too much of an ugly snapshot
2. There isn't really a way around it since we need to make sure it's a button and we still need to be doing a check there that _something_ still exists in the DOM.

Although, I am not claiming that this is the best way, if you have another approach that'd be cool 😎

chris-at-jobber:
@darryltec re: the time, `react-date-picker` has an optional [time input](https://reactdatepicker.com/#example-input-time) that we're not exposing here. If we did then the user would be able to update that time input, but since we're not I _assume_ it's fine to have it silently captured.

adambobadam:
@darryltec 
> Might be best if this is inline-block

Should be addressed here: https://github.com/GetJobber/atlantis/pull/774/commits/26e5ef5b3547e4b0d4ee2f35e6707f104a8d3637

If you think this works!

darryltec:
> @darryltec re: the time, `react-date-picker` has an optional [time input](https://reactdatepicker.com/#example-input-time) that we're not exposing here. If we did then the user would be able to update that time input, but since we're not I _assume_ it's fine to have it silently captured.

That explains why the time doesn't update regardless of what you do. This component is probably expecting a time picker when you return a time on the datas. 

Still odd that this component returns a time by default. Seems like something that someone would have to strip out all the time everytime they use this component.

packages/components/src/DatePicker/DatePicker.css
```diff
@@ -0,0 +1,97 @@
+.datePickerParentWrapper {
```
darryltec, line 1: Just a note: Don't need to act upon it. 

You don't have to make this as specific as possible as it won't clash with other classnames since those gets automagically adjusted to not class by css modules. Can just say wrapper.

adambobadam:
> > @darryltec re: the time, `react-date-picker` has an optional [time input](https://reactdatepicker.com/#example-input-time) that we're not exposing here. If we did then the user would be able to update that time input, but since we're not I _assume_ it's fine to have it silently captured.
> 
> That explains why the time doesn't update regardless of what you do. The component author of react datepicker is probably expecting a time picker to be used when you return a time on the datas.
> 
> Still odd that this component returns a time by default. Seems like something that someone would have to strip out all the time everytime they use this component.

@darryltec Do you think we should strip out time unless it's specified?  Or do you think just returning the raw format that the component itself is providing is fine?

darryltec (dismissed):
I get it now, that `raw` format is actually a `Date` object. 

image

It's probably best to document that as well in the `mdx` and `interface`.

Could be a fast follow tho.

_PS. Fast follow means it's a sub task of the original ticket and should be done soon. Hence fast. In case there are confusion there._ 😉

adambobadam:
@joelzwarrington @chris-at-jobber After discussing a bit with @darryltec we both thought it would be a good idea to remove the `raw` and `formatted` return and just have the component use a `raw` Date object.  This way it follows the React convention where the `onChange` returns what the `selected` prop expects.  It keeps things tidy and state is determined by the parent.  

Would be good to hear your thoughts/possible objections!

chris-at-jobber:
No concerns on my part

darryltec (approved):
This looks good to me. Hold off on merging tho until the parties involved are cool with it as well

robbyjobber:
This looks good to me.  The changes to utilizing a raw date object will work for what 🐝 's will need. 👍

PR #776, opened 2021-11-15
fix(components): use correct colors for disabled text

chris-at-jobber:
## Motivations

Atlantis FormFields were initially built with some legacy styling that has since been updated. This introduces usage of the intended colors for disabled states in FormField and its' inheriting components.

### Before:

- Uses an opacity of 0.6 on the entire input
- Value does not change color, but "dims" thanks to the use of opacity

image

### After:

- Uses our semantic color of `disabled` for the value when the input is disabled
- No longer using opacity value

image

### Changed

- CSS changes to remove opacity and introduce usage of `color-disabled` for the FormField's `value` text elements.

### Fixed

- FormField with a value no longer uses a combination of opacity-changes and `blue` to create a "disabled" appearance, instead uses our specified semantic colors.

## Testing

Add the following snippets into the sandboxes for each component. There are more components that use `FormField`, but this should be a representative sample:

### InputText

```
<Content>
<InputText disabled placeholder="Ship Name" />
<InputText disabled defaultValue="Rocinante" placeholder="Ship Name" />
<InputText size="small" disabled placeholder="Ship Name" />
<InputText size="small" disabled defaultValue="Rocinante" placeholder="Ship Name" />
<InputText size="large" disabled placeholder="Ship Name" />
<InputText size="large" disabled defaultValue="Rocinante" placeholder="Ship Name" />
<InputText prefix={{ icon:'search', label:'Search for' }} disabled placeholder="Ship Name" />
<InputText prefix={{ icon:'search', label:'Search for' }}  disabled defaultValue="Rocinante" placeholder="Ship Name" />
<InputText prefix={{ icon:'search', label:'Search for' }}  size="small" disabled placeholder="Ship Name" />
<InputText prefix={{ icon:'search', label:'Search for' }}  size="small" disabled defaultValue="Rocinante" placeholder="Ship Name" />
<InputText prefix={{ icon:'search', label:'Search for' }}  size="large" disabled placeholder="Ship Name" />
<InputText prefix={{ icon:'search', label:'Search for' }}  size="large" disabled defaultValue="Rocinante" placeholder="Ship Name" />
</Content>
```

### InputPassword

```
<InputPassword disabled placeholder="Ship Name" />
<InputPassword disabled defaultValue="Rocinante" placeholder="Ship Name" />
<InputPassword size="small" disabled placeholder="Ship Name" />
<InputPassword size="small" disabled defaultValue="Rocinante" placeholder="Ship Name" />
[...]

PR #777, opened 2021-11-15
feat(components): Use focus shadow instead of focus border

chris-at-jobber:
## Motivations

Similar to #776 , Atlantis inputs were designed with legacy styling in mind that we should be OK to move on from.

In particular, the lack of a focus "ring" and leveraging border-color in lieu led to issues where you could not clearly tell if you had focused an input with an error; this makes it harder to fix the error as it adds a bit of cognitive load.

### Before:

Input in isolation
image

Input focused + invalid (note that you can't see the "focus" color because the border has taken the "error" color)
image

Input in a group (note that bottom border is also hidden by subsequent input)
image

Input in a modal
image

### After:

Input in isolation:
image

Input focused + invalid
image

Input in a group:
image

Input in a modal (showing that z-index usage is not creating any conflicts with other "elevated" elements)
image

### Added

- Focus shadow styling to FormField

### Changed

- Focus state now uses "shadow" focus style, and (although not a visual change) relies on z-index to ensure focus ring is visible when Inputs are in an InputGroup

## Testing

- [x] Focus [InputText](https://deploy-preview-777--jobber-atlantis.netlify.app/components/input-text)
- [x] Focus [InputGroup](https://deploy-preview-777--jobber-atlantis.netlify.app/components/input-group)
- [x] Focus [Select](https://deploy-preview-777--jobber-atlantis.netlify.app/components/select)
- [x] Focus the input in a [Modal in the "with Actions" demo of Modal](https://deploy-preview-777--jobber-atlantis.netlify.app/components/modal/#with-actions)

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -53,7 +53,9 @@
 }
 
 .wrapper:focus-within {
-  --field--border-color: var(--color-focus);
+  position: relative;
+  z-index: var(--field--base-elevation);
+  box-shadow: var(--shadow-focus);
 }
 
 .miniLabel {


packages/components/src/FormField/FormField.css
```diff
@@ -53,7 +53,9 @@
 }
 
 .wrapper:focus-within {
-  --field--border-color: var(--color-focus);
+  position: relative;
+  z-index: var(--field--base-elevation);
```
chris-at-jobber, line 57: Referencing the CSS approach used by [ANT Financial](https://ant.design/components/input/#components-input-demo-group) to handle the same "show focus state even when subsequent inputs should overlap" situation

agglasaur: Cool

agglasaur:
Screen Shot 2021-11-15 at 3 57 02 PM

Gets a bit wonky with browser autofill but its still better than before and I am comfortable approving without that being addressed.

agglasaur (approved):

PR #778, opened 2021-11-16
fix(components): Allowlist accepted image types [JOB-37183]

JordanLos:
## Motivations

When using iOS image uploads _using the file system and not the photos app,_ `images/*` will allow `.heic` file formats which are not supported by most browsers. Changing the accept parameter to explicitly declare only accepted file types prevents uploading unsupported image types.

## Testing Bug Fix in Jobber Online
Bug related to unviewable `hiec` images coming in through Work Request Image Uploads (a feature only available to Fiber New franchise)[Jira Ticket here](https://jobber.atlassian.net/browse/JOB-37183)

**Get a Heic file to upload on iOS dev**
1. Take a picture on an iOS device and save it to Files (it will be saved as `.heic`)
2. In `app/features/work_request_image_upload_feature.rb` set `available?` to return `true` (this enables the work request image upload feature)
3. Setup an ngrok tunnel so you can access your dev environment
4. Open a work request in client hub - you should see the ability to add an image (against in ngrok)
Note: to do this, I had to go to `letter_opener`, then get the url for client hub, and replace `localhost.test` in the url with the ngrok tunnel url.   

**Reproduce the Bug**
1. Attempt to add the picture **from the files app, not the photos app**. The Photos app will automatically convert `heic` to `jpeg` and won't reproduce the bug. You should see the `heic` file as an option to upload - however adding it to the work request image will give a blank thumbnail (because its a `heic` file)

**Test the fix**
1. run `npm install @jobber/components@2.65.2-pre.1+71b61a99` - which includes this pre-release. Once its done - reload the page (updating the component)
2.  Try to upload the `hiec` image from the Files app as before and you should find it disabled and not available for uploading. 

 Pull in pre-release using `npm install @jobber/components@2.65.2-pre.1+71b61a99`
2.

## Choice of Images to Allow
This PR moves from allowing all image types to only certain ones, and based on the discussion with the FreshOpps designer, tech lead, and product managers there was a consensus that the following images formats are sufficient:
- JPEG (`jpg`,`jpeg`)
- PNG (`png`)

Chose note to support
- BMP because it is an uncompressed image type
- GIFs are typically treated as movies by iOS and do not have full support in Jobber Online (i.e. using a gif for an image does not produce an animation but just a static image)

## Changes
Change the `accept` from `images/*` (all image types) to a whitelist of accepted images (which exclude `heic`)

### Added
No new features

### Changed
Images types accepted by `InputFile`

### Deprecated
N/A

### Removed
N/A

### Fixed
Prevent unsupported `heic` images from getting uploaded.

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -82,7 +82,7 @@ interface InputFileProps {
 
   /**
    * Allowed File types.
-   * 
+   *
    * @param "images" - only accepts all types of image
    * @param "basicImages" - only accepts png, jpg and jpeg
    *
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -82,6 +82,10 @@ interface InputFileProps {
 
   /**
    * Allowed File types.
+   * 
+   * @param "images" - only accepts all types of image
+   * @param "basicImages" - only accepts png, jpg and jpeg
+   *
    * @default "all"
    */
   readonly allowedTypes?: "all" | "images" | "basicImages";
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -84,7 +84,7 @@ interface InputFileProps {
    * Allowed File types.
    * @default "all"
    */
-  readonly allowedTypes?: "all" | "images";
+  readonly allowedTypes?: "all" | "images" | "basicImages";
 
   /**
    * Allow for multiple files to be selected for upload.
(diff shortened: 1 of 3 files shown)


darryltec:
@JordanLos and I talked about this last week. Decided to add a new `allowedTypes` called `basicImages` since images is still legit

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -84,7 +84,7 @@ interface InputFileProps {
    * Allowed File types.
    * @default "all"
    */
```
darryltec, line 86: @JordanLos sorry, one last thing

```suggestion
   * Allowed File types.
   * 
   * @param "images" - only accepts all types of image
   * @param "basicImages" - only accepts png, jpg and jpeg
   *
   * @default "all"
   */
```

darryltec (approved):

PR #785, opened 2021-11-26
feat(components): InputPhoneNumber component

tyler-santos:
## Motivations
We wanted to introduce a formatted phone number input for our users to reduce the rate at which they enter bad data, and make it clear at a glance what we're asking for.

### Added
- Adds a new `InputPhoneNumber` component with a phone number mask applied to it
- Adds a new library for handling the input masking of numeric fields. We evaluated several option, this was the only one that had ongoing maintenance, a large amount of support and usage, and worked well with our `react-form-hooks`.

Does Not:
- Mask non numeric inputs. It's very easy to determine the cursor position when you limit the input values to just numeric values. There are not a lot of libraries that allow generic input masking with non-numeric characters.

### Changed
- Adds a `numberMask` type to the `FormField` component so that other inputs can leverage this masking behaviour

## Testing
This URL will take you to the input phone number component and you can play around with it there.

http://localhost:3333/components/input-phone-number

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -9,6 +9,7 @@ import React, {
 } from "react";
 import uuid from "uuid";
 import { Controller, useForm, useFormContext } from "react-hook-form";
+import NumberFormat from "react-number-format";
 import { FormFieldProps } from "./FormFieldTypes";
 import styles from "./FormField.css";
 import { FormFieldWrapper } from "./FormFieldWrapper";
@@ -40,6 +41,7 @@ export function FormField(props: FormFieldProps) {
     onFocus,
     onBlur,
     onValidation,
+    maskingProperties,
   } = props;
 
   const { control, errors, setValue, watch } =
@@ -134,6 +136,21 @@ export function FormField(props: FormFieldProps) {
                   ref={inputRef as MutableRefObject<HTMLTextAreaElement>}
                 />
               );
+            case "maskedNumber":
+              return (
+                <>
+                  <NumberFormat
+                    {...textFieldProps}
+                    {...maskingProperties}
+                    thousandSeparator={true}
+                    prefix={"$"}
+                    onValueChange={handleOnValueChange}
+                    getInputRef={inputRef}
+                    onChange={maskedOnChangeHandler}
+                  />
+                  {loading && <FormFieldPostFix variation="spinner" />}
+                </>
+              );
             default:
               return (
                 <>
@@ -165,6 +182,17 @@ export function FormField(props: FormFieldProps) {
           }
 
           onChange && onChange(newValue);
+          onControllerChange(event);
+        }
+        function handleOnValueChange(values: {
+          formattedValue: string;
+          originalValue: string;
+        }) {
+          const { formattedValue } = values;
+          onChange && onChange(formattedValue);
+        }
+
+        function maskedOnChangeHandler(event: ChangeEvent<HTMLInputElement>) {
           onControllerChange(event);
         }
 
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -3,13 +3,17 @@ import React, {
   FocusEvent,
   KeyboardEvent,
   MutableRefObject,
+  SyntheticEvent,
   useEffect,
   useImperativeHandle,
   useState,
 } from "react";
 import uuid from "uuid";
 import { Controller, useForm, useFormContext } from "react-hook-form";
-import NumberFormat from "react-number-format";
+import NumberFormat, {
+  NumberFormatProps,
+  NumberFormatValues,
+} from "react-number-format";
 import { FormFieldProps } from "./FormFieldTypes";
 import styles from "./FormField.css";
 import { FormFieldWrapper } from "./FormFieldWrapper";
@@ -139,12 +143,12 @@ export function FormField(props: FormFieldProps) {
             case "maskedNumber":
               return (
                 <>
-                  <NumberFormat
-                    {...textFieldProps}
-                    {...maskingProperties}
-                    onValueChange={handleOnValueChange}
-                    getInputRef={inputRef}
-                    onChange={maskedOnChangeHandler}
+                  <MaskedNumberInputInternal
+                    textFieldProps={textFieldProps}
+                    maskingProperties={maskingProperties}
+                    onChange={onChange}
+                    onControllerChange={onControllerChange}
+                    inputRef={inputRef}
                   />
                   {loading && <FormFieldPostFix variation="spinner" />}
                 </>
@@ -180,17 +184,6 @@ export function FormField(props: FormFieldProps) {
           }
 
           onChange && onChange(newValue);
-          onControllerChange(event);
-        }
-        function handleOnValueChange(values: {
-          formattedValue: string;
-          originalValue: string;
-        }) {
-          const { formattedValue } = values;
-          onChange && onChange(formattedValue);
-        }
-
-        function maskedOnChangeHandler(event: ChangeEvent<HTMLInputElement>) {
           onControllerChange(event);
         }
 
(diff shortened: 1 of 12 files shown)


packages/components/src/FormField/FormField.tsx
```diff
@@ -167,6 +184,17 @@ export function FormField(props: FormFieldProps) {
           onChange && onChange(newValue);
           onControllerChange(event);
         }
+        function handleOnValueChange(values: {
```
andrewjdmccann, line 187: Maybe we should called this `maskedOnValueChangedHandler` to make it clear it's related to the masking code.

packages/components/src/FormField/FormField.tsx
```diff
@@ -134,6 +136,21 @@ export function FormField(props: FormFieldProps) {
...
                 />
               );
+            case "maskedNumber":
+              return (
+                <>
+                  <NumberFormat
+                    {...textFieldProps}
+                    {...maskingProperties}
+                    thousandSeparator={true}
+                    prefix={"$"}
```
andrewjdmccann, line 146: Do we actually want this prefix?

packages/components/src/FormField/FormField.tsx
```diff
@@ -134,6 +136,21 @@ export function FormField(props: FormFieldProps) {
                   ref={inputRef as MutableRefObject<HTMLTextAreaElement>}
                 />
               );
+            case "maskedNumber":
+              return (
+                <>
+                  <NumberFormat
+                    {...textFieldProps}
+                    {...maskingProperties}
+                    thousandSeparator={true}
```
andrewjdmccann, line 145: Do we want this `thousand` separator?

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export function InputPhoneNumber({
+  country,
+  placeholder,
+  alwaysShowMask = true,
+  showCountryCode = false,
+  ...rest
+}: InputPhoneNumberProps) {
+  const [value, setValue] = React.useState("");
```
andrewjdmccann, line 29: The caller of the InputPhoneNumber should probably be the one who manages the state of `value`, and `onChange`. We don't need to define those in our component.

This input would be consumed like this:

```jsx
function MyEchoComponent() {
  const [value, setValue] = useState("");
  
  return (
    <>
      <InputPhoneNumber onChange={setValue} value={value} />
      <p>Echo: {value}</p>
    </>
  )
}
```

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const displayedCountryCode = showCountryCode
+    ? `+${countries[country].countryCode}`
+    : "";
+  const maskingProperties = {
+    allowEmptyFormatting: alwaysShowMask,
+    format: `${displayedCountryCode} ${countries[country].format}`,
+    mask: "_",
```
andrewjdmccann, line 36: Mockups didn't have a mask, just formatting. Do we actually want the input masked with `_`

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -0,0 +1,53 @@
...
+    prefix: `${displayedCountryCode}`,
+  };
+  return (
+    <FormField
+      value={value}
```
andrewjdmccann, line 41: `onChange` `value` and `placeholder` could all just be part of `rest`

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
tyler-santos: Done!

packages/components/src/FormField/FormField.tsx
tyler-santos: @andrewjdmccann I am leaving this one to your discretion :)

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
tyler-santos: Great questions, perhaps lets check with Rebecca or Chris on Monday.

kkimble006:
I tested on here: https://deploy-preview-785--jobber-atlantis.netlify.app/components/input-phone-number

kkimble006:
Hello @tyler-santos and @andrewjdmccann 

I tested on [here](https://deploy-preview-785--jobber-atlantis.netlify.app/components/input-phone-number).

Should this work for both North America and Great Britain (I don't see +44 or the mask for Great Britain, is this expected? it works if I start typing, but if I type a 4 I need to delete the 4)?
It also doesn't seem to actually format the GB number, other than separate out the country code, is this expectec?
Screen Shot 2021-11-29 at 2 22 47 PM
Screen Shot 2021-11-29 at 2 22 41 PM

I also noticed if I add phone number with country code to input where `alwaysShowMask` is false, it was kind of confusing. Not sure if this is something we want to do in the future - but if we could remove the country code in this specific case as the user is typing that would prevent this: (Especially for a copy/past situation)
Screen Shot 2021-11-29 at 2 24 20 PM

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export interface AllowedRegions {
+  NorthAmerica: RegionSettings;
+  GreatBritain: RegionSettings;
+  Unknown: RegionSettings;
+}
+
+const REGION_SETTINGS: AllowedRegions = {
+  NorthAmerica: { countryCode: 1, format: "(###) ###-####" },
+  GreatBritain: { countryCode: 44 },
```
kkimble006, line 17: I think this relates to my comment on the main description. Is there a reason we didn't add format for Great Britain?

andrewjdmccann: @kkimble006 From the reading that I was doing, we can't guarantee the number of digits in a GB phone number, so the formatting could change dynamically as the number is added. 

As a result I think formatting a GB number would be very, very hard. Instead I chose to just prefix it with `+44 `. We could try and do dynamic typing but I think that's an iteration on it's own and would get complicated.

packages/components/src/InputPhoneNumber/InputPhoneNumber.mdx
```diff
@@ -0,0 +1,73 @@
...
+---
+
+import { Playground, Props } from "docz";
+import { InputPhoneNumber } from ".";
+import AllowedCountries from "./InputPhoneNumber";
+import { Heading } from "../Heading";
+
+# Input Phone Number
+
+Input phone number is used in anywhere you need to accept phone numbers as
```
kkimble006, line 15: "is used anywhere" ?

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
andrewjdmccann: Here are all the formats:

GB formats

The varying number of digits inside the `(##)` and the digits in each gruoping afterward would make this super difficult.

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -0,0 +1,75 @@
...
+  Unknown: {},
+};
+
+// should it be form field or common form field?
```
kkimble006, line 21: Looks like `maskingProperties` is on FormField, so I think we need to use `FormFieldProps` here?

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
kkimble006: @andrewjdmccann should the "+44" still appear in the field? Like the "+1"?

andrewjdmccann: It should it's a bug if it isn't. Taking a look now.

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
andrewjdmccann: I don't think there is one. Will remove the comment, some inputs import from that, some import from FormField.

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
kkimble006: Let me know when that is working and I'll get more 👀 back on this PR!

andrewjdmccann: @kkimble006 Fixed. The +44 will now show up before typing a number.

andrewjdmccann:
> I also noticed if I add phone number with country code to input where alwaysShowMask is false, it was kind of confusing. Not sure if this is something we want to do in the future - but if we could remove the country code in this specific case as the user is typing that would prevent this: (Especially for a copy/past situation)

@kkimble006 I really like this idea, I'm looking to see if it's doable, but my gut says that we're at the mercy of the library maintainers here. There might be options to implement a custom formatter/unformatter, but I've been playing around with those... and they're very complicated to use and error prone.

kkimble006:
@andrewjdmccann LGTM! (I assumed Rebecca 👍 this already, or another designer?)

andrewjdmccann:
Closing this PR, it's something I think we should continue to explore, but the accessibility issues aren't something we can easily overcome.

We will not be continuing on with this component, but @kkimble006 thanks for the help in getting it reviewed.

PR #786, opened 2021-11-27
docs(components): add more detail to datepicker usage

chris-at-jobber:
## Motivations

Got a question of "when should I actually use a datepicker on its own, vs in an InputDate format?"

### Added

- Just a bit more of a tangible example use case for datepicker as a standalone

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.mdx
+++ b/packages/components/src/DatePicker/DatePicker.mdx
@@ -47,7 +47,7 @@ Use Datepicker when you are looking to provide the user with a highly visual
 interface for intuitive date selections.
 
 A classic use case for a standalone Datepicker would be as a means to navigate
-a more content-heavy calendar interface.
+within a larger calendar interface.
 
 It may be beneficial to supplement the Datepicker with alternate means of
 entering their date selection, such as providing a `date` input.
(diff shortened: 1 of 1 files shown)


packages/components/src/DatePicker/DatePicker.mdx
```diff
@@ -46,6 +46,9 @@ import { DatePicker } from "@jobber/components/DatePicker";
 Use Datepicker when you are looking to provide the user with a highly visual
 interface for intuitive date selections.
 
+A classic use case for a standalone Datepicker would be as a means to navigate
+a more content-heavy calendar interface.
```
agglasaur, line 50: "to navigate **to**"

agglasaur (changes requested):

packages/components/src/DatePicker/DatePicker.mdx
chris-at-jobber: I don't think so? I'm thinking of something like this

image

Or this
image

Where the Datepicker is a tool for navigating the calendar, not really navigating _to_ the calendar?

chris-at-jobber: Maybe navigating _within_ the calendar is more clear?

agglasaur: oh i see, yeah i definitely interpreted the original differently

agglasaur: `within` makes sense

chris-at-jobber:
```suggestion
within a larger calendar interface.
```

agglasaur (approved):

PR #787, opened 2021-11-29
feat(components): update toast visuals and timing [JOB-27572]

chris-at-jobber:
## Motivations

The Toast, while intended as a quick, non-critical message, is a bit too unobtrusive; admins are missing the success messages when they save changes, for example.

This is particularly problematic on larger/multiple monitor setups, which we know many of our admins are operating on.

### Changed

- Toast now uses the `reverse` colors from the theme for high contrast against the main application surface
- Toast now has longer min and max durations, and a longer auto-calculated duration

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Toast/Toast.tsx
+++ b/packages/components/src/Toast/Toast.tsx
@@ -80,7 +80,7 @@ export function Toast({
                 ariaLabel={"Hide Notification"}
                 onClick={handleToastClose}
                 type="tertiary"
-                variation="cancel"
+                variation="subtle"
               />
             </div>
           </div>
@@ -102,9 +102,9 @@ export function Toast({
   }
 
   function getTimeout() {
-    const min = 2000;
-    const max = 5000;
-    const time = message.length * 100;
+    const min = 5000;
+    const max = 10000;
+    const time = message.length * 200;
 
     if (time < min) return min;
     if (time > max) return max;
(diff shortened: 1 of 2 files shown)


packages/components/src/Toast/Toast.tsx
```diff
@@ -102,9 +102,9 @@ export function Toast({
   }
 
   function getTimeout() {
-    const min = 2000;
-    const max = 5000;
-    const time = message.length * 100;
+    const min = 5000;
+    const max = 10000;
+    const time = message.length * 200;
```
adambobadam, line 107: Perhaps too much of a smartboi comment (sorry!) figure I'll ask anyways.  Should time here also account for something like the variation type as well?

adambobadam (approved):

packages/components/src/Toast/Toast.tsx
chris-at-jobber: If the reasoning is like "stick around longer for error"...

TLDR we should actually remove error as an available Toast type and only use Banner for errors (but that's a separate issue, imo)

adambobadam: Okay gotcha!  Ya, not even an issue here, just figured I'd ask if something like `info` should stick around for slightly longer than `success`/`default`

PR #793, opened 2021-12-06
feat(components): RecurringSelect Component

ZakaryH:
## Motivations

We have an open source gem, [Recurring Select](https://github.com/GetJobber/recurring_select) that works, but is a very old approach to creating the front end that powers it, along with the general UX of a modal it presents to build the recurring schedule.

While doing some work around rebuilding the scheduling section in React, we ended up essentially re-creating this Recurring Select in React.

We'd like to add this to Atlantis for a couple main reasons:
- First: like I mentioned, there's our open source gem that people use for this - meaning we've already established that it is valuable for others, this is doing it in a more modern way and of course in React/JS 
- Second: we have interest from an internal team, as well as an external team who would like to leverage this functionality that we have built

### Main Questions
- What should the prop & inputs look like?
- What pieces need to be there vs what can be implemented on your own (eg. the summary section, that really doesn't need to be decided by us how it looks) instead could we expose the result and allow the consumer to choose how to show it? I have a newer comment on this, I feel like it's really not part of this particular component
- Related to the above, what exactly might the output look like, presumably we want the recurring schedule that has been built, which is really just data in a specific format
- Any design tweaks needed
- Accessibility changes: this is accessible as is, but it is only navigable with tabs whereas perhaps we'd want arrow key functionality?

### Note Worthy Details

Not to rush it or anything, but we do have a 3rd party who would like to use this component ASAP.

It is already quite accessible via keyboard.

All states & combinations of those states are accounted for with styling: focus, hover, selected, disabled.

All the screenshots below contain our implementation of how to display the summary in words and UI. Those were simply values passed into a `DescriptionList` component.

### Screenshots & examples

**most of these contain the summary section, but I probably shouldn't have included those - tying that in here is still pretty up for debate**

<details>
  <summary>Daily with our version of summary at bottom</summary>

  
image
</details>

<details>
  <summary>Weekly, random days with selected & hover states</summary>

  
image
</details>

<details>
  <summary>Weekly, disabled with selected and non selected</summary>

  
image
</details>

<details>
  <summary>Weekly with weekdays (clever wording comes from rrule), also some selected & focus state</summary>

  
image
</details>

<details>
  <summary>Monthly, day of month with some selected</summary>

  
image
</details>

<details>
  <summary>Monthly, day of month selected and disabled</summary>

  

image
</details>

<details>
  <summary>Monthly, day of week with some selected</summary>

  

image
</details>

<details>
  <summary>Monthly, day of week, selected and disabled</summary>

  

image
</details>

<details>
  <summary>Yearly</summary>

  

image
</details>

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


ZakaryH:
CI is unhappy for a number of reasons, still fixing things up and moving some stuff around

wanted to get the proposal up for early thoughts ASAP

deepvishwas:
Changed the title a bit to make ci happy

ZakaryH:
ok this is now rendering on the Schedule mdx page so you can play around with it

however the path I took to get this to render has raised a few questions around this "component"

it very much feels like something bigger than a component. I don't know if that's a `pattern` or what, but the way I'm thinking about it, it's almost its own npm package - much like recurring-select is its own gem

to me the value here is
- a nice, accessible UI/UX to select the custom recurring schedule
- the output or data, which used to be a form field value on the gem version

these are very tied together and I'm figuring out what that might look like in the Atlantis version

does this need to become a form hook, or form specific component? should we instead add all the individual pieces that make it up, which keeps all the business logic out?

the downside to the latter suggestion is that you then have to build out the system to achieve this again, so to me there's value in having it already figured out for you

the point I'm getting caught up on is really what this might look and behave like when implemented. let's say we do want to put it into Atlantis as a pre-built thing

```
<RecurringSelect>
```

that gets you the UI

but we also want the data that comes out of the selection. what might the code look like to achieve that?

in our original case, we have a `useState` at the top level that defines well, the state that is needed to render and then be updated by the UI. previously we passed this into the component and it gets called to update the state as needed. that state can't really live inside the component, because we want access to the version that is built after selections.

this puts us in a spot where we might have to move that specific `useState` hook to also be exported so that when you go to use `RecurringSelect` you can just grab it and pass that in

```
import { useScheduler } from "./useScheduler";

  const [schedulerState, setSchedulerState] = useScheduler();
  const onChangeSchedule = (partialNextSchedule) => setSchedulerState({ ...schedulerState, ...partialNextSchedule });

      <RecurringSelect
        onChange={onChangeSchedule}
        {...schedulerState}
      />


```

this starts to feel like a pretty big deviation from being a component. would we be ok with that? are there any other options I'm not thinking of here?

ZakaryH:
newest question: should `computeOccurence` be in here at all?

that's entirely tied to the summary component, which is entirely distinct from this. the recurrence rule you build with RecurringSelect can 100% affect it, but there's no guarantee you're interested in that portion

ZakaryH:
`lint_javascript` - I don't know why every other file is erroring in here and not just mine? also need to sort out how to import these in a way that makes it happy, or simply ignore the violations here if that makes sense

`lint_css` - I get what it wants and why, but I still disagree. organizing purely by selector specificity is not good for human readability imo. I like organizing by concept, eg. a block for disabled, a block for selected, a block for hover & focus etc.

`test` - snapshots are going to be displeased and need to be updated, I think that's all that's going on here

chriscarline:
@ZakaryH here are the visual changes that Chris M and I discussed. Essentially creating some visual space between the numbers, adding the corner radius, removing any of the coloured borders around the individual numbers, and then adding an overall border to the elements
image

chriscarline:
Week and day of week views are similar although day of week has borders in between the weeks and it stops before hitting the "1st, 2nd.." All of the spacing and radius stuff is consistent throughout the views.
image
image

joelzwarrington:
IMO this is a pattern component, similar to Page, but I don't think this contains any business logic, schedules are common between any business or product.

I think we keep it the way it's structured now. I didn't see anything odd with the props and the value provided to on change

joelzwarrington:
> IMO this is a pattern component, similar to Page, but I don't think this contains any business logic, schedules are common between any business or product.
> 
> I think we keep it the way it's structured now. I didn't see anything odd with the props and the value provided to on change

Following up on my previous comment - I do think the code to calculate the summary could be useful in a hook at a later date, but shouldn't get added here

ZakaryH:
@chriscarline new styles have been applied if you wanna check em out

packages/components/src/RecurringSelect/useScheduler.ts
```diff
@@ -0,0 +1,13 @@
+import { useState } from "react";
+import { DurationPeriod, SchedulerState } from "./types";
+
+export const useScheduler = (initialState: Partial<SchedulerState> = {}) => {
```
ZakaryH, line 4: this has been reduce down to a very small and simple function. does it still make sense to have as a hook? the `SchdulerState` is a single key now, `recurrence` and the `DurationPeriod` is the only other thing you really need, which can be imported on its own

also the name can be improved. it is no longer tied to the `Scheduler`. the name should reflect that

ZakaryH:
still adding some more examples and documentation to the mdx page, but the code is in a state I'm content with

I'd like to get some feedback on where it is at currently. there's some re-naming that could be done, assuming we don't remove it fully - so I'd like to decide on if we think some of these concepts/functions are even worth keeping around before I rename them

also on the topic of naming, does `RecurringSelect` still make sense with this slimmed down scope & behavior? it's arguably more a custom scheduler builder. if you want to make it recur based off the data you get back, that's up to you

packages/components/src/RecurringSelect/RecurringSelect.tsx
```diff
@@ -0,0 +1,211 @@
+/* eslint-disable import/no-internal-modules */
```
ZakaryH, line 1: had to do this here, and a couple similar ones

is this an acceptable solution, or is there a preferable way to solve this?

it is upset that we have this strcuture

```
RecurringSelect
--- WeeklySelect
--- MonthlySelect
--- OtherSelectComponents
```

then in this component we are importing them

chriscarline:
@ZakaryH Looking over the visuals I saw two things that could be adjusted, the rest is looking solid!

First, can we remove these little radius bumps that show up on day of week?
image

And secondly, let's reduce the height of the cells to essentially have space-smaller above and below the numbers. The width can continue to resize according to the window size as it's already doing. (I'm not sure what tomfoolery @chris-at-jobber did when we were pairing on this to get the sizing in the original screenshot I shared)
image

I will take a look at the descriptions too!

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
```diff
@@ -0,0 +1,37 @@
+.parentContainer {
+  /* useless, only here to avoid linting error */
+  display: block;
```
ZakaryH, line 3: this css crime is here because I need a selector high enough that it can know how many child components (divs) it has

in order to apply the class name I have to have a `className={styles.parentContainer}` 

in order to have a `.parentContainer` in the css, I can't leave it blank so I'm adding a completely useless piece of code to appease the linter :(((((((((((((((

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.container {
+  display: flex;
+  margin-bottom: calc(-1 * var(--border-base));
+}
+
+.parentContainer > div:nth-of-type(1) > div:nth-of-type(2) div {
```
ZakaryH, line 11: I'm not proud of everything below this

it's the only way to have the individual WeeklySelects *not* have their standard border radius that we do indeed want for individual cases

when they're stacked as 4, they each need unique border radius rules

this pairs with the `parentContainer` which is the only element in this component that can effectively keep count of how many of the WeeklySelects we have

we can't target their class names since those get generated but within the scope of the other component

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
ZakaryH: looking at this now, I care for it even less

if we can suppress a stylelint violation, I'd prefer that honestly

or if we can come up with a better overall solution that involves none of this, that also works

packages/components/src/RecurringSelect/RecurringSelect.mdx
```diff
@@ -0,0 +1,64 @@
...
+import { Avatar } from "@jobber/components/Avatar";
+import { useState } from "react";
+import { DurationPeriod, SchedulerState } from "./types";
+import { useScheduler } from "./useScheduler";
+
+# RecurringSelect
+
+<ComponentStatus stage="ready" responsive="yes" accessible="yes" />
+
+A `RecurringSelect` is used to provide an intuitive, accessible pre-built UI
```
chris-at-jobber, line 24: @chriscarline if you could flesh out the rest of the sections here (adding more "when to use / when not to use", Related Components, etc, that would be 😗 👌 - could also happen in a followup PR

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
chris-at-jobber: @zek is there a way the component could be smart enough to give classnames to first and last row containers? like



```
.MonthlyDayOfWeekSelect--selector {
  border-radius: 0; (or I guess we wouldn't even need this)
}

.MonthlyDayOfWeekSelect--container.MonthlyDayOfWeekSelect--first .MonthlyDayOfWeekSelect--selector {
  border-top-right-radius: var(--radius-large);
  border-top-left-radius: var(--radius-large);
}

.MonthlyDayOfWeekSelect--container.MonthlyDayOfWeekSelect--last  .MonthlyDayOfWeekSelect--selector {
  border-bottom-right-radius: var(--radius-large);
  border-bottom-left-radius: var(--radius-large);
}
```

Or something like that?

chris-at-jobber: Basically just wondering if there are any opportunities to get more classnames in there so it's a little less brittle

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
chris-at-jobber: If we can do something like my idea in the comment below, would that solve this issue of relying on a parent wrapper?

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.css
```diff
@@ -0,0 +1,84 @@
...
+.checkboxWrapper {
+  position: relative;
+}
+
+.checkboxWrapper:nth-child(n + 1):nth-last-child(-n + 1) {
+  grid-area: lastday;
+}
+
+/* kept present but hidden for accessibility */
+.container input[type="checkbox"] {
```
chris-at-jobber, line 30: For standards sake, can we follow the more common practice of CSS clip for this? (reference from [WebAIM](https://webaim.org/techniques/css/invisiblecontent/))

### CSS clip

```
{clip: rect(1px, 1px, 1px, 1px);
clip-path: inset(50%);
height: 1px;
width: 1px;
margin: -1px;
overflow: hidden;
padding: 0;
position: absolute;}
```

> This fairly modern technique will hide or clip content that does not fit into a 1-pixel visible area. Like off-screen content, it will be visually hidden but still readable by modern screen readers.

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.css
```diff
@@ -0,0 +1,84 @@
...
+}
+
+.dayCheckbox {
+  display: block;
+  padding: 1rem;
+  border-radius: var(--radius-large);
+  text-align: center;
+  user-select: none;
+  background: var(--color-white);
+  transition: 300ms;
```
chris-at-jobber, line 46: Can we use `--timing-base` here to be consistent with how we handle similar visual transitions in Button, etc?

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.css
```diff
@@ -0,0 +1,84 @@
...
+  top: 0;
+  left: 0;
+  width: 1rem;
+  height: 1rem;
+  opacity: 0;
+}
+
+.dayCheckbox {
+  display: block;
+  padding: 1rem;
```
chris-at-jobber, line 41:
```suggestion
  padding: var(--space-base);
```

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.css
```diff
@@ -0,0 +1,84 @@
...
+  cursor: pointer;
+}
+
+.container input[type="checkbox"]:focus + .dayCheckbox {
+  box-shadow: var(--shadow-focus);
+  background: var(--color-surface--hover);
+}
+
+.container input[type="checkbox"]:checked + .dayCheckbox {
+  color: var(--color-white);
```
chris-at-jobber, line 69:
```suggestion
  color: var(--color-text--reverse);
```

packages/components/src/RecurringSelect/components/WeeklySelect/WeeklySelect.css
```diff
@@ -0,0 +1,75 @@
...
+  opacity: 0;
+}
+
+.dayCheckbox {
+  display: block;
+  padding: 1rem;
+  border-radius: var(--radius-large);
+  text-align: center;
+  user-select: none;
+  background: var(--color-white);
```
chris-at-jobber, line 36:
```suggestion
  background: var(--color-surface);
```

darryltec (changes requested):
Stopping my review here for now. Got distracted by lots of things. I think we can break this baby up into a few more components but i'll leave that for a monday thing. Will do another round next week :D

packages/components/src/RecurringSelect/components/WeeklySelect/WeeklySelect.css
```diff
@@ -0,0 +1,75 @@
...
+  transition: 300ms;
```
chris-at-jobber, line 37:
```suggestion
  transition: var(--timing-base);
```

packages/components/src/RecurringSelect/components/WeeklySelect/WeeklySelect.css
```diff
@@ -0,0 +1,75 @@
...
+  background: var(--color-surface--hover);
+}
+
+.container input[type="checkbox"]:focus + .dayCheckbox {
+  box-shadow: var(--shadow-focus);
+  background: var(--color-surface--hover);
+}
+
+.container input[type="checkbox"]:checked + .dayCheckbox {
+  color: var(--color-white);
```
chris-at-jobber, line 50:
```suggestion
  color: var(--color-text--reverse);
```

packages/components/src/RecurringSelect/components/WeeklySelect/WeeklySelect.css
```diff
@@ -0,0 +1,75 @@
...
+  row-gap: var(--space-smaller);
+  column-gap: var(--space-smaller);
+}
+
+.checkboxWrapper {
+  position: relative;
+}
+
+/* kept present but hidden for accessibility */
+.container input[type="checkbox"] {
```
chris-at-jobber, line 21: General comment/question (and it's fine IMO if the answer is "we can't do it"), but could we extract up some of this CSS? We're writing the same styling in a few places throughout the larger component.

packages/components/src/RecurringSelect/components/WeeklySelect/WeeklySelect.css
```diff
@@ -0,0 +1,75 @@
...
+  top: 0;
+  left: 0;
+  width: 1rem;
+  height: 1rem;
+  opacity: 0;
+}
+
+.dayCheckbox {
+  display: block;
+  padding: 1rem;
```
chris-at-jobber, line 32:
```suggestion
  padding: var(--space-base);
```

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.css
```diff
@@ -0,0 +1,84 @@
...
+  opacity: 0;
+}
+
+.dayCheckbox {
+  display: block;
+  padding: 1rem;
+  border-radius: var(--radius-large);
+  text-align: center;
+  user-select: none;
+  background: var(--color-white);
```
chris-at-jobber, line 45:
```suggestion
  background: var(--color-surface);
```

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.css
chris-at-jobber:
```suggestion
  transition: var(--timing-base);
```

packages/components/src/RecurringSelect/RecurringSelect.css
```diff
@@ -0,0 +1,25 @@
...
+  align-self: center;
+  padding-right: 1rem;
+}
+
+.disabled {
+  color: var(--color-disabled);
+}
+
+.modifierText {
+  padding-left: 1rem;
```
chris-at-jobber, line 19:
```suggestion
  padding-left: var(--space-base);
```

packages/components/src/RecurringSelect/RecurringSelect.css
```diff
@@ -0,0 +1,25 @@
...
+  display: flex;
+}
+
+.container span {
+  right: var(--space-small);
+}
+
+.intervalText {
+  align-self: center;
+  padding-right: 1rem;
```
chris-at-jobber, line 11:
```suggestion
  padding-right: var(--space-base);
```

chris-at-jobber:
My review is mostly "replacing raw values with existing variables" in the CSS, but also from a visual design perspective the date sizing that @chriscarline flagged should get addressed. Can take another look when that's ready!

ZakaryH:
@chris-at-jobber that date sizing, was that something you already solved in the session with Chris C? there was some allusion to tomfoolery in an earlier comment

https://github.com/GetJobber/atlantis/pull/793#issuecomment-990232396

I ask because while changing the padding to smaller is quite trivial, maintaining the aspect ratio of a square as the container grows, and the width increases is less straightforward. to me at least

chris-at-jobber:
@ZakaryH we can do something like this (copy-paste from my DOM-mashing)

```
.DayOfMonthSelect--dayCheckbox--3Z7XC {
    display: flex;
    justify-content: center;
    align-items: center;
    padding: var(--space-small);
    aspect-ratio: 1 / 1;
    border-radius: var(--radius-large);
    text-align: center;
    -webkit-user-select: none;
    -moz-user-select: none;
    -ms-user-select: none;
    user-select: none;
    background: #fff;
    background: var(--color-white);
    -webkit-transition: .3s;
    transition: .3s;
```

There'll need to be some extra CSS for the Last Day but I think we already have special grid styling for that so should be OK?

image

chris-at-jobber:
I'm also OK with _not_ solving for aspect ratio and simply letting the dates get wider or narrower as the space allows

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
ZakaryH: I get what your idea is, and it almost works but it ends up not applying I believe due to the way the css definitions get exported, imported and ultimately used

even with specific class names assigned to the 1-4 instances, what happens is still comparable to what we have here

we can't write a class in the template, or React component of say `MonthlyDayOfWeekSelectFirst` and have that be applied. what happens is that the markup will get `.MonthlyDayOfWeekSelectFirst` as a class, but that's not what the exported css class is.

instead what happens is that React, or whoever really, adds a random affix to ensure uniqueness/scope. this means the styles never match and don't get applied

in order for the styles to apply, you must have a `MonthlyDayOfWeekSelectFirst` definition in your `css` which then gets added to the generated `css.d.ts` file for it

then you must actually reference that in your component via `styles.MonthlyDayOfWeekSelectFirst` , meaning the `--first` syntax becomes pretty awkward. I'm not even sure if you can do that, even with a `['MonthlyDayOfWeekSelect--first']` syntax, but I didn't try admittedly

so I can get that to work, but what happens is that we're back at square 1. arguably worse. since we don't actually care about `MonthlyDayOfWeekSelectFirst`, because it's the immediate child div that we want to apply styles to, that means we have no rule of say

```
.MonthlyDayOfWeekSelectFirst { content: "pls hlp"; } 
```

rather what we have, and need is 

```
.MonthlyDayOfWeekSelectFirst div { border-radius: 5px; }
```

 but we can't apply that second rule via `styles.MonthlyDayOfWeekSelectFirst div` that's not a thing. we can only apply `styles.MonthlyDayOfWeekSelectFirst`. this means we have to have that empty definition in the css, but the linter doesn't like that. so now we need 4 empty definitions, plus 4 custom border radii selectors on the child div, which is still pretty fragile imo, and adds even more clutter

TL;DR - we can do it but I don't know if it's actually any better :/

packages/components/src/RecurringSelect/RecurringSelect.tsx
ZakaryH: @darryltec any thoughts on how to best approach this? is the lint disable fine?

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
chris-at-jobber: https://media.giphy.com/media/xUPGcIm3J0teFwk1UI/giphy.gif

packages/components/src/RecurringSelect/components/WeeklySelect/WeeklySelect.css
ZakaryH: I'm getting so sad from these suggestions that I fully agree with, but run into pretty big blockers when trying to implement :(

so, I can indeed move the shared styles up into their own file that might look like this

```
RecurringSelect
-- DateCellCheckbox.css
-- components
---- DayOfMonthSelect
---- WeeklySelect
```

I can also import those top level styles into those `components`

as is tradition, things fall apart when I try to actually use those imported styles

in order to avoid polluting the global space with a rule like `input[type="checkbox"] { display: none }` we prefixed it with the scoped component class of `container` that you can see in the line this comment/thread was started off of

what the problem is here, is that when the styles are up in `DateCellCheckbox.css` `.container` is not the same `.container` as the one down in the component where we use the styles

this leaves us with 2 options:

import the `DateCellCheckbox.css` into the top level RecurringSelect component despite it having no actual interest in checkboxes at that level. then add a div that surrounds the element that will be swapped out with the various selects (weekly, monthly etc.). on this div, apply the `DateCellCheckbox.container` and now all the rules will be correctly referenced and scoped. once again we have the issue of a useless css class ruleset, because the linter doesn't like empty rules, but we don't actually have any css to apply here. all we want is the scoping and reference to the correct container :| (this is turning into a pattern)

OR perhaps add a new class to identify the `input[type="checkbox"]` which allows us to not need to scope it anymore

I went with approach 1 (it works!) because I was troubleshooting and trying to understand the exact mechanics of the issue, but I could swap to the second if we really want!

ZakaryH:
addressed the majority of feedback

chose to throw in the towel on one after giving it the ol' college try (less fragile DOM selectors & rules to handle border-radii on stacked WeeklySelects)

still a couple items I'm looking for feedback on
https://github.com/GetJobber/atlantis/pull/793#discussion_r766086168
https://github.com/GetJobber/atlantis/pull/793#discussion_r766112843

having trouble adding more examples that initialize with Monthly/Weekly, but that plus a description of the output is what I'd still like to add for documentation

packages/components/src/RecurringSelect/DateCellCheckbox.css
```diff
@@ -0,0 +1,80 @@
+.container {
+  display: block;
+}
+
+.checkboxWrapper {
+  position: relative;
+}
+
+/* kept present but hidden for accessibility */
+.container input[type="checkbox"] {
```
darryltec, line 10: hmm. can we add a class name on that input instead of nested styling?

ZakaryH: possibly, yes

however if, like the rest of the style changes I've been making in here, it turns out to be a huge pain

I'm leaving to become a farmer or something

chris-at-jobber: 🧑‍🌾

packages/components/src/RecurringSelect/RecurringSelect.mdx
```diff
@@ -0,0 +1,64 @@
+---
+name: RecurringSelect
+menu: Components
+route: /components/recuringselect
```
chris-at-jobber, line 4:
```suggestion
route: /patterns/recurringselect
```

chris-at-jobber: (Typo on the component name, and I think moving this one up to Patterns better surfaces the complexity it's handling)

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.tsx
```diff
@@ -0,0 +1,92 @@
...
+import checkBoxStyles from "../../DateCellCheckbox.css";
+import { DayOfMonth } from "../../types";
+
+interface DayOfMonthSelectProps {
+  selectedDays: Set<DayOfMonth>;
+  disabled: boolean;
+  onChange(selectedDays: Set<DayOfMonth>): void;
+}
+
+const daysInMonth = [
```
darryltec, line 12: alternatively `[...Array.from({length: 31}, (_, i) => i + 1), "LAST"]`

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.test.tsx
```diff
@@ -0,0 +1,81 @@
+/* tslint:disable no-relative-imports */
```
darryltec, line 1: do you need this?

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
darryltec: line 11 would add it back to the `d.ts` no? Have you tried removing it?

chris-at-jobber:
Just one thought, not a blocker but... is it weird that there's no control for the Yearly option? 
image

I know that this is modeled after the RecurringSelect gem and the expectation is that perhaps there's an external date selection happening, but it seems like in isolation I might want to say 

"Every 1 year on Dec. 25th" or "Every year on the last day of February" or something.

@joelzwarrington would that make sense in the context of the work your team is doing?

packages/components/package.json
```diff
@@ -46,6 +46,7 @@
     "react-markdown": "^4.1.0",
     "react-popper": "^2.2.5",
     "react-router-dom": "^5.2.0",
+    "rrule": "^2.6.8",
```
darryltec, line 49: From our chat. Don't need this anymore. uninstall it and once that's done, run `npm ci` again on the root folder since lerna is a silly goose that does something with the package locks

darryltec (changes requested):
Talked offline in. All in all, this works as expected. Some parts needs to be simplified tho

- The value that you pass in could just be the same type as what's getting returned back on the callback
- Some disable linter comments is not needed
- Warnings on imports should be _okay_ for now. We're aware of the issue
- Replace snapshot tests with a more specific test

^ I'll have another looksies once that's done. Maybe next time it's at the CSS land and tests ;)

chris-at-jobber:
Design wise, I am good with this, assuming @chriscarline is as well!

packages/components/src/RecurringSelect/RecurringSelect.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+describe("RecurringSelect", () => {
+  beforeEach(() => {
+    onChange = jest.fn();
+  });
+
+  it("should render the recurring select component", () => {
+    const tree = renderer
+      .create(
+        <RecurringSelect
+          data-testid="boop"
```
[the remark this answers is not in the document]
ZakaryH: I thought it did? but you're right, nothing is asking for it. I will remove it

packages/components/package.json
ZakaryH: exactly. this was needed for `computeOccurrence` which has been cut from the scope of this PR

can be removed now

packages/components/package-lock.json
```diff
@@ -115,39 +115,6 @@
 				}
 			}
 		},
-		"@jobber/design": {
```
ZakaryH: I'm not so sure about some of these changes, not what I was expecting or intending to do here

I removed rrule and did `npm ci`

chriscarline:
> Design wise, I am good with this, assuming @chriscarline is as well!

I'll have another look once the preview is up but I imagine it's all good since the last time I saw it

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
ZakaryH: after chatting with Darryl, it might not be as bad as I thought with the empty rules. that was my IDE being a slowpoke

I'll see if I still have the work I tried here stashed

packages/components/src/RecurringSelect/components/WeeklySelect/WeeklySelect.css
ZakaryH: jk went with option 2 and it's much much better

ZakaryH:
> > Design wise, I am good with this, assuming @chriscarline is as well!
> 
> I'll have another look once the preview is up but I imagine it's all good since the last time I saw it

everything is in wingdings and bright red, just like you asked

packages/components/package-lock.json
darryltec: yeah, that's some lerna magic that I still don't understand. Basically running `npm install` on the components folder yields a different result than doing it on the root folder. It all boils down to the `lerna bootstrap` doing it's magic on each package to link them up.

one day, we'll figure out what it's doing. but that day is not today haha. It also has a specific way of adding a package (that I just learned a week ago)

```
npx lerna add rrule@latest packages/components
```

which runs `npm install` on packages/components but with more magic like not doing this unintended change. probably.

packages/components/src/RecurringSelect/RecurringSelect.tsx
```diff
@@ -0,0 +1,201 @@
...
+      </div>
+      {isMonthly && (
+        <MonthlySelect
+          disabled={disabled}
+          onChange={onChangeType}
+          selectedMonthOption={recurrenceRule.type}
+        />
+      )}
+      <div className={checkBoxStyles.container}>
+        {currentRecurrenceComponent}
```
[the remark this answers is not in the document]
ZakaryH: do you feel like this is a nice-to-have, or a must do?

I ask because it's not obvious to me what problem that is solving. eg. is it to follow best practices, to make the component smaller, more readable etc.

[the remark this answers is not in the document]
darryltec: @ZakaryH you pretty much answered your question there haha. all of the above :P

packages/components/package-lock.json
ZakaryH: seems like this was incorrectly added, and now it is being removed as it should be

the git history says I added jobber/design to the /components level, which isn't what we want I'm pretty sure

joelzwarrington:
> Just one thought, not a blocker but... is it weird that there's no control for the Yearly option? image
> 
> I know that this is modeled after the RecurringSelect gem and the expectation is that perhaps there's an external date selection happening, but it seems like in isolation I might want to say
> 
> "Every 1 year on Dec. 25th" or "Every year on the last day of February" or something.
> 
> @joelzwarrington would that make sense in the context of the work your team is doing?

@chris-at-jobber I think ultimately we would use a single date selector (inside or outside of this component) if yearly was selected, so might make sense to include that change here maybe in a quick follow PR so that this PR can be closed out sooner?

chris-at-jobber:
@joelzwarrington makes sense to me! Agreed on not holding up this PR with more functionality.

ZakaryH:
feedback has been addressed

last 2 commits are there to add another example of the Month & the disabled state to the markdown since it's impossible to see what that looks like without adding something like a button to toggle the disabled state

from that, when I tried putting 2 on the same file, it got confused and always triggered one element, so I swapped to the `uuid` approach to make sure it's always referencing unique ids/labels

packages/components/src/RecurringSelect/DateCellCheckbox.css
```diff
@@ -0,0 +1,76 @@
...
+
+@supports (aspect-ratio: 1/1) {
+  .dayCheckbox {
+    /*
+     * disabling property-no-unknown here as this is a real property
+     * that stylelint just doesn't know about. We may be able to remove
+     * this when we update our stylelint to not care about IE11.
+     */
+    /* stylelint-disable-next-line property-no-unknown */
+    aspect-ratio: 1/1;
```
darryltec, line 37: @chris-at-jobber @chriscarline i'm assuming we good with this bois being big bois when it's on a wide parent?
image

same goes for the month
image

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const daysInMonth = [...Array.from({ length: 31 }, (_, i) => i + 1), "LAST"];
+
+export function DayOfMonthSelect({
+  selectedDays,
+  disabled,
+  onChange,
+}: DayOfMonthSelectProps) {
+  const onChangeDayOfMonth = (
```
darryltec, line 20: just a best practice thing in react and js land. it's kinda on most of the components here as well

`on` is what you'd use as a prop, `handle` is what you feed in to said prop. I.e., `onClick={handleClick}` 

also would prefer if it's a function below the return and not an arrow one. No biggie tho.

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.tsx
```diff
@@ -0,0 +1,66 @@
...
+              checked={isSelected}
+              value={day}
+            />
+            <label
+              className={checkboxStyles.dayCheckbox}
+              htmlFor={`${day}-${inputId}`}
+            >
+              {`${day === "LAST" ? "Last day" : day}`}
+            </label>
+          </div>
```
darryltec, line 61: could be it's own component as it's the exact idea as the one used in week. but not a big deal as well

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
```diff
@@ -0,0 +1,32 @@
...
+  border-top-left-radius: 0;
+}
+
+.order {
+  align-self: center;
+  flex: 1 1 0;
+}
+
+.selector {
+  flex: 9 1 0;
```
darryltec, line 31: this is probably the first time i've actually seen someone use 9 as the value.

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.css
```diff
@@ -0,0 +1,32 @@
+.container {
+  display: flex;
+  margin-bottom: calc(-1 * var(--border-base));
+}
```
darryltec, line 4: can do

```suggestion
.container {
  display: grid;
  grid-template-columns: var(--space-largest) 1fr;
  align-items: center;
  gap: var(--space-small);
  margin-bottom: calc(-1 * var(--border-base));
}
```

and you won't have to add the last 2 css class and probably cut down the amount of nth-child you have there

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+          const next: typeof selectedWeeks = [...selectedWeeks];
+
+          next.splice(index, 1, weekDay);
+
+          onChange(next);
+        };
+
+        return (
+          <div className={styles.container} key={`${index}${week}`}>
+            <div className={styles.order}>{`${getOrderString(index)}`}</div>
```
darryltec, line 51: Might wanna wrap this baby under `<Text />` if you do the grid styling i suggested, you don't even need the div.

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+      return "3rd";
+    }
+    case 3: {
+      return "4th";
+    }
+    default: {
+      return "";
+    }
+  }
+};
```
darryltec, line 31: can also do

```
const orderString = ["1st", "2nd", "3rd", "4th"];
```

and use it as 

```jsx
<Text>{orderString[index]}</Text>
```

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+        const onChangeWeek = (weekDay: Set<WeekDay>) => {
+          const next: typeof selectedWeeks = [...selectedWeeks];
+
+          next.splice(index, 1, weekDay);
+
+          onChange(next);
+        };
+
+        return (
+          <div className={styles.container} key={`${index}${week}`}>
```
darryltec, line 50: order string should be unique enough for a key, no?

packages/components/src/RecurringSelect/components/MonthlySelect/MonthlySelect.css
```diff
@@ -0,0 +1,7 @@
+.radioContainer div[role="radiogroup"] {
+  flex-direction: row;
```
darryltec, line 2: ah, sneaky!

darryltec (commented):
This actually looks good. I'll let you decide if you want to fix those comments now or on a follow up. Assuming you agree with them, that is.

Will do a QA pass tomorrow. I think it should be good!

darryltec:
also gonna assume that the input being able to show up to hundreds is intentional. Tho, the real issue is that the text slightly cut off tho

image

packages/components/src/RecurringSelect/RecurringSelect.tsx
```diff
@@ -0,0 +1,110 @@
...
+} from "./types";
+import { Option, Select } from "../Select";
+import { InputNumber } from "../InputNumber";
+import { InputGroup } from "../InputGroup";
+import { Content } from "../Content";
+import { Text } from "../Text";
+
+interface RecurringSelectProps {
+  disabled?: boolean;
+  recurrenceRule: RecurrenceRule;
```
darryltec, line 19: just for consistency with other components in the atlantis. 
```suggestion
  readonly disabled?: boolean;
  readonly recurrenceRule: RecurrenceRule;
```

also, this probably should be `value` since it's technically a huge input. Make it easy to guess what the prop be if most of our controlled components uses a `value` and `onChange` combo

packages/components/src/RecurringSelect/RecurringSelect.tsx
```diff
@@ -0,0 +1,110 @@
...
+  onChange(next: RecurrenceRule): void;
```
darryltec, line 20:
```suggestion
  onChange(value: RecurrenceRule): void;
```

consistency and also make it easier for people that would be using this to assume what that param be

packages/components/src/RecurringSelect/types.ts
```diff
@@ -0,0 +1,138 @@
...
+export type RecurrenceRule = {
+  interval: number;
+} & (
+  | RecurrenceRuleDay
+  | RecurrenceRuleWeek
+  | RecurrenceRuleYear
+  | RecurrenceRuleDayOfMonth
+  | RecurrenceRuleWeekDayOfMonth
+);
+export interface InitialStateMap {
```
darryltec, line 84: Just OCD on my part, either you put space in between or no space. Choose 1 home boi.

darryltec (changes requested):
QA'd it and it works! Just have a few question on the interaction tho

- Should you be able to leave the interval part empty? Should that throw an error?
image
- The moment you put a 0 on it, it gets hella angry, but then after unfocusing, it keeps the 0 but removes the error 🤔 
image
image
  - The layout weirdness fix is in [this pr](https://github.com/GetJobber/atlantis/pull/762)
- This may be a future thing but a validator prop would probs have to exist and work with the `<Form />` eventually

packages/components/src/RecurringSelect/RecurringSelect.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+        onChange={onChange}
+        disabled={false}
+      />,
+    );
+
+    const intervalEl = container.querySelector(
+      'input[name="schedule-recurrence-interval"]',
+    );
+    expect(intervalEl).not.toBeNull();
+    expect(intervalEl instanceof HTMLInputElement).toBe(true);
```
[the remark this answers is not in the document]
ZakaryH: we found that using `toBeDefined` returned `true` even if it was `null` so this was being extra safe with the results of the `container.querySelector`

packages/components/src/RecurringSelect/components/DayOfMonthSelect/DayOfMonthSelect.tsx
ZakaryH: I would like to follow up on this one :)

darryltec: sounds good!

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.tsx
ZakaryH: I was under the impression using `index` as a `key` while iterating stuff in React was not good. pretty sure that's the only reason here

darryltec: index on its own, yes. i actually meant using the `orderString[index]`. either way, it's no biggie

packages/components/src/RecurringSelect/components/MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect.tsx
```diff
@@ -2,7 +2,8 @@ import React from "react";
 import styles from "./MonthlyDayOfWeekSelect.css";
 import { Text } from "../../../Text";
 import { WeekDay } from "../../types";
-import { WeeklySelect } from "../index";
+// eslint-disable-next-line
+import { WeeklySelect } from "../WeeklySelect/WeeklySelect";
```
ZakaryH, line 6: tried to change this to `from "../index";` where we are exporting the components from `index.ts`

this allowed me to remove the eslint disable rule, which seems preferable

the issue there is that we then get a https://github.com/import-js/eslint-plugin-import/blob/v2.18.2/docs/rules/no-cycle.md

on the `export { MonthlyDayOfWeekSelect } from "./MonthlyDayOfWeekSelect/MonthlyDayOfWeekSelect";
` line in `components/index.tsx`

not exactly sure what the best course of action is here. we can either disable this and live with it, or we can stop exporting from the `index.ts` perhaps, but that means disabling eslint on a few other imports at which point if we're comparing 1 eslint disable vs plural, that seems like a clear cut winner to me

darryltec: Aside from the real issue of cycle dependency, this is basically a sign that the file structure combined with imports needs a bit of work

packages/components/src/RecurringSelect/RecurringSelect.test.tsx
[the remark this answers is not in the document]
darryltec: ah yeah, we've used `toBeDefined` in like 1 places. we either check instance of HTML because those can't be null, or `toBeInTheDocument()`

packages/components/src/RecurringSelect/DateCellCheckbox.css
darryltec: chris said yes

packages/components/src/RecurringSelect/RecurringSelect.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+        onChange={onChange}
+        disabled={false}
+      />,
+    );
+
+    expect(queryAllByText(/S/)).toBeTruthy();
+    expect(queryAllByText(/T/)).toBeTruthy();
+    expect(queryByLabelText(/F/)).toBeTruthy();
+    expect(queryByLabelText(/W/)).toBeTruthy();
+    expect(queryByLabelText(/M/)).toBeTruthy();
```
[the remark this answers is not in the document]
ZakaryH: while yes, that probably would be more specific I feel like the intent/value of this test as it stands now is more of a "did it render" test. snapshots are arguably better for that, but we decided to turf those which I agree with

so I'm leaning towards leaving this as is

ZakaryH:
that _should_ be everything from the last round of feedback, at least for dev things

there were a couple comments about validation, empty input, big numbers etc that I feel are more design/UX. if we decide we want that it can be implemented (probably)

chris-at-jobber:
I'm assuming that this [cropping of the digits](https://github.com/GetJobber/atlantis/pull/793#issuecomment-995326928) is an issue with InputNumber and not RecurringSelect? I think it's OK that the number can go into the hundreds, it's maybe unlikely that someone would need to but I don't think there's a great reason to stop it from happening.

And re: the aspect ratio, yes we're good with letting the day selections grow as the container gets wider.

darryltec (approved):
LGTM. The cycle error is a legit error tho so a follow up PR to fix that would be nice.

PR #794, opened 2021-12-06
chore(deps): Remove datepicker header type patch

darryltec:
## Motivations

- I thought this was the reason why the upgrade to node 16 was failing upon running install, but it's just one of the affected parts
- Anyhoo, this patch should be _patched out_ anyway so here's a PR

### Added

- 

### Changed

- Bumped @types/react-datepicker version to latest to grab the header types and not require patch

### Deprecated

- 

### Removed

- Patch package
- Patch file for react date picker header prop types

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/package.json
```diff
@@ -28,7 +26,7 @@
     "@std-proposal/temporal": "0.0.1",
     "@types/color": "^3.0.1",
     "@types/lodash": "^4.14.136",
-    "@types/react-datepicker": "^4.1.7",
+    "@types/react-datepicker": "^4.3.2",
```
adambobadam, line 29: Kind of just thinking out loud here, but confirmation might be cool.  IIRC When I worked on this with Eddy the patch was to get typing for the datepicker.  So I assume upgrading it takes care of that?

darryltec: Yup. Can test it like this, if the patch ain't there, typescript will get angry at you.

adambobadam (approved):
LGTM
- Started Atlantis from scratch
- Removed prop types to see if there any complaints and there were!
- Ran tests locally
- Checked DatePicker locally and in preview

Question because I just generally don't know: could this be brought in to Jobber now to see if it works fine there too?

darryltec:
> Question because I just generally don't know: could this be brought in to Jobber now to see if it works fine there too?

@adambobadam can always do a pre-release and install it on jobber. I doubt this will break that tho as the interface it's adding is used internally (DatePickerCustomHeader). Once it's compiled, it won't have a knowledge that said interface existed.

PR #796, opened 2021-12-09
chore(deps): Upgrade atlantis to node 14

darryltec:
## Motivations

Node 10 is dead. Long live Node 14!

### Added

- 

### Changed

- Bumped the minimum node version of Atlantis to 14 instead of 10
- Node 13+ has a full-icu support now and would use the proper date format per locale settings. https://stackoverflow.com/a/56624712
  - This forced us to update our test be more mindful of locale

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


chris-at-jobber (commented):
A smoke-test looks good, checked various components and design features, as well as made sure the some hooks are working as expected.

Will let an engineer do the code review 😃

darryltec:
Tests on iTerm seem to 💥  with the locale date string. WIll look into it!

packages/components/src/FormatDate/FormatDate.test.tsx
```diff
@@ -2,17 +2,51 @@ import React from "react";
...
-    const tree = renderer.create(<FormatDate date={value} />).toJSON();
-    expect(tree).toMatchSnapshot();
+describe("Different date values", () => {
+  const dates = {
+    CivilDate: new CivilDate(2020, 2, 26),
+    ISO8601DateString: "2019-03-30T00:45",
+    Date: new Date("2019-03-30T00:45"),
+  };
+
+  const mockDateResult = [`Feb 26, 2020`, `Mar 30, 2019`, `Mar 30, 2019`];
```
adambobadam, line 16: Repeat date here

darryltec: It's looping through 3 tests with different variables but return the exact date 😛

packages/components/src/FormatDate/FormatDate.test.tsx
```diff
@@ -2,17 +2,51 @@ import React from "react";
...
+    expect(mockDateSpy).toHaveBeenCalledWith(mockLocale, mockLocaleDateFormat);
+  });
+
+  it("should return the proper date format", () => {
+    expect(strFormatDate(expectedDate)).toBe(
+      new Date(expectedDateString).toLocaleDateString(
+        mockLocale,
+        mockLocaleDateFormat,
+      ),
+    );
```
adambobadam, line 50: Question: This test was added because after the upgrade Node supports ICU right?  So basically this test will break if we ever revert back to node 10?

darryltec: nope. this test was added because it's missing

packages/components/src/FormatDate/FormatDate.test.tsx
adambobadam: Oh understood!

packages/components/src/FormatDate/FormatDate.test.tsx
```diff
@@ -2,17 +2,51 @@ import React from "react";
...
+    ISO8601DateString: "2019-03-30T00:45",
+    Date: new Date("2019-03-30T00:45"),
+  };
+
+  const mockDateResult = [`Feb 26, 2020`, `Mar 30, 2019`, `Mar 30, 2019`];
+
+  Object.entries(dates).forEach(([inputType, value], index) => {
+    it(`renders a formatted date from ${inputType}`, () => {
+      const tree = renderer.create(<FormatDate date={value} />).toJSON();
+      expect(tree).toMatchInlineSnapshot(
```
adambobadam, line 21: Honest question: Can we use `toMatch` instead?  Don't we normally try to avoid snapshots?  What happens if they're inline?

darryltec: ah yup. totally! It basically just returns a string inline.

```jsx
expect(foo).toMatchInlineSnapshot(`"Foo"`)
```

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.test.tsx
```diff
@@ -4,142 +4,130 @@ import { cleanup } from "@testing-library/react";
...
-afterEach(cleanup);
 beforeEach(() => {
-  Date.now = jest.fn(() => 1593115122000);
+  jest.useFakeTimers("modern");
+  jest.setSystemTime(new Date(1593115122000));
+});
+
+afterEach(() => {
+  jest.useRealTimers();
+  cleanup();
```
adambobadam, line 14: Just wondering here!  It's not obvious right away, but why useFakeTimers before and then useRealTimers after?

darryltec: because instead of mocking what the `Date.now` returns (what it used to be), I mocked the system time so I can reset it back easily once i'm done.

Pasted_Image_2021-12-13__5_12_PM

adambobadam: Ah so that's why all the `Date.now()`s are replaced with `Date()`.  Was there a reason for this change having to do with Node? Sorry I'm being nosey 😛

darryltec: In a way. Basically removing redundancy resulted to this.

adambobadam: Thanks :)

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.test.tsx
```diff
@@ -4,142 +4,130 @@ import { cleanup } from "@testing-library/react";
...
-  expect(
-    renderer.create(<FormatRelativeDateTime date={testDate} />).toJSON(),
-  ).toMatchInlineSnapshot('"Jun 25, 2018"');
+  const dates = getMockDates(testDate);
+
+  Object.values(dates).forEach(value => {
+    expect(
+      renderer.create(<FormatRelativeDateTime date={value} />).toJSON(),
+    ).toEqual(
+      new Date("Jun 25, 2018").toLocaleDateString(undefined, {
```
adambobadam, line 114: Should `undefined` here be pulled out into a variable like in your other tests?

adambobadam:
Asking because I don't know.  

Just looking into other places where node is called.  For this PR do we need to double check where node is being called in commands? E.g., `postpublish` inside `package.json`, or since that doesn't directly affect any components do we need to worry about that in this PR?

packages/components/src/FormatRelativeDateTime/FormatRelativeDateTime.test.tsx
darryltec: nope. that one i'm testing differently than here. That one I actually cared about it and match it up with something, this one, not really

adambobadam: gotcha! Thanks!

adambobadam:
LGTM 👍

adambobadam (approved):

PR #797, opened 2021-12-09
chore(deps): Bump UUID to latest

darryltec:
### Added

- 

### Changed

- Updated UUID to v8.x to replace the deprecated 3.x

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Known issue that needs fixing
- [x] The test seems to fail on iTerm but not vscode terminal (FIXED!)

---

Random photo of Atlantis


adambobadam:
Checked out `InputText` and I got this badboi:

image

adambobadam:
Hmm interesting the tests pass, but there are definitely some uuid errors going on.  Are our tests legit enough?  Maybe a follow up issue?

darryltec:
> Hmm interesting the tests pass, but there are definitely some uuid errors going on. Are our tests legit enough? Maybe a follow up issue?

It passes because we're mocking what UUID returns on the tests, thus on the test it's always there. But on the real thing, the UUID disappeared because of how the way it works changed. Basically the mock wasn't updated to what the real world  would do.

> Are our tests legit enough?

It is. This one just required an in-depth knowledge with UUID's and it being mocked in the tests. Lesson here, avoid mocks unless you absolutely need to. 

> Why do we mock the UUID then?

Because we use snapshots on older components, so if we don't mock the UUID the snapshot test would always change every-time the test ran. Lesson here is to not use snapshot as a way to test things.

> How can we avoid it?

Add at least 1 test that doesn't require mocking the UUID

adambobadam:
```It passes because we're mocking what UUID returns on the tests, thus on the test it's always there. But on the real thing, the UUID disappeared because of how the way it works changed. Basically the mock wasn't updated to what the real world would do.```

Got it!!

**Comment by adambobadam** [comment]
LGTM, hard to test though right now since all the tests are failing locally for node v10.  So maybe I'll have another quick look when we figure that issue out!

**Comment by darryltec** [comment]
There seems to be a problem with the tests in this one as well. will figure that one out as well

**Comment by adambobadam** [comment]
@darryltec Is this failing for the same reason as master is right now?  Maybe some memory issue (hence the 137 error)?

**Comment by darryltec** [comment]
Yup!

File: packages/components/__mocks__/uuid.ts [hunk 1]
```diff
@@ -11,9 +6,5 @@ function fakeUUID() {
...
 }
 
-const uuid = fakeUUID as UUIDFunction;
-
-uuid.v1 = fakeUUID;
-
-// eslint-disable-next-line import/no-default-export
-export default uuid;
+export const v1 = fakeUUID;
+export const v4 = fakeUUID;
```
adambobadam, line 10: Question for my understanding:  How is `v4` being used here?  Throughout all the changes it looks like we needed to use only `v1`

darryltec: Did not dig into it. If you have it running, try removing it and run the test.

adambobadam: Tests still pass!

darryltec: huh, i guess it doesn't need to be there then

adambobadam: Seems that tests also break when you specifically use v4 instead of v1

darryltec: that's a problem for when you actually need to use v4 or v3 or other v's really.

adambobadam:
LGTM

- Did Atlantis smoke test IDs are still different for different components.
- Checked that IDs are disparate:
E.g., Radio vs. Menu:
google_b5f6d601-5dd6-11ec-8636-6317efc3e1ec
5d56d901-5dd6-11ec-a788-37dca2a739cb
- Ran tests locally on Node14

adambobadam (approved):

PR #806, opened 2021-12-15
feat(components): Update checkbox to use react components as a label

ChristianLee-Jobber:
## Motivations

One common use cases of checkboxes is for Terms of Services, where a link to the ToS will be embedded in to the checkbox description.
Screen Shot 2021-12-15 at 12 11 08

### Added

New props `onClick` and `children` are able to be passed to the checkbox

### Changed

In order to not introduce any breaking changes, there are now two types of checkbox props. One of them is essentially the old functionality with the label as a string, the other one allows children to be passed.

## Testing

A new section has been added to the checkbox documentation at the end called "With an embedded react component"

https://user-images.githubusercontent.com/81261618/146234276-228d1570-f68a-4109-8870-6cf6d70cc4b9.mov

The example with a link as a child component now allows for tab focus.

https://user-images.githubusercontent.com/81261618/146319690-5aa03249-8765-4e3f-8698-1f93a36b28bc.mov

---

Random photo of Atlantis


packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -97,6 +112,10 @@ export function Checkbox({
             <Text>{label}</Text>
           </span>
         )}
+
+        {children != undefined && (
```
jackson-zhipeng-chang, line 116: i didn't test it, but I think this can be simplified by doing `{childre &&`

chris-at-jobber (changes requested):
Accessibility issue - the user cannot focus child elements. In the example of including an `a` element in the label, it skips over the label to the next focusable element in the view

https://user-images.githubusercontent.com/39704901/146240804-f95fc1e9-090f-42d5-9c30-2e1b63b4e1f3.mov

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -48,10 +44,26 @@ interface CheckboxProps {
...
+  readonly label?: string;
   onChange?(newValue: boolean): void;
 }
 
+interface CheckboxChildrenProps extends BaseCheckboxProps {
+  /**
+   * Component children, which shows up as a label
+   */
+  readonly children?: JSX.Element;
+  onClick?(newValue?: boolean): void;
```
darryltec, line 62: What's the onclick for?

ChristianLee-Jobber: It allows the link to be clicked without ticking the box. The link will have an onclick and will use `e.preventDefault()` to prevent the call from going up the chain

ChristianLee-Jobber: Actually, I'm testing it again, it might not be necessary. I'll get back to you on that

jackson-zhipeng-chang: as @darryltec mentioned, It's a bit hard to tell the difference between `onChange` and `onClick` from the first glance. What do you think about renaming the `onClick` function to something like `onLinkClick`(something along that line) so that people can directly tell the difference between those two functions

ChristianLee-Jobber: It ended up not needing an onClick, so I actually just ended up removing that!

jackson-zhipeng-chang:
can you add some tests to cover some common use cases and invalid use cases?

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -92,11 +105,13 @@ export function Checkbox({
...
         </span>
 
-        {label != undefined && (
+        {label && (
           <span className={styles.label}>
             <Text>{label}</Text>
           </span>
         )}
+
+        {children && <span className={styles.label}>{children}</span>}
```
darryltec, line 114: can also write it this way

```suggestion
        {(label || children) && (
          <span className={styles.label}>
            {label ? <Text>{label}</Text> : children}
          </span>
        )}
```

between the `||` or duplicating `<span className={styles.label}>` i'd choose the `||` since it's much easier to understand than trying to make sure there's no difference between the 2 `span`

jackson-zhipeng-chang: I still like the way to keep the check of `label` & `children` separate. But I also see your point that we should avoid reusing the `<span>` here @darryltec 
So after chatting with @kingstonfung, Kingston suggested a good way to satisfy both our intentions

```
const labelText = label ? <Text>{label}</Text> : children;

{ labelText && (<span className={styles.label}>{labelText}</span>) }
```

jackson-zhipeng-chang: oh wait! I can see a problem with the suggestion 👆 , if we defined both label and children, then we will only be able to show one of them 🤔

ChristianLee-Jobber: I'm using XOR to prevent both label and children from being defined, because that is undefined behaviour

jackson-zhipeng-chang: oh okay then!

darryltec: i like that suggestion

packages/components/src/Checkbox/Checkbox.test.tsx
```diff
@@ -39,6 +40,37 @@ it("renders each variation of checked, defaultChecked and indeterminate", () =>
...
+});
+
+it("allows react components to be passed as children", () => {
+  const handleLinkClick = jest.fn();
+  const { getByText } = render(
+    <Checkbox description="Checkers">
+      <Text>
+        I agree to the
+        <a
+          style={{
```
jackson-zhipeng-chang, line 56: just wondering why did you have style code here? tests should not care about the style, and even if it cares, we shouldn't define a new style in the test

jackson-zhipeng-chang: I understand we should have `<a>` here, but I was just thinking we don't necessarily need to have it here. what do you think about that?

ChristianLee-Jobber: Yes, that makes sense to me. The test should work without it

packages/components/src/Checkbox/Checkbox.test.tsx
```diff
@@ -39,6 +40,37 @@ it("renders each variation of checked, defaultChecked and indeterminate", () =>
...
+        </a>
+      </Text>
+    </Checkbox>,
+  );
+  const TOSAnchorElement = getByText("Terms of Service");
+  expect(TOSAnchorElement).toBeInTheDocument();
+  fireEvent.click(TOSAnchorElement);
+  expect(handleLinkClick).toHaveBeenCalled();
+});
+
```
jackson-zhipeng-chang, line 73: also, could you add more tests to cover more cases? for example: 
1. passing in a regular element and the test should pass (already done 👍 )
2. passing in some random things, (e.g. plain text)
3. passing in some invalid values, (not sure what cases will be considered as invalid for this change, maybe plain text 👆 ?)
4. also some edge cases should be covered, e.g. passing in an empty element and so on

I didn't look super close to this, so some of my suggestions above may or may not make sense. Just use it as general guidance when you add more tests

packages/components/src/Checkbox/Checkbox.test.tsx
```diff
@@ -39,6 +40,37 @@ it("renders each variation of checked, defaultChecked and indeterminate", () =>
...
+        >
+          Terms of Service
+        </a>
+      </Text>
+    </Checkbox>,
+  );
+  const TOSAnchorElement = getByText("Terms of Service");
+  expect(TOSAnchorElement).toBeInTheDocument();
+  fireEvent.click(TOSAnchorElement);
+  expect(handleLinkClick).toHaveBeenCalled();
```
darryltec, line 71: interesting. is this to check if the onclick on a child still fires inside a checkbox? if so, when would it not be?

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -52,6 +48,22 @@ interface CheckboxProps {
...
+   * Label that shows up beside the checkbox.
+   */
+  readonly label?: string;
+}
+
+interface CheckboxChildrenProps extends BaseCheckboxProps {
+  /**
+   * Component children, which shows up as a label
+   */
+  readonly children?: JSX.Element;
```
darryltec, line 62: we normally use react element for this

```suggestion
  readonly children?: ReactElement;
```

packages/components/src/Checkbox/Checkbox.mdx
```diff
@@ -120,6 +123,57 @@ that the user has lost control of the interface.
...
+If you require more functionality than a simple string for the label, it is
+possible to pass a child component to the checkbox to use as a label. This
+allows us to embed a link into the checkbox label.
+
+<Playground>
+  {() => {
+    const [checked, setChecked] = useState(false);
+    const [isModalOpen, setIsModalOpen] = useState(false);
+    function handleOpenModal(e) {
+      e.preventDefault();
```
darryltec, line 137: In most cases, we put the most common scenario that the people can just copy paste. This one, however, needs a bit more context as to why there's a prevent default there. It's not really meant to prevent the default of the link, it's preventing the default of the `label` that holds this link.

```suggestion
      // Prevent the checkbox from being checked
      e.preventDefault();
```

darryltec (changes requested):

chris-at-jobber:
The good: I can now focus a child element (ie a link)

The less-good (specific to element usage here): I cannot activate the `a` element with the `return` key because the link is operating off of an onClick, instead of navigating; this should be a `button`. When I do this, it works:

image

packages/components/src/Checkbox/Checkbox.test.tsx
ChristianLee-Jobber: The invalid use case seems hard to test. I've looked at other components that accept children components and none of them seem to go that far either. I can definitely add plain text and an empty element though, thanks for the feedback!

ChristianLee-Jobber: I just double checked the tests in `Modal.test.tsx` actually. For testing children it only has one test "modal shows the children and a close button". Do we need to update the tests for Modal to cover more cases, or are we getting out of hand with testing possible children?

packages/components/src/Checkbox/Checkbox.test.tsx
ChristianLee-Jobber: Yeah, that was just me being very pedantic with ensuring that the child component behaves how I want it to. I've looked at other components that accept children and their tests regarding children components seem to not go too far. For example, I looked at `Modal.test.tsx` and it seems that what matters to them is that the child gets rendered and that's it.

This kind of relates to the comment that @jackson-zhipeng-chang left. I don't know how far we normally go with our tests as this is my first time modifying an Atlantis component, so I was trying to be very detailed.

packages/components/src/Checkbox/Checkbox.test.tsx
jackson-zhipeng-chang: I feel that we just need to care about this change and cover as many as possible cases for our new changes. I had another thought about adding empty component tests and other tests, I think you can just add the tests that you think make sense to you!

ChristianLee-Jobber: I understand. I liked your edge case test, so I added the test for an empty component. We can't do any tests with passing invalid input though - thanks to the magic of Typescript!

ChristianLee-Jobber:
@chris-at-jobber thanks for the tip. I added the `tabIndex` to the `<a>` element originally because it was not working without it. Your solution just works though, so I'll go with what you have

jackson-zhipeng-chang:
other than the suggestions from @chris-at-jobber and @darryltec, it looks good to me!

packages/components/src/Checkbox/Checkbox.test.tsx
darryltec: Ok yeah, just wanted to know the reason behind it. It was odd at first because it looked like you're testing if the onclick fires on the `a` tag but that's already a tried and true thing for raw HTML tags. 

That to me seems like a test that you would do when you do use it in a component that does exactly that interaction. On the other hand tho, it is a good test to see that the onClick within the children still fires if the parent does eventually prevent it from firing. Just making sure that I understand what you're doing and see if there's another way to test it/make it clearer by separating it on it's own `it`.

If you think this is a good test, then yeah, maybe put it on another `it("should still fire the onClick within the children")` or something like that so it's clear what exactly it's for.

ChristianLee-Jobber: Good call out, I separated the tests so that there is one for ensuring `onClick` works within the children, and another for rendering a non-empty child

packages/components/src/Checkbox/Checkbox.test.tsx
```diff
@@ -19,6 +20,23 @@ it("renders a disabled Checkbox", () => {
...
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+it("renders a checkbox with children", () => {
+  const childContent = <Text>Content</Text>;
+  const tree = renderer.create(<Checkbox>{childContent}</Checkbox>).toJSON();
+  expect(tree).toMatchSnapshot();
+});
```
darryltec, line 38: we generally avoid using snapshots in atlantis because of how it's hard to know exactly what you're testing. (Sorry if this is not clear, we're working on clearing that up soon)

This should make it more specific on what exactly are you testing.

```suggestion
it("renders a checkbox with children", () => {
  const { getByText } = render(<Checkbox>{<Text>Content</Text>}</Checkbox>);
  expect(getByText("Content")).toBeInTheDocument();
});
```

packages/components/src/Checkbox/Checkbox.test.tsx
```diff
@@ -19,6 +20,23 @@ it("renders a disabled Checkbox", () => {
...
+it("renders a checkbox with an empty element", () => {
+  const tree = renderer
+    .create(
+      <Checkbox>
+        <></>
+      </Checkbox>,
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
+});
```
darryltec, line 32: interesting test. a fragment is technically treated the same as any react element like `<Text />` so, not too sure if you should actually test it.

If you really want to test it, you can avoid that snapshot and add a `data-testid` on the label span and check via `expect(getByTestId("YOR_TEST_ID").children).tohaveLength(0)`if it has a child

packages/components/src/Checkbox/Checkbox.test.tsx
```diff
@@ -19,6 +20,23 @@ it("renders a disabled Checkbox", () => {
   expect(tree).toMatchSnapshot();
 });
 
+it("renders a checkbox with an empty element", () => {
```
darryltec, line 23: let's group the children tests under a `describe()`

darryltec (changes requested):
One last thing @ChristianLee-Jobber, I promise!

packages/components/src/Checkbox/Checkbox.test.tsx
jackson-zhipeng-chang: yeah, I first suggested testing the empty element as it was my first impression that it might be an edge case,  but after a second thought, I think it's eventually the same as other elements like `<Text />`. It was my bad to suggest that in the first place, so I will leave it to you @darryltec @ChristianLee-Jobber to decide whether to keep it or not. But I totally agree with @darryltec, it's the same as other elements

packages/components/src/Checkbox/Checkbox.test.tsx
ChristianLee-Jobber: Sounds good! I have also changed some of the other tests to not use snapshots anymore as well

darryltec: @ChristianLee-Jobber thank you sir! Muchas gracias!

packages/components/src/Checkbox/Checkbox.test.tsx
ChristianLee-Jobber: Thanks for bringing that up @darryltec. I have removed the test

darryltec (approved):
Looks real good mah dudes @ChristianLee-Jobber @jackson-zhipeng-chang 

@chris-at-jobber any last words on the design side?

chris-at-jobber:
LGTDesign

PR #807, opened 2021-12-15
fix(components): Prevent date picker from flickering / closing when click the activator again

darryltec:
## Motivations

On button activator, the date picker flickers when you click it
On InputDate, the date picker closes permanently when you click/focus on the field again

It shouldn't be like that. It should just keep the date picker open when clicking the activator again

### Added

- 

### Changed

- Added the `react-datepicker-ignore-onclickoutside` class name on our custom component
  - Investigated how [react date picker](https://reactdatepicker.com/) works and that's how it determines if the calendar stays open or not

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.test.tsx
+++ b/packages/components/src/DatePicker/DatePicker.test.tsx
@@ -65,3 +65,15 @@ it("always appears when inline", () => {
 
   expect(getByText("15")).toBeInstanceOf(HTMLDivElement);
 });
+
+it("should not add the `react-datepicker-ignore-onclickoutside` when inline", () => {
+  const { container, rerender } = render(<DatePicker onChange={jest.fn()} />);
+
+  const target = container.firstChild;
+  const className = "react-datepicker-ignore-onclickoutside";
+  expect(target).toHaveClass(className);
+
+  rerender(<DatePicker onChange={jest.fn()} inline />);
+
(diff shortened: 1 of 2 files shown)


adambobadam:
LGTM

- New test accounts for possible scenarios
- InputText component's behaviour is not affected either (blinking is removed, however)
- Tested component out in deploy preview

Question:  In other code bases there are tests to check that certain functionality still exists when third party packages are used.  Should we somehow also check that this class is still valid? Wouldn't be sure how to.

darryltec:
> Question @darryltec: In other code bases there are tests to check that certain functionality still exists when third party packages are used. Should we somehow also check that this class is still valid? Wouldn't be sure how to.

Since that class name is pretty embedded (and not exported) in the component, the only way we can test it is if the 3rd-party component itself uses that in one of the elements. But since we have a custom activator, the usage of the class name disappears so... in short... it would be nice, but we can't as of this moment.

adambobadam:
> Since that class name is pretty embedded (and not exported) in the component, the only way we can test it is if the 3rd-party component itself uses that in one of the elements. But since we have a custom activator, the usage of the class name disappears so... in short... it would be nice, but we can't as of this moment.

That makes sense!  Thanks, man!

adambobadam (approved):

PR #808, opened 2021-12-16
chore(components): Add test for datepicker to make sure custom css doesnt fail

darryltec:
## Motivations

We're overwriting and making use of classnames internal to ReactDatePicker, this ensures no surprises happens.

## Changes

- Wrote a test to ensure the classnames from ReactDatePicker that we're ~ab~ using still exists.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.test.tsx
+++ b/packages/components/src/DatePicker/DatePicker.test.tsx
@@ -1,5 +1,6 @@
 import React from "react";
-import { cleanup, fireEvent, render } from "@testing-library/react";
+import { act, cleanup, fireEvent, render } from "@testing-library/react";
+import ReactDatePicker from "react-datepicker";
 import { DatePicker } from "./DatePicker";
 
 afterEach(cleanup);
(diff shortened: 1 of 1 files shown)


packages/components/src/DatePicker/DatePicker.test.tsx
```diff
@@ -77,3 +72,48 @@ it("should not add the `react-datepicker-ignore-onclickoutside` when inline", ()
...
+      ".react-datepicker-wrapper",
+      ".react-datepicker-popper",
+      ".react-datepicker__header",
+      ".react-datepicker__month",
+      ".react-datepicker__day",
+      ".react-datepicker__day--outside-month",
+      ".react-datepicker__day--selected",
+      ".react-datepicker__day-names",
+      ".react-datepicker__day-name",
+    ];
```
adambobadam, line 100: Question: Just wondering if we do actually need to test these, or if we should just worry about just the ones that affect our system like the `react-datepicker-ignore-onclickoutside` class?

adambobadam: @darryltec Woops my bad!  All of these are in the `DatePicker.css` file!  So we gotta check these exist

adambobadam (approved):
LGTM

PR #810, opened 2021-12-16
chore(components): Make datepicker import css module instead of general css [JOB-38626]

robbyjobber:
## Motivations
- we are now importing the CSS module of react-datepicker into our component to unchoke our NEXTJS builds.

## Testing
- [ ] go into datepicker and input date and see that the CSS from react-datepicker is being properly applied 

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.css.d.ts
+++ b/packages/components/src/DatePicker/DatePicker.css.d.ts
@@ -1,4 +1,5 @@
 declare const styles: {
+  readonly "react-datepicker__month--selecting-range": string;
   readonly "datePickerWrapper": string;
   readonly "fullWidth": string;
   readonly "datePicker": string;
--- a/packages/components/src/DatePicker/DatePicker.css
+++ b/packages/components/src/DatePicker/DatePicker.css
@@ -1,3 +1,5 @@
+@import "react-datepicker/dist/react-datepicker.module.css";
+
 .datePickerWrapper {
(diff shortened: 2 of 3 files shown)


packages/components/src/DatePicker/DatePicker.css.d.ts
```diff
@@ -1,4 +1,5 @@
 declare const styles: {
+  readonly "react-datepicker__month--selecting-range": string;
```
darryltec, line 2: Was curious where did this come from. It was just some nested `scss` sillyness

Normall nested SCSS

https://github.com/Hacker0x01/react-datepicker/blob/e791571c2e8ba694b2080dab8ba49758844bc434/src/stylesheets/datepicker.scss#L483-L488

getting nested under `:global`

https://github.com/Hacker0x01/react-datepicker/blob/master/src/stylesheets/datepicker-cssmodules.scss

and because it's getting written as `.react-datepicker__month--selecting-range & ` it will be added before `:global`.

image

SCSS Protip: Look at the `&`

```scss
.foo {
  $.bar {}
}

/* output: .foo.bar {} */

.foo {
  .bar {}
}

/* output: .foo .bar {} */

.foo {
  .bar & {}
}

/* output: .bar .foo {} */
```

darryltec (approved):

darryltec:
For transparency, datepicker css does get rendered without css

image

PR #811, opened 2021-12-17
fix(components): Move import of ReactDatePicker CSS back to DatePicker.tsx

darryltec:
## Motivations

Release broke when #810 got introduced because the `bootstrap` (aka `npm ci`) removes `react-datepicker__month--selecting-range` from `.d.ts` but `npm start puts it back. When releasing, I believe it's only running `bootstrap` which leaves an uncommitted change which cancels out the release.

This PR ensures the `npm ci` and `npm start` doesn't clash

### Added

- 

### Changed

- Read title
- But this time, it's importing `.module.css` as opposed to just `.css`

It still uses the `:global` instead of without it. Which fixes the nextjs bug. We think.

image

Before #810, it was only using `.css` to import so it will give you a result with no `:global`

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.css
+++ b/packages/components/src/DatePicker/DatePicker.css
@@ -1,5 +1,3 @@
-@import "react-datepicker/dist/react-datepicker.module.css";
-
 .datePickerWrapper {
   display: inline-block;
 }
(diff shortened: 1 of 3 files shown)


darryltec:
Prerelease runs perfectly so we can assume the release would be fixed

image

adambobadam (approved):

PR #812, opened 2021-12-20
fix: DatePicker smart focus [JOB-39118, JOB-38255]

darryltec:
## Motivations

Focusing on the selected date when the date picker opens was removed when we added the input date. This brings it back. This also makes the date picker refocus on the activator.

### Added

- Adds back the autofocus on selected date
- Refocus on activator 
  - After selecting a date
  - After hitting escape (only when the calendar itself is focused) 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- When the date picker is opened
  - When a date is selected
    - It should move the focus to the selected date
    - Hitting `esc` key should put the focus back to the activator
    - Choosing another date should close the date picker and put the focus back to the activator
  - When a date is NOT selected
    - The focus should still be in the activator
    - Choosing a date should close the date picker and put the focus back to the activator
    - Tabbing to the date picker and hitting esc should put the focus back to the activator
  - Through InputDate
    - It should not do any of the above

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.test.tsx
+++ b/packages/components/src/DatePicker/DatePicker.test.tsx
@@ -111,6 +111,28 @@ describe("Ensure ReactDatePicker CSS class names exists", () => {
   });
 });
 
+describe("Smart autofocus", () => {
+  it("should focus on the selected date", async () => {
+    const { getByTestId } = render(
+      <DatePicker selected={new Date()} onChange={jest.fn} />,
+    );
+    await popperUpdate(() => fireEvent.click(getByTestId("calendar")));
+
+    expect(document.activeElement).toHaveClass(
+      "react-datepicker__day--selected",
+    );
+  });
+
+  it("should NOT focus on the selected date if there's none", async () => {
+    const { getByTestId } = render(<DatePicker onChange={jest.fn} />);
+    await popperUpdate(() => fireEvent.click(getByTestId("calendar")));
+
+    expect(document.activeElement).not.toHaveClass(
+      "react-datepicker__day--selected",
+    );
+  });
+});
+
 async function popperUpdate(event: Function) {
   event();
   // Wait for the Popper update() so jest doesn't throw an act warning
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -1,4 +1,10 @@
-import React, { MutableRefObject, ReactElement, useRef } from "react";
+import React, {
+  MutableRefObject,
+  ReactElement,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
 import classnames from "classnames";
 import ReactDatePicker from "react-datepicker";
 /**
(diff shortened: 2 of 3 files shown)


packages/components/src/DatePicker/DatePicker.test.tsx
```diff
@@ -111,6 +111,28 @@ describe("Ensure ReactDatePicker CSS class names exists", () => {
...
+describe("Smart autofocus", () => {
+  it("should focus on the selected date", async () => {
+    const { getByTestId } = render(
+      <DatePicker selected={new Date()} onChange={jest.fn} />,
+    );
+    await popperUpdate(() => fireEvent.click(getByTestId("calendar")));
+
+    expect(document.activeElement).toHaveClass(
+      "react-datepicker__day--selected",
+    );
```
adambobadam, line 123: Couple questions here:

Should there be
```
    expect(document.activeElement).toHaveFocus();
 ```
 
 Also, does it make sense to check if the calendar icon/activator is focused again once a date is selected?

File: packages/components/src/DatePicker/DatePicker.tsx [hunk 2]
```diff
@@ -21,6 +28,12 @@ interface BaseDatePickerProps {
    */
   readonly selected?: Date;
 
+  /**
+   * Determines if the focus moves to the selected date (if any) or back to
+   * the activator
+   */
+  readonly smartAutofocus?: boolean;
```
adambobadam, line 35: Wondering if this needs to be optional, since it's always either true or false in inputdate or datepicker

adambobadam:
Lookin pretty swanky, just a couple of questions!

packages/components/src/DatePicker/DatePicker.tsx
darryltec: @adambobadam let's say it's not optional. What would it do to the component when you do start using them?

packages/components/src/DatePicker/DatePicker.test.tsx
darryltec: > Should there be the following check in this test:

I'll let you check what `document.activeElement` is and then ask yourself that same question 😉 

> Also, does it make sense to check if the calendar icon/activator is focused again once a date is selected?

Already tested by the hook itself

adambobadam:
```
I'll let you check what document.activeElement is and then ask yourself that same question 😉
```
The reason why I asked is because even if the date is not "focused" and just "selected" it will still have the `react-datepicker__day--selected` class.

packages/components/src/DatePicker/DatePicker.tsx
adambobadam: Ahh, I guess what I was asking is if it's a necessary prop to expose since it's just hard-coded as true and false in two places and we don't use it anywhere else inside datepicker and inputdate

darryltec: In atlantis, yeah. Who knows how many places DatePicker would be used outside Atlantis. So if this is required, you'd always have to declare `smartAutofocus` on every usage. 

As opposed to being optional and set it to true, always 
image

Until the consumer says, y'know what, i don't want to

adambobadam (approved):
LGTM Talked with Darryl about my comments

PR #813, opened 2021-12-20
chore(components): Update package json to accommodate the DatePicker import

robbyjobber:
## Motivations
- we are now importing the CSS module of react-datepicker CSS into our component to unchoke our NEXTJS builds. The postcss-import is the plugin that handles the importation of external CSS into our CSS modules which is only run during `npm run build` and `npm start` in our local dev.  In production, this is not the case hence why it fails.
- `package.json` has now been updated to not run `npm run build:CSS` in components since we are already pushing up the CSS types created from the gatsby/webpack `postcss-import` plugin.  This script is the reason why it is breaking because it keeps overwriting what we currently have committed in our branch.
- this `package.json` configuration `"build:cssTypes": "tcm src",` creates the CSS types but is missing the import CSS types because it does not run the postcss-import plugin like this example`"build:css:icons": "postcss src/icons/*.css --dir icons/ --use postcss-import"` that exists in `packages/foundation`.  

charlie-day

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "prestart": "cd packages/components",
     "start": "USE_ATLANTIS_ALIASES=true docz dev",
     "start:withPrivateComponents": "USE_ATLANTIS_ALIASES=true PRIVATE_COMPONENTS=visible docz dev",
     "test": "jest",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "prestart": "cd packages/components && npm run build:cssTypes",
+    "prestart": "cd packages/components",
     "start": "USE_ATLANTIS_ALIASES=true docz dev",
     "start:withPrivateComponents": "USE_ATLANTIS_ALIASES=true PRIVATE_COMPONENTS=visible docz dev",
     "test": "jest",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -7,9 +7,8 @@
   "scripts": {
     "patch": "npx patch-package",
     "prebuild": "npm run patch",
-    "build": "npm run build:cssTypes && npm run build:rollup",
+    "build": "npm run build:rollup",
     "build:rollup": "rollup --config",
-    "build:cssTypes": "tcm src",
     "clean": "rm -rf dist/* tsconfig.tsbuildinfo",
     "prepublishOnly": "ts-node --project ../../tsconfig.bin.json scripts/entryPoints.ts",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "prestart": "cd packages/components && npm run build:cssTypes",
```
darryltec, line 7: any reason to keep this?

robbyjobber: @darryltec i don't think so.  Might be prudent to also remove that script in `packages/components` when we're relying on the plugin to build our CSS types anyways

robbyjobber: I'm waiting to see from circle if we actually need it.  Need some advise on how to confirm this in local as well.

adambobadam:
@robbyjobber Will this fail CI when it is merged?  

Doing an `npm install` (during check for manual release) removes 

```
readonly "react-datepicker__month--selecting-range": string;
```

Which when published might throw the same uncommitted changes error?

Maybe we push it to master just to see? 😛

robbyjobber:
@adambobadam could you try again.  I pushed some new changes

darryltec:
did a pre-release, seems to be importing the correct css. @adambobadam wanna try this with nextjs? 
```
npm i @jobber/components@2.71.1-pre.2+1588ab5
```

darryltec:
CI failing because of an outage https://status.circleci.com/

adambobadam:
> CI failing because of an outage https://status.circleci.com/

Ya defs! Looks like a gateway error, didn't seem related to the config!

adambobadam:
@robbyjobber @darryltec Seems to be working in nextjs!

image

darryltec:
CI and Docker is still borked tho

darryltec (changes requested):
Investigated a little bit more

- Found out that there's a rollup config under components and design that does the _real_ build with the correct css imports and whatnot that gets outputted under `dist` which the release uses.
- `tcm src` was there to make sure we have a `.d.ts` for every new css file before doing the rollup
- In short, we need to make sure that when rollup happens, it also renders the missing `.d.ts`

> When would a missing `.d.ts` happen?

When you generate a new component (`npm run generate` it'll add a css file but not the `.d.ts`) or create a new css file before running `npm start`, it'll throw an error every time you bootstraps since it can't find the `.d.ts` file for the new css you've added. Thus, the `tcm src` script.

TL;DR: Need to generate the `.d.ts` before or during the rollup stage of the build or else, it'll break on every css files that doesn't have `.d.ts` under components

robbyjobber:
circle is failing.  looking into it 👀

robbyjobber:
new pull request
https://github.com/GetJobber/atlantis/pull/818

PR #815, opened 2021-12-22
fix(components): Revert fix form field inconsistent spacing when used under Form and InputGroup JOB-37646

darryltec:
Reverts GetJobber/atlantis#762

Missed an issue with `maxLength` reducing the size. Since it's the holiday, it's much safer to revert than fix it.


robbyjobber (approved):
Revert LGTM

PR #816, opened 2021-12-22
chore: Parallelize jest test

darryltec:
## Motivations

@adambobadam noticed that we're using 1 out of the 200 available parallelized test. This PR adds a way to split that up into 4 parallels.

## To-do

- [ ] Ask CircleCI SME's if this would increase how much we pay for our usage? Since majority of of the time is used for set up than the actual testing.

### Added

- 

### Changed

- Parallelized CircleCi test step into 4

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,6 +118,7 @@ jobs:
 
   test:
     <<: *container_config
+    parallelism: 4
     steps:
       - *restore_repo
       - *sum_lockfiles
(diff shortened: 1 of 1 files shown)


darryltec:
Closing this because it only shaves off a few seconds 

- On average, the test runs for 4 mins
- Setup and `npm ci` takes over 3 mins
- The actual test run takes 40-ish seconds
- Even if you parallelize it, it still has to do setup on each so it'll shave off around 30 seconds or less

If we want to speed up the test then we should focus on only running `npm ci` once

.circleci/config.yml
```diff
@@ -118,15 +118,17 @@ jobs:
...
       - *restore_repo
       - *sum_lockfiles
       - *restore_npm_cache
       - *npm_install
       - run:
           name: "Run tests with jest"
-          command:
-            npm test -- --runInBand --no-cache --coverage --reporters=jest-junit
+          command: |
+            TEST=$(circleci tests glob **/packages/*/src/**/*.test.tsx | circleci tests split --split-by=timings)
```
darryltec, line 130: the `glob **/packages/*/src/**/*.test.tsx` is to avoid circle from including `node_modules`. I tried the `!(node_modules)` but I can't seem to get it working in circle.

darryltec:
Closing again since speeding this up doesn't actually speed up the overall time since build_docs still takes a while

PR #817, opened 2021-12-23
fix(components): Move focus to hook

adambobadam:
## Motivations

Hook was introduced because the DatePicker test [here](https://github.com/GetJobber/atlantis/blob/master/packages/components/src/DatePicker/DatePicker.test.tsx#L114) was a false positive because initially the DatePicker library focuses on mount, so it would be the case that the test will always pass.

To fix this @darryltec suggested creating a separate hook that handles focusing on the date hence `useFocusOnSelectedDate`   

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -1,6 +1,6 @@
 import React, { ReactElement, useEffect, useState } from "react";
 import classnames from "classnames";
-import { ReactDatePicker } from "react-datepicker";
+import ReactDatePicker from "react-datepicker";
 /**
  * Disabling no-internal-modules here because we need
  * to reach into the package to get the css file.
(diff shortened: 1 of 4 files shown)


packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -1,12 +1,6 @@
...
-  MutableRefObject,
-  ReactElement,
-  useEffect,
-  useRef,
-  useState,
-} from "react";
+import React, { ReactElement, useEffect, useState } from "react";
 import classnames from "classnames";
-import ReactDatePicker from "react-datepicker";
+import { ReactDatePicker } from "react-datepicker";
```
darryltec, line 3: That test failed because of this was changed to something that is not a component

robbyjobber (approved):
Code LGTM adam. Great job man 👏

PR #818, opened 2021-12-23
chore(components): Update package json to accommodate the DatePicker import

robbyjobber:
## Motivations
- we are now importing the CSS module of react-datepicker CSS into our component to unchoke our NEXTJS builds. The postcss-import is the plugin that handles the importation of external CSS into our CSS modules which is only run during `npm run build` and `npm start` in our local dev.  In production, this is not the case hence why it fails.
- `package.json` has now been updated to not run `npm run build:CSS` in components since we are already pushing up the CSS types created from the gatsby/webpack `postcss-import` plugin.  This script is the reason why it is breaking because it keeps overwriting what we currently have committed in our branch.
- this `package.json` configuration `"build:cssTypes": "tcm src",` creates the CSS types but is missing the import CSS types because it does not run the postcss-import plugin like this example`"build:css:icons": "postcss src/icons/*.css --dir icons/ --use postcss-import"` that exists in `packages/foundation`.  

charlie-day

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "prestart": "cd packages/components && npm run build",
     "start": "USE_ATLANTIS_ALIASES=true docz dev",
     "start:withPrivateComponents": "USE_ATLANTIS_ALIASES=true PRIVATE_COMPONENTS=visible docz dev",
     "test": "jest",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "prestart": "cd packages/components && npm run build:cssTypes",
+    "prestart": "cd packages/components && npm run build",
     "start": "USE_ATLANTIS_ALIASES=true docz dev",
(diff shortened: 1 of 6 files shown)


robbyjobber:
@darryltec 's changes looks good to me.  Tested on nextjs prerelease and looks good

darryltec (approved):
Just gonna log my broken thoughts here. We'll do a larger documentation on how exactly the install, start, and release happen and how it affects each other

- `npm ci` and `npm start` yields a different `.css.d.ts` file because the `ci` step has no knowledge of how to read `@import` while the `start` step does and outputs the correct `.css.d.ts`
- Had the assumption that the `build:cssTypes` aka `tcm src` step was there to make sure a `.d.ts` file exists before doing the rollup when you `generate` a new component with a `.css` file
  - Since `.css.d.ts` also gets autogenerated with webpacker when you run `npm start` and that most of those `.css.d.ts` are already generated and checked into master, it is a lot safer to update the generator to include a `.css.d.ts`
- This fixes the issue of mismatched `.css.d.ts` file when releasing as the release only runs `ci` and not `start`

Follow up work if we don't do this already: Probably fail the `build_docs` step in CI when there are uncommitted files like how the release step fails when the same happens.

darryltec:
Jobber online still works flawlessly

PR #821, opened 2022-01-04
fix(components): reorganize RecurringSelect structure to avoid circular dependencies

ZakaryH:
## Motivations

When building RecurringSelect, we came across a case where our imports required an eslint suppression. Without it, and following the suggestion to fix it - we would end up with a circular dependency which is not great. https://github.com/GetJobber/atlantis/pull/793#discussion_r770986036

I promised to fix it in a follow up, https://github.com/GetJobber/atlantis/pull/793#pullrequestreview-836555122

This is that.

## Changes

"Flattened" the `components` directory to avoid the circular dependency. The issue was stemming from the fact that we had individual component folders, importing and exporting in such a way that didn't jive with eslint.

Updated the imports in other files accordingly.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


ZakaryH:
@darryltec here's the follow up to the RecurringSelect PR. turned out to be quite small

obviously doesn't have to be you who reviews it, but you'll have the most context!

darryltec (approved):
LGTM

PR #827, opened 2022-01-11
fix: Allow space bar key deselect on single select chip

darryltec:
## Motivations

Since the single select chips are just an input radio behind the scenes, deselecting via space bar isn't really a default interaction for them. This make it support the ability to deselect a chip via a space bar press

Q: What about enter press?
A: Since radio's and checkboxes doesn't get selected on "enter" press, I didn't bother including it. You can, however, select both via a space bar so it seems natural to only support said key.

### Added

- Allow single select chips to be deselected on enter press
- Test that new interaction

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- You should be able to deselect a single select chip using the space bar

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chips/InternalChipSingleSelect.tsx
+++ b/packages/components/src/Chips/InternalChipSingleSelect.tsx
@@ -46,7 +46,7 @@ export function InternalChipSingleSelect({
 
     return (event: KeyboardEvent<HTMLInputElement>) => {
       if (event.key === " ") {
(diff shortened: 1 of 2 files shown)


chris-at-jobber (commented):
Design approved

adambobadam (approved):
LGTM

PR #831, opened 2022-01-14
refactor(hooks): remove totalCount as a requirement from useCollectionQuery

kkimble006:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
+++ b/packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
@@ -70,7 +70,7 @@ interface Collection {
     hasNextPage: boolean;
     [otherProperties: string]: unknown;
   };
-  totalCount: number;
+  totalCount?: number;
   [otherProperties: string]: unknown;
 }
 
@@ -256,9 +256,11 @@ function fetchMoreUpdateQueryHandler<TQuery>(
     );
   }
 
-  Object.assign(outputCollection, {
-    pageInfo: cloneDeep(nextCollection.pageInfo),
-    totalCount: nextCollection.totalCount,
+  const totalCount = outputCollection.totalCount != undefined ? { totalCount: nextCollection.totalCount } : {};
+
+  Object.assign(outputCollection, { ...{
+    pageInfo: cloneDeep(nextCollection.pageInfo) },
+    ...totalCount
   });
 
   return output;
@@ -296,9 +298,11 @@ function subscribeToMoreHandler<TQuery, TSubscription>(
     );
   }
 
+  const totalCount = outputCollection.totalCount != undefined ? { totalCount: outputCollection.totalCount + 1} : {};
+
   Object.assign(outputCollection, {
-    pageInfo: cloneDeep(outputCollection.pageInfo),
-    totalCount: outputCollection.totalCount + 1,
+    ...{pageInfo: cloneDeep(outputCollection.pageInfo)},
+    ...totalCount
   });
 
   return output;


packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
```diff
@@ -256,9 +256,11 @@ function fetchMoreUpdateQueryHandler<TQuery>(
     );
   }
 
+  const totalCount = outputCollection.totalCount != undefined ? { totalCount: nextCollection.totalCount } : {};
```
igor-m-jobber, line 259: Won't it make sense to return `undefined` or `0` instead of empty object?
This way it will make it easier to work on the front end if we need to build some logic around `totalCount`, because `0` and `undefined` are both falsy values.

kkimble006: The returned object needs to match what was passed in, in this case there is no `totalCount` so we shouldn't set it.

packages/hooks/src/useCollectionQuery/useCollectionQuery.ts
```diff
@@ -296,9 +298,11 @@ function subscribeToMoreHandler<TQuery, TSubscription>(
     );
   }
 
+  const totalCount = outputCollection.totalCount != undefined ? { totalCount: outputCollection.totalCount + 1} : {};
+
   Object.assign(outputCollection, {
-    pageInfo: cloneDeep(outputCollection.pageInfo),
-    totalCount: outputCollection.totalCount + 1,
+    ...{ pageInfo: cloneDeep(outputCollection.pageInfo) },
+    ...totalCount
```
davidr64, line 305: Why are you putting pageInfo into an object then expanding the object? Why not define it directly in the parent object. Similarily does totalCount need to be an object that gets expanded? Could you just store it as a number? 

would this work?
`
const totalCount = outputCollection.totalCount != undefined ?  totalCount: outputCollection.totalCount + 1 :  undefined;
Object.assign(outputCollection, {
    pageInfo: cloneDeep(outputCollection.pageInfo) ,
    totalCount
})
`

kkimble006: I don't think this would work, because you would need to wrap `{ totalCount: outputCollection.totalCount + 1 }` no?
Then we would end up with: 
```
Object.assign(outputCollection, {
pageInfo: cloneDeep(outputCollection.pageInfo) ,
{ totalCount: 5 }
})
```
In my solution I am just merging 2 hashes.

davidr64: Sorry messed up my example I meant for const totalCount to be a number not an object.

davidr64: Something like this https://jsfiddle.net/pcmjqsn9/1/

kkimble006: Oh but we don't always want totalCount to be present! Unless I am missing something about your implementation. Want to chat?

davidr64: That's fair. I figured it didn't matter if the key was added to the object if the value of the property was undefined regardless. It makes sense if you want to avoid having the key on the object at all.

PR #837, opened 2022-01-19
build: Speed Up CI Config Steps

adambobadam:
## Motivations
While investigating parallel test runs @darryltec noticed that `npm ci` was being called over and over at each step after `install_dependencies`.  This was probably due to a mistake in implementation of the caching inside `config.yml`    

Decrease the total time of Atlantis builds:
Before: 11m 21s (without release step) and ~14m 45s (with release step)
After: ~8m 37s (without release step)

## Changes

- Instead of using caching after `install_dependencies` we are persisting a workspace.  Caching works from branch to branch, but persisting a workspace works during a single pipeline instance.  This is why caching is still kept during `install_dependencies` since it could be the case that the cache will be hit on another branch that uses the same dependencies.  Commits [here](https://github.com/GetJobber/atlantis/pull/837/commits/186d26cb8521ab86fbdcc8cc5cc1b9ea07a49484) and [here](https://github.com/GetJobber/atlantis/pull/837/commits/55ddfb3f3c6e5ea1979f7255681fdeb1d222d7f4)

- The `medium+` resource class is now being used for `build_docs` because we are able to use the extra resources towards something else now that we shaved off ~4mins from the build.  This change shaves off ~1min of time too!  The heap size doesn't seem to be an issue as the RAM usage stays below 80% after several runs.  Commit [here](https://github.com/GetJobber/atlantis/pull/837/commits/7970ed41674e1db89cd122e460fe248260ff3f80)

- The max limit of 1400Mb (via `NODE_OPTIONS: --max_old_space_size=1400`) can be removed now that we are using the medium+ resource, since the RAM is not being maxed out.  This limit was initially set to mimic Node 10's smaller heap size, as Node 12+ begins killing the build with it's larger one.  See Jc's PR [here](https://github.com/GetJobber/atlantis/pull/805) for more detail

- Check out these times below:

## Before (on master w/release step too)
Screen Shot 2022-01-18 at 5 54 05 PM

## After
Screen Shot 2022-01-18 at 5 53 34 PM

---

Random photo of Atlantis

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,13 +137,12 @@ jobs:
           path: coverage
 
   build_docs:
+    resource_class: medium+
     <<: *container_config
     steps:
       - *attach_workspace
       - run:
           command: node -v && node -e 'console.log(`heap_size_limit = ${v8.getHeapStatistics().heap_size_limit / (1024 * 1024)} MB`)' && npm run build -- --base "/${BASE_PATH}/"
-          environment:
-            NODE_OPTIONS: --max_old_space_size=1400
       - store_artifacts:
           path: ~/atlantis/.docz/dist
           destination: docs
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,7 @@ jobs:
     steps:
       - *restore_repo
       - checkout
+      - *persist_to_workspace
       - save_cache:
           key: v1-repo-{{ .Branch }}-{{ .Revision }}
           paths:
@@ -72,7 +73,7 @@ jobs:
   install_dependencies:
     <<: *container_config
     steps:
-      - *restore_repo
+      - *attach_workspace
       - *sum_lockfiles
       - *restore_npm_cache
       - *npm_install
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,6 +141,7 @@ jobs:
     <<: *container_config
     steps:
       - *attach_workspace
+      - *generate_base_path
       - run:
           command: node -v && node -e 'console.log(`heap_size_limit = ${v8.getHeapStatistics().heap_size_limit / (1024 * 1024)} MB`)' && npm run build -- --base "/${BASE_PATH}/"
       - store_artifacts:
@@ -158,6 +159,7 @@ jobs:
     <<: *container_config
     steps:
       - *attach_workspace
+      - *github_authenticity
       - add_ssh_keys:
           fingerprints:
             - "39:99:9f:56:1c:70:0e:53:6a:ed:64:14:c3:a8:e2:ce"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -137,17 +137,12 @@ jobs:
...
-      - *restore_repo
-      - *sum_lockfiles
-      - *restore_npm_cache
-      - *npm_install
-      - *generate_base_path
+      - *attach_workspace
       - run:
           command: node -v && node -e 'console.log(`heap_size_limit = ${v8.getHeapStatistics().heap_size_limit / (1024 * 1024)} MB`)' && npm run build -- --base "/${BASE_PATH}/"
-          environment:
-            NODE_OPTIONS: --max_old_space_size=1400
```
jessevanherk, line 150: can you add a little bit more detail to the PR description about this change? You mentioned heap size but we should be explicit about this one.

.circleci/config.yml
```diff
@@ -162,11 +157,7 @@ jobs:
   release:
     <<: *container_config
     steps:
-      - *restore_repo
-      - *sum_lockfiles
-      - *restore_npm_cache
-      - *npm_install
-      - *github_authenticity
```
jessevanherk, line 169: why did github_authenticity get removed on this job?

adambobadam: Yeah, it's not part of the other setup steps and I don't think I should've removed it.  Thanks for that!

.circleci/config.yml
adambobadam: Added another bullet point!

darryltec:
I wonder if this part of the build docs is a leftover from https://github.com/GetJobber/atlantis/pull/45 where `deploy_docs` got added that makes use of that special workspace persistence below. Then `deploy_docs` got removed in favour of netlfy but the persistence was left https://github.com/GetJobber/atlantis/pull/189

image

@jessevanherk do you know?

_PS: We don't have to action upon it right now if this PR is good to go. All in all, it looks good to me and netlify didn't get angry anymore (gonna assume it's because of `generate_base_path` being added back in again)._

darryltec (approved):
LGTM. 

I wonder if we should test for false positives tho. Like, make each the step from `check_for_manual_release` up to the end fail.

adambobadam:
> LGTM.
> 
> I wonder if we should test for false positives tho. Like, make each the step from `check_for_manual_release` up to the end fail.

Sounds good! Wouldn't hurt to do a few more checks hey?

jessevanherk:
> I wonder if this part of the build docs is a leftover from #45 where `deploy_docs` got added that makes use of that special workspace persistence below. Then `deploy_docs` got removed in favour of netlfy but the persistence was left #189

@darryltec @adambobadam 

the `upload_artifacts` step takes the specified files and uploads them to circleCI so they're available in the `Artifacts` tab. Normally that's only used for making test results available for circleCI to parse and show in their UI, or for screenshots of a failed test.  I can't see how netlify or any other system would be fetching them from there, so it's actually probably a step we can remove for the build_docs job in a followup!

Here's what it looks like for this PR:

image

adambobadam:
@jessevanherk Thanks a bunch!  I was unsure if I should remove the artifact part of that step, but that clears it up!

PR #839, opened 2022-01-19
feat(hooks): Create the hooks index dynamically

eddysims:
## Motivations

There was an issue where to create a new hook you had to remember to export it from the `index.ts` file other wise it would not be available. This PR will automagically create the `index.ts` file so that no hook will ever be left behind.

### Added

- new script to create the index.ts file

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


scripts/createIndexFile.js
```diff
@@ -0,0 +1,57 @@
...
+const fs = require("fs");
+const path = require("path");
+
+const params = process.argv.slice(2);
+
+if (params.length > 1) {
+  throw Error(
+    "Enter only the direcory that you want to create an entry point for.",
+  );
+}
```
adambobadam, line 19: Should there be a check here if the length of params is 0?  It gives a type error later on otherwise.

scripts/createIndexFile.js
```diff
@@ -0,0 +1,57 @@
...
+/* eslint-env node */
+/* eslint-disable @typescript-eslint/no-var-requires */
+const fs = require("fs");
+const path = require("path");
+
+const params = process.argv.slice(2);
+
+if (params.length > 1) {
+  throw Error(
+    "Enter only the direcory that you want to create an entry point for.",
```
adambobadam, line 17:
```suggestion
    "Enter only the directory that you want to create an entry point for.",
```

adambobadam (approved):
LGTM!  Just a couple small things

PR #840, opened 2022-01-19
build: Remove extra persist, use ssh variables

adambobadam:
## Motivations
This PR is a follow up to https://github.com/GetJobber/atlantis/pull/837.

- There is another persist and some storing of artifacts that is slowing down the `build_docs` step. 
- Also, SSH pub key for `github_authentication` and a fingerprint are put into nice CI variables.

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #843, opened 2022-01-20
build: Revert back ssh key to make release pass

adambobadam:
## Motivations

Releases are failing now because of the SSH key variable added for this PR: https://github.com/GetJobber/atlantis/pull/840

More investigation is required into why it's failing.  So for now, it's best to unblock the release process until the solution can be finalized!!!

---

Random photo of Atlantis


darryltec (approved):
LGTM. Checked if the ssh and fingerprint as the same as before

PR #846, opened 2022-01-21
feat(components): add explicit "Heading" color to Card header

chris-at-jobber:
## Motivations

Card in Atlantis is not representative of desired output. It should use the `heading` color in the header, and because it had previously only been used in projects (Jobber) that had the desired color for `h3` elements set in global CSS, the production output was never an issue.

However as external applications are being built that leverage Atlantis, we should specify the heading color so that it does not inherit from whatever the external app's `body` color is set to.

### Added

- set `textColor` on Card's heading

---

Random photo of Atlantis


joelzwarrington (approved):
LGTM 👍

PR #847, opened 2022-01-24
feat(components): Support dismissible type on Chips

darryltec:
## Motivations

We want to have a consistent way of using chips for a larger set of options you can choose from

### Added

- Support dismissible type on `<Chips />`

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## TO DO
- [x] Render menu above the input when there are no space below
- [ ] A11y when adding/deleting
- [x] Add `,` comma shortcut key
- [x] Debounce search

---

image


packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissible.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+    });
+
+    chips
+      .filter(chip => !selectedChips.includes(chip))
+      .forEach(chip => {
+        expect(within(wrapperEl).queryByText(chip)).toBeInTheDocument();
+      });
+  });
+
+  it("should trigger the onChange callback when you select from available optipns", () => {
```
adambobadam, line 86:
```suggestion
  it("should trigger the onChange callback when you select from available options", () => {
```

packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissible.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+      mockIsInView.mockReturnValueOnce(true);
+    });
+
+    it("should trigger the onLoadMore callback", () => {
+      expect(handleLoadMore).toHaveBeenCalled();
+    });
+  });
+});
+
+it("should trigger the onclick callback when a chip gets clicked", () => {
```
adambobadam, line 131:
```suggestion
it("should trigger the onClick callback when a chip gets clicked", () => {
```

packages/components/src/Chips/utils/mdxUtils.ts
```diff
@@ -0,0 +1,46 @@
+import { uniq } from "lodash";
+import { useEffect, useState } from "react";
+
+export function useFakeOptionQuery() {
+  const [options, setOptions] = useState<string[]>([]);
+  const initialDataGetUrl = "https://swapi.dev/api/people/?format=json";
```
adambobadam, line 6: This is a funny idea 😛 I am wondering if we can just use an internal mock list to avoid the API call, or if this is something you wanted to mimic?

When I also hold down tab to select all of the characters that are present, it then takes a sec to reload, so I wonder if this call is related?

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
```diff
@@ -0,0 +1,195 @@
...
+import {
+  ChipDismissibleInputOptionProps,
+  ChipDismissibleInputProps,
+} from "../InternalChipDismissibleTypes";
+import { Icon } from "../../../Icon";
+import { ChipProps } from "../../Chip";
+
+const menuId = uuidV1();
+
+export function useInternalChipDismissibleInput({
```
adambobadam, line 19: Is the naming for this verbose?  Would something slightly simpler like `getOptionsAndActions` or something like that work? I saw `RevealButton` and a `useReveal` hook in mobile, but that was a short name to begin with, I wonder if this is based on a naming convention I am unaware of.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
```diff
@@ -0,0 +1,195 @@
...
+    },
+
+    handleOpenMenu: () => setMenuOpen(true),
+
+    handleCloseMenu: () => {
+      setMenuOpen(false);
+      setActiveIndex(0);
+    },
+
+    handleCancelBlur: () => setShouldCancelBlur(true),
```
adambobadam, line 66: Wondering if these ones would read better if they were called `cancelBlur` for example?  Maybe it's a naming convention I don't know about either though!

packages/components/src/Chips/InternalChipDismissible/hooks/useRepositionMenu.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function useRepositionMenu(attachTo: RefObject<Element | null>) {
+  const [positionElement, setPositionedElementRef] =
+    useState<HTMLElement | null>();
+  const popper = usePopper(attachTo.current, positionElement, {
+    modifiers: [
+      { name: "offset", options: { offset: [0, 8] } },
+      { name: "flip", options: { fallbackPlacements: ["top"] } },
+    ],
+    placement: "bottom",
```
adambobadam, line 12: Was `placement: "bottom"` a default key value here?  Just read this here: https://popper.js.org/react-popper/v2/render-props/

packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissibleInput.test.tsx
```diff
@@ -0,0 +1,215 @@
...
+  beforeEach(() => {
+    addButton = screen.getByRole("button", { name: "Add" });
+    fireEvent.click(addButton);
+  });
+
+  it("should not show the add button", () => {
+    expect(addButton).not.toBeInTheDocument();
+  });
+
+  it("should not show an input and menu", () => {
```
adambobadam, line 83: Should this be?

```suggestion
  it("should show an input and menu", () => {
```

adambobadam (commented):
Crazy good work!

I really like the separation of logic that `useInternalChipDismissibleInput.tsx` provides!  It made it easier to follow.

There are a few random things I thought of while using the component in the preview (the first two I already brought up with @chris-at-jobber and on the side, but just reiterating here), and they are probably all follow ups since none of them are show stopping, and are mostly just ideas:

- Leaving your mouse on the list of options, and then using the arrow keys gives you two focuses (discussed: not high priority right now)
- Would it make sense when clicking to select onto the parent element of the dismissible chips to have some focus on the chips or the plus sign? (discussed: not an input, and not really worth the effort) 
- Page up/page down functionality might be nice to control the selection of the menu (might be nowhere to be found in Jobber, but wondering if it's a worthwhile thought)
- The spinner gets cut off at the bottom when you hit the bottom of the menu's items, before loading more

packages/components/src/Chips/utils/mdxUtils.ts
darryltec: Got that from the `useCollectionQuery` but opted with a simpler pure js approach since there's a lot of setup involved with using that hook. The real API call is intended to give a better feel for it.

> When I also hold down tab to select all of the characters that are present, it then takes a sec to reload, so I wonder if this call is related?

like hold and click? that won't work since it resets on every select

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
darryltec: Eh, was trying to name it the same as the parent (+ "use") so no one uses it outside if it. Imagine all components having that simple name. Auto import is gonna be hella confusing.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
darryltec: i'd keep it as handle. Keep it consistent with the other functions that gets called in an `on` props

packages/components/src/Chips/InternalChipDismissible/hooks/useRepositionMenu.ts
darryltec: looks like it

darryltec:
> Leaving your mouse on the list of options, and then using the arrow keys gives you two focuses (discussed: not high priority right now)

Fixed

> Page up/page down functionality might be nice to control the selection of the menu (might be nowhere to be found in Jobber, but wondering if it's a worthwhile thought)

Haven't really done research in to those keyboard button, but I believe most modern laptop keyboard doesn't even have those keys

> The spinner gets cut off at the bottom when you hit the bottom of the menu's items, before loading more

Screenshot? Seems to work fine on the deploy preview



@adambobadam

packages/components/src/Chips/utils/mdxUtils.ts
adambobadam: > Fixed

Woot woot!

> Haven't really done research in to those keyboard button, but I believe most modern laptop keyboard doesn't even have those keys

I think you can maybe do the same thing with like a `fn + up arrow`

> Screenshot? Seems to work fine on the deploy preview

I tried again and it seems like if you scroll to the very bottom while loading it will show the whole thing, but cuts off half of it while it's loading?  Maybe it's a me problem, because that recording you took looks good!  Here's me interacting with it anyways:

https://user-images.githubusercontent.com/88052089/151081272-7973346d-bb5c-4bd3-9bf8-f5a91d0c0dc4.mov

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
adambobadam: Ya that was a thought I had around importing too (that there might be quite a bit of confusion).  It ain't a big deal either way, just thought the name was getting quite long, that's all :)

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
adambobadam: I see, it's a consistency thing okay!

adambobadam:
> Fixed

Woot woot!

> Haven't really done research in to those keyboard button, but I believe most modern laptop keyboard doesn't even have those keys

I think you can maybe do the same thing with like a fn + up arrow

> Screenshot? Seems to work fine on the deploy preview

I tried again and it seems like if you scroll to the very bottom while loading it will show the whole thing, but cuts off half of it while it's loading? Maybe it's a me problem, because that recording you took looks good! Here's me interacting with it anyways:

https://user-images.githubusercontent.com/88052089/151081760-d8a922a3-49b4-4024-b229-58a0d7213449.mov

packages/components/src/Chips/utils/mdxUtils.ts
adambobadam: >like hold and click? that won't work since it resets on every select

I mean like just hold tab key down and let all of the selections on the menu be chosen really fast like.  Although, weirdly enough I did just try it again and it wasn't an issue!

darryltec: > Should we make a note that there is a call being made in the mdx file

Whatcha worried about?

> I mean like just hold tab key down and let all of the selections on the menu be chosen really fast like. Although, weirdly enough I did just try it again and it wasn't an issue!

Ah yeah, not worried about that. There's no native interaction where you hold a tab and it doing something expected. (Like holding the shift and pressing a letter capitalizes it.)

darryltec:
> but cuts off half of it while it's loading

@adambobadam It's because you're not scrolled all the way down. firefox seems to stop scrolling all the way down at some point. Not a big deal tho since you can still scroll it down yourself again while it's loading.

feat_components___Support_dismissible_type_on_Chips_by_darryltec_·_Pull_Request__847_·_GetJobber_atlantis

darryltec:
> I think you can maybe do the same thing with like a fn + up arrow

Even then, remove the `fn` and you'll get to the same experience. haha. Also, page up/down is meant to control pages so, i believe, browsers would scroll you down. And as the name suggest (_page_) it's for making the page go down or up. So that's an anti-pattern.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
darryltec: I'd be concerned about the name if it's a re-useable thing by any component.

packages/components/src/Chips/utils/mdxUtils.ts
adambobadam: > Whatcha worried about?

I was just thinking the data fetching might take a while for some reason, and someone will wonder "hmm, I wonder why these characters are taking a while to load"

adambobadam:
> So your suggestion would be an anti-pattern.

Okay! Makes sense!  Was just thinking in the case of infinite scrolling sites, but I think you're right that it makes sense for pages :)

adambobadam (approved):

packages/components/src/Chips/utils/mdxUtils.ts
darryltec: Haven't heard that issue from `useCollectionQuery` so I wouldn't worry about it too much

PR #851, opened 2022-02-09
fix(docs): Rename iconStyles to iconStyleProperties

darryltec:
## Motivations

When removing the transformer in the mobile repo, I found out that the name `styles` has some weird effect on React Native metro config. Removing said config (because we don't need it) makes the `iconStyles` import disappear and RN throwing an error with it not existing. Renaming `iconStyles` to something else that doesn't refer to `styles` (like style) resolves it.

### Added

- 

### Changed

- Rename iconStyles to iconStyleProperties

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

image


adambobadam (approved):
LGTM!  There is a lot of context here and intermingling of the repos and it took me a bit to figure out the context.  Should there be some PR link from the work done in the mobile repo in the description?

PR #856, opened 2022-02-17
feat(components): expose minDate and maxDate props from react-datepicker (JOB-40742)

demoraesgui:
## Motivations

Provides the ability of disabling a range of dates. This could help the developer to make the user unable to select dates that could cause issues or edge cases.

image

### Added

- `minDate` and `maxDate` props to DatePicker and InputDate

### Changed

- added ability to disable dates by using said props

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.css
+++ b/packages/components/src/DatePicker/DatePicker.css
@@ -62,6 +62,11 @@
   background: var(--color-surface--background);
 }
 
+.datePicker :global(.react-datepicker__day--disabled) {
+  color: var(--color-disabled--secondary);
+  cursor: default;
+}
+
 .datePicker :global(.react-datepicker__month) {
   margin: 0;
 }
@@ -100,6 +105,20 @@
 .datePicker :global(.react-datepicker__day--selected):hover,
 .datePicker :global(.react-datepicker__day--selected):focus {
   background: var(--color-interactive--hover);
+}
+
+.datePicker :global(.react-datepicker__day--disabled):hover,
+.datePicker :global(.react-datepicker__day--disabled):focus {
+  box-shadow: none;
+  background: transparent;
+}
+
+.datePicker
+  :global(.react-datepicker__day--disabled.react-datepicker__day--outside-month):hover,
+.datePicker
+  :global(.react-datepicker__day--disabled.react-datepicker__day--outside-month):focus {
+  box-shadow: none;
+  background: var(--color-surface--background);
 }
 
 .month {
--- a/packages/components/src/DatePicker/DatePicker.mdx
+++ b/packages/components/src/DatePicker/DatePicker.mdx
@@ -99,6 +99,31 @@ pass a `React` component as the `activator`.
   }}
 </Playground>
 
+## Disabling dates
+
+If you want to enable for selection only for a set of dates you can pass
+`minDate` or `maxDate` to the component as props.
+
+<Playground>
+  {() => {
+    const [date, setDate] = useState(new Date("02/17/2022"));
+    return (
+      <Content>
+        <div>
+          <DatePicker
+            minDate={new Date("02/07/2022")}
+            maxDate={new Date("02/27/2022")}
+            selected={date}
+            inline
+            onChange={setDate}
+          />
+        </div>
+        {date && <DataDump defaultOpen data={date} />}
+      </Content>
+    );
+  }}
+</Playground>
+
 ## Accessibility
 
 At this point, the `DatePicker` component is a wrapped version of
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -13,6 +13,16 @@ import { useFocusOnSelectedDate } from "./useFocusOnSelectedDate";
 
 interface BaseDatePickerProps {
   /**
+   * The maximum selectable date.
+   */
+  readonly maxDate?: Date;
+
+  /**
+   * The minimum selectable date.
+   */
+  readonly minDate?: Date;
+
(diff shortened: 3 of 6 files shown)


packages/components/src/DatePicker/DatePicker.mdx
```diff
@@ -99,6 +99,31 @@ pass a `React` component as the `activator`.
   }}
 </Playground>
 
+## Disabling dates
```
chris-at-jobber, line 102: I think it would be more accurate to call this section "Restricting date ranges", as disabling the actions is just a means to enforce the range restriction.

chris-at-jobber (dismissed):
Instead of using two different colorsets to indicate a disabled state, can we use our `disabled` and `disabled--secondary` colours on the dates as well?

image

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -101,6 +113,8 @@ export function DatePicker({
         disabled={disabled}
         readOnly={readonly}
         onChange={handleChange}
+        maxDate={maxDate}
```
deepvishwas, line 116: I know this is just a range but do you need a simple check for minDate < maxDate ?
#nonBlockingComment

demoraesgui: I thought about this as well but what should be the behavior when the developer puts maxDate < minDate? What's your suggestion? 

This is what happens now

image

packages/components/src/DatePicker/DatePicker.css
```diff
@@ -62,6 +62,11 @@
   background: var(--color-surface--background);
 }
 
+.datePicker :global(.react-datepicker__day--disabled) {
+  color: var(--color-disabled--secondary);
+  cursor: default;
+}
+
```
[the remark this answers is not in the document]
chris-at-jobber: Sorry, to be specific: we are using inconsistent colors for the disabled month-changing buttons and the disabled dates. We should be using `color-disabled` for the text and icon elements, and `color-disabled--secondary` for the disabled elements that already have a background.

image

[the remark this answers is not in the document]
demoraesgui: Oh, gotcha!! I'll work on that, thanks for the explanation :D

packages/components/src/DatePicker/DatePicker.tsx
deepvishwas: Since it disables the entire range if by mistake we interchange min and max, I think a simple test should suffice

demoraesgui: But what I would be testing for? This behavior is exactly the same one for react-datepicker

image

A test for "minDate shouldn't be grater than maxDate" makes sense to me if we had some visual feedback or let's say if we tested the case of all dates being disable (this I wouldn't know how to implement) like querying for all the dates and checking if --disabled class is present? I'm really not sure what to implement 😕

I think what react-datepicker decided was to leave for the developer the responsibility to make sure the dates are correct. Do you think there will be a problem if we did that as well? or do you have a test scenario and some tips on how to implement it?

packages/components/src/DatePicker/DatePicker.css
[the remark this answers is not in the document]
demoraesgui: Done.

packages/components/src/DatePicker/DatePicker.tsx
deepvishwas: Also Can do something like this
```suggestion
         maxDate={maxDate && minDate && maxDate > minDate ? maxDate : undefined}
        minDate={maxDate && minDate && maxDate < minDate ? minDate : undefined}
```

demoraesgui: maxDate={maxDate && minDate && maxDate > minDate ? maxDate : undefined}, if minDate is not provided which is a valid case, maxDate the statement would return undefined right? The same case for minDate.

deepvishwas: I think you are right,  it makes sense to leave it to the user to select the correct range, 
my only concern is if a genuine mistake happens, we should fail gracefully.
(if we interchange the min and max by mistake it should not disable the calendar)

> But what I would be testing for? This behavior is exactly the same one for react-datepicker
> 
> image
> 
> A test for "minDate shouldn't be grater than maxDate" makes sense to me if we had some visual feedback or let's say if we tested the case of all dates being disable (this I wouldn't know how to implement) like querying for all the dates and checking if --disabled class is present? I'm really not sure what to implement 😕
> 
> I think what react-datepicker decided was to leave for the developer the responsibility to make sure the dates are correct. Do you think there will be a problem if we did that as well? or do you have a test scenario and some tips on how to implement it?

True, we don't have any such feedback also we are mocking dates, so that doesn't help either. 
I can't think of any better way to write a test what can we do though is fail gracefully if the dates are not in order. Regardless, this is by no means is a blocking comment.

demoraesgui: I see what you mean, I think this discussion is valuable. What do you meant by failing gracefully in this scenario? Unfortunately i believe that disabling any date makes since because the Union of both sets is all the dates, so i really don't know what would be a good approach on this, we could move this forward and start a discussion about this, have move people engaged as well...

demoraesgui:
@chris-at-jobber @deepvishwas Are there more changes you want me to implement?

chris-at-jobber (commented):
✅  Design approved! Thanks @demoraesgui

deepvishwas (approved):
LGTM 👍🏻

PR #858, opened 2022-02-18
fix: InputText pick prefix and suffix

deepvishwas:
## Motivations

Remove linting errors related to Prefix and Suffix 

## Changes
Picked prefix and suffix types 

Random photo of Atlantis


darryltec (approved):
- doesn't error out anymore
  - image
- prefix/suffix still exists
- suffix still enforces the `ariaLabel` if `onClick` is defined and vice-versa
- prefix doesn't have any of the above

PR #863, opened 2022-03-10
fix(components): Update tooltip to use react-popper

darryltec:
## Motivations

- An issue was raised around tooltip arrows aren't aligning to the middle on both Atlantis and Jobber
  - The Atlantis issue was pretty straightforward but, the Jobber issue was a mystery
- Updating tooltip to use `react-popper`'s `usePopper` instead of `@popperjs/core`'s `createPopper` is long overdue so that needed to be updated anyway
  - We hoped that it fixes all the issues since it also removes the cruft that came with using `createPopper` and that it did!

### Added

- 

### Changed

- replaced `createPopper` with `usePopper`

### Deprecated

- 

### Removed

- Transition out has been removed because it doesn't work well with `usePopper`. Lesser of the 2 ~evils~ things and whatnot

### Fixed

- 

### Security

- 

## Testing

- [ ] The arrow is rendered in the middle of the activator
  - it's on the activator instead of the tooltip because if the tooltip renders at the edge and there's no space, it will reposition itself but the arrow is aligned to the center of the activator
- [ ] Tooltips in Jobber are rendering the same as the previous test
  - to test this, install a prerelease of Atlantis with `npm install @jobber/components@2.78.1-pre.2+176f0bc`

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.tsx
+++ b/packages/components/src/Tooltip/Tooltip.tsx
@@ -2,14 +2,13 @@ import React, {
   ReactElement,
   ReactNode,
   useLayoutEffect,
-  useRef,
   useState,
 } from "react";
 import classnames from "classnames";
 import ReactDOM from "react-dom";
 import { motion } from "framer-motion";
-import { usePopper } from "react-popper";
 import styles from "./Tooltip.css";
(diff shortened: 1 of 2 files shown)


deepvishwas (approved):
LGTM 👍🏻

PR #868, opened 2022-03-16
docs(components): Document multiline loading={true} limitation 

NickBoers-Jobber:
* Document that loading={true} does not work when multiline={true}

## Motivations

Without these notes, users might be confused when `loading={true}` does not work.

### Added

- Notes that `loading={true}` does not work when `multiline={true}`.


chris-at-jobber (approved):
Looks good, thanks Nick!

PR #869, opened 2022-03-17
fix(components): Update popover elevation and width

leticiawanderley:
## Motivations
* Popover component was being displayed **under** some elements on Jobber online
* Additionally, the popover width was limited by the width of the element it was attached to

#### Before
popover-before

#### After
popover-after

### Added

- 

### Changed

* Replaced the component's previous z-index, `--elevation-base` with `--elevation-tooltip`
* Replace the component's previous width, `100%` with `max-content`

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

* [ ] The popover is rendered above other elements on the page
* [ ] On Jobber Online, the popover width depends on its content size rather than the reference element's width
---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Popover/Popover.css
+++ b/packages/components/src/Popover/Popover.css
@@ -20,8 +20,8 @@
 }
 
 .popover {
-  z-index: var(--elevation-base);
-  width: 100%;
+  z-index: var(--elevation-tooltip);
+  width: max-content;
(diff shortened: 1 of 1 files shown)


darryltec (approved):
:shipit:

PR #872, opened 2022-03-25
feat(components): Enable chip navigation with left and right arrow keys

adambobadam:
## Motivations

This PR adds left/right arrow key focus to dismissible chips, and improves the usage of the delete button.

#### Functionality:
- When focused on input, if delete is pressed and there is a preceding chip, it will just focus on this chip
- Left and right arrow keys work as expected with focus on next/prev chips
- Some styling is adjusted and some functionality is removed because active index is no longer being used
- If the next element is just an HTMLElement of some sort and not just a button the focus will be applied as well.

#### Tests:
- Some existing tests are changed due to the changes elsewhere that broke them
- A new test is added specifically for the left right arrow keys
- A new testing section for more overall behaviour of the component is added (the second describe) 
- Another (problematic) test was added for checking the deletion of a chip and having the focus be put on the preceding chip.  See the comment in the code for this one.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -1,4 +1,5 @@
 import { KeyboardEvent, MouseEvent, ReactElement, ReactNode } from "react";
+import { XOR } from "ts-xor";
 import { ChipButtonProps } from "./InternalChipButton";
 import { AvatarProps } from "../Avatar";
 import { IconProps } from "../Icon";
@@ -52,7 +53,9 @@ export interface InternalChipProps {
   /**
    * Callback for keyboard interaction with chips (e.g., chip deletion).
    */
-  onKeyDown?(event: KeyboardEvent<HTMLDivElement | HTMLButtonElement>): void;
+  onKeyDown?(
+    event: KeyboardEvent<XOR<HTMLDivElement, HTMLButtonElement>>,
+  ): void;
 
   /**
    * If the chip needs children (e.g., empty span for screen reading).
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -5,6 +5,11 @@ import { IconProps } from "../Icon";
 
 export interface InternalChipProps {
   /**
+   * Adds the ability to control focus via tabindex
+   */
+  readonly tabIndex?: number;
+
+  /**
    * Label of the chip.
    */
   readonly label: string;
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -1,4 +1,4 @@
-import { MouseEvent, ReactElement } from "react";
+import { KeyboardEvent, MouseEvent, ReactElement, ReactNode } from "react";
 import { ChipButtonProps } from "./InternalChipButton";
 import { AvatarProps } from "../Avatar";
 import { IconProps } from "../Icon";
@@ -48,4 +48,8 @@ export interface InternalChipProps {
   onClick?(
     event: MouseEvent<HTMLDivElement | HTMLInputElement | HTMLButtonElement>,
   ): void;
+
+  onKeyDown?(event: KeyboardEvent<HTMLDivElement | HTMLButtonElement>): void;
+
+  readonly children?: ReactNode;
 }
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -53,9 +53,7 @@ export interface InternalChipProps {
   /**
    * Callback for keyboard interaction with chips (e.g., chip deletion).
    */
-  onKeyDown?(
-    event: KeyboardEvent<XOR<HTMLDivElement, HTMLButtonElement>>,
-  ): void;
+  onKeyDown?(event: KeyboardEvent<HTMLDivElement | HTMLButtonElement>): void;
 
   /**
    * Adds an element or component between the label and suffix
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -1,5 +1,4 @@
 import { KeyboardEvent, MouseEvent, ReactElement, ReactNode } from "react";
-import { XOR } from "ts-xor";
 import { ChipButtonProps } from "./InternalChipButton";
 import { AvatarProps } from "../Avatar";
 import { IconProps } from "../Icon";
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -58,7 +58,7 @@ export interface InternalChipProps {
   ): void;
 
   /**
-   * If the chip needs children (e.g., empty span for screen reading).
+   * Adds an element or component between the label and suffix
    */
   readonly children?: ReactNode;
 }
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -5,11 +5,6 @@ import { IconProps } from "../Icon";
 
 export interface InternalChipProps {
   /**
-   * Adds the ability to control focus via tabindex
-   */
-  readonly tabIndex?: number;
-
-  /**
    * Label of the chip.
    */
   readonly label: string;
@@ -46,6 +41,16 @@ export interface InternalChipProps {
    * Adds a component on the right side of the label.
    */
   readonly suffix?: ReactElement<IconProps | ChipButtonProps>;
+
+  /**
+   * Allows the chip to be focused.
+   *
+   * **NOTE**: Clickable chips are focusable by default so you don't need to
+   * specify this prop.
+   *
+   * This has the same effect as HTML [tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex).
+   */
+  readonly tabIndex?: number;
 
   /**
    * Callback when the chip itself gets clicked.
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -1,4 +1,4 @@
-import { KeyboardEvent, MouseEvent, ReactElement, ReactNode } from "react";
+import { KeyboardEvent, MouseEvent, ReactElement } from "react";
 import { ChipButtonProps } from "./InternalChipButton";
 import { AvatarProps } from "../Avatar";
 import { IconProps } from "../Icon";
@@ -53,9 +53,4 @@ export interface InternalChipProps {
    * Callback for keyboard interaction with chips (e.g., chip deletion).
    */
   onKeyDown?(event: KeyboardEvent<HTMLDivElement | HTMLButtonElement>): void;
-
-  /**
-   * Adds an element or component between the label and suffix
-   */
-  readonly children?: ReactNode;
 }
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -43,6 +43,11 @@ export interface InternalChipProps {
   readonly suffix?: ReactElement<IconProps | ChipButtonProps>;
 
   /**
+   * The label for the interactive chip element for voice over
+   */
+  readonly ariaLabel?: string;
+
+  /**
    * Allows the chip to be focused.
    *
    * **NOTE**: Clickable chips are focusable by default so you don't need to
--- a/packages/components/src/Chips/ChipTypes.tsx
+++ b/packages/components/src/Chips/ChipTypes.tsx
@@ -15,7 +15,7 @@ export interface InternalChipProps {
   readonly active?: boolean;
 
   /**
-   * Makes the chip look and feels uninteractable.
+   * Makes the chip look and feel uninteractable.
    */
   readonly disabled?: boolean;
 
@@ -49,7 +49,13 @@ export interface InternalChipProps {
     event: MouseEvent<HTMLDivElement | HTMLInputElement | HTMLButtonElement>,
   ): void;
 
+  /**
+   * Callback for keyboard interaction with chips (e.g., chip deletion).
+   */
   onKeyDown?(event: KeyboardEvent<HTMLDivElement | HTMLButtonElement>): void;
 
+  /**
+   * If the chip needs children (e.g., empty span for screen reading).
+   */
   readonly children?: ReactNode;
 }
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
@@ -42,7 +42,7 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
           }
         >
           <span className={styles.visuallyHidden}>
-            Press delete to remove chip
+            Press delete or backspace to remove chip
           </span>
         </InternalChip>
       ))}
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
@@ -16,7 +16,6 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
     handleChipKeyDown,
     handleChipRemove,
     handleCustomAdd,
-    handleEmptyBackspace,
     handleWrapperKeyDown,
   } = useInternalChipDismissible(props);
 
@@ -52,7 +51,6 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
         options={availableChipOptions}
         onOptionSelect={handleChipAdd}
         onCustomOptionSelect={handleCustomAdd}
-        onEmptyBackspace={handleEmptyBackspace}
         onSearch={props.onSearch}
         onLoadMore={props.onLoadMore}
       />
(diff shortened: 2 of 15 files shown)


packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -21,23 +20,55 @@ export function useInternalChipDismissible({
...
+
     handleEmptyBackspace: () => {
-      actions.handleChipRemove(sortedVisibleChipOptions[selected.length - 1])();
+      actions.handleChipRemove(selected[selected.length - 1])();
+    },
+
+    handleWrapperKeyDown: (event: KeyboardEvent<HTMLDivElement>) => {
+      const target = event.target;
+      if (!(target instanceof HTMLElement)) return;
+      if (target instanceof HTMLInputElement && target.value) return;
```
timoteialbu, line 44: why are we doing this?

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -21,23 +20,55 @@ export function useInternalChipDismissible({
...
+        event.key === "ArrowRight" &&
+        target.nextElementSibling instanceof HTMLElement
+      ) {
+        target.nextElementSibling.focus();
+      }
+    },
+
+    handleChipKeyDown: (value: string) => {
+      return (event: KeyboardEvent<HTMLDivElement>) => {
+        if (event.key === "Backspace") {
```
timoteialbu, line 61: I think we should have `Delete` here too cause most mac keyboards don't have Backspace, just Delete

chris-at-jobber:
This is really close! There is an interaction missing on the delete side. 

From our AC:

Given that a user’s cursor is placed following a Chip
- pressing the delete key will set focus to the preceding Chip
- pressing the delete key once the Chip is focused will delete the Chip
- deleting the Chip is still announced to screen-readers

What's actually happening:

Given that a user’s cursor is placed following a Chip
- pressing the delete key will set focus and immediately delete the Chip
- deleting the Chip is still announced to screen-readers

It's a seemingly small extra step but it gives a bit of a buffer preventing the user from instantly removing the Chip.

An example is in [Angular chips.](https://material.angularjs.org/latest/demo/chips)

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -21,23 +20,55 @@ export function useInternalChipDismissible({
...
+          const target = event.target;
+          if (
+            target instanceof HTMLElement &&
+            target.nextElementSibling instanceof HTMLElement
```
timoteialbu, line 65: why are we doing nextElementSibling?
I think this is a good place to do some documentation through code.
Like set a variable
```
const elementWeNeedToFocusCauseBla = target.nextElementSibling

```

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -21,23 +20,55 @@ export function useInternalChipDismissible({
...
+      actions.handleChipRemove(selected[selected.length - 1])();
+    },
+
+    handleWrapperKeyDown: (event: KeyboardEvent<HTMLDivElement>) => {
+      const target = event.target;
+      if (!(target instanceof HTMLElement)) return;
+      if (target instanceof HTMLInputElement && target.value) return;
+      if (
+        event.key === "ArrowLeft" &&
+        target.previousElementSibling instanceof HTMLElement
```
timoteialbu, line 47: same comments for this and next if as https://github.com/GetJobber/atlantis/pull/872/files#r835424851

packages/components/src/Chips/ChipTypes.tsx
```diff
@@ -48,4 +48,8 @@ export interface InternalChipProps {
   onClick?(
     event: MouseEvent<HTMLDivElement | HTMLInputElement | HTMLButtonElement>,
   ): void;
+
+  onKeyDown?(event: KeyboardEvent<HTMLDivElement | HTMLButtonElement>): void;
```
timoteialbu, line 52: should we use XOR?

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
adambobadam: Totally a good question.  Keeping me on my toes! 😬 

The first one is that there is that there is an error when we try to access previous/next element sibling when we don't do the checks on the first highlighted line.

The second highlighted line there is to prevent using the left/right arrow keys when you start typing when there is an input.  I am assuming that Darryl and Chris worked out that functionality in a previous review and that was planned.

adambobadam: Resolving, but feel free to unresolve if this answer doesn't make you happy!!!

timoteialbu: For the second line, can we also make that into a self explanatory variable or add a comment? First preferred, but it not possible add comment.

For the first one, I guess my confusion is around why we need this escape hatch. I don't mean this from a TS perspective, which I get. Its more that if we didn't have a HTMLElement to begin with, then we would not attach this method to it in the first place.

adambobadam: @timoteialbu So for the first check, since the event is a a keyboard event on an html element you're thinking we don't really need that check to begin with?

Would `const target = event.target as HTMLElement;` and then removing the check make sense?

timoteialbu: I don't like forcing a cast like that just because this will hide future states where that type will change. I found something that we can do like this:

```
handleWrapperKeyDown: (
      event: KeyboardEvent<HTMLDivElement> & {
        target: HTMLDivElement;
      },
    ) => {
```

adambobadam: Damn okay! That makes sense!!

chris-at-jobber (commented):
Left review [here](https://github.com/GetJobber/atlantis/pull/872#issuecomment-1079187315)

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
```diff
@@ -3,33 +3,48 @@ import styles from "./InternalChipDismissible.css";
...
+            <InternalChipButton
+              icon="remove"
+              {...chip}
+              onClick={handleChipRemove(chip.value)}
+            />
+          }
+        >
+          <span className={styles.visuallyHidden}>
+            Press delete to remove chip
+          </span>
```
darryltec, line 46: This actually gets read separately on voice over. Basically it'll go
- 🔈  Mando Press delete to remove this chip. Group.
- _User press right arrow key_
- 🔈 Press delete to remove this chip
- _User press right arrow key_
- 🔈 End of group
- _User press right arrow key_
- _Finally goes to the next chip_

I think instead of passing in a children, we should pass in a computed `aria-label` where it's the chip label + "press delete" thing. That way we don't need this extra element and it's styling and it's just a string that gets added as an html attribute.

darryltec (changes requested):
QA: 

- ❌ Tabbing on multiselect chips requires 2 tabs now instead of 1 (compared with what's in [master](https://atlantis.getjobber.com/components/chips))
- ✔️ Pressing delete on a chip deletes it
- ❌ Deleting a chip doesn't get announced on voice over anymore
- ❌ Pressing delete on text input with no value focuses on the chip (if any) to the right of it.
- ✔️ Left and right arrow switcher between chips
- ❌ Pressing left and right with voice over doesn't immidiately go to the next chip

packages/components/src/Chips/ChipTypes.tsx
```diff
@@ -48,4 +49,16 @@ export interface InternalChipProps {
   onClick?(
     event: MouseEvent<HTMLDivElement | HTMLInputElement | HTMLButtonElement>,
   ): void;
+
+  /**
+   * Callback for keyboard interaction with chips (e.g., chip deletion).
+   */
+  onKeyDown?(
+    event: KeyboardEvent<XOR<HTMLDivElement, HTMLButtonElement>>,
+  ): void;
```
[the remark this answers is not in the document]
adambobadam: I think XOR is technically more correct here?  Just because for it only makes sense to have an event of one or the other.  I am on the fence though, because it seems like with that `ts-xor` library you can only do two xors.  

And I think
```
XOR<XOR<HTMLDivElement, HTMLInputElement>, HTMLButtonElement>
```
Seems overly complicated, so I think just doing the `OR` makes sense.

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
```diff
@@ -3,33 +3,48 @@ import styles from "./InternalChipDismissible.css";
...
           key={chip.value}
           {...chip}
+          onKeyDown={handleChipKeyDown(chip.value)}
           onClick={handleChipClick(chip.value)}
-          onRequestRemove={handleChipRemove(chip)}
-        />
+          suffix={
+            <InternalChipButton
+              icon="remove"
+              {...chip}
```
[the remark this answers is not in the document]
adambobadam: I don't think we fully need to do a spread, since there is just label, invalid, and disabled that aren't hard-coded, and the rest of the props passed in don't apply.  

Do you think spreading is still okay?  I would say it's okay, since there are enough props there to warrant it, but up to you!

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
```diff
@@ -3,33 +3,48 @@ import styles from "./InternalChipDismissible.css";
...
+              onClick={handleChipRemove(chip.value)}
+            />
+          }
+        >
+          <span className={styles.visuallyHidden}>
+            Press delete to remove chip
```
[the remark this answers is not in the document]
chris-at-jobber: Discussed with Adam and I don't have an answer beyond changing it to
- “Press backspace on Windows or delete on a Mac to remove Chip” which seems extremely annoying
- do a UserAgent check or some very specific @supports CSS query and change the label based on the response (Like check if os is mac or windows, then change the wording accordingly)

I’m asking in an accessibility-focused channel in the DS slack because I assume we're not the first people to have this problem, but Google did not help me.

(I would prefer to not hang up this PR on this question if it's the only outstanding issue because I think we have some decisions to make)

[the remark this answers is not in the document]
chris-at-jobber: (I’m also not sure the current state of UA-sniffing in browser APIs 😅 )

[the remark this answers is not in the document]
chris-at-jobber: I did get some feedback from a fairly well-known accessibility pro that saying something general like

"Press delete or backspace to remove {label}" would be a reasonable approach

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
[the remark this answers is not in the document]
darryltec: the goal is to make it predictable, spreading chip there makes it unpredictable when it starts supporting the same prop name but with a different implementation/expectation

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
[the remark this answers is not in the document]
darryltec: I like that. Some classic keep it simple!

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -21,23 +20,65 @@ export function useInternalChipDismissible({
...
     },
+
     handleEmptyBackspace: () => {
-      actions.handleChipRemove(sortedVisibleChipOptions[selected.length - 1])();
+      actions.handleChipRemove(selected[selected.length - 1])();
+    },
+
+    handleWrapperKeyDown: (
+      event: KeyboardEvent<HTMLDivElement> & {
+        target: HTMLDivElement;
```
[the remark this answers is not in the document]
adambobadam: Getting currentTarget and target confused as per our convo!  However, there should always be that add button rendered, so I am thinking that since it's always going to be a child element in the wrapper the target should always be there.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -21,23 +20,65 @@ export function useInternalChipDismissible({
...
+      if (
+        event.key === "ArrowRight" &&
+        nextElementToFocus instanceof HTMLElement
+      ) {
+        nextElementToFocus.focus();
+      }
+    },
+
+    handleChipKeyDown: (value: string) => {
+      return (event: KeyboardEvent<HTMLDivElement>) => {
```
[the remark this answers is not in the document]
adambobadam: Nope, could be a button too!

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
adambobadam: Updated here: https://github.com/GetJobber/atlantis/pull/872/commits/b2db210eb745cc816719f2794edb13b7322ac895, hope that works :)

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
```diff
@@ -111,7 +111,16 @@ export function useInternalChipDismissibleInput({
...
+        callbacks.Backspace = () => {
+          // If there's no text left to delete,
+          // and delete is pressed again, focus on a chip instead.
+          const target = computed.inputRef.current?.previousElementSibling;
+          if (target instanceof HTMLElement) {
+            target.focus();
+          } else {
+            onEmptyBackspace();
+          }
+        };
```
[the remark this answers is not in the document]
adambobadam: Hmm, you're right!  I think `onEmptyBackspace()` can be removed, because right now we put the focus on the chip now when there is an empty input.  Would just having the logic in the `handleKeyDown` function you've highlighted here make sense?

I think it would because the functionality of trying to delete something is related to the input right?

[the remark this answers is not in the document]
darryltec: Cool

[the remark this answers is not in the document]
adambobadam: This commit removes everything related to that functionality: https://github.com/GetJobber/atlantis/pull/872/commits/913c20e21c6e79cc8821ff0a5eda00329d812072

darryltec (changes requested):
Looking pretty good! Just a few things to think about

There's a pretty major screen reader issue where it'll always say "group" because we added tabindex on a non-interactable element (div). No biggie tho. need to just declare a role in there and a bunch more aria bits.

- Wrapper needs a `role` of `listbox` 
- Chips needs a `role` of `option`

Combination of those 2 gets rid of the `group` thing and make the screen readers read them properly

Also, while we're at it...
- Chips need to have `aria-setsize` and `aria-posinset` set to them 
  - So the non-seeing users know how many chips are there and at which chip are they focusing on

All-in-all, we're really close!!

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -21,23 +21,66 @@ export function useInternalChipDismissible({
...
+
+    handleChipKeyDown: (value: string) => {
+      return (event: KeyboardEvent<HTMLElement>) => {
+        if (event.key === "Backspace" || event.key === "Delete") {
+          const target = event.target;
+          if (
+            target instanceof HTMLElement &&
+            target.nextElementSibling instanceof HTMLElement
+          ) {
+            target.nextElementSibling.focus();
```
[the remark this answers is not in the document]
adambobadam: Sounds good!  @chris-at-jobber I think we talked about this one.  When you delete the last chip it goes to the add button.

chris-at-jobber:
Seeing some very inconsistent interactions on typing-to-add-new Chips. Looks like it could be related to how fast the list is filtering, but it's hit or miss. You can see both successful and unsuccessful attempts to type-to-add in this video:

https://user-images.githubusercontent.com/39704901/161647513-466912fa-3e35-4e3b-a931-4c9e02345043.mov

adambobadam:
> Seeing some very inconsistent interactions on typing-to-add-new Chips. Looks like it could be related to how fast the list is filtering, but it's hit or miss. You can see both successful and unsuccessful attempts to type-to-add in this video:

I wonder if this is having to do with the API call that the playground/documentation is making when using cloudflare.  It seems to work fine locally

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -98,6 +98,7 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
             role="listbox"
             id={menuId}
             className={styles.menuList}
+            data-testid="menu"
```
darryltec, line 101: Name this `chip-menu` because the chip test isn't the only test it's going to show up, It's also gonna show up on every test that implements `<Chips />`. Imagine adding a testid that is super generic that tests in jobber online uses them. You'd break all those test 😆 .

image

darryltec (changes requested):
One last thing. Once we confirm that @chris-at-jobber was just a blip, we can approve this. Code looks good to me.

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
adambobadam: hahaha good point, that makes sense!

darryltec (approved):
We're good. The issue that @chris-at-jobber found exists in master so we can follow up on that after this is merged

adambobadam:
Thanks @chris-at-jobber and @darryltec for the reviews and help!

PR #677, opened 2021-09-22
chore: Audit Atlantis packages

eddysims:
## Motivations

Atlantis packages are getting pretty far out of date. This PR will do an audit of vulnerabilities for `most` of the Atlantis npm packages.

There are some packages that need to be audited outside of this PR as they have a much much larger impact.

**Notes**

Updating the `typed-css-modules` caused a change to the way that css types are made. This causes changes in every `.css.d.ts`.

As part of this I have also ignored the `.css.d.ts` files in the `eslintignore` file as these are auto generated and we shouldn't care about linting them.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- pull down branch
- run all the `npm` scripts
- see they all work

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #880, opened 2022-04-14
fix(design): Add back missing icon colors types

darryltec:
## Motivations

For some reason, rollup doesn't want to bring over the `.css.d.ts` file even tho we're referencing it. Could have something to do with the way those `typed-css-module` exports the type with the old-style imports `export = styles`.

Anyhoo, this makes rollup copy those files in the correct place. It now autocompletes again.

image

Prop table colors now shows the correct type

image

### Added

- Added a plugin called `rollup-plugin-copy` to copy the missing `.css.d.ts` files over to `dist` so it gets exported

### Changed

- Make rollup use the `rollup-plugin-copy`
- Update icon sizing type to use the existing type file rather than hard coding it
- Updated `rollup-plugin-typescript2` to latest since it has issues/bug with async methods 
  - https://github.com/ezolenko/rollup-plugin-typescript2/issues/105

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Install the pre-release on the other repo and see that the colors now gets suggestions

```
npm i @jobber/design@0.24.1-pre.17+288721e0
```

---

Random photo of Atlantis

### DIFF
--- a/packages/design/package-lock.json
+++ b/packages/design/package-lock.json
@@ -20,11 +20,57 @@
 				"glob-to-regexp": "^0.3.0"
 			}
 		},
+		"@nodelib/fs.scandir": {
+			"version": "2.1.5",
+			"resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.5.tgz",
+			"integrity": "sha512-vq24Bq3ym5HEQm2NKCr3yXDwjc7vTsEThRDnkp2DK9p1uqLR+DHurm/NOTo0KG7HYHU7eppKZj3MyqYuMBf62g==",
+			"dev": true,
+			"requires": {
+				"@nodelib/fs.stat": "2.0.5",
+				"run-parallel": "^1.1.9"
+			},
+			"dependencies": {
+				"@nodelib/fs.stat": {
+					"version": "2.0.5",
+					"resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-2.0.5.tgz",
+					"integrity": "sha512-RkhPPp2zrqDAQA/2jNhnztcPAlv64XdhIp7a7454A5ovI7Bukxgt7MX7udwAu3zg1DcpPU0rz3VV1SeaqvY4+A==",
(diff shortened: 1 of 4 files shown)


timoteialbu:
This fixes Icon issues which is the AC of the story, but it would be nice to find the root cause of why rollup doesn't do this correctly, because if we do, then we can solve the other issues (for example `Content` has a prop `spacing` that also does not auto complete).

darryltec:
Oh I think I know why. There was a pretty big change with the typed-css-modules format. https://github.com/GetJobber/atlantis/pull/677/files

darryltec:
@timoteialbu thanks for that thought. I have an idea on how to fix this. it's gonna be a hella big PR but it's gonna fix it!! 80% sure on that one.

darryltec:
@timoteialbu that one is actually a lot harder to deal with as that issue goes back into v0 of atlantis.

This issue we're fixing is related but slightly different since it used to work. Why does it work? The history of the icon before the `keyof typeof` was it hardcoded all the colors.

I added another ticket to look at the bigger issue, but i think this should just go out as it's affecting atlantis, jobber online, and jobber mobile

timoteialbu:
> @timoteialbu that one is actually a lot harder to deal with as that issue goes back into v0 of atlantis.
> 
> This issue we're fixing is related but slightly different since it used to work. Why does it work? The history of the icon before the `keyof typeof` was it hardcoded all the colors.
> 
> I added another ticket to look at the bigger issue, but i think this should just go out as it's affecting atlantis, jobber online, and jobber mobile

If you made another ticket to look at it that sounds good to me!
I do wonder, would it be the worst to quickly add this `copy` stuff to all the other ones that are broken?

darryltec:
@timoteialbu thought about it. it would basically add extra files that we don't need on the release. We'd have to figure out which one gets referenced as a `keyof typeof` and then copy it over.

timoteialbu:
As discussed, we will have another ticket to try and fix the root cause so that we can fix the other type issues in the app. Good work on this!

PR #881, opened 2022-04-19
feat(components): only wrap Banner children in Text if they are a simple string. 

timoteialbu:
## Motivations

Wrapping all of the children of `Banner` in `Text` has unintended consequences when I want to build my banner with more than just text (ie button and other sections)

## Changes

Conditionally wrap the `children` of `Banner` in `Text` only if they are a `string`.

### Changed

- If you used the `Banner` in the past with more than just a simple string, this will affect how your component renders now.

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Banner/Banner.test.tsx
+++ b/packages/components/src/Banner/Banner.test.tsx
@@ -92,7 +92,23 @@ it("renders a banner with a primary 'learning' action when the type is 'notice'"
   expect(tree).toMatchSnapshot();
 });
 
-test("it should call the onClick when primaryAction is present", () => {
+it("does not wrap the its children in text if the children are not a simple string", () => {
+  const tree = renderer
+    .create(
+      <Banner
+        type="notice"
+        primaryAction={{
+          label: "smash me",
+        }}
+      >
+        <h3>Bruce</h3>
+      </Banner>,
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
+});
+
+it("it should call the onClick when primaryAction is present", () => {
   const onClick = jest.fn();
 
   const { getByText } = render(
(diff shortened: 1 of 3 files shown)


packages/components/src/Banner/Banner.test.tsx
```diff
@@ -92,7 +92,23 @@ it("renders a banner with a primary 'learning' action when the type is 'notice'"
...
+        type="notice"
+        primaryAction={{
+          label: "smash me",
+        }}
+      >
+        <h3>Bruce</h3>
+      </Banner>,
+    )
+    .toJSON();
+  expect(tree).toMatchSnapshot();
```
darryltec, line 108: a snapshot test? how dare you 😛

packages/components/src/Banner/Banner.tsx
```diff
@@ -83,7 +83,7 @@ export function Banner({
           role={type === "error" ? "alert" : "status"}
         >
           <div className={contentClassNames}>
-            <Text>{children}</Text>
+            <BannerChildren>{children}</BannerChildren>
```
darryltec, line 86: Nice touch!

packages/components/src/Banner/Banner.test.tsx
darryltec: You can probably do a find on `Bruce` and check if the parent is an instance of `HTMLDivElement`. If it gets wrapped in a `<Text />` then that's an `HTMLParagraphElement`

darryltec (approved):
Looks good to me. It would be nice if the snapshot test could be replaced with a more specific test but that shouldn't block this PR

packages/components/src/Banner/Banner.test.tsx
timoteialbu: I know, I figured your team will take this own and you can clean this garbage 😛 
But in the spirit of giving a shit I will change this!

darryltec (approved):
Release it!

PR #883, opened 2022-05-02
fix(components): set radio labels to base fontSize

chris-at-jobber:
## Motivations

Radio labels were inheriting font-size from the document body, making them larger than other input elements and `p` elements.

image

### Added

- Explicit font-size declaration on radio labels

### Changed

- Radio labels will no longer rely on their parent site's `body` to determine font-size

### Check RadioGroup on production
image
image

### Check locally/on [preview build](https://af270339.atlantis.pages.dev/)
image
image

- [ ] Radio labels are `base` (equivalent to 16px) on local/preview builds

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/RadioGroup/RadioGroup.css
+++ b/packages/components/src/RadioGroup/RadioGroup.css
@@ -11,8 +11,9 @@
 
 .label {
   display: inline-flex;
-  align-items: center;
+  font-size: var(--typography--fontSize-base);
   cursor: pointer;
+  align-items: center;
 }
 
 .input + .label::before {


tyler-santos:
@chris-at-jobber This looks good to me, just wondering if you have any concern of how this will affect people already using this and possibly having the look of this change unexpectedly? Also we don't have the ability to change the size of the labels. Is that something we might consider in the future? Thanks!

jackson-zhipeng-chang:
tested, the new font size is smaller than other elements' font size. LGTM

jackson-zhipeng-chang (approved):

chris-at-jobber:
@tyler-santos Good questions! Given that we use `base` throughout the product, I don't foresee any issues. This is mainly to reduce the visual inconsistencies seen in the [in-product example](https://github.com/GetJobber/atlantis/pull/883#issue-1223355817) where a very similar element like Checkbox, which often appears in the same context, has a different font-size. I generally think we should make components fairly inflexible in terms of visual treatment, so we can ensure consistency across the board more easily (and have to make fewer cleanup PRs like this one 😅 )

PR #888, opened 2022-05-05
chore: prevent lerna from checking our npm token as this fails when we have 2FA enabled, which we want to do for all accounts in our org [47168]

timoteialbu:
## Motivations
This will allow us to turn on account wide npm 2FA requirement.

## Changes
Disable lerna checking the npm token as per https://github.com/lerna/lerna/issues/2788#issuecomment-1062659948

Random photo of Atlantis

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "release-the-kraken": "npm ci && lerna publish",
     "release:unpublished-package": "npm ci && lerna publish from-package",
     "prerelease": "lerna publish --canary --dist-tag next --preid pre",
-    "ci:release": "lerna publish --yes",
+    "ci:release": "lerna publish --yes --no-verify-access",
     "version": "node ensureChangelogHeaders.js",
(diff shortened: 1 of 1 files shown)


darryltec:
What's this for?

.circleci/config.yml
```diff
@@ -162,7 +162,7 @@ jobs:
             if git log --max-count=1 --pretty=%s | grep -q "chore(release): Publish"; then
               echo "Skipping run on release commit."
             else
-              npm run ci:release
+              npm run ci:release -- --no-verify-access
```
darryltec, line 165: thoughts on putting that no verify access as a part of `ci:release` by default? I believe this is the only place we use that script on

darryltec (commented):
can we confirm that this wouldn't mess with the releases? It's been broken for over a week and we have recently fixed it again yesterday

timoteialbu:
Doesn't work, still asks for a one time password locally :(

PR #889, opened 2022-05-13
JOB-40406 Modify FormField Altantis Component For Thumbnail Functionality

dmcrorieGIT:
## Motivations

We needed the ability to view format files as thumbnails

## Changes

Added new props to allow format file to be displayed as a thumbnail

## Decision Justification

- we didn't add spacing to the component to allow for devs to set their own spacing

## Testing

In playground with a format file, set display to "compact".

## References

https://www.figma.com/file/b7LY6hVMwFOj5maidxPTMB/Design-System-Contribution-%5B-Thumbnail-%2B-Gallery-%5D?node-id=1420%3A19642

---

Random photo of Atlantis


packages/components/src/FormatFile/FormatFile.css
```diff
@@ -69,10 +89,49 @@
...
+
+.centered {
+  display: flex;
+  position: relative;
+  height: 100%;
+  margin: 0 var(--space-small);
+  align-items: center;
+}
+
+.overlay {
```
fagnanm, line 131: Reference comment above about this overlay conflicting with the original technique. Same overlay can be used for the progress bar and the delete action.

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -69,10 +89,49 @@
...
 .actionBlock {
   padding: 0 var(--file-card--base-padding) 0 0;
 }
+
+.thumbnail {
+  --message-weight: 400;
+  display: flex;
+  display: inline-block;
+  position: relative;
+  height: 50px;
```
fagnanm, line 106: Does this do something? Height/Width seem to be set in React below.

fagnanm (changes requested):
A few elements of the design where missed, noted here in the Figma file.

Screen Shot 2022-05-17 at 3 08 31 PM

Also because there are some minor differences between formatFile and what is in the thumbnail proposal, I'd like to use the proposal as the source of truth. Key differences:
- Background for the thumbnail/formatFile becomes `--color-surface--background`
- All file icons become the same greyBlue colour

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -37,8 +36,6 @@
 
 .imageBlock::before {
   content: "";
-  display: block;
```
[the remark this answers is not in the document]
dmcrorieGIT: I checked current atlantis, it doesn't look like the overlay is being added there either; here is two of the same image, one with progress bar and one without:

Screen Shot 2022-05-19 at 11 03 29 AM

It doesn't look like the image with the progress bar has an overlay 🤔

taylorvnoj:
Hey @fagnanm - @dmcrorieGIT and I have addressed the comments/changes above. Whenever you have a moment, if you wouldn't mind having another glance that would be greatly appreciated! Thank you!

packages/components/src/FormatFile/FormatFile.css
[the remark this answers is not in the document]
fagnanm: Likely is broken and we should fix

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -51,12 +51,61 @@
   margin: 0 auto;
 }
 
-.progress {
+.fileContentWrapper {
```
fagnanm, line 54: This is new to the component, it looks like in the original `FormatFile--imageBlock` handled both the image or the file styling. This allowed for simpler styling and markup as there didn't need to be a fileContentWrapper inside the image block. In the update this fileContentWarpper only applies to the thumbnail variant and not the original.

Screen Shot 2022-05-25 at 10 35 52 AM

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -51,12 +51,61 @@
...
+  display: flex;
+  width: inherit;
+  height: inherit;
+  align-items: center;
```
fagnanm, line 58: If you are using something like `margin: 0 auto;` on the icon then properties like align-items, justify-content won't do anything. I can't find where this is being set in the file, but it is there in Chrome inspector. The reason why the margin property isn't working to center is because this block has `flex-direction: column`, this means (vertical flexibility) so combined with `0 auto` won't center vertically because the `0` needs to be `auto` as well.

To my comment above, I believe the previous structure worked, but we are having to do this styling just for the thumbnail variant.

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -51,12 +51,61 @@
...
+  justify-content: center;
+  flex-direction: column;
+}
+
+.thumbailIconWrapper {
```
fagnanm, line 63: Pretty sure this isn't being used

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -51,12 +51,61 @@
...
+
+.thumbnailFilename {
   display: flex;
   position: absolute;
-  top: 0;
+  bottom: var(--space-small);
+  left: var(--space-small);
+}
+
+.thumbnailFilenameSmall {
```
fagnanm, line 74: This text size needs to be 10px instead of 12. 14 in the large thumbnail, 10 in the small.

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -51,12 +51,61 @@
...
+}
+
+.thumbnailNonImage::before {
+  content: "";
+  display: block;
+  position: absolute;
   right: 0;
   bottom: 0;
+  z-index: 1;
+  width: 20px;
```
fagnanm, line 88: This width and height should be 16px or `--space-base`

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -51,12 +51,61 @@
...
+  height: 20px;
+  background: linear-gradient(
+    270deg,
+    rgb(244, 244, 244) 38.46%,
+    rgba(244, 244, 244, 0) 100%
+  );
+}
+
+.thumbnail .fileContentWrapper {
```
fagnanm, line 97: If you move this `overflow` up to the FormatFile wrapper then you can use the previous structure for FormatFile--imageBlock for the image and the icons and won't need to absolute position another container that just clips the text.

fagnanm (commented):
Screen Shot 2022-05-25 at 10 23 25 AM

Couple of design things missed:
1. Thumbnails need `border-radius: --radius-base & border: --border-base`
2. When in progress upload there needs to be an overlay behind the progress bar using `--color-overlay`

ChristianLee-Jobber (commented):
I feel out of my depth looking at the CSS stuff, but the react code seems fine.

This might be a weird question: at what point does this justify having its own component? We seem to be checking the `display` type and serving 2 different components based on that value.

packages/components/src/FormatFile/FormatFile.css
taylorvnoj: We had initially only found in FontSizes.css that it went to `smaller` but I've done further digging to get the font to 10px using `font-size: calc(var(--base-unit) * 0.625);` 👍

packages/components/src/FormatFile/FormatFile.css
taylorvnoj: Adjusting this results in the below:
Screen Shot 2022-05-27 at 11 33 00 AM

packages/components/src/FormatFile/FormatFile.css
taylorvnoj: quick note on this - we've had to conditionally apply this to account for the difference in width & height for the large thumbnail. 
We have matched the figma to be 16px/`--space-base` on the default thumbnail however with larger font-size and increased margin in the larger thumbnail I needed to apply `--space-large` to cover the file name completely, otherwise we see the below:

Screen Shot 2022-05-27 at 11 54 42 AM

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -6,6 +6,7 @@
   --file-card--border-color: var(--color-border);
   --file-card--accent-color: var(--color-grey);
   --file-card--base-padding: var(--space-small);
+  --border-basebackground: var(--color-surface--background);
```
chris-at-jobber, line 9: Why are we defining a custom property here that we're only using in one place, as a background color? I'm only seeing this used on line 32.

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -26,37 +27,100 @@
   height: 100%;
   box-sizing: border-box;
   padding: 0;
-  background-color: var(--file-card--header-background-color);
-  background-position: center;
+  border: var(--border-base);
+  border-radius: var(--radius-base);
+  background-color: var(--border-basebackground);
```
chris-at-jobber, line 32:
```suggestion
  background-color: var(--color-surface--background);
```

packages/components/src/FormatFile/FormatFile.css
chris-at-jobber:
```suggestion
```

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -69,10 +133,24 @@
...
+  display: flex;
+  position: relative;
+}
+
+.deleteButton {
+  display: flex;
+  position: absolute;
+  top: var(--space-smaller);
+  right: var(--space-smaller);
+  z-index: 1;
```
chris-at-jobber, line 155: Seeing a lot of z-index usage in this component, is it possible to avoid it by ordering the DOM elements in a natural stacking order? It looks like this isn't even needed when I remove the style:
image

packages/components/src/FormatFile/FormatFile.mdx
```diff
@@ -35,9 +35,10 @@ import { FormatFile } from "@jobber/components/FormatFile";
         name: "myballisbigandroundIamrollingitontheground.png",
         type: "image/png",
         size: 213402324,
-        progress: 0.3,
+        progress: 0.4,
         src: () => Promise.resolve("https://source.unsplash.com/250x250"),
       }}
+      display="thumbnail"
       onDelete={() => alert("🧨BALEETED!💥")}
```
chris-at-jobber, line 42: I know this is pre-existing but can we change all of these examples to say "Deleted" instead of "🧨BALEETED!💥" ?

packages/components/src/FormatFile/FormatFile.mdx
```diff
@@ -61,11 +68,45 @@ following:
...
   or 2 files per row).
 
-## Props
+### Display
 
-<Props of={FormatFile} />
+Format files can be displayed as either a file row or a thumbnail.
 
----
+### Thumbnail
```
chris-at-jobber, line 75: Should bump this heading down a notch as it's nested under the "Display" section. We should also have a heading and explanation for when to use the "file row" display type.

packages/components/src/FormatFile/FormatFileDeleteButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+          type="tertiary"
+          icon="trash"
+          ariaLabel="Delete Thumbnail"
+          size={buttonSize}
+        />
+      </div>
+      <ConfirmationModal
+        title="Confirm Deletion"
+        message={`Are you sure you want to delete this thumbnail?`}
+        confirmLabel="Yes"
```
chris-at-jobber, line 37:
```suggestion
        confirmLabel="Delete"
```

packages/components/src/FormatFile/FormatFileDeleteButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+          variation="destructive"
+          type="tertiary"
+          icon="trash"
+          ariaLabel="Delete Thumbnail"
+          size={buttonSize}
+        />
+      </div>
+      <ConfirmationModal
+        title="Confirm Deletion"
+        message={`Are you sure you want to delete this thumbnail?`}
```
chris-at-jobber, line 36:
```suggestion
        message={`Are you sure you want to delete this file?`}
```

They're not just deleting the thumbnail but removing the whole file from the system so the error message should reflect this

packages/components/src/FormatFile/FormatFileDeleteButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <Button
+          onClick={() => setDeleteConfirmationOpen(true)}
+          variation="destructive"
+          type="tertiary"
+          icon="trash"
+          ariaLabel="Delete Thumbnail"
+          size={buttonSize}
+        />
+      </div>
+      <ConfirmationModal
```
chris-at-jobber, line 34: This ConfirmationModal should be of variation `destructive` [(reference)](https://1617698a.atlantis.pages.dev/components/confirmation-modal#destructive)

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -1,29 +1,69 @@
...
    * File upload details object. (See FileUpload type.)
    */
   readonly file: FileUpload;
 
+  /**
+   * To display as either a file row or thumbnail
+   *
+   * @default "file"
+   */
+  readonly display?: "file" | "thumbnail";
```
chris-at-jobber, line 34: I'd like to get some feedback from the design team on whether the "file" name feels natural.

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -26,37 +27,100 @@
...
+
+.thumbnailFilenameSmall {
+  display: flex;
+  position: absolute;
+  bottom: var(--space-smaller);
+  left: var(--space-smaller);
+  font-size: calc(var(--base-unit) * 0.625);
+}
+
+.thumbnailNonImagedefault::before {
```
chris-at-jobber, line 80: Is it possible to resolve some of this duplicated styling like...

```
.thumbnailNonImage::before {
  content: "";
  display: block;
  position: absolute;
  right: 0;
  bottom: 0;
  z-index: 1;
  width: var(--space-base);
  height: var(--space-base);
  background: linear-gradient(
    270deg,
    rgb(244, 244, 244) 38.46%,
    rgba(244, 244, 244, 0) 100%
  );
}

.thumbnailNonImage.thumbnailLarge::before {
  width: var(--space-large);
  height: var(--space-large);
}
```

Off the top of my head I can't recall if you can stack selectors for pseudo elements, but if we can that would be ideal so we're not writing the same CSS twice with two lines difference.

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -32,43 +72,103 @@ export function FormatFile({ file, onDelete }: FormatFileProps) {
...
+      {display === "file" && (
+        <div className={styles.formatFile}>
+          <div
+            className={imageBlockStyle}
+            style={style}
+            data-testid="imageBlock"
+          >
+            {!imageSource && (
+              <div className={styles.icon}>
+                <Icon name={iconName} color="greyBlue" />
```
MichaelParadis, line 90: Reading the description of the PR it says 

> we left icon colours for non image file types to be original colours, instead of changing to grey like in the thumbnail mockups

However it looks like you set the colour of the icons in both of the code paths. Can you update the description to remove this line?

chris-at-jobber (changes requested):
Some CSS and design changes - one of the Atlantis engineers will be doing a more code-focused review as well but there are some small fixes/opportunities I noticed.

packages/components/src/FormatFile/FormatFile.mdx
```diff
@@ -61,11 +68,45 @@ following:
...
+
+Used to display a cropped smaller version of the file or image that has been
+uploaded by the SP
```
MichaelParadis, line 78: Atlantis is not specific to Jobber so the SP should be changed

packages/components/src/FormatFile/FormatFile.css
MichaelParadis: I agree with @chris-at-jobber. If it is needed I would like to see it prefixed with `--file-card...`

packages/components/src/FormatFile/FormatFile.mdx
```diff
@@ -61,11 +68,45 @@ following:
 - FormatFile components should take up the full width of the parent container (1
   or 2 files per row).
 
-## Props
+### Display
 
-<Props of={FormatFile} />
+Format files can be displayed as either a file row or a thumbnail.
```
MichaelParadis, line 73: I almost think there should be an example of the file row like we have for thumbnail.

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -32,43 +72,103 @@ export function FormatFile({ file, onDelete }: FormatFileProps) {
...
+    : classNames(styles.imageBlock, styles.imageBlockOverlay);
 
   return (
-    <div className={styles.formatFile}>
-      <div className={styles.imageBlock} style={style} data-testid="imageBlock">
-        {!imageSource && (
-          <div className={styles.icon}>
-            <Icon name={iconName} />
+    <>
+      {display === "file" && (
```
MichaelParadis, line 81: This conditional seems to do a lot going on. I wonder if we can break these conditionals off into two internal components. One to handle the file preview and the other to handle the thumbnail

MichaelParadis: After looking at the element structure my gut tells me the conditionals can be change to just pass different classNames, style and testIds that could be determined before and set as `consts`

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -1,29 +1,69 @@
...
+   * The base dimensions of the thumbnail
+   *
+   * @default "default"
+   */
+  readonly displaySize?: keyof typeof sizeToDimensions;
+
+  /**
+   * Function to execute when format file is clicked
+   */
+  onClick?(event: React.MouseEvent<HTMLDivElement>): void;
```
MichaelParadis, line 46: onClick seems to only do something when `display="thumbnail` I think it should do something when it is a file row too. @chris-at-jobber thoughts?

MichaelParadis (changes requested):
Great job, looks good but I have a few comments and changes to make
 * Simplify the component, it seems to have a lot of duplication between the thumbnail and file preview. If not break them into their own internal components to make it a bit more readable
 * onClick doesn't seem to work on `display="file"`
 * The onClick seems to also be called if it is provided with an onDelete and you click the delete button. This could lead to a bad user experience if they click delete and are shown a preview and the delete modal for example.
 
A11y
* The delete button is focusable, however the thumbnail isn't. This means a screen reader can't open a preview for example.

packages/components/src/FormatFile/FormatFile.tsx
chris-at-jobber: That makes sense, I'd expect the same behaviour from both variations

packages/components/src/FormatFile/FormatFile.tsx
chris-at-jobber: Based on design team feedback, can we change `file` to `expanded` and thumbnail to `compact` ?

chris-at-jobber:
Just noticed something while referencing this against the Figma component..

1. the overlay to "mask" the filename is visible above the `loading` overlay
image

2. There's a uniform 8px (`space-small`) padding around the progress bar and filename. The Figma specifies 8, but is actually inset by 12 (11+1 for border). @fagnanm are you good with `small` all around? 
image

taylorvnoj:
Thank you @chris-at-jobber  & @MichaelParadis for looking things over. 

Major changes we've addressed:
1. Removed top level conditionals so that `expanded` and `compact` display share the same structure
2. Solve issue where clicking delete also triggers onClick; both `expanded` and `compact` both have onClick functionality
3. FormatFile is now hoverable and focusable
4. Solved filename gradient mask being visible in `loading` state

To address the padding around progress bar and filename, it is our understanding that it was strictly 8px on both sides however we can easily modify this if you have an alternative suggestion for spacing.

Thanks again for looking through our changes.  Would you mind taking another peek whenever you have a chance? 😄

chris-at-jobber (changes requested):
### A few hover/focus-related issues...

I can't seem to focus the FormatFile by clicking, only by using the `tab` key

We don't need a hover state unless/until there's an onClick or onDelete on the Thumbnail - it's there to indicate that the element is interactive
image

Also don't need a hover state on an in-progress as I don't think you'd be able to either preview or delete a mid-upload file? (If I'm off on that, let me know 😄)
image

When I hover an in-progress file, the masking overlay is visible overtop:
image
_(due to earlier note about in-progress not needing an overlay, this may not be an issue, but it seems like it could be revealing an underlying structural issue)_

Compact version with an `onDelete` should only show the onDelete when hovered or focused
image

I noticed that the thumbnail in the `expanded` display retains its' rounded corners, also on hover the overlay is not rounded in `compact` display:
image
image
image

### Props

We favour using `base` over `default` as a sizing prop name ([reference](https://atlantis.getjobber.com/guides/frontend-style#base-vs-default-or-medium-for-setting-the-default-option-in-a-range))

Nothing happens when `displaySize` is set to `large` on the `expanded` format. This isn't a use case we anticipate needing, but I'm wondering if this is being intentionally restricted.

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -32,46 +75,132 @@ export function FormatFile({ file, onDelete }: FormatFileProps) {
...
+          <Typography element="p" size="small" textColor="greyBlueDark">
+            {fileSize}
+          </Typography>
         </div>
       )}
+      {isComplete && onDelete && showDeleteButton && (
+        <>
+          <FormatFileDeleteButton
+            deleteButtonStyle={
+              IsFileDisplay(display) ? styles.actionBlock : styles.deleteButton
```
MichaelParadis, line 150: I think actionBlock and deleteButton might need to be renamed. The reason is that the Delete Button is not using the deleteButton style when it is a file. I think maybe it could be fileDeleteButton and imageDeleteButton maybe

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -1,29 +1,72 @@
...
   const isComplete = file.progress >= 1;
 
+  const thumbnailDimensions = sizeToDimensions[displaySize];
+  const [deleteConfirmationOpen, setDeleteConfirmationOpen] = useState(false);
+  const [showDeleteButton, setShowDeleteButton] = useState(
+    display === "compact" && displaySize === "default" ? false : true,
+  );
+
   const iconName = getIconNameFromType(file.type);
   const fileSize = getHumanReadableFileSize(file.size);
```
MichaelParadis, line 71: I think this should be named something to describe what it does better. style is too generic in my opinion

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -32,46 +75,132 @@ export function FormatFile({ file, onDelete }: FormatFileProps) {
...
   const style = imageSource ? { backgroundImage: `url(${imageSource})` } : {};
+  const imageBlockStyle = isComplete
+    ? styles.imageBlock
+    : classNames(styles.imageBlock, styles.imageBlockOverlay);
 
   return (
-    <div className={styles.formatFile}>
-      <div className={styles.imageBlock} style={style} data-testid="imageBlock">
+    <div
+      onMouseEnter={() => {
```
MichaelParadis, line 84: By using mouse events it is no longer possible to get the delete button to render using keyboard focus

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -32,46 +75,132 @@ export function FormatFile({ file, onDelete }: FormatFileProps) {
...
+        showDeleteButtonIfSmallThumbnail(
+          display === "compact" && displaySize === "default",
```
MichaelParadis, line 86: Maybe move this to a const 
`const isSmallThumbnail =  display === "compact" && displaySize === "default"`
 and reuse it

packages/components/src/FormatFile/FormatFile.tsx
MichaelParadis: I think onFocus and onBlur might solve this

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -32,46 +75,132 @@ export function FormatFile({ file, onDelete }: FormatFileProps) {
...
+function showDeleteButtonIfSmallThumbnail(
+  smallAndCompact: boolean,
+  setShowDeleteButton: React.Dispatch<React.SetStateAction<boolean>>,
+) {
+  if (smallAndCompact) {
+    setShowDeleteButton(true);
+  }
+}
+
+function IsFileDisplay(display: "expanded" | "compact") {
```
MichaelParadis, line 181: Should be `isFileDisplay`

MichaelParadis (changes requested):
Overall, it is good.
Main issue is some minor cleanup and A11y not working 100% with keyboard events.

packages/components/src/FormatFile/FormatFile.tsx
dmcrorieGIT: With just using onBlur and onFocus, we lose the delete button appearing on hover, which was the original intention. We will probably need both onFocus/onBlur as well as onMouseOver to get full functionality.

hover-doesnt-work

taylorvnoj:
Hello @chris-at-jobber & @MichaelParadis - thank you for providing another review!  

@dmcrorieGIT and I have addressed the suggested changes and are hoping you'd be able to take another look whenever you are able to 😄

packages/components/src/FormatFile/sizeToDimensions.ts
```diff
@@ -0,0 +1,12 @@
+export const sizeToDimensions = {
+  default: {
```
chris-at-jobber, line 2: Just wondering why this file is needed instead of doing the sizing with the CSS file and applying classnames? It looks like we're applying a `<style>` object instead of a class for this case only.

dmcrorieGIT: @chris-at-jobber Good callout! Switched to use classnames instead 😎

chris-at-jobber (changes requested):
When I move my hover from the image to the delete button, the parent loses hover-state and it's quite jarring. Also, can we add a transition to the hover overlay so that it doesn't snap in and out, but eases similar to our Button? (Button uses `transition: all var(--timing-base);` for reference).

https://user-images.githubusercontent.com/39704901/171951636-adeba304-cd37-47b9-b9a8-4371a4cf9386.mov

From the Figma, can we retain the overlay on focus as well? Currently it appears to only appear on hover.
image

For `expanded` displays that have an onDelete but not an onClick, can we remove the hover and focus states on the thumbnail? Currently it gives the impression that I could interact with this image.
image

For the focus styling itself...
- can we use the [focus shadow](https://atlantis.getjobber.com/colors#focus) instead of setting the outline value?
- can we set the focus to the entire `expanded` element, instead of only the thumbnail?

I've set up an overview sheet of all of the states for both compact and expanded displays with no events/onClick/onDelete/onClick + onDelete, if there are any questions let me know!

https://www.figma.com/file/b7LY6hVMwFOj5maidxPTMB/Design-System-Contribution-%5B-Thumbnail-%2B-Gallery-%5D?node-id=1869%3A18247

dmcrorieGIT:
> When I move my hover from the image to the delete button, the parent loses hover-state and it's quite jarring

@chris-at-jobber There might not be a way to get rid of that sadly. In order to get the other delete button functionality, we had to move the delete button outside of where the hover state is applied. If we move it back in, that will re-introduce those bugs. If we really want the hover state applied with hovering the delete button as well, we could do it but the code would be ugly as hell.

chris-at-jobber:
@dmcrorieGIT that example is of a progress-state thumbnail, which shouldn't have a focus state, and only has the overlay because progress has an overlay (not because of the focus state)

dmcrorieGIT:
> can we set the focus to the entire expanded element, instead of only the thumbnail?

@chris-at-jobber wouldn't this imply that they can click anywhere on the expanded row and that would trigger an onClick?

chris-at-jobber:
@dmcrorieGIT would they not be able to click anywhere on the component to trigger the onClick? If that's the case then as-is is fine, and we can update the behaviour later. I would have expected a click anywhere on the component to fire the onClick.

dmcrorieGIT:
@chris-at-jobber Right now we have expanded set so that they onClick is only applied when they click the thumbnail. This due to the way everything is structured: if we have it set to the whole element, we will re-introduce a bug where clicking the delete button will also trigger the onClick. We tried to find ways around this, but we were unable.

I agree that we should leave this as-is for now, and somebody can make a modification in a follow-up PR. So as for the focus, we will keep it set to only the thumbnail in the expanded variant.

taylorvnoj:
Hey @chris-at-jobber - we have made the following changes:
- persist hover state for image block when hovering delete
- added transition animation to hover overlay
- applied no hover/focus states for `expanded`
- applied `--shadow-focus` instead of outline value
- retained overlay on focus

Thank again for continuing to look this over! This is now ready for another review 😄

chris-at-jobber:
Few more things...

Focus state on the thumbnail in `expanded` format is getting clipped by the container:
image

With my keyboard I should be able to...
- fire the onClick event by pressing return/space (like a button)
  - in this video I am pressing `return` with it focused and nothing happens
- move focus from the thumbnail to the `delete` button by pressing the `tab` key
  - in this video, when I press the `tab` key, focus disappears, after a subsequent press focus moves to the next focusable element, skipping the delete button
https://user-images.githubusercontent.com/39704901/172258612-ed54cadd-5860-4050-8cb4-3f8d5b1007b8.mov

chris-at-jobber (changes requested):
Left comment with some changes!

chris-at-jobber:
@dmcrorieGIT these are pretty fundamental to the experience, we can't consider this a completed component without the ability to perform keyboard interactions. If you are committing to follow them up immediately in a separate PR I'm OK with it, but I wouldn't consider the work completed if the component doesn't work.

dmcrorieGIT:
> If you are committing to follow them up immediately in a separate PR I'm OK with it

@chris-at-jobber This sounds like a plan!

chris-at-jobber (commented):
Design approved on the condition that the [keyboard interaction bugs](https://jobber.atlassian.net/browse/JOB-48877) identified in the last review are immediately followed up on in a separate PR

dmcrorieGIT:
Closing this PR, here is the new PR --> https://github.com/GetJobber/atlantis/pull/898

PR #891, opened 2022-05-17
feat(components): Atlantis Component For Gallery

dmcrorieGIT:
## Motivations

I added a new component `Gallery` to atlantis, which displays a collection of thumbnails to the viewer

## Changes

* New component added
* Modified FormatFile so that it has a tab index, so that it can be focusable
* Modified FormatFile css so that it has a new class `thumbnailParent`, to get descendant combinator interactive selectors to work (e.g., hover)

### Added

- `Gallery` component

### Changed

- `FormatFile`:
    - Now tab focusable
    - New css classes allowing for focus and hover

## Testing

- Run atlantis on this PR locally
- Go to the Gallery component. Look at the test cases and/or even play around with the Playground to test your own changes
- Ensure that everything matches the figma mockup for Gallery --> https://www.figma.com/file/b7LY6hVMwFOj5maidxPTMB/Design-System-Contribution-%5B-Thumbnail-%2B-Gallery-%5D?node-id=1420%3A19639

## References

Figma Mockup --> https://www.figma.com/file/b7LY6hVMwFOj5maidxPTMB/Design-System-Contribution-%5B-Thumbnail-%2B-Gallery-%5D?node-id=1420%3A19639

---

Random photo of Atlantis


packages/components/src/Gallery/Gallery.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { useState } from "react";
+import styles from "./Gallery.css";
+import { GalleryProps } from "./GalleryTypes";
+import { LightBox } from "../LightBox";
+import { FormatFile } from "../FormatFile";
+import { Button } from "../Button";
+
+export function Gallery({ files, size = "default", max }: GalleryProps) {
```
chris-at-jobber, line 8: Similar to the FormatFile, can we change the `default` size to `base`?

packages/components/src/Gallery/Gallery.tsx
```diff
@@ -0,0 +1,74 @@
...
+                onClick={() => {
+                  handleThumbnailClicked(index);
+                }}
+                onDelete={file.onDelete}
+              />
+            </div>
+          );
+        })}
+        {max && files.length > max && !displayPastMax && (
+          <Button
```
chris-at-jobber, line 41: If we wrap a div around this Button with a width of {thumbnailSize (56 / 168)} and `display: flex`, and set the Button property to `fullWidth`, we'll get the square proportion that matches the design:

image
image

chris-at-jobber: We could also set the button to be `large` when the Gallery is large to make the "+N" label more prominent

chris-at-jobber (changes requested):
From the design side, just some changes to the size prop name to match the other components in the system and a tweak to how we handle the "max" button.

I did notice this issue and thought it's maybe worth raising:

- when I select a thumbnail from the "large" Gallery example with a PDF ([here](https://fe18aa9b.atlantis.pages.dev/components/gallery#sizes) in the preview build), the Gallery vanishes and there's a console error thrown:

https://user-images.githubusercontent.com/39704901/172463439-05af25b7-c48d-4847-8a99-d91582fc5545.mov

I'm not sure if this is an issue with the component, or possibly just the playground implementation, but I'm consistently able to reproduce it.

packages/components/src/Gallery/Gallery.tsx
```diff
@@ -0,0 +1,74 @@
...
+          return (
+            <div key={file.key}>
+              <FormatFile
+                file={{ ...file, src: () => Promise.resolve(file.src) }}
+                display="compact"
+                displaySize={size}
+                onClick={() => {
+                  handleThumbnailClicked(index);
+                }}
+                onDelete={file.onDelete}
```
ChristianLee-Jobber, line 35:
```suggestion
                onDelete={handleDelete(file)}
```
Sorry about the miscommunication earlier. Would it be possible for the `Gallery` to accept a function called `onDelete`. Then, for each individual `file` that was passed in, it will trigger that function on the file? So the code would look like what I suggested, but the `Gallery` would then require another input `prop`, the `handleDelete` method

packages/components/src/Gallery/Gallery.tsx
linds14sr20det: Hey @chris-at-jobber, I was just about to address this comment, changing this to `base` results in a lot of cascading changes. I can't seem to see where you suggested the changes in FormatFile, it still looks to use `default`

chris-at-jobber: Oh, it was left as a requested change [here](https://github.com/GetJobber/atlantis/pull/889#pullrequestreview-994232683) - I had thought it was addressed. We should try and get that fixed here to avoid introducing breaking changes to Jobber.

linds14sr20det: Ah I see. Okay we can do that rename here definitely, just was confused by the comment.

chris-at-jobber: Thanks!

packages/components/src/Gallery/Gallery.mdx
```diff
@@ -0,0 +1,242 @@
...
+        progress: 1,
+        src: "https://source.unsplash.com/250x250",
+      },
+    ]}
+  />
+</Playground>
+
+## Content Guidelines
+
+The Gallery was created to present images, specifically thumbnails, to the user.
```
paddle-sky, line 217: other gallery words being used are small g rather then large G

paddle-sky (approved):
Thanks for changing the Gs

chris-at-jobber (commented):
Design approved!

dmcrorieGIT:
@darryltec @MichaelParadis This is the component that builds of the modifications to FormatFile, could you review when you have a chance?

packages/components/src/Gallery/Gallery.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const [lightboxIndex, setLightboxIndex] = useState(0);
+  const [displayPastMax, setDisplayPastMax] = useState(false);
+
+  return (
+    <>
+      <div className={size === "large" ? styles.galleryLarge : styles.gallery}>
+        {files.map((file, index) => {
+          if (max && !displayPastMax && index + 1 > max) {
+            return <></>;
+          }
```
darryltec, line 21: This is gonna thrown an error on react since it doesn't have a key. TBH, not a fan of this since we're doing every bit of logic in JSX and that's a code smell.

What I would suggest is to add a const that determines whether to map all the files or only the first `n` files.

_Par exemple_
```tsx
const visibleFiles = displayPastMax ? files : files.slice(0, max);
```

This way you won't have to do the `if (max && !displayPastMax && index + 1 > max) {` logic since it won't render what's not there and fixes the key error that react throws.

Tho you would have to adjust the default value for `displayPastMax`.

packages/components/src/Gallery/GalleryTypes.d.ts
```diff
@@ -0,0 +1,70 @@
...
+   * The progress of a file upload between 0 and 1.
+   * - `0` represents Upload just started.
+   * - `1` represents a complete upload.
+   */
+  readonly progress: number;
+
+  /**
+   * The data url of the file.
+   */
+  src: string;
```
darryltec, line 65: `readonly`?

Also, curious as to why this is not a `Promise` while the FormatFile where it's being sent to is a `Promise`

packages/components/src/Gallery/GalleryTypes.d.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+interface GalleryItem extends File {
+  icon?: FileIconsNames;
```
darryltec, line 69: nit:

```suggestion
  readonly icon?: FileIconsNames;
```

packages/components/src/Gallery/GalleryTypes.d.ts
```diff
@@ -0,0 +1,70 @@
+import { IconNames } from "../Icon";
```
darryltec, line 1: also, mostly curious about this. why is this a `.d.ts` file? 🤔

packages/components/src/Gallery/GalleryTypes.d.ts
```diff
@@ -0,0 +1,70 @@
...
+export type Sizes = "small" | "base" | "large" | "extraLarge";
+
+export type FileIconsNames = Extract<
+  IconNames,
+  "file" | "excel" | "pdf" | "video" | "word"
+>;
+
+export interface GalleryProps {
+  /**
+   * The size of the Gallery and it's files
```
darryltec, line 12: Is it more?

> The size of the files and their spacing in the gallery

packages/components/src/Gallery/GalleryTypes.d.ts
```diff
@@ -0,0 +1,70 @@
...
+  files: File[];
+
+  /**
+   * The max number of thumbnails before no more thumbnails are displayed
+   * unless the user clicks an action to display the rest
+   */
+  max?: number;
+
+  /**
+   * onDelete callback - this function will be called when the delete action is triggered on a Gallery image
```
darryltec, line 29: nit: ah! you were doing so good on the previous doc string!! 😆 

```suggestion
   * onDelete callback - this function will be called when the delete action is 
   * triggered on a Gallery image
```

packages/components/src/Gallery/GalleryTypes.d.ts
```diff
@@ -0,0 +1,70 @@
...
+   */
+  onDelete?(file: File): void;
+}
+
+export interface File {
```
darryltec, line 34: Is this almost the same type as the `FileUpload`?

packages/components/src/Gallery/Gallery.css
```diff
@@ -0,0 +1,21 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: var(--space-small);
+}
+
+.galleryLarge {
+  display: flex;
+  flex-wrap: wrap;
+  gap: var(--space-base);
+}
```
darryltec, line 11: nit:

```suggestion
.gallery,
.galleryLarge {
  display: flex;
  flex-wrap: wrap;
  gap: var(--space-small);
}

.galleryLarge {
  gap: var(--space-base);
}
```

darryltec (changes requested):
This looks really good! Just a few notes on the code to make it real nice. 

A few questions for you tho
- Say you have 2 files.
  - The 1st file has finished uploading, 2nd is still uploading
  - You click the first file to preview it and it opens the Lightbox
  - What should happen when you press the right arrow key?
- Should the show button have a border? It looks odd on large image
- Also, should it have a height? Because this will happen image

A few jest testing ask
- it should call the onDelete with the file in question
- it should open the file in a new tab if it's not an image
- it should close the Lightbox if you trigger a close (since this is still handled on implementation rather than the component itself)
- what happens when `generateImagesArray` pushes undefined in the `filesToImageIndex`?

packages/components/src/Gallery/GalleryTypes.d.ts
dmcrorieGIT: We don't want this being a promise, since we need to access the string directly for the Lightbox to work; I originally had this as a promise, and I tried to have the LIghtbox "dig" for the string, but I wasn't able to get that to work.

packages/components/src/Gallery/GalleryTypes.d.ts
dmcrorieGIT: I'm not sure, this is just what the component generator created 🤷‍♂️

packages/components/src/Gallery/GalleryTypes.d.ts
dmcrorieGIT: Weird, I'm surprised this wasn't automatically fixed by the linter 🤔

packages/components/src/Gallery/GalleryTypes.d.ts
dmcrorieGIT: Very similar yes, but `src` is a raw string instead of a promise. I tried using type extension, but I couldn't get that to work.

packages/components/src/Gallery/GalleryTypes.d.ts
darryltec: Oh did it? Shiiii nvm then. I forget what Atlantis does now 😆

packages/components/src/Gallery/Gallery.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+it("renders a Gallery", () => {
+  const tree = renderer.create(<Gallery files={files} />);
+  expect(tree).toMatchSnapshot();
+});
+
+describe("when the Gallery is large", () => {
+  it("displays large thumbnails", () => {
+    const tree = renderer.create(<Gallery size="large" files={files} />);
+    expect(tree).toMatchSnapshot();
```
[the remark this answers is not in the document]
dmcrorieGIT: True, I'll check for classnames instead 😄

packages/components/src/Gallery/GalleryTypes.d.ts
darryltec: Hmm interesting. My only concern here is that it would mean we'll have some inconsistency
- `InputFile` returns a `Promise` `src` on any file you pick
- `FormatFile` expects that the file you give it has an `src` that is a `Promise`.
- `Gallery` which uses format file (which is tied to input file) is now not using a `src` `Promise` 

🤔 Not sure if that's okay or if we really need a promise in all of the file bits in atlantis 🤔

packages/components/src/Gallery/GalleryTypes.d.ts
darryltec: you can try this

```
interface File extends Pick<FileUpload, "all" | "the" | "common" | "props"> {
  readonly src: string;
}
```

if we decide that it's supposed to be string and not a promise

packages/components/src/Gallery/Gallery.mdx
```diff
@@ -0,0 +1,242 @@
...
+The Gallery can display as many images as the designs call for, but if the
+amount of images become too great, the `max` prop can be used to constrain the
```
darryltec, line 219:
```suggestion
amount of images becomes too great, the `max` prop can be used to constrain the
```

packages/components/src/Gallery/Gallery.mdx
```diff
@@ -0,0 +1,242 @@
...
+number of images shown.
+
+## Accessibility
+
+Each of the thumbnails the Gallery displays are tab navigatable. A hover state
+will also slighltly dim a particular thumbnail when the cursor hovers over it.
```
darryltec, line 225:
```suggestion
Each of the thumbnails the Gallery displays are tab navigable. A hover state
will also slightly dim a particular thumbnail when the cursor hovers over it.
```

packages/components/src/Gallery/Gallery.mdx
```diff
@@ -0,0 +1,242 @@
...
+
+## Responsiveness
+
+Responsiveness of thumbnails are controlled by the Gallery wrapper, thumbnails
+do not get resized. In other words, if the device rendering the Gallery is 500px
+wide, then taking into account padding it will display roughly three thumbnails
+beside eachother before wrapping subsequent thumbnails to a new line
```
darryltec, line 232:
```suggestion
beside each other before wrapping subsequent thumbnails to a new line
```

darryltec (commented):
few spelling bits

dmcrorieGIT:
@darryltec Thanks for the thorough review! I made some changes as per your feedback, and I added new test cases as per your suggestions.

Your questions about the design of this component (e.g., should the show more button have a border?) would be better answered by @chris-at-jobber, since he designed the component.

> what happens when `generateImagesArray` pushes `undefined` in the filesToImageIndex?

If `undefined` is pushed for a particular index, that means that when the corresponding "image" is not an image at all, but some other file type, so we don't want to try and open the lightbox with that file since it will fail. This is what the undefined check is doing on line 69; this could be technically redundant since non-images are screened on the method that calls `handleLightboxOpen`, but this needed to be added since the typescript linter will complain about the value possibly being undefined. We don't necessarily need to push `undefined` per se, but we need to fill it with something, otherwise the functionality of translating a file index to an image index will fail 😅 

All that being said, this is ready for another review when you have a moment! 😄

packages/components/src/Gallery/GalleryTypes.d.ts
dmcrorieGIT: Oh yeah, inheritance is much better! Made the change 😎

packages/components/src/Gallery/GalleryTypes.d.ts
dmcrorieGIT: I see what you're saying, consistency would be nice here. However, I'm not sure we can get a promise to work with the Lightbox, this might be a trade-off we have to make to get that functionality here unfortunately 😅

packages/components/src/Gallery/Gallery.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+      type: "application/pdf",
+      size: 233411234,
+      progress: 1,
+      src: pdfSrc,
+    };
+    const { findByTestId } = render(<Gallery files={[pdfFile, ...files]} />);
+
+    const internalPDFImage = await findByTestId("pdf");
+
+    fireEvent.click(internalPDFImage.parentElement.parentElement.parentElement);
```
darryltec, line 165: hehe, can actually just do

```suggestion
    fireEvent.click(internalPDFImage);
```

since that event bubbles up

also throws linting errors 
image

darryltec (changes requested):
One last critical thing as the types are resolving to `any` because it's not resolving the types. I think if we rename that to `.ts` without the `.d` on it, it'll get resolved

chris-at-jobber:
Re: the design questions

- If we can conditionally have a border on the "+N" button when large, and no border when small, that works for me
- "+N" button shouldn't collapse on a subsequent row
- 2 files, 1 still uploading, you press the right arrow key, scenario, the first few options that come to mind:   
  - (if this is possible) show the still-uploading image thumbnail without the progress bar/overlay - this feels like the most optimistic approach, and since we already have a rendered image that is being displayed behind the overlay/progressbar in the FormatFile, I would assume we can access it here?
  - do nothing (as the file doesn't exist to be previewed)
  - go to the "position" of the still-uploading image with an indicator that it's not uploaded yet, show image once uploaded

packages/components/src/Gallery/Gallery.test.tsx
dmcrorieGIT: Good call! I'll make the change!

darryltec:
Cloudflare is failing but the build docs is passing. I think we are safe to merge this 
image

darryltec:
I'm good to merge this now and @dmcrorieGIT will create the follow-up task for the design updates

chris-at-jobber:
@darryltec yeah nothing would change in terms of developer usage, so that makes sense

PR #895, opened 2022-06-01
feat(components): JOB-40406 Allow PUT requests in InputFile component

ChristianLee-Jobber:
## Motivations
Allowing the component to handle PUT requests makes it more flexible. ActiveStorage creates pre-signed URLs that are to be used with PUT requests, so it is required for our work.

## Changes
The `InputFile` component now allows PUT requests to be performed.

## Testing
In your local Atlantis playground, you can go to the `InputFile` page and modify it so that its `getUploadParams` prop/ method returns the following in a promise:
```Javascript
{
  url: "https://address_where_you_can_put.com",
  fields: {
    "Content-Type": "image/jpeg",
    "OTHER-APPROPRIATE-HEADERS": "appropriate_header"
  },
  httpMethod: "PUT" 
}
```

example:
 
Screen Shot 2022-06-01 at 16 10 21

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -207,12 +207,16 @@ export function InputFile({
     const fileUpload = getFileUpload(file, key);
     onUploadStart && onUploadStart({ ...fileUpload });
 
-    const uploadProgressUpdate = (progressEvent: any) => {
+    const handleUploadProgress = (progressEvent: any) => {
       onUploadProgress &&
         onUploadProgress({
           ...fileUpload,
           progress: progressEvent.loaded / progressEvent.total,
         });
+    };
+
+    const handleUploadComplete = () => {
+      onUploadComplete?.({ ...fileUpload, progress: 1 });
     };
 
     if (httpMethod === "POST") {
@@ -225,19 +229,19 @@ export function InputFile({
       axios
         .post(url, formData, {
           headers: { "X-Requested-With": "XMLHttpRequest" },
-          onUploadProgress: uploadProgressUpdate,
+          onUploadProgress: handleUploadProgress,
         })
         .then(() => {
-          onUploadComplete && onUploadComplete({ ...fileUpload, progress: 1 });
+          handleUploadComplete();
         });
     } else if (httpMethod === "PUT") {
       axios
         .put(url, file, {
           headers: fields,
-          onUploadProgress: uploadProgressUpdate,
+          onUploadProgress: handleUploadProgress,
         })
         .then(() => {
-          onUploadComplete && onUploadComplete({ ...fileUpload, progress: 1 });
+          handleUploadComplete();
         });
     }
   }
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -55,10 +55,17 @@ export interface UploadParams {
   readonly key?: string;
 
   /**
-   * Any extra fields to send with the upload POST.
+   * Any extra fields or headers to send with the upload.
    * If unspecified only the file will be included.
    */
   readonly fields?: { [field: string]: string };
+
+  /**
+   * The HTTP method which we wish to use for the upload
+   * If unspecified, it will default to POST
+   * The `fields` will then be used as headers for the request
+   */
+  readonly httpMethod?: "POST" | "PUT";
 }
 
 interface InputFileProps {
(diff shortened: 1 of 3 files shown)


packages/components/src/InputFile/InputFile.tsx
```diff
@@ -190,30 +197,49 @@ export function InputFile({
...
+      formData.append("file", file);
+
+      axios
+        .post(url, formData, {
+          headers: { "X-Requested-With": "XMLHttpRequest" },
+          onUploadProgress: uploadProgressUpdate,
+        })
+        .then(() => {
+          onUploadComplete && onUploadComplete({ ...fileUpload, progress: 1 });
+        });
```
darryltec, line 232: you can put this in a `const` so you won't have to repeat it

```ts
const handleUploadComplete = () => {
  onUploadComplete?.({ ...fileUpload, progress: 1 });
};

...
  .then(handleUploadComplete)
```

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -190,30 +197,49 @@ export function InputFile({
...
-          onUploadProgress &&
-          onUploadProgress({
-            ...fileUpload,
-            progress: progressEvent.loaded / progressEvent.total,
-          }),
-      })
-      .then(() => {
-        onUploadComplete && onUploadComplete({ ...fileUpload, progress: 1 });
-      });
+    const uploadProgressUpdate = (progressEvent: any) => {
```
darryltec, line 210: nit: we try to name those callbacks passed in `onCallback` props with `handleCallback` that way, it's easier to connect the dots just by looking at it.

```suggestion
    const handleUploadProgress = (progressEvent: any) => {
```

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -190,30 +197,49 @@ export function InputFile({
   }
 
   async function uploadFile(file: File) {
```
darryltec, line 199: Any ideas on how to simplify this function? Or like self-documenting. It's switching between 2 code paths now and it's getting a lil hard to read.

darryltec (changes requested):
Just a few thoughts

packages/components/src/InputFile/InputFile.tsx
MichaelParadis: You may be able to use the `axios(config)` or `axios.request(config)` api to simplify the branching code paths slightly

ChristianLee-Jobber: I think you still get branching logic like that. Because if it's a POST request, we have `formdata` to pass in. If it's a PUT request, we need to pass in `headers` and the file is the payload. We just move the branching logic to how we build out our `config` options for the `axios.request` method. But perhaps that can be moved into a separate function? I'll continue looking into this

ChristianLee-Jobber: I removed the branching logic from the `uploadFile` method, and created a new method called `createAxiosConfig` in order to generate the correct input for the `axios.request` method

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -122,6 +129,18 @@ interface InputFileProps {
...
 
+interface RequestParams extends Omit<UploadParams, "key"> {
+  /**
+   * The file being uploaded
+   */
+  file: File;
+
+  /**
+   *
+   */
```
darryltec, line 140: Did you mean to add a documentation here? also if we can explain why that's set as `any` that would be great! We try to avoid adding `any` as a type in Atlantis and if we have no choice, we explain why.

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -55,10 +55,17 @@ export interface UploadParams {
...
+   * Any extra fields or headers to send with the upload.
    * If unspecified only the file will be included.
    */
   readonly fields?: { [field: string]: string };
+
+  /**
+   * The HTTP method which we wish to use for the upload
+   * If unspecified, it will default to POST
+   * The `fields` will then be used as headers for the request
+   */
```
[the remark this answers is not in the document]
ChristianLee-Jobber: Yes, I agree. The wording is wrong. Thank you for that. I have fixed it up in the latest commit

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -190,31 +211,72 @@ export function InputFile({
...
+    const axiosConfig = createAxiosConfig({
+      url,
+      httpMethod,
+      fields,
+      file,
+      handleUploadProgress,
+    });
+    axios.request(axiosConfig).then(() => {
+      handleUploadComplete();
+    });
```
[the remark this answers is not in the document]
ChristianLee-Jobber: I did not know this, thanks for pointing it out! I have updated the code to match your suggestion

darryltec:
Great job @ChristianLee-Jobber !

PR #908, opened 2022-07-26
chore(deps): Update react markdown to latest

darryltec:
## Motivations

Update react markdown to latest so we can render it on storybook without it throwing 

```
Uncaught TypeError: Object(...).cwd is not a function
```

### Added

- 

### Changed

- Updated react markdown to latest

### Deprecated

- 

### Removed

- 

### Fixed

- Storybook failing to render it

### Security

- 

## Testing

- Ensure it still works like before

## Troubleshooting

Running jest would throw this error

```  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/en/ecmascript-modules for how to enable it.
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /atlantis/packages/components/node_modules/react-markdown/index.js:6
    export {uriTransformer} from './lib/uri-transformer.js'
    ^^^^^^

    SyntaxError: Unexpected token 'export'

    > 1 | import ReactMarkdown from "react-markdown";
        | ^
      2 | import React, { DetailedHTMLProps, HTMLAttributes } from "react";
      3 | import { Text } from "../Text";
      4 | import { Emphasis } from "../Emphasis";

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1350:14)
      at Object.<anonymous> (packages/components/src/Markdown/Markdown.tsx:1:1)
```

Here's an explanation on why that happens https://github.com/remarkjs/react-markdown/issues/635#issuecomment-900671613

and the potential solutions for it https://github.com/remarkjs/react-markdown/issues/635#issuecomment-956158474

---

Random photo of Atlantis

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -79,6 +79,8 @@ module.exports = {
   // A map from regular expressions to module names that allow to stub out resources with a single module
   moduleNameMapper: {
     "\\.css$": "identity-obj-proxy",
+    "react-markdown":
+      "<rootDir>/packages/components/node_modules/react-markdown/react-markdown.min.js",
   },
 
   // An array of regexp pattern strings, matched against all module paths before considered 'visible' to the module loader
--- a/package-lock.json
+++ b/package-lock.json
@@ -2747,7 +2747,7 @@
       "version": "file:packages/components",
       "dev": true,
       "requires": {
-        "@jobber/design": "^0.25.0",
+        "@jobber/design": "^0.25.2",
         "@jobber/formatters": "^0.2.0",
         "@jobber/hooks": "^1.11.0",
         "@popperjs/core": "^2.0.6",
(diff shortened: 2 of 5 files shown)


MichaelParadis (approved):
LGTM!

lmorgan29 (approved):
LGTM 😎

PR #918, opened 2022-09-16
chore(eslint) Removing `no-null` linter rule

gregb-jobber:
## Motivations

There was a lot of discussion around the rule and we eventually decided to take it out. Because we're using tools and libraries that aren't as strict about `null` as this rule it means there's lots of cases where `null` is valid and necessary value to assign to a variable. An example of this is null fields in GraphQL. If a field is nullable and you want to test the null case your test data will necessarily have `null` in it. The rule meant that we were regularly adding exceptions or doing other type acrobatics to make the linter happy.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chips/InternalChipDismissible/hooks/useInView.ts
+++ b/packages/components/src/Chips/InternalChipDismissible/hooks/useInView.ts
@@ -1,7 +1,7 @@
 import { useEffect, useRef, useState } from "react";
 
 export function useInView<T extends Element>() {
-  const ref = useRef<T>(null); // eslint-disable-line no-null/no-null
+  const ref = useRef<T>(null);
   const [isInView, setIsInView] = useState(false);
 
   if (window.IntersectionObserver) {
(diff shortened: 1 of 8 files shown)


gregb-jobber:
I don't understand what all the import errors are about.

MichaelParadis (commented):
Pretty sure this line could also be removed
https://github.dev/GetJobber/atlantis/blob/d4a5d571b02056cb9ce5773daffd854c911d96b8/packages/eslint-config/package.json#L17

MichaelParadis (approved):
LGTM!

gregb-jobber:
@MichaelParadis What needs to happen in the other repos to get this change? Does a new package get built when I merge this? And then I update the version in the other repos?

chris-at-jobber:
@gregb-jobber once this is merged, after ~15min there will be a new `npm` package available. 

You'll see it show up [both in Atlantis docs](https://atlantis.getjobber.com/packages-eslint-config-changelog) and the [npm page](https://www.npmjs.com/package/@jobber/eslint-config).

In any projects using Atlantis, run the install from the npm page to bump the versions.

PR #921, opened 2022-09-30
docs(components): DataTable Component Proposal

demoraesgui:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/proposals/DataTable/README.md
+++ b/docs/proposals/DataTable/README.md
@@ -1,0 +1,66 @@
+# DataTable
+
+The `DataTable` Component would give the users options for pagination, sorting,
+and column visibility by leveraging some sort of API or service.
+
+## Design Patterns
+
+Tables are used to organize tabular information to help users understand and
+quickly scan for information.
+
+Some scenarios for Tables include the following:
+
+- Index Pages
+- Dashboards
+
+### Sorting
+
+Sorting can be enabled on column headers to improve usability.
+
+The default sorted column should be based on contextual based on user needs.
+
+To indicate the column being sorted, use an up (ascending) or down (descending)
+arrow next to the column title. Only 1 column can be sorted at a time. User can
+click on a sorted header to reverse the sorting order. Arrow will rotate 180
+degrees.
+
+Not all columns will need sorting enabled.
+
+# Column Visibility
+
+Not 100% sure if this is something to be implemented at component level but at
+least should be facilitated somehow.
+
+## Interface
+
+```tsx
+<DataTable
+  data={[
+    { name: "Gui", awesome: "yes" },
+    { name: "Harry", awesome: "yes" },
+    { name: "Chris", awesome: "yes" },
+    { name: "Juan", awesome: "yes" },
+  ]}
+  columns={[
+    {
+      accessorKey: "name",
+      cell: info => <span>{info.getValue()}</span>,
+    },
+    {
+      accessorKey: "awesome",
+      cell: info => info.getValue(),
+    },
+  ]}
+/>
+```
+
+![DataTable Example](./example.png)
+
+## Props Table
+
+| name    | type             | default | required | description                                                                                                         |
+| ------- | ---------------- | ------- | -------- | ------------------------------------------------------------------------------------------------------------------- |
+| data    | T[]              |         | yes      | The data to passed into the DataTable                                                                               |
+| columns | ColumnDef\<T\>[] |         | yes      | The Column definition following @tanstack/react-table [ API ](https://tanstack.com/table/v8/docs/guide/column-defs) |
+
+## Accessibility
--- a/docs/proposals/DataTable/README.md
+++ b/docs/proposals/DataTable/README.md
@@ -93,6 +93,21 @@ interface DataTableProps<T> {
    * @memberof DataTable
    */
   sorting?: Sorting;
+
+  /**
+   * Force the height of the table, enables scrolling
+   */
+  height?: number;
+
+  /**
+   * Force the width of the table
+   */
+  width?: number;
+
+  /**
+   * Makes the header stick while scrolling.
+   */
+  stickyHeader?: boolean;
 }
 
 type Pagination = {
@@ -125,6 +140,21 @@ type Pagination = {
    * @type {boolean}
    */
   manualPagination: boolean;
+  /**
+   * The options to control the pagination
+   *
+   * @type {number[]}
+   * @memberof Pagination
+   */
+  itemsPerPage?: number[];
+  /**
+   * When manual pagination is enable the total numbers of
+   * records in the table should be provided.
+   *
+   * @type {number}
+   * @memberof Pagination
+   */
+  totalItems?: number;
 };
 
 type PaginationState = {
--- a/docs/proposals/DataTable/README.md
+++ b/docs/proposals/DataTable/README.md
@@ -31,7 +31,7 @@ Not all columns will need sorting enabled.
 Not 100% sure if this is something to be implemented at component level but at
 least should be facilitated somehow.
 
-## Interface
+## Usage
 
 ```tsx
 <DataTable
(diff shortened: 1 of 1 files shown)


docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
+# DataTable
+
+The `DataTable` Component would give the users options for pagination, sorting,
+and column visibility by leveraging some sort of API or service.
```
MichaelParadis, line 4: Maybe I missed some discussion somewhere but I don't see any mention of filtering. Was that cut from the scope?

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * Enables pagination, mostly follows:
+   * https://tanstack.com/table/v8/docs/api/features/pagination
+   *
+   * @type {Pagination}
+   */
+  pagination?: Pagination;
+
+  /**
+   * Enables pagination, mostly follows:
```
MichaelParadis, line 90: Should be `Enables Sorting`

docs/proposals/DataTable/README.md
lmorgan29: Is filtering a follow up?

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+- Index Pages
+- Dashboards
+
+### Sorting
+
+Sorting can be enabled on column headers to improve usability.
+
+The default sorted column should be based on contextual based on user needs.
+
+To indicate the column being sorted, use an up (ascending) or down (descending)
```
lmorgan29, line 22: are you wording this as an instruction for the engineer to manually control this arrow on their end? Or are you specifying that the component will control the visual of this arrow?

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+arrow next to the column title. Only 1 column can be sorted at a time. User can
```
lmorgan29, line 23: > Only 1 column can be sorted at a time

Is this a package limitation or just a limitation we are imposing here?

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+click on a sorted header to reverse the sorting order. Arrow will rotate 180
+degrees.
+
+Not all columns will need sorting enabled.
+
+# Column Visibility
+
+Not 100% sure if this is something to be implemented at component level but at
```
lmorgan29, line 31: it looks like we can pass this data to the table through state, though depending on the design around toggling visibility of table columns, this might be a separate component that gets this data and passes it to this table 🤔

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+   * Should follow the @tanstack/react-table ColumnDef.
+   * I'm not sure if we need to create an abstraction around this
+   * specific type, ColumnDef by itself is already very flexible and powerfull enough
+   * to enable us to configure the table any way we like.
+   * https://tanstack.com/table/v8/docs/guide/column-defs
+   *
+   * @type {ColumnDef<T>[]}
+   * @memberof DataTable
+   */
+  columns: ColumnDef<T>[];
```
lmorgan29, line 79: so we basically have infinite flexibility here eh? That's awesome. Hopefully the amount of options do not make this over whelming. Part of me likes the idea of abstracting this to make it easier for developers to use but I don't want to limit them. Wondering if we can utilize the Column Helper in a way that makes things easier?
https://tanstack.com/table/v8/docs/guide/column-defs#column-helpers

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * Enables pagination, mostly follows:
+   * https://tanstack.com/table/v8/docs/api/features/pagination
+   *
+   * @type {Pagination}
+   */
+  pagination?: Pagination;
```
lmorgan29, line 87: Without this prop we just display all data I assume?

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * When manually controlled represents the current pagination state of the table.
+   *
+   * @type {PaginationState}
+   */
+  state?: PaginationState;
+
+  /**
+   * When manually controlling pagination, you should supply a total pageCount value to the table if you know it.
+   * If you do not know how many pages there are, you can set this to -1. (Or should we error?)
```
lmorgan29, line 108: So, manual pagination is around our server side pagination? I am thinking we should try to enforce this as a positive number as we should know the count of records and page size right. Or maybe a dumb question, but is this something we can infer within the component based on the amount of data rows passed to the table, in addition to the selected page size?

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+  onPaginationChange?: (newPaginationState: PaginationState) => void;
+
+  /**
+   * Enables manual pagination. If this option is set to true,
+   * the table will not automatically paginate and instead will expect you
+   * to manually paginate the rows before passing them to the table.
+   *
+   * @type {boolean}
+   */
+  manualPagination: boolean;
```
lmorgan29, line 127: if this is true, is `onPaginationChange` a required prop? Feels like it

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+   * @type {SortingState}
+   */
+  state?: SortingState;
+
+  /**
+   * Enables manual sorting for the table.
+   *
+   * @type {boolean}
+   */
+  manualSorting: boolean;
```
lmorgan29, line 148: similar to my question above, does this being True mean that `onSortingChange` would be required? If so, our Typing should to reflect this to reduce engineer mistakes trying to get this to work (onSortingChange not provided when manualSorting is true)

lmorgan29 (commented):
Left a few comments. Would love to see more eyes on this from the engineers who will be using this in the near future

docs/proposals/DataTable/README.md
demoraesgui: From my understanding filtering would be accomplished by leveraging the API, meaning the DataTable would not handle filtering internally, at least not in this iteration. 

any thoughts @jsuazod ?

docs/proposals/DataTable/README.md
demoraesgui: I guess we should implement it on the Atlantis side and show the arrows if sorting is enabled on that specific column. I should update the wording here.

docs/proposals/DataTable/README.md
demoraesgui: It would be a limitation we would be imposing at this moment. The library has the capability to do [multi-sorting](https://tanstack.com/table/v8/docs/api/features/sorting#enablemultisort).

docs/proposals/DataTable/README.md
demoraesgui: From this [example](https://tanstack.com/table/v8/docs/examples/react/column-visibility), it seems that column visibility can be accomplished using the library on the client side, what I'm not 100% sure is how to handle this when the pagination is done manually by leveraging an API.

docs/proposals/DataTable/README.md
jsuazod: That is correct Gui, filtering on the client side could easily be on future iterations, for now, all the filters would need to be on the server and pass into the component data prop.

docs/proposals/DataTable/README.md
jsuazod: I agree with this approach, we will do some testing to make sure this is a working solution.

docs/proposals/DataTable/README.md
jsuazod: Yes, if no prop was passed in we will not show the pagination.

demoraesgui: That's the intention. 

```ts
pagination == undefined // no pagination
pagination != undefined // built-in or manually implemented pagination
```

and `manualPagination` is required.

docs/proposals/DataTable/README.md
demoraesgui: I think we are on the same page on this one @lmorgan29, if there's a valid use case for `manualPagination` without knowing the pageCount I also cannot see it. My feeling is that we should somehow enforce this to be rightfully set.

docs/proposals/DataTable/README.md
demoraesgui: Yes! we are not how to accomplish this if typescript yet though, I did a bit of research with conditional typing but still not able to figure this one out, I'm thinking that we should throw an exception in this case, similar to the `pageCount <= 0`

docs/proposals/DataTable/README.md
demoraesgui: similar to the answer above :)

docs/proposals/DataTable/README.md
```diff
@@ -0,0 +1,165 @@
...
+    {
+      accessorKey: "name",
+      cell: info => <span>{info.getValue()}</span>,
+    },
+    {
+      accessorKey: "awesome",
+      cell: info => info.getValue(),
+    },
+  ]}
+/>
```
kkimble006, line 54: Any ideas on what sorting methods would look like passed in? And the definition of ASC vs DESC?

demoraesgui: Sure, we are finishing the basic implementation, and not long after we will update the proposal addressing this and the other comments.

kkimble006: Apologies, I realize this is still in 'draft' state.

docs/proposals/DataTable/README.md
lmorgan29: Cool! Throwing an exception works if the typing gets too messy

Though, you could do something like this I think:

```
interface CondType1 {
  manualPagination: true;
  onPaginationChange: () => void;
}

interface CondType2 {
  manualPagination: false;
  onPaginationChange?: () => void;
}
type CondTyping = CondType1 | CondType2;
```

I guess the only thing would be having to modify the use of the Pagination type

docs/proposals/DataTable/README.md
lmorgan29: I'm cool with this being a follow up. I was just thinking where there are some other client side things built in to this proposal (sorting, pagination) this would be included. All good

docs/proposals/DataTable/README.md
lmorgan29: awesome ^

docs/proposals/DataTable/README.md
lmorgan29: Okay, I assume you're waiting on that because it adds complexity for something not being asked for yet?
Might be worth documenting that it is a limitation we are imposing, in case someone needs to build this in the future

docs/proposals/DataTable/README.md
lmorgan29: It seems like you can just modify the state of the table to hide certain columns?

docs/proposals/DataTable/README.md
lmorgan29: cool, thank you!

PR #922, opened 2022-10-06
docs(components): MultiSelect Component Proposal

CHPrado:
## Motivations

For the new QBO [design](https://www.figma.com/file/Dat89UMBtcroHoy2TFpqxA/QBO-Integration?node-id=318%3A75252) we will need filters that works like a drop down menu and the user will be able to select multiple options.

### Added

- MultiSelect Component proposal document

### Mockup

image

---

Random photo of Atlantis


docs/proposals/MultiSelect/MultiSelect.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Mockup
+
+!["MultiSelect Mockup"](./multi_select.png)
+
+## Props Table
+
+| name                | type                  | default   | description                                                                                                            |
+| ------------------- | --------------------- | --------- | ---------------------------------------------------------------------------------------------------------------------- |
+| `items`             | `Array?`              | undefined | An associative array that represents the labels and values that are to be associated with each option, and determines. |
```
chris-at-jobber, line 102: Could we use "options" here? It would more closely mirror the conventions for a single-Select

docs/proposals/MultiSelect/MultiSelect.md
```diff
@@ -0,0 +1,105 @@
...
+      - If unchecked again after that then all of the lower options should be
+        unchecked automatically
+
+    - The “All” checkbox is automatically marked checked if the user manually
+      checks all of the options below it
+
+    - The “All” checkbox is automatically marked unchecked if the user manually
+      unchecks all of the options below it
+
+- The text that appears in the box before a user clicks on it should be either
```
chris-at-jobber, line 39: How would we handle a long list of selected options? Does the trigger element grow, or would it truncate?

docs/proposals/MultiSelect/MultiSelect.md
```diff
@@ -0,0 +1,105 @@
...
+  “All **\_\_\_\_**” or a comma separated list of the current selections
+
+## Accessibility
```
chris-at-jobber, line 42: 👌

chris-at-jobber: Love these considerations. One thing I'm wondering is whether we can simply use an `html` select ([with multiple enabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#attr-multiple)) internally, as it should already have most of the accessibility we're looking for?

CHPrado: Hi Chris, the default `html select` only supports the `option html object` as child elements, so it is not possible to implement the requested design that required `checkboxes`.

docs/proposals/MultiSelect/MultiSelect.md
CHPrado: I think we could add a prop to fix the width but let the default behavior to adjust to its content, is that ok?

CHPrado: For the QBO design specifically we will show max two options then truncate the third
image

docs/proposals/MultiSelect/MultiSelect.md
chris-at-jobber: I'm just referencing the accessibility-facing side of things. We do similar things with our radio button, for example, where the actual element the user is interacting with is different from the visual element represented on-screen because of styling limitations with the underlying `html` element:
image

chris-at-jobber: In this example, the radio input is actually rendered offscreen, but is still present in the DOM - so the user gets the same accessibility, while we have some more control over the visual presentation of the element.

docs/proposals/MultiSelect/MultiSelect.md
```diff
@@ -0,0 +1,105 @@
...
+## Mockup
+
+!["MultiSelect Mockup"](./multi_select.png)
+
+## Props Table
+
+| name                | type                  | default   | description                                                                                                            |
+| ------------------- | --------------------- | --------- | ---------------------------------------------------------------------------------------------------------------------- |
+| `options`           | `Array?`              | undefined | An associative array that represents the labels and values that are to be associated with each option, and determines. |
+| `noneSelectedLabel` | `String?`             | undefined | The text that will be displayed when no options are selected.                                                          |
```
lmorgan29, line 103: so when this prop is not provided, do we should an empty box with no text?

CHPrado: talking with our designer he mentioned that it should show "All ****" or the selected options labels. If that is the case then I think we won't need this prop. Any thoughts?

upwardbinkie: We don't have a design for empty state so we were thinking it should show something like "Types" or "Select a type". If no value is provided I guess we could go with the empty box.

docs/proposals/MultiSelect/MultiSelect.md
CHPrado: I was taking a look at the RadioGroup component and I understood it simulates the radio behavior though css in the real component that the user is interacting with. I did some tests and it is possible to do the same with the default select and options elements if the selectable options were the only thing within the options container. 

But it seems like more will have to be implemented in the future on this component, this is a screenshot of the Finsights project that have more elements in the options container like search bar and reset button. 

image

For that reason I believe it is not possible to implement the default select with the default options elements with this design, so the options component all of the accessibility will have to be implemented from scratch.

docs/proposals/MultiSelect/MultiSelect.md
chris-at-jobber: OK so the behaviour is always to truncate?

docs/proposals/MultiSelect/MultiSelect.md
chris-at-jobber: Got it, thanks @CHPrado

docs/proposals/MultiSelect/MultiSelect.md
CHPrado: As for the current QBO design, yes. But looking at the Finsights design it show "2 selected", "3 selected". Maybe it could be a switch on/off feature?

chris-at-jobber: What if the content was just a "label"? Then the responsive behaviour is to always truncate overflow, but what's displayed in the element could be set by each team. 

So in your case...

`someFunctionToGetTheStatusesForTheLabel`

`label={ someFunctionToGetTheStatusesForTheLabel }`

And for the other usage

`label={ numberOfSelectedFilters + 'selected' }`

?

CHPrado: yes! I believe this is a good approach to leave the decision of what should be the label to the developer with a label parameter. 

Thank you

docs/proposals/MultiSelect/MultiSelect.md
upwardbinkie: @chris-at-jobber I added the accessibility requirements you suggested.

chris-at-jobber:
I think the proposal for the immediate concerns of "a standardized trigger for launching a standardized popover that contains checkboxes which allow the user to select >1 item from a list" make sense and I'm happy with the design and a11y considerations made here.

My only remaining question after some more thought would be if this is planned to eventually be extended to include the functionalities from [the mockup](https://github.com/GetJobber/atlantis/pull/922#discussion_r991321015) above:

- search and filter
- single-select (radio)
- pagination (view more)

If it is, I think there's potentially some additional thoughts needed around what to name this component, as it would then extend beyond a simple multi-select component and into more advanced options, or whether this is truly a generic design system component or something that will require more implementation-specific tailoring. But if we're not looking at those other use cases, I'm good to go ahead with adding this to the system as a `MultiSelect`.

trevor-olsen:
> * search and filter
> * single-select (radio)
> * pagination (view more)
> 
@chris-at-jobber after thinking about it for a little bit, I would argue that these kinds of functionalities actually don't belong in a MultiSelect dropdown component. They are pretty specific to the use case of a column filter on a table. So maybe the best path forward would be to continue with the component in this proposal being a pretty generic multi-select dropdown, and then the Finsights team can adapt that to make an additional new component with a name like TableFilter or something like that.

PR #860, opened 2022-02-22
feat(design): add new icons to atlantis -- (Minor) Feature Release

jlapstra:
## Motivations

- The filter Icon is needed for Communigrade work, and while including the icon we have included two others that were also added by the design team!
- Followed these directions: https://jobber.atlassian.net/wiki/spaces/DES/pages/1658585445/Technical+Guide+Adding+an+icon+to+Jobber+s+design+system#Adding-to-Atlantis

## Changes

- Three new icons have been added to atlantis

### Added

- PaperClip, Image and Filter icons

## Testing

Make sure that you can go to the Icons page in Atlantis and see three new icons: Paperclip, Image and Filter.

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #923, opened 2022-10-12
JOB-55676 Add Future, Filter, Recurring, Back to Today & Recurring Icons

JoshMacSween:
## Motivations
We needed to add new icons to help our users interact with our new features, for example, the `recurring` icon will be used when a user is attempting to delete a recurring task. 

## Changes
Several icons have been added to the repository. 

#### New Icons
- `Recurring`
- `Filter`
- `Future`
- `Back-to-Today`

---

Random photo of Atlantis


package-lock.json
```diff
@@ -2749,7 +2749,7 @@
       "requires": {
         "@jobber/design": "^0.25.2",
         "@jobber/formatters": "^0.2.0",
-        "@jobber/hooks": "^1.11.0",
+        "@jobber/hooks": "^1.12.0",
```
jcw-, line 2752: If you're just adding icons, it is unexpected to see dependency changes here - not sure if including this file in your PR was intentional?

packages/design/src/icons/iconMap.ts
```diff
@@ -5,552 +5,409 @@
...
+    "back-to-today": [
+      "M682.667 128c-23.565 0-42.667 19.103-42.667 42.667v42.667h-256v-42.667c0-23.564-19.103-42.667-42.667-42.667s-42.667 19.103-42.667 42.667v42.667h-85.333c-47.128 0-85.333 38.205-85.333 85.333v512c0 47.13 38.205 85.333 85.333 85.333h597.333c47.13 0 85.333-38.204 85.333-85.333v-512c0-47.128-38.204-85.333-85.333-85.333h-85.333v-42.667c0-23.564-19.102-42.667-42.667-42.667zM810.667 298.667v85.333h-597.333v-85.333h597.333zM213.333 469.333h597.333v341.333h-597.333v-341.333zM704 640c-35.345 0-64-28.655-64-64s28.655-64 64-64c35.345 0 64 28.655 64 64s-28.655 64-64 64z"
     ],
     "recurring": [
```
jcw-, line 19: The modifications to this file are extensive - much more than just adding the four or five new icons mentioned in the PR description. Are all the changes to the existing icons expected?

chris-at-jobber:
Naming suggestion:

- "back-to-today" can just be "today"

packages/design/src/icons/iconMap.ts
chris-at-jobber: This looks unexpected, for reference here's a previous icon update that only touches the vectors of existing icons:

https://github.com/GetJobber/atlantis/pull/860/files

That said there were some issues on the IcoMoon side of generating the updated icons so I wouldn't be surprised if something ended up being re-drawn :thinking:

The icons themselves look as expected when I can convince this to build locally:
image

chris-at-jobber: This could explain the abundance of snapshot failures as the icons are being redrawn marginally differently, even when the "right icon" is being rendered.

chris-at-jobber:
In the PR description it says we're adding `recurring` but that's already available:

image

https://atlantis.getjobber.com/components/icon

packages/design/src/icons/iconMap.ts
JoshMacSween: Thank you @jcw- for the feedback. It's strange that these changes are unexpected, as I only ran the command `npm run convertIcoMoon` on the json file in the `jobberIcons` folder as instructed in the documentation. There was also a similar change to the `package-lock.json` file in a previous PR that I found so I thought that I should push the `package-lock.json` file as well. Should it be removed?

JoshMacSween:
@chris-at-jobber I have been unable to get to the icons page in the CloudFlare build, which is one of the reasons I reached out. I just went to master and ran the same command and there were still extensive changes to the `iconMap` file, but no change to the `package-lock.json` file, so I feel like I should revert that change.

JoshMacSween:
> In the PR description it says we're adding `recurring` but that's already available:
> 
> image
> 
> https://atlantis.getjobber.com/components/icon



> Naming suggestion:
> 
> * "back-to-today" can just be "today"

I wonder if this would need to be changed in icoMoon, because if I just change it in `iconMap`, the next time someone generates the fonts, the name will be different. Similar to the double icon name `recurring`. Maybe it could be renamed to something like `recurring-item`? 🤔

chris-at-jobber:
What does the "other" recurring icon look like? I'm not aware of any other new ones being introduced. I would expect:

image

packages/design/src/icons/iconMap.ts
jcw-: i probably would (looks like you already have!), to keep this PR more straight forward / easier to test

chris-at-jobber:
@JoshMacSween @jcw- new PR with some notes based on Rodrigo and myself doing some investigation this afternoon:

https://github.com/GetJobber/atlantis/pull/924#issuecomment-1278238396

jcw-:
@chris-at-jobber makes sense - thanks for following up!

jcw-:
https://github.com/GetJobber/atlantis/pull/924

PR #932, opened 2022-11-09
feat(components): MultiSelect component

CHPrado:
## Motivations

Add new MultiSelect component to have the ability to select multiple options in a list

### Added

MultiSelect Component with the following features:
- Options displayed as a list of checkboxes;
- Custom default Label, Custom All options selected Label;
- The MultiSelect trigger element contains a downward-pointing-arrow to hint that it can be expanded;
- Users are able to focus the MultiSelect trigger element by using the Tab key;
- Users should be able to display the options by using the Space or Enter/Return keys;
- Users should be able to navigate through the options by using Tab or ArrowDown/ArrowUp keys and check/uncheck them by using Space or Enter/Return keys;
- Users can press the Escape key, click out of the component or click the trigger again to close the menu;
- Features and behavior are announced to the user when using a screen reader;

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/MultiSelect/DropDownMenu.css
+++ b/packages/components/src/MultiSelect/DropDownMenu.css
@@ -1,0 +1,28 @@
+.dropDownMenuContainer {
+  display: flex;
+  position: absolute;
+  z-index: var(--elevation-menu);
+  width: 100%;
+  max-height: 300px;
+  box-shadow: var(--shadow-base);
+  padding-bottom: var(--space-smaller);
+  border: var(--border-base) solid var(--color-border);
+  border-radius: var(--radius-large);
+  overflow: auto;
+  background-color: var(--color-surface);
+  flex-direction: column;
+  scroll-behavior: smooth;
+}
+
+.option {
+  width: 100%;
+  padding: var(--space-small) var(--space-base) 0;
+  cursor: pointer;
+  transition: all var(--timing-base);
+}
+
+.option:focus,
+.option.active {
+  background-color: var(--color-surface--hover);
+  outline-color: var(--color-focus);
+}
--- a/packages/components/src/MultiSelect/DropDownMenu.css
+++ b/packages/components/src/MultiSelect/DropDownMenu.css
@@ -7,7 +7,6 @@
   box-shadow: var(--shadow-base);
   margin: 0;
   padding: 0;
-  padding-bottom: var(--space-smaller);
   border: var(--border-base) solid var(--color-border);
   border-radius: var(--radius-large);
   overflow: auto;
@@ -21,8 +20,9 @@
 }
 
 .option {
+  display: flex;
   width: 100%;
-  padding: var(--space-small) var(--space-base) 0;
+  padding: var(--space-small) var(--space-base);
   cursor: pointer;
   transition: all var(--timing-base);
 }
--- a/packages/components/src/MultiSelect/DropDownMenu.css
+++ b/packages/components/src/MultiSelect/DropDownMenu.css
@@ -5,6 +5,8 @@
   width: 100%;
   max-height: 300px;
   box-shadow: var(--shadow-base);
+  margin: 0;
+  padding: 0;
   padding-bottom: var(--space-smaller);
   border: var(--border-base) solid var(--color-border);
   border-radius: var(--radius-large);
@@ -12,6 +14,10 @@
   background-color: var(--color-surface);
   flex-direction: column;
   scroll-behavior: smooth;
+}
+
+.dropDownMenuContainer li {
+  list-style: none;
 }
 
 .option {
--- a/packages/components/src/MultiSelect/DropDownMenu.css
+++ b/packages/components/src/MultiSelect/DropDownMenu.css
@@ -3,7 +3,7 @@
   position: absolute;
   z-index: var(--elevation-menu);
   width: 100%;
-  max-height: 300px;
+  max-height: calc(var(--space-extravagant) * 5);
   box-shadow: var(--shadow-base);
   margin: 0;
   padding: 0;
--- a/packages/components/src/MultiSelect/MultiSelect.css
+++ b/packages/components/src/MultiSelect/MultiSelect.css
@@ -1,0 +1,13 @@
+.multiSelectContainer {
+  position: relative;
+  width: 100%;
+}
+
+.multiSelectContainer:hover,
+.multiSelectContainer input {
+  cursor: pointer;
+}
+
+.multiSelectContainer.active {
+  box-shadow: var(--shadow-focus);
+}
--- a/packages/components/src/MultiSelect/MultiSelect.css
+++ b/packages/components/src/MultiSelect/MultiSelect.css
@@ -31,6 +31,8 @@
   caret-color: transparent;
 }
 
+.multiSelect:focus,
 .multiSelect.active {
   box-shadow: var(--shadow-focus);
+  outline-color: transparent;
 }
--- a/packages/components/src/MultiSelect/MultiSelect.css
+++ b/packages/components/src/MultiSelect/MultiSelect.css
@@ -34,5 +34,5 @@
 .multiSelect:focus,
 .multiSelect.active {
   box-shadow: var(--shadow-focus);
-  outline-color: transparent;
+  outline: none;
 }
--- a/packages/components/src/MultiSelect/MultiSelect.css
+++ b/packages/components/src/MultiSelect/MultiSelect.css
@@ -5,18 +5,15 @@
 
 .multiSelect {
   display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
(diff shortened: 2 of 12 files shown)


chris-at-jobber:
Some design and a11y notes:

Design: minor 🟡 
- I think in the proposal we were planning to truncate overflowing labels
image
- I don't think the arrow needs to flip when the menu opens. That's more of a Disclosure component pattern than a Select pattern

Accessibility: major 🔴 
- We should drop the use of a text input for this component. When I use a screen-reader, it announces the element as a text input, and it sets my cursor into the labels, which I cannot edit.
- I cannot access the selection options without using my mouse. When the user activates the selection menu, we should set focus to the first element in the list and the user should be able to navigate by keyboard, as well as close the menu by selecting the `esc` key. Maybe we consider using a `button` element internally?
- there's no indication to the user that they can use this element to make selections. For the menu, since we're unable to use native HTML `select`, we should consider using [aria-multiselectable](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable#:~:text=The%20aria%2Dmultiselectable%20attribute%20is,one%20item%20at%20a%20time.) or whatever is most appropriate to convey the expected usage to the user

chris-at-jobber (changes requested):
Some significant issues around semantic element usage and keyboard operation that need to be addressed, visually looking pretty good so far!

packages/components/src/MultiSelect/MultiSelect.css
```diff
@@ -0,0 +1,13 @@
+.multiSelectContainer {
+  position: relative;
+  width: 100%;
+}
+
+.multiSelectContainer:hover,
+.multiSelectContainer input {
+  cursor: pointer;
```
rodrigoeidelvein, line 8:
```suggestion
  cursor: pointer;
  caret-color: transparent;
```

rodrigoeidelvein (commented):
Code overall looks good! Nice work here!

Just a minor comment and a suggestion:
Can we dismiss the select when the user clicks outside of the component too? Currently we can only close it if we click again on the  `MultiSelect`

packages/components/src/MultiSelect/MultiSelect.css
CHPrado: Thank you for the suggestion, Rodrigo. It was implemented with the outside click

chris-at-jobber:
Seeing new issues introduced by using a `div` as the trigger: https://loom.com/share/7e71b910936c4353bcb9aea7ebb08c65

In summary:

- user needs to be able to focus the trigger element using the tab key
- trigger element should announce itself as an appropriate interactive element type when focused
- label should be announced when the trigger element is focused
- user needs to be able to activate the popup by pressing space or enter/return on the trigger element
- once opened, popup should be announced to screenreader
- focus should be set to the first checkbox element upon the popup opening
- user needs to be able to operate the checkboxes with the space key; `return` key is non-standard behaviour for checkbox selection
- when user presses `esc` to close the menu, focus needs to return to the trigger element

chris-at-jobber (dismissed):
Looking better! A few things left to address, visually + a11y-wise.

## Visual

### Focus state is incorrect/inconsistent

Focus:
image

Activated and focused:
image

Activated and focused within the popover:
image

The expected appearance in all 3 of those states would be:
image

### Hover state on checkboxes

When I hover the checkboxes in the popover, I'm seeing the focus state. Checkbox does not have a hover state other than the cursor changing; the background color of the list items changing is sufficient to convey to the user that they're hovering.
image

### Alignment of checkboxes

The spacing around the checkboxes is inconsistent (10 above, 6 below), should be `small` above and below:
image


## Accessibility

### Activator role
Currently the activator is read by screenreaders as "{Label(s)} button group". It's not apparent what the labels represent or that the button will offer multi-select functionality.

I think we should move the aria-multi-selectable role onto the activator, and add a "has pop-up" attribute, so that it's more apparent that:
- this button triggers a pop-up
- this button enables multi-select functionality

With the aria-multiselectable currently set to the popup, I feel it's less useful as the user is immediately set to focus on a checkbox, which implies that selections are available.

I think it would also be valuable to add a required `aria-label` string prop that will announce a label to screen-readers so they don't just hear the visible label (which is the currently selected contents).

chris-at-jobber:
@CHPrado this is great! One small refinement would be to use the value of `defaultLabel` as the `ariaLabel` value unless the user specifies otherwise. That way even when the user has made selections, they'll still hear the label for the options.

Using the current first example, when I have made selections and don't pass in ariaLabel I hear: 

"Warnings, Ignored"

If the defaultLabel value were passed in to the ariaLabel I would hear:

"Status: Warnings, Ignored"

And if I wanted to use something else for the ariaLabel, it could be something like

"Status filters: Warnings, Ignored"

I suppose another approach could just be to remove the ariaLabel as a prop and always use the `defaultValue` for the ariaLabel since that's already required, as that removes a bit of overhead from the consuming developer. What do you think?

packages/components/src/MultiSelect/MultiSelect.mdx
```diff
@@ -0,0 +1,134 @@
...
+html select would.
+
+<!--
+  How should the component behave on an iPhone 8? An iPad? A 1920x1080 monitor? How does the
+  component appear when the component itself is less than 375px wide? Less than 640px wide?
+  Less than 1200px wide? Greater than 1200px? Does it change when the device DPI is higher
+  or lower?
+-->
+
+## Mockup
```
chris-at-jobber, line 119: Can remove this section

packages/components/src/MultiSelect/MultiSelect.mdx
```diff
@@ -0,0 +1,134 @@
...
+options within a list of items.
+
+A user can click on the element to expand it and display the multiple options.
+The element will display the selected options or the placeholder text if none
+are selected.
+
+Every time an option is checked or unchecked a behavior can be triggered using
+the onChange prop
+
+<!--
```
chris-at-jobber, line 82: general: all of these commented out bits can be removed

packages/components/src/MultiSelect/MultiSelect.css
```diff
@@ -0,0 +1,38 @@
...
+}
+
+.multiSelect p {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.multiSelectContainer:hover,
+.multiSelectContainer.multiSelect {
```
chris-at-jobber, line 29: Do we still need these input-specific stylings since the component no longer uses a FormField internally?

packages/components/src/MultiSelect/DropDownMenu.css
```diff
@@ -0,0 +1,34 @@
+.dropDownMenuContainer {
+  display: flex;
+  position: absolute;
+  z-index: var(--elevation-menu);
+  width: 100%;
+  max-height: 300px;
```
chris-at-jobber, line 6: Can we replace this with a unit based our space variables? Will allow it to resize on browser zoom in/out.

packages/components/src/MultiSelect/MultiSelect.css
```diff
@@ -0,0 +1,38 @@
...
+  border-radius: var(--radius-base);
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.multiSelect p {
```
chris-at-jobber, line 22: Since we're using Text in the component and Text already has [truncation baked in](https://job-55182-multiselect-dropdo.atlantis.pages.dev/packages-components-src-text-text#maxlines) do we need this?

chris-at-jobber (commented):
@CHPrado also left some minor styling-code and documentation cleanup opportunities. Those could be fast-follows but I think there's some CSS that we no longer need with the changes from the past few rounds of revisions.

packages/components/src/MultiSelect/MultiSelect.css
CHPrado: yes, without it I can still see the blinking cursor in front of the text even though it is not an input text anymore

CHPrado:
> I suppose another approach could just be to remove the ariaLabel as a prop and always use the `defaultValue` for the ariaLabel since that's already required, as that removes a bit of overhead from the consuming developer. What do you think?

@chris-at-jobber I agree with this last part, so I'll remove the `ariaLabel` prop and use the `defaultValue` and the selected options so we could hear something like:

"Status: Warnings, Ignored"

chris-at-jobber (commented):
Design approved + all of the styling code looks great. I'd like to have an engineer look over the tests, so I've tagged @amia-real for a final code review.

amia-real (approved):
LGTM!

chris-at-jobber:
Thanks for your work on this @CHPrado !

PR #934, opened 2022-11-11
feat(components): Add footer row to Data Table component (JOB-57643)

danlaurent:
## Motivations

Add a footer row in our Data Table component to show the totals / conclusion of the current page data.

## Changes

Add footer row to Data Table component using TanStack column footers

### Added

- Footer Row as a Footer sub-component

### Changed

- The old Footer sub-component was renamed to Pagination

#### Why?
Since now we're adding a real Footer (using the `<tfoot />` tag, it's more readable / semantic to name it as a Footer, also the Pagination name for that component is more accurate since the component has only pagination behavior inside it.

---

Random photo of Atlantis


kyleherrmann (commented):
Tested and code LGTM! Just confirming our mobile view does not support a width less than 350px 🤔 I doubt we would ever encounter a screen size this small.

chris-at-jobber (commented):
Question: Is this a breaking change for anyone already using the Footer? (If yes, can the PR title please be updated to reflect?)

danlaurent:
> Question: Is this a breaking change for anyone already using the Footer? (If yes, can the PR title please be updated to reflect?)

No breaking changes, as it's just an internal sub-component rename, there is no impact in usage 😃

danlaurent:
> Tested and code LGTM! Just confirming our mobile view does not support a width less than 350px 🤔 I doubt we would ever encounter a screen size this small.

@kyleherrmann  In my tests I reached 260px without breaking the layout 🤔 
Screen Shot 2022-11-21 at 14 53 25

danlaurent:
> > 
> 
> Interesting, the left margin breaks in my browser, this is 300px.
> 
> Screenshot 2022-11-21 at 10 08 00 AM

@kyleherrmann  Ohh, that's probably because there is 2 margin / spacing around the table (the actual page one and the Playground one) 🤔 

But I'll ask Irem about the text behavior, if we truncate or wrap, good catch, thanks!

MichaelParadis (commented):
Looks good, some minor comments

MichaelParadis (approved):
LGTM!

packages/components/src/DataTable/DataTable.css
```diff
@@ -36,6 +36,20 @@
   padding: var(--space-base);
 }
 
+.table tfoot {
+  position: sticky;
+  z-index: 1;
+  width: 100%;
+  box-shadow: var(--shadow-base);
+  border-top: var(--border-thick) solid var(--color-border);
+  background-color: var(--color-white);
```
chris-at-jobber, line 45:
```suggestion
  background-color: var(--color-surface);
```

chris-at-jobber (dismissed):
Tiny change! That [color token for the background](https://github.com/GetJobber/atlantis/pull/934/files#r1032583694) should get replaced with the semantic token but otherwise looks good on my end

packages/components/src/DataTable/DataTable.css
danlaurent: @chris-at-jobber  Thanks! Just pushed this change :)

chris-at-jobber (approved):

chris-at-jobber:
Thanks @danlaurent !

PR #937, opened 2022-11-17
fix(component) Adjust dropdown height in DataTable compoment

CHPrado:
## Motivations

The height of the dropdown box was a little taller than we planned for:

image

## Changes

- Adjust items per page dropdown height in the footer;
- Remove svg placeholder from headers;

### Fixed

image

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataTable/Header.tsx
+++ b/packages/components/src/DataTable/Header.tsx
@@ -2,13 +2,13 @@ import { Row, Table, flexRender } from "@tanstack/react-table";
 import classNames from "classnames";
 import React from "react";
 import styles from "./DataTable.css";
-import { Sorting } from "./types";
+import { SortingType } from "./types";
 import { Icon } from "../Icon";
 
 interface HeaderProps<T> {
   table: Table<T>;
   stickyHeader?: boolean;
-  sorting?: Sorting;
+  sorting?: SortingType;
   onRowClick?: (row: Row<T>) => void;
 }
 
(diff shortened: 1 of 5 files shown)


packages/components/src/DataTable/Header.tsx
```diff
@@ -45,17 +45,12 @@ export function Header<T extends object>({
...
-                          height: "var(--space-large)",
-                        }}
-                      />
-                    )}
+                    {
+                      {
+                        asc: <Icon name="longArrowUp" color="green" />,
+                        desc: <Icon name="longArrowDown" color="green" />,
+                      }[header.column.getIsSorted() as string]
+                    }
```
lmorgan29, line 53: My only uncertainty here is I am seeing some column widths grow as I interact with it. This example https://atlantis.getjobber.com/components/data-table#table-with-footer-row does not play well, for example.

I did compare against master and do see that behaviour here on the up arrow specifically https://atlantis.getjobber.com/components/data-table#client-side-pagination-sorting-clickable-rows-and-fixed-height-with-sticky-header

Looks like it was already a problem, maybe slightly less bad on master. Any idea around fixing this?

demoraesgui: The svg was added to mitigate that and introduced the right-align bug.
I don't see a way to make the width of the column constant when you're sorting and the data can change completely. 
this also doesn't fix the right-align when sorting because the icon would still be there. Is just removing a bad workaround that introduced more bugs

CHPrado: @lmorgan29 [this](https://getjobber.slack.com/archives/C0405GDEASJ/p1668012586359169?thread_ts=1667945738.156329&cid=C0405GDEASJ) is the right-align bug commend that @demoraesgui is talking about

jsuazod: Hi Lori, I think this is the default behavior of a table, if the content changes the columns on the table should move to accommodate that content. Not sure if this is something that we can be fixed even if we added a min-width to the header that would only account for that specific width, if the content exceeds that default min-width the problem will come again.

lmorgan29 (approved):
LGTM, can figure out if we can optimize column resizing later

PR #941, opened 2022-11-30
fix: export tanstack context types for TS <4.0 compatibility

danlaurent:
## Motivations

In projects using TS<4.0 the props types of render functions for cell, footer and header are not recognized by TS automatically, so this adds those types to be used in order to define manually those types

## Changes

N/A

### Added

New exported types from tanstack

---

Random photo of Atlantis


kyleherrmann (approved):
Tested, code LGTM

PR #945, opened 2022-12-05
feat(component) add size property to MultiSelect component

CHPrado:
## Motivations

The MultiSelect component height was not matching the InputText component

### Added

added size property with `small` and `large` possible values

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/MultiSelect/MultiSelect.mdx
+++ b/packages/components/src/MultiSelect/MultiSelect.mdx
@@ -8,6 +8,7 @@ showDirectoryLink: true
 import { Playground, Props } from "docz";
 import { MultiSelect } from ".";
 import { useState } from "react";
+import { Divider } from "@jobber/components/Divider";
 
 # Multi Select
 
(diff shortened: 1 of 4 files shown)


packages/components/src/MultiSelect/MultiSelect.mdx
```diff
@@ -41,6 +41,52 @@ import { MultiSelect } from "@jobber/components/MultiSelect";
 
 <Props of={MultiSelect} />
 
+## Sizes
```
chris-at-jobber, line 44: These should be nested within the Usage Guidelines section (this "Sizes" header will become an H3, the "Small" and "Large" will become H4s) and have documentation on when to use each variation.

chris-at-jobber: Maybe something like

```suggestion
## Sizes

Adjusting the size of the `MultiSelect` can be used to help align with common adjacent elements such as InputText or Select.
```

And then you could even put both size examples into a single <Playground> to save vertical space.

lmorgan29 (approved):
LGTM

PR #951, opened 2022-12-13
feat(components): Add configurable column width to DataTable

danlaurent:
## Motivations

To be able to adjust the column width manually in order to emphasize column data

### Added

- Configurable column width options in column prop (`size, minSize, maxSize`)

## Testing

- Add `size, minSize, maxSize` to a `column` object and test if this style is applied to the column element

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataTable/DataTable.mdx
+++ b/packages/components/src/DataTable/DataTable.mdx
@@ -662,6 +662,7 @@ import { DataTable } from "@jobber/components/DataTable";
         accessorKey: "name",
         header: "Name",
         footer: "Totals",
+        enableResizing: false,
       },
       {
         id: "points",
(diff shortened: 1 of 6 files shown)


kkimble006:
@danlaurent can you added definitions for these new props: Add size, minSize, maxSize to a column object and test if this style is applied to the column element. Or link to the appropriate tanstack docs?

danlaurent:
> https://1e57cfd2.atlantis.pages.dev

@kkimble006  Yeah, sure!
You can follow [this example](https://1e57cfd2.atlantis.pages.dev/components/data-table#table-with-manual-column-sizing) that I added to the docs or the [tanstack docs](https://tanstack.com/table/v8/docs/api/features/column-sizing?from=reactTableV7&original=https://react-table-v7.tanstack.com/docs/api/useResizeColumns#column-def-options).
After that you can inspect the element and see if the styling properties were applied

danlaurent:
> @danlaurent I did have a question about the prop names themselves - given that this specifically controls column width, did you consider `width`, `minWidth` and `maxWidth`? Those would map more cleanly to the CSS underneath than `size` which I feel could lend itself to assumptions about having control over both height and width

@chris-at-jobber Yeah, I totally agree, I would prefer to use `width`, but as we're using the built in Column Sizing API from Tanstack, this is the nomenclature that the library is using, we're only consuming it. Here's the [docs](https://tanstack.com/table/v8/docs/api/features/column-sizing?from=reactTableV7&original=https://react-table-v7.tanstack.com/docs/api/useResizeColumns#size)

chris-at-jobber:
Oh so do we have any control over the props interface with DataTable? Or are we totally tied to TanStack's naming?

danlaurent:
> Oh so do we have any control over the props interface with DataTable? Or are we totally tied to TanStack's naming?

@chris-at-jobber I mean, we could add 3 extra "custom" props to this interface (by extending the library type) but I have some concerns:
- All other helper methods are using the same nomenclature (`getSize`, `startSize`)
- I didn't take a look on how the library built the helper methods, but by using custom props in the columnDef maybe those methods won't work properly

chris-at-jobber:
Ahhhhhh OK I see what you're saying. Thanks for the clarification!

packages/components/src/DataTable/DataTable.mdx
```diff
@@ -662,6 +662,7 @@ import { DataTable } from "@jobber/components/DataTable";
         accessorKey: "name",
         header: "Name",
         footer: "Totals",
+        enableResizing: false,
```
MichaelParadis, line 665: Is this needed?

MichaelParadis (approved):
LGTM! Minor comment but looks good

packages/components/src/DataTable/DataTable.mdx
danlaurent: Not needed, I'll remove it, thanks!

PR #953, opened 2022-12-21
feat(design): add syncAlert and apps icons 

taylorvnoj:
## Motivations
Upload two new icons into Atlantis

### Added
- `syncAlert` icon
- `apps` icon

### Changed
`Icon.mdx` was updated with the following:

`apps` was brought into the "Other" section:
Screenshot 2022-12-21 at 3 53 10 PM

`syncAlert` was brought into the "System actions" section:
Screenshot 2023-01-05 at 6 06 09 AM

---

Random photo of Atlantis


chris-at-jobber:
Technically this looks alright, but I have some design questions re: `syncAlert`:

1. Is the significant difference in proportions of the two icons intentional? I'm imagining a use case where perhaps a sync icon swaps for a syncAlert icon and the size increase looking jumpy.
2. Similarly, the gap between the sync arrows on the syncAlert icon seems larger than the gap on the sync icon - can that be standardized?
3. I can see why the rotation of the arrows has been changed to accommodate the exclamation but it's causing the arrows to render at _just off_ a 90-degree angle. Can those be adjusted to sit at an even 90 to avoid sub-pixel rendering?

chris-at-jobber (changes requested):
Noted 3 questions, the first is more up for discussion but there are some cleanups that I think should happen on the `syncAlert` before it goes in

taylorvnoj:
> Noted 3 questions, the first is more up for discussion but there are some cleanups that I think should happen on the `syncAlert` before it goes in

Thank you for chatting about this with me and George, @chris-at-jobber!  The new `syncAlert` icon has been pushed.

PR #969, opened 2023-01-10
feat(components): add modal size and child render support to confirmation modal

igor-m-jobber:
## Motivations
We had to use the <ConfirmationModal> component to display a warning to the user.
The message that had to be displayed could be fit in a `string` or `Markdown` format. We had to use JSX element. However, <ConfirmationModal> does not support displaying of the `children` in the component.

The confirmed design of the data that we need to display had to use a bigger size of the modal to fit all text nicely. But <ConfirmationModal> didn't have the ability to dynamically change the size of the displayed modal. This PR adds this functionality.

## Changes
- With the current proposal, <ConfirmationModal> can display either provided `message` or `children` component.
- Support for the modal size has been added.

### Added
- Added 2 new TS interfaces that must enforce displaying of information on the component

### Changed
- Updated `ConfirmationModalProps` structure to include support of the children component

### Deprecated
- n/a

### Removed
- n/a

### Fixed
- n/a

### Security
- n/a

## Testing
Added additional test that makes sure that either the `message` or `children` component is displayed.

1. Pass different modal sizes and make sure modal changes in size.
2. Try to add `message` and `children` props to the <ConfirmationModal> component. You should see TS error that won't let you do it.
3. Remove `message` prop from and make sure `children` are displayed.
3. Add back `message` prop and remove `children`. Make sure that `message` is displayed.

---

Random photo of Atlantis


packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -112,6 +114,16 @@ interface BaseConfirmationModalProps {
...
 
+  /**
+   * Size of the modal (small, large),
+   */
+  readonly size?: "small" | "large";
+
+  /**
+   * Child component. Not displayed if **message** prop is passed.
+   */
+  readonly children?: ReactNode;
```
danlaurent, line 125: @igor-m-jobber  I tested this in the mdx Playground and I was able to see both, the message and the children I added 🤔 
Possibly because the TS don't validate this file but, I was wondering if we should add an explicit validation for this behavior. Any thoughts?
Screen Shot 2023-01-12 at 16 18 07

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -222,7 +245,12 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
         onClick: handleAction("cancel"),
       }}
     >
-      <Content>{state.message && <Markdown content={state.message} />}</Content>
+      {state.message && (
+        <Content>
+          <Markdown content={state.message} />
+        </Content>
+      )}
+      {children}
```
danlaurent, line 253: Following my comment above, I was wondering if we should do something like this
```suggestion
      {state.message ? (
        <Content>
          <Markdown content={state.message} />
        </Content>
      ) : children}
```

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
igor-m-jobber: Ohhh! That is a good one @danlaurent!
You got it right! I was counting on the TS to show you the warning. So this way developer will have to make only 1 choice. But looks like TS doesn't work in the playground.
We definitely should add something. And your ternary operator code below should do the trick!

I will update the code. Thank you!

danlaurent (approved):
LGTM! 🚀

PR #973, opened 2023-01-13
feat(components): Add empty state behavior to DataTable component in Atlantis [JOB-58935]

kyleherrmann:
## Motivations

- To improve UX when viewing a data table with no content

## Changes

- Conditional rendering of text per designs

### Changed

- The `Body` component used by the `DataTable`

## Testing

- Run the Atlantis web app locally and navigate to the [empty state example](http://localhost:3333/components/data-table/#table-empty-state), validate the experience on mobile and desktop views.
---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataTable/Body.tsx
+++ b/packages/components/src/DataTable/Body.tsx
@@ -57,11 +57,7 @@ export function Body<T extends object>({
     <div className={classNames(styles.emptyState)} style={{ height }}>
       <div className={classNames(styles.emptyMessage)}>
         <Heading level={5}>No items found</Heading>
-        <Text>
-          No items to display for the
-          <br />
-          selected date range
-        </Text>
+        <Text>No items found, update your search or filter selection</Text>
       </div>
     </div>
   );
--- a/packages/components/src/DataTable/Body.tsx
+++ b/packages/components/src/DataTable/Body.tsx
@@ -2,13 +2,20 @@ import { Row, Table, flexRender } from "@tanstack/react-table";
 import classNames from "classnames";
 import React, { useCallback } from "react";
 import styles from "./DataTable.css";
+import { Heading } from "../Heading";
+import { Text } from "../Text";
 
 interface BodyProps<T> {
   table: Table<T>;
   onRowClick?: (row: Row<T>) => void;
+  height?: number;
 }
 
-export function Body<T extends object>({ table, onRowClick }: BodyProps<T>) {
+export function Body<T extends object>({
+  table,
+  onRowClick,
+  height,
+}: BodyProps<T>) {
   const bodyRowClasses = classNames({ [styles.clickableRow]: !!onRowClick });
 
   const handleRowClick = useCallback(
@@ -18,7 +25,8 @@ export function Body<T extends object>({ table, onRowClick }: BodyProps<T>) {
     },
     [onRowClick],
   );
-  return (
+
+  return table.getRowModel().rows.length ? (
     <tbody>
       {table.getRowModel().rows.map(row => {
         return (
@@ -45,5 +53,16 @@ export function Body<T extends object>({ table, onRowClick }: BodyProps<T>) {
         );
       })}
     </tbody>
+  ) : (
+    <div className={classNames(styles.emptyState)} style={{ height }}>
+      <div className={classNames(styles.emptyMessage)}>
+        <Heading level={5}>No items found</Heading>
+        <Text>
+          No items to display for the
+          <br />
+          selected date range
+        </Text>
+      </div>
+    </div>
   );
 }
--- a/packages/components/src/DataTable/Body.tsx
+++ b/packages/components/src/DataTable/Body.tsx
@@ -1,22 +1,21 @@
 import { Row, Table, flexRender } from "@tanstack/react-table";
 import classNames from "classnames";
-import React, { useCallback } from "react";
+import React, { ReactNode, useCallback } from "react";
 import styles from "./DataTable.css";
 import { Heading } from "../Heading";
-import { Text } from "../Text";
 
 interface BodyProps<T> {
   table: Table<T>;
   onRowClick?: (row: Row<T>) => void;
   height?: number;
-  emptyTableBodyMessage?: string;
+  children?: ReactNode | ReactNode[];
 }
 
 export function Body<T extends object>({
   table,
   onRowClick,
   height,
-  emptyTableBodyMessage,
+  children,
 }: BodyProps<T>) {
   const bodyRowClasses = classNames({ [styles.clickableRow]: !!onRowClick });
 
@@ -59,9 +58,7 @@ export function Body<T extends object>({
     <div className={classNames(styles.emptyState)} style={{ height }}>
       <div className={classNames(styles.emptyMessage)}>
         <Heading level={5}>No items found</Heading>
-        {emptyTableBodyMessage ? (
-          <Text>{emptyTableBodyMessage}</Text>
-        ) : undefined}
+        {children}
       </div>
     </div>
   );
--- a/packages/components/src/DataTable/Body.tsx
+++ b/packages/components/src/DataTable/Body.tsx
@@ -9,12 +9,14 @@ interface BodyProps<T> {
   table: Table<T>;
   onRowClick?: (row: Row<T>) => void;
   height?: number;
+  emptyTableBodyMessageElement?: string;
 }
 
 export function Body<T extends object>({
   table,
   onRowClick,
   height,
+  emptyTableBodyMessageElement,
 }: BodyProps<T>) {
   const bodyRowClasses = classNames({ [styles.clickableRow]: !!onRowClick });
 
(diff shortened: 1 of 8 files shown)


kyleherrmann:
General implementation should not change, meeting with @iremdesign to finalize UX this afternoon

chris-at-jobber (changes requested):
I don't think there should be a hard-coded empty state message unless we can make it more generic. Since the DataTable doesn't come with search-ability or filtering built in, the current message would be incorrect unless the implementation provides those features.

What do you think about passing in a {child} here instead so implementers have more control over the various ways you could present this? For example, in some cases you might want to provide a user with a CTA to add data, an illustrative element, contextual empty states, or other more explanatory content.

The approach in the Pagination seems fine as "no items" would hold true in any case where there are no items.

kyleherrmann:
> I don't think there should be a hard-coded empty state message unless we can make it more generic. Since the DataTable doesn't come with search-ability or filtering built in, the current message would be incorrect unless the implementation provides those features.
> 
> What do you think about passing in a {child} here instead so implementers have more control over the various ways you could present this? For example, in some cases you might want to provide a user with a CTA to add data, an illustrative element, contextual empty states, or other more explanatory content.
> 
> The approach in the Pagination seems fine as "no items" would hold true in any case where there are no items.

Thanks for the review Chris! I'll take the approach of passing in a child for the centered message.

iremdesign:
Hi Chris, not sure what you mean by this 

> DataTable doesn't come with search-ability or filtering built in

For most of the use cases especially in Reporting and Finbooks (Quickbooks integration), we will have filtering. I don't know when we will use the new `DataTable` component in other areas of Jobber. You are right that some reports won't have search or filtering functionality but it's not too worrisome at this point. I think this copy already went through multiple iterations to make it as generic as possible and it should work for us and Finbooks' use case #1. You can see their ticket here and use cases listed -> [(https://jobber.atlassian.net/browse/JOB-59627)]

@kyleherrmann from your perspective, is passing a child something you can do with ease to help other teams such as Finbooks to help with their implementation of other states such as the empty state #2 (due to typos)? We wanted to settle on a leaner empty state with generic copy because Jobber Built Apps team wasn't sure if they could easily modify the `DataTable` component

iremdesign:
I cannot use Github lol https://jobber.atlassian.net/browse/JOB-59627

packages/components/src/DataTable/Body.tsx
```diff
@@ -2,13 +2,22 @@ import { Row, Table, flexRender } from "@tanstack/react-table";
...
 import React, { useCallback } from "react";
 import styles from "./DataTable.css";
+import { Heading } from "../Heading";
+import { Text } from "../Text";
 
 interface BodyProps<T> {
   table: Table<T>;
   onRowClick?: (row: Row<T>) => void;
+  height?: number;
+  emptyTableBodyMessage?: string;
```
alexander-meyer, line 12: Are we choosing to limit this child element to text only? It seemed like based on Chris's comment there could be other types of content we may wish to provide (e.g. "an illustrative element")

alexander-meyer (commented):
Looks good on the test, just had one question.

packages/components/src/DataTable/Body.tsx
kyleherrmann: @iremdesign

chris-at-jobber:
@iremdesign whether the current usages of DataTable _leverage_ search and filtering in their respective implementations is outside of the responsibility of the DataTable component. DataTable in and of itself does not have search or filtering "baked in", and so the component should not by default reference functionality that exists outside of itself.

That's why I am suggesting the DataTable component should allow for a child component to be passed in for the empty state, so that any team can optimize for their use case, rather than making a static empty state that has hard and fast styling or functionality limitations, or has to be made overly generic when context would help (ie search / filter messaging).

packages/components/src/DataTable/Body.tsx
chris-at-jobber: That's correct, my suggestion was intended to be that this prop should actually accept a child component rather than a string – this way, a wider range of use cases can be supported for individual teams and we don't need to be overly prescriptive at the system level.

kyleherrmann:
@chris-at-jobber Just pushed the implementation using child elements, let me know your thoughts when you have a minute

packages/components/src/DataTable/Body.tsx
```diff
@@ -45,5 +54,12 @@ export function Body<T extends object>({ table, onRowClick }: BodyProps<T>) {
         );
       })}
     </tbody>
+  ) : (
+    <div className={classNames(styles.emptyState)} style={{ height }}>
+      <div className={classNames(styles.emptyMessage)}>
+        <Heading level={5}>No items found</Heading>
```
chris-at-jobber, line 60: This heading should be removed as well, and I think the divs around it are no longer needed? I don't think there should be any styling or UI elements in here that consumers would have to override to achieve their intended design. I can see maybe keeping the `emptyState` div on line 58 remaining as a generic "we can assume users will want to drop something in the middle" container that's easy enough to do custom work within, with a few styling tweaks.

packages/components/src/DataTable/DataTable.css
```diff
@@ -100,3 +100,28 @@
   background: var(--color-surface--background);
   cursor: pointer;
 }
+
+.emptyState {
+  display: flex;
+  width: 100%;
+  text-align: center;
```
chris-at-jobber, line 107: Shouldn't these two lines be using

```
justify-content: center;
align-items: center;
```

if the goal is to always keep the empty state content centered?

chris-at-jobber (changes requested):
A few small changes!

packages/components/src/DataTable/DataTable.mdx
```diff
@@ -707,6 +711,70 @@ import { DataTable } from "@jobber/components/DataTable";
...
+          </Typography>
+        ),
+        cell: info => <Typography align="end">{info.getValue()}</Typography>,
+      },
+    ]}
+  >
+    <div className={classNames(styles.emptyMessage)}>
+      <Heading level={5}>No items found</Heading>
+      <Text>Update your search or filter selection</Text>
+    </div>
```
alexander-meyer, line 774: Based on the changes you made, I believe this would now be captured as a React node and provided as the `children` argument to the above `<DataTable>` element.

kyleherrmann: I believe children is a reserved word and are typically passed between the opening and closing tags of the component no? 🤔

chris-at-jobber:
@kyleherrmann as I mentioned on Slack, removed some CSS and made some adjustments such that I don't believe the `height` prop on the `Body` needed to be passed anymore - also added some usage guidance around empty states to the docs. 

@alexander-meyer are you [thinking](https://github.com/GetJobber/atlantis/pull/973#discussion_r1071637651) of something like making `emptyState` a prop on the `DataTable` and then users of this component would pass their "empty state component" in? That makes sense to me.

alexander-meyer (approved):
@chris-at-jobber @kyleherrmann my apologies, I had misunderstood what the intended direction was. Chris, that was indeed what I had in mind, but I'm certainly on board with the approach of simply passing in the desired empty state as a child element to the `<DataTable>` instance.

chris-at-jobber:
FYI I've tagged the engineers from my team to take a look at the failing CSS linter when they're in tomorrow :x:

kyleherrmann:
> Design approved, @MichaelParadis is going to do a code review on the Atlantis side! Thanks @kyleherrmann

Great, thank you for the reviews all! Would it be possible to get a rough timeline on when this code may be merged? It is our end of sprint today and I am just reviewing remaining subtasks.

MichaelParadis (changes requested):
Minor comments, although I noticed tests are missing for this change. Would you be able to add some @kyleherrmann?

packages/components/src/DataTable/DataTable.tsx
```diff
@@ -60,6 +60,11 @@ export interface DataTableProps<T> {
    * Enables row click action. The provided callback will be executed when the row is clicked.
    */
   onRowClick?: (row: Row<T>) => void;
+
+  /**
+   * The elements to display when the data table is empty
+   */
+  children?: ReactNode | ReactNode[];
```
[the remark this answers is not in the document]
MichaelParadis: See parent review. I just saw the other comments

packages/components/src/DataTable/Body.tsx
```diff
@@ -1,14 +1,20 @@
...
 import classNames from "classnames";
-import React, { useCallback } from "react";
+import React, { ReactNode, useCallback } from "react";
 import styles from "./DataTable.css";
 
 interface BodyProps<T> {
   table: Table<T>;
   onRowClick?: (row: Row<T>) => void;
+  height?: number;
+  children?: ReactNode | ReactNode[];
```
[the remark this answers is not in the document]
MichaelParadis: See parent review. I just saw the other comments

kyleherrmann:
> Minor comments however, not too sure because this was discussed before so I will be fine to resolve them. I noticed tests are missing for this change. Would you be able to add some @kyleherrmann?

Do you recommend any tests apart from ensuring the provided elements are rendered when there is no data?

MichaelParadis (approved):
LGTM!

PR #989, opened 2023-01-20
refactor(components): Adjust banner alignment

rodrigoeidelvein:
## Changes

- Changed banner alignment

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Banner/Banner.css
+++ b/packages/components/src/Banner/Banner.css
@@ -10,7 +10,7 @@
   padding: var(--flash--inset);
   color: var(--color-black);
   background-color: var(--color-grey--lightest);
-  align-items: flex-start;
+  align-items: center;
 }
 
 .bannerContent {
--- a/packages/components/src/Banner/Banner.css
+++ b/packages/components/src/Banner/Banner.css
@@ -21,6 +21,10 @@
   align-items: flex-start;
 }
 
+.bannerChildren {
+  align-self: center;
+}
+
 .dismissibleSpacing {
   padding-top: var(--space-smallest);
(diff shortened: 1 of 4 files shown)


packages/components/src/Banner/Banner.css
```diff
@@ -10,7 +10,7 @@
   padding: var(--flash--inset);
   color: var(--color-black);
   background-color: var(--color-grey--lightest);
-  align-items: flex-start;
+  align-items: center;
```
chris-at-jobber, line 13: @rodrigoeidelvein this breaks the alignment from what we expect / what was previously there

You can see this pushes the dismiss button to the center of the banner:
image

It should remain aligned to the top:
image

chris-at-jobber (changes requested):
Needs to solve the problem without breaking existing reflow behaviour

chris-at-jobber:
Wide looks good ✅ 
image

Narrow looks good ✅ 
image
 
There's this intermediate break where I think the styling for a "narrow" banner is added, but the content is not long enough to wrap:
image

I wouldn't call this a blocker since we already see this behaviour (see example), but wanted to flag it.
image

chris-at-jobber (approved):
The changes look good to me. We'll definitely want to test this one "in the wild" with more complex usages, but this seems like a reasonable way to encapsulate the alignment into separate containers. It looks like you may have inadvertently also resolved an issue where custom child content could break!

### Before
Bullet list reflows on same line as paragraph, not good:
image

### After
Bullet list is beneath paragraph, as expected!
image

rodrigoeidelvein:
@chris-at-jobber that was all calculated 🤣

PR #992, opened 2023-01-21
refactor(components): Update typography for Button

taylorvnoj:
## Motivations
Match new style for Button component because the typography has been updated

## Changes
Before:
image

After:
image

### Changed
- updated casing, spacing and font-family of typography within the button
- snapshot tests that use button have been updated

### Testing
- check Button component locally and compare with Figma
- can check casing by updating playground
---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -16,8 +16,8 @@
   border: var(--border-base) solid var(--button--color-variation);
   border-radius: var(--radius-base);
   color: var(--button--color-variation);
-  fill: var(--button--color-variation);
   -webkit-font-smoothing: antialiased;
+  letter-spacing: 0.2px;
   text-align: center;
   text-decoration: none;
   vertical-align: middle;
@@ -25,13 +25,15 @@
   cursor: pointer;
   transition: all var(--timing-base);
   appearance: none;
+  fill: var(--button--color-variation);
   align-items: center;
   justify-content: center;
 }
 
 .button * {
   color: inherit;
-  fill: inherit !important; /* required to over-ride very specific fill from <Icon> */
+  fill: inherit !important;
+  /* required to over-ride very specific fill from <Icon> */
 }
 
 .button:hover,
@@ -263,6 +265,7 @@ a.button.disabled:hover {
   0% {
     background-position: 0 0;
   }
+
   100% {
     background-position: var(--space-larger) var(--space-larger);
   }
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -17,7 +17,7 @@
   border-radius: var(--radius-base);
   color: var(--button--color-variation);
   -webkit-font-smoothing: antialiased;
-  letter-spacing: 0.2px;
+  letter-spacing: var(--typography--letterSpacing-loose);
   text-align: center;
   text-decoration: none;
   vertical-align: middle;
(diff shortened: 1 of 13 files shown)


packages/components/src/Button/Button.css
```diff
@@ -16,22 +16,24 @@
   border: var(--border-base) solid var(--button--color-variation);
   border-radius: var(--radius-base);
   color: var(--button--color-variation);
-  fill: var(--button--color-variation);
   -webkit-font-smoothing: antialiased;
+  letter-spacing: 0.2px;
```
chris-at-jobber, line 20: We should use the value from the [design package](https://github.com/GetJobber/atlantis/blob/15f604cc51ee6881a5ed9c1be01081c76cf5ce6e/packages/design/src/typography.css#L3) here
```suggestion
  letter-spacing: var(--typography--letterSpacing-loose);
```

packages/components/src/Button/Button.css
```diff
@@ -16,22 +16,24 @@
...
   text-align: center;
   text-decoration: none;
   vertical-align: middle;
   background-color: var(--button--color-surface);
   cursor: pointer;
   transition: all var(--timing-base);
   appearance: none;
+  fill: var(--button--color-variation);
```
chris-at-jobber, line 28: Did this move itself on save (auto-formattting?), or was this a specific decision?

packages/components/src/Button/Button.tsx
```diff
@@ -162,8 +162,9 @@ function ButtonInternals({ label, icon, size = "base" }: ButtonProps) {
       {icon && <Icon name={icon} size={size} />}
       <Typography
         element="span"
-        textCase="uppercase"
+        textCase="capitalize"
```
chris-at-jobber, line 165: I know I wrote the ticket but thinking about we shouldn't actually use `capitalize` here. I'd like to follow the same guidelines we have for [mobile Button](https://atlantis-mobile.jobber.dev/?path=/docs/atlantis-actions-button--button) around "proper" title-casing – it does add more work on the implementation-in-product side as we'll have a bunch of labels to tweak, but we should be able to catch those instances fairly easily.

chris-at-jobber (changes requested):
One change to a variable used, and one larger change!

packages/components/src/Button/Button.css
taylorvnoj: @chris-at-jobber yes, this moved itself. I noticed there's quite a few things my vsCode changes which go against the Atlantis linter, so I'll be making sure those settings are adjusted

packages/components/src/Button/Button.mdx
```diff
@@ -129,9 +129,9 @@ Destructive actions remove something from Jobber.
 #### Primary
 
 A Primary destructive action will permanently destroy an object carrying
-signficant data, such as a job, quote, or client. This destructive action should
-be contained in a [ConfirmationModal](/components/confirmation-modal) triggered
-by clicking a secondary or tertiary destructive button.
+significant data, such as a job, quote, or client. This destructive action
```
taylorvnoj, line 132: This was just to fix a typo (spelling of `significant`)

packages/components/src/Button/Button.tsx
taylorvnoj: Good call out - I removed the `textCase` variable and updated the `.mdx` file to match the guidelines we have for mobile.  Like we discussed, once we are updating the layouts in the future ticket, we will look at `<Button>` components in JO, Dev Center, etc. I don't mind doing that when the time comes 👍

chris-at-jobber (approved):
Looks great, thanks @taylorvnoj !

PR #1007, opened 2023-01-26
refactor(components): Remove letterSpacing-loose letter spacing from Buttons

taylorvnoj:
## Motivations
There should be no letter-spacing on the small buttons

### Changed
Before:
Screenshot 2023-01-26 at 2 05 59 PM

After:
Screenshot 2023-01-26 at 4 02 03 PM

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -17,6 +17,7 @@
   border-radius: var(--radius-base);
   color: var(--button--color-variation);
   -webkit-font-smoothing: antialiased;
+  letter-spacing: var(--typography--letterSpacing-loose);
   text-align: center;
   text-decoration: none;
   vertical-align: middle;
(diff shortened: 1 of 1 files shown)


chris-at-jobber (changes requested):
This removes the letter-spacing from all buttons - the change should only apply to `small` size buttons

packages/components/src/Button/Button.css
```diff
@@ -177,11 +176,13 @@ a.button.disabled:hover {
 .base {
   min-height: calc(var(--base-unit) * 2.25);
   padding: 0 calc(var(--space-base) - 4px);
+  letter-spacing: var(--typography--letterSpacing-loose);
```
chris-at-jobber, line 179: I think we can reduce the amount of CSS here if we leave the `letterSpacing` declaration as-is on line 20 and set `letter-spacing: 0` on the small button class on line 173

chris-at-jobber (approved):
🎉

PR #1018, opened 2023-02-01
refactor(design): Cleanup function to export SCSS tokens

rodrigoeidelvein:
## Motivations
Refactor the function that exports the design tokens to SCSS.

## Changes
- Just clean up and refactor.

## Testing
You can build the file by running:
1. `cd packages/design`
2. `npm run build:foundation`

The exported tokens should be the same as `master`

---

Random photo of Atlantis

### DIFF
--- a/packages/design/jobberStyle.js
+++ b/packages/design/jobberStyle.js
@@ -156,33 +156,7 @@ function getResolvedSCSSVariables(cssProperties) {
   const allKeys = Object.keys(cssProperties);
 
   return allKeys.reduce((acc, cssVar) => {
-    const customPropertyValue = customProperties["--" + cssVar];
-    const variableType = getVariableType(cssVar);
-    let propertyValue = "";
-
-    switch (variableType) {
-      case "simple":
-        propertyValue = `${resolvedCssVars[cssVar]}`;
-        break;
-      case "calc":
-        {
-          const calcRegexResult =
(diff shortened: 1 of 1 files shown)


packages/design/jobberStyle.js
```diff
@@ -154,41 +154,41 @@ function getResolvedCSSVars(cssProperties) {
...
-      );
-
-      return [...acc, `$${cssVar}: ${resolvedShadowProperty};`];
-    } else {
-      return acc;
+    const customPropertyValue = customProperties["--" + cssVar];
+    const variableType = getVariableType(cssVar);
+    let propertyValue = "";
+
+    switch (variableType) {
```
darryltec, line 163: thoughts on breaking this `switch` out into its own function? Separate those concerns of reducing and figuring out what value?

```js
const propertyValue = getPropertyValue();
if (propertyValue) {
  return [...acc, `$${cssVar}: ${propertyValue};`];
}

return acc;
```

IMO, it makes the reducer simpler and straight to the point. if there's a property value, add it; otherwise, don't. That way you won't have to reassign a variable. Up to you tho

rodrigoeidelvein: I liked this idea, I pushed the changes related to your comment. Thanks!

darryltec (approved):
Can confirm, no difference in what gets outputted. Nice cleanup!

PR #1022, opened 2023-02-03
fix(components): Remove default value for prop size in ConfirmationModal

igor-m-jobber:
## Motivations
In the [PR](https://github.com/GetJobber/atlantis/pull/969) that added additional functionality to the `ConfirmationModal` we assign the default `size` property to the `small` value. This led to only 2 sizes being available for the modal size: `small` and `large`. 

We want to ensure that if `size` property is not supplied, none of the styles for the size modal will be applied, meaning that it will be the default size.

## Changes
Remove the default value for the `size` prop in the `ConfirmationModal` component.

### Added
n/a

### Changed
The `small` size prop was a default value. Now, `size` doesn't have default value.

### Deprecated
n/a
### Removed
n/a
- 

### Fixed
n/a
- 

### Security
n/a
- 

## Testing
Go to the Confirmation modal and open it. Should open the modal.
If you provide `small` or `large` size value - the size of the modal should change correspondingly.  

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
+++ b/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
@@ -175,7 +175,7 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
     onCancel,
     onRequestClose,
     variation = "work",
-    size = "small",
+    size = undefined,
     children,
   }: ConfirmationModalProps,
   ref: Ref<ConfirmationModalRef>,
--- a/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
+++ b/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
@@ -175,7 +175,7 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
     onCancel,
     onRequestClose,
     variation = "work",
-    size = undefined,
+    size,
     children,
   }: ConfirmationModalProps,
   ref: Ref<ConfirmationModalRef>,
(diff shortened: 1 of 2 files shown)


MichaelParadis (approved):
Code looks good. Going to wait on @chris-at-jobber 's input before merged though

chris-at-jobber:
@MichaelParadis this looks alright, but this seems like a sign that we should add an explicit `base` size property to `Modal`. The issue here seems less that the `ConfirmationModal` defaults to `small` than that there's no way to use the standard modal size if desired.

MichaelParadis:
@chris-at-jobber it looks like the reason for Modal not having a base size is by default it appears to default to 100% width to fit the content while small and large limits the size. I could be wrong though

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Modal/Sizes.css#L1

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Modal/Modal.css#L33-L45

chris-at-jobber:
Modal has a max-size at default: 

Defined [here](https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Modal/Modal.css#L39) as `max-width: --modal-width`  

And the base value is set [here](https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Modal/Modal.css#L2) as `--modal--width: calc(var(--base-unit) * 37.5);`

igor-m-jobber:
Hey @MichaelParadis @chris-at-jobber. 
Wondering if there is anything that blocks this PR from being merged into the master.

Asking because there is a good chance that Comms_experience will be working on the ticket that will require the correct modal size to be used in the modal.

Thank you for the help! : )

MichaelParadis:
@igor-m-jobber we are going to address the modal size thing but I will merge this in right now

igor-m-jobber:
@MichaelParadis yeah. I saw your conversation and I think it's a great idea.
Thank you for merging this PR!

PR #1023, opened 2023-02-03
fix(components): JOB-61635 InputNumber on iOS not allowing decimals

MichaelParadis:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Fixed in ability to add decimals on iOS for mobile because the inputMode wasn't set https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

### Security

- 

## Testing

- Open on iOS, set `keyboard` to `decimal`. Verify that it is possible to see the decimal on the keyboard

---
IMG_93A4C4444CF0-1

Random photo of Atlantis


packages/components/src/InputNumber/InputNumber.tsx
```diff
@@ -18,6 +18,11 @@ interface InputNumberProps
       | "defaultValue"
     > {
   value?: number;
+
+  /**
+   * Allow keyboards for mobile to display decimal option
+   */
+  allowDecimals?: boolean;
```
chris-at-jobber, line 25: Would allowing this component to `Pick` the keyboard prop with a default of `numeric` have the same effect?

MichaelParadis: Actually, I think that may would be better. Let me update things

chris-at-jobber:
I'm able to test that the correct keyboards are called when the value is changed.

One note (possibly for future investigation as this can be replicated on production): When I have typed a `.` in the InputNumber, the minilabel collapses down into overlapping with the entered value: 

https://user-images.githubusercontent.com/39704901/216725243-db1fc116-f64e-4d87-8db1-2991dadceed1.mov

dhaberst:
@MichaelParadis @chris-at-jobber This lgtm!

PR #1025, opened 2023-02-07
refactor(components): replace base colors with semantic colors

chris-at-jobber:
## Motivations

Our components should use semantic values whenever possible to ensure they benefit from systemic visual updates. While we've made efforts on this front in the past, this PR catches some that slipped through the gaps as the colors were defined in the component properties rather than in CSS.

### Changed

- replaced "base" palette values with semantic colors in various instances of Typography

## Testing

- inspect the colors in the components
- you can go into the root values in the inspector and adjust the values for `--color-text` and/or `--color-text--secondary` you should observe the components will now update accordingly.

---

Random photo of Atlantis


rodrigoeidelvein (approved):
LGTM!

PR #1026, opened 2023-02-08
feat(components): Add Customizable Header Prop to Atlantis Web Card Component

frankiepoon:
## Motivations

We need to be able to support a customizable header in order to add button(s) to the [`Card`](https://atlantis.getjobber.com/components/card) component. We are taking a similar approach to [`Card`](https://atlantis-mobile.jobber.dev/?path=/docs/atlantis-layouts-and-containers-card--card) component in `jobber-mobile`, which allows `HeaderProp`. 

### Added
- add a new `HeaderProps` type that allows for the title to be set along with a single `Button` or an entirely custom React component

### Changed

- update tests and documentation

### Deprecated

- deprecate the `title` prop

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- [Add `header` prop to `Card` to create single button ](http://localhost.test:3333/components/card#header-with-button)
- [Add `header` prop to `Card` to render custom component](http://localhost.test:3333/components/card#header-with-custom-component)
- no changes to existing functionality (e.g. deprecated `title` should still render a header title)

https://user-images.githubusercontent.com/94410044/217611014-5776277e-430c-41c4-a12c-44d78188b037.mov

---

Random photo of Atlantis


packages/components/src/Card/Card.tsx
```diff
@@ -46,13 +45,12 @@ interface HeaderActionProps {
...
-  readonly label: string;
-  readonly size?: "small" | "base" | "large";
-  readonly type?: "primary" | "secondary" | "tertiary";
-  readonly icon?: IconNames;
-  readonly onClick?: () => void;
-}
+
+type ButtonAction = Pick<
+  ButtonProps,
+  "label" | "size" | "type" | "icon" | "onClick"
```
chris-at-jobber, line 51: Is there any reason not to allow all Button props here? `variation` in particular, but also options like `external`,  `disabled`, the full suite of `aria{Property}`, etc.

packages/components/src/Card/Card.tsx
```diff
@@ -53,11 +85,31 @@ export function Card({
...
+
   const cardContent = (
     <>
-      {title && (
+      {showCustomCompHeader && header?.customHeader}
+
+      {!showCustomCompHeader && showCardHeader && (
         <div className={styles.header}>
-          <Heading level={3}>{title}</Heading>
+          <div>
```
chris-at-jobber, line 99: What is this `div` doing?

packages/components/src/Card/Card.css
```diff
@@ -56,6 +56,12 @@
 }
 
 .header {
+  display: flex;
+  flex-direction: row;
   padding: var(--card--base-padding);
   background-color: var(--color-surface--background);
 }
+
+.headerButton {
```
chris-at-jobber, line 65: Minor but I think this should be renamed to `headerAction` to match the prop names

packages/components/src/Card/Card.css
```diff
@@ -56,6 +56,12 @@
...
+  margin-left: auto;
```
chris-at-jobber, line 66: Does this class actually need to exist if this is all it's doing? It seems like we could just add an equivalent `gap` to the `header` class, no?

packages/components/src/Card/Card.css
```diff
@@ -56,6 +56,12 @@
 }
 
 .header {
+  display: flex;
+  flex-direction: row;
```
chris-at-jobber, line 60: Do we need to specify a flex-direction? Web flex-direction's [initial value should be row](https://jobber.atlassian.net/browse/JOB-61896).

packages/components/src/Card/Card.tsx
```diff
@@ -53,11 +85,31 @@ export function Card({
...
+            <Heading level={3}>{header?.title || title}</Heading>
+          </div>
+          {header?.action?.label && (
+            <div className={styles.headerButton}>
```
chris-at-jobber, line 103: Related to the CSS, I think we should see if we can remove the need for this div + its' associated class.

chris-at-jobber (commented):
Some initial thoughts from a design/usage perspective! I'll tag the engineers from the team to have a look as well.

packages/components/src/Card/Card.tsx
frankiepoon: I'm also leaning towards that idea, however am uncertain if this would be better (friendlier) to make `Card` take smaller set of parameters

packages/components/src/Card/Card.tsx
frankiepoon: this needs a cleanup 👍

packages/components/src/Card/Card.css
frankiepoon: the `margin-left: auto` was really meant for right aligning the button, which is now better handled by `header` class with `justify-content: space-between`, and so the `headerButton` class can be removed.

chris-at-jobber: We'll still need the gap, as the title can currently run right into the action before wrapping: 
image

chris-at-jobber:
I'm going to take this question back to the design team, but I don't think this is the behaviour we want when the Button size changes.

image

packages/components/src/Card/Card.tsx
```diff
@@ -4,23 +4,49 @@ import { XOR } from "ts-xor";
...
   /**
    * The `accent`, if provided, will effect the color accent at the top of
    * the card.
    */
   readonly accent?: keyof typeof colors;
   readonly children: ReactNode | ReactNode[];
+}
+
+interface CardHeaderProps {
+  readonly title?: never;
```
davidr64, line 21: I believe it is not necessary to add `title:Never` as the XOR utility type handles that for us. Personally it made it more confusing for me that title was defined here.

packages/components/src/Card/Card.tsx
```diff
@@ -4,23 +4,49 @@ import { XOR } from "ts-xor";
...
+}
 
+interface CardTitleProps {
   /**
-   * The title of the card.
+   * @deprecated Use header instead
    */
   readonly title?: string;
+
+  readonly header?: never;
```
davidr64, line 35: Similar I don't believe this is adding anything to our type system. Please correct me if I am wrong though.

packages/components/src/Card/Card.css
frankiepoon: As i was adding `gap`, i realized that the wrapping would vertically stretch out the button. To counteract this, I added `align-content: top`, which raises the question as to whether this should be another prop. This is likely something to consider along with your comments about the size of buttons.  
image

packages/components/src/Card/Card.tsx
```diff
@@ -4,23 +4,49 @@ import { XOR } from "ts-xor";
...
+}
+
+type CardProps = CardCommonProps & XOR<CardHeaderProps, CardTitleProps>;
```
davidr64, line 38: 2 things are confusing me while reading the types. I think they are pre-existing issues that extra types have made more confusing.

1. The order of the types. Normally I like defining lower level interfaces first and use them below. In here we seem to have a mix of defining an interface using other interfaces that are defined both above and below. It would be nice if we could do one or the other. 
2. The names are a little tricky to follow, though I am not sure how much can be done about that.

davidr64: Actually, on the names it might help to reduce the number of named types. If we don't need to export the `HeaderProps` which I don't think we do, you could not create it and instead assign header directly.

```
interface CardHeaderProps {
  readonly title?: never;

  /**
   * The header props of the card.
   */
  readonly header?: XOR<HeaderActionProps, HeaderCustomProps>;
}
```

packages/components/src/Card/Card.tsx
```diff
@@ -53,11 +80,19 @@ export function Card({
...
 
+  const { showCustomCompHeader, showCardHeader } = checkHeaderVisibility(
+    title,
+    header,
+  );
+
   const cardContent = (
     <>
-      {title && (
+      {showCustomCompHeader && header?.customHeader}
```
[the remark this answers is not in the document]
davidr64: That might be a designer / atlantis question

davidr64 (changes requested):
Some suggestions and questions that we might want to forward to design unless you have already checked.

packages/components/src/Card/Card.tsx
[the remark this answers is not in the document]
davidr64: @chris-at-jobber do we want any styling on the custom headers or would we prefer to leave that to whoever is using the custom header?

packages/components/src/Card/Card.tsx
frankiepoon: image

davidr64: Sorry this comment was not about the `deprecated`  tag it was about defining `header: never`

packages/components/src/Card/Card.tsx
frankiepoon: the `HeaderProps` type is used in `checkHeaderVisibility(...)`, but the `export` is not required

davidr64 (approved):
There is an outstanding design question but code wise I am good with how it works now. I could see that design question going either way.

packages/components/src/Card/Card.tsx
```diff
@@ -4,23 +4,46 @@ import { XOR } from "ts-xor";
...
+}
+
+interface HeaderActionProps {
+  readonly title?: string;
+  readonly action?: ButtonProps;
+}
+
+interface HeaderCustomProps {
+  readonly customHeader?: React.ReactNode;
+}
```
davidr64, line 26: We probably want descriptions on these fields

davidr64 (changes requested):
Sorry really minor thing as I was reflecting on it.

packages/components/src/Card/Card.tsx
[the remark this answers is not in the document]
chris-at-jobber: I think I would leave custom headers up to the consumer

davidr64 (approved):

packages/components/src/Card/Card.mdx
```diff
@@ -112,3 +113,62 @@ Card titles should be sentence-cased.
...
+  </Card>
+</Playground>
+
+### Header with custom component
+
+<Playground>
+  <Card
+    header={{
+      customHeader: <Banner type="notice">Bananas are on sale!</Banner>,
+    }}
```
darryltec, line 157: We should skip the object and go straight to

```ts
    header={<Banner type="notice">Bananas are on sale!</Banner>}
```

if it's a custom component. This also removes the need for an XOR since TS (and JS) can differentiate between an object or a function/component `HeaderActionProps | ReactNode`

Also, we shouldn't use banner unless the designer wants that to happen. Is that the intended use for it?

packages/components/src/Card/Card.tsx
```diff
@@ -4,23 +4,55 @@ import { XOR } from "ts-xor";
...
+
+interface CardTitleProps {
+  /**
+   * @deprecated Use header instead
+   */
+  readonly title?: string;
+}
+
+type CardProps = CardCommonProps & XOR<CardHeaderProps, CardTitleProps>;
```
darryltec, line 53: Instead of doing all of this and XOR, have we thought about 

```ts
interface CardProps {
  ...
   /**
    * The title of the card.
    * 
    * @deprecated
    */
  readonly title?: string;

   /**
    * The header props of the card.
    */
  readonly header: string | HeaderProps | ReactNode
  /
}
```

that way it's always either one of those but not all of them. Don't need the XOR. On the code side, a classic `typeof` should do the trick.

Also simplifies the usage when you don't really need any of the advanced stuff and just wants a title.


```ts
<Card header="Bananas" />
<Card header="{ title: "Bananas", action: {...} }" />
<Card header="<MyBananaHeader />" />
```

darryltec (changes requested):

packages/components/src/Card/Card.mdx
chris-at-jobber: ➕ to a more realistic example. I can see something like a `Heading level={4}` and a `subtle` Button with a cross Icon and no grey background being something that gets used for cases like inline messages.

frankiepoon: @darryltec  not a specific design we intend to have, I don't have a concrete example using custom component other a banner or adding multiple action buttons. @chris-at-jobber does it work with this sample 
image

chris-at-jobber: That looks good, if we could make the button `tertiary` instead so there's no border that would look ideal 👌

chris-at-jobber:
Design update: having some discussions that, pending the outcome, could impact the interface for either the `action`, or the CSS for the header. Will keep this PR updated as that moves along!

packages/components/src/Card/tests/Card.test.tsx
```diff
@@ -50,6 +51,82 @@ it("renders a card", () => {
   `);
 });
 
+it("renders a card with button", () => {
```
darryltec, line 54: Almost forgot about this, is there a way to avoid the snapshot? I know we do that on other tests but those tests are also 3-4 years old. We've learned our lesson there to not use snapshots. 😁

packages/components/src/Card/tests/Card.test.tsx
```diff
@@ -50,6 +51,82 @@ it("renders a card", () => {
...
+  const tree = renderer
```
darryltec, line 55: Almost forgot about this, is there a way to avoid the snapshot? I know we do that on other tests but those tests are also 3-4 years old. We've learned our lesson there to not use snapshots. 😁

frankiepoon:
@JordanLos and I added some changes to refactor the header rendering logic, while further design updates are expected to be around the action props.

chris-at-jobber:
Update from discussions with design:

### Action props interface

We will introduce _some_ restrictions to the properties of Button being passed through, and encourage teams to use `CustomHeader` for all other usages. 

There's no appetite from the team to expose `size` or `fullWidth` as 
- `size` and `fullWidth` introduce myriad new design concerns as the header will align differently with each size, and there are currently no use cases the team has that would merit the additional complexity
- we could end up pursuing a card-specific "action" UI that doesn't leverage these properties

We will keep the `type` property exposed as there are existing examples using all three "types" of button mocked up already.

### Visual alignment

We'll keep the action aligned to the top, with the first line of the header.

image

frankiepoon:
> Update from discussions with design:
> 
> ### Action props interface
> We will introduce _some_ restrictions to the properties of Button being passed through, and encourage teams to use `CustomHeader` for all other usages.
> 
> There's no appetite from the team to expose `size` or `fullWidth` as
> 
> * `size` and `fullWidth` introduce myriad new design concerns as the header will align differently with each size, and there are currently no use cases the team has that would merit the additional complexity
> * we could end up pursuing a card-specific "action" UI that doesn't leverage these properties
> 
> We will keep the `type` property exposed as there are existing examples using all three "types" of button mocked up already.
> 
> ### Visual alignment
> We'll keep the action aligned to the top, with the first line of the header.
> 
> image

@chris-at-jobber are `size` and `fullWidth` the only two fields not to be exposed ? do we expect default values on them ?
For `online booking`, we are looking to have `icon`, `label`, `type`, while other fields are not relevant. 

On the alignment, the screenshot seems to be `h3` and a `button` (`size="small"`), but now that `size` is not exposed, the default Button size which is `base` would cause some height misalignment like below. Just want to make sure button `size` can be overwritten as `small` in `Card` ? 

-button (default size = `base`)

image

-button (overwritten as size=`small`)
image

chris-at-jobber:
That's correct, the action size should be set internal to the Card to always be `small`, and `fullWidth` should be fine to leave out as the default is false. 

The rest of the props are likely to be needed for any "button-esque" interactive element (`onClick`, to/url/ external, `loading`, `disabled`, etc) or offer accessibility features that may be needed.

davidr64 (changes requested):
Some minor requested changes. I would like @darryltec or some other confirmation that the move away from snapshots is good. I would understand if we wanted to stick with snapshots for design based components like this.

packages/components/src/Card/Card.tsx
```diff
@@ -55,15 +85,45 @@ export function Card({
...
+        </div>
+      );
+    }
+    // Case 3: Custom Header
+    if (React.isValidElement(header)) {
+      return header;
+    }
+    // Case 4: Default Header Props
+    return (
+      !!header && (
```
davidr64, line 116: Is the `!!` required. I am just asking for my own learning

packages/components/src/Card/Card.mdx
```diff
@@ -112,3 +114,72 @@ Card titles should be sentence-cased.
...
+        style={{
+          display: "flex",
+          justifyContent: "space-between",
+          padding: "var(--card--base-padding)",
+          gap: "var(--space-base)",
+          alignItems: "flex-start",
+        }}
+      >
+        <Heading level={4}>Bananas varieties</Heading>
+        <Button onClick={() => alert("Dismiss")} icon="cross" type="teritary" />
```
chris-at-jobber, line 166:
```suggestion
        <Button onClick={() => alert("Dismiss")} icon="cross" type="tertiary" />
```

Just a typo on the prop name

packages/components/src/Card/Card.mdx
```diff
@@ -112,3 +114,72 @@ Card titles should be sentence-cased.
...
+      <Text>
+        This is the familiar yellow banana sold in United States’ supermarkets.
+        There are also other sizes of Cavendish, including Dwarf and Giant,
+        though they are difficult to find.
+      </Text>
+    </Content>
+  </Card>
+</Playground>
+
+### Header with custom component
```
chris-at-jobber, line 151: Can we add some guidance around when to use a custom component? Something like "Use the provided header title and action properties when possible. If you require UI elements not supported by the card's header, a custom component can be passed in."

packages/components/src/Card/Card.mdx
```diff
@@ -112,3 +114,72 @@ Card titles should be sentence-cased.
     </Content>
   </Card>
 </Playground>
+
+### Header with button
```
chris-at-jobber, line 118:
```suggestion
### Header with action

Use the header action when there's an action tied directly to the Card's contents, such as adding more items. It comes with pre-defined styling and layout to standardize how Card actions look across the product.
```

packages/components/src/Card/Card.mdx
```diff
@@ -112,3 +114,72 @@ Card titles should be sentence-cased.
...
+
+<Playground>
+  <Card
+    header={{
+      title: "Banana varieties",
+      action: {
+        label: "Buy",
```
chris-at-jobber, line 125:
```suggestion
        label: "New Banana",
```

packages/components/src/Card/Card.mdx
```diff
@@ -112,3 +114,72 @@ Card titles should be sentence-cased.
...
+        type: "primary",
+        icon: "add",
+        onClick: () => {
+          alert("Buy Bananas!");
```
chris-at-jobber, line 129:
```suggestion
          alert("New banana added");
```

packages/components/src/Card/Card.tsx
```diff
@@ -55,15 +85,45 @@ export function Card({
...
+        <div className={styles.header}>
+          {header?.title && <Heading level={3}>{header?.title}</Heading>}
+          {header?.action && (
+            <Button {...(header?.action as ButtonProps)} size="small" />
```
chris-at-jobber, line 120: Is it possible to _default_ the type to `tertiary`? I'd prefer not to make `primary` the default treatment as you could quickly build out a pretty overwhelming view with >1 cards with primary actions.

packages/components/src/Card/Card.tsx
```diff
@@ -86,25 +83,45 @@ export function Card({
...
-          <Heading level={3}>{header?.title || title}</Heading>
-          {header?.action?.label && <Button {...header.action} />}
-        </div>
-      )}
+      {CardHeader && CardHeader()}
       {children}
     </>
   );
 
+  function CardHeader() {
```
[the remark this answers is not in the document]
darryltec: I would just move this to its own component. Separate concerns and would probably clean it up as well

darryltec:
> I would like @darryltec or some other confirmation that the move away from snapshots is good. 

Great question. Multiple points there. 

- First off, moving away from snapshots is good! The idea behind snapshots is great but it requires a lot of human interaction, honor system, and overall understanding if it actually broke something. 
  - Chances are, if you're updating something as old as this, you wouldn't have the same context as to why that broke without finding the person that wrote it. Compared to it being a real assertion test that tells you what exactly is it testing.
- Atlantis (repo and team) internally swore off snapshots for years we just haven't had the time to clean up.
- We did talk about maybe being okay with a snapshot in this repo but we leaned toward "It's either we all have it or we don't". We don't want the "_rules for thee but not for me_" type thing.

> I would understand if we wanted to stick with snapshots for design based components like this.

True, but there's a much better way to write snapshot tests other than jest blob of generated code since it creates more problems and work for us. I can confidently say that we've never said the words "Thank god there were snapshots" since it's just autogenerated and it's so easy to miss changes that are actually breaking.

We're also spiking out implementing [chromatic](https://www.chromatic.com/) for a real, actual image, snapshot testing that would catch problems visually.

If, at some point, a bug does come up. Then we write more test. Rinse and repeat until we have tests that are hella strong. Typescript also helps us from making silly mistakes so that makes removes some FUD.

darryltec:
@frankiepoon merge master, please! There's a pretty huge change that happened recently. Thank you!

packages/components/src/Card/Card.tsx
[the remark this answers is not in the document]
frankiepoon: thanks both, updating as its own component looks cleaner to me.

packages/components/src/Card/Card.tsx
```diff
@@ -86,25 +83,47 @@ export function Card({
...
+    // Case 2: String Header
+    if (typeof header === "string") {
+      return (
+        <div className={styles.header}>
+          {header && <Heading level={3}>{header}</Heading>}
+        </div>
+      );
+    }
+    // Case 3: Custom Header
+    if (React.isValidElement(header)) {
```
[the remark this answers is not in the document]
frankiepoon: credit to @darryltec 👍

packages/components/src/Card/Card.tsx
frankiepoon: it's not required here, I"m removing it to keep consistent with the rest

packages/components/src/Card/Card.tsx
frankiepoon: added default Button type to be `tertiary` if it's not specified in the header action props. Not sure if we need to mention this level of details in the documentation.

davidr64 (approved):

davidr64:
@darryltec can we clear your requested changes?

darryltec:
@davidr64 Planning to review it tomorrow AM as well. So we'll keep that there for now.

packages/components/src/Card/Card.tsx
```diff
@@ -1,26 +1,57 @@
...
 
+interface HeaderActionProps {
   /**
    * The title of the card.
    */
   readonly title?: string;
+  /**
+   * The action props that renders into a button on the card header.
+   */
+  readonly action?: Omit<ButtonProps, "size" | "fullWidth">;
```
darryltec, line 27: I think we should change this to be a component instead of always a `<Button />`. Since (based on current design vision and usage) the action is either a `<Button />` or `<Menu />`.

Colin_Blue_-_Jobber

@chris-at-jobber what do you think?

Also prevents us, Atlantis, from having to deprecate this, do a breaking change, update all usage in Jobber Online, mass QA, and release coordination. Makes it easy to extend it too if we do decide to support others like Switch. #futureProofing

---

As for how, you can do it like [InternalChipAffix](https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Chips/InternalChipAffix.tsx)

https://github.com/GetJobber/atlantis/blob/f6dbeb11a162ad2a88556a116b07e4b3d974c972/packages/components/src/Chips/InternalChipAffix.tsx#L11

Setting different defaults can be done in this area
https://github.com/GetJobber/atlantis/blob/f6dbeb11a162ad2a88556a116b07e4b3d974c972/packages/components/src/Chips/InternalChipAffix.tsx#L21-L31

packages/components/src/Card/Card.tsx
```diff
@@ -1,26 +1,57 @@
...
+  /**
+   * The `accent`, if provided, will effect the color accent at the top of
+   * the card.
+   */
+  readonly accent?: keyof typeof colors;
+  readonly children: ReactNode | ReactNode[];
+  /**
+   * @deprecated
+   * The title of the card. Use header instead.
+   *
```
darryltec, line 42: get it straight to the point

```suggestion
   * @deprecated - Use header instead.
```

packages/components/src/Card/Card.mdx
```diff
@@ -112,3 +114,80 @@ Card titles should be sentence-cased.
...
+contents, such as adding more items. It comes with pre-defined styling and
+layout to standardize how Card actions look across the product.
+
+<Playground>
+  <Card
+    header={{
+      title: "Banana varieties",
+      action: {
+        label: "New Banana",
+        type: "primary",
```
darryltec, line 130: I'd remove this. make sure the copy-paster doesn't think that you need to have this.

darryltec (changes requested):
Looking really good! Just little bits and pieces to think about. We can talk about this in-depth if you want.

Also, not sure if this is just my local but I can't run this locally. I get stuck with an endless loop of 
image

Does that happen to you?

packages/components/src/Card/Card.tsx
chris-at-jobber: That does make sense to me - my concern is Menu right now is pretty opinionated about [it's default activator element](https://github.com/GetJobber/atlantis/blob/4fade5ab2a69d2e13bd645f0a0189ca257b61704/packages/components/src/Menu/Menu.tsx#L96-L105), so for users to easily to match the style they'd need to pass in Menu and a custom activator that defaults to `small` `tertiary` Button.

Could it be simplified so that if > 1 action is passed, the consumer has to provide a menu label but we take care of the "custom" activator and menu options internally? (Or, can the implementation right now be done in a way that allows for us to layer that in easily?)

packages/components/src/Card/Card.mdx
chris-at-jobber: Good call, we should _default_ to showing the default 😄

packages/components/src/Card/Card.tsx
chris-at-jobber: Nice!

I have some side thoughts about making Menu a bit more flexible but we can talk about those another time 😄

darryltec: ah good thing is, it won't change the implementation here so we're safe and should be talked about in another time and place haha

packages/components/src/Card/Card.tsx
```diff
@@ -1,26 +1,57 @@
...
+  /**
+   * The header props of the card.
+   */
+  readonly header?: HeaderProps;
 }
 
-interface LinkCardProps extends CardProps {
+type CardHeaderProps = Pick<CardProps, "title" | "header">;
+
+type LinkCardProps = CardProps & {
```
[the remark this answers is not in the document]
frankiepoon: i wasn't aware of the distinction, the guideline is now clear to me

packages/components/src/Card/Card.tsx
frankiepoon: updated the action to support `Button` and `Menu`, keeping the same restriction on `Button`. Was chatting with @darryltec for `Menu`, i am not planning to implement any specific restriction on the props. Also neither are not included in the documentation, I suppose to encourage using the `header` directly with title or custom component.

packages/components/src/Card/Card.tsx
frankiepoon: straigtened

darryltec (approved):
Looks good to me! There are some few bits that needs to be cleaned up but since it's not show-stopping, we can do a follow up PR to fix those!

PR #1045, opened 2023-02-17
docs(components): Convert another bundle to Storybook

rodrigoeidelvein:
## Motivations
Convert the first bundle of components to Storybook.

### Added
Storybook pages for:
- InputText
- Checkbox
- DescriptionList
- Content
- FormField
- LightBox
- StatusLabel
- Spinner

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/LightBox/LightBox.stories.mdx
+++ b/packages/components/src/LightBox/LightBox.stories.mdx
@@ -1,0 +1,110 @@
+import { useState } from "react";
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
+import { Button } from "@jobber/components/Button";
+import { LightBox } from "@jobber/components/LightBox";
+
+<Meta title="Components/LightBox" component={LightBox} />
+
+# Light Box
+
+LightBox is a component designed to display an array of images to the user. A
+user clicks on a control and is shown a carousel style presentation of the
+defined images.
+
+This component is a wrapper of the frontend-collective's react-image-lightbox
+third party library. This component is a wrapper of the
+[frontend-collective's react-image-lightbox](https://github.com/frontend-collective/react-image-lightbox).
+
+```ts
+import { LightBox } from "@jobber/components/LightBox";
+```
+
+<Canvas>
+  <Story name="LightBox">
+    {() => {
+      const [open, isOpen] = useState(false);
+      return (
+        <>
+          <Button label="😈 Click me!" onClick={() => isOpen(true)} />
+          <LightBox
+            open={open}
+            images={[
+              {
+                title: "😃 Happy Face",
+                url: "https://i.imgur.com/6Jcfgnp.jpg",
+                caption: "This is a happy face",
+              },
+              {
+                title: "😮",
+                url: "https://i.imgur.com/OsLxYkT.jpg",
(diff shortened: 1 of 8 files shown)


packages/components/src/LightBox/LightBox.stories.mdx
```diff
@@ -0,0 +1,111 @@
+import { useState } from "react";
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
+import { Button } from "@jobber/components/Button";
+import { LightBox } from "@jobber/components/LightBox";
+import { ExampleWithHooks } from "mdxUtils/ExampleWithHooks";
+
+<Meta title="Components/LightBox" component={LightBox} />
+
+# Light Box
+
```
MichaelParadis, line 10: This seems to be missing the other styling when the lightbox is open
Storybook:
image


Docs currently:

image

MichaelParadis (changes requested):
Minor issue with the light box story

packages/components/src/LightBox/LightBox.stories.mdx
MichaelParadis: Also doesn't show on the canvas tab

rodrigoeidelvein: I knew I'd have nightmares with this

rodrigoeidelvein: This looks like an issue with the library `react-image-lightbox`, but it's already deprecated and it's been a while since the last update.

I tried upgrading it to `5.1.4` because they fixed [this issue](https://github.com/frontend-collective/react-image-lightbox/issues/136) that is related to the lightbox being inside an iFrame (our case), but no success :(

MichaelParadis: Are we able to use the library that is recommended instead?
https://www.npmjs.com/package/react-photoswipe-gallery

We may want to add it to our list of packages to upgrade because it is no longer maintained.

I feel like this should be a blocker but I will defer to @chris-at-jobber

chris-at-jobber: Yeah this is a pretty bad representation of the component and this will also impact viewers of Gallery as that component consumes Lightbox

chris-at-jobber: If we can split it out to solve independently and land the rest of these docs, that could be OK for now. Would really like InputText to be available 😄

rodrigoeidelvein: I removed LightBox from this PR so we can land the other ones.

MichaelParadis (approved):
Overall looks good. I did notice that if I opened a link in a new tab it wouldn't work.
The url would be something like `http://localhost:6005/iframe.html?path=/docs/components-inlinelabel--inline-label` which doesn't seem to be correct. 
I know storybook has a link addon that may solve this.

Since we have the follow up subtask I don't think this is a blocker for this PR
https://storybook.js.org/addons/@storybook/addon-links

PR #1060, opened 2023-02-22
docs(components): Improve Atlantis branding on Storybook

rodrigoeidelvein:
## Motivations
Change the `a` tags colors and use our own `Heading` and `Text`.

### Note
This will temporarily show up since the existing readme needs it for docz

image

https://github.com/GetJobber/atlantis/blob/c30d262b2be0c862ee349f7a13ec63ce615f1e30/docs/README.md?plain=1#L1-L4

## Changes
- Change the `a` tags colors and use our own `Heading` and `Text`.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


.storybook/preview.js
```diff
@@ -8,4 +15,14 @@ export const parameters = {
       order: ["Components", "Design", "*"],
     },
   },
-}
+  docs: {
+    components: {
+      wrapper: props => <Content>{props.children}</Content>,
+      h1: props => <Heading {...props} level={1} />,
```
chris-at-jobber, line 21: Does this break the clickable/copyable heading URLs? @darryltec mentioned that might be the reason those don't work in our mobile Storybook instance.

darryltec: most likely, i'll bring em back on my PR. Seems like we have to do some custom bits on it to make it play nice

chris-at-jobber: We could potentially also just drop the CSS into the `preview-head` 🤷 

I think the crucial adjustments would be

`{ color: var(--color-heading); font-family: var(--typography--fontFamily-display); }`

darryltec: They use semantic tags anyway so we probably could just target those. The alternative is yeah, implement our own link anchor

chris-at-jobber:
Seeing discrepancies across pages:

❌  **Intro** 
image

✅  **Components**
image

✅ **Design**
image

✅ **Basically any other page**
image

chris-at-jobber:
I just saw that PR allowing `.md` to be imported so I'm assuming that's the reason for the difference on the ReadMe 😄 - can we resolve that?

chris-at-jobber:
I also noticed that the links in the content are `interactive` but the links in the TOC sidebar are not - is that something we can update?

image

chris-at-jobber:
One last piece / maybe separate from this work, the shadow from the "collapsed" TOC sidebar is visible on narrow viewports:

image

darryltec:
@chris-at-jobber those thangs you saw should be fixed now

chris-at-jobber:
I see this note in the description - how temporary is this? 

image

If we haven't already, we need to add an explicit note to the followup ticket that removes Docz to clean this up

chris-at-jobber (commented):
Design approved ✅

rodrigoeidelvein:
I know that I'm the PR creator, but @darryltec took over this, LGTM!

darryltec (approved):
Rodrigo's changes also looks good!

PR #1069, opened 2023-02-23
fix: JOB-62724 Update RadioGroup alignment

sabidhasan:
## Motivations
Style was updated for alignment around multiline children for the `RadioGroup` component. According to the [Contribution Guidelines](https://github.com/GetJobber/atlantis/blob/80ab88b320ffbb94af1dfd20c055ebd771c4bcaa/docs/CONTRIBUTING.md#bugfix), this is probably a BugFix, because the alignment is of the radio button when there are multiple children is broken. This _does_ affect the visual layout for the existing example in Docz that shows the component with user icons (this is shown in the before and after sections below).

We will do an audit of existing usage in Jobber to make sure this is working as intended and no consumers are affected.

### Changed

- RadioGroup.css:
Added a top margin and updated alignment to be flex start instead of vertically centered. Used a `margin-top` of `smallest` for the radio button.

---
Before
image

image

After
image

image

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/RadioGroup/RadioGroup.css
+++ b/packages/components/src/RadioGroup/RadioGroup.css
@@ -13,7 +13,7 @@
   display: inline-flex;
   font-size: var(--typography--fontSize-base);
   cursor: pointer;
-  align-items: center;
+  align-items: flex-start;
 }
 
 .input + .label::before {
@@ -22,7 +22,7 @@
   width: var(--space-base);
   height: var(--space-base);
   box-sizing: border-box;
-  margin: 0 var(--space-small) 0 0;
+  margin: var(--space-smaller) var(--space-small) 0 0;
   border: var(--border-thick) solid var(--color-interactive);
   border-radius: 100%;
   background-color: var(--color-surface);
--- a/packages/components/src/RadioGroup/RadioGroup.css
+++ b/packages/components/src/RadioGroup/RadioGroup.css
@@ -22,7 +22,7 @@
   width: var(--space-base);
   height: var(--space-base);
   box-sizing: border-box;
-  margin: var(--space-smaller) var(--space-small) 0 0;
+  margin: var(--space-smallest) var(--space-small) 0 0;
   border: var(--border-thick) solid var(--color-interactive);
   border-radius: 100%;
   background-color: var(--color-surface);


jlelford (commented):
I see the comment about doing an audit to determine if this breaks any existing instances, does that mean I should hold feedback for the arrival window child object for a different PR?

As for making changes to the component I'll defer to @chris-at-jobber but I think this looks as expected.

packages/components/src/RadioGroup/RadioGroup.css
```diff
@@ -22,7 +22,7 @@
   width: var(--space-base);
   height: var(--space-base);
   box-sizing: border-box;
-  margin: 0 var(--space-small) 0 0;
+  margin: var(--space-smaller) var(--space-small) 0 0;
```
chris-at-jobber, line 25: This margin isn't needed, it actually breaks the alignment:
 
image

chris-at-jobber:
```suggestion
  margin: 0 var(--space-small) 0 0;
```

sabidhasan: I think it might be conditionally required, because it also looks off when there are multiple children and the margin is provided. LMK if that makes sense?

chris-at-jobber (changes requested):
Small change to the styling around the radio indicator, otherwise looks good!

Once you've made the change, if you'd like a pre-release to test this in the product post in #team_atlantis and we'll get you a test version to do QA with.

packages/components/src/RadioGroup/RadioGroup.css
chris-at-jobber: Do you have an example?

sabidhasan: Hey @chris-at-jobber thanks for the review! This is what I am seeing when trying things out locally.

# Flex Center, Margin Top 0
This is as it currently is in the published version of this package in `master`.

image

# Flex Start, Margin Top Space-Smaller
This is what it looks like with the proposed change in this PR

image

# Flex Start, Margin Top 0
This is what it looks like with the suggested change:

image

In conclusion, none of these seem pixel-perfect to me, but I am definitely not a designer and am very open to other suggestions/feedback on how to get around this. In the meantime, I have updated my work that needs these changes to use a workaround (basically I am doing the same thing as what I am proposing in this PR, except only scoped to the component/UI needing this change). Once we have a path forward here, I can go back and remove my temporary fix.

Thanks in advance for feedback!

chris-at-jobber:
```suggestion
  margin: var(--space-smallest) var(--space-small) 0 0;
```

sabidhasan: Sweet - change made!

sabidhasan:
> does that mean I should hold feedback for the arrival window child object for a different PR?

Yes please. That should come under the PR that is using these changes and implementing the UI. Thanks 🙏

chris-at-jobber (approved):
Looks good! Thanks @sabidhasan !

PR #1070, opened 2023-02-23
fix(components): Allow "input" Chips to have hover states

chris-at-jobber:
## Motivations

Chip has a designed hover state
image

But it's not present in the rendered component. Thanks to @iremdesign and @rebecca-li for raising this!

Looking into the code, there's a check for `clickable` to add the hover-state styling which appears to only look for an `onClick`, but this ignores that a common usage of Chips for selection does not involve an `onClick`.

I have tried to solve this in a CSS-first way. I suspect there's maybe a "simpler" fix on the React side, but my understanding is that the logic of "inputs" is actually only raised at the Chip level and not the InternalChip level.

### Fixed

- An inactive Chip now has a hover state that includes a background-color change.
image

## Testing

- Interact with the Chips!

---

Random photo of Atlantis


rodrigoeidelvein (approved):
I'll let the chips say this:
image

LGTM! Thanks for fixing it, Chris

PR #1075, opened 2023-02-24
docs: Copy changelog to storybook

darryltec:
### Added

- Stories for changelog

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


taylorvnoj:
I just have two quick questions:

1. Can we move where the Change Log section is in the sidebar? I wonder if it should go after packages to not break up the flow of Design, Content, etc. And should it be Change Log with a space, or one word Changelog.
2. Under the Contents section on the right, looking at Components Change log for example, the list is super long. Is there a way to organize that by months/years?

Happy to approve otherwise!

darryltec:
> Can we move where the Change Log section is in the sidebar? I wonder if it should go after packages to not break up the flow of Design, Content, etc.

Totes can!

>  And should it be Change Log with a space, or one word Changelog.

These guys says 1 word

image

and they package is for changelog (we don't use them) so I kinda believe them haha

> Under the Contents section on the right, looking at Components Change log for example, the list is super long. Is there a way to organize that by months/years?

Quick answer no since this is just a doc that a bot updates. You'd actually see this (or the MD file) in every github repo 😆

taylorvnoj:
Thanks @darryltec! 

Cool, I believe those guys too 😆  In that case, can we then go in the `CHANGELOG.md` files for each and make that heading one word also?

Screenshot 2023-03-10 at 10 35 35 AM

darryltec:
yah gots it!

taylorvnoj (approved):
Wooo thank you! 🚀

PR #1095, opened 2023-03-02
refactor(components): remove react-image-lightbox

rodrigoeidelvein:
## Motivations
Refactoring the component `LightBox` to remove the library `react-image-lightbox` and create our own Lightbox. This implementation is a merge of an [example](https://codesandbox.io/s/framer-motion-image-gallery-pqvx3?from-embed=&file=/src/Example.tsx) from the `framer-motion` docs and the old implementation with the external library.

This doesn't change the consumption of the component, the props are still the same, and the behavior and UI has small differences but we can adjust if needed.

## Changes

- Removed `react-image-lightbox`
- Create our own Lightbox

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing
You can test the new component on the page for the `Lightbox` and `Gallery` on Storybook.

Pre-release: `npm i @jobber/components@3.3.2-pre.1+fbcb0d57`

---

Random photo of Atlantis


packages/components/src/LightBox/LightBox.tsx
```diff
@@ -43,43 +42,90 @@ interface LightBoxProps {
...
+    onRequestClose({ lastPosition: currentImageIndex });
+  }, "Escape");
+
+  useOnKeyDown(handleMovePrevious, {
+    key: "ArrowLeft",
+  });
+
+  useOnKeyDown(handleMoveNext, {
+    key: "ArrowRight",
+  });
```
darryltec, line 88: image

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -43,43 +42,90 @@ interface LightBoxProps {
...
 
   useEffect(() => {
     setCurrentImageIndex(imageIndex);
-  }, [imageIndex]);
+  }, [imageIndex, open]);
 
   return (
     <>
```
darryltec, line 95: only have 1 element inside so you don't need the fragment

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -43,43 +42,90 @@ interface LightBoxProps {
...
+      <AnimatePresence initial={false}>
+        {open && (
+          <LightBoxWrapper key="lightbox">
+            <Toolbar>
+              <Title>{images[currentImageIndex].title}</Title>
+              <ButtonDismiss
+                ariaLabel="Close lightbox"
+                onClick={() => {
+                  onRequestClose({ lastPosition: currentImageIndex });
+                }}
```
darryltec, line 105: As a best practice, it's much better to put this in a function somewhere and reference it here as `onClick={handleClick}` since named functions are much more performant than anonymous

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -92,4 +138,86 @@ export function LightBox({
...
+  children: ReactNode;
+}
+
+function Title({ children }: ChildrenProps) {
+  return <span className={styles.title}>{children}</span>;
+}
+
+function Toolbar({ children }: ChildrenProps) {
+  return <div className={styles.toolbar}>{children}</div>;
+}
```
darryltec, line 175: I think we don't need to do this since it's just aliasing an element anyway. Save us a component runaround. Same goes for the below as well. What do you think?

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -43,43 +42,90 @@ interface LightBoxProps {
...
+              <Title>{images[currentImageIndex].title}</Title>
+              <ButtonDismiss
+                ariaLabel="Close lightbox"
+                onClick={() => {
+                  onRequestClose({ lastPosition: currentImageIndex });
+                }}
+              />
+            </Toolbar>
+            <ImagesWrapper handleImageWrapperClick={handleImageWrapperClick}>
+              <PreviousButton onClick={handleMovePrevious} />
```
darryltec, line 109: I think we should just use our button here so it's consistent with our style. Either [work tertiary](https://atlantis.getjobber.com/?path=/story/components-button--button&args=label:;variation:work;type:tertiary;icon:arrowLeft&globals=backgrounds.value:!hex(F8F8F8)) or [subtle secondary](https://atlantis.getjobber.com/?path=/story/components-button--button&args=label:;variation:subtle;type:secondary;icon:arrowLeft&globals=backgrounds.value:!hex(F8F8F8))

packages/components/src/LightBox/LightBox.css
```diff
@@ -18,3 +18,61 @@
...
+  justify-content: space-between;
+  flex-direction: column;
+}
+
+.toolbar {
+  display: flex;
+  padding: var(--space-small) var(--space-base);
+  color: var(--color-greyBlue--lighter);
+  font-size: var(--typography--fontSize-large);
+  background-color: rgba(0, 0, 0, 0.5);
```
darryltec, line 66: I wonder if you'd get the same effect if you use the `var(--color-overlay)` here since it would be 2 overlay on top of each other

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -92,4 +138,86 @@ export function LightBox({
   function handleMoveNext() {
     setCurrentImageIndex((currentImageIndex + 1) % images.length);
   }
+
+  function handleImageWrapperClick(
+    event: React.MouseEvent<HTMLDivElement, MouseEvent>,
```
darryltec, line 143: Since we're destructuring the other ones like ReactNode instead of React.ReactNode, mind following that pattern for `MouseEvent`?

darryltec (changes requested):
This looks really really good and, quite frankly, much better than the package.

https://user-images.githubusercontent.com/15986172/222461970-9ecb58ab-95d8-4def-89c3-f3f183893e62.mp4

Few little things tho
- Animations seems to be stuck at always going left even if i want it right
  - Swiping is the same
- If the image is wide rather than tall, clicking the space above it doesn't close the lightbox, clicking the space beside works. Most likely because the overlay is the same height as the image. 
  - My suggestion, follow how Modal's do their overlay. It might simplify the function for it too
- Keyboard shortcuts work tho!! 👌🏼 
- That animation is sliiiiiick

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -43,43 +42,90 @@ interface LightBoxProps {
...
-          onMoveNextRequest={handleMoveNext}
-        />
-      )}
+      <AnimatePresence initial={false}>
+        {open && (
+          <LightBoxWrapper key="lightbox">
+            <Toolbar>
+              <Title>{images[currentImageIndex].title}</Title>
+              <ButtonDismiss
+                ariaLabel="Close lightbox"
```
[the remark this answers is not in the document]
rodrigoeidelvein: I kept the same aria labels that `react-image-lightbox` used, but I can update it

packages/components/src/LightBox/LightBox.tsx
rodrigoeidelvein: I did it to make it easier to read and understand the main component, but I'm fine with just having the divs directly there too

darryltec:
Also, while you're here, can you wire up the args in the story so we can modify it through the documentation

packages/components/src/LightBox/LightBox.tsx
rodrigoeidelvein: I think for this case `React.MouseEvent` is different than `MouseEvent`. `MouseEvent` refers to the DOM, while the other is React-specific. I tried to change its import from React, but then I get an error on `onEndDrag` because it's referencing DOM's MouseEvent.

packages/components/src/LightBox/LightBox.css
```diff
@@ -18,3 +18,61 @@
 .wrapper:global(.ril__outer .ril__toolbarItem) {
```
chris-at-jobber, line 18: Do we still need this? I _assume_ this is coming from the old package

rodrigoeidelvein: Nice catch!

packages/components/src/LightBox/LightBox.tsx
darryltec: I mean like import it above 

```
import React, { MouseEvent, ReactNode, useEffect, useState } from "react";
```

packages/components/src/LightBox/LightBox.css
```diff
@@ -18,3 +18,61 @@
...
+.next,
+.prev {
+  display: flex;
+  position: absolute;
+  top: calc(50% - 20px);
+  z-index: 2;
+  width: 20px;
+  height: 34px;
+  padding: 30px 20px;
+  font-size: 18px;
```
chris-at-jobber, line 31: Since the `next` and `prev` use `Icon` now, what is this text styling doing?

packages/components/src/LightBox/LightBox.css
rodrigoeidelvein: It's not the _same_, it looks even better

chris-at-jobber: Can we just go with surface--reverse here?

packages/components/src/LightBox/LightBox.css
```diff
@@ -18,3 +18,61 @@
...
+  height: 100%;
+  background-color: var(--color-overlay);
+  justify-content: space-between;
+  flex-direction: column;
+}
+
+.toolbar {
+  display: flex;
+  padding: var(--space-small) var(--space-base);
+  color: var(--color-greyBlue--lighter);
```
chris-at-jobber, line 64: Anywhere we use this text color let's replace with `var(--color-text--reverse--secondary)`

packages/components/src/LightBox/LightBox.css
```diff
@@ -18,3 +18,61 @@
...
+  font-weight: bold;
+  user-select: none;
+  background-color: rgba(0, 0, 0, 0.2);
+  cursor: pointer;
+  justify-content: center;
+  align-items: center;
+}
+
+.next {
+  right: 0px;
```
chris-at-jobber, line 41: Nitpick put if we're going to do this can we say `0` instead of `0px`? We use `0` elsewhere in the file.

packages/components/src/LightBox/LightBox.css
```diff
@@ -18,3 +18,61 @@
...
   color: var(--color-greyBlue--lighter);
 }
+
+.next,
+.prev {
+  display: flex;
+  position: absolute;
+  top: calc(50% - 20px);
+  z-index: 2;
+  width: 20px;
```
chris-at-jobber, line 28: These proportions are very random-seeming. Can we make them square? Would also like to round the corners (will add those comments on the next and prev classes below)

packages/components/src/LightBox/LightBox.css
chris-at-jobber: Would like the border radius for this to be

`var(--radius-base) 0 0 var(--radius-base)`

packages/components/src/LightBox/LightBox.css
```diff
@@ -18,3 +18,61 @@
...
+  justify-content: center;
+  align-items: center;
+}
+
+.next {
+  right: 0px;
+}
+
+.prev {
+  left: 0px;
```
chris-at-jobber, line 45: Would like the border radius for this to be

`0 var(--radius-base) var(--radius-base) 0`

packages/components/src/LightBox/LightBox.css
```diff
@@ -18,3 +18,61 @@
...
+}
+
+.lightboxWrapper {
+  display: flex;
+  position: fixed;
+  top: 0;
+  left: 0;
+  z-index: 10;
```
chris-at-jobber, line 53: We have defined values for z-index, can we use those in this file?

chris-at-jobber: https://job-62726-replace-react-imag.atlantis.pages.dev/?path=/docs/design-elevations--page#z-index-elevation

packages/components/src/LightBox/LightBox.css
chris-at-jobber: image

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -92,4 +138,86 @@ export function LightBox({
...
+    </div>
+  );
+}
+
+interface NavButtonProps {
+  onClick: () => void;
+}
+
+function PreviousButton({ onClick }: NavButtonProps) {
+  return (
```
chris-at-jobber, line 210: Why are we returning `div` instead of `button` elements here?

chris-at-jobber: (and on the Next "button" as well)

packages/components/src/LightBox/LightBox.tsx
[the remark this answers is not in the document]
chris-at-jobber: Could change it to be "image gallery" to be more descriptive?

Other question - does Lightbox 
- trap focus when launched (ie focus is set to the Lightbox and the user cannot tab behind to the underlying DOM elements) 
- use the `return focus to activator` hook when closed?

(I think it should if it doesn't)

packages/components/src/LightBox/LightBox.tsx
rodrigoeidelvein: I did, but than I get type errors

packages/components/src/LightBox/LightBox.css
chris-at-jobber: I just realized that won't look great with the ButtonDismiss 🙃

chris-at-jobber:
So this is new with the animation, but is it possible to preserve the previous image and slide it off the screen? Like this...

https://www.figma.com/proto/rIIhulZvcp9M82lNOCGv16/branch/v6fPuhEvgY67f5COdkyxr7/Product%2FOnline?page-id=13845%3A0&node-id=17450%3A67239&viewport=-60%2C-117%2C0.25&scaling=scale-down&starting-point-node-id=17450%3A67239

chris-at-jobber:
Note that when I drag it's a bit closer to this, but on arrow key or button click, the current image just blinks out of existence and the next/prev image slides in

chris-at-jobber:
@rodrigoeidelvein  that looks great!

chris-at-jobber (commented):
Design approved, thanks for the work here!

rodrigoeidelvein:
@chris-at-jobber that was a great improvement! Added it!

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -43,53 +47,170 @@ interface LightBoxProps {
...
+  });
 
   useEffect(() => {
     setCurrentImageIndex(imageIndex);
-  }, [imageIndex]);
+  }, [imageIndex, open]);
 
   return (
-    <>
+    <AnimatePresence initial={false}>
```
darryltec, line 108: we should portal this to the body since we have no control over where this get put on the component. Right now it'll render on the DOM wherever you put it on the code. The problem with that is if the parent have some `transform` property, all `position:fixed` acts like `position absolute`. Try wrapping it in a div with a `transform: translate(0)`. 

image

darryltec (changes requested):
Few lil bits for you homes! This is close!!

packages/components/src/LightBox/LightBox.css
```diff
@@ -1,20 +1,66 @@
...
-  background-color: var(--color-overlay);
+  justify-content: space-between;
+  flex-direction: column;
 }
 
-.wrapper:global(.ril__outer .ril__captionContent) {
-  color: var(--color-greyBlue--lighter);
+.toolbar {
+  display: flex;
+  z-index: var(--elevation-modal);
```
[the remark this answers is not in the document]
rodrigoeidelvein: I tried to make the structure:
```
.lightbox
  .overlay
  .toolbar
  .content
    .prev
    img
    .next
  .toolbar
```

But the overlay still renders above the toolbar, I think the `position: fixed` does it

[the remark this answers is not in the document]
darryltec: try adding a `z-index: var(--elevation-base)` on the toolbar

darryltec:
Something odd happened with the switcherroo. Works for a bit then not

https://user-images.githubusercontent.com/15986172/222762243-d36f1012-57f5-4e3f-91d4-e5fa3af32421.mp4

Big image doesn't resize. We can fix this with some CSS magic. basically don't use position absolute on this babies

image

darryltec (approved):
LGTM

PR #1100, opened 2023-03-02
feat(hooks): Add useFocusTrap hook

rodrigoeidelvein:
## Motivations

Promoting `useFocusTrap` to `@jobber/hooks`. It's currently being used only in `Modal`, but will be useful for another component soon.

## Changes

- Adds `useFocusTrap` to `@jobber/hooks`

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


darryltec (commented):
This looks good. Do you want to update the usage on Modal? in this PR? It's exactly the same code

packages/hooks/src/useFocusTrap/useFocusTrap.stories.mdx
```diff
@@ -0,0 +1,89 @@
...
+                zIndex: "var(--elevation-modal)",
+              }}
+            >
+              <Overlay onClick={() => setOpen(false)} />
+              <Modal>
+                <Button label="first button" />
+                <Button label="second button" />
+                <Button label="third button" />
+                <Button label="close" onClick={() => setOpen(false)} />
+              </Modal>
```
darryltec, line 46: Got confused about what this modal is, I thought it was our Atlantis modal. We shouldn't do that tho since people copy paste and they might thing they have to do that on every Atlantis Modal

I think we can simplify the example to just be the buttons. Like, don't need to position fixed the parent, literally just show the buttons on the canvas (maybe wrapped in a card?) and let people tab around and see that it's trapped within that box.

rodrigoeidelvein: Ooh, that makes sense! I'll change the example, it's really confusing using the same names

darryltec (approved):
we gucci, tis magnifique! thanks for doing this

PR #1105, opened 2023-03-03
chore(hooks): Upgrade graphql and @apollo/query

rodrigoeidelvein:
## Motivations

Upgrading `@apollo/client` and `graphql` to prepare Atlantis for the React 18 upgrade. I had to upgrade `graphql` too because there was a [type error ](https://github.com/apollographql/apollo-client/issues/9167) when I only upgraded `@apollo/client`

## Changes
- Bump `@apollo/client` to `^3.7.10`
- Bump `graphql` to `^15.5.0`

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/hooks/package.json
```diff
@@ -38,7 +38,7 @@
     "zxcvbn": "^4.4.2"
   },
   "peerDependencies": {
-    "@apollo/client": "^3.3.16",
+    "@apollo/client": "^3.7.10",
```
darryltec, line 41: this is gonna be interesting tho. I wonder if @GetJobber/frontend-foundations peeps are updating apollo in JO. If they are, then this should be fine. Otherwise, this might be a breaking change.

adambobadam: Don't recall this needing to be upgraded in JO - can't see any FeF subtasks for it.  Still at 3.3.16 in JO.

darryltec: yeah, we're bumping this because it's only for R16,17 
image

adambobadam: Oooh okay, dang 😬

darryltec: also, based on the test that failed on this PR, it might break some shiiiiiii

timoteialbu: Thanks for pointing this out. [Testing right now](https://github.com/GetJobber/Jobber/pull/34139) to see what this means for JO. I do not understand why `npm ls` does not surface that up.

darryltec: actually just found this out after upgrading to node 18 on my react 18 branch. That's the first error I got 😆

timoteialbu: So weird. Damn you `npm` 👊

adambobadam: graphql-ruby-client doesn't support this apollo client version in JO, so I got this [PR](https://github.com/GetJobber/Jobber/pull/34147) to bump it

timoteialbu: I got a PR open for this. I wonder if we need to be specific about the version or if we can/should do `>=3.7.10` or something. I also wonder why the peer Dependency is different than the "real" dependency.

adambobadam:
Yo @rodrigoeidelvein Why did you switch from 3.7 to 3.6 there? Is it because it's the minimum that still works in Atlantis with react 18?

rodrigoeidelvein:
> Yo @rodrigoeidelvein Why did you switch from 3.7 to 3.6 there? Is it because it's the minimum that still works in Atlantis with react 18? Also how come the peer and dev dependencies are mismatched? Trying to sort it out in my noggin' 😆

3.6.0 is the first one that supports react 18 and 16. I tried this version to see if the tests were breaking on latest, but it still breaks :( 

I'll try to figure out the `devDependencies` and `peerDependencies` versions here on Atlantis, it's confusing everybody 😅

timoteialbu:
> > Yo @rodrigoeidelvein Why did you switch from 3.7 to 3.6 there? Is it because it's the minimum that still works in Atlantis with react 18? Also how come the peer and dev dependencies are mismatched? Trying to sort it out in my noggin' 😆
> 
> 3.6.0 is the first one that supports react 18 and 16. I tried this version to see if the tests were breaking on latest, but it still breaks :(
> 
> I'll try to figure out the `devDependencies` and `peerDependencies` versions here on Atlantis, it's confusing everybody 😅

For now, I am deploying v 3.7.10 on staging3 to test with. I can change that later to the one you are using. Not sure if there will be issues because atlantis is on another version.

packages/hooks/package.json
darryltec: to answer your question there @timoteialbu I think it's mostly to make sure that what we use in Atlantis is the same as what we would use in other places since this whole gql thing is finicky... i think.

timoteialbu: Would we ever get rid of `useCollectionQuery` in favor of something apollo has built in so that we do not have this dependency to begin with? I suppose its not the worst dependency to have cause its used in one place and its easy enough to update atlantis, but still.

package.json
```diff
@@ -83,6 +83,7 @@
     "eslint-plugin-mdx": "^1.16.0",
     "glob": "^7.1.4",
     "graphql": "^15.8.0",
+    "graphql-ws": "^5.12.0",
```
timoteialbu, line 86: what is this for?

packages/hooks/package.json
darryltec: that is a question that I do not have an answer to. IMO, we should but that's a question best answered by the people that uses this. Maybe ask in #engineering, #tech-mobile/web?

package.json
darryltec: peer dep
image

adambobadam (commented):
Had a go at testing this out in JO on Timotei's [branch](https://github.com/GetJobber/Jobber/pull/33342) using the prerelease @darryltec made!

- Checked `PromotionListLoader` and seemed to be functioning fine!
- Looking good in Atlantis!

Todo:
- Need to check with Comms for the conversation/chat components (ran through the 2way sms setup a couple days ago, but got stuck with Igor on the setup before the part where I could potentially QA this)
- Need to check with Otterbots/Bizops on DistributionChannel (tried populating them there but they're blank)

adambobadam:
Further QA done:

Reached out to Bizops [here](https://getjobber.slack.com/archives/C26DWUMMX/p1678294851720089) and checked the prerelease again on Timotei's branch.

Steps to check distribution channel/notes after pairing with @BronwynThompson:
- Added distribution channels and checked that `useCollectionQuery` was present and the search functionality was working.
- Added >20 accounts to a distribution channel and checked that `useCollectionQuery` was working there as well.

PR #1106, opened 2023-03-03
refactor(components): Simplify Banner styling [JOB-46520]

chris-at-jobber:
## Motivations

Banner has some very aggressive selector-targeting CSS that is both susceptible to responsive edge cases and hard to reason about.

There is conditional styling that applies and removes margins on different axes of different elements based on container width. Since the Banner uses flex, `gap` is a natural way to resolve these issues in a much simpler fashion. Removing `> *` selectors is also a great way to prevent unintended breakage of child elements which we cannot predict.

In addition, internally the Banner uses
- function and class names internally that reference a legacy Jobber `flash` component
- non-semantic "base" color values that will not respond to theming

### Changed

- Internal styling and function names of Banner for better editability

### Removed

- generic selector targeting

### Fixed

- Cases where the text does not wrap but the Banner has conditional styling applied no longer pushes the 'action' down out of alignment with the text

## Testing

Plug in some long and short content, try it on a few different breakpoint sizes, with and without action, with and without dismiss.

---

Random photo of Atlantis


chris-at-jobber:
@taylorvnoj Would love to have you take a look at this one as I _think_ it can simplify the work you're looking at in the in-Jobber "global" banner

taylorvnoj (approved):
Much cleaner! LGTM 🎉

PR #1109, opened 2023-03-06
build(deps): Bump React to 18 [JOB-62732]

darryltec:
## Motivations

We are out of date and we need to get to 18!

### Added

- 

### Changed

- Bumped React and React DOM version to 18
- Peer deps have been updated to >= 16.8 since Atlantis can still support them right now
  - And the fact that we can just send this to JO master (instead of living in a feature branch there) since we don't use any specific R18 features

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


taylorvnoj:
All components work as expected in Storybook. Will use the pre-release shortly to test out JO

taylorvnoj (approved):
Also did a smoke test in JO and everything appeared to be normal. 

LGTM but I think another set of 👀 would be a good idea!

PR #1115, opened 2023-03-08
fix: use update instead of forceUpdate on popper

darryltec:
## Motivations

`forceUpdate` from popper was too forceful that it'll crash the whole page and throw an error around `flushSync` when you use dismissible chips. This PR should make it less forceful while still keeping the same functionality.

### Added

- 

### Changed

- replaced `forceUpdate` to `update` in dismissible chips component
  - The reason why we needed to update is to reposition the options when you've selected enough chips that it goes into multiple lines. This is so that it doesn't cover the other selected chips.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Selecting many chips and going into multiple lines should reposition the options and not cover the selected chips.

---

Random photo of Atlantis


ajpeng (approved):
Works great 👍

PR #1119, opened 2023-03-14
feat(components): Add highlightDates and onMonthChange props to DatePicker

TJKruitbosch:
## Motivations

We would like to be able to highlight dates and detect a month change event on the Atlantis `DatePicker` component. These props and functionality are available in the underlying [`ReactDatePicker`](https://reactdatepicker.com/) component, this PR is just exposing them in this wrapper.

The existing `DatePicker` component is already replacing most of the calendar styles from the underlying component to match our own style guide. To implement the highlighted dates, the `--color-surface--active` colour was set as their background. This matches how we're using `--color-surface--hover` for the focus/hover state and provides a state visually _between_ the default (`--color-surface`) and the selected (`--color-interactive`) states in a visual hierarchy.

### Added

- Added highlightDates and onMonthChange props to DatePicker component

## Testing

[Additional section](https://datepicker-highlightdates.atlantis.pages.dev/?path=/docs/components-datepicker--date-picker#highlighted-dates) added to the Storybook docs demonstrating the `highlightDates` functionality. Additional jest test added to the test suite to confirm `onMonthChange` behaviour.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -141,6 +141,14 @@ export function DatePicker({
     </div>
   );
 
+  /**
+   * The onChange callback on ReactDatePicker returns a Date and an Event, but
+   * the onChange in our interface only provides the Date. Simplifying the code
+   * by removing this function and passing it directly to the underlying
+   * component breaks tests both here and downstream (i.e. the pattern
+   * `expect(onChange).toHaveBeenCalledWith(date)` is commonly used and would
+   * fail).
+   */
   function handleChange(value: Date /* , event: React.SyntheticEvent */) {
     onChange(value);
   }


packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -123,6 +135,8 @@ export function DatePicker({
         onCalendarOpen={handleCalendarOpen}
         onCalendarClose={handleCalendarClose}
         dateFormat={["P", "PP", "PPP", "MMM dd yyyy", "MMMM dd yyyy"]}
+        highlightDates={highlightDates}
+        onMonthChange={handleMonthChange}
```
darryltec, line 139: can just skip the function and go straight to 
```suggestion
        onMonthChange={onMonthChange}
```

right?

TJKruitbosch: Yeah, that should be fine; was just mimicking what was done with `onChange`/`handleChange`. Which...while we're here, should that one get cleaned up too?

darryltec: that would be great! less code to do some run around the code

TJKruitbosch: Updated the `onMonthChange` handler. Tried to cleanup `onChange`/`handleChange`, but it turns out the ReactDatePicker gives both date and an event as parameters. Which theoretically should be a non-breaking change, but that breaks tests both here (DatePicker and InputDate) and would probably cause downstream breaks in tests (when others write something like `expect(onChange).toHaveBeenCalledWith(date)` in their component tests).

darryltec: Oh yah, i knew there was a reason why we did it!

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -123,11 +135,13 @@ export function DatePicker({
...
         onCalendarClose={handleCalendarClose}
         dateFormat={["P", "PP", "PPP", "MMM dd yyyy", "MMMM dd yyyy"]}
+        highlightDates={highlightDates}
+        onMonthChange={onMonthChange}
       />
     </div>
   );
 
-  function handleChange(value: Date) {
+  function handleChange(value: Date /* , event: React.SyntheticEvent */) {
```
darryltec, line 144: any reason why you kept that comment?

TJKruitbosch: So that the next person to come along and think "oh, this isn't doing anything I can remove it" might realize that there's that small difference in the call signature. That said, maybe it doesn't really communicate _why_ the difference matters 🤔

darryltec: yeah, i'd leave a description around why

TJKruitbosch: Added a description for why `handleChange` is necessary here.

darryltec (approved):
we goooooood!

PR #1139, opened 2023-03-17
fix(components): define and standardize lineheight in FormField

chris-at-jobber:
## Motivations

Because there's no defined lineheight on the fieldffix labels, but there is one on the `input` element, the fieldaffix labels are vulnerable to external stylesheets "leaking in" and causing unsightly inconsistencies when components using FormField are consumed by another project.

image

### Changed

- Added a CSS custom prop to the input and used wherever line-height was already being defined, then added it on the fieldaffix where there was no definition. I'm also open to the idea of using `inherit` on the fieldaffix level, or other suggestions for a cleaner way to solve this! 🤷 

### Fixed

- FormField and assorted consumers will no longer have misaligned baselines when the application has a global stylesheet with a different default.

## Testing

- Set a line-height on the body of the application and see that the FormField alignment stays as expected
- Run a pre-release into Jobber and observe that affixes and values line up as expected

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -27,6 +27,7 @@
   --field--padding-right: var(--space-base);
 
   --field--value-color: var(--color-heading);
+  --field--value-lineHeight: calc(var(--base-unit) * 1.25);
 
   --field--border-color: var(--color-border);
   --field--background-color: var(--color-surface);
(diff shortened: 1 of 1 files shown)


chris-at-jobber:
Would love input (pun intended) on my approach here - is there a cleaner way to get the same result?

darryltec (approved):
I can't think of a cleaner way to do this! This one works well!

PR #1142, opened 2023-03-20
build: Use NPM Workspace instead of Lerna to bootstrap

darryltec:
## Motivations

Lerna is planning to deprecate [bootstrap](https://github.com/lerna/lerna/discussions/3410) on their V7 since NPM can do it in a much better way. With the Node 18 upgrade, the need to move has become more apparent since it totally broke the way it worked for us.

### Known issues
- It'll throw a dependency warning and that's related to storybook upgrade. That comes later.

### Added

- 

### Changed

- Switch Lerna magic to NPM Workspace
  - In doing so, we can delete the packages package-lock.json since they're managed on the root now. No more confusing package-lock changes
- Had to work around NPM's lack of topological order when bootstrapping
  - Basically, npm would run `prepare` on each packages in alphabetical order but components need design to be built first.
  - This works around that limitation by firing the scripts manually through `npm run [script] -w` which respects the order under `workspaces`
  - https://github.com/npm/rfcs/issues/548 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Run `nvm use`
- Run `npm ci` successfully with no error. Warning is fine.
- Run `npm start` and storybook runs locally

---

Random photo of Atlantis

[JOB-63633]

[JOB-63633]: https://jobber.atlassian.net/browse/JOB-63633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rodrigoeidelvein:
@darryltec running `npm ci` is throwing the following error for me:
`npm ERR! Cannot read property '@std-proposal/temporal' of undefined`

I deleted all `node_modules` and I still got the error

darryltec:
Cloudflare failed because it's trying to build on node 18, but it can't 😆

darryltec:
@rodrigoeidelvein cloudflare failed becauase it's building on node 18 which it can't haha. Cloudflare fires `npm run build`

rodrigoeidelvein (commented):
I'm going to make a list of things that I noticed here, just to see if I understood correctly:
- Removed the `sum_lockfiles` because there's no `package-lock.json` in `packages/*` anymore? Are they all in the root `package-lock?`
- `--no-optional == -imot=optional`?
- You reorganized the npm command in the root `package.json`, and renamed some of them, but the only thing that changed inside the commands is the `export NODE_OPTIONS=--openssl-legacy-provider` in Storybook?
- Why did some snapshots change? 
- Loved the bonus refactor on `FormatRelativeDateTime` tests, was it failing?

packages/components/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "rollup --config",
     "clean": "rm -rf dist/* tsconfig.tsbuildinfo",
     "prepublishOnly": "ts-node --project ../../tsconfig.bin.json scripts/entryPoints.ts",
-    "prepare": "npm run clean; npm run build",
+    "bootstrap": "npm run clean; npm run build",
```
MichaelParadis, line 11: Reading this: https://lerna.js.org/docs/getting-started#package-dependency-management
It looks like the `@jobber` packages should be "*" is there a reason that wasn't changed?

darryltec: No reason other than it also worked and still symlinked so I didn't touch it haha

MichaelParadis: Do you think we should update it to match what lerna expects?

darryltec: most likely yeah

darryltec: It works! 
image

rodrigoeidelvein (approved):
Niiiice! Looks good to me!

- npm ci works
- storybook runs and builds

I just would like an approval from @MichaelParadis too

MichaelParadis (approved):
LGTM!

PR #1157, opened 2023-03-28
refactor(components): Replace Temporal library

rodrigoeidelvein:
## Motivations

Removing `@std-proposal/temporal` and adding `@js-temporal/polyfill` as the old one was deprecated. The new library is an alpha version too, but it's still being maintained and it's the recommended library by the creators of `@std-proposal/temporal.`

## Changes

- Uninstalled `@std-proposal/temporal`
- Installed `@js-temporal/polyfill`
- Replaced its usages and types

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Tests are passing and Storybook pages are loading in the components that were refactored.

Pre-release:  `@jobber/components@4.2.2-pre.8+9dd6a2c7`

---

Random photo of Atlantis


JeffMarvin:
@rodrigoeidelvein Just curious... we have some components (eg. DatePicker) that use `Date` -- is the idea to keep those as-is for now?

rodrigoeidelvein:
> @rodrigoeidelvein Just curious... we have some components (eg. DatePicker) that use `Date` -- is the idea to keep those as-is for now?

Good question, maybe we should accept both `Date` and `Temporal.PlainDateTime` every where, because the new Temporal is still a proposal, so we probably shouldn't rely just on that too... what do you think?

packages/components/src/InputTime/civilTimeConversions.tsx
JeffMarvin, line 1: This file name reflects the old "Civil Time" naming. I wonder if this file and its functions could be renamed. Not sure where they all get used/called, so let me know if that makes the scope blow up! (E.g. are they used in Jobber Online??)

JeffMarvin:
> > @rodrigoeidelvein Just curious... we have some components (eg. DatePicker) that use `Date` -- is the idea to keep those as-is for now?
> 
> @JeffMarvin, Good question, maybe we should accept both `Date` and `Temporal.PlainDateTime` every where, because the new Temporal is still a proposal, so we probably shouldn't rely just on that too... what do you think?

In most places, it looks like we are allowing both... only in `InputTime` I think we aren't... my gut feel is that we shouldn't add it at this point, and wait to see if there is a demand to do it. Also, not sure if they will keep the name, but **hopefully** the standard goes through and we can just remove the polyfill without another round of name changes (`Temporal.*` -> `NewName.*`)!

packages/components/src/InputTime/civilTimeConversions.tsx
MichaelParadis: I think it would be a good idea to rename this. The functions aren't directly exported in the package so they are only used in Atlantis.

rodrigoeidelvein: What do you think about `atlantisTimeConversions`?

rodrigoeidelvein: or `plainTime*` to be closer to the package/proposal we're using?

MichaelParadis: Yeah, that sounds good to me

rodrigoeidelvein:
After discussing with the team, we evaluated that it will be better to keep the deprecated library for now as it’s widely used in Atlantis and Jobber Online and there are risks of changing the time object, adding support for the old package or replace it.

The solution would be to change the library, which is deprecated, for another polyfill of a tc39 proposal that should NOT be used in production.

It’s better to wait for tc39 to release the temporal proposal than having to change it again once they do it.

PR #1159, opened 2023-03-28
feat: Add uploadUrl to FileUpload

gregorijobber:
## Motivations

We need this to be able to use the upload expense receipt (following the same as the mobile side)

### Added

- `uploadUrl` is now a prop from `FileUpload` type following the same pattern as the mobile

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -39,7 +39,7 @@ export interface FileUpload {
   /**
    * Url file was uploaded to, used to send in mutation to Jobber Online
    */
-  uploadUrl?: string;
+  readonly uploadUrl?: string;
 
   /**
    * The data url of the file.
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -37,6 +37,11 @@ export interface FileUpload {
   readonly progress: number;
 
   /**
+   * Url file was uploaded to, used to send in mutation to JO
+   */
+  uploadUrl?: string;
+
+  /**
    * The data url of the file.
    */
   src(): Promise<string>;
@@ -225,18 +230,23 @@ export function InputFile({
     } = await getUploadParams(file);
 
     const fileUpload = getFileUpload(file, key);
-    onUploadStart && onUploadStart({ ...fileUpload });
+    onUploadStart &&
+      onUploadStart({
+        ...fileUpload,
+        uploadUrl: url,
+      });
 
     const handleUploadProgress = (progressEvent: any) => {
       onUploadProgress &&
         onUploadProgress({
           ...fileUpload,
           progress: progressEvent.loaded / progressEvent.total,
+          uploadUrl: url,
         });
     };
 
     const handleUploadComplete = () => {
-      onUploadComplete?.({ ...fileUpload, progress: 1 });
+      onUploadComplete?.({ ...fileUpload, progress: 1, uploadUrl: url });
     };
 
     const axiosConfig = createAxiosConfig({
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -37,7 +37,7 @@ export interface FileUpload {
   readonly progress: number;
 
   /**
-   * Url file was uploaded to, used to send in mutation to JO
+   * Url file was uploaded to, used to send in mutation to Jobber Online
    */
   uploadUrl?: string;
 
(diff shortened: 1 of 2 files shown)


packages/components/src/InputFile/InputFile.tsx
```diff
@@ -36,6 +36,11 @@ export interface FileUpload {
    */
   readonly progress: number;
 
+  /**
+   * Url file was uploaded to, used to send in mutation to JO
+   */
+  uploadUrl?: string;
```
darryltec, line 42: we mark most of our props as `readonly` to ensure we don't randomly overwrite it some place else and for overall consistency as well

```suggestion
  readonly uploadUrl?: string;
```

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -36,6 +36,11 @@ export interface FileUpload {
    */
   readonly progress: number;
 
+  /**
+   * Url file was uploaded to, used to send in mutation to JO
```
darryltec, line 40: Outside of Jobber, no one would know what JO is

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -225,18 +230,23 @@ export function InputFile({
     } = await getUploadParams(file);
 
     const fileUpload = getFileUpload(file, key);
-    onUploadStart && onUploadStart({ ...fileUpload });
+    onUploadStart &&
+      onUploadStart({
+        ...fileUpload,
+        uploadUrl: url,
```
darryltec, line 236: should this be guarded? Don't add it unless it actually exists?

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -225,18 +230,23 @@ export function InputFile({
     } = await getUploadParams(file);
 
     const fileUpload = getFileUpload(file, key);
```
darryltec, line 232: `uploadUrl` probably should be added here since this gets spread on all the places that need it anyway. Fewer duplication and missed code

darryltec (changes requested):
this is more of a `feat:` as per 

> feat: a commit of the type feat introduces a new feature to the codebase

on the PR template.

Also, few little things for you

packages/components/src/InputFile/InputFile.tsx
ryanb-jobber: Would it make sense to add `uploadUrl` to the `fileUpload` returns from `getFileUpload`? Rather than appending on each use?

ryanb-jobber: Nevermind you already made that change 😄

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -36,6 +36,11 @@ export interface FileUpload {
    */
   readonly progress: number;
 
+  /**
+   * Url file was uploaded to, used to send in mutation to Jobber Online
```
darryltec, line 40: Ah, better, but what I mean is that it's not just Jobber Online who uses this open source project. 

We need to be clear on where did this URL came from and what is it for. Looking at the code it is just a thing that came from what you passed in the `getUploadParams`, right? In theory, you would have this already without doing a change here but in a sense of ease-of-use, sure let's add it.

Also raises the question of whether this should be optional if `getUploadParams` and the `url` it returns are both required?

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -224,8 +229,14 @@ export function InputFile({
       httpMethod = "POST",
     } = await getUploadParams(file);
 
-    const fileUpload = getFileUpload(file, key);
-    onUploadStart && onUploadStart({ ...fileUpload });
+    const fileUpload = {
+      ...getFileUpload(file, key),
+      uploadUrl: url || undefined,
```
darryltec, line 234: I was thinking more on the lines of

```
{
  ...
  ...(url && { uploadUrl: url })
}
```

That's the reason why the test changed. Since it will be there now wether it's undefined or nah.

but if the answer above is "this should always be there" then don't need to do this.

packages/components/src/InputFile/InputFile.tsx
gregorijobber: I copied that from mobile code, but yes, as `getUploadParams` and `url` will always be defined so this one is never undefined and not optional

darryltec: That's fair, but, in the context of this repo, how should we write that? 🤔

gregorijobber: `Url file was uploaded to` maybe is enough? As this is just the base url not the full file url

darryltec: ah, that prop should've been named `baseUrl` but for the sake of consistency, let's stick with `url`.

You can also use the existing one where this came from as an inspiration

https://github.com/GetJobber/atlantis/blob/8f33bd6f75d9fb4e241e98dbe2ac46588dbc665d/packages/components/src/InputFile/InputFile.tsx#L47

> Base URL where file was sent as a POST. This is the same URL as what's returned in `getUploadParams`.

packages/components/src/InputFile/InputFile.tsx
darryltec: remove the `|| undefined` since it wont be undefined now

darryltec (approved):
:shipit:

PR #1161, opened 2023-03-29
feat(components): Add new “money” icons

BenA-jobber:
## Motivations
Adding the `money` icon so that we can display it for the Jobber Money feature

### Added
added `money` icon and updated storybook. `money` icon is placed under `transaction`
Screenshot 2023-03-29 at 11 55 30

## Testing
- [ ] `npm run start`
- [ ] check that the new `money` icon exists

---

Random photo of Atlantis


BenA-jobber:
> @BenA-jobber it looks correct but a bit small

hmm. i think it should be the same size as all the other icons, which is set to 1024 (pixels?). @taylorvnoj can you verify this please :)

darryltec:
I guess in relation to the wallet icon, it looks a tad bit small. Might want to double check how it got exported from figma. If that's correct, then check how it got imported in icomoon. the white-space around the icons is normally the culprit as to what makes it smaller than the other even though it's the same size

darryltec:
Oh, ha! It is actually built smaller so @MadsCogen1 might wanna adjust it from the grid to accommodate the circle or leave it as is

image

image

MadsCogen1:
Oh I see, I think this should be fine @BenA-jobber

taylorvnoj (approved):
Tested in `jobber-mobile` and everything looks good! Thanks Ben! 🚀

PR #1175, opened 2023-04-05
fix: Use chip label instead of value on aria-label [JOB-66426]

rodrigoeidelvein:
## Motivations

In order to improve accessibility on dismissible chips, we are changing the `ariaLabel` to use the chip `label` instead of the `value` to represent what the user is actually reading instead of the value of the chip that can be an ID.

## Changes

- Changed dismissible chip to use `label` instead of `value`

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


darryltec (approved):
LGTM!

PR #1176, opened 2023-04-05
fix(components): fix datepicker validations

JeffMarvin:
## Motivations
This fixes an `InputDate` bug that causes the field to be marked as invalid when it shouldn't. E.g. when it starts empty and is required, picking a date from the DatePicker will still mark the field as invalid even though it fills in a date.

Repro Steps: 
1. Start with an `InputDate` initialized to have no value, but with a validation requiring the field (See Testing below for a validation example). 
2. Select the `InputDate` to launch the `DatePicker` and select a date.
- **Expected:**
The date should fill the `InputDate` 
- **Observed:**
The date is filled in, but the invalid styling is applied and the validation error is shown.

https://user-images.githubusercontent.com/6612596/230157835-9fe66223-b041-496f-b943-bfd9b766d25d.mp4

### Fixed

- Fixed an issue that was causing the InputDate field to mark the field as invalid when a date was selected from the DatePicker.

## Testing

Full tests tbd. 
1. Load storybook canvas for `InputDate`
3. Add the following validation 
```
{
    "required": {
      "message": "Start time required",
      "value": true
    }
  }
```
4. Clear the date field. 
5. OBSERVE: The proper validation error message appears, because the field is empty but required.
6. Click the InputDate to launch the DatePicker and select a date.
7. OBSERVE: The validation error message disappears correctly.
8. Click the InputDate to launch the DatePicker and select another date.
9. OBSERVE: The validation error message does not appear. 
10. Clear the date field. 
11. OBSERVE: The proper validation error message appears, because the field is empty but required.

---

Random photo of Atlantis


JeffMarvin:
There is currently an issue where setting the value using the imperative handle works, but some intermediary process is also very briefly changing the state, resulting in a very brief "blip" of incorrect content briefly visible in the field.
Apr-05-2023 10-46-16
Apr-05-2023 10-37-06

packages/components/src/InputDate/InputDate.tsx
```diff
@@ -40,10 +41,17 @@ interface InputDateProps
...
   return (
     <DatePicker
       selected={inputProps.value}
-      onChange={inputProps.onChange}
+      onChange={val => {
+        inputProps?.onChange(val);
+        formFieldActionsRef.current?.setValue(
+          // Text Input has a string type
+          `${val.getMonth()}/${val.getDay()}/${val.getFullYear()}`,
+        );
```
darryltec, line 53: instead of manually parsing the date, since the format of it is highly dependent on the system's settings, you can move this bit down to the activator

Cursor

Fixes the issue and we won't have to deal with date formatting

JeffMarvin: Nice!! Ok, on it.

packages/components/src/FormField/FormField.tsx
```diff
@@ -66,7 +67,7 @@ export function FormField(props: FormFieldProps) {
```
JeffMarvin, line 67: Should we proactively set the `{shouldValidate:true}` in the [useEffect here](https://github.com/GetJobber/atlantis/pull/1176/files#diff-7664eb8fb7178dac2dae4e5ccb337b655a800ab7dade9a2ec9501c0d428c78f7R64) as well??

rodrigoeidelvein: I think @darryltec or @MichaelParadis would have a better idea on it than myself about it

packages/components/src/InputDate/InputDate.tsx
```diff
@@ -40,10 +41,11 @@ interface InputDateProps
 }
 
 export function InputDate(inputProps: InputDateProps) {
+  const formFieldActionsRef = useRef<FieldActionsRef>(null);
   return (
     <DatePicker
       selected={inputProps.value}
-      onChange={inputProps.onChange}
```
rodrigoeidelvein, line 46: I think we can remove this optional chaining here

rodrigoeidelvein (changes requested):
I tested it and it works well!
I just added a nitpick and I think Darryl or Michael will have a better answer to your question about the `{shouldValidate:true}`, so it's up to you if you want to wait for an answer or just merge it and discuss it on another PR.

packages/components/src/FormField/FormField.tsx
MichaelParadis: I think so. Mainly because the revalidate mode will be `onChange` so we should trigger validations since `setValue` doesn't by default

JeffMarvin:
@rodrigoeidelvein Ok, I've made these two tweaks.
I want to test the form in a few real world scenarios (in Jobber Online) and also to catch a video of the fixed behaviour. There's an issue preventing me from bringing Atlantis into JO right now, that @MichaelParadis and I discovered on Friday. I'll work to get that fixed and then merge this to test it.

rodrigoeidelvein (approved):
LGTM!

PR #1182, opened 2023-04-11
fix(components): Fix label z-index and autofill style on Inputs [JOB-63282]

rodrigoeidelvein:
## Motivations

When autofilling an Atlantis input using the browser's default behavior, the label disappeared and the input's background was set to blue.

## Changes

- Removed the reassigned of the CSS variable `--field--base-elevation` inside `miniLabel`
- Added the styling to remove the autofill background style ([SO answer](https://stackoverflow.com/a/14205976))

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Input's autofill behavior

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -62,7 +62,6 @@
 }
 
 .miniLabel {
-  --field--label-elevation: calc(var(--field--base-elevation) + 1);
   --field--placeholder-color: var(--color-text--secondary);
   --field--placeholder-offset: var(--space-smallest);
   --field--placeholder-transform: 0;


fagnanm (commented):
LGTM, tested in Safari, Chrome, Edge, mobile Safari. Just waiting for my Android device to have some power and will test there as well.

Changing the z-index on the minilabel actually fixes another issue on multi-line text areas where the value overlays the label on scroll. Eventually we should add a background to the mini label so that it doesn't conflict with a longer scrolling value.

fagnanm (approved):
Have tested on Android as well and looks good

MichaelParadis (approved):
LGTM!

PR #1191, opened 2023-04-17
fix: Update positioning of Autocomplete Menu

darryltec:
## Motivations

It gets cut off in the storybook because it's rendered within a parent with an `overflow: hidden`. Best practice around floating elements like a menu is always to throw it to the body.

### Before
image

## After
image

### Added

- 

### Changed

- Portal Autocomplete Menu to the body
- Replaced positioning logic of Autocomplete Menu

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Autocomplete/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu.tsx
@@ -1,6 +1,8 @@
-import React, { useEffect, useRef, useState } from "react";
+import React, { RefObject, useEffect, useState } from "react";
 import classnames from "classnames";
 import useEventListener from "@use-it/event-listener";
+import { createPortal } from "react-dom";
+import { usePopper } from "react-popper";
 import { AnyOption, Option } from "./Option";
 import styles from "./Autocomplete.css";
 import { Text } from "../Text";
@@ -16,6 +18,7 @@ interface MenuProps {
   readonly visible: boolean;
   readonly options: Option[];
   readonly selectedOption?: Option;
+  readonly attachTo: RefObject<Element | null>;
   onOptionSelect(chosenOption: Option): void;
 }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -176,3 +160,18 @@ function isGroup(option: AnyOption) {
   if (option.options) return true;
   return false;
 }
+
+function useRepositionMenu(attachTo: MenuProps["attachTo"]) {
+  const [positionElement, setPositionedElementRef] =
+    useState<HTMLElement | null>();
```
MichaelParadis, line 166: Shouldn't this be `useRef` instead?

darryltec: probably. I copied this from another place where it actually needs to use a state since it gets unmounted and mounted back in when it shows up. But this whole component gets unmounted anyway. so probably.

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -16,6 +18,7 @@ interface MenuProps {
   readonly visible: boolean;
   readonly options: Option[];
   readonly selectedOption?: Option;
+  readonly attachTo: RefObject<Element | null>;
```
MichaelParadis, line 21: I feel like there should be a comment for this prob to explain what this does

packages/components/src/Autocomplete/Menu.tsx
MichaelParadis: `useRef` breaks it as we saw

MichaelParadis (approved):
LGTM!

PR #1201, opened 2023-04-24
build: Enable the use of npm pack

darryltec:
## Motivations

Found out that `npm pack` is a thing and it gives you a local version of what gets released in npm. Kinda like prereleases, but local.

### Recommended way

This is documented in the read me

```
npm run pack @jobber/components
```

Would run the bootstrap first to ensure you have all the changes before packing. This is good for the times you updated design or hooks so it gets packed up with components.

The issue however is it takes a while since it has to bootstrap first.

### Only when you know what you're doing

If you're just changing one of the packages, you can pack that directly without the bootstrap

```
npm pack -w @jobber/components
```

### Added

- Add `prepack` and `postpack` scripts to add entry points when packing the package.

### Changed

- cleanup scripts in components package.json to be self documenting

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- run `npm run pack -w @jobber/components`
  - It should generate a `.tgz` file
- install file in your project `npm i path/to/file/generated-file-name.tgz`

---

Random photo of Atlantis


packages/components/package.json
```diff
@@ -6,10 +6,14 @@
...
   "scripts": {
     "build": "rollup --config",
-    "clean": "rm -rf dist/* tsconfig.tsbuildinfo",
-    "prepublishOnly": "ts-node --project ../../tsconfig.bin.json ../../scripts/entryPoints.ts",
     "bootstrap": "npm run clean; npm run build",
-    "postpublish": "ts-node --project ../../tsconfig.bin.json ../../scripts/entryPoints.ts clean"
+    "clean": "rm -rf dist/* tsconfig.tsbuildinfo",
+    "entryPoints:build": "ts-node --project ../../tsconfig.bin.json ../../scripts/entryPoints.ts",
+    "entryPoints:clean": "npm run entryPoints:build clean",
+    "prepack": "npm run entryPoints:build",
```
MichaelParadis, line 13: I think prepack needs to be added to `@jobber/hooks` now that we changed how they are built too

MichaelParadis (changes requested):
Minor change

packages/components/package.json
darryltec: oh shit you right!

MichaelParadis (approved):
LGTM!

PR #1209, opened 2023-04-25
feat: Create InputPhoneNumber

darryltec:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -0,0 +1,49 @@
...
+      | "prefix"
+      | "suffix"
+    > {
+  readonly value: string;
+  readonly onChange: (value: string) => void;
+
+  /**
+   * Shows a "required" validation message when the component is left empty.
+   */
+  readonly required?: boolean;
```
darryltec, line 24: Surprised that we don't do this and force people to write their own message all the time. Like if we can have min/max that adds validation, why shouldn't we do this?

rodrigoeidelvein:
Should we add a maximum length to it? Maybe have a default value and allow it to be configurable as a prop (just an idea) so we don't get into states like this:

image

darryltec:
@rodrigoeidelvein that is actually a valid state. Phone numbers are just so wildly different that it's much safer to keep it open. Tho [E.164](https://en.wikipedia.org/wiki/E.164#:~:text=Plan%2Dconforming%20telephone%20numbers%20are%20limited%20to%20only%20digits%20and%20to%20a%20maximum%20of%20fifteen%20digits) states that up to 15 digits max.

rodrigoeidelvein:
I got into this weird state when using `InputPhoneNumber` in a `InputGroup`:
image

darryltec:
@rodrigoeidelvein Yah, that'll happen because mdx is not throwing an error and it's acting like it's an uncontrolled component.

image

on a real project,  ts would've thrown an error and stopped compiling

packages/components/src/InputPhoneNumber/InputPhoneNumber.stories.mdx
```diff
@@ -0,0 +1,78 @@
...
+          value={value}
+          onChange={setValue}
+        />
+      );
+    }}
+  </ExampleWithHooks>
+</Canvas>
+
+## Dealing with existing data
+
```
JeffMarvin, line 60: I wonder if we should allow the user an implicit kind of "revert" behaviour if they start editing the field when they didn't mean to.
Here's a use case I am imagining. The field has some existing data ("555-TUNA") and they accidentally edit the field (or could this even happen if they tab through?)... Currently, we would format the data with the best guess we have, but they have lost the original data. If they blurred the field, could we revert to their existing invalid data?
Maybe that makes sense after we add validation? I.e. if you first interact with the field when it contains existing, invalid data, and you edit it but don't produce valid data.

Hmm, on second thought, are we wanting to require them to enter a valid number? Is the intent of this field to only be used in contexts where we enforce "valid" inputs? 

Going to drop this note here, but happy to discuss. I could see a potential future where we want formatting but not strict validation. But maybe not yet... ?

darryltec: Yeah, left it as is since it will be used on newer features! Older features tho, it will be a bit problematic when it suddenly changes and you have a non-number value in there.

It only autoformats it when you do try and change the value. Focus and blur won't trigger it.

> Hmm, on second thought, are we wanting to require them to enter a valid number? Is the intent of this field to only be used in contexts where we enforce "valid" inputs?

Yes. We want phone numbers that consists of numbers

packages/components/src/InputPhoneNumber/InputMask.tsx
```diff
@@ -0,0 +1,104 @@
+import React, { ReactElement, cloneElement, useState } from "react";
+import styles from "./InputMask.css";
+import { FormFieldProps } from "../FormField";
+
+export interface InputMaskProps {
+  /**
+   * Determines the masking pattern. By default, a `*` is used to be the value
+   * you're expecting the user would type.
+   */
```
JeffMarvin, line 9: Is this a regex? A string we treat specially? Consider being more precise and specific about the pattern format. 
Also, could we add an example usage in this comment?

packages/components/src/InputPhoneNumber/InputPhoneNumber.stories.mdx
```diff
@@ -0,0 +1,78 @@
...
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
+import { ExampleWithHooks } from "mdxUtils/ExampleWithHooks";
+import { InputPhoneNumber } from "@jobber/components/InputPhoneNumber";
+
+<Meta title="Components/InputPhoneNumber" component={InputPhoneNumber} />
+
+# InputPhoneNumber
+
+Automatically format your user's phone number as they type.
+
```
JeffMarvin, line 11: Maybe we should add a warning about the current scope/design/assumptions about country code... that we're not including handling of it for now... and that we're assuming North American phone formats only...

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  return (
+    <InputMask pattern={pattern} strict={false}>
+      <FormField
+        {...props}
+        type="tel"
+        validations={{
+          required: {
+            value: Boolean(required),
+            message: `${placeholder || "This"} is required`,
```
JeffMarvin, line 42: I like this in general, no notes to change it. I also look forward to new and wonderful ways that this breaks unintentionally. (E.g. "Please enter your phone number is required")

darryltec: ha! I was gonna say "good thing we have rules!" but I don't think we have one.

packages/components/src/InputPhoneNumber/InputMask.tsx
darryltec: Hmmm I am picking up what you're putting down.

I'll add an example since this needs to be paired with the pattern to actually understand what's up haha

```tsx
<InputMask pattern="(n*n=n)" delimiter="n" />
```

Typing `122` on the field would give you a value of `1*2=2`

packages/components/src/InputPhoneNumber/InputPhoneNumber.stories.mdx
darryltec: Tru tru. Should it be up top or here?
image
https://feat-input-phone-number.atlantis.pages.dev/?path=/docs/components-inputphonenumber--input-phone-number#with-country-code-as-a-prefix

packages/components/src/InputPhoneNumber/InputMask.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  function getMaskingInfo(value: string) {
+    const patternChars = pattern.split("");
+    const specialChars = patternChars.filter(char => char !== delimiter);
+
+    const cleanValueChars = value
+      .split("")
+      .filter(char => !specialChars.includes(char))
+      // Since this is only used in phone number, we can restrict it to
```
JeffMarvin, line 84:
```suggestion
      // Since this is only used in InputPhoneNumber, we can restrict it to
```

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
JeffMarvin: Yeah, I like it - I wonder if the "automatic" validation error message is why we didn't do it (see my comment below). Let's see how it works here and then move the pattern over to other inputs if it works?

packages/components/src/FormField/FormField.css
```diff
@@ -135,13 +139,19 @@
   position: relative;
 }
```
JeffMarvin, line 141: I get why/how we're using this for the InputMask but it feels a bit too generic for all "children" of FormField. (Not that they are used any other way right now, though... but it just feels like a detail we could forget or could be confusing a little down the road when someone else works in here.) 
I don't have any specific recommendations, other than maybe using the particular positioning here only for specific children or something?

packages/components/src/InputPhoneNumber/InputMask.tsx
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export function InputMask({
+  children,
+  delimiter = "*",
+  pattern,
+  strict = true,
+}: InputMaskProps) {
+  const { value: inputValue, onChange } = children.props;
+  const [hasMask, setHasMask] = useState(!inputValue);
```
JeffMarvin, line 39: I think I understand what this line is doing (well enough, lol) but a comment here might make this a bit quicker for devs to understand.

packages/components/src/InputPhoneNumber/InputMask.tsx
```diff
@@ -0,0 +1,104 @@
...
+  readonly delimiter?: string;
+
+  /**
+   * Allow/prevent users adding from more value than the desired pattern.
+   * For example, your pattern could accept 10 characters. If it's strict, your
+   * users can only type 10 characters.
+   *
+   * @default true
+   */
+  readonly strict?: boolean;
```
JeffMarvin, line 27: `strict` can mean a lot in this context - consider using something more specific to what this currently is doing, e.g. `limitToPatternLength` ?? naming is hard... `restrictLengthToPattern`? You get the idea.

packages/components/src/InputPhoneNumber/InputPhoneNumber.stories.mdx
JeffMarvin: I think up top, for anyone "first discovering we have this" / "considering this for use in their feature".

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
JeffMarvin: This is a fine thing for us to be opinionated about, so having labels be a noun of the "thing" being inputted is ok. I wonder if we have anywhere we can/should communicate that?

JeffMarvin (changes requested):
Great stuff! 
I didn't look at the tests yet, but had a bunch of comments/suggestions that I wanted to get over to you.

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
darryltec: yeah, considering placeholder/label is a pretty big part of input, I think it would be rare to see this (_pun intended_). I think we can start with design and see what they think

packages/components/src/FormField/FormField.css
darryltec: Hmmm, lemme think about that tho. On one hand `position:relative` makes it more clear when you're dealing with an absolute child (which is where relative really needs to exist) since it'll be contained within it. Rather than it not existing and your child is floating outside the bounds (sometimes off screen) of what you're expecting.

packages/components/src/InputPhoneNumber/InputMask.tsx
darryltec: yeah, trying to get a semantic/declarative naming out of it too rather than an imperative one. Something to think about tho 🤔

packages/components/src/InputPhoneNumber/InputPhoneNumber.stories.mdx
darryltec: > and that we're assuming North American phone formats only...

I need to confirm this from someone who is actually not in NA. iOS and Android seems to not care about the format and both chose `(***) ***-****` as _the_ format for everyone. Changing your phone's region, which changes a lot of your locale settings, somehow keeps the phone format the same. So this might just be a classic, if it's good enough for Apple and Google, it's good enough for us!

packages/components/src/InputPhoneNumber/InputMask.tsx
```diff
@@ -0,0 +1,116 @@
...
+  children,
+  delimiter = "*",
+  pattern,
+  strict = true,
+}: InputMaskProps) {
+  const { value: inputValue, onChange } = children.props;
+  const [isMasking, setIsMasking] = useState(!inputValue);
+  const stringifiedValue = String(inputValue || "");
+  const placeholderValue = pattern
+    .replace(new RegExp(`\\${delimiter}`, "g"), "_")
```
MichaelParadis, line 54: Since this is only for InputPhoneNumber this isn't needed but if we add masking to other inputs we should probably allow consumers to change the placeholder character

darryltec: the prop doc explains that further and we do want this to be used by any input eventually

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
darryltec: It adds a validation https://github.com/GetJobber/atlantis/blob/2fc03fc89ff6ef3f187f75da59c8bfaadf40f7a8/packages/components/src/InputNumber/InputNumber.tsx#L67-L85

Think of it as a super quick consistent messaging thing for everything and you only ever reach for validation prop when you're doing some crazy validations

MichaelParadis: Ahh then ignore my comment

packages/components/src/InputPhoneNumber/InputMask.tsx
```diff
@@ -0,0 +1,116 @@
...
+    .slice(stringifiedValue.length);
+
+  const inputMask = (
+    <div className={styles.mask} aria-hidden="true">
+      <span className={styles.hiddenValue}>{stringifiedValue}</span>
```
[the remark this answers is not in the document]
darryltec: it's to space the mask out correctly since every character has a different width, it's much easier to add an invisible spacer with the same value as the inout. Run the code and try removing it

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -0,0 +1,49 @@
...
+    Pick<
+      FormFieldProps,
+      | "autocomplete"
+      | "onEnter"
+      | "onFocus"
+      | "onBlur"
+      | "validations"
+      | "readonly"
+      | "prefix"
+      | "suffix"
```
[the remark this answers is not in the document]
darryltec: oh true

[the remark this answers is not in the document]
darryltec: oh, name is part of the CommonFieldProps so it should already have it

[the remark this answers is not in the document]
MichaelParadis: Ahh, good catch, see it

packages/components/src/InputPhoneNumber/InputMask.tsx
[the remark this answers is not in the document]
MichaelParadis: Ahh, that explains a lot. Thank you

MichaelParadis (approved):
LGTM!

packages/components/src/InputPhoneNumber/InputMask.tsx
JeffMarvin: My first instinct would be that "strict" meant it had to conform to ALL validations or else be ... blank/empty probably? 
"Strict" being false might then mean that it tried to format but allowed any input values to be set for the field. 

I wonder if there are different checks that will make up our Mask checks in general: only allow certain characters, limit to the pattern length, only allow the precise number. And then, what to do when they aren't valid: have the field be blank, have the field go to a default or previous value (e.g. date pickers usually default to today or the original input date if they can't parse the input to a date). We might want to have more of a config object (almost like the validations themselves) to allow for flexible specification. 

That's a lot of thoughts beyond what we need right now. But it still makes me think that `strict` isn't quite the prop we want here.

darryltec: i'm thinking it would be private to form field. No one should really be using `InputMask` or `FieldMask` on its own. It's a baked-in component inside the `FormField`

packages/components/src/FormField/FormField.css
JeffMarvin: I trust your call here. :)

packages/components/src/InputPhoneNumber/InputMask.tsx
JeffMarvin: This looks great, thanks!

packages/components/src/InputPhoneNumber/InputPhoneNumber.stories.mdx
JeffMarvin: Interesting! That would be excellent (and not entirely surprising) -- it was pre-internet when I was travelling in Europe and maybe things have standardized since then!

packages/components/src/InputPhoneNumber/InputMask.tsx
```diff
@@ -4,14 +4,26 @@ import { FormFieldProps } from "../FormField";
...
    * Change the delimiter when you need to have a `*` as a value that the input
-   * returns.
+   * returns. For example, you want your pattern to have a `*` in it.
+   *
+   * ```
+   * <InputMask pattern="n*n=n" delimiter="n" />
+   * ```
+   *
+   * It would now replace `n` with the value you type and end up with `1*2=2` if
+   * you give a value of `122`.
```
JeffMarvin, line 26: 🎉

packages/components/src/InputPhoneNumber/InputMask.tsx
darryltec: Would you be okay if we open this conversation again on the "implement on formfield" task? Or a follow up. I think this component will evolve the more we go through improving it (adding validations, changing the UX a bit)

packages/components/src/InputPhoneNumber/InputMask.tsx
```diff
@@ -48,7 +48,7 @@ export function InputMask({
   strict = true,
 }: InputMaskProps) {
   const { value: inputValue, onChange } = children.props;
-  const [hasMask, setHasMask] = useState(!inputValue);
+  const [isMasking, setIsMasking] = useState(!inputValue);
```
JeffMarvin, line 51: Excellent. Clear variable names > comments!

packages/components/src/InputPhoneNumber/InputMask.tsx
JeffMarvin: Yup, absolutely. We won't get it right just by pondering ideas, we can iterate as we go! :)

JeffMarvin (approved):
Great foundation for masking here, and lots of thought to making this work well out of the box without having to be too complex/elaborate. Nice work!

packages/components/src/InputPhoneNumber/InputMask.tsx
darryltec: yea! I like this kind of comment! makes me really think about what I'm doing

PR #1215, opened 2023-04-26
fix(components): calculate Z index for autocomplete 

Brett-A-T-Anderson:
## Motivations

There was an issue where the autocomplete options would appear behind the modal. Screenshot: 
image

## Changes

Now it will appear in front of the modal so it can be selected. This is a quick fix so the component can be used. 

### Security

- 

## Testing

- Go to anchor as test@example.com
- Go to an account 
- Go to update account
- Scroll down to the distribution channel component 
image
- Attempt to type in there, and you'll see it appear in front now. 

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Autocomplete/Autocomplete.css
+++ b/packages/components/src/Autocomplete/Autocomplete.css
@@ -6,7 +6,7 @@
 .options {
   display: none;
   position: absolute;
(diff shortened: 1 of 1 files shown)


MichaelParadis (approved):
LGTM!

PR #1217, opened 2023-04-27
chore(hooks): Check if dropping hooks dev dependency version works [JOB-67899]

MichaelParadis:
## Motivations

This PR downgrades the `apollo/client` version to match jobber-mobile at `3.7.3`.  

We should test `hooks` in jobber-mobile and JO to be sure that in doing this version downgrade, `hooks` intended behaviour continues!

### Deprecated

`@apollo/client` was downgraded from `3.7.10` to `3.7.3`

## Testing

Places where Atlantis Hooks are being used in JO and jobber-mobile should be tested to make sure nothing is broken with hooks.

## ✨  Use these for testing:

jobber-mobile PR here with pre-release installed: https://github.com/GetJobber/jobber-mobile/pull/7216

Android Bitrise Build here for testing: https://github.com/GetJobber/jobber-mobile/pull/7216#issuecomment-1527763167

Jobber online PR here with pre-release installed: https://github.com/GetJobber/Jobber/pull/35373

---

Random photo of Atlantis

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   },
   "homepage": "https://github.com/GetJobber/atlantis#readme",
   "devDependencies": {
-    "@apollo/client": "^3.7.10",
+    "@apollo/client": "^3.7.3",
     "@babel/core": "^7.4.5",
     "@babel/preset-env": "^7.4.5",
     "@babel/preset-react": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   },
   "homepage": "https://github.com/GetJobber/atlantis#readme",
   "devDependencies": {
-    "@apollo/client": "^3.7.3",
+    "@apollo/client": "^3.7.10",
     "@babel/core": "^7.4.5",
     "@babel/preset-env": "^7.4.5",
     "@babel/preset-react": "^7.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,7 @@
         "jest-environment-jsdom": "^29.5.0"
       },
       "devDependencies": {
-        "@apollo/client": "^3.7.10",
+        "@apollo/client": "^3.7.3",
         "@babel/core": "^7.4.5",
         "@babel/preset-env": "^7.4.5",
         "@babel/preset-react": "^7.0.0",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
   },
   "homepage": "https://github.com/GetJobber/atlantis#readme",
   "devDependencies": {
-    "@apollo/client": "^3.7.10",
+    "@apollo/client": "^3.7.3",
```
darryltec, line 56: don't really need to downgrade this, no? Like the peer deps said 3.7.3 and above? or are we saying only 3.7.3?

MichaelParadis: No technically but I did so to keep it consistent with the peer dependency so we can actually verify that it will work within Atlantis instead of building something using a newer version that breaks on older versions. I can revert though

darryltec: but JO is 3.7.10

darryltec: also, tbf, this would eventually install 3.7.13 on other projects when npm starts deduping them

MichaelParadis: @darryltec I can undo the change here. 
Looking at the following libraries it seems like they don't pin patch versions and will generally unpin (*) or have it set to a major version. See below:

1. [Reanimated](https://github.com/software-mansion/react-native-reanimated/blob/main/package.json#L86-L94)
2. [Flashlist](https://github.com/Shopify/flash-list/blob/main/package.json#L54-L57)
3. [Popper](https://github.com/floating-ui/react-popper/blob/master/package.json#L60-L63)
4. [React Navigation Native stack](https://github.com/react-navigation/react-navigation/blob/main/packages/native-stack/package.json#L59-L64)
5. [React Navigation drawer](https://github.com/react-navigation/react-navigation/blob/main/packages/drawer/package.json#L64-L71)
6. [React Navigation Stack](https://github.com/react-navigation/react-navigation/blob/main/packages/stack/package.json#L61-L67)
Thoughts on just having it be `^3.0.0` to match the others?

darryltec: That's fine by me!

MichaelParadis: Sounds good, I will push the change right now

MichaelParadis: @darryltec updated

darryltec (approved):
LGTM

PR #1218, opened 2023-04-27
feat: Create package for mobile components

rodrigoeidelvein:
## Motivations
To unify our design system, we'll now have a package for the mobile components! For this first version, I imported `Icon` and `Divider`. 

### Added

- `components-mobile` package!
- `Icon` and `Divider` components

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

You can pack this package and install it on jobber mobile.
Assuming you have `atlantis` and `jobber-mobile` in the same folder:
On atlantis root:
`npm pack -w @jobber/components-native`

After packing it, you can run on jobber-mobile root:
`npm install ../atlantis/jobber-components-native-0.0.0.tgz`

Also, I created a draft PR with the prerelease installed in jobber-mobile: https://github.com/GetJobber/jobber-mobile/pull/7212 and there you can download the builds from Bitrise

---

Random photo of Atlantis


docs/guides/index.tsx
darryltec, line 1: I know i added this, but we don't need this. It's for testing purposes 😆

.eslintrc.js
```diff
@@ -3,12 +3,14 @@ require("@jobber/eslint-config/patch-eslint-plugin-resolution.js");
 
 const packageAliases = [
   ["@jobber/components", "./packages/components/src"],
+  ["@jobber/components-native", "./packages/components-native/src"],
```
darryltec, line 6: are we going for 

```js
import { Icon } from "@jobber/components-native"
```

if so, we don't need this.

(PS: I also added this, I'm sorry 😬 )

package.json
```diff
@@ -60,6 +61,7 @@
     "@babel/preset-typescript": "^7.3.3",
     "@csstools/postcss-global-data": "^1.0.3",
     "@jobber/components": "*",
+    "@jobber/components-native": "*",
```
darryltec, line 64: we probably shouldn't install it yet since we don't use em right now.

packages/components-native/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Usage
+
+Import and render an Atlantis component:
+
+```tsx
+import React from "react";
+import { Icon } from "@jobber/components-native";
+
+export default function MyComponent() {
```
darryltec, line 28:
```suggestion
export function MyComponent() {
```

because we don't really do default in the jobber

packages/components-native/src/Divider/Divider.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { View } from "react-native";
+import { styles } from "./Divider.style";
+
+interface DividerProps {
+  /**
+   * The weight of the divider.
+   *
+   * @default "base"
+   */
+  readonly size?: "base" | "large" | "largest";
```
darryltec, line 11: I wonder if

```suggestion
  readonly size?: keyof typeof styles;
```

would work here 🤔

tsconfig.json
darryltec, line 1: I think you can revert the changes on this file again. it was added to support `@jobber/components-native/Icon` which we don't do anymore, yeah?

darryltec (commented):
Few little bits for you mah dude. Mostly reversing the silly things I added haha

rodrigoeidelvein:
I got into this weird state when using `InputPhoneNumber` in a `InputGroup`:
image

package.json
MichaelParadis: Do we need `esm` for `components-native`? Wouldn't this affect the storybook stuff if it isn't installed?

darryltec: it won't since we don't reference it in the docs. It's esm but still a js file not mjs.

darryltec: as for

> Do we need esm 

If you look at react native's package, they use esm instead of common js. Not sure of RN tree shake things but esm is what's tree shakeable

packages/components-native/src/Divider/Divider.tsx
MichaelParadis: That would key rid of `base` and add `divider`. We may be able to do something like
```
export const styles = StyleSheet.create({
  divider: {
    height: JobberStyle["space-minuscule"],
    margin: 0,
    borderBottomWidth: JobberStyle["border-base"],
    borderBottomColor: JobberStyle["color-border"],
  },
  sizes: {
    large: {
      borderBottomWidth: JobberStyle["border-thick"],
      opacity: 0.875,
    },
    largest: {
      borderBottomWidth: JobberStyle["space-small"],
      opacity: 0.375,
    },
  }
});
```

```
readonly size?: keyof typeof styles.sizes;
```
Though

package.json
MichaelParadis: Ahh, good point the storybook stuff is in a separate issue

Also makes sense about the RN package stuff. Thanks for the clarification

package.json
```diff
@@ -113,6 +115,7 @@
     "prop-types": "^15.7.2",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-native": "^0.71.7",
```
MichaelParadis, line 118: I think we should match the Jobber mobile version instead of this version

packages/components-native/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "@types/node": "*",
+    "@types/react": "*",
+    "@types/react-native": "*",
+    "metro-react-native-babel-preset": "^0.76.0",
+    "typescript": "*"
+  },
+  "peerDependencies": {
+    "@babel/core": "*",
+    "react": "*",
+    "react-native": "*"
```
MichaelParadis, line 38: I am hesitant about an unpinned dependency here. If when packaging it grabs from the main package.json this is fine but if not we should probably have a version specified

packages/components-native/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "dependencies": {
+    "@jobber/design": "*",
+    "react-native-svg": "^12.4.4",
+    "ts-xor": "^1.1.0"
+  },
+  "devDependencies": {
+    "@types/node": "*",
+    "@types/react": "*",
+    "@types/react-native": "*",
```
MichaelParadis, line 31: Same hesitation as below types that don't match the RN version being used can cause issues. If it would grab it from the parent package.json I would specify a version there

The same applies for React versions here too

packages/components-native/package.json
MichaelParadis: Also applies to the react version

packages/components-native/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "metro-react-native-babel-preset": "^0.76.0",
+    "typescript": "*"
+  },
+  "peerDependencies": {
+    "@babel/core": "*",
+    "react": "*",
+    "react-native": "*"
+  },
+  "eslintIgnore": [
```
MichaelParadis, line 40: Isn't this not needed because of the `.eslintignore` file in the main repo?

packages/components-native/tsconfig.types.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist/types",
```
MichaelParadis, line 4: I think this file can be moved into the one `tsconfig` file using this tsconfig option
https://www.typescriptlang.org/tsconfig#declarationDir

packages/components-native/src/Divider/Divider.tsx
MichaelParadis: This would work though

```diff
diff --git a/packages/components-native/src/Divider/Divider.style.ts b/packages/components-native/src/Divider/Divider.style.ts
index 72c1d6ce..001d218e 100644
--- a/packages/components-native/src/Divider/Divider.style.ts
+++ b/packages/components-native/src/Divider/Divider.style.ts
@@ -9,6 +9,9 @@ export const styles = StyleSheet.create({
     borderBottomWidth: JobberStyle["border-base"],
     borderBottomColor: JobberStyle["color-border"],
   },
+});
+
+export const sizeStyles = StyleSheet.create({
   large: {
     borderBottomWidth: JobberStyle["border-thick"],
     opacity: 0.875,
diff --git a/packages/components-native/src/Divider/Divider.tsx b/packages/components-native/src/Divider/Divider.tsx
index 03ebf72f..efe09d06 100644
--- a/packages/components-native/src/Divider/Divider.tsx
+++ b/packages/components-native/src/Divider/Divider.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { View } from "react-native";
-import { styles } from "./Divider.style";
+import { sizeStyles, styles } from "./Divider.style";
 
 interface DividerProps {
   /**
@@ -8,14 +8,14 @@ interface DividerProps {
    *
    * @default "base"
    */
-  readonly size?: "base" | "large" | "largest";
+  readonly size?: "base" | keyof typeof sizeStyles;
 }
 
 export function Divider({ size = "base" }: DividerProps): JSX.Element {
   const style = [
     styles.divider,
-    size === "large" && styles.large,
-    size === "largest" && styles.largest,
+    size === "large" && sizeStyles.large,
+    size === "largest" && sizeStyles.largest,
   ];
   return <View testID="Divider" style={style} />;
 }

```

rodrigoeidelvein: What if we just change the `divider` property to `base`? See the new changes

package.json
rodrigoeidelvein: Good idea!

packages/components-native/tsconfig.types.json
rodrigoeidelvein: I also removed `tsconfig.esm.json`.

packages/components-native/package.json
rodrigoeidelvein: Do you think we should put the same version as we have in jobber-mobile?

MichaelParadis: I think so. We may want a couple patch versions below but I think matching jobber-mobile for now is good. We don't really want to have things be broken on other versions

rodrigoeidelvein: I got an peer dependency issue because `react-native@0.69.10` (version installed on jobber-mobile), has a peer dependency on `react@18.0.0` and we currently have `react@18.2` installed.

I think it's safe for us to use the latest version in `components-native`. I added a minimum peer dependency version to be the same one as jobber-mobile

packages/components-native/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "./dist",
```
MichaelParadis, line 4:
```suggestion
    "outDir": "./dist/esm",
```
Or update the package.json

rodrigoeidelvein: Good catch! I updated the `package.json`

packages/components-native/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Octopusapp Inc.
```
MichaelParadis, line 3: Year should be updated, however they all should be so this one can be done and we can update the other ones in another PR

MichaelParadis (approved):
LGTM!

PR #1225, opened 2023-05-04
fix(components): fix the styling of DescriptionList component

jeyhuno:
## Motivations
Currently DescriptionList component width for `key` is `60%` and width for `value` is `40%`. I believe it should be the opposite since usually the keyse are shorter than values

## Changes
Updated `dd` element width from `40%` to `60%`
Updated `dt` element width from `60%` to `40%`

Before:
Screenshot 2023-05-04 at 11 41 37 AM

After:
Screenshot 2023-05-04 at 11 41 50 AM


darryltec (approved):
LGTM, thanks @jeyhuno

PR #1235, opened 2023-05-10
feat(components-native): Import Typography to components-native

rodrigoeidelvein:
## Motivations
Import `Typography` to the package `components-native` .

### Added

- Create Typography in components-native

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


darryltec (changes requested):
few lil bits for you

rodrigoeidelvein:
The pre-release version:
`@jobber/components-native@0.1.3-pree.2+23faa88d`

To install the new version run:
`npm install @jobber/components@4.14.1-import-typ.5+4700e8f4`

packages/components-native/package.json
```diff
@@ -22,7 +22,8 @@
   },
   "dependencies": {
     "@jobber/design": "^0.38.2",
-    "react-native-svg": "^12.4.4",
+    "react-native-gesture-handler": "^2.10.0",
+    "react-native-svg": "^13.9.0",
```
darryltec, line 26: I'm guessing this is trying to match the mobile repo?

packages/components-native/src/Typography/Typography.style.ts
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * Reusable typography tokens to ensure consistency for any client facing texts.
+ */
+export const typographyStyles: { [index: string]: TextStyle } =
+  StyleSheet.create({
+    // This follows a pattern of
+    // { fontFamily }{ fontStyle }{ fontWeight }
+    baseRegularRegular: {
+      fontFamily: "inter-regular",
+    },
```
darryltec, line 33: Oh, you do need to export this on its own const like before since it's not gonna work for web and storybook due to some fun react native performance thing

darryltec: now that I think about it, maybe this should be under utils (typographyTokens) since it's shared through a lot of other components

darryltec (changes requested):
one last thing!

packages/components-native/package.json
rodrigoeidelvein: Exactly

packages/components-native/src/Typography/Typography.style.ts
rodrigoeidelvein: So, we need to export both of them? The object with the values and the const with `Stylesheet.create`?

darryltec (approved):
LGTM

PR #1241, opened 2023-05-11
fix(components): Fix Select component hit area

rodrigoeidelvein:
## Motivations

The `Select` component what not opening the options if the user clicked on the label. With this fix, I created a class for the select, similar to what we do with textarea and changed the z-index of the input that was behind the label, making it impossible to trigger the select by clicking on the label.

### Added

- 

### Changed

- Add a class specific for the select and changed its z-index

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -177,6 +177,10 @@
   resize: vertical;
 }
 
+.select .input {
+  z-index: var(--field--label-elevation);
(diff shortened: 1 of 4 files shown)


taylorvnoj (approved):
Nice solution, works beautifully 🚀  LGTM

PR #1247, opened 2023-05-12
refactor(components-native): Add tests to components-native's Typography

rodrigoeidelvein:
## Motivations
Adding tests to `Typography` on components-native.

I had to install the same version of `react-native-gesture-handler` that we're using on jobber-mobile because they introduced a rule that doesn't make sense for us. 

https://github.com/software-mansion/react-native-gesture-handler/commit/ad626405d966b14966c14d45a0a0c141ed6251e6#diff-10cc5c6712f3dadbd9f8ff4fd42423bf6aa26d6678a909ea908971826e7b115bR403-R409

Now, `GestureDetector` needs to be wrapped on a `GestureHandlerRootView`, but it's already being declared in the root of the app, we don't need it on a simple `Typography` component. Anyways, just venting.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37771,22 +37771,6 @@
         "nullthrows": "^1.1.1"
       }
     },
-    "node_modules/react-native-gesture-handler": {
-      "version": "2.10.0",
-      "resolved": "https://registry.npmjs.org/react-native-gesture-handler/-/react-native-gesture-handler-2.10.0.tgz",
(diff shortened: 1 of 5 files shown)


taylorvnoj (approved):
YUP :angry-corgi: 🚀 🚢

PR #1248, opened 2023-05-12
feat(components-native): Import Text to components-native

taylorvnoj:
## Motivations

Import `Text` to `components-native`

### Added

 
- `Text` RN component now exists
- test & snapshot files also added

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/Text/index.ts
+++ b/packages/components-native/src/Text/index.ts
@@ -1,0 +1,1 @@
+export { Text } from "./Text";
--- a/packages/components-native/src/Text/index.ts
+++ b/packages/components-native/src/Text/index.ts
@@ -1,1 +1,1 @@
-export { Text } from "./Text";
+export { Text, TextLevel } from "./Text";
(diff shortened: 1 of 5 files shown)


packages/components-native/src/Text/index.ts
```diff
@@ -0,0 +1 @@
+export { Text } from "./Text";
```
rodrigoeidelvein, line 1:
```suggestion
export { Text, TextLevel } from "./Text";
```

There's a place in jobber-mobile where it imports TextLevel, we should export it too.

rodrigoeidelvein (changes requested):
Just a small thing

rodrigoeidelvein (changes requested):
Just a small thing

packages/components-native/src/Text/index.ts
taylorvnoj: Great call out!

rodrigoeidelvein (approved):
LGTM!

PR #1249, opened 2023-05-12
feat: inline eslint module resolution

seanhealy:
## Motivations

Make using shared eslint slightly easier.

You may want to double check but for us it's working to require the eslint module resolution patch in the eslint config that's being shared rather than needing to do it everywhere that is consuming it.

### Added

- 

### Changed

- Move eslint patching into the core shared eslint config so you don't need to add the `require` line everywhere you want to use it.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,3 @@
-// eslint-disable-next-line import/no-internal-modules
-require("@jobber/eslint-config/patch-eslint-plugin-resolution.js");
-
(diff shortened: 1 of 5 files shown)


seanhealy:
Hmmm… not sure what is different but it seems this approach isn't working here. 👋

PR #1253, opened 2023-05-12
feat(components-native): Import ActivityIndicator to components-native

rodrigoeidelvein:
### Added

- `ActivityIndicator` component in components-native

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/index.ts
+++ b/packages/components-native/src/index.ts
@@ -2,3 +2,4 @@ export * from "./Icon";
 export * from "./Divider";
 export * from "./Typography";
 export * from "./Text";
+export * from "./ActivityIndicator";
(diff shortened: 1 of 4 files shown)


packages/components-native/src/index.ts
```diff
@@ -2,3 +2,4 @@ export * from "./Icon";
 export * from "./Divider";
 export * from "./Typography";
 export * from "./Text";
+export * from "./ActivityIndicator";
```
taylorvnoj, line 5: Do we need to export the ActivityIndicatorProps also?

taylorvnoj (changes requested):

packages/components-native/src/index.ts
rodrigoeidelvein: We have a wrapper for react-native's `ActivityIndicator` to make it use grey blue as the default color. We don't need to export `ActivityIndicatorProps` as it comes from `react-native`

taylorvnoj (approved):
LGTM

PR #1254, opened 2023-05-15
feat(components): Allow GridCell' s to only be usable in Grids

MichaelParadis:
### Added

- 

### Changed

- Made it so `Grid.Cell` can only be used inside of a grid component

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Install pre-release.
- Add `Grid.Cell` that is not wrapped in a `Grid`. Verify an error occurs
- Add `Grid.Cell` that is wrapped in a `Grid`. Verify an error doesn't occur.

---

Random photo of Atlantis


darryltec:
Oooh, that's a cool approach. What happens when you do it like so?

```html
<Grid>
  <Grid.cell>
    <Grid.cell> </Grid.cell>
  </Grid.cell>
<Grid>
```

would it accept it and not throw an error?

JeffMarvin:
@MichaelParadis Is there any functional difference/benefit to using this approach vs. the `Chip.tsx` with `useAssert` hook that we have used in the Chips component? (https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Chips/Chip.tsx)

I find that approach a bit more understandable on the surface, so I'm curious what we gain with this approach. :)

MichaelParadis:
@darryltec @JeffMarvin Updated to use the Chips style of handling it

packages/components/src/Grid/GridCell.tsx
```diff
@@ -0,0 +1,16 @@
+// Only need the component to pass in the data as a react children
+
+import { useAssert } from "@jobber/hooks/useAssert";
+import React, { PropsWithChildren } from "react";
+import { GridCellProps } from "./InternalGridCell";
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function GridCell(props: PropsWithChildren<GridCellProps>) {
+  // Throw error when <Chip /> gets used outside of the <Chips />
```
darryltec, line 9: Oopsies 

```suggestion
  // Throw error when <Grid.Cell /> gets used outside of the <Grid />
```

packages/components/src/Grid/Grid.tsx
```diff
@@ -36,7 +37,9 @@ export function Grid({
 
   return (
     <div data-testid={GRID_TEST_ID} className={classnames}>
-      {children}
+      {React.Children.map(children, gridCell => {
+        return <InternalGridCell {...gridCell.props} />;
```
darryltec, line 41: nit

```suggestion
      {React.Children.map(children, { props } => {
        return <InternalGridCell {...props} />;
```

darryltec (changes requested):
1 very minor oops

packages/components/src/Grid/GridCell.tsx
```diff
@@ -0,0 +1,16 @@
+// Only need the component to pass in the data as a react children
+
+import { useAssert } from "@jobber/hooks/useAssert";
+import React, { PropsWithChildren } from "react";
+import { GridCellProps } from "./InternalGridCell";
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function GridCell(props: PropsWithChildren<GridCellProps>) {
```
darryltec, line 8: underscore did not work for this eh?

```suggestion
export function GridCell(_: Props>) {
```

darryltec (approved):
:shipit:

PR #1259, opened 2023-05-16
refactor: Replace deprecated breakpoint usage with new breakpoints

JeffMarvin:
## Motivations
This refactors usage of deprecated breakpoints to use the new breakpoints. 

Since the new breakpoints define different responsive boundaries, some considerations have been made to remain compatible with existing usage in Jobber Online. 
    1. Drawer is hardcoded to use `>= 640px` instead of the new breakpoint values. This is to retain compatibility with the existing layouts which are all using the old breakpoint.
    2. Menu takes a more conservative approach, and moves from a bottom fullscreen display mode to an "inline" mode at a smaller breakpoint than before (`>=490px` instead of `>=640px`). It still uses the new breakpoints. (See screenshots below.) This is because there is plenty of room in a 500px width layout for the inline view, and menus appearing full screen and at the bottom at near 768px width layouts looked odd.

### Changed
- Storybook preview docs - show the right hand sidebar at widths `>=768px` instead of `>=1024px`
- Components that were using the deprecated breakpoint CSS variables are now using the new breakpoint values. 
  - `Drawer`
    - make the visible Drawer column wider at `>= 1440px` instead of `>= 1024px`
  -  `GridCell` corrected accidental usage of `medium-screens` - `grid-column-end` should now get proper sizing for this breakpoint
  - `Menu` - move from menu displaying at the bottom of the screen to a relative position near the button that triggered it at widths `>=490px` instead of `>= 640px`
  - `Modal` - grow left & right padding from base to large at widths  `>= 768px` instead of `>= 640px`
  - `Pagination` changed gap in pagination buttons to affect `< 490px` instead of `< 640px`

### Removed

- The deprecated breakpoint CSS variables have been removed. 

## Testing
Jobber Online has been tested for usage of these components, and adjustments have been made as noted above. 
Example of Menu appearing inline in a width <=768px

---

Random photo of Atlantis


darryltec (approved):
LGTM!

darryltec (approved):
We good!

PR #1266, opened 2023-05-17
feat(components-native): Import ProgressBar 

taylorvnoj:
## Motivations

- `ProgressBar` needs to be imported to `components-native`

### Added

- `ProgressBar` is now migrated
- `SBProvider` to help stories using `react-intl` actually show up in Storybook

### Removed

 
- removed `ProgressAppointmentTitle` from the main `ProgressBar` component; this should be a Shared Component instead so I removed the related logic from the snapshots & `ProgressBar.test.tsx`

---

Random photo of Atlantis


packages/components-native/src/ProgressBar/messages.ts
```diff
@@ -0,0 +1,29 @@
...
+  noVisitsCompleted: {
+    id: "noVisitsCompleted",
+    defaultMessage: "0 visits completed",
+    description: "Progress bar title when no visits are completed",
+  },
+  visitsComplete: {
+    id: "visitsComplete",
+    defaultMessage: "{title}",
+    description: "Progress bar title when a visit/visits are completed",
+  },
```
rodrigoeidelvein, line 18: We can remove these, they were being used by `ProgressAppointmentTitle`

packages/components-native/src/ProgressBar/ProgressBar.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const current = 1;
+  const bar = renderProgressBarComponent({
+    ...defaultSetupProps,
+    current: current,
+    total: 2,
+  });
+  expect(bar).toMatchSnapshot();
+});
+
+it("renders green CompletedProgress bar and renders 'visits' as the smart string in the progress bar title", () => {
```
rodrigoeidelvein, line 42: We should change these test names to not expect "visits" as we are not testing `ProgressAppointmentTitle` here anymore. As the snapshot is testing the color, we could add something like this below the snapshot:
`expect(bar.getByLabelText("2 of 3 complete")).toBeDefined();`

Then it will be easier to read what the test is expecting.

packages/components-native/src/ProgressBar/ProgressBar.style.ts
```diff
@@ -0,0 +1,34 @@
...
+  },
+  centeredContent: {
+    justifyContent: "center",
+    alignItems: "center",
+  },
+  spacedContent: {
+    justifyContent: "space-between",
+    alignItems: "flex-start",
+    flexDirection: "row",
+  },
```
rodrigoeidelvein, line 29: We can remove these because they were being used only by `ProgressAppointmentTitle`

rodrigoeidelvein (changes requested):
Just need to remove some leftovers from `ProgressAppointmentTitle` and we're good!

packages/components/src/Card/Card.stories.mdx
```diff
@@ -1,175 +1,327 @@
 import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
-import { Card } from "@jobber/components/Card";
-import { Text } from "@jobber/components//Text";
-import { Heading } from "@jobber/components//Heading";
-import { Content } from "@jobber/components//Content";
-import { Button } from "@jobber/components/Button";
-import { InputText } from "@jobber/components/InputText";
-import { InputGroup } from "@jobber/components/InputGroup";
+import { ActionItem, ActionItemGroup, Card } from "@jobber/components-native";
```
rodrigoeidelvein, line 2: 🤔 I'm assuming you didn't want to commit this file?

rodrigoeidelvein (changes requested):
Just a couple more things

I think you didn't want to commit the changes to `Card.stories`, it's replacing the content with the mobile stories

packages/components/src/Card/Card.stories.mdx
taylorvnoj: Oh wow - I'll see myself out 😆

rodrigoeidelvein (approved):
LGTM! :shipit:

PR #1276, opened 2023-05-19
build: Upgrade react-hook-form to v7.43.7 [JOB-65817]

taylorvnoj:
## Motivations

`react-hook-form` needs to be upgraded

## Changes

`react-hook-form` upgraded from `6.15.8` => `7.43.7`

### Changed

  
- `errors` has been moved into `formState` object

## Testing

Pre-releases available to install in JO for testing: `npm install @jobber/components@4.17.2-install-re.6+30145e27`

---

Random photo of Atlantis


MichaelParadis (commented):
Looks good but going to get another person to review my changes

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
```diff
@@ -39,7 +39,11 @@ export function InputPhoneNumber({
 }: InputPhoneNumberProps) {
   const { placeholder, validations, pattern = "(***) ***-****" } = props;
   const errorSubject = placeholder || "Phone number";
-  const getValues = useFormContext();
+  const { getValues } =
+    useFormContext() != undefined
+      ? useFormContext()
+      : // If there isn't a Form Context being provided, get a form for this field.
+        useForm({ mode: "onTouched" });
```
rodrigoeidelvein, line 46: I wonder if getting `getValues` from `useForm` is valuable here. I tried it a couple ways in Storybook and it always returned an empty object when using `InputPhoneNumber` without a form. The validation still worked when changed it to:

```ts
const { getValues } = useFormContext();
```

Maybe it's valuable for because it's using the `Controller` component.

But it's just a thought, any ideas around here? Did you find a use case where it was problematic?

rodrigoeidelvein (changes requested):
Just found a typo and a question

packages/components/src/InputPhoneNumber/InputPhoneNumber.tsx
rodrigoeidelvein: @MichaelParadis Just tagging you here as you're not the PR creator

MichaelParadis: I was seeing an error without it on in the `InputPhoneNumber` story because it isn't in a `Form` so there isn't a context available

PR #1277, opened 2023-05-22
feat(components-native): Import Flex (before it was Grid) to components-native

rodrigoeidelvein:
### Added

- `Flex` component in components-native. Before it was called `Grid`, but it actually implements Flex behind the scenes, so we will have another component to match the web `Grid` in the future.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


darryltec (changes requested):
This looks good. Tho, we should rename this to `Flex` since it isn't really a `Grid` component anymore like the one on web. Planning to create a Web counterpart for this since we need it in web as well

rodrigoeidelvein:
> This looks good. Tho, we should rename this to `Flex` since it isn't really a `Grid` component anymore like the one on web. Planning to create a Web counterpart for this since we need it in web as well

It makes sense! I'll change it

darryltec (approved):
We goooooood!

PR #1284, opened 2023-05-25
fix(components): BREAKING CHANGE: Prevent menu from appearing onChange in freeform Autocomplete fields

TJKruitbosch:
## Motivations
When a freeform Autocomplete field is modified without focusing it, due to e.g. browser address autofill, the dropdown menu still appears even though the user never focused the field. This change makes that behaviour unique to just Autocomplete fields that do not allow freeform changes.

### Before

https://github.com/GetJobber/atlantis/assets/872151/ec7ce8ed-0b28-4056-aefd-1fee220737b7

### After

https://github.com/GetJobber/atlantis/assets/872151/035d3986-a429-405b-b74f-04e6084e558f

### Removed

Autocomplete no longer shows menu options based solely on an onChange event when allowFreeform is true

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Autocomplete/Autocomplete.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.tsx
@@ -146,8 +146,9 @@ export function Autocomplete({
     updateInput(newText);
     if (allowFreeForm) {
       onChange({ label: newText });
+    } else {
+      setMenuVisible(true);
     }
-    setMenuVisible(true);
   }
 
   function handleInputBlur() {


packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -146,8 +146,9 @@ export function Autocomplete({
     updateInput(newText);
     if (allowFreeForm) {
       onChange({ label: newText });
+    } else {
+      setMenuVisible(true);
```
darryltec, line 150: Hmmm... this only removes the issue for freeform ones. The non-freeform ones would still have this issue. no?

Thinking about it more, by restricting it to non-freeform, you basically removed this functionality for the freeform ones. Which raises the question of, if it works fine for non-freeform ones, do we actually need this bit of code?

darryltec: Also yeah, looking at the code, there's a focus and blur callback. Focus shows the menu and blur hides it (or when there are no more options and you're still focused). 

Kinda weird that we have to open it when you change the value of it. If you're focused, it's already open anyway. If you're not focused, don't really need to open it. 🤔 

I think we can just remove it instead of keeping it and creating 2 code paths for it

darryltec (changes requested):
I think we can just remove this part altogether. Also, not really a breaking change, more of a bug fix 😁

TJKruitbosch:
> I think we can just remove this part altogether. Also, not really a breaking change, more of a bug fix 😁

@darryltec I don't mind taking that part off, but I'm certain it's going to break at least a few tests in the main repo. Specifically because I wrote a test recently that waits on the autocomplete menu to appear before ending the test (so Jest doesn't complain about state changes outside of a test), and I'm not going to take the bet that I'm the only one who has done that 😉. If/when this gets updated in the main Jobber repo, that update PR is gonna need to fix some things.

darryltec:
That's a fair thing to say. But, in the sense of atlantis realm, it isn't a thing that requires a major version bump. For example, if polaris/material fixes a bug that your tests rely on, would you have expected them to mark that their changes are breaking?

Also, we can try that by triggering a pre-release and testing it there to see if it does break. Instructions [here](http://localhost:6005/?path=/docs/introduction--page#pre-release) 
image

packages/components/src/Autocomplete/Autocomplete.tsx
TJKruitbosch: > The non-freeform ones would still have this issue. no?
I was uncertain if it's a problem in that case, to be honest. I assumed this code was here in the first place in case a React-controlled non-freeform field received an input change event. The `onChange` callback won't be called, so the field itself won't update its value, and now the user is wondering why "nothing" happened if there's no autocomplete menu. 

Might be a moot point though...this browser autofill is probably not going to usually pop up unless it's used in an Address Input kind of form, which I assume you'd always want to be freeform anyway?

darryltec: Nah, I tried it. it actually puts the value in there and then the component deletes them on focus. That's a bug on its own tho.

You're probably right. It won't usually show up but the browser does rely on the label of the input to determine the autocomplete-ability of it so it's hard to assume that it would only happen on address fields

TJKruitbosch: Might be outside the scope of this quick little bugfix, but for non-freeform fields maybe it should do some kind of "if the new value is in the valid list of options, fire onChange with the Option, else show some error-like user feedback"? I'm kind of uncertain about what the best UX here is for a field that only allows certain values being programmatically set to a value that may or may not be on the list of allowed ones. 

Are you still okay with treating "it shows the dropdown menu" as bug-behaviour across the board here and removing/fixing it?

darryltec: Nope. I'd treat it as a different bug. One that needs more UX thought. Like if you're gonna open the menu and clearing it, you should focus them on that field as well. 

But it isn't an intentional "I want it to be this value" but more of a side-effect of a multi-autocomplete so that'll be more confusing than just stopping the onChange from firing. Even more, it is the browser adding the value so it is a disconnect between the state of react and the HTML input. Like you know how uncontrolled components work in web where you can keep typing even without a `state` or firing `onChange`

darryltec (approved):
LGTM!

PR #1289, opened 2023-05-26
fix: enable prefix and suffix on InputNumber

darryltec:
### Added

- 

### Changed

- enabled prefix and suffix on input number because... i think we just forgot

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputNumber/InputNumber.tsx
+++ b/packages/components/src/InputNumber/InputNumber.tsx
@@ -17,6 +17,8 @@ interface InputNumberProps
       | "readonly"
       | "defaultValue"
       | "keyboard"
+      | "prefix"
+      | "suffix"
     > {
   value?: number;
 }


JeffMarvin:
👀

JeffMarvin:
LGTM - I didn't test this much (for any weird interference between anything that can be placed in the suffix and the browser's up/down UI arrows) but I think we'll discover those as usage increases. Approving because a team needs this and also because it is so simple.

JeffMarvin (approved):

darryltec:
@JeffMarvin I tested it. https://enable-prefix-suffix-on-inpu.atlantis.pages.dev/?path=/story/components-inputnumber--input-number&args=size:small;inline:true;maxLength:4;suffix.label:hrs It should be fine and not overlap since the real input is always getting pushed. No overlays here. Just pure html stacking

PR #1292, opened 2023-05-26
fix(hooks): Revert initial state on useFormState

rodrigoeidelvein:
## Motivations

We changed the initial state of `useFormState` because `react-hook-form` changed their state to always start `isValid` as `false`: https://github.com/react-hook-form/react-hook-form/blob/master/CHANGELOG.md#changed-23

The problem is that some forms in Jobber-Online uses Checkboxes that are not tied to the form state yet, then it's not triggering the state changes in the Form, causing it to break some functionalities and tests.

### Added

- 

### Changed

- Reverted the initial state of `useFormState` to be `isValid: true`

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/hooks/src/useFormState/useFormState.ts
+++ b/packages/hooks/src/useFormState/useFormState.ts
@@ -3,7 +3,7 @@ import { useState } from "react";
 export function useFormState() {
   const [formState, setFormState] = useState({
     isDirty: false,
-    isValid: false,
+    isValid: true,
   });
 
   return [formState, setFormState] as const;


MichaelParadis:
This seems to not be resolving the issue. Form is valid but if a checkbox changes to make a Form valid it is still not being set to valid.

rodrigoeidelvein:
> This seems to not be resolving the issue. Form is inValid and if a checkbox changes to make a Form valid the validity of the form isn't being updated

Yeah, it's not changing the Checkbox behavior tied to the Form. This PR is just reverting the initial state of this hook. Even if we create a PR changing the Checkbox behavior, we would also have to change the feature code where it's being used.

rodrigoeidelvein:
Solving it in a different way here: https://github.com/GetJobber/atlantis/pull/1296

rodrigoeidelvein:
Solving it in a different way here: https://github.com/GetJobber/atlantis/pull/1296

PR #1312, opened 2023-06-06
feat(components-native): Import Select to components-native

MichaelParadis:
### Added
- `Select` now imported into `components-native`

## Testing

On Atlantis root:
```sh
npm pack -w @jobber/components-native
```

On jobber-mobile root (assuming you have them in the same folder): 
```sh
npm install ../atlantis/jobber-components-native-0.21.1.tgz
```
iOS testing:
 - Then you can replace the usages in jobber-mobile for Select.
Android Testing:
Since Android doesn't have a custom select do the following
- Revert 386b181d2e7993731d28a2c42fc4bd7384542d1a
- Update a component to test the multiply native function with
```
  import { multiply } from '@jobber/components-native';

  const [result, setResult] = React.useState<number | undefined>();

  React.useEffect(() => {
    multiply(3, 7).then(setResult);
  }, []);

  return (
    <View style={styles.container}>
      <Text>Result: {result}</Text>
    </View>
  );
```

---

Random photo of Atlantis


rodrigoeidelvein:
@MichaelParadis If Android doesn't require a custom select, we can probably delete the whole Android folder, right?

.gitignore
```diff
@@ -7,3 +7,15 @@ node_modules
...
+# Android/IJ
+#
+.classpath
+.cxx
+.gradle
+.idea
+.project
+.settings
+local.properties
+android.iml
```
rodrigoeidelvein, line 21: We can probably get rid of these too as it is related to the android folders

packages/components-native/package.json
```diff
@@ -2,6 +2,16 @@
...
+  "description": "React Native implementation of Atlantis",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/GetJobber/atlantis.git"
+  },
+  "author": "The Frends",
+  "bugs": {
+    "url": "https://github.com/GetJobber/atlantis/issues"
+  },
+  "homepage": "https://github.com/GetJobber/atlantis#readme",
```
rodrigoeidelvein, line 14: Nice! Thanks for adding it

rodrigoeidelvein (changes requested):
Just a nitpick on the gitignore file. Other than that, it looks good!

packages/components-native/package.json
MichaelParadis: No worries, it was needed for the podspec file

.gitignore
MichaelParadis: Sure thing

MichaelParadis: @rodrigoeidelvein removed in e5e79cf007f4e92caea078782dafd9a29233114a

rodrigoeidelvein (approved):
LGTM! Tested it on Android and iOS and with screen reader on and off

PR #1317, opened 2023-06-07
refactor(eslint): Update lodash eslint rule and imports

taylorvnoj:
## Motivations

Remove `lodash` per-method packages in favour of the main `lodash` package
In future v5 of `lodash`, these per method packages will be removed anyway.

### Added

- eslint rule to force require methods in the import; ex. `import debounce from "lodash/debounce";`

### Changed

- updated all the relevant `lodash` imports
- uninstalled specific `lodash` methods (debounce, chunk, etc)

## Testing

- update an import to only import `from "lodash";` and see the error appear with eslint rule explaining the issue on hover
- you can also run `npm pack -w @jobber/components-native` on the branch and install in another repo to test components

---

Random photo of Atlantis


.eslintrc.js
```diff
@@ -25,7 +25,17 @@ module.exports = {
...
      */
     "import/no-relative-parent-imports": "warn",
-    "no-restricted-imports": "warn",
+    "no-restricted-imports": [
+      "error",
+      {
+        paths: [
+          {
+            name: "lodash",
+            message: "Import [module] from lodash/[module] instead",
```
JeffMarvin, line 34: Minor suggestion: add a period on the end to make it more readable
```suggestion
            message: "Import [module] from lodash/[module] instead.",
```

JeffMarvin:
@taylorvnoj I noticed the following error when I ran `npm ci` on the overall repo (but happens specifically when `packages/components` runs `npm run build`):
```
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
lodash/sortBy (imported by src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts)
lodash/omit (imported by src/DatePicker/DatePickerActivator.tsx, src/InputDate/InputDate.tsx)
lodash/get (imported by src/List/List.tsx)
lodash/groupBy (imported by src/List/List.tsx)
```

Following https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency made it a bit more clear, and I see that we list explicit external modules in `packages/components/rollup.config.js` so maybe add the relevant `lodash/(modules)` in that `external` list? (Also, not sure but I *think* if we should also remove the `lodash` entry from that list due to the linting rule we added.)

taylorvnoj:
Great catch @JeffMarvin thank you.  I removed the `"lodash"` entry because you are right, we won't have that as a standalone import statement anymore.

JeffMarvin (approved):

PR #1318, opened 2023-06-07
feat: Extend Glimmer features

darryltec:
### Added

- Supported `reverseTheme` prop to render the glimmer on a really dark background
- Added `Glimmer.Header`, `Glimmer.Text`, `Glimmer.Button` with default styling and a bit of semantic meaning

### Changed

- Changed the default styling to adapt to the background color

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Glimmer/Glimmer.stories.mdx
+++ b/packages/components/src/Glimmer/Glimmer.stories.mdx
@@ -1,19 +1,12 @@
-import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
+import { ArgsTable, Canvas, Meta, Source, Story } from "@storybook/addon-docs";
 import { Banner } from "@jobber/components/Banner";
 import { Card } from "@jobber/components/Card";
 import { Content } from "@jobber/components/Content";
 import { Divider } from "@jobber/components/Divider";
+import { Tab, Tabs } from "@jobber/components/Tabs";
 import { Glimmer } from ".";
 
-<Meta
-  title="Components/Glimmer"
-  component={Glimmer}
-  subcomponents={{
-    "Glimmer.Header": Glimmer.Header,
-    "Glimmer.Text": Glimmer.Text,
-    "Glimmer.Button": Glimmer.Button,
-  }}
-/>
+<Meta title="Components/Glimmer" component={Glimmer} />
 
 # Glimmer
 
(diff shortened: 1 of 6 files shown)


packages/components/src/Glimmer/Glimmer.stories.mdx
```diff
@@ -33,13 +34,52 @@ import { Glimmer } from "@jobber/components/Glimmer";
...
+## Colors
+
+Glimmers adapts to the background you put it in. If you have a grey background,
+the `Glimmer` will automatically darken itself so it's still noticeable.
+
+<Canvas style={{ background: "var(--color-surface--background)" }}>
+  <Glimmer />
+</Canvas>
+
+If you have a darker or coloured background, you can toggle the `reverseTheme`
```
JeffMarvin, line 57: Why `reverseTheme` and not `darkTheme`? Mostly curious, but I think Dark Mode/Theme is more standard...

packages/components/src/Glimmer/Glimmer.stories.mdx
```diff
@@ -33,13 +34,52 @@ import { Glimmer } from "@jobber/components/Glimmer";
...
+to switch the colors so it's still noticeable.
+
+<Canvas style={{ background: "var(--color-surface--reverse)" }}>
+  <Glimmer reverseTheme={true} />
+</Canvas>
+
+## Semantic blocks
+
+A pre-made blocks can be used to indicate a certain component
```
JeffMarvin, line 66:
```suggestion
A pre-made block can be used to indicate a certain component.
```

packages/components/src/Glimmer/Glimmer.stories.mdx
darryltec: no idea. was making it consistent with the mobile text prop 
image

packages/components/src/Glimmer/Glimmer.tsx
```diff
@@ -46,15 +62,100 @@ export function Glimmer({
...
+interface GlimmerHeaderProps extends Omit<GlimmerProps, "shape" | "size"> {
+  /**
+   * Adjust the size of the `Glimmer.Header`.
+   *
+   * @default 3
+   */
+  readonly level?: 1 | 2 | 3 | 4 | 5;
+}
+
+Glimmer.Header = function GlimmerHeader({
```
JeffMarvin, line 95: I am blind... where are we exporting `Glimmer.Header` and some of the other new components?? I wonder if these new `Glimmer.*` components would be easier to find in their own files.

JeffMarvin (changes requested):
Good stuff, nice update! I had a few questions/suggestions.

packages/components/src/Glimmer/Glimmer.stories.mdx
taylorvnoj: Although I understand why `darkTheme` sounds more familiar, `--color-surface--reverse` connects to `reverseTheme`, pointing to that colour rather than, just black.

packages/components/src/Glimmer/Glimmer.tsx
JeffMarvin: Oh! OHH! I get it now. lol
I do still think there would be value in having GlimmerHeader come from its own file... I think that is possible, right?

packages/components/src/Glimmer/Glimmer.stories.mdx
JeffMarvin: Ok, makes sense. 👍🏻

packages/components/src/Glimmer/Glimmer.tsx
JeffMarvin: It's not a deal breaker... I'll approve this and if you feel like changing it you can just ping me for re-review.

JeffMarvin (approved):

packages/components/src/Glimmer/Glimmer.tsx
darryltec: lemme think about it. i'll leave it for now tho since it won't change how you use it

PR #1321, opened 2023-06-09
fix(components): Fix initial resize of InputText when rows are passed in

rodrigoeidelvein:
## Motivations

`InputText` was not initially resizing if the user passed the `rows` prop with a `min` and `max` values because `rowRange.min` is passed to `FormField` and we just resized when the `onChange` was called.

Now, it will use `useLayoutEffect` to resize it on the first render too and it will handle the resizing if the input value is longer than the minimum row value.

### Added

- 

### Changed

- `InputText` now resizes in the initial render if the consumer passes the `rows` prop

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputText/InputText.tsx
+++ b/packages/components/src/InputText/InputText.tsx
@@ -1,4 +1,10 @@
-import React, { Ref, createRef, forwardRef, useImperativeHandle } from "react";
+import React, {
+  Ref,
+  createRef,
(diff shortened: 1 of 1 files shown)


MichaelParadis (approved):
LGTM!

PR #1328, opened 2023-06-14
feat(components-native): Import contentOverlay to components-native

taylorvnoj:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/ContentOverlay/ContentOverlay.style.ts
+++ b/packages/components-native/src/ContentOverlay/ContentOverlay.style.ts
@@ -1,0 +1,70 @@
+import { StyleSheet } from "react-native";
+import { tokens } from "../utils/design";
+
+const modalBorderRadius = tokens["radius-larger"];
+const titleOffsetFromHandle = tokens["space-small"] + tokens["space-smallest"];
+
+export const styles = StyleSheet.create({
+  handle: {
+    width: tokens["space-largest"],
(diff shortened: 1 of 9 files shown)


rodrigoeidelvein (commented):
Hey Taylor, my environment is broken today, so I still didn't test the code, but here are some considerations for the code itself:

- I see that the `ContentOverlay.tsx` file is handling a lot of things in a single file, maybe we could separate some parts into smaller components and the types into a `types.ts` file.
- Maybe a better way of pulling the booleans of `shouldShowDismiss` into some separate constants to be easier to read. Something like:
```jsx
const isFullScreenOrTopPosition = fullScreen || (!adjustToContentHeight && position === "top");

const shouldShowDismiss = showDismiss || isScreenReaderEnabled || isFullScreenOrTopPosition;
```

taylorvnoj:
@rodrigoeidelvein Thanks for the great feedback! I separated the types out and refactored `shouldShowDismiss` like you suggested.  I tried to separate some of the logic to create components but was having a hard time (for example at times I 
 created dependency cycles).  I was also looking at other examples but wasn't able to cleanly separate logic but I'm all ears if you have other suggestions.
 BTW another good place to do some testings is in the 'on my way' message

rodrigoeidelvein (approved):
LGTM! Tested on iOS and Android :shipit:

PR #1329, opened 2023-06-14
docs(components): Link proposal

devinserink:
### **DO NOT MERGE** 
_this PR is a proposal only_

## Motivations
Create a simple component that will standardize behaviour of external and internal text based hyperlinks on web.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import { cleanup, render } from "@testing-library/react";
+import { Link } from ".";
+
+afterEach(cleanup);
+
+it("renders a Link", () => {
+  const { container } = render(<Link text="Foo" url="https://getjobber.com" />);
+  expect(container).toMatchSnapshot();
```
JeffMarvin, line 9: Maybe these came from the generator... (if so, we should probably update to not use snapshots)...
There's a better way to test these using React Testing Library... I'll try to find some examples in our code and link them here.

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { PropsWithChildren } from "react";
+
+interface LinkProps {
+  readonly external?: boolean;
+  readonly url: string;
+}
+
```
JeffMarvin, line 7: This implementation wraps `<Link>` around `children` but the storybook and test files seem to use a different, self closing tag version of `<Link />` that takes a "title" prop. Would be good to pick one and be consistent. (The tests could just be scrapped for now or treated as WIP with a comment. But the story would be good to update with the proposed approach.)

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,20 @@
...
+export function Link({
+  external = false,
+  url,
+  children,
+}: PropsWithChildren<LinkProps>) {
+  const target = external ? "_blank" : "_self";
+
```
JeffMarvin, line 14: There is quite a bit inside the Atlantis Button component that might be worth considering in your initial implementation. I don't understand all of it, but it might queue up some good discussion of things we want to include out of the box for <Link>.

packages/components/src/Link/Link.test.tsx
devinserink: Appreciate that - and yeah, I just kept the basics from the template, and removed one that felt unapplicable.

JeffMarvin: Page has some tests for when links appear in the intro, they are probably a good pattern to follow: 
https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Page/Page.test.tsx#L144-L191

packages/components/src/Link/Link.tsx
devinserink: Hmm, I looked through button - it does some fancy prop handling, but I don't think that's helpful here unless this component gets a lot more complex.  Interestingly, the button does render `<a>` elements when provided a URL, but it also still renders them to look like buttons, so while semantically questionable, it doesn't seem to be an overlap.

That said, from what I see with the Button, it looks like adding client routing support will be really easy!

JeffMarvin: From button, I'm thinking we might want to include these props just to support the same expressibility of aria attributes:
 ```
    "aria-controls": ariaControls,
    "aria-haspopup": ariaHaspopup,
    "aria-expanded": ariaExpanded,
```    
e.g. docs:    https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded

File: packages/components/src/Link/Link.tsx [hunk 4]
```diff
@@ -0,0 +1,22 @@
...
+  url,
+  children,
+}: PropsWithChildren<LinkProps>) {
+  const linkProps = {
+    ...(external && { target: "_blank" }),
+    href: url,
+    ...(ariaLabel && { "aria-label": ariaLabel }),
+  };
+
+  return <a {...linkProps}>{children}</a>;
```
JeffMarvin, line 21: Since the props are only used in one place, I suggest this as a bit more direct and easier to read:
```suggestion
  return (
    <a
      href={url}
      {...(external && { target: "_blank" })}
      {...(ariaLabel && { "aria-label": ariaLabel })}
    >
      {children}
    </a>
  );
```

JeffMarvin (approved):
Proposal is looking good! I am going to mark this as approved, but suggest the following: 
- I had a suggestion to consider in the implementation about some aria attributes to bring it in parity with Button. (If you agree about adding the other aria attributes mentioned in the comment, those can get included in the *.stories documentation when the code is implemented.)
- I had a suggestion about the code used to set up the props. 

Since this is intended to be a proposal, we don't want to ship the code by approving and merging this. This PR is a useful document for the process though, so I suggest we mark it as approved but include a "DO NOT MERGE" warning at the top of the PR description. (We can then mark this as closed.) 
You could create a new PR for the implementation and then include a link between the two in the descriptions of each of these two PRs to maintain the record of discussion.

packages/components/src/Link/Link.tsx
devinserink: I'd looked through those: they aren't intended for use on <a> (or links in general).  It's a definite anti-pattern / misuse of the component if these did end up being relevant to a use case I think.

JeffMarvin: Hmm, curious where you see that. IMHO, these can be on links that trigger popups or control or expand other content in the DOM. This section of [aria-expanded](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#associated_roles) lists it as relevant for anything with a `link` role, which an <a> automatically has. 
This article has an example that uses one of them on a link. https://www.digitala11y.com/aria-haspopup-properties/

We can continue the discussion/gather other input on the implementation PR though!

PR #1330, opened 2023-06-15
fix(components): InputNumber doesn't run custom validations

gregb-jobber:
## Motivations

If you define a custom validation function on InputNumber like this,
```
    <InputNumber
      placeholder="Custom validation function"
      validations={{
        validate: validationHandler,
      }}
    />
```

The validation function won't run. This was because there's a validation function already defined internally for the component and it doesn't merge the user-defined function with the internal function.

### Added

- 

### Changed

Added a function to merge the validation functions together.
I had to add the function because if I just spread the `validate` property like this,
```
      validations={{
        ...props.validations,
        validate: {
          ...props.validations?.validate,
          getOverLimitMessage,
        },
```
it fails in the cases where there's no `validate` value defined.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputNumber/InputNumber.test.tsx
+++ b/packages/components/src/InputNumber/InputNumber.test.tsx
@@ -199,6 +199,70 @@ test("allows custom validation", async () => {
   });
 });
 
+test("it should call the custom validate function if provided", async () => {
+  const validationHandler = jest.fn();
+  const expectedValidationValue = Math.floor(
+    Math.random() * Number.MAX_SAFE_INTEGER,
+  );
+  const { getByLabelText } = render(
+    <InputNumber
+      value={expectedValidationValue}
+      placeholder="Custom validation function"
+      validations={{
+        validate: validationHandler,
+      }}
+    />,
+  );
+
+  const input = getByLabelText("Custom validation function");
+
+  input.focus();
+  input.blur();
+
+  await waitFor(() => {
+    expect(validationHandler).toHaveBeenCalled();
+    // expect(validationHandler).toHaveBeenCalledWith(expectedValidationValue);
+    // Expected: 12
+    // Received: 12, {"generatedName--123e4567-e89b-12d3-a456-426655440063": 12}
+  });
+});
+
+test("it should use the custom validate object if provided", async () => {
+  const validationHandler = jest.fn();
+  const validationHandler2 = jest.fn();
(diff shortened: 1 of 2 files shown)


packages/components/src/InputNumber/InputNumber.test.tsx
```diff
@@ -199,6 +199,70 @@ test("allows custom validation", async () => {
...
+  const input = getByLabelText("Custom validation function");
+
+  input.focus();
+  input.blur();
+
+  await waitFor(() => {
+    expect(validationHandler).toHaveBeenCalled();
+    // expect(validationHandler).toHaveBeenCalledWith(expectedValidationValue);
+    // Expected: 12
+    // Received: 12, {"generatedName--123e4567-e89b-12d3-a456-426655440063": 12}
```
gregb-jobber, line 226: I'd like to do `toHaveBeenCalledWith` to be more precise about what's going on, but I haven't been able to figure out how to get Jest to match a string within `toHaveBeenCalledWith`. I'm asking the testing guild, but it won't change the code, so this can be reviewed.

packages/components/src/InputNumber/InputNumber.tsx
```diff
@@ -57,11 +58,29 @@ function InputNumberInternal(
...
+  function customValidators(
+    validators?: RegisterOptions["validate"],
+  ): RegisterOptions["validate"] {
+    if (validators == null) {
+      return getOverLimitMessage;
+    } else if (typeof validators === "function") {
+      return {
+        customValidation: validators,
+        getOverLimitMessage,
+      };
```
darryltec, line 75: This seems to not fire `getOverLimitMessage` anymore. Same goes for the one below

Also, good to have a test case for when you have a custom validator the min/max should still work.

darryltec (changes requested):
Found a bit of an issue

packages/components/src/InputNumber/InputNumber.tsx
darryltec: Turns out, if one of your `validate` functions returns a `false` it'll prevent the other validations from happening.

I think the only change here is the test to make sure the internal validations still runs and document the validation quirkiness

darryltec (approved):
:shipit:

PR #1334, opened 2023-06-16
feat(components): Link implementation

devinserink:
## Motivations

Implementation for #1329 

> Create a simple component that will standardize behaviour of external and internal text based hyperlinks on web.

### Added

New link component that renders <a> elements for internal and external hyperlinks.

### Changed

- 

### Deprecated

Discussion ongoing around the overlapping functionality in markdown, and the relevance of that component going forward, but this is out of the scope of this PR.

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Go to component page in Storybook and confirm component loads and renders.

---

Random photo of Atlantis


packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,24 @@
...
+  readonly external?: boolean;
+  readonly url: string;
+}
+
+export function Link({
+  ariaLabel,
+  external = false,
+  url,
+  children,
+}: PropsWithChildren<LinkProps>) {
```
JeffMarvin, line 14: It occurred to me that we might want an `onClick` event as well. Thoughts?

devinserink: Considered that.. it opens the door to a lot of anti-patterns.  The legit use case I can see might be needing to add dynamic behavior around a "you'll lose your work if you leave" type dialog.  But: that's kind of an anti-pattern to (vs e.g. using window.beforeunload where you can do such checks in one place instead of trying to intercept anything that might take you off the page).

Are there other (commendable) use cases we can see right now?  Otherwise, I'm inclined to opine in favour of YAGNI.

devinserink: Here's what autopilot had to say about it when I started typing this sentence 😂 

> Adding a prop for OnClick was considered, but rejected as it would be too easy to abuse. If you need to trigger an action, use a `<Button>` instead.

devinserink:
Reopening Aria discussion from proposal:

I added the aria-expanded attribute: that makes sense - I misread that one as being an element that controlled expansion, not being subject to it.

Still not sure about popup.  That kind of action feels like the domain of buttons or other controls... I'm not sure if we'd want links doing that.  The exception I can think of is the sort of save on exit type dialog: but is that something that should be added to ALL controls that might trigger what is ultimately not a function of the control so much as the page?  I couldn't find a clear answer.

packages/components/src/Link/Link.tsx
JeffMarvin: Is this kind of the guideline that you are thinking? https://polaris.shopify.com/components/navigation/link#submitting-data

i.e. links should navigate, and from a UI perspective other things (like submitting data, or triggering other dynamic behaviour) should look like buttons (or, at least not look like links?)
I think this makes sense, curious what others think:  @fagnanm @darryltec

packages/components/src/Link/Link.stories.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+```ts
+import { Link } from "@jobber/components/Link";
+```
+
+<Canvas>
+  <Story
+    name="Link"
+    args={{
+      text: "Link Component Notes",
```
JeffMarvin, line 16: This prop doesn't match the component anymore, I think it can be removed.

packages/components/src/Link/Link.tsx
devinserink: Yeah - essentially.  "buttons" are the place where I would want to see flexibility (to whatever extent that Button component supports those use cases is a separate concern)..  My thinking is that links are a fairly special/consistent thing users expect to behave a certain way.

fagnanm: Yes that makes sense. TBH it's never come up on the design team, may have to push on it a little more if we had a button style that also looked like a link in the future.

JeffMarvin: Oh nice! I was going to check active usages... The real question is do we want these to be `<Link>`'s under the hood, or should they be `<Button>`'s styled to look like links?

darryltec: we have a shiny new eslint rule for a11y where it will force you to use a button because screen readers and keyboard

packages/components/src/Link/Link.stories.mdx
devinserink: Ah yes, this vestigial prop has been surgically removed.  Good catch.

JeffMarvin: 🔪

packages/components/src/Link/Link.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+  </Story>
+</Canvas>
+
+## Props
+
+<ArgsTable of={Link} story="Link" />
+
+## Usage guidelines
+
+The `<Link>` will render an `<a>` tag with the appropriate `href`, `rel` and
```
JeffMarvin, line 30: I don't think this sentence is accurate anymore.

packages/components/src/Link/Link.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+Try to ensure the text of the `<Link>` is descriptive of the destination. Avoid
+using "click here" or "learn more" as the text of the link, and instead use the
+title of the destination page or a short description of the content.
+
+If it's not possible to make the visible text sufficiently descriptive without
+additional context, set the `ariaLabel` prop to a descriptive string for
+screenreaders. Do _not_ set this prop if it will be redundant to the link text.
+
+## Notes
```
JeffMarvin, line 59: I think this Notes section could be excluded, as it is more for maintainers than users of the library. Another idea would be to move it to a file in this directory (e.g. `Link.design_notes.txt` or something) to preserve these thoughts for anyone else coming to modify or contribute to the component later. (I will float this idea past the team and see what they think!)

packages/components/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import { cleanup, render } from "@testing-library/react";
+import { Link } from ".";
+
+afterEach(cleanup);
```
JeffMarvin, line 5: This isn't needed - Jest does this automatically. (See https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-cleanup)

packages/components/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+
```
JeffMarvin, line 6: We would typically move the content being `render`ed and `getBy`ed into a variable, for readability and to be more robust (as it uses the same variable in the comparison). E.g. 
```
const testURL = "https://getjobber.com";
const testLinkText = "This is a link";

describe("when a Link is rendered", () => {
  it("opens the link in the same tab if external is left at default value of false", () => {
    const { getByText } = render(<Link url={testURL}>{testLinkText}</Link>);
    const link = getByText(testLinkText);
    expect(link.getAttribute("target")).toBeNull();
  });
...
// other tests
});
```

packages/components/src/Link/Link.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+  it("opens in a new tab if external is set to true", () => {
+    const { getByText } = render(
+      <Link url="https://getjobber.com" external={true}>
+        This is a link
+      </Link>,
+    );
+    const link = getByText("This is a link");
+    expect(link.getAttribute("target")).toBe("_blank");
+  });
+});
```
JeffMarvin, line 24: Consider a test that ensures the URL is coming through correctly... e.g. 
```
it("links to the correct url provided", () => {
    const { getByText } = render(<Link url={testURL}>{testLinkText}</Link>);
    const link = getByText(testLinkText);
    expect(link.getAttribute("href")).toBe(testURL);  
});
```

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { PropsWithChildren } from "react";
+
+interface LinkProps {
+  readonly ariaLabel?: string;
+  readonly ariaExpanded?: boolean;
+  readonly external?: boolean;
+  readonly url: string;
```
JeffMarvin, line 7: Small nitpick - consider moving this line for `url` to the top, since it is the only required property.

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export function Link({
+  ariaLabel,
+  ariaExpanded,
+  external = false,
+  url,
+  children,
```
JeffMarvin, line 15: Same small style suggestion: consider moving `url` and `children` up to the top of this list.

JeffMarvin (changes requested):
Great stuff here! I had a few suggestions, feel free to ping me if it is useful to discuss any of them in a chat.

JeffMarvin (approved):
Atlantis - now with Links!
Nice work @devinserink! :D

PR #1335, opened 2023-06-19
feat(components-native): Import Menu to components-native

rodrigoeidelvein:
### Added

- `Menu` component in components-native

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components-native/src/AtlantisContext/AtlantisContext.tsx
```diff
@@ -52,6 +57,7 @@ export const defaultValues: AtlantisContextProps = {
   },
   floatSeparators: { group: ",", decimal: "." },
   currencySymbol: DEFAULT_CURRENCY_SYMBOL,
+  headerHeight: 50,
```
taylorvnoj, line 60: Where does 50 come from? 
Is this something to do with `@react-navigation/stack`?

taylorvnoj (commented):
Looks good - just a small suggestion and a question

packages/components-native/src/AtlantisContext/AtlantisContext.tsx
rodrigoeidelvein: No, it's just a random number, because the `headerHeight` differs from device to device and platform.

rodrigoeidelvein: Whoops, yes, it's the value of `useHeaderHeight` from `@react-navigation/stack`, but 50 is just a default number here

rodrigoeidelvein: I changed it to 0 because if the user don't pass this value to the context, we understand that actually there's no header to worry about. Does it make sense?

taylorvnoj: Yes that makes more sense to me!

taylorvnoj (approved):
LGTM! 🚀

PR #1341, opened 2023-06-26
feat(components-native): Migrate ThumbnailList

MichaelParadis:
### Added

- components-native: migrate `ThumbnailList`

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

On Jobber Mobile using  `test-thumbnail-list` verify thumbnail list works as expected.

---

Random photo of Atlantis


packages/components-native/src/ThumbnailList/ThumbnailList.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { styles } from "./ThumbnailList.style";
+import { FormatFile } from "../FormatFile";
+import { CreateThumbnail, File } from "../FormatFile/types";
+
+export enum RowCount {
+  TwoRows = 2,
+  ThreeRows = 3,
+}
+
+export interface ThumbnailListProps {
```
taylorvnoj, line 13: Should we put the above `enum` and this `interface` in a separate `types.ts` file?  It's not like this file is too big but just wondering if you think it's worth it based on patterns of other components.

taylorvnoj (commented):
Quick contemplation!

packages/components-native/src/ThumbnailList/ThumbnailList.tsx
MichaelParadis: Makes sense to me. I can update it

taylorvnoj (approved):
Thanks for making that change!

PR #1361, opened 2023-07-05
refactor(components-native): Remove react-native-localize dependency

rodrigoeidelvein:
## Motivations

We were just using one function from `react-native-localize`, and we actually can replace that function call with a native one from `Intl`.

### Added

- 

### Changed

- The default `timezone` value from `AtlantisContext` is now using the timezone from the native `Intl`

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


darryltec (approved):
:shipit:

PR #1362, opened 2023-07-06
docs(components-native): Add FormatFile mobile stories

taylorvnoj:
## Motivations

Documentation between web and mobile now need to be unified. This adds Mobile stories for `FormatFile`.

### Added
 
- `FormatFile` stories for mobile are now `Image`, `PDF` and `Video` to match mobile documentation

---

Random photo of Atlantis


taylorvnoj:
Update: I created a ticket to look into the constant loading state.  There is a fix I noted in that JIRA ticket but it only fixes the problem on Storybook... bringing that into the app then breaks the loading state there.

PR #1381, opened 2023-07-14
docs: add IconButton docs

darryltec:
## Motivations

What's that? Oh yeah, docs migration.

### Added

- IconButton docs baby!

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


rodrigoeidelvein (commented):
It looks good, the only thing that I found is that I get an error when I try to set a badge for the IconButton.

Currently, on Atlantis Mobile, we can't change the props in the props table, but now we have the option and it's not working... maybe it's an easy fix?

darryltec:
it is not. it needs a component

rodrigoeidelvein:
@darryltec maybe we can think of making this prop only a string instead of a component as a future improvement for this

rodrigoeidelvein (approved):
LGTM!

PR #1388, opened 2023-07-17
docs: add Google Analytics tracking

darryltec:
## Motivations

Because we want to know how much traffic we getting and where

### Added

- installed `react-ga4` package instead of @storybook/google-analytics since it doesn't support GA4.
- Tracked page visits through manager js
- Also included errors since that's what the @storybook/google-analytics does 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Install this [chrome extension](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en)
- Ensure your browser extensions that blocks trackers are disabled
- Once done, click the google analytics extension to turn it on. It should have an ON badge
- It should start console logging the infos when viewing pages

Or just trust that it works

image
image

---

Random photo of Atlantis

### DIFF
--- a/.storybook/manager.js
+++ b/.storybook/manager.js
@@ -1,9 +1,11 @@
 import React from "react";
 import { addons, types } from "@storybook/addons";
+import { STORY_CHANGED, STORY_ERRORED, STORY_MISSING } from '@storybook/core-events';
 import { SidebarLabel } from "./components/SidebarLabel";
 import theme from "./theme";
 import favicon from "./assets/favicon.svg";
 import { Playground } from "./components/Playground";
+import ReactGA from "react-ga4";
 
 import "./assets/css/manager.css";
 import "@jobber/design/foundation.css";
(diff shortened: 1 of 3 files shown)


.storybook/manager.js
```diff
@@ -39,3 +41,27 @@ addons.register("code/tab", () => {
     render: ({ key, active }) => <>{active && <Playground key={key} />}</>,
   });
 });
+
+addons.register('google-analytics', (api) => {
+  ReactGA.initialize("G-V1N3TQVQB5");
```
rodrigoeidelvein, line 46: Should we put it in a environment variable?

rodrigoeidelvein (commented):
Maybe we need to do a environment check and only send the tracking events in the production environment

.storybook/manager.js
darryltec: it's smart not to. GA tracking would only really track when it matches the URL of atlantis.getjobber.com

rodrigoeidelvein: Nice! 

And about this GA code, should we move it to a environment variable or is it okay to keep it public?

darryltec: okay to be public. it's pretty much gonna be public no matter what since it'll show up in the DOM

rodrigoeidelvein (approved):
Worked well using the extension, the events are showing up on my console!

LGTM!

PR #1390, opened 2023-07-18
docs(components): Update InputPassword docs

taylorvnoj:
## Motivations

InputPassword docs are updated to include web and mobile stories

### Changed

 
The mobile stories required an `export` on `InputPasswordProps` to make it available to Storybook.  If you notice that the `export default` code in `Mobile.stories.tsx` is erroring out, you just need to do a `npm run build -w @jobber/components-native` and the error will be gone

---

Random photo of Atlantis


docs/components/InputPassword/Mobile.stories.tsx
```diff
@@ -7,6 +7,8 @@ export default {
   component: InputPassword,
   parameters: {
     viewMode: "story",
+    previewTabs: { code: { hidden: false } },
```
JeffMarvin, line 10: Is our default for this setting `true` or do we have no default, i.e. always need to specify it?

taylorvnoj: @JeffMarvin  Yup! In `../.storybook/preview.tsx` it's marked `true` by default.  

My understanding is when we implemented the code tab, we wanted to hide it by default because the mobile stories didn't have the code tab fixed yet.  If we took this away now, or changed the default to false, the code tab would appear everywhere which I don't think we are ready for yet.  
It will show up on the Docs page and then show the code for the first example.
And because not everything is finalized yet, shipping a change now might not be best/add confusion to the user.  (I know you aren't asking that yet but wanted to share some considerations)

JeffMarvin: Sounds good, thanks for the additional context! :)

JeffMarvin (approved):
LGTM! Nice work!

PR #1397, opened 2023-07-19
chore(components): Deprecate Toast Actions 

dlrussel:
## Motivations

Toast actions aren't being used and therefore should be removed.

### Deprecated
- `action` and `actionLabel` props on `Toast`

## Testing

1.  Install the pre-release 
2. Use Toast component and pass in the `action` and `actionLabel` props 
3. Verify that you see the deprecation warning in the your editor (note that it will only show in a tooltip and you won't actually be able to see the strikethrough due to [this typescript issue](https://github.com/microsoft/TypeScript/issues/39374))
4. Make sure you can see the deprecation comment in storybook as shown below 
image

---

Random photo of Atlantis


taylorvnoj:
I am pondering on one thing.  
In other @deprecated props we have given an alternative to use, which I know is not the case here.  I am however, wondering if we should give some context or just state that it's a misuse to have an action in a Toast so users understand why it's deprecated.  I also think the current message could be confusing if not changed after the next major release - leaving the user to wonder if it was removed or not.  Let me know what you think 😃

dlrussel:
@taylorvnoj Ya I can definitely add some more context around why it's being depreciated if you think that's something that would be valuable. I think considering that no one is currently using those props we may be good to just depreciate them and then remove them entirely. Is there a reason why we wouldn't remove them during the next major version?

taylorvnoj:
@dlrussel

> considering that no one is currently using those props we may be good to just depreciate them and then remove them entirely.

Very true. I was thinking they'd still be listed for some reason, probably because other deprecated props are still in the code, but for this case where it's not even being used...this makes a lot more sense.  We'll just have to remember to remove them when the time comes

Thanks Diane!

taylorvnoj (approved):
LGTM!

PR #1399, opened 2023-07-19
docs: use babel-ts instead of babel as parser

darryltec:
## Motivations

Noticed that Autocomplete Code is not loading and it was due to the prettier parser being set to babel instead of babel-ts. it didn't know what to do with `useState<State>()`

### Added

- 

### Changed

- change prettier parser from babel to babel-ts

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Autocomplete Code tab should load
- Others should load as well

---

Random photo of Atlantis


taylorvnoj (approved):
good catch!

PR #1423, opened 2023-07-25
docs(components): Update Tabs docs

rodrigoeidelvein:
## Motivations

Updating Tabs docs, it's web only.

### Added

- 

### Changed

- Updated Tabs documentation

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


taylorvnoj:
For the empty Controls section, I was thinking we could do one of two things:

1. If we change `ComponentStory<typeof` to `Tab>` instead of tabs, we can put the `args` in the export for *one* tab. Looks slightly strange in VSCode however in my opinion, it makes Storybook look a bit better and the code tab still works.  The first label does have `{ }` around it in the first tab though 🤔 
Screenshot 2023-07-26 at 7 52 52 AM

2. We utilize `StoryDetails` to put a disclaimer like in mobile stories, which explains that the component Tabs is an args-less component - it's the Tabs within that take in args instead

I guess a third option would be to explain this in a `Content guidelines` section of the docs page but I'm leaning toward something being done in the Canvas example.

rodrigoeidelvein:
Hey @taylorvnoj! I liked the first option, having a control is better than nothing. And having the `{ }` around the first label shouldn't be an issue. Thanks for the suggestion

taylorvnoj (approved):
Love it! 🚢

PR #1428, opened 2023-07-26
docs(components): Update Checkbox docs

rodrigoeidelvein:
## Motivations

Updating Docs for Checkbox, it's mobile and web.

### Added

- 

### Changed

- Update docs for Checkbox

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


docs/components/Checkbox/Mobile.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+    />
+  );
+};
+
+export const Controlled = ControlledTemplate.bind({});
+Controlled.args = {
+  label: "Check me out",
+};
+
+export const Indeterminate = BasicTemplate.bind({});
```
dlrussel, line 42: I think it may be worth adding more information/context around this example in the docs section explaining when it would be used and linking to this example. Mobile storybook has the context for this so it's probably worth porting over.

docs/components/Checkbox/Web.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+  disabled: true,
+  defaultChecked: true,
+};
+
+const ChildrenAsLabelTemplate: ComponentStory<typeof Checkbox> = args => {
+  const [checked, setChecked] = useState(false);
+  return (
+    <Checkbox {...args} checked={checked} onChange={setChecked}>
+      <Text>
+        I agree to the
```
dlrussel, line 81: This is missing a space between the text and the link 
image

docs/components/Checkbox/Web.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+        <button
```
dlrussel, line 82: I don't think this is the right use case for a button. We should probably be using a link here.

dlrussel (changes requested):

docs/components/Checkbox/Mobile.stories.tsx
rodrigoeidelvein: Good point! I added this section that we currently have on mobile to the updated docs, it applies for both Web and Mobile components. I also added a section for mobile's CheckboxGroup

docs/components/Checkbox/Web.stories.tsx
rodrigoeidelvein: Nice catch!

docs/components/Checkbox/Web.stories.tsx
rodrigoeidelvein: I agree! I ended up just copying and pasting from the previous documentation and forgot that we now have a Link component. Thanks, I changed it there

rodrigoeidelvein:
Thanks for the detailed review @dlrussel ! I addressed them, let me know what you think

dlrussel (approved):
🚢

PR #1432, opened 2023-07-26
docs(components): Update docs for Chips

rodrigoeidelvein:
## Motivations

Updating the docs for Chips. It's a tricky one for web and mobile. I added `Tabs` to separate things that differentiate between web and mobile, let me know if you have a better idea for handling it.

### Added

- 

### Changed

- Update docs for Chips

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


docs/components/Chips/Chips.stories.mdx
```diff
@@ -42,39 +31,6 @@ Unlike
 selected single-select Chip can be de-selected by the user, leaving all
```
dlrussel, line 31: The above link for radio is broken and should be `[Radio](../?path=/docs/components-forms-and-inputs-radiogroup-docs--page)`

docs/components/Chips/Mobile.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+export const InactiveBackgroundColor = BasicTemplate.bind({});
+InactiveBackgroundColor.args = {
+  label: "Inactive Chip",
+  accent: "client",
+  icon: "clients",
+  onPress: () => alert("Clients"),
+  accessibilityLabel: "Select clients",
+  isDismissible: true,
```
dlrussel, line 99: Should an inactive chip be dismissible? If it's inactive I would assume that it should be able to be interacted with. It might be helpful if there was a use case for inactive in the docs.

dlrussel (changes requested):

docs/components/Chips/Mobile.stories.tsx
rodrigoeidelvein: Good question, I was testing it here and the prop `inactiveBackgroundColor` just changes the color, the user still can click and interact with the Chip 🤔 

I will remove the `isDismissible` to avoid confusion, it was just a leftover from the previous example

docs/components/Chips/Chips.stories.mdx
rodrigoeidelvein: Nice catch!

docs/components/Chips/Mobile.stories.tsx
dlrussel: hm if you can still interact with it then should it be called inactive? I'm confused as to how this is meant to work.

rodrigoeidelvein: I'm confused too. I can raise a ticket so we can review this behavior and fix it

rodrigoeidelvein: Looks like it's only actually disabled when a `onPress` callback is not passed: https://github.com/GetJobber/atlantis/blob/f73435c6443b80638b1b7558dfba9a4b8f79bdd1/packages/components-native/src/Chip/Chip.tsx#L111

rodrigoeidelvein: Oh, ok. I think I got it. The Chip being inactive doesn't mean it's disabled. Because you can have multiple chips and select some of them to be "active", or "selected". You can still dismiss the chip if it's inactive, because it's just not select at the moment.

dlrussel: oh ok sounds good.

dlrussel (approved):

PR #1435, opened 2023-07-27
docs(components): fix InputFile code examples

JeffMarvin:
## Motivations

Some code tab examples were broken for `InputFile` and the "Variations and Sizes" example didn't behave very nicely, it just hid the controls. 

### Changed
- Improved the "Variations and Sizes" example allowing a default to handle the args.  Also stopped hiding the controls since the user can just turn them back on. 

### Fixed
- Code tab is working for all InputFile examples now. 

## Testing
- [x] Confirm the Code sandbox works properly for all examples. 

---

Random photo of Atlantis

### DIFF
--- a/docs/components/InputFile/Web.stories.tsx
+++ b/docs/components/InputFile/Web.stories.tsx
@@ -14,18 +14,20 @@ export default {
   component: InputFile,
   parameters: {
     viewMode: "story",
-    options: { showPanel: true },
     previewTabs: {
       code: {
         hidden: false,
+        extraImports: {
+          "@jobber/components/InputFile": [
+            "FileUpload",
+            "InputFile",
+            "updateFiles",
+          ],
+        },
       },
     },
   },
 } as ComponentMeta<typeof InputFile>;
-
-function fetchUploadParams() {
-  return Promise.resolve({ url: "https://httpbin.org/post" });
-}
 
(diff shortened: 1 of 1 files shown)


taylorvnoj:
Looks really great Jeff! 

The only thing I'm thinking about is the way `<InputFile />` looks on line 13 of the sandbox.  On first glance, it looks odd to have a component there without any `args`.  On the other hand, the whole point is to display the default component.  

Putting a `buttonLabel="File Upload"` arg in there would fill out the sandbox - but then that would be redundant since it's not needed.

I'm open to leaving it as is.  I'm curious what you think about it 😄 

Screenshot 2023-07-27 at 7 27 38 AM

JeffMarvin:
> Looks really great Jeff!
> 
> The only thing I'm thinking about is the way `<InputFile />` looks on line 13 of the sandbox. On first glance, it looks odd to have a component there without any `args`. On the other hand, the whole point is to display the default component.
> 
> Putting a `buttonLabel="File Upload"` arg in there would fill out the sandbox - but then that would be redundant since it's not needed.
> 
> I'm open to leaving it as is. I'm curious what you think about it 😄

I think my gut feel is to leave it as is, because adding anything else will mean it isn't the default.

taylorvnoj (approved):
LGTM!

PR #1453, opened 2023-07-31
docs(components): Update Heading docs

taylorvnoj:
## Motivations

Heading web and mobile need to be combined into one docs section

### Added

The `levels` prop within Heading for web vs. mobile is vastly different - I decided to lay that out in a table with tabs for Web and Mobile in an effort to condense the information.  Then I put all `level`s within each respective story with the first Heading being the `{...args}`.   

## Testing

- [ ] make sure code tab works for each story

---

Random photo of Atlantis


docs/components/Heading/Heading.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+There are a few caveats around copying text on Android and iOS that you can read
+under the
+[Typography](../?path=/docs/components-text-and-typography-typography-docs--page#platform-considerations)
+documentation.
+
+## Accessibility
+
+The Heading component has the accessibilityRole of 'Header' and will be
+announced to assistive technology as "Header: {text content}".
```
darryltec, line 68: Do we need to explain this?

Also the `{text content}` string would break on storybook 7 so maybe use [] or just ay text content really.

darryltec (commented):
lil tiny thing

docs/components/Heading/Heading.stories.mdx
taylorvnoj: I took it from the mobile docs. I could see that most people might assume this already, but I'm not sure it's taking anything away from the documentation by being there. It's another way of saying, use Heading instead of big bold text to give context for assistive technology users.

darryltec: IMO, it's one of those implied things where if we start explaining them it will be a thing we have to explain everywhere. Or in other words, assume everything is working unless otherwise stated.

taylorvnoj: Fair enough. I removed that section.

darryltec (approved):
We good!

PR #1465, opened 2023-08-01
feat(components): InputText Props for Autocomplete

jessedoyle:
## Motivations

The `Autocomplete` component utilizes `InputText` behind the scenes to provide the primary input element that is rendered.

We've encountered a scenario where the following props need to be passed down to the underlying InputText element:

* `suffix` - To style the input with an icon suffix.
* `prefix` - To enable a clear action for the input.
* `inputRef` - A reference to the input element in the DOM so that focus can be activated programatically.

## Changes

This PR exposes the props listed above on `Autocomplete` and forwards them to the `InputText` child. As the current `name` prop is also forwarded to the input, I opted to remove the explicit reference in favour of forwarding the prop via the spread operator.

### Changed

- Add the `suffix`, `prefix` and `inputRef` props to the `Autocomplete` component.

## Testing

The additional props are editable in the [StoryBook documentation](https://feat-autocomplete-props.atlantis.pages.dev/?path=/docs/components-forms-and-inputs-autocomplete-web--basic). Happy to add additional stories to the docs if necessary!

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Autocomplete/Autocomplete.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.tsx
@@ -12,7 +12,15 @@ type OptionCollection = XOR<Option[], GroupOption[]>;
 interface AutocompleteProps
   extends Pick<
     FormFieldProps,
-    "size" | "onBlur" | "onFocus" | "invalid" | "name" | "validations"
+    | "invalid"
+    | "inputRef"
+    | "name"
+    | "onBlur"
+    | "onFocus"
+    | "prefix"
+    | "size"
+    | "suffix"
+    | "validations"
   > {
   /**
    * Initial options to show when user first focuses the Autocomplete
@@ -59,11 +67,8 @@ interface AutocompleteProps
   readonly placeholder: string;
 }
 
-/**
- * Max statements disabled here to make room for the
- * debounce functions.
- */
-// eslint-disable-next-line max-statements
+// Max statements increased to make room for the debounce functions
+/* eslint max-statements: ["error", 14] */
 export function Autocomplete({
   initialOptions = [],
   value,
@@ -76,8 +81,8 @@ export function Autocomplete({
   placeholder,
   onBlur,
   onFocus,
-  name,
   validations,
+  ...inputProps
 }: AutocompleteProps) {
   const [options, setOptions] = useState(initialOptions);
   const [menuVisible, setMenuVisible] = useState(false);
@@ -106,8 +111,8 @@ export function Autocomplete({
         placeholder={placeholder}
         onFocus={handleInputFocus}
         onBlur={handleInputBlur}
-        name={name}
         validations={validations}
+        {...inputProps}
       />
       {menuVisible && (
         <Menu
--- a/packages/components/src/Autocomplete/Autocomplete.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.tsx
@@ -12,8 +12,8 @@ type OptionCollection = XOR<Option[], GroupOption[]>;
 interface AutocompleteProps
   extends Pick<
     FormFieldProps,
-    | "invalid"
     | "inputRef"
+    | "invalid"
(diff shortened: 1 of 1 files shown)


packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -106,8 +110,8 @@ export function Autocomplete({
         placeholder={placeholder}
         onFocus={handleInputFocus}
         onBlur={handleInputBlur}
-        name={name}
```
jessedoyle, line 109: The `name` prop is included in the `inputProps` spread object, so we no longer need to explicitly pass it to the child.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -59,10 +67,6 @@ interface AutocompleteProps
   readonly placeholder: string;
 }
 
-/**
- * Max statements disabled here to make room for the
```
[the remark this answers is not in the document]
jessedoyle: Are we okay removing this comment? It feels like we should probably add context as to why we're disabling lint checks in our commit messages.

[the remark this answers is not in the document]
darryltec: I think the current standard right now (Jobber wise) is to increase the limit on this 1 function instead of disabling it and then adding a quick summary as to why.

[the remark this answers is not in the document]
jessedoyle: Sure, I'll add the comment back in! Going to amend the commit.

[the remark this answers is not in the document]
jessedoyle: Pushed an amendment [here](https://github.com/GetJobber/atlantis/compare/8cfc2ba5a39e01b1a81cc142ccbe38b5271332c3..c72d41bb76d278b663577a40b6f5714589eb47be)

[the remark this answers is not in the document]
darryltec: muchas gracias mi amigo!

packages/components/src/Autocomplete/Autocomplete.tsx
jessedoyle: @darryltec - It looks like the `invalid` prop is in a similar scenario. Any issues if I give it the same treatment here?

darryltec: nope. no issue at all. clean that thang up!

jessedoyle: Done in e25b098, thanks!

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -100,14 +104,13 @@ export function Autocomplete({
       <InputText
         autocomplete={false}
         size={size}
-        invalid={invalid}
```
darryltec, line 103: This is weird. maybe this is doing something? doesn't seem to highlight red anymore

image

compared to master

image

darryltec (changes requested):
One tiny thing then we good!

packages/components/src/Autocomplete/Autocomplete.tsx
jessedoyle: Hmm interesting - I'll check it out!

jessedoyle: Not seeing the same behaviour on the [storybook preview link](https://feat-autocomplete-props.atlantis.pages.dev/): 

storybook

Validations and error highlighting appear to be functioning as expected?

darryltec: Oh dang, you right!. probs just my local being a silly goose!

darryltec (approved):
:shipit:

PR #1471, opened 2023-08-03
feat(components): Permit using children on RadioOption with label 

TJKruitbosch:
## Motivations

For upcoming work, we would like to have an RadioGroup that has complex RadioOptions that include a label, description, as well as secondary form elements associated with that option as children. For example, consider a group with _Option A_, _Option B_, and _Option C_ as options, but a user selecting _Option A_ must also select from a series of Chips under it, and a user selecting _Option B_ must fill out an InputNumber field.

Screen Shot 2023-08-03 at 3 50 18 PM

[Figma Prototype Link](https://www.figma.com/proto/vOZ51PFeJdaJTkx2pzfGRU/%5BOLB%5D-Maximum-Drive-Time?page-id=10%3A32789&type=design&node-id=182-80076&viewport=-906%2C-759%2C0.19&t=M4jQkAuf2DgjKQF1-1&scaling=min-zoom&starting-point-node-id=182%3A79936) for internal reviewers

### Changed

- RadioOption now supports providing label, description, and children on the same option.

## Testing

Modify the storybook story/example ["Radio Option with Children"](https://atlantis.getjobber.com/?path=/story/components-forms-and-inputs-radiogroup-web--radio-option-with-children) to also output the User's name as a label. 

```diff
           {users.map(user => (
-            <RadioOption value={user.name} key={user.name}>
+            <RadioOption value={user.name} key={user.name} label={user.name}>
               <Avatar initials={user.initials} name={user.name} />
             </RadioOption>
           ))}
```
Screen Shot 2023-08-03 at 3 54 58 PM

---

Random photo of Atlantis


packages/components/src/RadioGroup/RadioOption.tsx
darryltec, line 1: Haven't really dove into this file but, I wonder if we can simplify this now and not have restrictions anymore. I think we're more mature with react and a11y that restrictions like this might be more confusing than helpful. The moment you reach for only the `children` prop, you are responsible for what goes in there and are saying "I don't want the default interaction"

What do you think?

TJKruitbosch: Taking another look at this with fresh post-weekend eyes...I think the "you must provide _at least one of_ `label` and `children`" restriction still makes sense here. The component itself doesn't make a lot of sense visually without _something_ in the label area. That said, I think the types are a little over-complex for expressing that, and I can give simplifying it a shot.

darryltec: Coolio! Thanks homie!

TJKruitbosch: I gave refactoring the prop types here a shot. There's now just one type for the "internal" options, just one human-readable type for the external options, and one intermediate type solely for enforcing the "you must provide at least one of label or children" rule. If it turns out in the future that I'm totally wrong here (it happens, no big deal 🤷‍♂️) and that RadioOptions without labels or children should be permitted, it should be straightforward to just remove that intermediate type and use `BaseRadioOptions` directly. What do you think?

darryltec (changes requested):
Haven't tested this but a few little bits I saw. Thanks for simplifying the types on this!

packages/components/src/RadioGroup/RadioOption.tsx
```diff
@@ -83,6 +86,11 @@ export function InternalRadioOption({
           </Text>
         </div>
       )}
+      {children && label && (
```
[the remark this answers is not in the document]
TJKruitbosch: No strong feelings against that, I just live in constant fear of the max-lines lint rule ruining my day. 

How about `shouldRenderIndependentChildren`?

packages/components/src/RadioGroup/RadioOption.tsx
```diff
@@ -4,50 +4,53 @@ import { v1 as uuidv1 } from "uuid";
 import styles from "./RadioGroup.css";
 import { Text } from "../Text";
 
-interface BaseRadioOptions {
+type BaseRadioOptions = PropsWithChildren<{
```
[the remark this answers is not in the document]
TJKruitbosch: Only thing I disagree with here is that documentation saying more about why not to use this prop and nothing about what it does or why you would use it. Do you mind if I re-frame it to start with what this does and what it's for before getting into using the other props if possible as best practice?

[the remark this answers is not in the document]
darryltec: Sure thing!

The only reason why it went with do not use is children is the most straightforward prop there is and reaching for it is pretty common and known. Compared to a label prop existing. So if you're reaching for children, get them out of there fast if they don't know that label exists. If they do know that label exist, they already know what they want to use it for.

packages/components/src/RadioGroup/RadioOption.tsx
```diff
@@ -1,54 +1,65 @@
...
+  readonly label?: string;
+
   /**
-   * Further description of the label.
+   * Provide children elements to render a custom label component if `label` is
+   * not provided, or associated additional content with the radio button if
+   * `label` is provided.
+   *
+   * Prefer using `label` and `description` over adding child elements if the
+   * content of either would be text.
```
darryltec, line 34: I think we can simplify this and get straight to the point.

```suggestion
   * Render a custom label or additional content below the `label`
   * and `description`.
   *
   * Prefer using `label` and `description` over adding child elements if the
   * content of either would be a string.
```

darryltec (changes requested):
I think this looks pretty good and works!! Gonna trigger a pre-release to make sure we don't hit any typing issues within Jobber. 

While we're waiting, what do you think about the "Radio option with children" being renamed as "Custom Radio Option Content" and update it closer to your screenshot. I believe, that's how it would be used most of the time compared to avatars.

maybe go with

- Fixed time
Select from the 15 mins interval
[ 15 mins ] [ 30 mins ] [ 45 mins ] [ 60 mins ]
- Specific time
Set your reminder down to the minute
Remind me at [ 42 | mins ] and [ 3 | sec ] interval

darryltec (approved):
Loooooks good homie! Thanks for contributing to improve Atlantis! Did not see type issues in Jobber

PR #1472, opened 2023-08-03
fix(components): Fix Tabbing on InputDate

ZakaryH:
## Motivations

There are issue with keyboard behavior
- Tabbing will never leave the date picker while open
- With the enableTabLoop disabled, the date picker does not close when losing focus

## Changes

Added a listener to close the datepicker on tab press if none of the elements in the date picker, including the input or custom activator, are the currently focused element.

### Added
Close on Tab press.

### Changed
Consecutive Tab presses no longer loop, locked in the open date picker.

### Deprecated

N/A

### Removed

Technically removed the debatable behavior of keeping focus trapped in the open dialog. There are arguments for that pattern even if I personally don't love it.
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

### Fixed

Inability to Tab out of date picker to close.

### Security
N/A

## Testing
**DatePicker**
As a standalone component, invoke it with your activator. Give focus to your activator with the keyboard which should immediately open the date picker.

Tab through however many elements separate your current focus position and the date picker. Be aware this can be anywhere between 0 and N elements depending entirely on your custom input/activator. 

For example the DatePicker "Basic" example will immediately put focus on a day in the calendar, meaning the next tab will close it.

Tabbing out of the input/activator should keep the DatePicker open. Shift tabbing focus entirely out of both the picker and activator should close the picker. Tabbing past the Next/Previous month buttons and the days which are treated as a single tabbable group should also close the date picker.

**Input Date**
Pretty much the same thing as above except with InputDate we know what structure to expect and starting from the input it should take 5 tabs to lose focus in the foward direction.

Focus the input with keyboard, once again launching the picker. Tab again to focus the icon within the initial input. Tab again to focus the picker and beyond to close it. Same idea for Shift Tab, blurring away from the initial input should also close the picker.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.test.tsx
+++ b/packages/components/src/DatePicker/DatePicker.test.tsx
@@ -158,7 +158,7 @@ describe("Tabbing behavior", () => {
     jest.useRealTimers();
   });
 
-  it("tabbing should not hide the calendar when inline", async () => {
+  it("should not hide the calendar when inline and tabbing", async () => {
     const monthChangeHandler = jest.fn();
     const { queryByText } = render(
       <DatePicker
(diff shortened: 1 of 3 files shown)


ZakaryH:
Kapture 2023-08-03 at 21 36 32

PR #1473, opened 2023-08-04
feat(components-native): Add Icon prop to Banner

taylorvnoj:
## Motivations

Update Banner to take an `icon` prop on web and mobile

### Added

- `children` is now being returned in **mobile** Banner
- `icon` added as a prop to both **web** and **mobile** banners
- Updated the Upcoming Breaking Changes page to include mobile's deprecated Banner props

### Web
`icon` prop exists in web banner:
Screenshot 2023-08-11 at 10 25 25 AM

Screenshot 2023-08-11 at 10 25 36 AM

### Mobile
`text` & `details` are **deprecated** : 
I initially debated between deprecating `text` or using `ts-xor`.  I decided to go with the former because:
1. I want to, as closely as possible, standardize the way `Banner` is used across web and mobile without completely removing the `text` prop.  We should encourage consumers to use `children` instead
2. A `children` prop is used *far more* than `text` in other components
3. Rarely, if (hopefully) ever, would someone use a banner without any text in it.  Naturally they will put the text as a child.  If they see the `text` prop, they will see it is deprecated and be guided to use `children` instead.  

This means that I have restructured the order of how `children`, `text` and `details` are displayed. Order is: 
1. `children`
2. `text`
3. `details`

It is important to note that these three props are all within a content container.  Using one or all three will render them in order with proper styling.  See here:

#### Using `children`
Screenshot 2023-08-11 at 8 12 52 AM

#### Using `text`
Screenshot 2023-08-11 at 8 13 17 AM

#### Using `children` with `details`
Screenshot 2023-08-11 at 8 11 05 AM

#### And even if someone decides to use `details` alone

Screenshot 2023-08-08 at 2 14 47 PM

I also refactored the mobile stories for Banner to use `children` instead of the `text` prop to promote this change

## Testing

Mobile:
- [ ] test using different props (`children` vs `text`) with `details`, `icon` and make sure styling stays as it always has

Web:
- [ ] test `icon` prop in banner with `action`, etc at different breakpoints to make sure nothing gets janky 

---

Random photo of Atlantis


packages/components/src/Banner/BannerIcon.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
```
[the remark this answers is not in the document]
taylorvnoj: Sounds good - I debated this for a while so I'll change it

packages/components-native/src/Banner/Banner.tsx
```diff
@@ -21,8 +24,18 @@ export function Banner({
...
     >
       <Content childSpacing="small">
-        <Text level="textSupporting">{text}</Text>
-        {details && <TextList items={details} level="textSupporting" />}
+        <View style={styles.bannerContent}>
+          {icon && <BannerIcon icon={icon} />}
+          <View style={styles.contentContainer}>
+            {children}
+            {text && (
+              <div style={styles.textContainer}>
```
MichaelParadis, line 32: React Native doesn't have `div`

packages/components-native/src/Banner/types.ts
```diff
@@ -29,22 +29,29 @@ export interface BannerProps {
...
   /**
-   *  Optional custom content that can be displayed below the text and details
+   * Custom content to be displayed in the banner
    */
-  readonly children?: ReactNode;
+  readonly children?: React.ReactElement | string;
 
   /**
-   * Text to display below the 'text' prop. Will display a single message or bullet points of messages if multiple are provided
+   * **Deprecated**: Use `<TextList level="textSupporting" />` instead
```
MichaelParadis, line 37:
```suggestion
   * **Deprecated**: Use `children` with a <TextList level="textSupporting" />` instead
```

packages/components-native/src/Banner/types.ts
```diff
@@ -29,22 +29,29 @@ export interface BannerProps {
...
   /**
-   * Text to display below the 'text' prop. Will display a single message or bullet points of messages if multiple are provided
+   * **Deprecated**: Use `children` with a `<TextList level="textSupporting" />` instead
+   * @deprecated
    */
   readonly details?: string[];
 
   /**
-   * The primary text to display in the banner
+   * **Deprecated**: Use `children` instead
```
MichaelParadis, line 43:
```suggestion
   * **Deprecated**: Use `children`with a `<Text level="textSupporting" />` or a `string` instead
```
Minor suggestion. May need to do locally in case of a lint error

taylorvnoj: Nice catch - done here https://github.com/GetJobber/atlantis/pull/1473/commits/db39a634c619ef8d3a54e54b3982597febf41543

MichaelParadis (approved):
LGTM!

PR #1478, opened 2023-08-09
build(components-native): Forbid html elements in components-native

MichaelParadis:
## Motivations

It is easy to accidentally use html elements in `@jobber/components-native` which do not work. This PR adds an eslint rule to prevent this

### Added

- Added eslint rule to `@jobber/components-native` to prevent accidentally using html elements in React Native

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Verify disallowed elements can't be used in `packages/components-native`
- Verify disallowed elements can be used in `packages/components`
- Verify disallowed elements can be used in `docs/` storybook files

---

Random photo of Atlantis


packages/components-native/.eslintrc.js
```diff
@@ -6,5 +6,18 @@ module.exports = {
         allow: ["**"],
       },
     ],
+    "react/forbid-elements": [
```
darryltec, line 9: thoughts on dropping this on eslintrc under root and targeting both components-native and Mobile.stories.tsx?

```js
module.exports = {
  ...
  overrides: [
    {
      files: ["packages/components-native/**", "docs/**/Mobile.stories.tsx"],
      rules: {
        ...
      },
    },
  ]
}

Comment (MichaelParadis) [reply packages/components-native/.eslintrc.js]: @darryltec updated it

File: docs/components/FormatFile/Mobile.stories.tsx [hunk 2]
```diff
@@ -14,9 +15,9 @@ export default {
 } as ComponentMeta<typeof FormatFile>;
 
 const BasicTemplate: ComponentStory<typeof FormatFile> = args => (
-  <div style={{ width: 290 }}>
+  <View style={{ flex: 1 }}>
```
darryltec, line 18: you could probably just remove this. flex 1 would make it full width which the formatfile does it be default

MichaelParadis: @darryltec I will take a look, right now dealing with the Toast lint error

docs/components/Toast/Mobile.stories.tsx
```diff
@@ -20,11 +21,14 @@ export default {
     },
   },
   decorators: [
-    Story => (
-      <div style={{ height: "90vh" }}>
-        <Story />
-      </div>
-    ),
+    Story => {
+      const height = window.innerHeight * 0.9;
```
MichaelParadis, line 25: Did this because `vh` doesn't work with React Native. May need to update this when we get things running through expo

docs/components/FormatFile/Mobile.stories.tsx
MichaelParadis: @darryltec updated it

.eslintrc.js
```diff
@@ -71,5 +71,23 @@ module.exports = {
...
+        "react/forbid-elements": [
+          "error",
+          {
+            forbid: [
+              { element: "div", message: "Use `<View>` instead" },
+              { element: "span", message: "Use `<View>` instead" },
+              { element: "button", message: "Use `<Button/>` instead" },
+              { element: "a", message: "Use `<AutoLink/>` instead" },
+              { element: "img", message: "Use `<Image/>` instead" },
+              { element: "input", message: "Use a Form input instead" },
```
darryltec, line 86:
```suggestion
              { element: "input", message: "Use one of our `Input` components instead" },
```

docs/components/Toast/Mobile.stories.tsx
darryltec: This only works if you load that page directly. If you're browsing from another place to here, that height is the initial height of the iframe since it animates from really big to really small. 

I'd actually suppress the error here since decorators are internal to storybook and won't show on the code tab.

darryltec (approved):
:shipit:

PR #1479, opened 2023-08-09
docs: Added upcoming breaking changes to changelogs

MichaelParadis:
## Motivations

We have deprecated things but not really kept track of them somewhere. This PR adds a page for tracking these upcoming breaking changes

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/changelog/breaking-changes.stories.mdx
+++ b/docs/changelog/breaking-changes.stories.mdx
@@ -1,0 +1,35 @@
+import { Meta } from "@storybook/addon-docs";
+
+<Meta
+  title="Changelog/Upcoming Breaking Changes"
+  parameters={{ previewTabs: { canvas: { hidden: true } } }}
+/>
+
+# Upcoming Breaking Changes
+
+## `@jobber/components@5.x.x`
+
+### Autocomplete
+
+Removing `inputRef` from `Autocomplete` component. Use `ref` instead.
+
+Before:
+
+```tsx
+const inputRef = React.useRef(null);
+<Autocomplete inputRef={inputRef} />;
+```
+
+After:
+
+```tsx
+const inputRef = React.useRef(null);
+<Autocomplete ref={inputRef} />;
+```
+
+### showToast
+
+Removing `action` and `actionLabel` from `showToast`.
+
+Since Toasts are shown for a limited amount of time it doesn't make sense to
+show an action that may disappear before the user has a chance to click it.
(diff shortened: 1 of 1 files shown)


docs/changelog/breaking-changes.stories.mdx
```diff
@@ -0,0 +1,35 @@
...
+  parameters={{ previewTabs: { canvas: { hidden: true } } }}
+/>
+
+# Upcoming Breaking Changes
+
+## `@jobber/components@5.x.x`
+
+### Autocomplete
+
+Removing `inputRef` from `Autocomplete` component. Use `ref` instead.
```
darryltec, line 14: If our reasoning behind this is that `inputRef` doesn't work then we can safely assume the other ones that exist on other input doesn't work as well. Otherwise, we risk having more prop inconsistencies since it would still exist on some.

docs/changelog/breaking-changes.stories.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+Before:
+
+```tsx
+const inputRef = React.useRef(null);
```
darryltec, line 19:
```suggestion
const inputRef = useRef(null);
```

docs/changelog/breaking-changes.stories.mdx
```diff
@@ -0,0 +1,35 @@
...
+<Autocomplete inputRef={inputRef} />;
+```
+
+After:
+
+```tsx
+const inputRef = React.useRef(null);
```
darryltec, line 26:
```suggestion
const inputRef = useRef(null);
```

MichaelParadis: Fair enough. Updating now

docs/changelog/breaking-changes.stories.mdx
darryltec: Doesn't that defeat the purpose of this? We're saying we want to expose the breaking change we want to do but now you're saying we shouldn't 🤔

MichaelParadis: I think we should expose breaking changes that have been done and are in the version that is on NPM. Mainly so people can make changes ahead of time. For example if we said that this was for every input, and people wanted to fix code ahead of the breaking change then functionality would break because it hasn't been done yet

MichaelParadis: I think this should be for more actionable breaking changes, i.e. things people can fix ahead of time

PR #1481, opened 2023-08-10
fix(components): Prevent Closing Parent on DatePicker ESC

ZakaryH:
## Motivations

When the InputDate, or DatePicker is used within another component that closes on ESC key press, when the DatePicker is open pressing ESC will close both the picker and the parent component. This is not what we want it to do in that scenario. It should only close the picker, and require another ESC if you want to close say the Modal.

### Added

N/A

### Changed

N/A

### Deprecated

N/A

### Removed

N/A

### Fixed

While the date picker is open, pressing ESC will only close the picker. If there is some parent component that can also be closed on ESC, it will remain open.

### Security

N/A

## Testing

Setup a component that listens to ESC presses for closing, such as a `Modal`, inside there add an `InputDate` or `DatePicker`. Activate the date picker so that the picker is open. Now press ESC and only the picker should close.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -176,6 +176,8 @@ function useEscapeKeyToCloseDatePicker(
 
   const escFunction = (event: KeyboardEvent) => {
     if (event.key === "Escape" && open) {
+      // Close the picker ourselves and prevent propagation so that ESC presses with the picker open
+      // do not close parent elements that may also be listening for ESC presses such as Modals
       pickerRef.current?.setOpen(false);
       event.stopPropagation();
     }
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, useEffect, useState } from "react";
+import React, { ReactElement, useEffect, useRef, useState } from "react";
 import classnames from "classnames";
 import ReactDatePicker from "react-datepicker";
 import { XOR } from "ts-xor";
@@ -112,6 +112,7 @@ export function DatePicker({
   const datePickerClassNames = classnames(styles.datePicker, {
     [styles.inline]: inline,
   });
+  const { pickerRef } = useEscapeKeyToCloseDatePicker(open, ref);
 
   if (smartAutofocus) {
     useRefocusOnActivator(open);
@@ -121,6 +122,7 @@ export function DatePicker({
   return (
     <div className={wrapperClassName} ref={ref}>
       <ReactDatePicker
+        ref={pickerRef}
         calendarClassName={datePickerClassNames}
         showPopperArrow={false}
         selected={selected}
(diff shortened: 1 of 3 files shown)


packages/components/src/InputDate/InputDate.test.tsx
```diff
@@ -180,9 +173,44 @@ it("displays the calendar when input is focused with a click", () => {
...
 
-  act(() => {
-    fireEvent.click(input);
-  });
+  fireEvent.click(input);
 
   expect(getByText("15")).toBeInTheDocument();
 });
+
+describe("when InputDate is used within a Modal", () => {
```
ZakaryH, line 181: this is definitely more of an integration test than purely a unit test, but I can't think of another effective way to verify this behavior since in isolation a test asserting that on ESC press the picker closes would pass even without this PR

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -165,3 +167,28 @@ export function DatePicker({
...
+function useEscapeKeyToCloseDatePicker(
+  open: boolean,
+  ref: React.RefObject<HTMLDivElement>,
+): { pickerRef: React.RefObject<ReactDatePicker> } {
+  const pickerRef = useRef<ReactDatePicker>(null);
+
+  const escFunction = (event: KeyboardEvent) => {
+    if (event.key === "Escape" && open) {
+      pickerRef.current?.setOpen(false);
+      event.stopPropagation();
```
darryltec, line 180: should probably explain this since it has a big role in this whole thang

ZakaryH: shore!

packages/components/src/InputDate/InputDate.test.tsx
darryltec: I guess, this would also work on just the normal datepicker that's triggered by a button. So maybe this test should still be in the DatePicker test. 

Alternatively, you are testing that it shouldn't trigger the escape on elements outside of the DatePicker. So you could test it like so

```tsx
const handleEscape = jest.fn()
const handleKeyDown = e => e.key === "Escape" && handleEscape()
window.addEventListener("keydown", handleKeyDown )
...

<div onKeyDown={handleKeyDown}>
  <DatePicker />
</div>

...

expect(handleEscape).not.toHaveBeenCalled()
```

ZakaryH: correct. this will also apply to the base DatePicker. I think it's simply more likely to have an `InputDate` inside a `Modal` than a `DatePicker` just because of what the components do, but it's entirely possible that you _could_  have it nested in something else listening to ESC.

I'm fine with adding another one like your example in DatePicker, but I do want to have at least 1 test that specifically covers the case we are seeking to address to begin with!

ZakaryH: hmmm I just noticed it auto focuses the picker upon opening for the base DatePicker, so my line of `nextMonthButton.focus();` in the test isn't necessary

even without that line it passes. happy to remove it if we think it's doing more harm than good.

darryltec: let's do it on a different PR. this one be good!

darryltec (approved):
:shipit:

PR #1484, opened 2023-08-11
chore(components): Restrict React DatePicker Package to Patch Versions

ZakaryH:
## Motivations

While doing some InputDate/DatePicker work recently I noticed that newer versions of react-datepicker` have some minor layout differences that mildly break.

The wrapper class in versions under 4.14.0 have this `width: 100%` that plays nice with our `fullWidth` behavior
https://github.com/Hacker0x01/react-datepicker/blob/v4.10.1/src/stylesheets/datepicker.scss#L8

However after that, it's gone
https://github.com/Hacker0x01/react-datepicker/blob/v4.14.0/src/stylesheets/datepicker.scss

Then the `InputDate` is only ever about as wide as the button + the string representation of the date, and not full width like we want it to be with this code.

```
export function InputDate(inputProps: InputDateProps) {
  const formFieldActionsRef = useRef<FieldActionsRef>(null);
  return (
    <DatePicker
      selected={inputProps.value}
      onChange={inputProps.onChange}
      disabled={inputProps.disabled}
      readonly={inputProps.readonly}
      fullWidth={!inputProps.inline}
       ...
```

The fix won't be the worst, we'll just need to shim in some of our own styles to make it be full width again. It's just an easy thing to miss since the changelog doesn't make it super obvious imo.
https://github.com/Hacker0x01/react-datepicker/releases/tag/v4.14.0

So, with that in mind I'm going to make it a bit harder to inadvertently break this by changing it from a caret which accepts patch and minor to the tilde that only accepts patch versions.

### Added

n/a

### Changed

`^` -> `~`

### Deprecated

n/a

### Removed

n/a

### Fixed

n/a

### Security

n/a

## Testing

Everything should be exactly the same. If it's not, that's deeply concerning.

---

Random photo of Atlantis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45693,7 +45693,7 @@
         "framer-motion": "6.3.16",
         "lodash": "^4.17.21",
         "react-countdown": "^2.3.2",
-        "react-datepicker": "^4.10.0",
+        "react-datepicker": "~4.10.0",
         "react-dropzone": "^11.0.2",
         "react-hook-form": "^7.43.7",
         "react-markdown": "^8.0.3",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -39,7 +39,7 @@
     "framer-motion": "6.3.16",
     "lodash": "^4.17.21",
     "react-countdown": "^2.3.2",
-    "react-datepicker": "^4.10.0",
+    "react-datepicker": "~4.10.0",
     "react-dropzone": "^11.0.2",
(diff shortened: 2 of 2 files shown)


scotttjob (approved):
I'm going back through old PRs to see about approving anything hanging around for a while.

I think this change should be fine to either merge as-is, or we can close the PR and create a follow-up ticket to fully address the issues that Zak mentioned along with upgrading the package to 4.13 (or above) as Darryl mentioned.

PR #1488, opened 2023-08-15
chore(components): Add counter and title to DataList

rodrigoeidelvein:
## Motivations

Adding the title and counter to DataList. They're both optional.

When it's loading, it will show the Glimmer instead of the results string.

### Added

- 

### Changed

- Added title and counter to DataList

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

You can remove this line and then DataList will show up in Storybook:
https://github.com/GetJobber/atlantis/blob/c48f3d35e38846864f624b50c0e0481e2c19c737/packages/components/src/DataList/DataList.stories.tsx#L15

Use the controls to test the behavior.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -1,3 +1,12 @@
 .wrapper {
   flex: 1;
 }
+.titleContainer {
+  display: flex;
+  align-items: center;
+}
+
+.results {
+  width: 80px;
+  margin-left: var(--space-small);
+}
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -1,15 +1,18 @@
 .wrapper {
   flex: 1;
 }
+
 .titleContainer {
   display: flex;
   align-items: center;
+  margin-bottom: var(--space-small);
 }
 
 .results {
-  width: 80px;
+  min-width: 80px;
   margin-left: var(--space-small);
 }
+
 .header {
   position: sticky;
   top: 0;
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -8,11 +8,6 @@
   margin-bottom: var(--space-small);
 }
 
-.results {
-  min-width: 80px;
-  margin-left: var(--space-small);
-}
-
 .header {
   position: sticky;
   top: 0;
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -60,6 +60,7 @@ const Template: ComponentStory<typeof DataList> = args => {
       nextFetchPolicy: "cache-first",
       client: apolloClient,
     },
+
     getCollectionByPath(items) {
       return items?.allPeople;
     },
@@ -97,6 +98,7 @@ const Template: ComponentStory<typeof DataList> = args => {
         created: "Created",
       }}
       loading={loadingInitialContent}
+      totalCount={data?.allPeople.totalCount}
     >
       <DataList.Layout>
         {(item: DataListItemType<typeof mappedData>) => (
@@ -126,12 +128,12 @@ const Template: ComponentStory<typeof DataList> = args => {
 };
 
 export const Basic = Template.bind({});
-Basic.args = {
-  title: "All Clients",
-  loading: false,
-  showCount: true,
-  totalCount: 50,
-};
+// Basic.args = {
+//   title: "All Clients",
+//   loading: false,
+//   showCount: true,
+//   totalCount: 50,
+// };
 export const EmptyState = Template.bind({});
 EmptyState.args = {
   data: [],
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -131,7 +131,6 @@ export const Basic = Template.bind({});
 Basic.args = {
   title: "All Clients",
   loading: false,
-  showCount: true,
 };
 export const EmptyState = Template.bind({});
 EmptyState.args = {
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -12,10 +12,19 @@ export default {
     viewMode: "story",
   },
   // Comment this out to make it show up in storybook
+  excludeStories: ["Basic"],
   decorators: [
     // Detach from Storybook's layout
     Story => (
-      <div>
+      <div
+        style={{
+          position: "fixed",
+          inset: 0,
+          overflow: "auto",
+          display: "flex",
+          flexDirection: "column",
+        }}
+      >
         <Story />
       </div>
     ),
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -12,7 +12,6 @@ export default {
     viewMode: "story",
   },
   // Comment this out to make it show up in storybook
-  excludeStories: ["Basic"],
   decorators: [
     // Detach from Storybook's layout
     Story => (
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -88,6 +88,8 @@ const Template: ComponentStory<typeof DataList> = args => {
 
   return (
     <DataList
+      loading={loadingInitialContent}
+      totalCount={data?.allPeople.totalCount}
       {...args}
       data={(args.data as typeof mappedData) || mappedData}
       headers={{
@@ -97,8 +99,6 @@ const Template: ComponentStory<typeof DataList> = args => {
         homePopulation: "Home world population",
         created: "Created",
       }}
-      loading={loadingInitialContent}
-      totalCount={data?.allPeople.totalCount}
     >
       <DataList.Layout>
         {(item: DataListItemType<typeof mappedData>) => (
@@ -128,12 +128,11 @@ const Template: ComponentStory<typeof DataList> = args => {
 };
 
 export const Basic = Template.bind({});
-// Basic.args = {
-//   title: "All Clients",
-//   loading: false,
-//   showCount: true,
-//   totalCount: 50,
-// };
+Basic.args = {
+  title: "All Clients",
+  loading: false,
+  showCount: true,
+};
 export const EmptyState = Template.bind({});
 EmptyState.args = {
   data: [],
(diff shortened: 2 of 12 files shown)


packages/components/src/DataList/DataList.stories.tsx
```diff
@@ -97,6 +98,7 @@ const Template: ComponentStory<typeof DataList> = args => {
         created: "Created",
       }}
       loading={loadingInitialContent}
+      totalCount={data?.allPeople.totalCount}
```
darryltec, line 101: if you move these 2 above `{...args}` you should be able to override them with the `Basic.args`

packages/components/src/DataList/DataList.css
```diff
@@ -1,7 +1,15 @@
 .wrapper {
   flex: 1;
 }
+.titleContainer {
```
darryltec, line 4: nit: 
```suggestion
}

.titleContainer {
```

packages/components/src/DataList/DataList.css
```diff
@@ -1,7 +1,15 @@
...
+  display: flex;
+  align-items: center;
+}
 
+.results {
+  width: 80px;
+  margin-left: var(--space-small);
+}
 .header {
```
darryltec, line 13: nit: 
```suggestion
}

.header {
```

packages/components/src/DataList/DataList.css
```diff
@@ -1,7 +1,15 @@
 .wrapper {
   flex: 1;
 }
+.titleContainer {
+  display: flex;
+  align-items: center;
+}
```
darryltec, line 7: need a 
```suggestion
  align-items: center;
  margin-bottom: var(--space-small);
}
```

packages/components/src/DataList/DataList.css
```diff
@@ -1,7 +1,15 @@
...
 
+.results {
+  width: 80px;
```
darryltec, line 10: This would be better as min-width because this would happen

image

which should make the glimmer still work

```suggestion
  min-width: 80px;
```

packages/components/src/DataList/DataList.tsx
```diff
@@ -39,6 +45,18 @@ export function DataList<T extends DataListObject>({
...
     <div className={styles.wrapper}>
       {/* List title and counter */}
+      <div className={styles.titleContainer}>
+        {title && <Heading level={3}>{title}</Heading>}
+        {showCount && (
+          <div className={styles.results}>
+            {loading ? (
+              <Glimmer size="auto" shape="rectangle" />
+            ) : (
+              <Text>({totalCount} results)</Text>
```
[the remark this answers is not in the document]
darryltec: also

```suggestion
              <Text>({totalCount.toLocaleString()} results)</Text>
```

should make it look nice when you get to thousands

image

packages/components/src/DataList/DataList.types.ts
```diff
@@ -32,4 +32,19 @@ export interface DataListProps<T extends DataListObject> {
...
   readonly children: ReactElement | ReactElement[];
+  /**
+   * The title of the DataList
+   */
+  readonly title?: string;
+
+  /** Should the datalist display the count of items on the DataTable?
+   * @default true
+   */
+  readonly showCount?: boolean;
```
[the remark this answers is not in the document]
rodrigoeidelvein: I love this! Thanks for the suggestion. I removed the `showCount` prop, let me know what you think.

packages/components/src/DataList/DataList.test.tsx
```diff
@@ -5,12 +5,101 @@ import {
...
+            </Grid>
+          )}
+        </DataList.Layout>
+      </DataList>,
+    );
+    expect(screen.getByText("All Clients")).toBeInTheDocument();
+    expect(screen.queryByText("(10 results)")).toBeNull();
+  });
+
+  it("should render the Glimmer when loading", () => {
```
darryltec, line 61: Keep in mind that the moment you do the loading ticket, this will break. Is this testing the loading in general? Or the loading of the totalCount?

packages/components/src/DataList/DataList.tsx
```diff
@@ -39,6 +44,20 @@ export function DataList<T extends DataListObject>({
   return (
     <div className={styles.wrapper}>
       {/* List title and counter */}
+      <div className={styles.titleContainer}>
+        {title && <Heading level={3}>{title}</Heading>}
+        {totalCount !== null && (
```
darryltec, line 49: curious. what is this logic trying to do?

If it's undefined, it's gonna go through and render a (0 results) text or a glimmer.

packages/components/src/DataList/DataList.tsx
```diff
@@ -12,13 +12,18 @@ import {
...
+import { Heading } from "../Heading";
 
 export function DataList<T extends DataListObject>({
   data,
   headers,
   loading = false,
   filterApplied = false,
   children,
+  title,
+  totalCount = 0,
```
darryltec, line 26: if you implement the data list without specifying the totalCount

```tsx
<DataList data={...} headers={...}>
  ...
</DataList>
```

It'll always say 0 results

packages/components/src/DataList/DataList.tsx
```diff
@@ -39,6 +44,20 @@ export function DataList<T extends DataListObject>({
...
+        {title && <Heading level={3}>{title}</Heading>}
+        {totalCount !== null && (
+          <div className={styles.results}>
+            {loading ? (
+              <Glimmer size="auto" shape="rectangle" />
+            ) : (
+              <Text variation="subdued">
+                ({totalCount.toLocaleString()} results)
+              </Text>
+            )}
```
darryltec, line 57: we should find a way to avoid this ternary

packages/components/src/DataList/DataList.types.ts
```diff
@@ -32,4 +32,17 @@ export interface DataListProps<T extends DataListObject> {
    */
   readonly filterApplied?: boolean;
   readonly children: ReactElement | ReactElement[];
+  /**
```
darryltec, line 35:
```suggestion

  /**
```

packages/components/src/DataList/DataList.types.ts
```diff
@@ -32,4 +32,17 @@ export interface DataListProps<T extends DataListObject> {
...
+   */
+  readonly title?: string;
+
+  /**
+   * Total number of items in the DataList.
+   *
+   * This renders an "N result" text with the DataList
+   * that helps users know how many items they have
+   * in the list
+   **/
```
darryltec, line 46:
```suggestion
   */
```

packages/components/src/DataList/DataList.stories.tsx
```diff
@@ -21,7 +21,7 @@ export default {
     viewMode: "story",
   },
   // Comment this out to make it show up in storybook
-  excludeStories: ["Basic", "EmptyState"],
+  // excludeStories: ["Basic", "EmptyState"],
```
darryltec, line 24: Don't forget to update this

```suggestion
  excludeStories: ["Basic", "EmptyState"],
```

darryltec (changes requested):
The logic seems a bit off.

What do you think about this scenarios?
- When I don't set the `totalCount` prop, it shouldn't show the results text
  - The same goes for glimmer when loading. It shouldn't show up
- When I do add a specific `totalCount` number, it should add an `(N results)` text.
  - NEW: It shouldn't show the glimmer regardless since we know there's a number already. No point in adding an extra glimmer.
- When the total count is set to `null` 
  - ...and the data is loading, it should add a glimmer
  - ...and the data is not loading, it should show nothing

packages/components/src/DataList/DataList.test.tsx
rodrigoeidelvein: I added a test ID and queried within it. What do you think?

packages/components/src/DataList/DataList.test.tsx
```diff
@@ -74,6 +129,20 @@ describe("DataList", () => {
     });
   });
 
+  it("should render a title when it's provided", () => {
```
darryltec, line 132: I wonder if you should test the inverse of this

packages/components/src/DataList/DataList.tsx
```diff
@@ -55,5 +64,32 @@ export function DataList<T extends DataListObject>({
   );
 }
 
+function ResultsCounter({
```
darryltec, line 67: I'd put this in its own file. I see the appeal of putting everything in the same file but it's also a slippery slope into a gigantic file that is harder to maintain. Remember the phrase "One file, one component".

Also, there's something about the name that feels misleading. It isn't really counting anything internally. It's more presentation. What about `TotalCount` or `DataListTotalCount` if you want to follow the pattern under the components folder?

packages/components/src/DataList/DataList.tsx
```diff
@@ -55,5 +64,32 @@ export function DataList<T extends DataListObject>({
...
+  }
+
+  if (typeof totalCount === "number") {
+    output = (
+      <Text variation="subdued">({totalCount.toLocaleString()} results)</Text>
+    );
+  }
+
+  return (
+    <div className={styles.results} data-testid="results">
```
darryltec, line 88: This isn't really broadcasted but i'd like to make that testID more unique and specific to Atlantis. 

[Internal post about it](https://jobber.atlassian.net/wiki/spaces/TEAMATL/pages/2476933462/How+should+we+declare+test+ID+s+in+Atlantis+components)

Let's think about testing in general. That test ID is too generic that someone outside of atlantis would easily write a testID called "results" in their component. They then get confused as to why jest is complaining about finding 2 elements with that test ID. 😆

darryltec (commented):
Just a few thoughts to ponder on. If you agree with them let's do the change. Otherwise, I'll approve it.

packages/components/src/DataList/DataList.test.tsx
rodrigoeidelvein: Any ideas how this could be done? Because if a title is not provided, there's nothing to search for.

packages/components/src/DataList/DataList.tsx
rodrigoeidelvein: Agreed! Moved it. Thanks for the suggestion.

I created it on the main file because it was not actually being exported as `DataListLayout` or `EmptyState`, but it does make sense to keep one component per file to keep it organized.

packages/components/src/DataList/DataList.test.tsx
darryltec: hmmm. `within` + test id? you already have the title container div haha

darryltec (approved):
LGTM

PR #1492, opened 2023-08-17
chore(components): make the data list list part happen

darryltec:
## Motivations

This creates the initial way of setting the list layout so we can build upon this.

### Added

- DataList layout capability that determines the list item and header layout

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/DataList/DataList.stories.tsx
```diff
@@ -12,43 +21,113 @@ export default {
     viewMode: "story",
   },
   // Comment this out to make it show up in storybook
-  excludeStories: ["Basic"],
+  // excludeStories: ["Basic"],
```
darryltec, line 24: Note to self: Don't forget to comment out
```suggestion
  excludeStories: ["Basic"],
```

packages/components/src/DataList/DataList.stories.tsx
```diff
@@ -12,43 +21,113 @@ export default {
...
+    </DataList>
+  );
+};
+
+export const Basic = Template.bind({});
 
+export const EmptyState = Template.bind({});
 EmptyState.args = {
-  items: [],
+  data: [],
```
MichaelParadis, line 132: Minor thing, but because it uses the `Template` loading can't be set to true and stay there because of the `useCollectionQuery` setting it to false to demonstrate the EmptyState not showing while loading. Non-blocking though

packages/components/src/DataList/DataList.stories.tsx
```diff
@@ -12,43 +21,113 @@ export default {
...
 } as ComponentMeta<typeof DataList>;
 
-const Template: ComponentStory<typeof DataList> = args => (
-  <DataList {...args} />
-);
+const Template: ComponentStory<typeof DataList> = args => {
+  const {
+    data,
+    /* See useCollectionQuery for example on how to load more */
+    // refresh,
```
MichaelParadis, line 51: When we do the infinite scroll we should show this off (later thing)

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -0,0 +1,143 @@
...
+    Object.keys(item).reduce((acc, key: keyof DataListElements) => {
+      const currentItem = item[key];
+
+      if (!currentItem) {
+        return acc;
+      }
+
+      if (key === "tags" && Array.isArray(currentItem)) {
+        acc[key] = (
+          // TODO: Create a component specific for this with the experience we want
```
MichaelParadis, line 51: Just confirming this todo is a later thing and not in this PR.

darryltec: yup. added a subtask for it in the same story

packages/components/src/DataList/DataList.stories.tsx
darryltec: totally!

packages/components/src/DataList/DataList.stories.tsx
darryltec: oh, hmm. lemme check!

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+      return acc;
+    }, {} as DataListElements),
+  );
+}
+
+/**
+ * Generate the header elements with the default styling
+ */
+export function generateHeaderFromData<T extends DataListObject>(
```
MichaelParadis, line 80: Super nitpicky, but I am not a big fan of the `FromData` mainly because this is actually from the `headers` prop and the `data` prop is different and isn't used in the `headers` besides for helping with the typescript magic.

I can't think of a better name so if you have nothing I am not going to block it for this

darryltec: hmm yeah, i'll think on it. not crazy about it either

MichaelParadis: @darryltec thoughts on changing the `headers` prop to `columns`?

darryltec: Thought about this more. I think we should defer that back to the sorting story. Mostly because if our main reason is for it to be flexible for when sorting comes, it could also go either way. Following the pattern of "simple props" declaring sort should be on the sort prop like so `sort={{sortable: ["label", "address"], onSort: (sorted) => void}}`

MichaelParadis (approved):
LGTM!

PR #1495, opened 2023-08-22
chore(components): add data list tags pattern

darryltec:
## Motivations

There's a pretty specific tags pattern in the list where it overflows and counts the ones that aren't visible/half-visible to convey that there are more tags.

This aims to add that pattern.

This is the logic behind it. Say you have 5 tags
- If only 3 are visible, it'll add a + 2 count
- If all 5 are visible, it won't add the + counter
- If 4 are visible and the last is half visible, it'll count that as not visible and add +1
  - This is a designer-approved logic.

image

`IntersectionObserver` is also not available in Jest so a plugin needs to be added to mock it within the test.

### Added

- DataList Tag component
- Also test

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Resize your screen to the point that the attributes column content doesn't fit
- It should start counting the labels that gets cut-off/not visible

---

[In Atlantis we use Github's built-in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -72,10 +72,7 @@ const Template: ComponentStory<typeof DataList> = args => {
   //   loadingNextPage,
   // );
 
-  const dummyString =
-    "Lorem ipsum, dolor sit amet consectetur adipisicing elit. Laudantium obcaecati nemo totam harum, repudiandae nisi necessitatibus velit consequuntur corporis similique".split(
-      " ",
-    );
+  const randomTags = ["SW", "commercial", "pets", "fence"];
   const items = data?.allPeople.edges || [];
   const mappedData = items.map(({ node }) => ({
     label: node.name,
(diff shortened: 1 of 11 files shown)


packages/components/src/DataList/DataList.stories.tsx
```diff
@@ -21,7 +21,7 @@ export default {
     viewMode: "story",
   },
   // Comment this out to make it show up in storybook
-  excludeStories: ["Basic", "EmptyState"],
+  // excludeStories: ["Basic", "EmptyState"],
```
darryltec, line 24: Don't forget to uncomment this out

```suggestion
  excludeStories: ["Basic", "EmptyState"],
```

taylorvnoj:
Nice - this looks really great.

Quick question - is the columns overlapping here a known issue?
Screenshot 2023-08-22 at 1 15 04 PM

darryltec:
@taylorvnoj yup. that's a job for the responsive of either passing in your responsive grid or what Michael is doing.

taylorvnoj (approved):
LGTM - don't forget your comment suggestion and I assume we'll remove the `randomTags` later/before actually exposing this

PR #1496, opened 2023-08-22
chore(components): better keys for data list items

darryltec:
## Motivations

We shouldn't use the index as a key for mapped children. This [react doc](https://react.dev/learn/rendering-lists#why-does-react-need-keys) goes into detail as to why.

Gist of it:
- Say you have 10 items
- If you insert an item between 4 and 5 you'd now have 11 items
- Since the 5th element is now the 6th and so on, it'll rerender the previous 5 to 10 elements since the index changed.
- Now imagine you have 1,000 and you inserted between 1 and 2

### Added

- Explicitly set and ID as per react's suggestion
- Added label as a required key
- Added tags as an optional key

### Changed

- adjusted types accordingly

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.tsx
+++ b/packages/components/src/DataList/DataList.tsx
@@ -53,8 +53,7 @@ export function DataList<T extends DataListObject>({
       </div>
       {layout &&
         elementData.map((child, i) => (
-          // TODO: Don't use index as key. Might have to force an ID on the data JOB-76773
-          <div className={styles.listItem} key={i}>
+          <div className={styles.listItem} key={data[i].id}>
             {layout(child)}
           </div>
         ))}
--- a/packages/components/src/DataList/DataList.types.ts
+++ b/packages/components/src/DataList/DataList.types.ts
@@ -11,7 +11,18 @@ export type DataListItemTypeFromHeader<
 > = Record<keyof THeader, ReactElement>;
 
 export interface DataListObject {
+  /**
+   * The ID of the data.
+   *
+   * This is used as a key when looping through the data to prevent accidental
+   * rerender when the order of data changes.
+   */
   readonly id: string | number;
+
+  /**
+   * Styles the string as an emphasized text that differs from other keys.
+   */
+  readonly label?: string | ReactElement;
   readonly [key: string]: ReactNode | Date;
 }
 
@@ -24,20 +35,22 @@ export interface DataListProps<T extends DataListObject> {
   readonly headers: DataListHeader<T>;
 
   /**
-   * Tell the DataList if the data loading
+   * Shows the loading state of the DataList.
+   *
    * @default false
    */
   readonly loading?: boolean;
 
   /**
    * Temporary prop for setting default state for if filters are applied
+   *
    * @default false
    */
   readonly filterApplied?: boolean;
   readonly children: ReactElement | ReactElement[];
 
   /**
-   * The title of the DataList
+   * The title of the DataList.
    */
   readonly title?: string;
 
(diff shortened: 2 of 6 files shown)


packages/components/src/DataList/DataList.tsx
```diff
@@ -53,8 +53,7 @@ export function DataList<T extends DataListObject>({
       </div>
       {layout &&
         elementData.map((child, i) => (
-          // TODO: Don't use index as key. Might have to force an ID on the data JOB-76773
-          <div className={styles.listItem} key={i}>
+          <div className={styles.listItem} key={data[i].id}>
```
ZakaryH, line 56: this feels weird to use the index to lookup the id rather than using `child.id` when we are already iterating over the data set

I see in `generateListItemElements` we're modifying this returned array to be something else entirely, so I get why it's necessary with the way this is set up... buuut that is something we control

one thing we could do is instead change the data structure returned from `generateListItemElements` to have 2 keys: `id` and `elements` or something like that. then we'd be able to do `child.id` and then `layout(child.elements)`

that said my argument is mostly "this feels weird". I haven't tested this out with a massive list to see if there are any actual performance concerns with an extra lookup on a big data set

ZakaryH:
also I know it's not the concern of this PR, but I'm not sure how I feel about not doing any validation or logic around the amount of `Grid.cell` s passed in wrt to their `size`

we have a concept of `MAX_COLUMNS` but we're not keeping track of what we've been given for any given size (xs, s, md etc. ) so you can pass it a sum of sizes that exceed 12 and it will still try (and fail) to render it nicely. though maybe the fact that it breaks is enough for you to realize "oh I shouldn't do that"

packages/components/src/DataList/DataList.types.ts
```diff
@@ -5,12 +5,36 @@ export type DataListItemType<T extends DataListObject[]> = Record<
...
+  /**
+   * Creates a styled list of tags that overflows and is truncated with a +N.
+   */
+  readonly tags?: string[];
+
+  /**
+   * Support any key. The keys in this object are used as way to determine the
+   * keys you can use on the DataList header, layout, etc.
+   */
+  readonly [key: string]: ReactNode | Date;
```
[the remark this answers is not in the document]
darryltec: oh sorry. that key isn't talking about the key of this PR. that one is for the data not the component key.

packages/components/src/DataList/DataList.tsx
darryltec: I honestly, just didn't want to fight with types on this one haha

packages/components/src/DataList/DataList.types.ts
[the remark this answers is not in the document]
ZakaryH: yeah I know that's the React key while iterating whereas this is a separate key and this behavior was existing - but I still feel the same way about it

we most definitely do not need to deal with it here in this PR!

packages/components/src/DataList/DataList.tsx
ZakaryH: fair.

ok after doing what I hope is a reasonable experiment at seeing the performance, to me it looks negligible to lookup the id again rather than using an id that we have provided

image

ZakaryH (approved):
for the scope of what this is, LGTM

ZakaryH:
@darryltec oh whoops I didn't notice there were conflicts. happy to re-approve once those are un-conflicted!

packages/components/src/DataList/DataList.tsx
darryltec: oooh fancy

darryltec:
@ZakaryH funnily enough, that conflict was from my other PR. Present Derl 🔫 Past Derl

ZakaryH (approved):
still LGTM but I'm not even sure that matters since the approval was never removed after the conflict resolution commit!

darryltec:
It matters on my ❤️ @ZakaryH

PR #1497, opened 2023-08-22
chore(components): JOB-75752 DataList responsivenes

MichaelParadis:
### Added

- Ability to add `DataListLayouts` for various breakpoints
- Added ability to hide the header on specific breakpoints

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

 - Verify the layout changes based on the size prop for `DataList.Layout`s
 - Verify that the `header` can be hidden for specific `DataList.Layout`s

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.const.ts
+++ b/packages/components/src/DataList/DataList.const.ts
@@ -1,2 +1,14 @@
+/**
+ *  Grab the type from breakpoints const
+ */
+type ValuesOf<T extends readonly unknown[]> = T[number];
+
+export type Breakpoints = ValuesOf<typeof BREAKPOINTS>;
+
 export const EMPTY_FILTER_RESULTS_MESSAGE = "No Results for Selected Filters";
 export const EMPTY_FILTER_RESULTS_ACTION_LABEL = "Clear Filters";
+
+/**
+ * Breakpoints that we support
+ */
+export const BREAKPOINTS = ["xs", "sm", "md", "lg", "xl"] as const;
--- a/packages/components/src/DataList/DataList.const.ts
+++ b/packages/components/src/DataList/DataList.const.ts
@@ -12,3 +12,11 @@ export const EMPTY_FILTER_RESULTS_ACTION_LABEL = "Clear Filters";
  * Breakpoints that we support
  */
 export const BREAKPOINTS = ["xs", "sm", "md", "lg", "xl"] as const;
+
+export const BREAKPOINT_SIZES: Record<Breakpoints, number> = {
+  xs: 0,
+  sm: 490,
+  md: 768,
+  lg: 1080,
+  xl: 1440,
+};
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -17,4 +17,22 @@
 .listItem {
   padding: var(--space-small);
   border-bottom: var(--border-base) solid var(--color-border);
+  @media (--small-screens-and-up) {
+    display: var(--list-item-sm-display);
+  }
+
+  @media (--medium-screens-and-up) {
+    display: var(--list-item-md-display);
+  }
+
+  @media (--large-screens-and-up) {
+    display: var(--list-item-lg-display);
+  }
+
+  @media (--extra-large-screens-and-up) {
+    display: var(--list-item-xl-display);
+  }
+}
+.hide {
+  display: none;
 }
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -9,6 +9,27 @@
   border-bottom: var(--border-thick) solid var(--color-border);
   background-color: var(--color-surface);
 }
+.headerContent {
+  @media (--small-screens-and-below) {
+    display: var(--data-list-xs-display);
+  }
+
+  @media (--small-screens-and-up) {
+    display: var(--data-list-sm-display);
+  }
+
+  @media (--medium-screens-and-up) {
+    display: var(--data-list-md-display);
+  }
+
+  @media (--large-screens-and-up) {
+    display: var(--data-list-lg-display);
+  }
+
+  @media (--extra-large-screens-and-up) {
+    display: var(--data-list-xl-display);
+  }
+}
 
 .headerLabel > * {
   font-weight: 500;
@@ -19,25 +40,22 @@
   border-bottom: var(--border-base) solid var(--color-border);
 
   @media (--small-screens-and-below) {
-    display: var(--list-item-xs-display);
+    display: var(--data-list-xs-display);
   }
 
   @media (--small-screens-and-up) {
-    display: var(--list-item-sm-display);
+    display: var(--data-list-sm-display);
   }
 
   @media (--medium-screens-and-up) {
-    display: var(--list-item-md-display);
+    display: var(--data-list-md-display);
   }
 
   @media (--large-screens-and-up) {
-    display: var(--list-item-lg-display);
+    display: var(--data-list-lg-display);
   }
 
   @media (--extra-large-screens-and-up) {
-    display: var(--list-item-xl-display);
+    display: var(--data-list-xl-display);
   }
-}
-.hide {
-  display: none;
 }
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -25,25 +25,3 @@
   padding: var(--space-small);
   border-bottom: var(--border-base) solid var(--color-border);
 }
-
-.layout {
-  @media (--small-screens-and-below) {
-    display: var(--data-list-xs-display);
-  }
-
-  @media (--small-screens-and-up) {
-    display: var(--data-list-sm-display);
-  }
-
-  @media (--medium-screens-and-up) {
-    display: var(--data-list-md-display);
-  }
-
-  @media (--large-screens-and-up) {
-    display: var(--data-list-lg-display);
-  }
-
-  @media (--extra-large-screens-and-up) {
-    display: var(--data-list-xl-display);
-  }
-}
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -17,6 +17,11 @@
 .listItem {
   padding: var(--space-small);
   border-bottom: var(--border-base) solid var(--color-border);
+
+  @media (--small-screens-and-below) {
+    display: var(--list-item-xs-display);
+  }
+
   @media (--small-screens-and-up) {
     display: var(--list-item-sm-display);
   }
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -2,6 +2,12 @@
   flex: 1;
 }
 
+.titleContainer {
+  display: flex;
+  align-items: center;
+  margin-bottom: var(--space-small);
+}
+
 .header {
   position: sticky;
   top: 0;
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -10,7 +10,7 @@
   background-color: var(--color-surface);
 }
 
-.headerLabel {
+.headerLabel > * {
   font-weight: 500;
 }
 
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -9,6 +9,7 @@
   border-bottom: var(--border-thick) solid var(--color-border);
   background-color: var(--color-surface);
 }
+
 .headerContent {
   @media (--small-screens-and-below) {
     display: var(--data-list-xs-display);
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -16,28 +16,6 @@
   background-color: var(--color-surface);
 }
 
-.headerContent {
-  @media (--small-screens-and-below) {
-    display: var(--data-list-xs-display);
-  }
-
-  @media (--small-screens-and-up) {
-    display: var(--data-list-sm-display);
-  }
-
-  @media (--medium-screens-and-up) {
-    display: var(--data-list-md-display);
-  }
-
-  @media (--large-screens-and-up) {
-    display: var(--data-list-lg-display);
-  }
-
-  @media (--extra-large-screens-and-up) {
-    display: var(--data-list-xl-display);
-  }
-}
-
 .headerLabel > * {
   font-weight: 500;
 }
@@ -45,7 +23,9 @@
 .listItem {
   padding: var(--space-small);
   border-bottom: var(--border-base) solid var(--color-border);
+}
 
+.layout {
   @media (--small-screens-and-below) {
     display: var(--data-list-xs-display);
   }
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -11,6 +11,7 @@
 .header {
   position: sticky;
   top: 0;
+  z-index: var(--elevation-base);
   padding: var(--space-small);
   border-bottom: var(--border-thick) solid var(--color-border);
   background-color: var(--color-surface);
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -207,6 +207,7 @@ const BreakpointTemplate: ComponentStory<typeof DataList> = args => {
         homePopulation: "Home world population",
         created: "Created",
       }}
+      headerVisibility={{ xs: false, lg: true }}
     >
       <DataList.Layout size="lg">
         {(item: DataListItemType<typeof mappedData>) => (
@@ -223,7 +224,7 @@ const BreakpointTemplate: ComponentStory<typeof DataList> = args => {
           </Grid>
         )}
       </DataList.Layout>
-      <DataList.Layout size={"xs"} showHeader={false}>
+      <DataList.Layout size={"xs"}>
         {(item: DataListItemType<typeof mappedData>) => (
           <Card>
             <Content>
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -76,6 +76,8 @@ const Template: ComponentStory<typeof DataList> = args => {
   // );
 
   const items = data?.allPeople.edges || [];
+
+  const randomTags = ["SW", "commercial", "pets", "fence"];
   const mappedData = items.map(({ node }) => ({
     label: node.name,
     home: node.homeworld.name,
@@ -83,7 +85,11 @@ const Template: ComponentStory<typeof DataList> = args => {
       node.gender,
       node.hairColor?.split(", "),
       node.skinColor?.split(", "),
-    ],
+      ...randomTags.slice(
+        Math.round(Math.random() * 2),
+        Math.round(3 + Math.random() * 4),
+      ),
+    ].filter(t => t !== "n/a"),
     homePopulation: node.homeworld.population?.toLocaleString(),
     created: new Date(node.created),
   }));
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -23,7 +23,7 @@ export default {
     viewMode: "story",
   },
   // Comment this out to make it show up in storybook
-  excludeStories: ["Basic", "EmptyState", "Breakpoints"],
+  // excludeStories: ["Basic", "EmptyState", "Breakpoints"],
   decorators: [
     // Detach from Storybook's layout
     (Story, { viewMode }) => {
@@ -227,23 +227,9 @@ const BreakpointTemplate: ComponentStory<typeof DataList> = args => {
         {(item: DataListItemType<typeof mappedData>) => (
           <Card>
             <Content>
-              <Grid>
-                <Grid.Cell size={{ xs: 8 }}>
-                  <Grid>
-                    <Grid.Cell size={{ xs: 12 }}>{item.label}</Grid.Cell>
-                    <Grid.Cell size={{ xs: 12 }}>{item.tags}</Grid.Cell>
-                  </Grid>
-                </Grid.Cell>
-                <Grid.Cell size={{ xs: 4 }}>
-                  <Grid>
-                    <Grid.Cell size={{ xs: 12 }}>{item.home}</Grid.Cell>
-                    <Grid.Cell size={{ xs: 12 }}>
-                      {item.homePopulation}
-                    </Grid.Cell>
-                  </Grid>
-                </Grid.Cell>
-                <Grid.Cell size={{ xs: 12 }}>{item.created}</Grid.Cell>
-              </Grid>
+              {item.label}
+              {item.home}
+              {item.created}
             </Content>
           </Card>
         )}
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -62,6 +62,7 @@ const Template: ComponentStory<typeof DataList> = args => {
       nextFetchPolicy: "cache-first",
       client: apolloClient,
     },
+
     getCollectionByPath(items) {
       return items?.allPeople;
     },
@@ -89,6 +90,8 @@ const Template: ComponentStory<typeof DataList> = args => {
 
   return (
     <DataList
+      loading={loadingInitialContent}
+      totalCount={null}
       {...args}
       data={(args.data as typeof mappedData) || mappedData}
       headers={{
@@ -98,7 +101,6 @@ const Template: ComponentStory<typeof DataList> = args => {
         homePopulation: "Home world population",
         created: "Created",
       }}
-      loading={loadingInitialContent}
     >
       <DataList.Layout>
         {(item: DataListItemType<typeof mappedData>) => (
@@ -127,7 +129,10 @@ const Template: ComponentStory<typeof DataList> = args => {
 };
 
 export const Basic = Template.bind({});
-
+Basic.args = {
+  title: "All Clients",
+  loading: false,
+};
 export const EmptyState = Template.bind({});
 EmptyState.args = {
   data: [],
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -23,7 +23,7 @@ export default {
     viewMode: "story",
   },
   // Comment this out to make it show up in storybook
-  excludeStories: ["Basic", "EmptyState", "Breakpoints"],
+  // excludeStories: ["Basic", "EmptyState", "Breakpoints"],
   decorators: [
     // Detach from Storybook's layout
     (Story, { viewMode }) => {
@@ -213,7 +213,7 @@ const BreakpointTemplate: ComponentStory<typeof DataList> = args => {
           </Grid>
         )}
       </DataList.Layout>
-      <DataList.Layout size={"xs"} showHeader={true}>
+      <DataList.Layout size={"xs"} showHeader={false}>
         {(item: DataListItemType<typeof mappedData>) => (
           <Card>
             <Content>
--- a/packages/components/src/DataList/DataList.stories.tsx
+++ b/packages/components/src/DataList/DataList.stories.tsx
@@ -23,7 +23,7 @@ export default {
     viewMode: "story",
   },
   // Comment this out to make it show up in storybook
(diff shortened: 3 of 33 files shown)


packages/components/src/DataList/DataList.test.tsx
```diff
@@ -96,6 +113,11 @@ describe("DataList", () => {
     it("should render the specified layout", () => {
       expect(screen.getAllByTestId(layoutWrapper)).toHaveLength(3);
       expect(screen.getAllByTestId(layoutItem)).toHaveLength(6);
+      BREAKPOINTS.forEach(size => {
+        expect(
+          getCssVarForListItem(`--data-list-${size}-display`, layoutWrapper),
+        ).toEqual(new Array(3).fill("block"));
```
MichaelParadis, line 119: Probably going to change this to be the length of mock data. Same as below

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -144,3 +163,89 @@ export function generateDataListEmptyState({
...
+    const layoutSize = layout.props.size || "xs";
+    const showHeader = layout.props.showHeader ?? true;
+    const largerBreakpointsToHide = sizePropsOfLayouts?.filter(
+      size => BREAKPOINTS.indexOf(size) > BREAKPOINTS.indexOf(layoutSize),
+    );
+    const cssVars = getCSSVariablesForBreakpoints(
+      layoutSize,
+      largerBreakpointsToHide,
+    );
+    console.log("showHeader", showHeader);
```
MichaelParadis, line 212: Remove this console log

packages/components/src/DataList/DataList.test.tsx
MichaelParadis: Actually because it renders the layout in the header as well. It would be mockData.length + 1. Not the biggest fan of that

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -144,3 +164,88 @@ export function generateDataListEmptyState({
...
+          style={cssVars}
+        >
+          {layoutChildren(child)}
+        </div>
+      );
+    });
+  });
+}
+
+export function renderDataListHeader<T extends DataListObject>(
```
MichaelParadis, line 198: Not a big fan of having so much duplicated code. I think it may be worth looking at once we add sorting support to see what can be done.

MichaelParadis: Cleaned it up

packages/components/src/DataList/DataList.css
```diff
@@ -25,3 +25,25 @@
...
+  }
+
+  @media (--large-screens-and-up) {
+    display: var(--data-list-lg-display);
+  }
+
+  @media (--extra-large-screens-and-up) {
+    display: var(--data-list-xl-display);
+  }
+}
```
darryltec, line 49: I wouldn't use CSS to deal with this. 

You're rendering twice (or 5, if you hit all breakpoints) the amount of elements and functions on the page. CSS is hiding them and preventing it from being painted, but the functions and side effects that comes with those components still exists and are running.

packages/components/src/DataList/DataList.tsx
```diff
@@ -47,17 +50,8 @@ export function DataList<T extends DataListObject>({
...
-        {headerData && layout?.(headerData)}
-      </div>
-      {layout &&
-        elementData.map((child, i) => (
-          // TODO: Don't use index as key. Might have to force an ID on the data JOB-76773
-          <div className={styles.listItem} key={i}>
-            {layout(child)}
-          </div>
-        ))}
+      {headerData && dataListHeaderContent?.map(_layout => _layout)}
```
darryltec, line 53: There should only be 1 header, no?

packages/components/src/DataList/DataList.stories.tsx
```diff
@@ -142,3 +143,110 @@ export const EmptyState = Template.bind({});
...
+                <Grid.Cell size={{ xs: 6 }}>{item.home}</Grid.Cell>
+              </Grid>
+            </Grid.Cell>
+            <Grid.Cell size={{ xs: 2 }}>{item.homePopulation}</Grid.Cell>
+            <Grid.Cell size={{ xs: 3 }}>{item.tags}</Grid.Cell>
+            <Grid.Cell size={{ xs: 2 }}>{item.created}</Grid.Cell>
+          </Grid>
+        )}
+      </DataList.Layout>
+      <DataList.Layout size={"xs"} showHeader={true}>
```
darryltec, line 216: I wonder if `showHeader` is better as `withHeader` english and semantically speaking.

Also, is supposed to be `false`?

packages/components/src/DataList/DataList.tsx
MichaelParadis: Because it renders based on the layout it could change for different layouts instead of just the first layout

packages/components/src/DataList/DataList.stories.tsx
MichaelParadis: It should be `false`. I feel `showHeader` makes more sense but I am not super attached to it

packages/components/src/DataList/DataList.css
MichaelParadis: I initially was conditionally rendering this using a `useMediaQueries` hook. I can update it to use that

packages/components/src/DataList/DataList.utils.tsx
darryltec, line 1: should probs just delete the `getCompoundComponent` if we're not using it

MichaelParadis: It is used for the Empty State at the moment

packages/components/src/DataList/DataList.css
MichaelParadis: @darryltec I am doing something to update this, however this is already happening with GridCells so I am not sure if it something to address at the moment.

darryltec: > however this is already happening with GridCells so I am not sure if it something to address at the moment.

usage is a bit different no? for grid, you can't really hide the element

taylorvnoj:
I'm still seeing overlap in the 2nd and 3rd columns. The headers start to merge at the `423` mark and column content at `398`.  

Screenshot 2023-08-22 at 4 48 08 PM

packages/components/src/DataList/DataList.css
MichaelParadis: @darryltec I updated it.

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -137,3 +165,152 @@ export function generateDataListEmptyState({
...
+
+  const initialMatches = BREAKPOINTS.reduce(
+    (previous, breakpoint) => ({
+      ...previous,
+      [breakpoint]: mediaQueries[breakpoint].matches,
+    }),
+    {} as Record<Breakpoints, boolean>,
+  );
+
+  const [matches, setMatches] = useState(initialMatches); // one-time, instantaneous check  useEffect(() => {
```
MichaelParadis, line 290: remove that comment because it is wrong

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -137,3 +165,152 @@ export function generateDataListEmptyState({
...
+  );
+}
+
+export function sortSizeProp(sizeProp: Breakpoints[]) {
+  return sizeProp.sort(
+    (a, b) => BREAKPOINTS.indexOf(a) - BREAKPOINTS.indexOf(b),
+  );
+}
+
+export function useGridLayoutMediaQueries() {
```
MichaelParadis, line 273: Chose this approach because `useResizeOberserver` would cause the entire table to rerender a bunch as the window resizes as opposed to this approach which only rerenders when a mediaQuery matches

packages/components/src/DataList/DataList.tsx
MichaelParadis: This may be able to be changed since I took a different approach with handling layouts. It is a small change so look at the mediaquery hook approach and let me know

packages/components/src/DataList/DataList.utils.tsx
MichaelParadis: We could add something like this to our hooks package but it is far outside the scope of this subtask

taylorvnoj:
@MichaelParadis 

> that doesn't seem to be on the breakpoints story. It also may be something with Grid @darryltec may have more information.

I posed the same question to Darryl on his PR and directed it to you 😆  https://github.com/GetJobber/atlantis/pull/1495#issuecomment-1688621497

>  This is visible on master before these changes

Yes, because we hadn't looked at responsiveness yet. Is there a way to truncate at least the header of the second column.  Screenshot just to display that the last three column headings will truncate:
Screenshot 2023-08-23 at 10 31 56 AM

packages/components/src/DataList/DataList.stories.tsx
```diff
@@ -143,3 +144,105 @@ export const EmptyState = Template.bind({});
...
+            </Grid.Cell>
+            <Grid.Cell size={{ xs: 2 }}>{item.homePopulation}</Grid.Cell>
+            <Grid.Cell size={{ xs: 3 }}>{item.tags}</Grid.Cell>
+            <Grid.Cell size={{ xs: 2 }}>{item.created}</Grid.Cell>
+          </Grid>
+        )}
+      </DataList.Layout>
+      <DataList.Layout size={"xs"} showHeader={false}>
+        {(item: DataListItemType<typeof mappedData>) => (
+          <Card>
```
taylorvnoj, line 228: I like this story a lot - but I'm wondering if we should be putting them into cards.  The figma only has the divider on the mobile web views without an extra box around the line item.  Should we mimic that?

packages/components/src/DataList/DataList.tsx
```diff
@@ -24,14 +27,25 @@ export function DataList<T extends DataListObject>({
...
+  const dataListHeaderContent = renderDataListHeader(
+    allLayouts,
+    headerData,
+    mediaMatches,
+  );
+  const dataListContent = renderDataListItems(
+    allLayouts,
+    elementData,
+    mediaMatches,
+  );
```
darryltec, line 48: I wonder if these things should be encapsulated in its own component now rather than functions that returns an element.

Basically separate concerns.
- `<DataListHeader />` only deals with header bits
- `<DataListItems />` only deals with the items
- Both of them access' the media matches

packages/components/src/DataList/DataList.const.ts
```diff
@@ -1,2 +1,22 @@
...
+ */
+export const BREAKPOINTS = ["xs", "sm", "md", "lg", "xl"] as const;
+
+export const BREAKPOINT_SIZES: Record<Breakpoints, number> = {
+  xs: 0,
+  sm: 490,
+  md: 768,
+  lg: 1080,
+  xl: 1440,
+};
```
darryltec, line 22: doesn't have to be in this ticket but we should generate those breakpoints as a design token so we can reuse them in JS

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -137,3 +165,171 @@ export function generateDataListEmptyState({
...
+  elementData: DataListItemType<T[]>[],
+  mediaMatches?: Record<Breakpoints, boolean>,
+) {
+  return renderDataListLayouts(
+    layouts,
+    (layout: React.ReactElement<DataListLayoutProps<T>>) => {
+      return elementData.map((child, i) => {
+        // TODO: Don't use index as key. Might have to force an ID on the data JOB-76773
+        return (
+          <div className={styles.listItem} key={`${i}-${child.id}`}>
```
darryltec, line 184: hmmm. that key isn't gonna work since it is still tied to the index and therefore would rerender the moment that index changes when another item gets inserted.

darryltec: see PR description https://github.com/GetJobber/atlantis/pull/1496

packages/components/src/DataList/DataList.utils.tsx
darryltec: agreed

packages/components/src/DataList/components/DataListLayout/DataListLayout.tsx
```diff
@@ -1,8 +1,24 @@
...
+   * The breakpoint at which the layout should be displayed. It will be rendered until a layout with a larger breakpoint is found.
+   * @default "xs"
+   */
+  readonly size?: Breakpoints;
+
+  /**
+   * Determines if a header is shown for the layout
+   * @default true
+   */
+  readonly showHeader?: boolean;
```
darryltec, line 21: Thinking more about this. Might be better if it's handled through the `DataList` rather than the `Layout`. Like so `headerVisibility={{ xs: false, large: true }}` (prop name TBD).

Thinking about that scenario when someone only uses 1 layout and passes in a pure CSS responsive layout. They wouldn't be able to remove that header.

Happy to talk about this IRL

packages/components/src/DataList/DataList.const.ts
darryltec: oh, not talking about that weird hook. talking about the CSS variables that determines the breakpoints that we document https://atlantis.getjobber.com/?path=/docs/design-breakpoints--page

taylorvnoj:
Okay hid and marked my comment as Off Topic since it's Grid related 👍  Will leave the final approval to @darryltec

packages/components/src/DataList/DataList.tsx
```diff
@@ -23,14 +27,15 @@ export function DataList<T extends DataListObject>({
...
-  const layout = getCompoundComponent<DataListLayoutProps<T>>(
+  const allLayouts = getCompoundComponents<DataListLayoutProps<T>>(
     children,
     DataListLayout,
-  )?.props.children;
+  );
 
-  const elementData = generateListItemElements(data);
   const headerData = generateHeaderElements(headers);
+  const mediaMatches = useGridLayoutMediaQueries();
```
darryltec, line 38: do we save any performance by calling this hook here and passing it in as props on the 2 components? Rather than calling this hook on each components?

packages/components/src/DataList/DataList.types.ts
```diff
@@ -72,4 +73,13 @@ export interface DataListProps<T extends DataListObject> {
...
+
+  /**
+   * Determine if the header is visible at a given breakpoint. If one isn't provided,
+   * it will use the value from the next smallest breakpoint that has a value.
+   * @default { xs: true, sm: true, md: true, lg: true, xl: true }
+   */
+  readonly headerVisibility?: { [Breakpoint in Breakpoints]?: boolean };
 }
+
+export { Breakpoints } from "./DataList.const";
```
darryltec, line 85: gotta admit, did not expect this to be down here. Consistency-wise, should it be down here?

packages/components/src/DataList/DataList.utils.tsx
darryltec: Actually, I wonder if this should be in its own hook file rather than with the utils? Also, just curious, why is it called grid?

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeader.tsx
```diff
@@ -0,0 +1,51 @@
...
+  layouts,
+  mediaMatches,
+  headerData,
+  headerVisibility,
+}: {
+  layouts: React.ReactElement<DataListLayoutProps<T>>[] | undefined;
+  mediaMatches?: Record<Breakpoints, boolean>;
+  headerData?: DataListItemTypeFromHeader<T, DataListHeaderType<T>>;
+  headerVisibility: NonNullable<DataListProps<T>["headerVisibility"]>;
+}) {
```
darryltec, line 24: nit: mostly because we've been consisitent around putting these things on interfaces.

protip: highlight +  `cmd + .` > extract

image

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeader.tsx
```diff
@@ -0,0 +1,51 @@
...
+          (key): key is Breakpoints => !!mediaMatches?.[key as Breakpoints],
+        );
+        const sortedVisibleBreakpoints = sortSizeProp(matchingMediaQueries);
+        // Determines if the header should be visible based on the headerVisibility
+        const showHeader = sortedVisibleBreakpoints.reduce(
+          (previousVisibility, breakpoint) => {
+            return headerVisibility[breakpoint] ?? previousVisibility;
+          },
+          true,
+        );
```
darryltec, line 39: Does this need to be inside this function? It seems to not rely anything on the param from the render layout

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeader.tsx
```diff
@@ -0,0 +1,51 @@
...
+}: {
+  layouts: React.ReactElement<DataListLayoutProps<T>>[] | undefined;
+  mediaMatches?: Record<Breakpoints, boolean>;
+  headerData?: DataListItemTypeFromHeader<T, DataListHeaderType<T>>;
+  headerVisibility: NonNullable<DataListProps<T>["headerVisibility"]>;
+}) {
+  return (
+    <DataListLayoutInternal
+      layouts={layouts}
+      renderLayout={(layout: React.ReactElement<DataListLayoutProps<T>>) => {
```
darryltec, line 28: To save us some pain in the future, do you need to declare that type? it seems to be inferring it correctly.

packages/components/src/DataList/components/DataListLayoutInternal/DataListLayoutInternal.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { BREAKPOINTS, Breakpoints } from "../../DataList.const";
+import { DataListObject } from "../../DataList.types";
+import { sortSizeProp } from "../../DataList.utils";
+import { DataListLayoutProps } from "../DataListLayout/DataListLayout";
+
+/**
+ * Determine is layout is visible by checking a media query matches for the
+ * visible sizes of the layout and there isn't a larger layout that should be rendered
+ */
+function isLayoutVisible({
```
darryltec, line 11: gotta admit, did not expect to see this function first. For consistency's sake again, probs should drop this below the component that this file name is referencing.

packages/components/src/DataList/components/DataListLayoutInternal/index.ts
```diff
@@ -0,0 +1,2 @@
+export { DataListItems } from "./DataListItems";
+export { DataListHeader } from "./DataListHeader";
```
darryltec, line 2: don't really know how I feel about these 2 files being exported in a totally different named folder. It is an internal thing tho, so maybe that's okay.

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -137,3 +163,55 @@ export function generateDataListEmptyState({
...
+      const handler = (e: MediaQueryListEvent) =>
+        setMatches(previousMatches => ({
+          ...previousMatches,
+          [breakpoint]: e.matches,
+        }));
+      mediaQueries[breakpoint].addEventListener("change", handler);
+      return { ...previous, [breakpoint]: handler };
+    }, {} as Record<Breakpoints, (e: MediaQueryListEvent) => void>);
+    return () => {
+      BREAKPOINTS.map(breakpoint => {
```
darryltec, line 203: semantically, this is a `forEach`, no? we aren't really using what `map` is returning

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -137,3 +163,55 @@ export function generateDataListEmptyState({
...
+  const [matches, setMatches] = useState(initialMatches);
+  // Set up mediaQuery event handlers
+  useEffect(() => {
+    const handlers = BREAKPOINTS.reduce((previous, breakpoint) => {
+      const handler = (e: MediaQueryListEvent) =>
+        setMatches(previousMatches => ({
+          ...previousMatches,
+          [breakpoint]: e.matches,
+        }));
+      mediaQueries[breakpoint].addEventListener("change", handler);
```
darryltec, line 199: okay, this threw me off for a bit since it is more of a side effect than what reduce is meant to do.

should this be created outside of the effect and then foreach it to add the listeners like the return?

darryltec (changes requested):
This is close! Just some code structure things. Let me know what you think!

packages/components/src/DataList/DataList.utils.tsx
darryltec:
```ts
const handlers = reduce // <--- here

useEffect(() => {
  const handlers = reduce // <--- or here
  
  handers.forEach(...) // add event listeners

  return () => {
    handers.forEach(...) // add event listeners
  }
})
```

My point here is to separate concerns and let that function inside the reduce do 1 thing. Right now it is doing 2 things and it is an easily missable thing down the line when a bug happens

rodrigoeidelvein: I see. That makes sense, I'll apply it

packages/components/src/DataList/components/DataListLayoutInternal/index.ts
rodrigoeidelvein: As they're internal components from this component, I think it's okay. I extracted the props to an interface tho

packages/components/src/DataList/DataList.tsx
rodrigoeidelvein: Honestly I don't know. I can take some time to test it soon

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -32,6 +34,23 @@ export function getCompoundComponent<T>(
...
+  type: ReactElement<T>["type"],
+): ReactElement<T>[] {
+  const childrenArray = Children.toArray(children);
+  const elements = childrenArray.filter(
+    (child): child is ReactElement<T> =>
+      isValidElement<T>(child) && child.type === type,
+  );
+
+  // Comply with the return type without casting it
+  return elements || [];
```
darryltec, line 51: filter returns an empty array if it finds nothing so you actually don't need this

```suggestion
  return elements;
```

packages/components/src/DataList/components/DataListLayout/DataListLayout.tsx
```diff
@@ -1,8 +1,18 @@
 import React from "react";
-import { DataListItemType, DataListObject } from "../../DataList.types";
+import {
+  Breakpoints,
+  DataListItemType,
+  DataListObject,
+} from "../../DataList.types";
 
 export interface DataListLayoutProps<T extends DataListObject> {
   children: (item: DataListItemType<T[]>) => JSX.Element;
```
darryltec, line 9: not on you, but forgot to readonly this
```suggestion
  readonly children: (item: DataListItemType<T[]>) => JSX.Element;
```

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+  DataListProps,
+} from "../../DataList.types";
+import { sortSizeProp } from "../../DataList.utils";
+import { DataListLayoutProps } from "../DataListLayout/DataListLayout";
+
+interface DataListHeaderProps<T extends DataListObject> {
+  layouts: React.ReactElement<DataListLayoutProps<T>>[] | undefined;
+  mediaMatches?: Record<Breakpoints, boolean>;
+  headerData?: DataListItemTypeFromHeader<T, DataListHeaderType<T>>;
+  headerVisibility: NonNullable<DataListProps<T>["headerVisibility"]>;
```
darryltec, line 18: nit: consisitency. same goes for the other ones

```suggestion
  readonly layouts: React.ReactElement<DataListLayoutProps<T>>[] | undefined;
  readonly mediaMatches?: Record<Breakpoints, boolean>;
  readonly headerData?: DataListItemTypeFromHeader<T, DataListHeaderType<T>>;
  readonly headerVisibility: NonNullable<DataListProps<T>["headerVisibility"]>;
```

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+    },
+    true,
+  );
+
+  return (
+    <DataListLayoutInternal
+      layouts={layouts}
+      renderLayout={layout => {
+        return (
+          (showHeader && headerData && (
```
darryltec, line 45: This is why I like keeping the code organized, it makes it easy to spot something like this.

This check and how it returns a fragment if one of these disappears. should this be caught earlier? 

Don't render the `DataListLayoutInternal` if we don't want it to show (`showHeader`) and there's nothing to show (`headerData`) 🤔

darryltec (approved):
saw some pretty minor stuff but that could be a fast follow PR's

darryltec (changes requested):
missed a spot

packages/components/src/DataList/DataList.tsx
MichaelParadis: The performance impact would be minimal and I assume the code running once and a stack allocation for a reference to an object would be faster than running code twice

packages/components/src/DataList/hooks/useLayoutMediaQueries.ts
```diff
@@ -0,0 +1,53 @@
+import { useEffect, useState } from "react";
+import { BREAKPOINTS, BREAKPOINT_SIZES, Breakpoints } from "../DataList.const";
+
+export function useLayoutMediaQueries() {
```
MichaelParadis, line 4: @darryltec the name grid was stuck in my head for quite a while and it slipped in

packages/components/src/DataList/DataList.tsx
darryltec: Tru! My only concern was the prop drilling part where we have this reusable hook that can be used by the components instead of passing it in as a prop. If saving a bit of a performance outweighs the impact of prop drilling (which is a totally valid thing to do) when we don't need to then I am good with this.

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeader.tsx
rodrigoeidelvein: I added an early return for these conditionals.

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -32,6 +34,23 @@ export function getCompoundComponent<T>(
...
+  children: ReactElement | ReactElement[],
+  type: ReactElement<T>["type"],
+): ReactElement<T>[] {
+  const childrenArray = Children.toArray(children);
+  const elements = childrenArray.filter(
+    (child): child is ReactElement<T> =>
+      isValidElement<T>(child) && child.type === type,
+  );
+
+  // Comply with the return type without casting it
```
darryltec, line 50:
```suggestion
```

packages/components/src/DataList/components/DataListLayoutInternal/DataListLayoutInternal.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { DataListObject } from "../../DataList.types";
+import { sortSizeProp } from "../../DataList.utils";
+import { DataListLayoutProps } from "../DataListLayout/DataListLayout";
+
+interface DataListLayoutInternalProps<T extends DataListObject> {
+  layouts: React.ReactElement<DataListLayoutProps<T>>[] | undefined;
+  renderLayout: (
+    layout: React.ReactElement<DataListLayoutProps<T>>,
+  ) => JSX.Element;
+  mediaMatches?: Record<Breakpoints, boolean>;
```
darryltec, line 12: y'kow what it is!

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Breakpoints } from "../../DataList.const";
+import styles from "../../DataList.css";
+import { DataListObject } from "../../DataList.types";
+import { generateListItemElements } from "../../DataList.utils";
+import { DataListLayoutProps } from "../DataListLayout/DataListLayout";
+
+interface DataListItemsProps<T extends DataListObject> {
+  layouts: React.ReactElement<DataListLayoutProps<T>>[] | undefined;
+  mediaMatches?: Record<Breakpoints, boolean>;
+  data: T[];
```
darryltec, line 12: 😏

darryltec (approved):
can ship this and follow up so your task can be unblocked

PR #1499, opened 2023-08-24
feat(components): Add DataTable loader skeleton [JOB-76695]

zevpermack:
## Motivations

- There have been complicated ways of introducing loading states to the DataTable component externally, with different designs implemented
- As a company it would be nice to have consistent design for the loading state of the DataTable and as a user be able to simply be able to pass in whether the component is in a loading state or not.

## Changes

- The amount of loaders shown is based off the `itemsPerPage` property in the pagination prop object. The number of loader rows will match the currently selected value.
- If there is no pagination prop passed, the value will default to 10 as per [tan stack documentation](https://tanstack.com/table/v8/docs/api/features/pagination#:~:text=Resets%20the%20page%20size%20to%20its%20initial%20state.%20If%20defaultState%20is%20true%2C%20the%20page%20size%20will%20be%20reset%20to%2010%20regardless%20of%20initial%20state.)

### Added

- Loading state to DataTable component

### Changed

- The loading prop is optional and defaults to false, so there are no breaking changes to previous uses.

## Testing

- Check that loading prop displays the design as intended and that displaying data is otherwise unaffected. 
- There are tests for the default loading state and when pagination prop is provided with non-default values

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataTable/Body.tsx
+++ b/packages/components/src/DataTable/Body.tsx
@@ -2,18 +2,21 @@ import { Row, Table, flexRender } from "@tanstack/react-table";
 import classNames from "classnames";
 import React, { ReactNode, useCallback } from "react";
 import styles from "./DataTable.css";
+import { Glimmer } from "../Glimmer";
 
 interface BodyProps<T> {
   table: Table<T>;
   onRowClick?: (row: Row<T>) => void;
   height?: number;
   emptyState?: ReactNode | ReactNode[];
+  loading?: boolean;
 }
 
 export function Body<T extends object>({
   table,
   onRowClick,
   emptyState,
+  loading,
 }: BodyProps<T>) {
   const bodyRowClasses = classNames({ [styles.clickableRow]: !!onRowClick });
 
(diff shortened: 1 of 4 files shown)


packages/components/src/DataTable/Body.tsx
```diff
@@ -25,34 +28,57 @@ export function Body<T extends object>({
...
-            key={row.id}
-            onClick={handleRowClick(row)}
-            className={bodyRowClasses}
-          >
-            {row.getVisibleCells().map(cell => {
+  const loaderRows = table.getState().pagination.pageSize;
+  const loaderColumns = table.getAllColumns().length;
+  return (
+    <>
+      {!loading ? (
```
rodrigoeidelvein, line 35: I think it would be good to do an early return instead of adding another ternary to this `return`

It can be as simple as 
```
  if (loading) {
    return (
      <tbody>
        {Array.from(Array(loaderRows).keys()).map(row => (
          <tr key={row}>
            {Array.from(Array(loaderColumns).keys()).map(arr => (
              <td key={arr}>
                <Glimmer />
              </td>
            ))}
          </tr>
        ))}
      </tbody>
    );
  }
```

It you want to go further, this could be extracted to its own component, something like `DataTableLoadingState`, but it's up to you, I see that in this file there's a pattern of doing ternaries and this would be an improvement, but not necessary for now.

rodrigoeidelvein (changes requested):
Hey Zev, thanks so much for adding the loading state to DataTable, it will definitely be a huge improvement in this component!

just some thoughts in my review

packages/components/src/DataTable/Body.tsx
```diff
@@ -2,18 +2,21 @@ import { Row, Table, flexRender } from "@tanstack/react-table";
 import classNames from "classnames";
 import React, { ReactNode, useCallback } from "react";
 import styles from "./DataTable.css";
+import { Glimmer } from "../Glimmer";
 
 interface BodyProps<T> {
   table: Table<T>;
   onRowClick?: (row: Row<T>) => void;
   height?: number;
```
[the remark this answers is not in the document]
zevpermack: True, removed 👍

packages/components/src/DataTable/Body.tsx
zevpermack: Good point, I agree! Changed the pattern let me know what you think.

packages/components/src/DataTable/BodyLoading.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import { Table } from "@tanstack/react-table";
+import { Glimmer } from "../Glimmer";
+
+interface BodyLoadingProps<T> {
+  table: Table<T>;
+}
+
+export function BodyLoading<T extends object>({ table }: BodyLoadingProps<T>) {
+  const loaderRows = table.getState().pagination.pageSize;
```
zevpermack, line 10: @rodrigoeidelvein Made the changes asked for. One more thing that came up is whether we should consider putting a cap on the maximum amount of glimmers that get shown. For example if someone decided their minimum items per page was 1000, should we cap the amount of loaders that we show at 100? 150? 200? Or do you think this is not relevant?

rodrigoeidelvein: I don't think we need it. If someone is putting 1000 items per page, they're aware of the performance problems they'll encounter 😆

rodrigoeidelvein (approved):
LGTM! Thanks for adding this feature and doing a campground cleanup!

PR #1500, opened 2023-08-24
docs(components-native): Add InputSearch documentation

taylorvnoj:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/components/InputSearch/InputSearch.stories.mdx
+++ b/docs/components/InputSearch/InputSearch.stories.mdx
@@ -1,0 +1,40 @@
+import { Meta } from "@storybook/addon-docs";
+import { Figma } from "storybook-addon-designs/blocks";
+
+<Meta title="Components/Forms and Inputs/InputSearch/Docs" />
+
+# InputSearch
+
+The InputSearch component allows the user to input a keyword to search from
+associated data. It handles debouncing for improved performance when making
+queries.
+
+## Design & usage guidelines
+
+InputSearch will provide a consistent experience for users searching for content
+within the Jobber app. It uses familiar iconography and interactions to make
+searching a simpler experience.
+
+## Content guidelines
+
+The label for InputSearch should help guide the user to find their results. It
+can be helpful to direct the user to use certain types of queries for a given
+search, such as "Search by clients and properties".
+
+It is recommended to use the `search` icon as a prefix to aid users in
+recognizing the functionality of the input.
+
+InputSearch will take up the full available width of its' container.
+
+## Accessibility
+
+Most of the accessibility concerns for InputSearch are handled by `InputText`.
+It is announced as an input and the label and value are read to assistive
+technology. Users can navigate to the clear button to remove the entered value.
+
+## Mockup
+
+<Figma
+  collapsable
+  url="https://www.figma.com/file/avvgu5SkbBvS8lGVePBsqO/Product%2FMobile?node-id=25435%3A166491"
+/>
--- a/packages/components-native/src/InputSearch/InputSearch.tsx
+++ b/packages/components-native/src/InputSearch/InputSearch.tsx
@@ -49,11 +49,11 @@ function SearchInputInternal(
   }: InputSearchProps,
   ref: Ref<InputTextRef>,
 ) {
-  const delayedSearch = debounce(onDebouncedChange, wait);
+  const delayedSearch = debounce(() => onDebouncedChange(value), wait);
   const handleChange = (newValue = "") => onChange(newValue);
 
   useEffect(() => {
-    delayedSearch(value);
+    delayedSearch();
     return delayedSearch.cancel;
   }, [value, delayedSearch]);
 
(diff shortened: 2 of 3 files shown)


docs/components/InputSearch/InputSearch.stories.mdx
```diff
@@ -0,0 +1,40 @@
...
+## Content guidelines
+
+The label for InputSearch should help guide the user to find their results. It
+can be helpful to direct the user to use certain types of queries for a given
+search, such as "Search by clients and properties".
+
+It is recommended to use the `search` icon as a prefix to aid users in
+recognizing the functionality of the input.
+
+InputSearch will take up the full available width of its' container.
```
ZakaryH, line 27:
```suggestion
InputSearch will take up the full available width of its container.
```

packages/components-native/src/InputSearch/InputSearch.tsx
```diff
@@ -49,11 +49,11 @@ function SearchInputInternal(
   }: InputSearchProps,
   ref: Ref<InputTextRef>,
 ) {
-  const delayedSearch = debounce(onDebouncedChange, wait);
+  const delayedSearch = debounce(() => onDebouncedChange(value), wait);
```
ZakaryH, line 52: I'm trying to better understand this change. looking at the `lodash` versions we have:

`"lodash": "^4.17.21",`
VS
`"lodash": "^4.17.20",`

where there's nothing that _should_ have changed around this

https://github.com/lodash/lodash/wiki/Changelog
image

when I go back to how this was with `delayedSearch(value)` and no wrapping arrow function around `onDebouncedChange` my linter is happy with that, which is definitely what I'd expect since `onDebouncedChange` is already function

>while building the mobile storybook example I discovered that I needed to make the first argument to debounce a function.

what do you mean by that? like the invocation in `Mobile.stories.tsx` was complaining about props, or that was fine but the example wasn't debouncing the provided `onDebouncedChange`?

actually on that note, unless I'm missing something, I don't see us passing all of the required props to our instance of `<InputSearch>`. `onChange` and `onDebouncedChange` are supposed to be mandatory.

ZakaryH: to be clear when I try out these changes in an actual RN context, they do seem work the same as before - I'd just like to know if we can get away with not changing them at all!

taylorvnoj: @ZakaryH  I agree! It seems very silly considering the documentation and changelog have nothing.

The only reason I noticed this was the storybook example in the browser.
The original code on the `"lodash": "^4.17.21"` version gives me this

Screenshot 2023-08-25 at 4 32 55 PM

taylorvnoj: command clicking `debounce` leads to the same result in the `function.d.ts` yet in jobber-mobile it's fine and the code is the same

taylorvnoj: By 'result' I mean that it shows the first argument to `debounce` is a function

ZakaryH:
ok yeah I believe the problem is with the Basic example arguments

`onChange` and `onDebouncedChange` are mandatory but are not being passed in for this example. of course `onDebouncedChange` is the function that will be passed to the `_.debounce` call

the reason TS isn't getting upset about that is because we're tricking it into thinking `args` has everything it needs. you can see here the type of `args` is a combination of `InputSearchProps` and some ref stuff

image

so it's thinking sure, you're promising me you'll give me an `onChange` and `onDebouncedChange` since those are required in `InputSearchProps` - I am ok with this

but then when we go to build those `args` we do

```
export const Basic = BasicTemplate.bind({});
Basic.args = {
  placeholder: "Search",
  prefix: {
    icon: "search",
  },
};
```

notice that it doesn't have those functions, so when this compiles and tries to run - `debounce` is given `props.onDebouncedChange` which will be `undefined` and it correctly throws this error

I'm not sure about this `bind` pattern honestly. I think it makes it very easy to get around TypeScript and run into errors like this, and it's not obvious to me what benefit it provides. that's very much another discussion though, we don't need to worry about that here

to solve this all we need to do is
- provide the story with the right args
- revert the changes to the actual component

once I did that, it happily renders! (I probably wouldn't actually use these functions though, they make it hard to see what the component is doing)

image

ZakaryH (changes requested):
figured out the issue with the `debounce`

let's fix that up. aside from that, this is lookin good!

ZakaryH:
one other thing I want to make sure to mention is the fact that in our example invocation we must add an `onChange` that actually changes the `value` that we pass back into the components otherwise you'll never see the `onDebouncedChange` fire because that's in a `useEffect` that only watches `value` and `delayedSearch` where `delayedSearch` really shouldn't be changing

this makes me have thoughts & questions about this components design overall, but again that's not something we need to be worrying about with this PR!

ZakaryH (approved):
LGTM!

PR #1503, opened 2023-08-28
feat(components): Support semantic HTML elements for Content root element

mattander:
## Motivations

The spacing support provided by `Content` is great, but using the component currently means sacrificing potential semantic HTML regions. This feature would allow you to still get all the benefits of `Content` but with a more semantic element where applicable. 

## Changes
`Content` now accepts a `type` prop to use a semantic HTML5 element as the root element. Still defaults to `<div>`

## Testing

Try providing a `type` prop to a `Content`. The type must be a valid HTML5 content container. Try a `<section>`.

Added a unit test to verify the section renders as expected. 
---

❓ 😕 Random Picture of Atlanta
image

### DIFF
--- a/docs/components/Content/Web.stories.tsx
+++ b/docs/components/Content/Web.stories.tsx
@@ -21,7 +21,7 @@ export default {
 const BasicTemplate: ComponentStory<typeof Content> = args => (
   <Content {...args}>
     <Card title="About me">
-      <Content tagName="section" {...args}>
+      <Content type="section" {...args}>
         <Heading level={2}>Sign up!</Heading>
         <Text>
           Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis
--- a/docs/components/Content/Web.stories.tsx
+++ b/docs/components/Content/Web.stories.tsx
@@ -21,7 +21,7 @@ export default {
 const BasicTemplate: ComponentStory<typeof Content> = args => (
   <Content {...args}>
     <Card title="About me">
-      <Content {...args}>
+      <Content tagName="section" {...args}>
         <Heading level={2}>Sign up!</Heading>
         <Text>
           Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis
--- a/packages/components/src/Content/Content.test.tsx
+++ b/packages/components/src/Content/Content.test.tsx
@@ -45,8 +45,8 @@ it("renders a Content with a small spacing", () => {
   `);
 });
 
-it("renders with a semantic tag when a valid tagName is set", () => {
-  const { container } = render(<Content tagName="section">A section!</Content>);
+it("renders with a semantic tag when a valid type is set", () => {
+  const { container } = render(<Content type="section">A section!</Content>);
 
   expect(container).toMatchInlineSnapshot(`
     <div>
--- a/packages/components/src/Content/Content.test.tsx
+++ b/packages/components/src/Content/Content.test.tsx
@@ -44,3 +44,17 @@ it("renders a Content with a small spacing", () => {
     </div>
   `);
 });
+
+it("renders with a semantic tag when a valid tagName is set", () => {
+  const { container } = render(<Content tagName="section">A section!</Content>);
+
+  expect(container).toMatchInlineSnapshot(`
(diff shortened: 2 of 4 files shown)


docs/components/Content/Web.stories.tsx
```diff
@@ -21,7 +21,7 @@ export default {
 const BasicTemplate: ComponentStory<typeof Content> = args => (
   <Content {...args}>
     <Card title="About me">
-      <Content {...args}>
+      <Content type="section" {...args}>
```
darryltec, line 24: I'd drop that type down to line 47 so the storybook controls shows that value as well

packages/components/src/Content/Content.test.tsx
```diff
@@ -44,3 +44,17 @@ it("renders a Content with a small spacing", () => {
...
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      <section
+        class="padded base"
+      >
+        A section!
+      </section>
+    </div>
+    `);
+});
```
darryltec, line 60: I know we still use snapshots on older test but i'd prefer if we don't add more of them (or help us clean them up). We don't like snapshots as much here.

packages/components/src/Content/Content.tsx
```diff
@@ -11,10 +11,29 @@ interface ContentProps {
    * @default base
    */
   readonly spacing?: keyof typeof spacings;
+
+  /**
+   * The html tag name to use as the container element
+   * Defaults to 'div'
+   * Must be a valid HTML5 element that accepts children
```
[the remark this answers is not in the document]
mattander: Fixed.

docs/components/Content/Web.stories.tsx
mattander: Ahh thanks. Haven't done much with Storybook yet.

packages/components/src/Content/Content.test.tsx
mattander: Ah good to know. Just going off what was there.

docs/components/Content/Web.stories.tsx
mattander: Done.

packages/components/src/Content/Content.test.tsx
mattander: I cleaned the rest up too.

docs/components/Content/Web.stories.tsx
```diff
@@ -45,4 +45,5 @@ const BasicTemplate: ComponentStory<typeof Content> = args => (
 export const Basic = BasicTemplate.bind({});
 Basic.args = {
   spacing: "small",
+  type: "section",
```
darryltec, line 48: hmmmmmm, now that I think about it. We shouldn't set our default example to a section since we can assume that the code for that (under the code tab) is what's gonna be copy-pasted everywhere.

If we really want to showcase that, we should add another example. IMO, it showing up under controls is enough to tell people that they can do this.

image

```suggestion
```

packages/components/src/Content/Content.test.tsx
darryltec, line 1: sweeeet!

packages/components/src/Content/Content.tsx
```diff
@@ -11,10 +11,27 @@ interface ContentProps {
    * @default base
    */
   readonly spacing?: keyof typeof spacings;
+
+  /**
+   * The html tag name to use as the container element
+   * @default 'div'
+   */
```
darryltec, line 18:
```suggestion
  /**
   * Change the wrapping element to be one of the available
   * semantic tags.
   * 
   * @default 'div'
   */
```

darryltec (changes requested):
Two super minor things and this is 👌🏼

docs/components/Content/Web.stories.tsx
mattander: Yeah I was honestly wondering the same thing. I don't know that this really needs to be demo'd given that it's not a visible change.

darryltec (approved):
Ayt, we good! Thanks for contributing!! Ship it

PR #1505, opened 2023-08-29
fix: update autolink built in messages to be translateable

darryltec:
## Motivations

We need to allow translation on AutoLink's built-in messages. Our current implementation of it, while simple and clever, was too clever for our taste since it relies on a 3rd party package returning the correct value. 

This aims to expand on the code so it's easier to understand at every dev level.

### Added

- Needed strings for autolink

### Changed

- use I18n instead of react-intl

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


MichaelParadis (approved):
LGTM!

PR #1509, opened 2023-08-30
feat: Combobox Feature Branch

taylorvnoj:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


ZakaryH:
to whomst'ever is reviewing this, please note that if you have feedback around
- ComboboxContent being massive
- ComboboxContent having hooks in there
- the hooks having less ideal grouping of logic
- being integration test heavy

these points, and more, have been addressed in https://github.com/GetJobber/atlantis/pull/1550

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -0,0 +1,369 @@
...
+            <Text variation="subdued">No results for {`“${searchValue}”`}</Text>
+          </div>
+        )}
+
+        {!optionsExist && (
+          <div className={styles.emptyStateMessage}>
+            <Text variation="subdued">
+              {getZeroIndexStateText(props.subjectNoun)}
+            </Text>
+          </div>
```
ZakaryH, line 134: not sure about having these two divs inside the `ul` as direct children. I do not believe that is a best practice

>[Content model](https://html.spec.whatwg.org/multipage/dom.html#concept-element-content-model):
Zero or more [li](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element) and [script-supporting](https://html.spec.whatwg.org/multipage/dom.html#script-supporting-elements-2) elements.
https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element

taylorvnoj: Update here: We discussed adjusting this in Zak's current PR for `ComboboxContent` refactor

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useAssert } from "@jobber/hooks/useAssert";
+import { ComboboxContent } from "./components/ComboboxContent";
+import { ComboboxAction } from "./components/ComboboxAction";
+import { ComboboxContextProvider } from "./ComboboxProvider";
+import {
+  ComboboxTriggerButton,
+  ComboboxTriggerChip,
+} from "./components/ComboboxTrigger";
+
+export interface ComboboxProps {
```
scotttjob, line 11: Naive, new person question: We have props here, but a types file as well in this directory. Is there a reason we use both patterns? I'm happy to see a type file, but I would want to see all the types in there (including props) just to make sure we're separating concerns (types in the type file, components in the tsx files). If the pattern though is props with the component, every other type in the type file, that's okay too! :)

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,80 @@
...
+  readonly children: ReactElement | ReactElement[];
+}
+
+export const COMBOBOX_TRIGGER_COUNT_ERROR_MESSAGE =
```
scotttjob, line 15: Naive new person question: Do constants normally live with the root component, or is there ever a constants.ts file to house them?

packages/components/src/Combobox/ComboboxProvider.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { RefObject, useRef, useState } from "react";
```
scotttjob, line 1: Very much a fan of this pattern to avoid prop drilling, great work.

Naive, philosophical new person question: Do we provide any way of consumers to override this internal state? I can definitely see a future where consumers of the combobox component will want to open/close it some other way than a trigger button, or say, manually set the search field content because the user clicked on a specific button within a UI to represent a common search term. I'm spitballing on the use cases of overriding internal state (there's a hundred others), but I'm definitely a fan of allowing consumers to override (it increases extensibility without much effort on our behalf). That said, if we've made the collective decision to not allow override for specific reasons, I'm always open to having my mind changed especially for our use cases.

packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
```diff
@@ -0,0 +1,32 @@
+.actionContainer {
```
scotttjob, line 1: This CSS makes me happy. Great stuff.

packages/components/src/Combobox/hooks/useComboboxAccessibility.ts
```diff
@@ -0,0 +1,112 @@
+import { useEffect, useRef } from "react";
+import { useRefocusOnActivator } from "@jobber/hooks/useRefocusOnActivator";
+import { useFocusTrap } from "@jobber/hooks/useFocusTrap";
+import { usePopper } from "react-popper";
```
scotttjob, line 4: We should reconsider our usage of `react-popper` moving forward. It's not a library I'm familiar with and when I looked into it, I found it's based on @popperjs/core, but that team has recently deprecated that library in favour of the new floating-ui, which has their own react-bindings if we chose to move over. `react-popper` itself hasn't seen an update in over a year, so I think, it's entered a soft deprecated state of its own where the maintainers realize there's no future for it.

That said, if you dig into the core of `floating-ui` it's only a handful of functions that I understand pretty well, so if we wanted, we could roll our own solution that mimics the react-popper API so we don't really have to migrate, just roll our own solution and start using that instead.

packages/components/src/Combobox/components/ComboboxContent/ComboboxList.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { SetStateAction } from "react";
```
scotttjob, line 1: There are folders for ComboboxAction, ComboboxContent, ComboboxTrigger. There are sub-folders in ComboboxTrigger for ComboboxTriggerButton and ComboboxTriggerChip. Big fan of the organization.

My naive new person question though: in ComboboxContent, we have ComboboxList and ComboboxSearch. Should those actually be in their own top-level folders? Or should they be in sub-folders under ComboboxContent but prefixed like ComboboxTrigger? Or is there another pattern at play that I'm completely missing (that's entirely possible and most likely).

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const {
+    searchValue,
+    setSearchValue,
+    setSelectedElement,
+    filteredOptions,
+    optionsListRef,
+  } = useComboboxContent(props.options, open);
+
+  const { popperRef, popperStyles, attributes } = useComboboxAccessibility(
```
scotttjob, line 59: Naive, new person style question: In the top-level Combobox file we inlined the hooks there with the file, but here we're moving them to the hooks directory. I'm personally a fan of moving things into their own files like we're doing here, but is there a reason we're not doing that in the top-level file? Maybe there's a good reason I'm missing.

packages/components/src/Combobox/Combobox.css.d.ts
```diff
@@ -0,0 +1,5 @@
+declare const styles: {
```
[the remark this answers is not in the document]
taylorvnoj: They are! One time during this PR the `css.d.ts` file didn't autogenerate.

The situation was: In Content folder we have Content.tsx and a Content.css - why isn't the css.d.ts file autogenerating?

After talking with Darryl, he laid out a tldr process for these to generate on their own:
- create the css file
- import the styles file even though nothing exist inside
- make sure you are using the component (in my example, that would be the `Combobox.Content`
and if you have storybook running, storybook webpack will create it after save.

So the reason why we didn't see the `css.d.ts` file at first was because we weren't actually using `Combobox.Content` anywhere.  9 times out of 10 you won't run into this. Hope this gives more context rather than confusion!

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,274 @@
+import { cleanup, fireEvent, render } from "@testing-library/react";
+import React from "react";
+import {
+  COMBOBOX_REQUIRED_CHILDREN_ERROR_MESSAGE,
+  COMBOBOX_TRIGGER_COUNT_ERROR_MESSAGE,
+  Combobox,
+} from "./Combobox";
+import { ComboboxOption } from "./Combobox.types";
+
+afterEach(cleanup);
```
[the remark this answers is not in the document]
taylorvnoj: Good question! I would say from personal experience that Copilot helps a LOT in test writing.  Of course we don't want to rely on Copilot - there could be a false positive, etc but I find Copilot is most helpful for writing tests (for me personally).

packages/components/src/Combobox/Combobox.tsx
taylorvnoj: I have a couple guesses but not an informed answer - I think this would be a good team discussion to bring up at Office Hours or after stand up so that we're all informed 😃

packages/components/src/Combobox/Combobox.tsx
taylorvnoj: This is another good topic to bring up with the whole team because it seems to be a changing pattern...or it depends.

The only `constants.ts` files we have are related to `components-native` components.
I just noticed that `DataList` has a `DataList.const.ts` file that houses a lot of constants.
Older components seem to have the constants live within the root component.

taylorvnoj:
@dlrussel @ZakaryH  - some really great points brought up here by @scotttjob 

I've answered a few but please add where I missed anything.  Also, there are some other comments that I didn't address - I know either of you would have more informed answers.

packages/components/src/Combobox/Combobox.test.tsx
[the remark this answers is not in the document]
ZakaryH: I was literally just gushing about this to someone the other day! most of these I type out the `it` assertion, and even then Copilot sometimes offers solid test statements, from there Copilot fills out the entire body of the test

sometimes the set up or way it tests the thing needs a minor adjustment, but for simple things that don't involve any specific implementation knowledge it is so so soso sososossosoos nice to just run the generated test code, maybe make sure it fails when it should and move onto the next one

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,80 @@
...
+  );
+  return (
+    <ComboboxContextProvider>
+      {triggerElement}
+      {contentElement}
+    </ComboboxContextProvider>
+  );
+};
+
+function useComboboxValidation(children: ReactNode): {
```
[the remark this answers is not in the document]
ZakaryH: good questions!

as I recall, the way we got here was more or less:

"it might be nice if combobox didn't care about the order you gave the subcomponents"

eg.

```
<Combobox>
  <Combobox.Trigger/>
  <Combobox.Content/>
</Combobox>
```
is just as valid as

```
<Combobox>
  <Combobox.Content/>
  <Combobox.Trigger/>
</Combobox>
```

to achieve that, we needed to identify which child was which. now that we know which child is which and how many we got, we thought why not add a bit of validation to improve the developer experience? 

there was also a 3rd thing we tried to get out of this setup that ended up not panning out. the goal of it was to avoid the use of a `Context` but yeah we didn't do it in the end

but you're right, this is more code that _could_ have bugs. is it worth it? debatable. I agree that if Atlantis makes (or has already made) a philosophical choice that says "if you do something silly, you're on your own" that can inform these implementation decisions, and it simplifies many of these discussions around how defensively we need to code/how much we need to protect people from themselves

maybe it's perfectly fine for our code to totally break if you give us two `Triggers` since that's something we've explicitly said we don't support!

packages/components/src/Combobox/Combobox.tsx
ZakaryH: to be perfectly honest I think it might have been partly out of fear

at one point we tried to be tidy little typers with our types all defined in once place and we got ourselves into a fun little dependency cycle hell 🙃 

though, that was with a bit of a different setup... now with all the types in a single file, that _shouldn't_ be the case, and it's probably a bit of cleanup we should have done because I agree it would be much nicer to have all the exported types/interfaces colocated

if an interface is only used locally and would never ever need to be used outside, then I'm a little less certain it should be in the types file but I could easily be convinced otherwise on that

packages/components/src/Combobox/Combobox.tsx
ZakaryH: exactly what Taylor said

in mobile, where I've been living for the past year ish, we do `constants.ts` but I didn't see many/any in Atlantis. despite the fact that I did the opposite here, I do like separating them into their own file especially when they are user facing messages for translations

packages/components/src/Combobox/ComboboxProvider.tsx
ZakaryH: hmm those are interesting scenarios. my question with those would be: is that a combobox we're talking about or maybe something else at that point?

I'm not totally opposed to opening things up to consumers, but we've gotta have some limits. I wouldn't want us to enable too much customization to the point where the behavior is quite different depending on how much of the default behavior has been overridden

packages/components/src/Combobox/hooks/useComboboxAccessibility.ts
ZakaryH: good to know! yeah we've been using `react-popper` for a while now, I'm sure when it was introduced it was much more actively maintained. sounds like a bit of tech debt for us to replace all those instances with the newer library and hope it doesn't have the same problem in a year or two

packages/components/src/Combobox/components/ComboboxContent/ComboboxList.tsx
ZakaryH: it's like you read my mind!

https://github.com/GetJobber/atlantis/pull/1550#issuecomment-1739910249

still looking for feedback on this topic

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: nope, no reason. just an oversight that I missed while moving those other two into their own files. the top level hook should also have been moved into there. good catch!

PR #1510, opened 2023-08-31
feat(components): Combobox Folder Reorg & Validation

ZakaryH:
## Motivations

We need one main thing right now, to have separated files to unblock parallelization. That's the point of the folder re-org.

While doing that, it didn't make sense to move files we didn't want anymore so we tried to swap the Context/Provider pattern for a local `useState` inside `Combobox`. That ended up not working out and we went back to `Context` to manage `open` and `setOpen` for the child `Trigger` to use.

We had already done the work to setup the structure for the local state, and it was pretty simple to add the validation for this level. actually this is the only validation required at all because it is perfectly fine to pass an empty array of `options` to `Combobox.Content` and `Combobox.Action` is also optional, so I don't believe anything will be necessary in there beyond basic prop validation that we will get from TypeScript.

With that, I'm fine with removing the validation in this PR if we feel it is too much. I can add that separately. But if we're ok with it, it's all working!

## Changes

- Folder structure

### Added

- Validation for development environment specifically

## Testing

The storybook example should still be functional to the same degree it is on the parent feature branch. It will look goofy but that's known.

Try invoking the compound component, in your local dev environment either in the story or some arbitrary other `.tsx` file, without the necessary elements.

**Requirements**
- Exactly one Trigger, whether that's a `TriggerChip` or a `TriggerButton`
- Exactly one Content

Anything other than that combination should throw an error in development. I'm not too sure how to do this in production easily....

One could make a local packaged version of Atlantis with this branch, pull that into Jobber Online and use it there, and maybe hack your node env to report as `production`. Then, come up with some code that dynamically populates the `Combobox` in such a way that it will result in missing or erroneous elements. This should not throw an error.

---

Random photo of Atlantis


packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -1,7 +1,127 @@
...
+            <></>
+          </Combobox.Content>
+        </Combobox>,
+      );
+    }).not.toThrow();
+  });
+  it("throws an error if there is no Trigger element", () => {
+    expect.assertions(1);
+    let error;
+    try {
```
ZakaryH, line 23: looking for feedback on this approach. it works, but I'm not sure I love it

initially I tried

```
  it("throws an error if there is no Trigger element", () => {
    expect.assertions(1);
    let error;
    const expectation = render(
      <Combobox>
        <Combobox.Content options={[]} onSelection={jest.fn()}>
          <></>
        </Combobox.Content>
      </Combobox>,
    );
    expect(expectation).toThrow(
      "Combobox must have a Trigger and Content element",
    );
  });
```

which I like significantly more. the problem is that the thrown error stops execution and we never make it to the `expect`. it complains about an ErrorBoundary which is fair, but I'd prefer not to build and implement that here since that does not appear to exist currently

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,191 +1,71 @@
...
-   */
+import React, { ReactNode } from "react";
+import { useAssert } from "@jobber/hooks/useAssert";
+import { Content } from "./components/Content";
+import { Action } from "./components/Action";
+import { TriggerButton, TriggerChip } from "./components/Trigger";
+import { ComboboxContextProvider } from "./ComboboxProvider";
+
+export interface ComboboxProps {
   readonly children: React.ReactNode;
```
darryltec, line 9: you're not expecting a random string to be put in as children, yeah? If so, 

```suggestion
  readonly children: ReactElement | ReactElement[];
```

so you won't have to check if the `child` is not a string

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,191 +1,71 @@
...
+  renderTrigger: ReactNode;
+  renderContent: ReactNode;
+} {
+  const childrenArray = React.Children.toArray(children);
+  let renderTrigger: ReactNode, renderContent: ReactNode;
+
+  childrenArray.forEach(child => {
+    if (isTriggerElement(child)) {
+      if (renderTrigger) {
+        useAssert(true, "Combobox can only have one Trigger element");
```
darryltec, line 34: in retrospect, `useAssert` should probably not be a hook (:past-derl:). But since this starts with `use` we'll treat it like a hook and say we's breaking dem [hook rulez](https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level)

> Don’t call Hooks inside loops, conditions, or nested functions

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,191 +1,71 @@
...
-  const filteredOptions = props.options.filter(option =>
-    option.toLowerCase().includes(searchValue.toLowerCase()),
-  );
-
-  function handleSearch(event: React.ChangeEvent<HTMLInputElement>) {
-    setSearchValue(event.target.value);
-  }
+function useComboboxValidation(children: ReactNode): {
+  renderTrigger: ReactNode;
+  renderContent: ReactNode;
```
darryltec, line 26: thoughts on calling them thangs as `triggerElement` and `contentElement` it isn't really trying to render but more of finding and returning an element.

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,191 +1,71 @@
...
+  });
 
-  function handleSelection(event: React.MouseEvent<HTMLButtonElement>) {
-    const selected = event.currentTarget.innerText;
-    setSelectedOption(selected);
-    setOpen(false);
-    setSearchValue("");
-    props.onSelection(selected);
+  if (!renderTrigger || !renderContent) {
+    useAssert(true, "Combobox must have a Trigger and Content element");
```
darryltec, line 44: Maybe more clearer if we refer to the component name?

```suggestion
    useAssert(true, "Combobox must have a Combobox.Trigger and Combobox.Content children");
```

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,191 +1,71 @@
...
+    React.isValidElement(child) &&
+    typeof child !== "string" &&
+    typeof child.type !== "string" &&
+    (child.type.name == "TriggerButton" || child.type.name == "TriggerChip")
   );
-};
-
-const Action = (props: ActionProps) => {
+}
+function isContentElement(child: ReactNode): boolean {
```
darryltec, line 61:
```suggestion
}

function isContentElement(child: ReactNode): boolean {
```

let em breathe

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,191 +1,71 @@
...
 
-      {/* only render actions if their are any */}
-      {props.children && <div className={styles.actions}>{props.children}</div>}
-    </div>
+function isTriggerElement(child: ReactNode): boolean {
+  return (
+    React.isValidElement(child) &&
+    typeof child !== "string" &&
+    typeof child.type !== "string" &&
+    (child.type.name == "TriggerButton" || child.type.name == "TriggerChip")
```
darryltec, line 58: can also skip the string check and match it with the component itself

```suggestion
    React.isValidElement(child) &&
    (child.type === TriggerButton || child.type === TriggerChip)
```

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,191 +1,71 @@
...
-        textColor="green"
-        fontWeight="bold"
-      >
-        {props.label}
-      </Typography>
-    </button>
+    React.isValidElement(child) &&
+    typeof child !== "string" &&
+    typeof child.type !== "string" &&
+    child.type.name == "Content"
```
darryltec, line 66: same here

```suggestion
    React.isValidElement(child) &&
    child.type === Content
```

packages/components/src/Combobox/ComboboxProvider.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { useEffect, useRef, useState } from "react";
+
+export const ComboboxContext = React.createContext(
+  {} as { open: boolean; setOpen: (open: boolean) => void },
+);
+
+export interface ComboboxProviderProps {
+  readonly children: React.ReactNode;
+}
+export function ComboboxContextProvider(
```
darryltec, line 10:
```suggestion
}

export function ComboboxContextProvider(
```

packages/components/src/Combobox/components/Action/Action.css
```diff
@@ -0,0 +1,12 @@
...
+  padding: var(--space-small) var(--space-base);
+  border: none;
+  text-align: left;
+  background-color: var(--color-surface);
+}
+
+.action:hover,
+.action:focus {
+  background-color: var(--color-surface--hover);
+}
```
darryltec, line 12: consistency wise, no focus glow? 
image

packages/components/src/Combobox/components/Content/Content.tsx
```diff
@@ -0,0 +1,89 @@
...
+interface ContentProps {
+  /**
+   * options for the combobox
+   */
+  readonly options: string[];
+
+  /**
+   * onSelection handler.
+   */
+  onSelection: (selection: string) => void;
```
darryltec, line 16:
```suggestion
  readonly onSelection: (selection: string) => void;
```

packages/components/src/Combobox/components/Content/Content.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * actions
+   */
+  readonly children: React.ReactNode;
```
darryltec, line 21: if this isn't gonna accept a string and just a list of elements 

```suggestion
  readonly children: ReactElement | ReactElement[];
```

packages/components/src/Combobox/components/Content/Content.tsx
```diff
@@ -0,0 +1,89 @@
...
+    setSearchValue(event.target.value);
+  }
+
+  function handleSelection(event: React.MouseEvent<HTMLButtonElement>) {
+    const selected = event.currentTarget.innerText;
+    setSelectedOption(selected);
+    setSearchValue("");
+    props.onSelection(selected);
+  }
+  return (
```
darryltec, line 43:
```suggestion
  }

  return (
```

packages/components/src/Combobox/components/Content/Content.tsx
```diff
@@ -0,0 +1,89 @@
...
+  function handleSearch(event: React.ChangeEvent<HTMLInputElement>) {
+    setSearchValue(event.target.value);
+  }
+
+  function handleSelection(event: React.MouseEvent<HTMLButtonElement>) {
+    const selected = event.currentTarget.innerText;
+    setSelectedOption(selected);
+    setSearchValue("");
+    props.onSelection(selected);
+  }
```
darryltec, line 42: nit: mostly because atlantis web is consistent on where the function goes. and it goes below the JSX. bit. Mostly when you're reading it, you get the context from the JSX first before you get to the nitty-gritty details.

packages/components/src/Combobox/components/Content/Content.tsx
```diff
@@ -0,0 +1,89 @@
...
+          type="search"
+          className={styles.searchInput}
+          placeholder="Search teammates"
+          onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+            handleSearch(event)
+          }
+          value={searchValue}
+        />
+
+        {/* if there is a search value show the clear icon */}
```
darryltec, line 56: do you still need to explain this?

packages/components/src/Combobox/components/Content/Content.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * actions
+   */
+  readonly children: React.ReactNode;
+}
+
+export function Content(props: ContentProps): JSX.Element {
+  const [searchValue, setSearchValue] = useState<string>("");
+  const [selectedOption, setSelectedOption] = useState<string>("");
```
darryltec, line 26: why is this a state? Should the consumer decide which one is selected?

packages/components/src/Combobox/components/Content/Content.tsx
```diff
@@ -0,0 +1,89 @@
...
+          role="button"
+          tabIndex={0}
+          key={option}
+          onClick={handleSelection}
+        >
+          {option}
+        </button>
+      ))}
+
+      {/* if there are no results matching the search value show no results item */}
```
darryltec, line 82: this one too do we need to explain this?

packages/components/src/Combobox/components/Content/Content.tsx
```diff
@@ -0,0 +1,89 @@
...
+      {filteredOptions.length === 0 && <p>No results for {searchValue}</p>}
+
+      {/* only render actions if their are any */}
```
darryltec, line 85: and also this one

packages/components/src/Combobox/components/Trigger/TriggerButton.tsx
darryltec, line 1: should we be actually restricting this to either just be a button or chip? Or just 1 type and variation of button?

packages/components/src/Combobox/components/Trigger/types.ts
darryltec, line 1: Should probably name this `Trigger.types.ts` mostly because we'll end up with mobile style of 
image

packages/components/src/Combobox/index.ts
```diff
@@ -1 +1,3 @@
-export { Combobox } from "./Combobox";
+export { Combobox, ComboboxProps } from "./Combobox";
+
+export { ComboboxContextProvider } from "./ComboboxProvider";
```
darryltec, line 3: Finally! The inverse of my nitpicks

```suggestion
export { Combobox, ComboboxProps } from "./Combobox";
export { ComboboxContextProvider } from "./ComboboxProvider";
```

darryltec (commented):
Got carried away here. Just a few thoughts here and there.

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,191 +1,71 @@
...
-  onSelection: (selection: string) => void;
-
-  /**
-   * actions
-   */
+import React, { ReactNode } from "react";
+import { useAssert } from "@jobber/hooks/useAssert";
+import { Content } from "./components/Content";
+import { Action } from "./components/Action";
+import { TriggerButton, TriggerChip } from "./components/Trigger";
```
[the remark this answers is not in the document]
ZakaryH: that's totally fair. if we can avoid adding more "noise" to imports like that, we should

so I took a look at what y'all did for DataList and I'm just gonna steal your own ideas and use them here!
https://github.com/GetJobber/atlantis/blob/master/packages/components/src/DataList/DataList.tsx#L82C1-L82C34

so what we'll do for Combobox's Content is....

`Content.tsx` -> `ComboboxContent.tsx`
```
export function ComboboxContent() {
...
```

`Combobox.tsx`
```
import { ComboboxContent } from "components/ComboboxContent";

...

Combobox.Content = ComboboxContent
```

and that keeps the   s i g n a t u r e   exactly the same for the consumer 🎉 

```
import { Combobox } from "@jobber/component/Combobox"
...
<Combobox>
  <Combobox.Content />
</Combobox>
```

packages/components/src/Combobox/Combobox.tsx
ZakaryH: true we're not expecting it no, but there's nothing stopping someone from doing so

I feel... I dunno, dishonest? about telling Typescript "hey this will only be this thing" when I can't possibly guarantee what someone might pass as `children`

if I promise TS that it'll only ever be an element, and we code around that but then we do get a string. it will most certainly break on something. 

so to me the question becomes how much do we want to protect people from themselves? if someone gives us nonsense and the component falls flat on its face, is that fine? or should we take measures to more elegantly handle incorrect input?

darryltec: Typescript will stop someone from doing that no? Like, that is the point of typescript. We said "only elements" and consumer said "nah, I want a string." The component should fail at that point if you intentionally suppress TS since we specifically said element only please.

IMO, supporting unsupported behaviour would create more problems and bugs for us.

ZakaryH: k I replied to you but changed my mind because I didn't understand something correctly. I think your suggestion makes sense! I'd even be tempted to go one step further with it and say

`children: ReactElement<ContentProps> | ReactElement<ContentProps>[] | ReactElement<TriggerProps> | ReactElement<TriggerProps>[]`

idk, that might be overkill. I'll give it a quick try here.

darryltec: That works too! Only advantage there is if you do drill down on it, you get the correct type of props

darryltec: Could also do the typescript way of recursive typing. Cmd + click ReactNode and see how it declared both single and array type without ever typing `this | this[]` 😉

packages/components/src/Combobox/Combobox.tsx
ZakaryH: true true true. kinda surprised TS didn't yell at me about that one.

ohhh I see how it's being used with conditions now! I just happened to look at all the usages where it was just `true` as the first arg, but you can obviously have it evaluate an expression there and then either throw it or not throw it

I'll update that, and yeah I agree. you could make a good argument for this being a util rather than a hook for what it needs to do. I don't think it being a hook adds a whole lot other than limitation of how it can be used.

packages/components/src/Combobox/Combobox.tsx
ZakaryH: I'm not too opinionated there. I've seen variable naming that prefixes with `render` for stuff like this, but I get what you're saying, technically it's already been rendered out where you invoked this component with `Combobox.Content` and at this point, we're dealing with the element

with that in mind, I'm fine with changing it to `blahElement`

packages/components/src/Combobox/Combobox.tsx
ZakaryH: yeah shore! never hurts to be more specific

the only thing is that we don't have a `Combobox.Trigger` the actual component will be either `Combobox.TriggerChip` or `Combobox.TriggerButton` and I'm not sure that we'd want to merge those together to make a single `Trigger` with props, since each one will want pretty different things

packages/components/src/Combobox/Combobox.tsx
ZakaryH: if it's gonna inflate my number of lines of codes added, you don't have to ask me twice. I'll give them all the space they could ever want.

packages/components/src/Combobox/Combobox.tsx
darryltec: oh, yeah missed that part. Hmmm maybe not then haha

packages/components/src/Combobox/Combobox.tsx
darryltec: yeah, we don't have the linter for that yet. we ain't as fancy as the mobile haha

packages/components/src/Combobox/Combobox.tsx
ZakaryH: what in the world. I swear I tried that at one point and it did not work, but yeah you're right that does the trick!

packages/components/src/Combobox/Combobox.tsx
ZakaryH: yuh yuh yuh

packages/components/src/Combobox/Combobox.tsx
darryltec: hey man, our performance is rated through number of commits and line added 

/s

packages/components/src/Combobox/components/Action/Action.css
ZakaryH: pls ignore styles for now, we have a follow up subtask to make everything beautiful for real!

this current stuff is just from our PoC

packages/components/src/Combobox/Combobox.tsx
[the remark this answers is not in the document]
darryltec: yeah, that, idk. haven't looked into it. That might also be an editor thing than a ts thing tho. 🤔

packages/components/src/Combobox/components/Action/Action.css
darryltec: ayt! Consider it ignored!

packages/components/src/Combobox/components/Content/Content.tsx
ZakaryH: fair enough! consider them moved

packages/components/src/Combobox/components/Content/Content.tsx
ZakaryH: probably not. no. I'll delete it.

goodnight sweet prince

packages/components/src/Combobox/components/Content/Content.tsx
ZakaryH: good question

not as of right now but coming soon in another subtask, yes!

imagine using a combobox for something in the edit state. you'd already have a selection, so we need to support the consumer being able to pre-set what is selected via `setSelectedOption`

darryltec: oh good. as long as there's subtask and we don't forget about it!

packages/components/src/Combobox/components/Content/Content.tsx
ZakaryH: I get it. you hate comments. it's fine, I'll just never write a comment ever again

darryltec: #selfDocumentingCode

packages/components/src/Combobox/components/Trigger/TriggerButton.tsx
ZakaryH: I'm in favour of 2 distinct components

I say that because I feel the props of a button are going to be different from the props of a chip. where a button has different variations/colors etc and a chip won't have those

so you'd now have to have conditional prop validation based on what you chose for the `type` prop on a single `Trigger`  which I believe is possible... probably.

also, thinking ahead to the future, there's a high probability of us doing something like a `TriggerSelect` or `TriggerTextInput`. the reason we're saving them for later is because they are quite different in UI and UX, where I am very skeptical we'll be able to make a single `Trigger` that has a sensible API for all those cases. I feel like it's going to get quite bloated and ugly if we go down that road. I could totally be wrong, but that's my gut feeling!

packages/components/src/Combobox/components/Trigger/types.ts
ZakaryH: hmmm

I'm not sure I think it's that bad having all those `types.ts` since on the end of all those results it'll say which file it's for, plus navigating to the types in your directory is pretty painless if your IDE has any kind of GUI for folders. 

I've been in mobile for the past year where those come from and personally I've never found it to be a pain point where I can't find the `types` I was looking for. but who knows, maybe others don't share my views.

is that where you're coming from, that it makes it difficult to find the `types` you are after?

I'm also wondering, let's say we do add a prefix for these types. would we be ok with ending up with a `Trigger.types` `Content.types` `Blah.types` etc. that sounds a lot like the problem we had with `Content` being an export when `Content` is already a thing

that's a lot of `Thing.types` files for a single component, and a lot of room for collision which might be the problem we're trying to avoid with `types`

I think either way, if we stick with `types.ts` or add a prefix, we should probably bring that up to `components/Combobox/` that way we have a single types file for all of the component, identical to the mobile/components-native setup. so if we do wanna prefix it, I'd suggest `Combobox.types.ts` and yeah put it up at the top level of this component's directory

packages/components/src/Combobox/index.ts
ZakaryH: no breathing 4 u

packages/components/src/Combobox/components/Trigger/types.ts
darryltec: > is that where you're coming from

yup

> that sounds a lot like the problem we had with Content 

not really since it will be named `ComboboxContent`

> yeah put it up at the top level of this component's directory

I like that better

ZakaryH: gotcha. ok I will put the file at the top level! on the topic of the naming though I'm not convinved. I still don't think it's hard to find the right types even if every file is named `types.ts`

thinking about how you would interact with this file, or find this file I think it's pretty much going to be either
- looking at the folder structure
- using some kind of fuzzy file search

the first case, I'd argue it's very self evident. the single `types.ts` file in the folder for the component you're interested in will be the correct one

the second case, what's the practical difference between leaving it as `types.ts` vs prefixing it with `Combobox.types.ts`?

if you start searching, you will need to add a prefix to your search term - that prefix will be `Combobox` or however little of that string is required to be unique. with the existing setup you're going to get the exact same result because it will match on the folder name
image


and if you add the prefix, it's the same term with the same result
image

so we're reducing consistency by adding a new naming convention that only applies to Atlantis (both mobile and online just do `MyComponent/types.ts`, it's a bit of a redundant naming pattern eg. `Combobox/Combobox.types.ts` and it's not obvious to me what it adds from a functional perspective

packages/components/src/Combobox/Combobox.tsx
ZakaryH: k I've stopped calling this conditionally and in a loop, although I'm not convinced the latter is entirely necessary

https://stackoverflow.com/questions/53906843/why-cant-react-hooks-be-called-inside-loops-or-nested-function#comment94658381_53906907
https://stackoverflow.com/questions/53546553/is-foreach-guaranted-to-loop-in-the-same-order-everytime-over-the-same-data-set

I'm pretty sure the order will always be respected here and we could get away with keeping it in the loop, but whatever moving it outside is trivial - the only downside is a bit uglier code imo

packages/components/src/Combobox/Combobox.tsx
ZakaryH: went ahead with half of this suggestion where it'll specify `Combobox.Content` but just `Trigger`

lmk if you think that's better or worse!

packages/components/src/Combobox/components/Trigger/types.ts
darryltec: That's fair. The goal really is to ensure we're doing things consistently within this repo. 

> the second case, what's the practical difference between leaving it as types.ts vs prefixing it with Combobox.types.ts?

Practical wise, it's much more clear that that first result there is the one you're looking for without having to scan through it to make sure it is indeed under combobox folder. This one is not really a great example since combobox is a unique name on itself.

Consistency wise, (mostly in the atlantis realm since it is maintained by 1 team compared to many) if we look at types file, right now, you have 3 patterns. Either `types.ts`, `Comp.types.ts`, or `CompTypes.ts` where there should only be 1.

> Should that even be a thing?

Have you seen a `styles.css` in Atlantis? I believe there's only 1  and it's under the design package that holds the most generic CSS in Atlantis.

Anyway, we're so consistent CSS that no one really questions if we should use `styles.css` instead of `Comp.css`.

Test files is another example.

As for index.ts, that one has a purposeful reason as to why it is just named `index.ts`. Mostly because that's an entry file so you won't have to repeat the name on the import.

ZakaryH: >This one is not really a great example since combobox is a unique name on itself.

yeah true, let's try something less unique. for example `button` which could be either `Button` or `ButtonGroup`
image

vs prefixed

image

so I see what you mean that you can stop reading earlier - I just feel like that's such a teeny tiny difference but I agree it is faster. if redundant names and consistency across all our projects aren't a concern, then I agree this would be preferable but on the topic of consistency....

>Consistency wise, (mostly in the atlantis realm since it is maintained by 1 team compared to many) if we look at types file, right now, you have 3 patterns. Either types.ts, Comp.types.ts, or CompTypes.ts where there should only be 1.

100% agree, which is my argument haha!

searching, you can see that `ThingTypes.ts` is the outlier compared to the majority of instances that follow `type.ts`

image

which is why I'm pushing back on adopting not even the lesser used pattern, but being the first to introduce a brand new naming pattern altogether since what we have today is `ThingTypes.ts` and `types.ts`. this is us adding a 3rd right now 🙀

>As for index.ts, that one has a purposeful reason as to why it is just named index.ts. Mostly because that's an entry file so you won't have to repeat the name on the import.

oh totally, I know the technical reason why it's that way - but I'm saying that if our argument is around files becoming hard to find when they share the same name, causing you to spend more time scanning through the full directory of a result, then we already have that exact problem with `index.ts`

so my question is if it's ok there, or it's not a problem there, what's different about `types.ts` that makes it a problem? like if we really wanted to, I'm sure we could create a setup that would enable us to do `Combobox.index.ts` and get the same behavior of not repeating the name on the import

darryltec: the sudden burst of `types.ts` is due to speedily moving the components native without adjusting the names. If you remove it from the equation, you'll have the 3 equally

darryltec: if consistency throughout is the question then there are more `FileName.css`, `FileName.types.ts`, `FileName.test.tsx

ZakaryH: fair, but even with those removed from the equation I don't see any instances of `Thing.type.ts`

it's either `ThingType.ts` or `type.ts` so we are still adding a new, 3rd way of expressing the same concept

darryltec: not really a new one, see https://github.com/GetJobber/atlantis/pull/1510#discussion_r1316335778

ZakaryH: can you show me an example of a `FileName.types.ts`?

darryltec: I mean, you're right there are not that much example of `.types.ts` but in the grand scheme of things where everything is named after the component with the other much more common files like css and test. 

The question now is should we continue on with types being just types compared to other files in the folder. Go bigger. it's not just the types file anymore.

ZakaryH: mkay, at this point I'm pretty confident we both understand each other's position and just have different opinions on the naming!

at the end of the day I'm just visiting, so I'm happy to go with `Thing.types.ts` if it's not a concern that we're adding that 3rd naming convention. I do see how that would be consistent with other files (`.css` `.tsx` etc.), it's just not consistent with other _type_ files

darryltec: yeah. the good thing is, next time we hit this discussion point, i can send them in this thread haha

ZakaryH: pushed!

packages/components/src/Combobox/Combobox.test.tsx
darryltec: yeah, i think this should be fine

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,193 +1,84 @@
...
-
-      {/* if there are no results matching the search value show no results item */}
-      {filteredOptions.length === 0 && <p>No results for {searchValue}</p>}
-
-      {/* only render actions if their are any */}
-      {props.children && <div className={styles.actions}>{props.children}</div>}
-    </div>
+    React.isValidElement(child) &&
+    typeof child !== "string" &&
+    typeof child.type !== "string" &&
```
darryltec, line 67: still had to check if it's a string eh?

packages/components/src/Combobox/components/Action/Action.tsx
darryltec, line 1: Consistency wise, this should also be prefixed like content. kinda odd now. (DataList is doing the same thing but we haven't found the time to actually rename that dang thing)

packages/components/src/Combobox/components/Trigger/TriggerButton.tsx
darryltec, line 1: same here with the action comment

darryltec (approved):
This looks good. The comment I have can be done in a different PR

packages/components/src/Combobox/Combobox.tsx
ZakaryH: ugh, no we definitely do not now that we've said it's either one or more `ReactElement`s

that's a quick fix tho

packages/components/src/Combobox/components/Action/Action.tsx
ZakaryH: agreed. I will double check with the others if we wanna do that in the individual subtasks for adjusting the Action & Trigger to reduce conflicts rather than doing it here

packages/components/src/Combobox/components/Trigger/TriggerButton.tsx
ZakaryH: also same as the other response

taylorvnoj: I just fixed this in my Trigger refactor PR and will do it in Action as well 👍

PR #1512, opened 2023-09-01
feat(components): implement ability to add a data list filters

darryltec:
## Motivations

Looking at a large set of data requires some form of filtering ability along with its microinteractions. This adds a place for the filters in DataList.

### Added

- An ability to pass in `DataList.Filters` to the DataList
- In the event that the amount of filters is bigger than the space we have, the filters would overflow with a fade indicator
- Added a context for DataList to reduce the amount of prop drilling we do which reduces the type issues we'd have in the future.
- Added a hook called `useInView` to check if an element is visible on the screen
  - In the essence of time, opted to put the documentation for it on another PR.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

To test the overflow, set the story size to small mobile

---

Random photo of Atlantis


rodrigoeidelvein (approved):
Actually, I'm going to approve and merge to unblock the search feature work. We still should do a round of design reviews in the whole component.

PR #1514, opened 2023-09-05
feat(components): add search ability for data list

darryltec:
## Motivations

What's the most common filter in any list? That's right. Search! This adds search functionality to the `DataList` by adding another compound component called `DataList.Search`. Not implementing this would remove the search functionality.

### Added

- DataListSearch component
- Allow the context to be the one-stop shop for grabbing the compound components in the future

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.const.ts
+++ b/packages/components/src/DataList/DataList.const.ts
@@ -20,3 +20,5 @@ export const BREAKPOINT_SIZES: Record<Breakpoints, number> = {
   lg: 1080,
   xl: 1440,
 };
+
+export const SEARCH_DEBOUNCE_DELAY = 500;
--- a/packages/components/src/DataList/DataList.tsx
+++ b/packages/components/src/DataList/DataList.tsx
@@ -5,10 +5,16 @@ import {
   DataListLayout,
   DataListLayoutProps,
 } from "./components/DataListLayout";
-import { DataListObject, DataListProps } from "./DataList.types";
+import {
+  DataListFiltersProps,
+  DataListObject,
+  DataListProps,
+  DataListSearchProps,
+} from "./DataList.types";
 import {
   generateDataListEmptyState,
   generateHeaderElements,
+  getCompoundComponent,
   getCompoundComponents,
 } from "./DataList.utils";
 import { DataListTotalCount } from "./components/DataListTotalCount";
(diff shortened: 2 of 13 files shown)


packages/components/src/DataList/DataList.const.ts
```diff
@@ -20,3 +20,5 @@ export const BREAKPOINT_SIZES: Record<Breakpoints, number> = {
   lg: 1080,
   xl: 1440,
 };
+
+export const SEARCH_DEBOUNCE_DELAY = 500;
```
rodrigoeidelvein, line 24:
```suggestion
export const SEARCH_DEBOUNCE_DELAY = tokens["timing-slowest"];
```

What do you think? Is it the same type of timing?

rodrigoeidelvein (commented):
I noticed that when setting the viewport to mobile on storybook web, it looks weird because it renders the scroll bar at the bottom of the filters, however it looks good when using a mobile device.
Small mobile (macbook)
image

Mobile device
image

I also just added a suggestion to use a token instead of a direct value, but I'm not sure if it's the same type of timing there.

packages/components/src/DataList/DataList.const.ts
darryltec: it can yeah, i believe i used it in one of the files here

packages/components/src/DataList/DataList.tsx
```diff
@@ -18,32 +24,49 @@ import {
...
-  totalCount,
-  headerVisibility = { xs: true, sm: true, md: true, lg: true, xl: true },
-}: DataListProps<T>) {
-  const allLayouts = getCompoundComponents<DataListLayoutProps<T>>(
+function InternalDataList() {
+  const {
+    data,
+    headers,
+    loading = false,
+    filterApplied = false,
```
MichaelParadis, line 62: No action needed at the moment, but do we still need this `filterApplied` prop now that the filters compound component has been added

darryltec: Gonna reevaluate on the other ticket about filtering and empty state

MichaelParadis: @darryltec sounds good :)

darryltec:
Alright, alignment fixed and token used @rodrigoeidelvein 

image

rodrigoeidelvein (approved):
LGTM!

PR #1516, opened 2023-09-06
feat(components): Adjust Combobox Content Position Dynamically

ZakaryH:
## Motivations

Currently the content and list of the opened Combobox always goes below the Trigger. This works if there's enough space there, but that won't always be the case. Popper enables us to define the preferred placement, and provide a fallback if that doesn't work - so we're doing that!

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


ZakaryH:
closing because git got weird and it's just 1 commit that I need anyways

PR #1518, opened 2023-09-06
feat(components): ComboboxContent 

ZakaryH:
## Motivations

Currently the content and list of the opened Combobox always goes below the Trigger. This works if there's enough space there, but that won't always be the case. Popper enables us to define the preferred placement, and provide a fallback if that doesn't work - so we're doing that!

Aside from that, we need to clean up and adjust a handful of other things about ComboboxContent that are required.
### Added

using `popper` to handle the adjustment of placement when the opened content can't fit below

close the open content on esc press

prop to configure the search placeholder text. defaults to "Search"
### Changed
use ReactDOM to place opened content elsewhere, not as a sibling to avoid overflow hiding it

updated how we close on clicks outside with the ReactDOM change

changed the shape of the expected `options` prop, since we will always need something unique (an id) to determine which item is selected since we cannot only rely on the label/string value. there may be times it will not be unique, plus we need to make sure we have a unique `key` for the `map` we're doing

## Testing

if you head over to the storybook instances, a quick way to try this out is to put a big ol div in the document above the Combobox like so
```
const ComboboxChip: ComponentStory<typeof Combobox> = args => {
  return (
    <div>
      <div style={{ width: "100%", height: "100vh", background: "#29d" }}></div>

      <Combobox {...args}>
      ...
```

this makes it so there's no room for the opened content to go below, and the popper behavior will kick in placing it above!

Kapture 2023-09-06 at 12 39 32

feel free to try it with other situations you can think of!

**ESC to close**
open it, and press ESC to close it

**Clicking outside to close**
should do exactly what it used to do, we're doing it in a different way now is all

**Searching**
should work the same as before with the new data shape

**Close on selection**
pretty simple. when you make a selection it should close and if you've done _something_ with the `onSelection` it'll get called with the thing you picked

---

Random photo of Atlantis


packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -19,6 +17,10 @@
   visibility: hidden;
 }
 
+.option {
+  display: block;
```
ZakaryH, line 21: not a necessary change here. having the options be block makes it easier to fill up the vertical space and cause the content to flip to the other side when testing this out

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -1,42 +1,66 @@
...
+    option.label.toLowerCase().includes(searchValue.toLowerCase()),
   );
 
-  return (
-    <div className={classnames(styles.content, !open && styles.hidden)}>
+  const template = (
+    <div
+      ref={popperRef}
+      data-testid="combobox-content"
+      className={classnames(styles.content, !open && styles.hidden)}
```
ZakaryH, line 54: I tried doing something higher above this component that would prevent it from rendering at all in `Combobox`

```
    <ComboboxContextProvider>
      {triggerElement}
      <ComboboxContentWrapper>
         {contentElement}
      </ComboboxContentWrapper/>
    </ComboboxContextProvider>

...
const ComboboxContentWrapper = (props: {children: ReactNode}): JSX.Element {
  const { open } = React.useContext(ComboboxContext);

  return (
    <>
     {open && (
       {props.children}
     )}
   </>
  )
}
```

so that way yeah it just doesn't render at all. we have to have a component wrapping it because since this is where the provider is first provided, it can't be used in `Combobox` yet

the problem here is that popper seems to get super lost and fails to adjust the location of the popper element which results in it being overtop over the trigger since it's `absolute top: 0 left: 0` and that's relative to the first parent, not relative to the container due to the `ReactDOM` placing it out there

packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -12,3 +12,8 @@ export interface ComboboxTriggerButtonProps
     Pick<ButtonProps, "type" | "variation" | "icon" | "iconOnRight"> {}
 
 export type ComboboxTriggerChipProps = ComboboxTriggerProps;
+
+export interface ComboboxOption {
+  id: string | number;
```
ZakaryH, line 17: I had this as only string, but decided to open it up because it's pretty normal to use numbers as ids too and I can't think of a solid reason we'd not want to support that

docs/components/Combobox/Web.stories.tsx
```diff
@@ -23,21 +23,21 @@ const ComboboxButton: ComponentStory<typeof Combobox> = args => {
...
-          "Frodo Baggins2",
-          "Pippin Took2",
-          "Merry Brandybuck2",
-          "Sam Gamgee2",
-          "Bilbo Baggins3",
-          "Frodo Baggins3",
-          "Pippin Took3",
-          "Merry Brandybuck3",
-          "Sam Gamgee3",
+          { id: "1", label: "Bilbo Baggins" },
```
MichaelParadis, line 26: Really like this approach, it will make handling server data more easily

packages/components/src/Combobox/Combobox.css
```diff
@@ -0,0 +1,12 @@
+.overlay {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  z-index: var(--elevation-menu);
+
+  @media (--small-screens-and-up) {
+    background-color: transparent;
```
MichaelParadis, line 10: Is there a reason I am missing as to why this isn't needed on all screens? I assume I am missing something but wanted to be sure

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -1,42 +1,66 @@
...
-    option.toLowerCase().includes(searchValue.toLowerCase()),
+    option.label.toLowerCase().includes(searchValue.toLowerCase()),
   );
 
-  return (
-    <div className={classnames(styles.content, !open && styles.hidden)}>
+  const template = (
+    <div
+      ref={popperRef}
+      data-testid="combobox-content"
```
MichaelParadis, line 53: For the testId here and other places we should follow our guide on naming testing ids
https://jobber.atlassian.net/wiki/spaces/TEAMATL/pages/2476933462/How+should+we+declare+test+ID+s+in+Atlantis+components

packages/components/src/Combobox/Combobox.css
ZakaryH: ah good catch. I pulled that same overlay styling from another instance where it uses a darker opacity on very small screens, but I deleted those styles and forgot to adjust this bit

given the state it is in right now, you're right I'll move that out of the media query and apply it all the time

MichaelParadis (approved):
LGTM!

PR #1521, opened 2023-09-07
feat(components): Ability to set selected option for combobox

dlrussel:
## Motivations

Add ability to set selected option for the combobox component. We may want the ability to pass in an initially selected value for the selected option for a combobox.

Also a small refactor to update the option element to be an input type="radio" instead of a button because this is more semantically correct and it will be easier to implement multiselect moving forward.

Feel free to let me know if the prop name I chose isn't ideal.

### Added

- Added a prop called `preSelectedOption` to set the default selected value from the combobox options.
- Added a test to assure that the selected option is set as the preselected option on render when a value is passed to the new prop.
- Added a test to throw an error if the preselected option provided is not a valid option.

## Testing
- Try passing in a non random value to the `preSelectedOption` prop and confirm that an error gets thrown
- Pass in a value that does exist in the options list and ensure that it is selected on initial render of the component.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -40,7 +40,7 @@ interface ComboboxContentProps {
    * @default ""
    * @optional
    * @type string
-   * */
+   */
   readonly selected?: string | number;
 }
 
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -14,7 +14,7 @@ import { Icon } from "../../../Icon";
 import { ComboboxContext } from "../../ComboboxProvider";
 import { ComboboxOption } from "../../Combobox.types";
 
-interface ComboboxContentProps {
+export interface ComboboxContentProps {
   /**
    * List of selectable options to display.
    */
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -14,7 +14,7 @@ import { Icon } from "../../../Icon";
 import { ComboboxContext } from "../../ComboboxProvider";
 import { ComboboxOption } from "../../Combobox.types";
 
-export interface ComboboxContentProps {
+interface ComboboxContentProps {
   /**
    * List of selectable options to display.
    */
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -34,6 +34,14 @@ interface ComboboxContentProps {
    * Placeholder text to display in the search input. Defaults to "Search".
    */
   readonly searchPlaceholder?: string;
+
+  /**
+   * pre selected option
+   * @default ""
+   * @optional
+   * @type string
+   * */
+  readonly preSelectedOption?: string;
 }
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -23,7 +23,7 @@ interface ComboboxContentProps {
   /**
    * Callback function invoked upon the selection of an option. Provides the selected option as an argument.
    */
-  readonly onSelection: (selection: ComboboxOption) => void;
+  readonly onSelect: (selection: ComboboxOption) => void;
 
   /**
    * Optional action button(s) to display at the bottom of the list.
@@ -41,7 +41,7 @@ interface ComboboxContentProps {
    * @optional
    * @type string
    * */
-  readonly setSelection?: string;
+  readonly selected?: string;
 }
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
@@ -116,7 +116,7 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
 
   function handleSelection(selection: ComboboxOption) {
     setSelectedOption(selection);
-    props.onSelection(selection);
+    props.onSelect(selection);
     setSearchValue("");
     setOpen(false);
   }
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -41,7 +41,7 @@ interface ComboboxContentProps {
    * @optional
    * @type string
    * */
-  readonly preSelectedOption?: string;
+  readonly setSelection?: string;
 }
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -41,7 +41,7 @@ interface ComboboxContentProps {
    * @optional
    * @type string
    * */
-  readonly selected?: string;
+  readonly selected?: string | number;
 }
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
@@ -55,11 +55,8 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
     popperRef,
     popperStyles,
     attributes,
-  } = useComboboxContent();
-
-  const filteredOptions = props.options.filter(option =>
-    option.label.toLowerCase().includes(searchValue.toLowerCase()),
-  );
+    filteredOptions,
+  } = useComboboxContent(props.selected, props.options);
 
   const template = (
     <div
@@ -76,16 +73,20 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
       />
       <div className={styles.list}>
         {filteredOptions.map(option => (
-          <button
-            className={classnames(
-              styles.option,
-              option.id === selectedOption?.id && styles.selectedOption,
-            )}
-            key={option.id}
-            onClick={() => handleSelection(option)}
-          >
+          <label key={option.id}>
+            <input
+              type="radio"
+              className={classnames(
+                styles.option,
+                option.id === selectedOption?.id && styles.selectedOption,
+              )}
+              checked={option.id.toString() === selectedOption?.id.toString()}
+              value={option.label}
+              name={option.label}
+              onChange={() => handleSelection(option)}
+            />
             {option.label}
-          </button>
+          </label>
         ))}
 
         {filteredOptions.length === 0 && (
@@ -165,7 +166,10 @@ function Search(props: {
   }
 }
 
-function useComboboxContent(): {
+function useComboboxContent(
+  selected: string | number | undefined,
+  options: ComboboxOption[],
+): {
   searchValue: string;
   setSearchValue: React.Dispatch<React.SetStateAction<string>>;
   selectedOption: ComboboxOption | null;
@@ -177,12 +181,20 @@ function useComboboxContent(): {
   popperRef: React.RefObject<HTMLDivElement>;
   popperStyles: { [key: string]: React.CSSProperties };
   attributes: { [key: string]: { [key: string]: string } | undefined };
+  filteredOptions: ComboboxOption[];
 } {
+  const defaultValue = options.find(
+    option => option.id.toString() === selected?.toString(),
+  );
   const [searchValue, setSearchValue] = useState<string>("");
   const [selectedOption, setSelectedOption] = useState<ComboboxOption | null>(
-    null,
+    defaultValue || null,
   );
   const { open, setOpen, wrapperRef } = React.useContext(ComboboxContext);
+
+  const filteredOptions = options.filter(option =>
+    option.label.toLowerCase().includes(searchValue.toLowerCase()),
+  );
 
   useOnKeyDown(() => {
     if (open) {
@@ -217,5 +229,6 @@ function useComboboxContent(): {
     popperRef,
     popperStyles,
     attributes,
+    filteredOptions,
   };
 }
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -57,10 +57,10 @@ const ComboboxButton: ComponentStory<typeof Combobox> = args => {
           "Sam Gamgee3",
 >>>>>>> 4fec09b3 (Revert "Refactor to use a ref")
         ]}
-        onSelection={selection => {
+        onSelect={selection => {
           console.log(selection);
         }}
-        setSelection="Bilbo Baggins"
+        selected="Bilbo Baggins"
       >
         <Combobox.Action
           label="Add a teammate"
@@ -99,7 +99,7 @@ const ComboboxChip: ComponentStory<typeof Combobox> = args => {
           { id: "4", label: "Merry Brandybuck" },
           { id: "5", label: "Sam Gamgee" },
         ]}
-        onSelection={selection => {
+        onSelect={selection => {
           console.log(selection);
         }}
       >
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -42,6 +42,7 @@ const ComboboxButton: ComponentStory<typeof Combobox> = args => {
         onSelection={selection => {
           console.log(selection);
         }}
+        preSelectedOption="Bilbo Baggins"
       >
         <Combobox.Action
           label="Add a teammate"
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -42,7 +42,7 @@ const ComboboxButton: ComponentStory<typeof Combobox> = args => {
         onSelection={selection => {
           console.log(selection);
         }}
-        preSelectedOption="Bilbo Baggins"
+        setSelection="Bilbo Baggins"
       >
         <Combobox.Action
           label="Add a teammate"
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -23,7 +23,6 @@ const ComboboxButton: ComponentStory<typeof Combobox> = args => {
       />
       <Combobox.Content
         options={[
-<<<<<<< HEAD
           { id: "1", label: "Bilbo Baggins" },
           { id: "2", label: "Frodo Baggins" },
           { id: "3", label: "Pippin Took" },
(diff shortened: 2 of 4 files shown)


packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -19,17 +20,33 @@ interface ComboboxContentProps {
...
    */
   readonly children: ReactElement | ReactElement[];
+
+  /**
+   * pre selected option
+   * @default ""
+   * @optional
+   * @type string
+   * */
+  readonly preSelectedOption?: string;
```
ZakaryH, line 30: with the name, I think this makes perfect sense when we're using it for editing

what I'm wondering about is other ways you might want to use this prop. specifically I'm thinking about how you might clear a selection. I feel like that's a case we should support when I think of team assignment, or tags on a report - both of those it's perfectly valid to have nothing selected to either keep it unassigned or see your report with no tags reducing your results

maybe we don't want to use this prop to support that use case/flow which is totally fair but I think we _could_

I'm imagining the following
- click the trigger
- open the list
- choose something, setting the internal selection and doing whatever on the consumer side with that returned selection
- decide hey I don't want that anymore
- use a `setSelection` prop that will, as it suggests, set the internal selection - in this case setting it to `null`

to me that feels flexible where you could use it for editing and clearing? although what I've described is definitely more of a forwardedRef/imperativeHandle pattern where you'd do `myComboboxRef.current.setSelection(null)` or `myComboboxRef.current.setSelection(userThatWasAlreadySelected)`

now that I've typed it out though, do we even care about the internal state of this component? like imagine I have that form and I've used the value returned from `Combobox`. 

I can clear that myself... since `Combobox` has no visual representation of what's selected, what even would be the point of clearing the component's internal state? 🤔

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -19,17 +20,33 @@ interface ComboboxContentProps {
...
 }
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
+  const { preSelectedOption } = props;
```
ZakaryH, line 34: very much a style/preference thing, but to me the value of doing `props: ComboboxContentProps` is that when you use it via `props.preSelectedOption` it becomes very evident that it came from a prop rather than a hook or a local variable!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -19,17 +20,33 @@ interface ComboboxContentProps {
...
 
   const filteredOptions = props.options.filter(option =>
     option.toLowerCase().includes(searchValue.toLowerCase()),
   );
+  let invalidPreSelectedOption = false;
+  if (preSelectedOption && !props.options.includes(preSelectedOption)) {
+    invalidPreSelectedOption = true;
+  }
 
+  useAssert(invalidPreSelectedOption, "Preselected option must be in options");
```
ZakaryH, line 49: wondering how heavy handed we need to be here. if someone puts `"nonsense.jpg"` as the `preSelectedOption` what happens?

the component still renders happily, they can still use it as normal once they open it and choose something that **is** in the list

the only thing that doesn't work of course is that nothing is preselected. is that punishment enough or enough incentive to give it something it can find & use?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -19,17 +20,33 @@ interface ComboboxContentProps {
...
+  const [selectedOption, setSelectedOption] = useState<string>(
+    preSelectedOption || "",
+  );
   const { open } = React.useContext(ComboboxContext);
 
   const filteredOptions = props.options.filter(option =>
     option.toLowerCase().includes(searchValue.toLowerCase()),
   );
+  let invalidPreSelectedOption = false;
+  if (preSelectedOption && !props.options.includes(preSelectedOption)) {
```
ZakaryH, line 45: I think this could be reduced down to a one liner

```
 const shouldGetUpset: boolean =
    !!props.preSelectedOption &&
    !props.options.includes(props.preSelectedOption);
```

ZakaryH (commented):
one code style suggestion. not mandatory by any means.

one suggestion for reducing some conditional assignment

question about `useAssert`

question about `radio` inputs

larger question about if this should be a prop like we have now, or if we want to consider a `useImperativeHandle` approach to either directly expose the internal `setSelectedOption` or maybe some modified version of it, which could then work for both edit and clearing. if we even care about clearing that is, I'm not entirely sure we do!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
dlrussel: Hm I like this thought process. I  agree that the `setSelection` naming convention would be more flexible so I do think that's a good refactor. Although combobox technically wouldn't be linked to a visual representation of the output wouldn't we still want to clear the state of the combobox if we wanted to clear the visual output considering than one would be linked to the other?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
dlrussel: ah yes! good point we should stay consistent here 👍🏻

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: yeah... thinking on this a bit more, while we could get away with not clearing the internal state for single select things will definitely get awkward with multi select where there will be an intermediary visual representation of what is selected via the check boxes & count, even if it also has no final output

if we solely rely on the consumer to manage clearing/updating the data outside the component, you could end up in weird states where the piece of the form using `Combobox` might say "nothing selected" but then you open up the combobox and it still shows your previous selections

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
dlrussel: yup! great suggestion!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -55,17 +72,19 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
...
-        <button
-          className={classnames(
-            option === selectedOption && styles.selectedOption,
-          )}
-          role="button"
-          tabIndex={0}
-          key={option}
-          onClick={handleSelection}
-        >
+        <label htmlFor={option} key={option}>
```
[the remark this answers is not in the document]
dlrussel: I don't think that it makes sense to keep these as buttons. Semantically a radio makes more sense here because a selection is taking place. 
In terms of the implementation for multiselect I was thinking that we would have a prop that would set a variable to dictate the type for the input and either make it radio or checkbox. that way the logic could stay relatively simple.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
dlrussel: hm ya I considered this definitely open to opinions on how strict to be here. I figured we would want a pre selected option that doesn't exist to to throw an error but open to your opinion 🤔

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -19,17 +20,31 @@ interface ComboboxContentProps {
...
    */
   readonly children: ReactElement | ReactElement[];
+
+  /**
+   * pre selected option
+   * @default ""
+   * @optional
+   * @type string
+   * */
+  readonly setSelection?: string;
```
darryltec, line 30: oh hey, it's me!

curious about the intent for this prop. 

Is it to determine which option is selected? Kinda like inputs have `value` and paired with `onChange`. In combobx terms, it's `setSelection` and `onSelection`?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -11,19 +11,29 @@ interface ComboboxContentProps {
   readonly options: string[];
 
   /**
-   * onSelection handler.
+   * onSelect handler.
```
darryltec, line 14: Hmm... we should explain this better for the newer peeps who aren't familiar with the system. Also stating intent.

> Callback that fires whenever an option is selected.

And probably something along the lines of the selected params is the label of the option?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -11,19 +11,29 @@ interface ComboboxContentProps {
...
    */
-  readonly onSelection: (selection: string) => void;
+  readonly onSelect: (selection: string) => void;
```
darryltec, line 16: nit: mostly because it follows the prop name that it eventually fills out

```suggestion
  readonly onSelect: (selected: string) => void;
```

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -11,19 +11,29 @@ interface ComboboxContentProps {
...
   /**
    * actions
    */
   readonly children: ReactElement | ReactElement[];
+
+  /**
+   * pre selected option
+   * @default ""
+   * @optional
+   * @type string
```
darryltec, line 27: hmm... we haven't added the whole @optional and @type before. mostly because typescript handles those for us automatically. Should we add those?

Also, calling this "pre selected option" isn't really true in a sense that it is the selected option whenever the user interacts with it (assuming it is implemented correctly)

Maybe

> The selected value on the combobox content

or something

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -11,19 +11,29 @@ interface ComboboxContentProps {
...
+   * */
```
darryltec, line 28:
```suggestion
   */
```

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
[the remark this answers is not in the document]
darryltec: just an fyi, radio's will automatically move the selected option to the next or previous using the arrow keys. Not sure if we want that. Do we want that? That seems different than other menu-type UI where you are finding an option before selecting it (HTML select, autocomplete).

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -11,19 +11,29 @@ interface ComboboxContentProps {
...
+  readonly selected?: string;
 }
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
```
darryltec, line 32: I actually don't mind the `prop` vs destructured (`{ onChange }`) as much but since we've been very consistent around it too. Does this land on "depends on the context" style or "nah, we need to be consistent" side? I actually don't know 🤔

darryltec (commented):
just a few food for thoughts.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -11,19 +11,29 @@ interface ComboboxContentProps {
   readonly options: string[];
```
darryltec, line 11: unrelated, but probs should ask it now. 

@ZakaryH with your time as a person who implements Atlantis, for components and UI's like this, is it important to know the id of the thing you're selecting too? 

I know on a user selector for chips, a team needed to have the ID on the data as well since finding the ID based on the name (label) is not fun.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -11,19 +11,29 @@ interface ComboboxContentProps {
...
+   * @type string
+   * */
+  readonly selected?: string;
 }
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
   const [searchValue, setSearchValue] = useState<string>("");
-  const [selectedOption, setSelectedOption] = useState<string>("");
+  const [selectedOption, setSelectedOption] = useState<string>(
+    props.selected || "",
```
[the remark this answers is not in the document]
darryltec: Actually, now that I see it, do we need to track the state internally?

ZakaryH:
I was trying the flow where we'd want to clear the selection, and it won't let me use `null` which is fair. we did type it as optional AKA undefined, or string or number

image

are we ok with `undefined` being used in two different ways here?

one where you totally omit the `selected` prop and never gave anything at all

and another where you explicitly do `selected={someState}` where in the case that we've cleared it `someState` is now `undefined`

given how we've configured the types, `undefined` becomes the only way people would be able to use the clearing selection functionality. that might be a bit too opinionated? it would force a very specific data type for their state

eg.

```
function MyCoolComponent() {
   const [selectedHat, setSelectedHat] = useState<Hat | undefined>(undefined);
   const options = [{id: 1, label: "fedora"}, {id:2, label: "baseball cap"}];

  return (
   <button onClick={() => setSelectedHat(undefined)>Clear Selection</button>
   <Combobox>
      <Combobox.TriggerChip>
      <Combobox.Content
        options={options}
         selected={selectedHat.id}
         onSelection={(selection) => setSelectedHat(selection)}
       >
          <Combobox.Action>
      </Combobox.Content>
  </Combobox>
  )

}

```

personally in my mind I do treat `null` and `undefined` differently. `undefined` to me is that it doesn't exist at all whereas `null` is like yeah we know about it, but it hasn't been interacted with or it hasn't been set _yet_

also I'm realizing that this makes the consumer have to do some extra work here in the clearing case... they can't just set `selectedHat` to be `undefined` or `null` because if they do, then the prop doing `selectedHat.id` that will blow up asking for `id` :/

can we improve that by asking for the entire object? so `selected` would now accept an entire `ComboboxOption`, and then internally we'd know to just grab the `id` off that for our comparisons. that lets the consumer have simpler code of `selected={selectedHat}` and they can easily clear with a `setSelectedHat(undefined)` (or maybe `null` if we decide to support that)

I know we really want to get this merged, and we do have a separate task to look at the clearing flow - so perfectly valid to say "let's deal with this later" and leave this as is for the purposes of this PR!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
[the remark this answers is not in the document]
ZakaryH: interesting, this radio button setup we have currently is doing something unexpected where I can't give focus via tabs to the radios at all

I can't tell if I'm being dumb, but I cannot for the life of me get the focus to move to the radios and using the arrows only ever scrolls either the inner list, or the outer Storybook page. it never changes between the radios/labels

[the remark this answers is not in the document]
darryltec: oh right. that involves some fun CSS trickery. https://atlantis.getjobber.com/?path=/story/components-selections-chips-web--basic

Since you don't focus on the label, you focus on the input radio

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: I strongly prefer the `prop.thing` syntax and avoid destructuring (when it comes to props specifically, hooks and other stuff is fine)

it's definitely from my old Ember days where you'd be in a component and see `option` but have absolutely no way of knowing where `option` came from

is it 
- an internal property?
- a property being passed in via Ember's version of props
- a mixin?

new Ember solves this by prefixing props with `@thing` and then internal state with `this.thing` making it much quicker to know where the values are coming from, which I think is really nice!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: it is! we've changed the shape of `options` in another PR to include an id now

cause yeah it's entirely possible for you to have two clients with the name `Jimothy Stevens`, and while that might be confusing to choose from in a list, we do need to have some way to know which one we selected

then it'd be up to the consumer to figure out which is the right one if all we're showing is the name, but that wouldn't be a concern of this component. not yet anyway!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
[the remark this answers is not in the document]
ZakaryH: technically no. for single select you're right, we could get away with not having an internal state because you click, it selects, closes and you're done

however for multi select since you'll be able to click on something, click on something else, maybe change your mind and undo one of them etc. there's all sorts of changes happening to the selections that we would not want to inform the consumer of until you are done and happy with your choices. otherwise that could needlessly trigger all sorts of updates to anything listening to your selections

so we'll eventually need this because the internal and external states will have valid cases where they are temporarily out of sync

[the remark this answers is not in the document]
ZakaryH: though now that I say it, I feel like there could be cases where you do want to immediately see the selections happening live

almost feels like that could be a prop specifically for multi select. should it trigger `onSelect` the second you interact with options, or should it wait until you've closed it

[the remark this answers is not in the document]
darryltec: I believe most of the cases would be instant. If it does get to this point

> otherwise that could needlessly trigger all sorts of updates to anything listening to your selections

That's the throttling pattern where the consumer can also implement and that we don't have to worry about it until we hear otherwise.

```tsx
<MyComponent value={value} onChange={handleChange} />

...
const throttleFetch = throttle(doSomething, 300);

function handleChange(newValue) {
  // cancel existing func
  doSomething.cancel()

  // set new values
  setValue(newValue);

  // try again
  doSomething();
}
```

Our components need to be predictable and part of it being predictable is not handling a state that the user can change within it since you risk the internal state and external state being out-of-sync.

You also have the part where you have outside forces that can change the selected value on that component. That means you would have to sync the state internally. One example of that is saved filters that can be applied anytime. Not just on mount.

Plus, single select deffo don't need this so let's start with simple

darryltec:
@ZakaryH is that due to the internal state? If we remove it and let the component react to the prop changes rather than us having to calculate the state, that clearing issue would go away, no?

As for null vs undefined, I agree with the distinction but in this component, I don't think we need to if we follow the controlled component way. If you tell the component that prop doesn't exist (`undefined` in react is also special) it will revert back to whatever no value means.

As for the `selected.hat` TS will tell you `selected` could be undefined thus making you write `selected?.hat` which returns undefined to the prop.

dlrussel:
@ZakaryH I see your point. I say lets merge this because it's just going into a feature branch and address the concern in its own pr.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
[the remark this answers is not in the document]
ZakaryH: I'm not convinced a `throttle` or `debounce` works there though

like think about how people use these types of lists. I am assigning team members, and doing _something_ expensive with those selections, so I want to have a high degree of certainty that I have the final results before calculating based on the selections

if I have even 15 entires in the list, I start out and I pick Adam. next, I want to pick Zak. how long does it take me to either scroll down to the bottom of the list, or to use the search and type it in and then select that user?

I'd wager it's going to be more than 300ms, so you're going to have to bump that up significantly. I don't even know what value you'd use because sometimes if all your selections are beside one another, 300ms is fine - but if you need to spend time getting to your next selection it totally depends on how well the user can navigate our UX

if you make that debounce too high, it's going to feel laggy or unresponsive for those quick cases, and you can't guarantee it'll even be enough for users who make selections very slowly

that's where I'm like ok, there's no way we can know - we should offer the options to just wait until it closes. I do understand your argument about consistency and predictability but I don't see how someone could make a good UX using this component in this scenario (btw it is a real scenario I know that is coming soon, not just something I've made up!)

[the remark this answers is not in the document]
darryltec: That's a good point! 

> like think about how people use these types of lists. I am assigning team members, and doing something expensive with those selections, so I want to have a high degree of certainty that I have the final results before calculating based on the selections

Doesn't this land on a "Do what you need, double check the other _fields_, then click save when you're done" pattern?

I do get where you're coming from tho. We need to ensure we only support and build UX's that we understand and would do right now. Majority of the use cases in Jobber where you select things are still all under 1 big save button (settings, edit pages).

I don't doubt that last scenario tho. It will happen for sure. When it does, we should support that at that point since it comes with a lot of micro-interaction that other (majority of) use cases wouldn't.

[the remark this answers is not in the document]
ZakaryH: >Doesn't this land on a "Do what you need, double check the other fields, then click save when you're done" pattern?

typically yes, you're right. the case I have in mind is atypical in that it is doing live, async stuff to update your UI and generate further options that you would then choose, make sure you're happy with and then save. so it might be a complete outlier that we do not want to design for. also this is coming up relatively soon as I understand it, which is why I'd like to do some thinking about it now rather than waiting because the solution doesn't feel obvious to me

it feels like we're in a bit of a pickle with two conflicting constraints. we want these flows to feel familiar so that users learn it once and know it for all of the product. however; in this instance the cost is too great to have that same behavior so we need to do something different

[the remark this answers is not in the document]
darryltec: Agreed! Which is also a design conundrum! Since consistency should also be their responsibility.

PR #1523, opened 2023-09-08
feat(components): Combobox Search

ZakaryH:
## Motivations

A few minor adjustments and improvements to the search functionality of the combobox.

## Changes

updated the way we are applying the gradient to the top and bottom of the list area when it has enough items to be scrollable. the previous way resulted in the `.search` div having a larger height than intended due to the pseudo elements, causing the centering of the X clear button to be off-center.

changed the markup around the clear button to be an actual semantic button. added a focused state here to make it easier to see when interacting with it via keyboard.

added a QoL/UX improvement to give the search input focus again after clearing it either with keyboard or clicks because I found it annoying to have to re-focus it after clearing

updated some class names to be a bit clearer

## Testing

not a ton new to test here besides the re-focus after clearing

aside from that it should look OK, though we still have another subtask to finalize the styles on everything

searching should filter results if there are applicable options that match the string
- should be case insensitive
- should have a no-result message
- should reset after a selection
- search should feel fine to use with keyboard only

---

Random photo of Atlantis


packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -67,39 +69,43 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
...
-          {option.label}
-        </button>
-      ))}
-
-      {filteredOptions.length === 0 && <p>No results for {searchValue}</p>}
-
-      {props.children && <div className={styles.actions}>{props.children}</div>}
+        {filteredOptions.length === 0 && <p>No results for {searchValue}</p>}
+      </div>
+      <div className={styles.actions}>{props.children && props.children}</div>
```
ZakaryH, line 98: not sure if I want to stick with this

the reason I'm keeping the wrapping div even if there were no actions is because that's where I put the `actions:before` to have the gradient

I'm quite sure there's a way to do the same thing with the new `.list` div I added to wrap the options. very open to not rendering the `.actions` and doing it another way!

taylorvnoj: We're going to have some pretty sticky merge conflicts haha - I had to update this quite a bit in my [Actions PR](https://github.com/GetJobber/atlantis/pull/1519/files#diff-3dc00b67c89a4fd17c5f5e14c8a5f794ee1b676a3453f7a29316513f38b86b31) to adjust for styling on the last Action.  It's going to be worth going back after everything is in the feature branch and combing through it once more

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -51,56 +39,69 @@
...
+  background-color: var(--color-surface--background);
+  transform: translateY(-50%);
+  justify-content: center;
+  align-items: center;
 }
 
-.selectedOption {
-  background-color: var(--color-surface--background);
+.clearSearch:focus {
+  border: 1px solid var(--color-focus);
```
taylorvnoj, line 61: Can you just do `box-shadow: var(--shadow-focus);` and get everything you need here?  You might also need `outline: none;` since it's a button.

taylorvnoj: Okay actually pulled and tested it - `box-shadow` is probably best imo.  `outline: none` isn't needed

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -67,39 +69,43 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
...
-          onClick={() => handleSelection(option)}
-        >
-          {option.label}
-        </button>
-      ))}
-
-      {filteredOptions.length === 0 && <p>No results for {searchValue}</p>}
-
-      {props.children && <div className={styles.actions}>{props.children}</div>}
+        {filteredOptions.length === 0 && <p>No results for {searchValue}</p>}
```
taylorvnoj, line 96: Can we add the " " surrounding the {searchValue} 
Screenshot 2023-09-08 at 12 05 01 PM

taylorvnoj:
Okay sorry to keep adding comments before you even put it up for review 😆 

As far as styling goes there are some little changes we'll have to make like the clear search icon padding-right, font-family for search, colour of "No results..." text  but I don't mind leaving that for the last styling ticket.  Those would definitely be in the refinement stage.  If you also want to ignore the focus css on the clear, that's ok too.

Otherwise looking really great - I love giving the input search focus again after a clear 👏

ZakaryH:
@taylorvnoj no need to apologize at all, this **is** ready for review so that's exactly what I'm looking for !

I have been bouncing around with a few other things this morning and haven't moved it out of draft yet is all

ZakaryH:
the squash and merge is causing headaches here where git thinks all the commits from the parent I branched off (that have already been merged) belong to this branch :| 

I could fix this with a rebase and keep this branch, but honestly I just don't want to do that when it's a single commit. gonna make a new branch and PR and just link to this one

ZakaryH:
closing this because I don't want to deal with git right now. please come and visit me at 

https://github.com/GetJobber/atlantis/pull/1526

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: added this on the other branch!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
ZakaryH: yeah I wasn't sure if border was the right thing to do here. I should have checked other existing patterns. `box-shadow` looks good to me and I've added it on the other branch

PR #1524, opened 2023-09-08
fix(components): adjust no result state of data list

darryltec:
## Motivations

Needed to officialize the `filtered` prop and follow the code pattern for the empty state. It used to also use a state to deal with filters but we can keep it pure react and let it react to change rather than us having to manually set a state.

This also changes how you define empty states in the data list. You can now set a custom message for filters empty state. If you don't specify one, there are default messages for both empty and filtered.

### Added

- Contain empty state in a DataListEmptyState component

### Changed

- 

### Deprecated

- 

### Removed

- utility to generate empty state

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


rodrigoeidelvein (approved):
LGTM! Tested in Storybook with the new `filtered` prop and it works well! Code-wise looks good too!

PR #1525, opened 2023-09-08
feat(components): allow DataList to infinite scroll

darryltec:
## Motivations

The current implementation of DataList forces you to load everything in one go. This allows the user to fire a callback to load more once the user has scrolled closer to the end.

Instead of listening to scroll–which we don't know the element that is scrolling–we watch an invisible element at the end of the list if it's visible using the `useInView` hook. Once visible, this triggers the `onLoadMore` callback for the consumers to start loading more items.

### Added

- `onLoadMore` callback

### Changed

- 

### Deprecated

- 

### Removed

- The ability for the tags to reset everytime the component updates. This removes the flicker that is happening whenever you load more. This isn't really needed as the component is reactive enough to rerender whenever a change happens.

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/components/DataList/Web.stories.tsx
+++ b/docs/components/DataList/Web.stories.tsx
@@ -43,7 +43,7 @@ const Template: ComponentStory<typeof DataList> = args => {
     data,
     /* See useCollectionQuery for example on how to load more */
     // refresh,
-    // nextPage,
+    nextPage,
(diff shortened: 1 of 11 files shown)


MichaelParadis (approved):
LGTM!

PR #1526, opened 2023-09-08
feat(components): Combobox Search

ZakaryH:
## Motivations

A few minor adjustments and improvements to the search functionality of the combobox.

## Changes

updated the way we are applying the gradient to the top and bottom of the list area when it has enough items to be scrollable. the previous way resulted in the `.search` div having a larger height than intended due to the pseudo elements, causing the centering of the X clear button to be off-center.

changed the markup around the clear button to be an actual semantic button. added a focused state here to make it easier to see when interacting with it via keyboard.

added a QoL/UX improvement to give the search input focus again after clearing it either with keyboard or clicks because I found it annoying to have to re-focus it after clearing

updated some class names to be a bit clearer

## Testing

not a ton new to test here besides the re-focus after clearing

aside from that it should look OK, though we still have another subtask to finalize the styles on everything

searching should filter results if there are applicable options that match the string
- should be case insensitive
- should have a no-result message
- should reset after a selection
- search should feel fine to use with keyboard only

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -326,6 +326,6 @@ describe("ComboboxContent Search", () => {
     const searchInput = getByPlaceholderText("Search");
     fireEvent.change(searchInput, { target: { value: "Bilbo" } });
 
-    expect(getByText("No results for Bilbo")).toBeInTheDocument();
+    expect(getByText('No results for "Bilbo"')).toBeInTheDocument();
   });
 });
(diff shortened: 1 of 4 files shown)


ZakaryH:
clean git history version of https://github.com/GetJobber/atlantis/pull/1523

ZakaryH:
@taylorvnoj just pushed up one more commit that I'd like to get reviewed & unofficially approved even though this is already technically approved and I could merge it 

I'm feeling that the ComboboxContent is getting a bit more bloated than I'd like. I don't really want to make a full separate component where I'd do `export Search` because then it's available elsewhere, so I'm making a small local component to help keep the markup in `ComboboContent` at least a baby bit cleaner and move some concerns that are very specific to Search into its own space

I think if we ever build a `InputSearch` type component this also makes it easy to replace, and I could definitely see that being a thing since there are common bits of accessibility & UX around search inputs that would be valuable to have across the board when we're doing searches. having a component you can plop in and use would be much nicer than having to build it yourself each time.

taylorvnoj (approved):
chefs kiss on that local search component. ship it.

PR #1527, opened 2023-09-11
feat(components): add 'back to top' on DataList

darryltec:
## Motivations

Once you've hit the end of the list and there's nothing to load more, we want the users to have easy access to go back to the top when there are more than 50 items in the list. Mostly because it's only helpful when you have a large set of scrollable data.

This implements a button to go back to the top but only if the component is not loading more. There's also a target on how far top it goes back to. As per the design requirements, we want it to go back to the top of the list but not necessarily to the title of the DataList or the top of the scrollable parent.

### Added

- Back to top button
- A target to go back to top since see motivations above

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Keep scrolling down until there's nothing to load

---

Random photo of Atlantis


rodrigoeidelvein (approved):
LGTM!

PR #1528, opened 2023-09-11
feat(components): Combobox zero index state

taylorvnoj:
## Motivations

When there are no `options` loaded into Combobox, there should be a zero index state.  

### Added

- an optional prop called `contentNoun` that is meant to describe what the options are within Combobox (example `teammates`, `tax rates`, etc)
- established a variable for whether options exist or not

Zero index state will either return a default or customized string:
Default:
Screenshot 2023-09-11 at 3 19 57 PM

Setting "teammates" to `contentNoun`:
Screenshot 2023-09-11 at 3 20 36 PM

## Testing

- in the Combobox `Web.stories.tsx` file, comment out the array of options and see the default zero index state message
- now add `contentNoun` to the `Combobox.Content` and pick a noun like "teammates" or "tax rates" and see the customized zero index state message appear
- make sure that search and filter still work as intended

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -411,7 +411,7 @@ describe("Combobox Zero Index State", () => {
           <Combobox.TriggerButton label="Select a tax rate" />
           <Combobox.Content
             options={[{ id: "1", label: "10%" }]}
-            onSelection={jest.fn()}
+            onSelect={jest.fn()}
           />
         </Combobox>,
       );
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -371,3 +371,38 @@ describe("Combobox selected value", () => {
     expect(option).toBeChecked();
   });
 });
+
+describe("Combobox Zero Index State", () => {
+  describe("when no options exist and contentNoun is provided", () => {
+    it("should render the correct zero index state text", () => {
+      const contentNoun = "teammates";
+      const { getByText } = render(
+        <Combobox>
+          <Combobox.TriggerButton label="Select a teammate" />
+          <Combobox.Content
+            options={[]}
+            onSelect={jest.fn()}
+            subjectNoun={contentNoun}
+          />
+        </Combobox>,
+      );
+
+      expect(
+        getByText("You don't have any teammates yet."),
+      ).toBeInTheDocument();
+    });
+  });
+
+  describe("when no options exist and contentNoun is not provided", () => {
+    it("should render the default zero index state text", () => {
+      const { getByText } = render(
+        <Combobox>
+          <Combobox.TriggerButton label="Select a tax rate" />
+          <Combobox.Content options={[]} onSelect={jest.fn()} />
+        </Combobox>,
+      );
+
+      expect(getByText("No options yet.")).toBeInTheDocument();
+    });
+  });
+});
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -387,9 +387,7 @@ describe("Combobox Zero Index State", () => {
         </Combobox>,
       );
 
-      expect(
-        getByText("You don't have any teammates yet."),
-      ).toBeInTheDocument();
+      expect(getByText("You don't have any teammates yet")).toBeInTheDocument();
     });
   });
 
@@ -402,7 +400,7 @@ describe("Combobox Zero Index State", () => {
         </Combobox>,
       );
 
-      expect(getByText("No options yet.")).toBeInTheDocument();
+      expect(getByText("No options yet")).toBeInTheDocument();
     });
   });
 });
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -373,16 +373,16 @@ describe("Combobox selected value", () => {
 });
 
 describe("Combobox Zero Index State", () => {
-  describe("when no options exist and contentNoun is provided", () => {
+  describe("when no options exist and subjectNoun is provided", () => {
     it("should render the correct zero index state text", () => {
-      const contentNoun = "teammates";
+      const subjectNoun = "teammates";
       const { getByText } = render(
         <Combobox>
           <Combobox.TriggerButton label="Select a teammate" />
           <Combobox.Content
             options={[]}
             onSelect={jest.fn()}
-            subjectNoun={contentNoun}
+            subjectNoun={subjectNoun}
           />
         </Combobox>,
       );
@@ -391,7 +391,7 @@ describe("Combobox Zero Index State", () => {
     });
   });
 
-  describe("when no options exist and contentNoun is not provided", () => {
+  describe("when no options exist and subjectNoun is not provided", () => {
     it("should render the default zero index state text", () => {
       const { getByText } = render(
         <Combobox>
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -403,4 +403,20 @@ describe("Combobox Zero Index State", () => {
       expect(getByText("No options yet")).toBeInTheDocument();
     });
   });
+
+  describe("when options do exist", () => {
+    it("should not render default zero index state text", () => {
+      const { queryByText } = render(
+        <Combobox>
+          <Combobox.TriggerButton label="Select a tax rate" />
+          <Combobox.Content
+            options={[{ id: "1", label: "10%" }]}
+            onSelection={jest.fn()}
+          />
+        </Combobox>,
+      );
+
+      expect(queryByText("No options yet")).not.toBeInTheDocument();
+    });
+  });
 });
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -42,6 +42,11 @@ interface ComboboxContentProps {
    * @type string
    */
   readonly selected?: string | number;
+
+  /**
+   * The encapsulating noun for the content of the combobox. Should be pluralized.
+   */
+  readonly subjectNoun?: string;
 }
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
@@ -58,6 +63,8 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
     filteredOptions,
   } = useComboboxContent(props.selected, props.options);
 
+  const optionsExist = props.options.length > 0;
+
   const template = (
     <div
       ref={popperRef}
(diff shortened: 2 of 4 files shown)


packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -34,6 +34,11 @@ interface ComboboxContentProps {
    * Placeholder text to display in the search input. Defaults to "Search".
    */
   readonly searchPlaceholder?: string;
+
+  /**
+   * The encapsulating noun for the content of the combobox. Should be pluralized.
+   */
+  readonly contentNoun?: string;
```
taylorvnoj, line 41: not tied to this prop name - very open to suggestions!

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -329,3 +329,36 @@ describe("ComboboxContent Search", () => {
...
+        <Combobox>
+          <Combobox.TriggerButton label="Select a tax rate" />
+          <Combobox.Content options={[]} onSelection={jest.fn()} />
+        </Combobox>,
+      );
+
+      expect(getByText("No options yet")).toBeInTheDocument();
+    });
+  });
+});
```
ZakaryH, line 364: what do you think about one more test case, where we do have a > 0 length of options, and we assert that the `No options yet` should **not** be found?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: off the top of my head some other synonyms could be....
- topic
- subject
- noun
- theme
- concept

I kinda like topic and subject the best. wondering about if we need to prefix with `content` when it's a prop of the content component

that said, I do like noun because it's a very.... grammatical? word. to me, it heavily implies that its usage is tied to wording, to sentence structure

hmmm, what do you think about a combination of say `subjectNoun` or `topicNoun`? or is that too redundant?

also I have one more suggestion that might change how we wanna name or prefix is

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -34,6 +34,11 @@ interface ComboboxContentProps {
    * Placeholder text to display in the search input. Defaults to "Search".
    */
   readonly searchPlaceholder?: string;
```
ZakaryH, line 36: now that we're providing a noun prop, I feel like this search placeholder prop is obsolete and it could leverage that same noun prop to either say 

just `Search` if you didn't give us a noun

or `Search Teammates` if you did!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -141,6 +151,14 @@ function Search(props: {
   }
 }
 
+function getZeroIndexStateText(contentNoun?: string) {
+  if (contentNoun) {
+    return `You don't have any ${contentNoun} yet`;
+  }
+
+  return "No options yet";
```
ZakaryH, line 159: definitely more of a designer thing, but this wording feels very sterile. it makes me think of

image


if this was the recommended wording we settled on though, feel free to ignore me!

ZakaryH (commented):
looking good!

a couple suggestions/questions around adding another test case, wording, and possibly re-using the prop for the search placeholder value to reduce the amount of props

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
taylorvnoj: `subjectNoun` actually makes perfect sense to me - subject nouns are the subject of a sentence so saying "I'm choosing a 13% tax rate" means the subject noun would be "tax rate". Boom. Done.

packages/components/src/Combobox/Combobox.test.tsx
taylorvnoj: Yes - great idea!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
taylorvnoj: Yeah like I mentioned yesterday in office hours, this is the verbiage we got from Design.  I will still point this out as something to check in the official design review to double check, since this suggestion was mentioned in a slack thread

packages/components/src/Combobox/Combobox.test.tsx
taylorvnoj: Done here - 5b612edce4bbc27891df8ecb55c74dfb70917cff

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
taylorvnoj: yes - great idea! I did this here 252b47d936722c0d731d5d51c3119092945f9835
learned about nullish coalescing operator `??` today. Let me know if there's a different approach you were thinking

ZakaryH:
one thing that is probably exactly what we want, but I'll mention it just the same

image


when you have no options, and search for something that doesn't get a result, the "no options" takes priority over the "no search results"

I think this is A-OK for today; it does make me wonder if that will always be true though. maybe once we enable custom handlers on search terms, if you'd want to return those results rather than looking at the initial options you were passed. a problem for future us. maybe.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -31,9 +31,9 @@ interface ComboboxContentProps {
   readonly children?: ReactElement | ReactElement[];
 
   /**
-   * Placeholder text to display in the search input. Defaults to "Search".
+   * The encapsulating noun for the content of the combobox. Should be pluralized.
```
ZakaryH, line 34: very smol nitpick/suggestion

could we mention where/how it is used? perhaps say "The encapsulating noun for the content of the combobox. Used in the empty state, and search placeholder. Should be pluralized."

the only reason I say that is because I think it's reasonable to assume that the Trigger and Action buttons also use the subject for their text content, but they don't. those are just open ended `label`s

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: nice!

so the nullish coalescing operator has 1 case that is arguably worse with our particular usage here due to the fact that the prop is a string

https://stackoverflow.com/a/61481631

you can see in that example `"" || "Some default"` returns `"Some default"` whereas `"" ?? "Some default"` returns `""`

of course it's nonsense to give us an empty string and to the user it will look identical, but if you inspect it - we are actually putting a space at the end of the placeholder
image

I don't think it's the worst, but I also don't think it really adds anything. that plus the fact that the logic we have for the empty message is `if (subjectNoun)` which behaves like `subjectNoun || "Deafult"` I'd be in favour of having those two be in sync as far as how they treat the absence of the `subjectNoun`

tl;dr - I think `||` is actually better here!

taylorvnoj: Ohhh I see! Okay `subjectNoun={props.subjectNoun || ""}` makes more sense so I'll change that for sure.

I'm confused about the space happening after search though -  `"Search "` - how is that happening?

ZakaryH: it's because of how we are using the value here
`placeholder={`Search ${props.subjectNoun}` || "Search"}`

so when `props.subjectNoun` is an empty string, it's not actually that value that is creating the space rather it's the space between `Search` and `${props.subjectNoune}` that we're seeing

the fact that the `subjectNoun` is present by this logic and making it go down that path is what causes the unexpected result

ZakaryH: on that note, we should adjust how this is working because this will never evaluate the right hand side of the `||` since we will always have a non empty string on the left hand side

rather than conditionally passing in A or B to 
```
<Search
  subjectNoun={ A ?? B}
...
```

I think this could be simplified by always passing in `props.subjectNoun`. now this can be undefined since it's optional, so `subjectNoun` on `Search` becomes optional too

then, we handle all the logic in one place inside `Search` where we'd do something like
```
placeholder={props.subjectNoun ? `Search ${props.subjectNoun}` : "Search"
```

I'm also tempted to change the name of the Search prop from `subjectNoun` -> `placeholder` because that's a lot more tailored to how it will be used. `subjectNoun` is a lot more broad, and while makes sense as a name for the `Combobox.Content` prop, it's less clear what it is going to do in the context of `Search`

that's a super nitpicky thing though and I don't feel that strongly about it

taylorvnoj: Yaaay ok - by changing how we're using `subjectNoun`, that gets rid of the `"Search "` problem because now it can actually get to the default like you said.  Love that approach and not having near duplicate `||` everywhere 👍 

I'm ok with changing `subjectNoun` to `placeholder` because I agree with you - my only FUD is... even though I will update the description of the prop to explain the usage, I could image someone seeing the word `placeholder` and assuming that would be the entire string.  I guess they'd find out pretty quickly that it's only the noun part but... hmmm.  I think I'll ask the rest of the team for opinions on that one, see what the general consensus is.

taylorvnoj:
> when you have no options, and search for something that doesn't get a result, the "no options" takes priority over the "no search results"

@ZakaryH I see - so you mean, if there are no options of course we see "No options yet".  But if we put something in the search bar, then that message should change to "No results for...". Basically anytime a search is done without a corresponding option, we see the "No results" message.

If we think this will need to change in the future, I think it's worth fixing now...

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
taylorvnoj: Yup, agreed.  Thing about me Zak, is I love a smol nitpick. Bring the nit every time.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
taylorvnoj: Update - I totally misunderstood this!  Changed the name of the `Search` prop here d40760ef585d85207fd671d2df27391c31a2ff0c

Sorry Zak!  I understand now and get why we'd want to rename this in Search.

taylorvnoj:
To wrap up some of the loose ends here after chatting with Zak:

In regards to overriding "No options yet" with "No results for.." when there are no options, the decision is to not include this logic in this PR now.  The code surrounding this area will likely change in the near future so for now, let's keep it simple with the standard "No options..." text.

The internal `Search` component `subjectNoun` prop was renamed to `placeholder`.  This doesn't affect the naming of `subjectNoun` within `ComboboxContentProps` - `placeholder` just now calls that prop within `Search`

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -371,3 +371,52 @@ describe("Combobox selected value", () => {
...
+  });
+
+  describe("when options do exist", () => {
+    it("should not render default zero index state text", () => {
+      const { queryByText } = render(
+        <Combobox>
+          <Combobox.TriggerButton label="Select a tax rate" />
+          <Combobox.Content
+            options={[{ id: "1", label: "10%" }]}
+            onSelection={jest.fn()}
```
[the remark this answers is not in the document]
taylorvnoj: Strange - vsCode did have the red underline but very strange the linter was ok with it.

ZakaryH (approved):
thanks for making all the changes

LGTM!

PR #1529, opened 2023-09-12
feat(components): JOB-78452 Select DataList row

MichaelParadis:
## Motivations

- Added ability to select rows in the DataList.
- Made it on the consumer to keep track of the selectedRows. This is to match with other parts of the DataTable such as filters
- I was not able to add tests for the visibilty of the Checkboxes because there is a bug with JS-Dom where it doesn't cascade CSS styling.

### Added

- Added Ability to Select rows in DataList. This is needed for bulk actions in the future

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

After:

Large Hover:

image

image
NOTE: Due to the way layouts can be customized however consumers want I was not able to align the checkbox to the label on multiline rows

Small Screens:
image

Note: Some of the alignment issues with the label and checkbox will be resolved https://jobber.atlassian.net/browse/JOB-75935

Here is the alignment with label without the Grid workaround to line up with the inline label
image

---

Random photo of Atlantis


packages/components/src/DataList/DataList.css
```diff
@@ -35,6 +35,30 @@
...
+
+.listItem .selectable > :first-child {
+  visibility: visible;
+  @media (--medium-screens-and-up) {
+    visibility: hidden;
+  }
+}
+
+.listItem:hover .selectable > :first-child,
+.listItem:focus .selectable > :first-child,
```
darryltec, line 57: `listItem` isn't focusable yet so this won't do anything. If this is meant to show the checkbox the moment you focus on the checkbox, `focus-within` is what you're looking for

packages/components/src/DataList/DataList.css
```diff
@@ -35,6 +35,30 @@
...
+
+  @media (--medium-screens-and-up) {
+    align-items: center;
+  }
+}
+
+.listItem .selectable > :first-child {
+  visibility: visible;
+  @media (--medium-screens-and-up) {
+    visibility: hidden;
```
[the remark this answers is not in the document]
darryltec: if you use `opacity` and add a simple `transition` on it, it'll have a better feeling when hovering on wider screens

packages/components/src/DataList/DataList.types.ts
```diff
@@ -97,6 +97,21 @@ export interface DataListProps<T extends DataListObject> {
...
   readonly onLoadMore?: () => void;
+
+  /**
+   * Possible actions for selected items in the DataList.
+   */
+  readonly actions?: Record<string, (items: DataListObject["id"][]) => void>;
+
+  /**
+   * Callback when an item row is clicked.
+   */
```
darryltec, line 108: i believe it is a callback when the checkbox is clicked

packages/components/src/DataList/DataList.types.ts
```diff
@@ -97,6 +97,21 @@ export interface DataListProps<T extends DataListObject> {
...
+  readonly onSelectChange?: (items: DataListObject["id"][]) => void;
```
darryltec, line 109: - As per my other comment, rename this to `onSelect`. 
- Should create a type alias for `DataListObject["id"]` since you are using it on multiple places now
- You should also use `T["id"]` since that narrows it down more. 
  - If the consumer has only passed in a `string` as an id...
    - `T["id"]` will be typed as only `string`
    - `DataListObject["id"]` will always be `number | string` regardless

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
```diff
@@ -38,3 +43,36 @@ export function DataListItems<T extends DataListObject>({
     />
   );
 }
+
+function ListItemInternal<T extends DataListObject>({
```
darryltec, line 47: Put this on its own file. One file one component.

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
```diff
@@ -38,3 +43,36 @@ export function DataListItems<T extends DataListObject>({
...
+  item: T;
+}) {
+  const { selectedItems, actions, onSelectChange } = useDataListContext();
+
+  if (actions) {
+    return (
+      <div
+        className={classNames(styles.selectable, [
+          selectedItems.length > 0 && styles.selected,
+        ])}
```
darryltec, line 61: forgot that you could do arrays (or no array) with the `&&` to conditionally add CSS. Majority of the time it is written as 

```ts
classNames(styles.selectable, {
  [styles.selected]: selectedItems.length,
})
```

packages/components/src/DataList/context/DataListContext/DataListContext.tsx
```diff
@@ -6,6 +7,9 @@ export const defaultValues = {
   data: [],
   headers: {},
   children: [],
+  actions: undefined,
+  onSelectChange: noop,
```
darryltec, line 11: this prop is optional so regardless of it having a `noop` or not being defined here, typescript will still force you to guard it. In short, don't really need to do this

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    });
+    renderItems();
+    const checkbox = screen.getByRole("checkbox");
+
+    await userEvent.click(checkbox);
+
+    expect(onSelectChangedMock).toHaveBeenCalledTimes(1);
+    expect(onSelectChangedMock).toHaveBeenCalledWith([]);
+  });
+});
```
darryltec, line 86: Should add one more test for multiple selected items and see if it does add onto the array instead of overwriting it

packages/components/src/DataList/DataList.tsx
```diff
@@ -67,6 +67,7 @@ export function DataList<T extends DataListObject>(props: DataListProps<T>) {
         filterComponent,
         layoutComponents,
         emptyStateComponents,
+        selectedItems: props.selectedItems ?? [],
         ...props,
```
darryltec, line 71: won't this get overwritten? 

image

darryltec (changes requested):
Just some thoughts here and there. Might need to talk to matt about the alignment of things. Also (not really because of this), we should fix that tags wrapping issue with `white-space: nowrap`

docs/components/DataList/Web.stories.tsx
```diff
@@ -81,9 +81,16 @@ const Template: ComponentStory<typeof DataList> = args => {
     created: new Date(node.created),
   }));
 
+  const [selectedItems, setSelectedItems] = useState<Array<number | string>>(
```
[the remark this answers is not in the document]
MichaelParadis: Good catch

docs/components/DataList/Web.stories.tsx
```diff
@@ -81,9 +81,16 @@ const Template: ComponentStory<typeof DataList> = args => {
...
+    [],
+  );
+
   return (
     <DataList
       {...args}
+      actions={{}}
```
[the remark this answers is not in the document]
MichaelParadis: Sounds good, I will add a temporary prop that we can remove in the future

docs/components/DataList/Web.stories.tsx
```diff
@@ -81,9 +81,16 @@ const Template: ComponentStory<typeof DataList> = args => {
...
+      selectedItems={selectedItems}
+      onSelectChange={setSelectedItems}
```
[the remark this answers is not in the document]
MichaelParadis: Sounds good

packages/components/src/DataList/DataList.css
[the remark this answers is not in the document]
MichaelParadis: Hmm good point, I will use opacity and a transition

docs/components/DataList/Web.stories.tsx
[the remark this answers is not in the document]
darryltec: no, don't add a temporary prop. use the prop we have

packages/components/src/DataList/DataList.types.ts
MichaelParadis: I was running into issues relating to consumers with only numbers getting an error of number[] | undefined not compatable with number[] | string[] | undefined. number not compatable with string

packages/components/src/DataList/DataList.css
```diff
@@ -35,6 +35,33 @@
   border-bottom: var(--border-base) solid var(--color-border);
 }
 
+.selectable {
+  display: grid;
+  column-gap: var(--space-small);
+  align-items: flex-start;
+  grid-template-columns: max-content minmax(0px, auto);
```
darryltec, line 42: You sent me a message about `auto` not working well. Especially on smaller screens.

The answer to that is because, like flex, grid also adds a `min-width` of `auto` on the target element [by default](https://atlantis.getjobber.com/?path=/story/components-selections-chips-web--basic). Unfortunately, some elements that are meant to not wrap and overflow aren't working well with it. So the fix is to tell that child to have a `min-width` of 0. Or what you did here. Which cancels out that issue.

packages/components/src/DataList/DataList.css
```diff
@@ -35,6 +35,33 @@
...
+
+  @media (--medium-screens-and-up) {
+    align-items: center;
+  }
+}
+
+.listItem .selectable > :first-child {
+  opacity: 1;
+  --transition-properties: var(--timing-base) ease-in-out;
```
fagnanm, line 51: Can we speed this up, feels a little bit slow

fagnanm (commented):
> Due to the way layouts can be customized however consumers want I was not able to align the checkbox to the label on multiline rows
This will still exist after the label changes are complete as the labels will only go from 25.7px tall down to 24px.

Can we not add top margin to the checkbox at the smaller size?

packages/components/src/DataList/DataList.types.ts
```diff
@@ -97,6 +97,16 @@ export interface DataListProps<T extends DataListObject> {
...
+
+  /**
+   * Callback when an item checkbox is clicked.
+   */
+  readonly onSelect?: (items: T["id"][]) => void;
+
+  /**
+   * The list of Selected Item ids
+   */
+  readonly selected?: Array<DataListObject["id"]>;
```
darryltec, line 109: Does this need to be `T["id"]` as well?

MichaelParadis: @darryltec I was running into an issue. See https://github.com/GetJobber/atlantis/pull/1529#discussion_r1324651923

packages/components/src/DataList/DataList.css
MichaelParadis: @darryltec thanks for the explanation as to what is going on

packages/components/src/DataList/components/DataListLayoutInternal/DataListItemInternal.tsx
```diff
@@ -0,0 +1,39 @@
...
+import styles from "../../DataList.css";
+import { DataListObject } from "../../DataList.types";
+
+export function ListItemInternal<T extends DataListObject>({
+  children,
+  item,
+}: {
+  children: JSX.Element;
+  item: T;
+}) {
```
darryltec, line 14: for cleanliness sake, new interface, please.

packages/components/src/DataList/components/DataListLayoutInternal/DataListItemInternal.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { selected, onSelect } = useDataListContext();
+
+  if (typeof selected !== "undefined" && onSelect) {
```
darryltec, line 17: that an interesting approach. Out of curiosity, why does checking `selected !== undefined` not working?

packages/components/src/DataList/components/DataListLayoutInternal/DataListItemInternal.tsx
```diff
@@ -0,0 +1,39 @@
...
+      >
+        <Checkbox
+          checked={selected?.includes(item.id)}
+          onChange={() => {
+            if (selected?.includes(item.id)) {
+              onSelect?.(selected?.filter(id => id !== item.id));
+            } else {
+              onSelect?.([...selected, item.id]);
+            }
+          }}
```
darryltec, line 32: do you mind on putting this under a `handleChange` function down below? 

Mostly to not muddy up this JSX. Better clean it now than it growing to an insane amount because no one wants to do it anymore in fear of "breaking" it

packages/components/src/DataList/components/DataListLayoutInternal/DataListItemInternal.tsx
MichaelParadis: Honestly, I think it was a small remnant of when you compare types for narrowing you do this but undefined works. There is some stuff related to undeclared variables with it but that isn't a concern here
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#typeof_operator_and_undefined

packages/components/src/DataList/DataList.types.ts
darryltec: That's odd that one of the prop works but the other doesn't. 

Where are you seeing the error? I changed the example id from gql string to just map index numbers. Which changed the `T` id to a `number` instead of a string and it seems to work fine.

MichaelParadis: I was seeing it in the tests in `DataList.test.tsx`

darryltec: oh, that's because of the spread. since it is just grabbing a `typeof DataList` it defaults to number or string then tries to pass it in. more of a test setup problem than a consumer problem

MichaelParadis: @darryltec I was able to fix it in 9312b4d73eef3b2a4f60f234a4605152d7d85206

docs/components/DataList/Web.stories.tsx
```diff
@@ -81,9 +81,13 @@ const Template: ComponentStory<typeof DataList> = args => {
     created: new Date(node.created),
   }));
 
+  const [selectedItems, setSelectedItems] = useState<Array<string>>([]);
+
   return (
     <DataList
       {...args}
+      selected={selectedItems}
+      onSelect={setSelectedItems}
```
darryltec, line 90: actually, we should not implement/expose this yet because the "select all" isn't there yet and this will be misaligned on the example 

image

MichaelParadis: I can remove it once we get finished the review before merging. That way we can test and before approving it

MichaelParadis: @darryltec not sure if it pinged you

darryltec: agreed! yup. I get pinged on everything!

MichaelParadis:
@fagnanm 

> Can we not add top margin to the checkbox at the smaller size?

My one concern with this is that due to Layouts on smaller sizing being fully customizable adding padding may cause issues in the future

fagnanm:
> @fagnanm
> 
> > Can we not add top margin to the checkbox at the smaller size?
> 
> My one concern with this is that due to Layouts on smaller sizing being fully customizable adding padding may cause issues in the future

I think we'll be okay because there should always be a first line that the checkbox will line up with. If not we can adjust as necessary when the time comes.

MichaelParadis:
@fagnanm I added some padding. Let me know what you think

packages/components/src/DataList/components/DataListLayoutInternal/DataListItemInternal.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Checkbox } from "../../../Checkbox";
+import { useDataListContext } from "../../context/DataListContext";
+import styles from "../../DataList.css";
+import { DataListObject } from "../../DataList.types";
+
+interface ListItemInternalProps<T extends DataListObject> {
+  children: JSX.Element;
+  item: T;
+}
+export function ListItemInternal<T extends DataListObject>({
```
darryltec, line 12: mostly because file name and grouping

```suggestion
}

export function DataListItemInternal<T extends DataListObject>({
```

darryltec (commented):
just a few minute things that I noticed.

Mostly code quality/QOL things for when time has passed and we need to quickly scan the code (with our eyes and brain) and update things. But we end up easily missing things because things aren't _stylistically_ grouped together/obvious. (Pretty much the same argument as 1 liner if's)

fagnanm:
> @fagnanm I added some padding. Let me know what you think

@MichaelParadis looks good. Also the rest of the PR looks good design-wise.

packages/components/src/DataList/components/DataListLayoutInternal/DataListItemInternal.tsx
```diff
@@ -0,0 +1,42 @@
...
+      >
+        <Checkbox
+          checked={selected?.includes(item.id)}
+          onChange={handleChange}
+        />
+        {children}
+      </div>
+    );
+  }
+  return children;
```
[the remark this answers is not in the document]
darryltec: What about the space between? keep that code grouping clean and consistent as well. Can't just do it in 1 place but not the other.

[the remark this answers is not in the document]
darryltec: oh sorry. wrong comment. I mean this one https://github.com/GetJobber/atlantis/pull/1529#discussion_r1324957964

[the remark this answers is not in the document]
MichaelParadis: @darryltec pushing right now

packages/components/src/DataList/DataList.css
```diff
@@ -35,6 +35,35 @@
...
+  @media (--medium-screens-and-up) {
+    align-items: center;
+  }
+}
+
+.listItem .selectable > :first-child {
+  padding-top: var(--space-smaller);
+  opacity: 1;
+  transition: opacity var(--transition-properties);
+  --transition-properties: var(--timing-quick) ease-in-out;
```
darryltec, line 53: do you intend to reuse this?

darryltec (approved):
:shipit:

packages/components/src/DataList/DataList.css
MichaelParadis: At the moment no. However we _may_ want it in the future. We can remove it if it isn't needed during select all

PR #1530, opened 2023-09-12
feat(components): Refine styling on Combobox

taylorvnoj:
## Motivations

This PR is to refine and finalize styling on the Combobox component.

### Added

Action & Options:
- pointer cursor on hover
- focus state 
- `var(--space-smaller)` gap between each
- `checkmark` icon on selected option

Filter & zero index state messges:
- proper text color and styling

### Fixed

Search:
- long text in the search field no longer goes behind the `clearSearch` button

## Testing

Find strange happenings, discrepancies & try to break things!

**Note:** fade on scroll after search and before actions, is being pushed to another ticket

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
@@ -8,8 +8,7 @@
 .actionButton {
   position: relative;
   width: 100%;
-  height: calc(var(--space-base) * 2.75);
-  padding: 0;
+  padding: var(--space-small) 0;
   border: none;
   text-align: left;
   background-color: var(--color-surface);
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
@@ -9,7 +9,7 @@
   position: relative;
   width: 100%;
   height: calc(var(--space-base) * 2.75);
-  padding: var(--space-small) 0;
+  padding: 0;
   border: none;
   text-align: left;
   background-color: var(--color-surface);
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
@@ -8,7 +8,7 @@
 .actionButton {
   position: relative;
   width: 100%;
-  height: 44px;
+  height: calc(var(--space-base) * 2.75);
   padding: var(--space-small) 0;
   border: none;
   text-align: left;
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
@@ -8,6 +8,7 @@
 .actionButton {
   position: relative;
   width: 100%;
+  min-height: calc((var(--space-largest) - var(--space-small)));
   padding: var(--space-small) 0;
   border: none;
   text-align: left;
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
@@ -13,8 +13,10 @@
   border: none;
   text-align: left;
   background-color: var(--color-surface);
+  cursor: pointer;
 }
 
+.actionButton:hover,
 .actionButton:focus {
   border-radius: var(--radius-large);
   outline: none;
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
@@ -11,19 +11,19 @@
   min-height: calc((var(--space-largest) - var(--space-small)));
   padding: var(--space-small) 0;
   border: none;
+  border-radius: var(--radius-large);
   text-align: left;
   background-color: var(--color-surface);
   cursor: pointer;
+  transition: all var(--timing-base);
 }
 
 .actionButton:focus {
   box-shadow: var(--shadow-focus);
-  border-radius: var(--radius-large);
   background-color: var(--color-surface--background);
 }
 
 .actionButton:hover {
-  border-radius: var(--radius-large);
   background-color: var(--color-surface--background);
 }
 
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
@@ -16,10 +16,14 @@
   cursor: pointer;
 }
 
-.actionButton:hover,
 .actionButton:focus {
+  box-shadow: var(--shadow-focus);
   border-radius: var(--radius-large);
-  outline: none;
+  background-color: var(--color-surface--background);
+}
+
+.actionButton:hover {
+  border-radius: var(--radius-large);
   background-color: var(--color-surface--background);
 }
 
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
@@ -8,6 +8,7 @@
 .actionButton {
   position: relative;
   width: 100%;
+  height: 44px;
   padding: var(--space-small) 0;
   border: none;
   text-align: left;
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
@@ -1,7 +1,7 @@
 .content {
   z-index: var(--elevation-tooltip);
   max-width: var(--popover--width);
-  max-height: calc(var(--space-largest) * 8.75);
+  max-height: calc(var(--space-largest) * 8.25);
   min-width: calc(var(--space-extravagant) * 3.75);
   box-shadow: var(--shadow-base);
   margin-top: var(--space-small);
@@ -28,12 +28,14 @@
 
 .option {
   display: flex;
-  justify-content: space-between;
-  align-items: center;
+  min-height: calc((var(--space-largest) - var(--space-smaller)));
+  box-sizing: border-box;
   padding: var(--space-small);
   color: var(--color-heading);
   font-weight: 500;
   cursor: pointer;
+  justify-content: space-between;
+  align-items: center;
 }
 
 .selectedOption,
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
@@ -25,8 +25,8 @@
   top: 50%;
   right: var(--space-base);
   z-index: var(--elevation-tooltip);
-  width: calc(var(--space-small) * 3);
-  height: calc(var(--space-small) * 3);
+  width: var(--space-large);
+  height: var(--space-large);
   padding: 0;
   border: none;
   border-radius: var(--radius-circle);
@@ -52,17 +52,18 @@
   min-height: calc((var(--space-largest) - var(--space-smaller)));
   box-sizing: border-box;
   padding: var(--space-small);
+  border-radius: var(--radius-large);
   color: var(--color-heading);
   font-weight: 500;
   cursor: pointer;
   justify-content: space-between;
   align-items: center;
+  transition: all var(--timing-base);
 }
 
 .selectedOption,
 .option:hover,
 .option:focus {
-  border-radius: var(--radius-large);
   background-color: var(--color-surface--background);
 }
 
@@ -103,11 +104,6 @@
     rgba(255, 255, 255, 1),
     rgba(255, 255, 255, 0)
   );
-}
-
-.selectedOption {
-  border-radius: var(--radius-large);
-  background-color: var(--color-surface--background);
 }
 
 .actions::before {
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
@@ -1,7 +1,8 @@
 .content {
   z-index: var(--elevation-tooltip);
   max-width: var(--popover--width);
-  max-height: 400px;
+  max-height: calc(var(--space-extravagant) * 6.25);
+  min-width: calc(var(--space-extravagant) * 3.75);
   box-shadow: var(--shadow-base);
   margin-top: var(--space-small);
   border: var(--border-base) solid var(--color-border);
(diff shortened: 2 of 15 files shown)


packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -21,16 +22,24 @@
...
   list-style: none;
 }
 
 .option {
-  padding: var(--space-small) var(--space-base);
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: var(--space-small);
+  color: var(--color-blue);
```
chris-at-jobber, line 34: Can we use the semantic value here? 
```suggestion
  color: var(--color-heading);
```

packages/components/src/Combobox/components/ComboboxTrigger/ComboboxTriggerChip/ComboboxTriggerChip.css
```diff
@@ -1,10 +1,11 @@
 .trigger {
   display: inline-flex;
-  gap: var(--space-small);
   padding: var(--space-small) var(--space-base);
   border: var(--border-base) solid rgb(230, 234, 235);
```
chris-at-jobber, line 4: this isn't happening in this PR but where is this random RGB value coming from? Can we use one of our system colors?

packages/components/src/Combobox/components/ComboboxTrigger/ComboboxTriggerChip/ComboboxTriggerChip.tsx
```diff
@@ -17,7 +17,7 @@ export function ComboboxTriggerChip(
         setOpen(!open);
       }}
     >
-      <Typography fontWeight="semiBold" textColor="heading">
+      <Typography fontWeight="semiBold" textColor="blue">
```
chris-at-jobber, line 20: This color shouldn't be changed from the semantic to a base value

packages/components/src/Combobox/components/ComboboxTrigger/ComboboxTriggerChip/ComboboxTriggerChip.css
chris-at-jobber: Looks like `color-border` would work here

chris-at-jobber (commented):
Is it possible to adjust the height of the available scrollable area for options to make overflow more apparent?

**Current - no indication that there are more items**
image

**Edited - more items is hinted by fader overlapping last item**
image

chris-at-jobber (changes requested):
Few small changes, looking really good!

packages/components/src/Combobox/components/ComboboxTrigger/ComboboxTriggerChip/ComboboxTriggerChip.css
taylorvnoj: I feel like I saw this rgb value in the figma at some point but `color-border` does the trick!

taylorvnoj:
Thanks @chris-at-jobber! I made those changes here 3740827348b039ba7d1b38aab73f54a5ccc69149 
Let's wait for the last of the PRs to be merged before we do another review 👍

chris-at-jobber:
Thanks for the updates @taylorvnoj ! Two remaining issues:

1️⃣ Looks like the options are a bit shorter than expected. In the Figma they're equivalent to 44px high, here they're:

- 37px high unselected
- 40px high selected (probably because the icon bumps them up) 

image

Whether you want to address that here, or leave for @ZakaryH as he'll have some CSS wrangling around options to do on his end anyways, I'll leave to you two to sort.

2️⃣ The actions don't have a hover or focus state. They should have the same hover and focus state as a [tertiary button](https://atlantis.getjobber.com/?path=/story/components-actions-button-web--basic&args=type:tertiary), including having a `cursor: pointer` when hovered.

chris-at-jobber:
@taylorvnoj if you set it as a min-height with some padding baked into it you should get the long-label-wrapping starting from the desired minimum. In this example, I used the `calc` on the `min-height`, left the padding as-is, and set `box-sizing: border-box`:

image

taylorvnoj:
Hey @chris-at-jobber - couple of things 😄 

017b619c3df1709cfdc4beb8feac6a90fd51a72d - commit with changes to height of the options. I also adjusted the height of the actions since the figma calls for a `40px` height.   Because of these changes I changed the height of the content once again to show more of the last option in the TriggerButton example, to show that the list continues.

4b7773547bd7f645b66c8b0145c307351c3b2d3a - hover and pointer cursor on actions.  I spoke with Laura about the focus state on Actions and the conclusion was that focus on actions will not have the yellow glow.  Do you want to loop her in and revisit this?

chris-at-jobber:
Changes are mostly good!

Do you have a reference for the `40px` height? In the Figma I'm referencing, the options are `44px` high:
image

I don't have context on the decision to drop the focus state from the actions, so we can revisit that later 👍

taylorvnoj:
@chris-at-jobber Options are 44px, Actions are 40px - both changes were done in that second to last commit

Screenshot 2023-09-22 at 1 41 51 PM

chris-at-jobber:
Ohhhh that's my bad, I conflated `actions` and `options` 😅 carry on! The working software looks great 👌 

Also, just confirmed expected hover and focus states with Laura, they've been added to [the Figma](https://www.figma.com/file/gQNDrpJYRFAwfbP3jRO0an/DSC-%5B-Combobox-%5D?type=design&node-id=2423%3A29916&mode=design&t=POFNRyuNC5xTy5fs-1)

taylorvnoj:
Perfect - thanks @chris-at-jobber I'll make that change here too!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -1,76 +1,93 @@
 .content {
   z-index: var(--elevation-tooltip);
   max-width: var(--popover--width);
-  max-height: 400px;
+  max-height: calc(var(--space-largest) * 8.75);
```
dlrussel, line 4: where are these arbitrary numbers used for these calculations coming from??

taylorvnoj: Based on [this comment](https://github.com/GetJobber/atlantis/pull/1530#pullrequestreview-1636377932) we need to have a max-height that shows _some_ of the next option.  There isn't a clean way to calculate this using only tokens.

Let me know if you have an alternative for the calc. I know we talked about setting 'magic numbers' but I think in this case it's a bit different.  In our `spacing.css` file and our tokens in `foundation.css` we sometimes have to calc with hard numbers.  

tldr; multiplying by 8.75 looks the best 😆

chris-at-jobber: Stylistically, maybe we could define these numbers at the top of the CSS file so it's more clear in-use? Like

`--optionsOverflowHeight: calc(var(--space-largest) * 8.75)`

and then in usage, 

`max-height: var(--optionsOverflowHeight);`

?

taylorvnoj: Looks like we do that in components like `Button` where specific calculations are used multiple times over.

There are two instances in this css file where we use the same calculation twice.  I'm down to set those variables in a `root` at the top of the file - how do you feel about that @dlrussel?

taylorvnoj: Diane and I talked about this over zoom. I don't want to speak for her so please @dlrussel pipe in if I get this wrong - but her concern is more so about using a hardcoded, seemingly random number in the calculation _and_ for the height.  The suggestion to use variables in a `root` won't solve for that (and I am thinking we don't need that in this file).

But we do need to have a `max-height` on the options list.  With Zak's PR now merged, a `260px` max-height looks best in the options list (final decision is with Chris).

Returning to Diane's original question, my reasoning for why the calculation and height is arbitrary is because:
- we want a slight cut off on the (fifth?) option to show the fade; afaik the easiest way to do this is through css rather than our .tsx file
- there is no single token that will get us 260px
- having `calc(var(--space-small) * 32.5)` is in my opinion, faster to calculate and easier to read than trying to divide and subtract a longer list of tokens together to somehow get to `260px`

Like I said to Diane I completely understand her hesitancy and dislike for calculations like this.  Have a think on it - if there's another way to get there let's do it

taylorvnoj:
Future overflow indicator (gradient) work is documented in JOB-79443.

packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
```diff
@@ -8,15 +8,22 @@
...
   padding: var(--space-small) 0;
   border: none;
   text-align: left;
   background-color: var(--color-surface);
+  cursor: pointer;
 }
 
 .actionButton:focus {
+  box-shadow: var(--shadow-focus);
+  border-radius: var(--radius-large);
```
chris-at-jobber, line 21: Couple of really minor state things that would be nice to see addressed as a fast-follow 

- is there a reason we're manipulating the border-radius on a per-state basis? As far as I can tell we could simplify this and the hover state on line25 by setting the border-radius on `.actionButton`
- can we set both the options and actions to have a `transition: all var(--timing-base);` applied to match [Button](https://github.com/GetJobber/atlantis/blob/e46abe091488d7bacf653ac8fd7ffb73b575e8aa/packages/components/src/Button/Button.css#L26C38-L26C38)?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -1,75 +1,93 @@
...
   visibility: hidden;
 }
 
+.clearSearch {
+  display: flex;
+  position: absolute;
+  top: 50%;
+  right: var(--space-base);
+  z-index: var(--elevation-tooltip);
+  width: calc(var(--space-small) * 3);
```
chris-at-jobber, line 28: Fast-follow: `var(--space-small) * 3` is the same output (24px) as `var(--space-large)` so we can do away with these calcs

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -1,75 +1,93 @@
...
   cursor: pointer;
+  justify-content: space-between;
+  align-items: center;
 }
 
-.option:hover {
+.selectedOption,
+.option:hover,
+.option:focus {
+  border-radius: var(--radius-large);
```
chris-at-jobber, line 65: Same as in the actions, 

- is there a reason not to set the border-radius on the `option`, rather than modifying it based on state?
- should have the same transition property applied

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -87,34 +105,34 @@
...
   left: 0;
-  width: 100%;
   height: var(--space-base);
-  background-image: linear-gradient(
-    to bottom,
-    rgba(255, 255, 255, 0),
-    rgba(255, 255, 255, 1)
+  background: linear-gradient(
+    180deg,
+    rgba(255, 255, 255, 0) 0%,
```
chris-at-jobber, line 123: Unrelated and not a fast-follow as it's more of a "should we do this", but can/should we codify this gradient styling? We're using it all over the place in ComboBox and DataList

chris-at-jobber: Note for me: we should also have guidance around when to use the fade

chris-at-jobber (commented):
Design approved! My comments/notes are all non-blocking.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
taylorvnoj: wow Taylor 😵‍💫

packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css
taylorvnoj: No reason, I got too lost in the css sauce. Thank you for pointing that out!

These updates plus the calc mishap are fixed here: 2303facb98cd03322324a4c0273528c94aadba93

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
taylorvnoj: For anyone reading, we had a quick convo about this in Office Hours today.  We'll look to extract this in someway - Chris, would you like me to put some sort of a ticket for this consideration in the backlog?

chris-at-jobber: I think that makes sense, thanks!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -1,75 +1,94 @@
...
- */
+* so please be careful if you change this class name, or the css rules within.
+*/
 .content.hidden {
   display: none;
   visibility: hidden;
 }
 
+.clearSearch {
+  display: flex;
```
dlrussel, line 23: Ok so we can remove this line

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -1,75 +1,94 @@
...
+  right: var(--space-base);
+  z-index: var(--elevation-tooltip);
+  width: var(--space-large);
+  height: var(--space-large);
+  padding: 0;
+  border: none;
+  border-radius: var(--radius-circle);
+  background-color: var(--color-surface--background);
+  transform: translateY(-50%);
+  justify-content: center;
```
dlrussel, line 35: we can also remove this line and the one below

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -1,75 +1,94 @@
...
-  * so please be careful if you change this class name, or the css rules within.
- */
+* so please be careful if you change this class name, or the css rules within.
+*/
 .content.hidden {
   display: none;
   visibility: hidden;
 }
 
+.clearSearch {
```
dlrussel, line 22: Can we add  ` cursor: pointer;` to this style block 👆🏻 It was historically only used when hovering over links but is now frequently used to indicate a clickable element, such as a button!

dlrussel (approved):
A few small comments but this looks really good ✅

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
taylorvnoj: Oh great catch! Thank you!

PR #1532, opened 2023-09-13
feat(components): Combobox Accessibility

ZakaryH:
## Motivations

We should optimize for, or at least offer a comparable experience as far as quality for users interacting with the combobox with keyboard only, or a screen reader.

### Added

**Arrow Key Navigation**
the options are no longer tab-able. to traverse them you use the arrow keys to focus (but not select!) the item. then to select them, you use enter or space.

**Scrolling to Selected**
if you have a list that is big enough to scroll, and then select something far enough in the list that requires scrolling. when you re-open the combobox that selected item should be in view and there shouldn't be an obvious scroll animation happening.

a side effect of the changes to achieve this was that the list is now the scrollable area with overflow: scroll, not the entire container. also, the search input is no longer sticky. it can just be a regular ol' div in the standard document flow. this makes the scrolling calculations much simpler.

**Voice Over Experience**

the combobox button should be read out and identified as a combobox that has a listbox that can be opened. then once it's open, it should read out the elements all nice like and when you traverse the options it should indicate if you are on a selected option or not!

**Focus Trapping**

Per accessibility best practices with dialogs/modals - we should "trap" the focus so that you can't accidentally tab out of the dialog you've opened. It will loop infinitely within the open dialog until you close it either by pressing ESC, making a selection or choosing an action. On that last one, we haven't fully decided if it will close on action activation. I don't think it does that right now.

**Refocusing on Activator**

When you do close the combobox, we want to put the focus back where you were when you initially started your combobox adventure.

<details>
  <summary>Removed: Skipping to Actions</summary>

**Skipping to Actions**

While playing around with this, with like 20 items in the list it was super annoying to have to tab through each and every single one of them when all I wanted was to get to the actions. Though now that we have the focus trapping it's an option to shift-tab backwards from the top and quickly get to them, so honestly this might not be necessary anymore. 

I kinda want to scrap it now because I'm going to have to bug a designer about how it should look, along with the code it adds - I'm just not convinced it's worth it now. **Looking for feedback on this.**

Kapture 2023-09-12 at 22 45 40
</details>

## Testing

Using a screen reader or your keyboard, interact with the combobox. It should feel good navigating around with tabs, shift-tabs, arrow down, arrow up, ESC, return, and spacebar presses.

Focus should remain trapped in the open combobox. 

Focus should return to the button/chip that you opened it with to begin with, also this particular bit won't be exclusive to keyboard usage. Even clicking outside to close the combobox should exhibit the same behavior.

Screen readers should announce that the trigger chip and button are combobox elements, announce that there is a search input on open, announce the listbox of options, announce which one is selected and that the rest are not, announce the actions as a group separate from the listbox

There's a known deficiency here where each of the options well be read off as "text element" rather than something interactive. I'm not sure how to get around this honestly. Even Polaris' implementation does the same thing. I feel like it's good enough though since it's announced as part of a listbox and each item is an option.

---

Random photo of Atlantis


chris-at-jobber:
@ZakaryH one immediate concern is that the combobox itself doesn't _appear_ to have a role or be announced effectively. When I activate the combobox from the trigger element (the button in this case), what's announced is:

image

I'm not entirely sure if this would be universally applicable, but perhaps auto-focusing the search input, _then_ having the skip to actions, might be more appropriate?

chris-at-jobber:
I'm very fresh to this project but I'm not seeing any elements in the assorted ComoboBox components labeled with the [combobox](JOB-75937/JOB-77630/accessibility-stuff) role - was that a consideration at any point?

ZakaryH:
@chris-at-jobber good catches!

yeah Taylor also suggested the focus-search-on-open, so I'll be adding that shortly

unless you really love the skip-to-actions, I think I'm gonna remove it. now that we have the focus trapping and you can easily shift tab backwards to quickly get to the actions, I think that's a much simpler way to achieve the same goal

oh interesting. I had no idea that `combobox` was such a widespread or commonly used term to the point that it's a bonafide aria role. that absolutely makes sense to incorporate into our combobox

ZakaryH:
@chris-at-jobber k this is looking/sounding much better with that combobox role, great suggestion!

image

also, are you our resident accessibility expert? 

I feel like this is at a place I'm happy with now, minus the whole radio vs button thing we have in progress, so I wouldn't mind getting at least a partial review to speed things up - but I'm not sure if there's a best person to ask on the team

chris-at-jobber:
That preview looks good, I'll take a pass before EOD!

chris-at-jobber (commented):
Design approved! I know we're developing the options separately but the overall experience feels significantly better than the first pass, thanks for your work on this @ZakaryH

ZakaryH:
things are getting interesting trying to do the arrow keys to navigate, because you wanna do a few things
- disable tabbing
- allow arrows to focus without selecting
- allow scrolling with arrows when you approach the max height of the list and there are more elements

it's 100% possible, in fact I can see some examples of how we're doing this in Autocomplete - so I have no doubts we can achieve this

however it's going to be more work. at this point I'd like to follow up with that functionality if that's alright with everyone. lemme know what you think though! is this fine to use tabs for now, or do we really want arrows to be the way to navigate this for v1?

chris-at-jobber:
> do we really want arrows to be the way to navigate this for v1

If by v1 you mean "the first user-facing-in-Jobber implementation", then yes we need to have the arrow navigation for that.

If by v1 you mean "the first version of this that we merge into Atlantis master", then I think it's OK for now.

ZakaryH:
@chris-at-jobber that's fair. using arrow keys does feel a lot better

I was more thinking it would be an improvement along the way to multiselect. though I can't guarantee when that will be worked on, so I understand why it might be best to add it now. I'll try and just do that.

ZakaryH:
with the last batch of changes here's where we are at:

**Arrow navigation**
you may no longer tab to get to the options. instead, you will use and and down arrows to highlight or focus (not select!) and then use `enter` or `space` to select that item

this should work when there's a small list

this should work when there's a big list, and you must scroll. while moving down the list it should scroll with you, keeping your focused item in view

this should work when you use the filter to reduce the list to a smaller subset, it should go to the first position and not get lost or anything

this should work with screen readers, where it will not only allow you to navigate with arrows, but also have it read out the selected or not selected state

this aligns with parts of this guide that I have conveniently decided to follow. please don't ask me about the rest
https://get-evinced.com/blog/creating-accessible-combo-boxes/

**Selection on Enter/space**

this should not interfere with the other actions in the combobox such as the actions at the bottom if they are provided, nor  the little X to clear a search input. you should always be able to interact with those both before and after having moved focus into the listbox with the arrow keys


**Known issues**
I need to, or at least want to, deal with the fact that you can have this scenario
image

where one item is selected, and that will have its own proper style in the other PR
one item is focused with the arrow navigation
one item is in the hover state

there should only be one selected, and then one hover OR focused. I have an idea for how to do this, but if you think this is OK for now I am not going to argue

chris-at-jobber:
With the work Taylor's doing on the selected styling, I'm not _overly_ concerned about the state overlap. The quickest fix I'd look at would be adding a real focus indicator with our [focus shadow](https://job-75937-job-77630-accessib.atlantis.pages.dev/?path=/docs/design-colors--page#focus), but that could be a followup IMO.

chris-at-jobber:
@ZakaryH What I'm seeing on the build is that there's actually no focus state - in this video, once I turn VoiceOver on (roughly the 00:20 mark) you can see that I am in fact successfully navigating the options.

https://github.com/GetJobber/atlantis/assets/39704901/542eb416-f376-4816-b138-29cfc858b10a

I can see it in your examples, so I'm wondering if maybe I'm somehow looking at an outdated build - but I did _just_ open this PR up and open the cloudflare build.

ZakaryH:
@chris-at-jobber whoopsy! 😅  yeah you're right

I have a local change that I didn't push up because Taylor is doing the styling for real in the other PR

```
.option:focus {
}
```

I added this for myself so I could see the focus changing, because as you've noticed - without it it's just using your imagination

chris-at-jobber:
So for the purposes of this PR just review the keyboard + assistive tech functionality, visuals to come in Taylor's PR? In that case, 👌

ZakaryH:
@chris-at-jobber correct! but honestly it's going to be tough to see if it's doing the right thing without the colors. there's quite a few scenarios I had to fix where the focus wasn't where I wanted it to be, and I'd like to see how you feel about where the focus goes after you close it after making a selection, or filtering/unfiltering and then move the arrows again

so I can just push up a commit with that color to make it easier to QA and play around with, then remove it after OR you can make the change locally if your dev environment is good to go. either works for me!

chris-at-jobber:
One bit that's a little funky:

1. Select option lower in the list (scrolled to select)
2. Combobox closes
3. Re-open combobox
4. Option you selected is in view, focused (nice)
5. Press `up` arrow
6. First option is selected

I was a bit surprised it didn't go up to the option above my selection.

Video of me walking through this:

https://github.com/GetJobber/atlantis/assets/39704901/74658309-0152-4f95-9851-b2faedf60b62

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -134,10 +154,19 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
...
   placeholder?: string;
   searchValue: string;
+  open: boolean;
   setSearchValue: Dispatch<SetStateAction<string>>;
 }): JSX.Element {
   const searchRef = useRef<HTMLInputElement>(null);
 
+  useEffect(() => {
+    if (props.open) {
+      setTimeout(() => {
```
ZakaryH, line 164: required to focus the search input on open. you cannot `focus()` properly on something that has the css property `visibility: hidden`

what seems to be happening is a race condition where we trigger the open, focus the search and then add the css class that changes the visibility

adding this timeout of 0 moves this operation down to the bottom of the call stack or whatever it's called for the JS event loop, giving it enough time to do its thing

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -134,10 +154,19 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
...
+        searchRef.current?.focus();
+      }, 0);
```
ZakaryH, line 166: required to focus the search input on open. you cannot focus() properly on something that has the css property visibility: hidden

what seems to be happening is a race condition where we trigger the open, focus the search and then add the css class that changes the visibility

adding this timeout of 0 moves this operation down to the bottom of the call stack or whatever it's called for the JS event loop, giving it enough time to do its thing

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -184,6 +213,84 @@ function getZeroIndexStateText(subjectNoun?: string) {
...
+  open: boolean,
+  selectionCallback: (selection: ComboboxOption) => void,
+  containerRef: React.RefObject<HTMLDivElement>,
+  filteredOptions: ComboboxOption[],
+  optionsListRef: React.RefObject<HTMLUListElement>,
+): void {
+  const hasOptionsVisible = open && filteredOptions.length > 0;
+  const [focusedIndex, setFocusedIndex] = useState<number | null>(null);
+
+  useEffect(() => setFocusedIndex(null), [open, filteredOptions.length]);
```
ZakaryH, line 226: we want to discard any previous arrow presses that might have set a focused index because with a new set of filteredOptions, there may be a new option in the first position and we wanna start fresh from `null`. same idea for when it closes. discard any index changes we made.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -184,6 +213,84 @@ function getZeroIndexStateText(subjectNoun?: string) {
...
+      containerRef.current?.addEventListener("keydown", handleContentKeydown);
+    }
+
+    return () => {
+      containerRef.current?.removeEventListener(
+        "keydown",
+        handleContentKeydown,
+      );
+    };
+  }, [open, optionsListRef, focusedIndex, filteredOptions]);
```
ZakaryH, line 239: this is a bit aggressive, but what happens without these dependencies is that the handler added to the event gets stale. it's almost a snapshot of the state when we attached it, but we need up to date versions of these vars when doing the logic in `handleContentKeydown`

I feel like there's gotta be a better way to do this but I'm blanking at the moment

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -34,9 +34,12 @@
   background-color: var(--color-surface--background);
 }
 
+.option:focus {
+  background-color: rgb(34, 153, 221);
```
ZakaryH, line 38: reminder: delete this before merging. only have it so that I can easily see the focus moving around.

ZakaryH:
@chris-at-jobber ok, those changes we did in the call have been applied. I committed the focus state for easier testing but I'll remove it before merging

didn't fuss with the optimal arrow key behavior after a selection, nor the double hover/focus state as we discussed

I'm pretty sure that's everything now! 🤞

chris-at-jobber (commented):
Design approved! Great work.

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -251,12 +263,13 @@ describe("Combobox selected value", () => {
   it("has no selected option when a null selected value is passed", () => {
     const { getByText } = render(<ClearSelectionCombobox />);
 
-    fireEvent.click(getByText("Button"));
-
     const option = getByText("Bilbo Baggins");
-    expect(option).toHaveClass("selectedOption");
     const clearButton = getByText("Clear Selection");
+
+    expect(option).toHaveClass("selectedOption");
```
ZakaryH, line 269: had to shuffle around the order of interactions here

this `expect` where we check the class only works because the open state is being managed with visibility right now. if this isn't rendered it'll stop working which is perhaps not ideal

what was causing this test to fail was that we only opened the dialog to check, then cleared but never closed the dialog, which means we depend on some re-renders as `open` changes to get the correct values.  

you wouldn't be able to do this in real life. your tabs & focused would either be trapped in the dialog, or you click outside and it closes

well actually now that I say it out loud, you could do this if you gave the combobox an option to clear the selection which I think would be super weird and not how this component should be used at all. even still, you could do that - and you'd probably hit the bug this test was hitting

ZakaryH: nevermind, even though that's a weird thing to do it works without causing a bug

image

I'm willing to chalk it up to the test environment behaving differently than real life 🤷

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -203,7 +323,10 @@ function useComboboxContent(options: ComboboxOption[]): {
     }
   }, "Escape");
 
-  const popperRef = useRef<HTMLDivElement>(null);
+  useRefocusOnActivator(open);
```
ZakaryH, line 326: so I don't love having this in here. it's arguably a lot more accessibility related than it is generic behavior of Combobox. I tried moving it into the `useComboboxAccessibility` but...

the problem is that if I call this **after** `useFocusTrap` then the wrong `document.activeElement` gets used as you can see here 

Components___Combobox___Combobox___Web_-_Clear_Selection_⋅_Storybook

and then it just doesn't work because the focus doesn't return to the activator it goes into the now closed dialog :| 

now I'm sure there's a way to make this order independent, but I'm not sure I want to fix that right now - so lemme know what you think about this setup. I could move all this `popperRef` and `useFocusTrap` into the other hook

same with the `useOnKeyDown` honestly too. that's accessibility focused.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
dlrussel: This is an awesome solution! 💡

dlrussel:
@ZakaryH is the expected behaviour that when you tab out of the search, the focus goes to the actions instead of the options?

dlrussel:
Another thing to consider here is that the green text colour that we are using for our action links fails the color contrast guidelines for accessibility when used on a white background. I know that design pattern is something we use in many other places in jobber but moving forward it may not be something we want to continue to implement. cc @chris-at-jobber

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -50,47 +53,70 @@ interface ComboboxContentProps {
...
     <div
       ref={popperRef}
+      id="ATL-Combobox-Content"
       data-testid="ATL-Combobox-Content"
+      tabIndex={0}
       className={classnames(styles.content, !open && styles.hidden)}
       style={popperStyles.popper}
       {...attributes.popper}
     >
       <Search
```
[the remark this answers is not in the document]
dlrussel: Should we perhaps add a label and visually hide it?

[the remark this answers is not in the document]
ZakaryH: so if you don't provide the `subjectNoun` prop it'll default to `Search` as the placeholder which does get read out by VoiceOver. on top of that since we're using `type="search"` on the input it also calls it out as a `search text field`

then, if you do provide a `subjectNoun` the placeholder becomes even more specific or contextual eg.

image

personally I feel like that's enough? to me the only thing missing is perhaps a description of what happens when you search, but that's going to get pretty verbose. if someone was already familiar with comboboxes, it's probably implied how the search would be used, but I have doubts about how many people think about comboboxes

dlrussel (approved):
Looks good ✅

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
ZakaryH: DELETED

PR #1533, opened 2023-09-13
feat(components): Add sorting to DataList

rodrigoeidelvein:
## Motivations

Adds sorting to `DataList`. The consumer can define a initial state for the sorting, a callback and which columns can be sorted.

### Added

- Added a prop called `sorting` to DataList, where the consumer will have access to the `onSort` callback, state and can select which columns are sortable.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/DataList/context/DataListSortContext/DataListSortContext.tsx
```diff
@@ -0,0 +1,56 @@
...
+  children: React.ReactNode;
+}) {
+  const { sorting: sortingProp } = useDataListContext();
+
+  const [sortingState, setSortingState] = useState<Sorting>(
+    sortingProp?.initialState || {
+      direction: "none",
+      key: "",
+    },
+  );
```
darryltec, line 32: best practice on this is to never track the state of a component that could be updated outside of it. You risk making the component does something different than what is in the main state that the consumer uses.

Also easy to spot it of it's broken. If you click it and it does nothing, you, the consumer hasn't implemented it properly. Compared to it updating the state internally but actually not the main state outside of it. 

In short, keep it to be one source of truth instead of 2

docs/components/DataList/Web.stories.tsx
```diff
@@ -95,6 +95,14 @@ const Template: ComponentStory<typeof DataList> = args => {
...
       }}
       onLoadMore={nextPage}
+      sorting={{
+        initialState: {
+          key: "label",
+          direction: "desc",
+        },
+        onSort: sorting => console.log(sorting),
+        sortable: ["label", "home"],
+      }}
```
darryltec, line 105: This is why I like seeing the component signature first before writing any code.

The fact that you have a prop called `initialState` made me assume that we are tracking a state internally (we are) and that we shouldn't because you risk having the internal and external state being out of sync and causing a lot of headaches for both user and dev.

Also, rule of thumb, the moment you name a prop "initial" in Atlantis, that should be an indicator that you are doing something that's not right.

rodrigoeidelvein: Should I just keep the `sortable` and `onSort` options then? I think it would be useful to have a way to define a initial state for the sorting, I can think of someone wanting to have a "created" column sorted by default, for example.

darryltec: Let's say I have a button somewhere on the page that the DataList doesn't know about. Say it's a feature that saves your filters. Clicking the button does a bunch of like set a filter, search, and also includes updating the state where I determine what sorting the data list should have.

- The user loads the page with the data list
- The DataList renders and it says the initial sorting is nothing
- The user clicks "give me my saved filter"
- How would the data list know you want it to have a different sort?

rodrigoeidelvein: Yeah, I see. In this case, the consumer could set the desired sorting as to the `initialState` and it would update it. Maybe it is not "initial", so we can change the prop name to just "state", or have a way of updating the state from the outside? 🤔

darryltec: yup. states that are configurable by the user and dev should always be outside of the reusable component. We shouldn't track that state, the component should react* to the prop you give it.

_* this explains why react is called react because they want you to write reactive components

chris-at-jobber:
@rodrigoeidelvein does this need a design review, or still getting feedback on the code?

docs/components/DataList/Web.stories.tsx
```diff
@@ -95,6 +104,14 @@ const Template: ComponentStory<typeof DataList> = args => {
         created: "Created",
       }}
       onLoadMore={nextPage}
+      sorting={{
+        state: sortingState,
+        onSort: sorting => {
+          console.log(sorting);
+          setSortingState(sorting);
+        },
```
darryltec, line 112: do you still need the `console.log`? The setSortingState kinda tells you that it is working no?

packages/components/src/DataList/DataList.test.tsx
darryltec, line 1: - should probably test if the `state` updates the UI when you change the value.
- should also test if the onSort fires with the correct params

packages/components/src/DataList/DataList.types.ts
```diff
@@ -44,6 +44,11 @@ export type DataListHeader<T extends DataListObject> = {
   readonly [K in keyof T]?: string;
 };
 
+export interface DataListSorting {
+  readonly key: string;
```
darryltec, line 48: have you tried narrowing this down to whatever keys is available on the data/header?

packages/components/src/DataList/DataList.types.ts
```diff
@@ -97,6 +102,19 @@ export interface DataListProps<T extends DataListObject> {
...
+
+  readonly onSortingChange?: (sorting: DataListSorting) => void;
+
+  /**
+   * `sortable`: List of keys that are sortable.
+   * `initialState`: The initial state of the sorting.
+   * `onSort`: The callback function when the user sorting a column.
+   */
+  readonly sorting?: {
+    sortable: string[];
```
darryltec, line 114: same here, i think you can narrow this down to whatever keys is available on the data/header?

packages/components/src/DataList/DataList.types.ts
```diff
@@ -97,6 +102,19 @@ export interface DataListProps<T extends DataListObject> {
...
+    state?: DataListSorting;
+    onSort: (sorting: DataListSorting) => void;
```
darryltec, line 116: readonly please

packages/components/src/DataList/DataList.types.ts
```diff
@@ -44,6 +44,11 @@ export type DataListHeader<T extends DataListObject> = {
   readonly [K in keyof T]?: string;
 };
 
+export interface DataListSorting {
+  readonly key: string;
+  readonly direction: "asc" | "desc" | "none";
```
darryltec, line 49: what's the `none` for?

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -89,13 +89,7 @@ export function generateHeaderElements<T extends DataListObject>(
...
     (acc, key) => ({
       ...acc,
-      [key]: (
-        <div className={styles.headerLabel}>
-          <Text variation="subdued" maxLines="single" size="small">
-            {headers[key]}
-          </Text>
-        </div>
-      ),
+      [key]: <DataListHeaderTile headers={headers} headerKey={key} />,
```
darryltec, line 92: can grab that `headers` prop from the context, no?

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const isSortable = sorting?.sortable.includes(headerKey);
+  const sortingState = sorting?.state;
+
+  function toggleSorting(sortingKey: string) {
+    if (sortingKey === sortingState?.key) {
+      sorting?.onSort({
+        direction: sortingState.direction === "asc" ? "desc" : "asc",
+        key: sortingKey,
+      });
```
darryltec, line 28: i wonder if you remove the rest and keep this, would it still work the same?

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -0,0 +1,72 @@
...
+    } else {
+      sorting?.onSort({ direction: "asc", key: sortingKey });
+    }
+  }
+
+  return (
+    <div
```
darryltec, line 35: shouldn't use div on anything interactable as this would only work for mouse and tap but not accessibility

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -0,0 +1,72 @@
...
+      className={classnames(styles.headerLabel, {
+        [styles.sortable]: isSortable,
+      })}
+      onClick={isSortable ? () => toggleSorting(headerKey) : undefined}
```
darryltec, line 39: don't know how to feel about that ternary 🤔

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+function SortingArrows({ order }: { order: "asc" | "desc" | "none" }) {
+  if (order === "none") return <></>;
+
+  return (
+    <div>
+      <span className={styles.arrowUp}>
+        <Icon
+          name="arrowUp"
+          color={order === "asc" ? "greyBlue" : "blue"}
```
darryltec, line 59: since this and line 66 pretty much have the same logic except for the value, can we simplify this and avoid that ternary?

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -0,0 +1,72 @@
...
+import styles from "./DataListHeaderTile.css";
+import { DataListHeader, DataListObject } from "../../DataList.types";
+import { Text } from "../../../Text";
+import { Icon } from "../../../Icon";
+import { useDataListContext } from "../../context/DataListContext";
+
+interface DataListHeaderTileProps<T extends DataListObject> {
+  headers: DataListHeader<T>;
+  headerKey: string;
+}
```
darryltec, line 12: today on

> I wonder what Darryl would say about this types

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -0,0 +1,72 @@
...
+        {headers[headerKey]}
+      </Text>
+      {sortingState?.key === headerKey && (
+        <SortingArrows order={sortingState.direction} />
+      )}
+    </div>
+  );
+}
+
+function SortingArrows({ order }: { order: "asc" | "desc" | "none" }) {
```
darryltec, line 51: probs should be on its own file. could also just be on the same folder

darryltec (changes requested):
haven't tested, but I noticed a few things

docs/components/DataList/Web.stories.tsx
rodrigoeidelvein: Yes, but it's a way of showing the results, as we do for `onSearch`

packages/components/src/DataList/DataList.types.ts
rodrigoeidelvein: For when the consumer wants to reset the sorting direction, but now thinking better, just setting the key to an empty string should work too

packages/components/src/DataList/DataList.utils.tsx
rodrigoeidelvein: oh yeah, good catch!

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
rodrigoeidelvein: I don't know... maybe I could extract the colors to constants called `activeColor` and `inactivateColor`, but the ternary is very straightforward and works well for inverting the colors depending on the order. Do you have any suggestions?

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
rodrigoeidelvein: What element could be usable here? Would a `span` be accessible in this case?

chris-at-jobber:
Referencing the [Figma](https://www.figma.com/file/TuoXq9l7Mg1Xfusj6qnx31/%5BJobber-2025-System%5D-List-Views?type=design&node-id=1822%3A120110&mode=design&t=9cz65z2CHYAVhrPv-1) the sort icon is out of alignment:
image

Also in comparison to the Figma

- the headers are using `secondary` color when the expected is `text`
- headers are using `small` text when the expected is `base`

chris-at-jobber:
For small screens, it [looks like](https://www.figma.com/file/TuoXq9l7Mg1Xfusj6qnx31/%5BJobber-2025-System%5D-List-Views?type=design&node-id=1822%3A120517&mode=design&t=9cz65z2CHYAVhrPv-1) the plan is to render a "sort" action in the datalist actions barm for small devices. Is that being implemented in a separate PR?

rodrigoeidelvein:
> For small screens, it [looks like](https://www.figma.com/file/TuoXq9l7Mg1Xfusj6qnx31/%5BJobber-2025-System%5D-List-Views?type=design&node-id=1822%3A120517&mode=design&t=9cz65z2CHYAVhrPv-1) the plan is to render a "sort" action in the datalist actions bar. Is that being implemented in a separate PR?

Yes, it will be on a separate PR

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
rodrigoeidelvein: lol it does work 🤦 thanks for catching it

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
rodrigoeidelvein: Extracted it to a handler function

packages/components/src/DataList/DataList.types.ts
rodrigoeidelvein: I narrowed it down, it gets only the headers' keys, but the `value` of `DataListContext.Provider` is complaining about it

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
rodrigoeidelvein: Added a button to sortable columns

docs/components/DataList/Web.stories.tsx
```diff
@@ -60,6 +64,13 @@ const Template: ComponentStory<typeof DataList> = args => {
     },
   });
 
+  const [sortingState, setSortingState] = useState<DataListSorting | undefined>(
+    {
+      key: "label",
+      direction: "asc",
+    },
+  );
```
darryltec, line 72: maybe we should start undefined on this one since this actually does nothing to the example right now as the test API we use doesn't support that.

packages/components/src/DataList/DataList.types.ts
darryltec: @rodrigoeidelvein , forgot to update this to be `Array<keyof DataListHeader<T>>` as well.

Also, maybe we should use `(keyof DataListHeader<T>)[]` here and on line 112 for consistency

packages/components/src/DataList/__tests__/DataList.utils.test.tsx
```diff
@@ -129,9 +129,11 @@ describe("Datalist utils", () => {
       const headerElements = generateHeaderElements({ name });
       render(headerElements?.name || <></>);
 
+      screen.logTestingPlaygroundURL();
```
darryltec, line 132: don't need this

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.css
```diff
@@ -0,0 +1,18 @@
+.headerLabel {
+  display: flex;
+  align-items: center;
+}
+
+button.headerLabel {
```
darryltec, line 6: i think we can just put this on the other one too. No need to differentiate between them as they should have the same styling regardless

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
darryltec: @rodrigoeidelvein add readonly please. Not only for consistency, but also for being semantically correct as props are readonly and we should set them as such

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -0,0 +1,56 @@
...
+      direction: sortingState?.direction === "asc" ? "desc" : "asc",
+      key: sortingKey,
+    });
+  }
+
+  function handleOnClick() {
+    if (!isSortable) return;
+
+    toggleSorting(headerKey);
+  }
```
darryltec, line 39: Consistency-wise, drop these boys down below return. Pretty minor thing but reading the JSX first makes it easier to understand what this things are for

packages/components/src/DataList/components/DataListHeaderTile/SortingArrows.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import styles from "./SortingArrows.css";
+import { Icon } from "../../../Icon";
+
+export function SortingArrows({ order }: { order: "asc" | "desc" }) {
+  const activeColor = "greyBlue";
```
darryltec, line 6: should be grey blue lighter. If that doesn't exist on the icon color, make it support it.

packages/components/src/DataList/components/DataListHeaderTile/SortingArrows.css
darryltec, line 1: sneaky!

packages/components/src/DataList/context/DataListContext/DataListContext.tsx
```diff
@@ -7,6 +7,13 @@ export const defaultValues = {
   headers: {},
   children: [],
   selected: [],
+  sorting: {
+    sortable: [],
+    state: undefined,
+    onSort: () => {
+      return;
+    },
+  },
```
darryltec, line 16: sorting is optional, don't need to define this here

darryltec (approved):
Those comments I have are not show-stopping so I we could add those later on

packages/components/src/DataList/components/DataListHeaderTile/SortingArrows.tsx
rodrigoeidelvein: Good catch, added it!

rodrigoeidelvein:
@darryltec I addressed most of your comments, only the types of `key` on `DataListSorting` that I kept out. Would we have to export `DataListObject` too? Typescript started to complain on my tests and on `Web.stories.tsx` that it needs an argument to the interface

darryltec:
@rodrigoeidelvein oh yeah. Good point! We should leave it then.

packages/components/src/DataList/DataList.types.ts
darryltec: After Rodrigo's last comment. this actually makes sense as a string. Leaving it as is

darryltec (approved):
Still LGTM! Will leave the final approval to @chris-at-jobber and let him merge it when it's all good!

chris-at-jobber:
Question on the interaction for displaying the sort icon. 

In [DataTable](https://job-76897-job-78448-add-data.atlantis.pages.dev/?path=/story/components-lists-and-tables-datatable-web--manual-sorting) 
- we show the icon at all times, with a third "unsorted" state. Here, we only show the sort icon once you've already sorted a column
- the entire header is clickable to trigger sort. Here, only the label is interactive
- the icon is slightly larger in DataTable (see screenshot)
image


I'm wondering if these discrepancies are intentional?

rodrigoeidelvein:
> Question on the interaction for displaying the sort icon.
> 
> In [DataTable](https://job-76897-job-78448-add-data.atlantis.pages.dev/?path=/story/components-lists-and-tables-datatable-web--manual-sorting)
> 
> * we show the icon at all times, with a third "unsorted" state. Here, we only show the sort icon once you've already sorted a column
> * the entire header is clickable to trigger sort. Here, only the label is interactive
> * the icon is slightly larger in DataTable (see screenshot)
>   image
> 
> I'm wondering if these discrepancies are intentional?

@chris-at-jobber, I just looked into [DataTable's code ](https://github.com/GetJobber/atlantis/blob/259b11a82ef7f36c15f854a49aa52e8a9c72a295/packages/components/src/DataTable/SortIcon.tsx)and it's not using Atlantis' Icon component, that's why it's different.

About the entire header being clickable, DataTable has a hover state that changes the whole header color, DataList doesn't have that, I think it would be weird to have a pointer in a blank space without telling the user what is currently "active". But that's up to implementation, we could make the whole header element selectable on DataList too

darryltec:
@chris-at-jobber you coll with this going out and then updating the design when that's decided? The component is in alpha mode so drastic changes are okay to happen in a minor version

chris-at-jobber:
I'm ok with leaving the interaction states as-is and revisiting, but can we use the same SVG for the sorting here as we do in DataTable?

darryltec:
@chris-at-jobber Yup! I'll change it on my other PR here for the small screen sorting

PR #1534, opened 2023-09-14
docs(components): Add Combobox docs - Alpha

taylorvnoj:
## Motivations

Update the docs page for `Combobox`.

Note: UI changes are coming in another PR so please focus the review on the Docs page content.

### Added

Combobox is now labelled `Alpha` and a banner was added to the docs page to explain this component should be used with caution.

Completed sections for `Combobox` docs:
- Design & usage guidelines
- Content guidelines
- Accessibility 

### Removed

The top level `index.ts` is removed because we don't want Combobox to be used anywhere yet.

---

Random photo of Atlantis


docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -1,92 +1,94 @@
...
+searching through data.
 
 ## Design & usage guidelines
 
-The Combobox will... _Add a brief description of the design and usage of the
-component_
+The `Combobox` component's primary function is to facilitate option selection
+and searching within a list of items. It should be used in scenarios where users
+need to choose from a set of options, with the added benefit of filtering and
+customization.
```
dlrussel, line 18: I think using the word customizations here is a little too vague. I think we should be clear about what types of customizations we are talking about and probably mention something about actions.

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -1,92 +1,94 @@
...
-  the provided content diverges from what is expected (image aspect ratio, amount of text, video
-  vs audio, filesize)? Are there any important references to the [Product Vocabulary](https://atlantis.getjobber.com/guides/product-vocabulary)
-  that should be made regarding this component? What happens when content is unavailable or unreliable
-  (poor connection, empty states, flaky GPS data?)
--->
+#### Trigger
 
-## Related components
+The trigger is the user-facing interface to open and close the `Combobox` and
+initiate option selection. The choice between them depends on the design and
```
dlrussel, line 27: What do you think about mentioning the 2 different trigger options that are currently available. I like the idea of making that information clear here. Thoughts?

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -1,92 +1,94 @@
...
+user experience requirements.
 
-<!--
-  What are some related components in Atlantis?
--->
+There are two primary options for trigger elements of the `Combobox`,
```
dlrussel, line 30: I think we should move this up because it's arguably the most important information IMO but I also don't think we should use the word `primary` here since it makes me interpret that in a way that leads me to believe there are perhaps `secondary` options . 🤔

docs/components/Combobox/Combobox.stories.mdx
taylorvnoj: Yup that is a little vague - what about something like this: 

```
## Design & usage guidelines

The `Combobox` component's primary function is to facilitate option selection
and searching within a list of items.

It should be used in scenarios where users need to choose from a set of options,
with the added benefit of filtering.

The `Combobox` also has flexibility to allow for custom actions, such as adding
a new item to the list of options or managing a selection.
```

docs/components/Combobox/Combobox.stories.mdx
taylorvnoj: This is right below:

```
There are two primary options for trigger elements of the Combobox, TriggerButton and TriggerChip. You must have one trigger component.
```

I was planning on linking each web example - so where it says TriggerButton is a link to that web example.

docs/components/Combobox/Combobox.stories.mdx
taylorvnoj: By move up do you mean within this subheading of `Trigger` or higher in the doc?

Within this subheading of `Trigger` I think the order of the two sentences makes sense. First sentence is defining what Trigger means and the second is letting the reader know there are two types of Triggers.

Agree with the word primary! Let me find an alternative way to word this...

docs/components/Combobox/Combobox.stories.mdx
dlrussel: love that ✅

docs/components/Combobox/Combobox.stories.mdx
dlrussel: ok makes sense! Sounds good to me!

dlrussel (approved):
🚢

PR #1536, opened 2023-09-14
feat(components): Add optional role to Button

ZakaryH:
## Motivations

Short version is that accessibility guidelines for Combobox wants the trigger element to have a `role=combobox`

>The combobox role is set on input that controls another element, such as a listbox or grid, that can dynamically pop up to help the user set the value of the input.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

here are a couple examples, one as a button
image

one as an input
image

so in our Combobox implementation this needs to go on our Trigger(s), but it needs to be on the literal `button` itself, not only something wrapping them.

for the `TriggerChip` that's fine since we're using a `button` but for `TriggerButton` that's using an Atlantis `Button` and thus we need to add an optional `role` prop to it which is where things got interesting

**Longer Explanation**
I initially added a default for `role` to be `"button"` but that was not desirable because it would add `role="button"` to elements that ended up being rendered as `a` tags for the  react-router `Link`, `a` tags for normal links, and `submit` type buttons which is fine, but redundant because it already had an implicit `button` role

from there I removed the default and considered if we should even let the `role` be set at all on these other button types. for the `href` or `url` type buttons, I do not think these should support setting some other role. I think the implicit `link` role is preferred

for the react routing links, I'm on the fence. they do kinda behave as buttons, and maybe we should be setting a role on them - however I'm not sure those make sense to accept a prop. if it's being used as a routing button, it's not a combobox and so you would never need to override it with this prop

for submit buttons, same idea. I can't imagine using a submit button to launch a combobox and so I've decided it doesn't make sense to be a prop for those either.

now, due to how the types are set up here with `XOR`s and `&`s I had limited options of how to implement this in such a way that only a small subset of the buttons would have `role` as a supported prop.

### Added

added an optional `role` prop to a subset of button types. this will allow for us to do `role="combobox"` on a `Button` that isn't a submit button, isn't a `url` button nor a `to` button

## Testing

try adding a button somewhere that has the following, which is what we'll need for Combobox's TriggerButton
```
  <Button
    label="booton"
    onClick={() => console.log("clicky click")}
    role="combobox"
    ariaExpanded={false}
    ariaHaspopup={true}
    ariaControls="lmao"
    icon={"arrowDown"}
    iconOnRight={true}
    variation="subtle"
    type="primary"
  />
```
this should be valid according to TypeScript, and result in a `role="combobox"` on the final HTML output
image

for other buttons, they should _not_ be getting a custom role, or a role at all other than their implicit one provided by say the `a` or `button` tag. this should be true of the final HTML and even before that, TypeScript should complain if you try to have a `url` or `to` or `submit` so try that in your editor please and thank you!

image
image

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.test.tsx
+++ b/packages/components/src/Button/Button.test.tsx
@@ -143,7 +143,7 @@ it("renders button type='submit'", () => {
 });
 
 it("routes when buttons are clicked", () => {
-  const { queryByText } = render(
+  const { getByText, queryByText } = render(
     <Router>
       <Button label="One" to="/" />
       <Button label="Two" to="/two" />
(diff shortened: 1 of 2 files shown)


packages/components/src/Button/Button.test.tsx
```diff
@@ -166,15 +166,33 @@ it("routes when buttons are clicked", () => {
...
+describe("Button role", () => {
+  it("should have a role of button when role not provided", () => {
+    const { getByRole } = render(<Button label="hello" />);
+    expect(getByRole("button")).toBeInstanceOf(HTMLButtonElement);
+  });
+  it("should not have a role of button when role not provided and it's a link", () => {
+    const { queryByRole, getByRole } = render(
+      <Button label="hello" url="myspace.com" />,
+    );
+    expect(queryByRole("button")).not.toBeInTheDocument();
```
dlrussel, line 191: What is the reason for using queryByRole here and getByRole above?

dlrussel (approved):
Just one question about the tests but this is good to go ✅

packages/components/src/Button/Button.test.tsx
ZakaryH: hey! yeah so `getBy*` and `queryBy*` each have their own use cases they are best suited for

>getBy...: Returns the matching node for a query, and throw a descriptive error if no elements match or if more than one match is found (use getAllBy instead if more than one element is expected).

so if you **are expecting something** to be there, you'd reach for `getBy*`

VS

>queryBy...: Returns the matching node for a query, and return null if no elements match. This is useful for asserting an element that is not present. Throws an error if more than one match is found (use queryAllBy instead if this is OK).

https://testing-library.com/docs/queries/about

so for this particular test I want to make sure this instance of `Button` that turns into an `<a>` doesn't have the `button` role AKA I want to search for something I expect to not exist - so `queryBy`

then the second assertion I'm making I do want it, and expect it, to exist so I use `getBy`

-------

as for my actual assertion logic you could definitely say that I'm not even passing in a `role="button"` to this invocation, and if I did - the test would in fact fail because it does end up adding the role to the output HTML

image

so arguably my test should/could be "Typescript should get upset if given a role for an incompatible button type" but that's testing our implementation of Typescript and the XOR package. we've done nothing to stop you ignoring Typescript and putting a `role` on a `Button` it doesn't belong on

with a quick search, it seems like it is possible to write tests for Typescript types, but we'd need to add a few packages & setup
https://www.olivare.net/blog/testing-ts-type-defs

PR #1538, opened 2023-09-15
feat(components): Allow actions to be added on a single DataList item

darryltec:
## Motivations

With a list of items, we need to add a way for the users to trigger actions on them. This allows the user to define what actions can a user do in a single list item.

### What's in it?
- Allow consumers to implement actions
- Menu set up to allow us to show a context menu but not implement it yet

### What's not in it
- Moving the actions with the layout on smaller screens.
  - That's on a different ticket
- Right-click context menu

### Added

- Component signature for adding an action on each item

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- It should show the first 2 actions, but only if those 2 actions have an icon
- If none of the actions have icons, it shouldn't expose it
- All remaining actions that aren't exposed should show under the overflow more menu
- It should trap the focus when you open the overflow menu
- It should close the menu when you press escape

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListAction/DataListAction.tsx
+++ b/packages/components/src/DataList/components/DataListAction/DataListAction.tsx
@@ -1,17 +1,40 @@
 import React from "react";
+import classNames from "classnames";
 import styles from "./DataListAction.css";
-import { DataListActionProps, DataListObject } from "../../DataList.types";
+import {
+  DataListActionProps,
+  DataListObject,
+  InternalDataListActionProps,
+} from "../../DataList.types";
 import { Typography } from "../../../Typography";
 import { Icon } from "../../../Icon";
 
-export function DataListAction<T extends DataListObject>({
+export function DataListAction<T extends DataListObject>(
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _: DataListActionProps<T>,
+) {
+  return null;
+}
+
+export function InternalDataListAction<T extends DataListObject>({
   label,
   icon,
-}: DataListActionProps<T>) {
+  destructive,
+  onClick,
+  item,
+  withIconOffset,
+}: InternalDataListActionProps<T>) {
   return (
-    <button className={styles.action}>
-      <div>{icon && <Icon name={icon} />}</div>
-      <Typography textColor="blue">
+    <button
+      className={classNames(styles.action, {
+        [styles.withIconOffset]: withIconOffset,
+      })}
+      onClick={() => onClick?.(item)}
+    >
+      <div>
+        {icon && <Icon name={icon} color={destructive ? "critical" : "blue"} />}
+      </div>
+      <Typography textColor={destructive ? "critical" : "blue"}>
         <span className={styles.label}>{label}</span>
       </Typography>
     </button>
--- a/packages/components/src/DataList/components/DataListAction/DataListAction.tsx
+++ b/packages/components/src/DataList/components/DataListAction/DataListAction.tsx
@@ -1,0 +1,8 @@
+import { DataListActionProps, DataListObject } from "../../DataList.types";
+
+export function DataListAction<T extends DataListObject>(
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _: DataListActionProps<T>,
+) {
+  return null;
+}
--- a/packages/components/src/DataList/components/DataListAction/DataListAction.tsx
+++ b/packages/components/src/DataList/components/DataListAction/DataListAction.tsx
@@ -22,11 +22,13 @@ export function InternalDataListAction<T extends DataListObject>({
   onClick,
   item,
 }: InternalDataListActionProps<T>) {
+  const color = destructive ? "critical" : "blue";
+
   return (
     <button className={styles.action} onClick={() => onClick?.(item)}>
-      {icon && <Icon name={icon} color={destructive ? "critical" : "blue"} />}
+      {icon && <Icon name={icon} color={color} />}
 
-      <Typography textColor={destructive ? "critical" : "blue"}>
+      <Typography textColor={color}>
         <span className={styles.label}>{label}</span>
       </Typography>
     </button>
--- a/packages/components/src/DataList/components/DataListAction/DataListAction.tsx
+++ b/packages/components/src/DataList/components/DataListAction/DataListAction.tsx
@@ -1,5 +1,4 @@
 import React from "react";
-import classNames from "classnames";
 import styles from "./DataListAction.css";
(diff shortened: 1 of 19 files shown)


darryltec:
@chris-at-jobber you kenat. I'd actually add a subtask for that baby as this PR is getting hella huge

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
```diff
@@ -27,11 +30,24 @@ export function DataListItems<T extends DataListObject>({
...
             {elementData.map((child, i) => {
+              const item = data[i];
+
               return (
-                <div className={styles.listItem} key={`${data[i].id}`}>
+                <div
+                  onMouseEnter={() => setHover(item.id)}
+                  onMouseLeave={() => setHover(undefined)}
+                  className={styles.listItem}
+                  key={`${item.id}`}
```
rodrigoeidelvein, line 40: Why is it in template strings?

packages/components/src/DataList/components/DataListItemActions/DataListItemActions.tsx
```diff
@@ -0,0 +1,103 @@
...
+      animate="visible"
+      exit="hidden"
+      transition={{
+        duration: TRANSITION_DURATION_IN_SECONDS,
+        delay: TRANSITION_DELAY_IN_SECONDS,
+      }}
+      className={styles.menu}
+    >
+      {exposedActions.map(({ props }) => {
+        if (!props.icon) return <></>;
```
rodrigoeidelvein, line 62: Might return `null` instead of an empty fragment as we don't want to render anything

packages/components/src/DataList/components/DataListActionsMenu/DataListActionsMenu.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <>
+          <div className={styles.overlay} onClick={onRequestClose} />
+
+          <motion.div
+            ref={setRef}
+            variants={variants}
+            initial="hidden"
+            animate="visible"
+            exit="hidden"
+            transition={{ duration: tokens["timing-base"] / 1000 }}
```
rodrigoeidelvein, line 47: You can use the `TRANSITION_DURATION_IN_SECONDS` const

packages/components/src/DataList/components/DataListAction/DataListAction.tsx
```diff
@@ -0,0 +1,34 @@
...
+  icon,
+  destructive,
+  onClick,
+  item,
+}: InternalDataListActionProps<T>) {
+  return (
+    <button className={styles.action} onClick={() => onClick?.(item)}>
+      {icon && <Icon name={icon} color={destructive ? "critical" : "blue"} />}
+
+      <Typography textColor={destructive ? "critical" : "blue"}>
```
rodrigoeidelvein, line 29: As these both are using the same logic, could be extracted into a constant

rodrigoeidelvein (approved):
Some comments, but none of them are blockers...

Approving in advance

packages/components/src/DataList/components/DataListAction/DataListAction.tsx
darryltec: oh, you's right! good catch!

packages/components/src/DataList/components/DataListActionsMenu/DataListActionsMenu.tsx
darryltec: I knew there was 1 more haha. good catch!

packages/components/src/DataList/components/DataListItemActions/DataListItemActions.tsx
darryltec: oh tru!

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
darryltec: hmmm can't remember, i think it has something to do with key's needing to be a string.

darryltec:
@rodrigoeidelvein ill do them on the following tasks! Will wait for @chris-at-jobber to do the design :stamp:

chris-at-jobber:
Seeing some inconsistencies with the mockup on the on-hover menu

(Current PR on left, Figma on right)
image

- hover menu is sitting too high relative to the row (mockup appears to be 10px higher, code is 20px higher)
- hover menu should have a border

chris-at-jobber:
Border radius also looks a bit off, Figma is 8 and the code _appears_ to be 4.

chris-at-jobber:
Last one, referencing [this mockup](https://www.figma.com/file/TuoXq9l7Mg1Xfusj6qnx31/%5BJobber-2025-System%5D-List-Views?type=design&node-id=1822%3A121415&mode=design&t=G59PzTU7igyj0hyj-1) the menu doesn't use any icons. I see that you removed the offset so when a user does add an icon it keeps the text aligned to the left so maybe that's fine for now? I can see people wanting to use icons in these menus so don't have a strong opinion as of right now.

darryltec:
> hover menu is sitting too high relative to the row (mockup appears to be 10px higher, code is 20px higher)

Thought it was 11 which seems oddly specific.

Screenshot 2023-09-15 at 2 16 36 PM

Anyhoo, I eyeballed it to 25% of the size. But forgot that `top` is relative to the parent so on taller bois, it goes even more up. haha. 

Our spacing token either goes 8 or 16 tho so we gotta stick with those numbers.

> hover menu should have a border

yup! totally missed that part haha

darryltec:
@chris-at-jobber RE: Icon in menus. That mockup seems to be missing those first 2 actions from the hover. Which does have an icon. Action still supports an icon, just doesn't offset the non-icon ones. Pretty easy to bring back if we think we should bring back the offset.

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
darryltec: looks like it doesn't have to be in template strings haha

chris-at-jobber:
@darryltec agreed on the tokens, let's use `space-small` aka 8 for the offset.

darryltec:
@chris-at-jobber done-zo!

chris-at-jobber (commented):
Design approved! Actions feel much more connected to their row now.

darryltec:
Eyooooo shipit!

PR #1539, opened 2023-09-15
feat(components): support select all and deselect all on DataList

darryltec:
## Motivations

Since we can select 1 item, we need to allow for selecting and deselecting all along with telling the user how many have been selected.

This also updates the sorting icon as the alignment has been messed up with the addition of taller header.

### Added

- Added a checkbox on the header to allow implementation of "select all"
- Added a counter of how many have been selected
- Added a "Deselect all" button when 1 or more have been selected

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


chris-at-jobber:
Nice use of the animated switcher on the transition 👌 

When I click the selected "all" checkbox, it's a bit surprising that it remains checked, rather than de-selecting all (see GMail's inbox for example). Is that intentional? I get that there's a "deselect all" action, but I'd expect the checkbox to work as well.

darryltec:
@chris-at-jobber was just being lazy and needed to put this up to get Rodrigo to look at it. But, yeah I'll add it on the example.

chris-at-jobber:
Beauty, thanks!

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeaderCheckbox.tsx
```diff
@@ -0,0 +1,72 @@
...
+        switched={Boolean(selected.length)}
+        initialChild={children}
+        switchTo={
+          <div className={styles.headerBatchSelect}>
+            <Text>
+              <b>{selected.length} selected</b>
+            </Text>
+            <Button
+              label="Deselect All"
+              onClick={() => onSelect?.([])}
```
darryltec, line 50: still debating if there should be an `onDeselectAll` prop that returns `[]` always and forever

docs/components/DataList/Web.stories.tsx
```diff
@@ -254,9 +268,17 @@ const Template: ComponentStory<typeof DataList> = args => {
...
-    if (loadingNextPage) return "loadingMore";
+    if (loadingNextPage || loadingIDs) return "loadingMore";
     return args.loadingState;
   }
+
+  async function handleSelectAll() {
+    if (totalCount === selected.length) return setSelected([]);
+
+    const idsQuery = await getIDs();
+    const ids = idsQuery?.data?.allPeople?.edges?.map(({ node }) => node.id);
```
taylorvnoj, line 279: All the `?` jumped out 😆 

Is it too much to split up to avoid nesting?

```
    const edges = idsQuery?.data?.allPeople?.edges;
    if (!edges) return;

    const ids = edges.map(({ node }) => node.id);

File: packages/components/src/DataList/components/DataListLayoutInternal/DataListHeaderCheckbox.tsx (hunk 1 above)
Comment (taylorvnoj) [reply packages/components/src/DataList/components/DataListLayoutInternal/DataListHeaderCheckbox.tsx:50]: Good question.  I think it being called onSelect still implies that it can handle selecting and deselecting - you are selecting all either way.

File: packages/components/src/DataList/components/DataListLayoutInternal/__tests__/DataListHeaderCheckbox.tsx [hunk 3]
```diff
@@ -0,0 +1,156 @@
+import * as React from "react";
```
taylorvnoj, line 1: should this file have a `.test`?

taylorvnoj (commented):
Just played around with the select options and watching the `Select All` next to the counter.  When I'm not using Select All and I'm just selecting random items in the table, seeing the `-` in the top checkbox makes me feel like thats indicating a subtraction (like when you Select All and then remove some individual items).  That being said, I noticed that gmail is the exact same so I'm assuming we emulated that. Don't necessarily want to suggest a change here as I'm no designer, just wanted to throw it out there and curious about how we came to those choices.

https://www.loom.com/share/6abc0cd92c07420f9c3649f6a6e95b8f?sid=0f9b8ea2-3cd9-45ba-a91a-fc42bfa1e4cd

chris-at-jobber:
The line for the [indeterminate](https://atlantis.getjobber.com/?path=/docs/components-selections-checkbox-web--indeterminate#indeterminate) state is a pattern across Jobber, and is found in other systems as well. Mozilla has an example [here](https://developer.mozilla.org/en-US/play), for reference.

docs/components/DataList/Web.stories.tsx
darryltec: it's still gonna return undefined tho haha

packages/components/src/DataList/components/DataListLayoutInternal/__tests__/DataListHeaderCheckbox.tsx
darryltec: yup! good catch

darryltec:
@taylorvnoj ^ what chris said. Also a thing that raw HTML checkboxes do now apparently https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_checkbox_indeterminate click set state 
image

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeaderCheckbox.tsx
darryltec: I like that reason. I'll keep it as is!

taylorvnoj:
@chris-at-jobber @darryltec Oh perfect! Thank you!

docs/components/DataList/Web.stories.tsx
taylorvnoj: 😆 ok truee

darryltec: hmm now that I think of it, we promoting good code so imma do that suggestion.

taylorvnoj (approved):
LGTM!

PR #1540, opened 2023-09-18
feat(components): Refactor options as li elements

dlrussel:
## Motivations

Refactor to update the combobox options as li elements instead of radio inputs.
We made this decision due to accessibility concerns with the fact that when using a radio input when using the arrow keys to switch the selected value an option would be made without using spacebar or enter to confirm a selection and the combobox content would close automatically. This created a weird experience for the user.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
@@ -19,8 +19,17 @@
   visibility: hidden;
 }
 
+.optionsList {
+  display: flex;
+  padding: var(--space-base);
+  flex-direction: column;
+}
+
 .option {
-  display: block;
+  border: none;
+  border-radius: var(--radius-large);
+  text-align: left;
+  background: none;
 }
 
 .search {
@@ -60,10 +69,6 @@
 
 .clearSearch:focus {
   box-shadow: var(--shadow-focus);
-}
-
-.list {
-  padding: var(--space-base);
 }
 
 .search::after {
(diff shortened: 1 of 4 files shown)


taylorvnoj:
LGTM - @ZakaryH if you wanna do the approval that would be lovely 😄

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
```diff
@@ -19,8 +19,19 @@
...
+.optionsList {
+  margin: 0;
+  padding: 0;
+  list-style: none;
+}
+
 .option {
-  display: block;
+  padding: var(--space-small) var(--space-base);
+  cursor: pointer;
```
ZakaryH, line 30: we'll need to have this rule in `.option:hover` for it to have the desired effect, but we can make a note to do that in the styling PR, or we could do it now. I don't mind terribly either way!

ZakaryH (approved):
I'm sad we can't use `button` because that gets us the click & return/space bar activation for free. using `li` we will need to handle the keyboard events ourselves. I can do that in my accessibility PR to get this merged ASAP to unblock the other tasks. 

this LGTM - but maybe change the title of the PR since we're not using buttons now!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
taylorvnoj: I've got it in my styling PR 👍

PR #1541, opened 2023-09-19
feat(components): move sorting to a button when the header disappears

darryltec:
## Motivations

We've recently implemented sorting and it is closely tied to the header showing up. Unfortunately, you can hide that header on any breakpoint which removes the sorting capability even if you have implemented it.

This PR moves the sorting capability to a button whenever the header disappears. That way the user have access to it at all times.

### Added

- Move the sorting capability to a button with the filters when the header is not shown
- Capability to import from `@jobber/components` instead of `../../../../../../../`

### Changed

- Renamed `direction` to `order` since it is more common to hear "sort order" than "sort direction"

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Adjust story size to large mobile <br /> 
image
- Click this <br />
image
  - That should swap between big enough that it's not a button, and small enough that it is a button

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
+++ b/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
@@ -4,7 +4,7 @@
 
   position: absolute;
   /* Inputs are off by 1 when put beside a button */
-  top: calc(var(--space-small) - 1px);
+  top: calc(var(--space-small) + 1px);
   right: var(--button-offset);
   visibility: hidden;
   width: 0;
--- a/jest.web.config.js
+++ b/jest.web.config.js
@@ -11,6 +11,7 @@ module.exports = {
     "react-markdown":
       "<rootDir>/node_modules/react-markdown/react-markdown.min.js",
     "^@jobber/hooks/(.*)$": "<rootDir>/packages/hooks/dist/$1",
+    "^@jobber/components/(.*)$": "<rootDir>/packages/components/src/$1",
   },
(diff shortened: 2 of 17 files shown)


chris-at-jobber:
Is there a way we can ensure the popover remains onscreen while it's open? I'm able to swipe-scroll the actionbar which leads to the popover being shuffled away:
image

darryltec:
@chris-at-jobber so, it turns out, you can. Not sure what problems this would cause on other poppers tho. Also, the arrow doesn't go away so it eventually points to the search haha

Sep-20-2023 06-57-08

dlrussel (approved):
Code looks good and I followed the QA instructions and everything looks good to me!

packages/components/src/DataList/components/DataListSearch/DataListSearch.css
```diff
@@ -4,7 +4,7 @@
 
   position: absolute;
   /* Inputs are off by 1 when put beside a button */
-  top: calc(var(--space-small) - 1px);
+  top: calc(var(--space-small) + 1px);
```
dlrussel, line 7: everything looks good to me so I approved the PR. just curious though why we need this??

darryltec: oh right. should've raised that on the description. The off by 1 comment is actually miscalculated. it shouldn't be a subtraction. Otherwise, you'd get this

image


compared to addition

image

PR #1542, opened 2023-09-19
feat(components): add context menu on data list

darryltec:
## Motivations

We've implemented a hover menu but there's 1 problem to hover menu's on DataList, you will have to tunnel your mouse from left to right to get to that hover menu. 

So instead of tunneling your way to the left, you can just right click anywhere on that item and get the same actions

### Added

- Context menu on the data list item
- Add active background when the context menu is triggered so you know which one has been right clicked

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -54,7 +54,7 @@
 }
 
 .listItem.active {
-  --data-list-item-active-color: rgb(243, 245, 246);
+  --data-list-item-active-color: var(--color-surface--background);
   background-color: var(--data-list-item-active-color);
 }
 
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -50,11 +50,12 @@
   position: relative;
   padding: var(--space-small);
   border-bottom: var(--border-base) solid var(--color-border);
-  transition: all var(--timing-base) ease-in-out;
+  transition: all var(--timing-base);
 }
 
 .listItem.active {
-  background-color: rgb(243, 245, 246);
+  --data-list-item-active-color: rgb(243, 245, 246);
+  background-color: var(--data-list-item-active-color);
 }
 
 .selectable {
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -50,6 +50,11 @@
   position: relative;
   padding: var(--space-small);
   border-bottom: var(--border-base) solid var(--color-border);
+  transition: all var(--timing-base) ease-in-out;
+}
+
+.listItem.active {
+  background-color: rgb(243, 245, 246);
 }
 
 .selectable {
--- a/packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
+++ b/packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from "react";
 import { AnimatePresence } from "framer-motion";
+import classNames from "classnames";
 import { DataListLayoutInternal } from "./DataListLayoutInternal";
 import { DataListItemInternal } from "./DataListItemInternal";
 import { Breakpoints } from "../../DataList.const";
@@ -7,6 +8,9 @@ import styles from "../../DataList.css";
 import { DataListLayoutProps, DataListObject } from "../../DataList.types";
 import { generateListItemElements } from "../../DataList.utils";
 import { InternalDataListItemActions } from "../DataListItemActions";
+import { useDataListContext } from "../../context/DataListContext";
+import { DataListActionsMenu } from "../DataListActionsMenu";
+import { InternalDataListAction } from "../DataListAction";
 
 interface DataListItemsProps<T extends DataListObject> {
   readonly layouts: React.ReactElement<DataListLayoutProps<T>>[] | undefined;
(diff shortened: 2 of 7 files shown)


chris-at-jobber:
I'm able to get into a _fun_ state where:

- more actions hover menu is open
- right-click context menu is open
- browser right-click context menu is open

image

Is it possible to communicate to the various menus to close? Since the "left-click outside" behaviour is working, I'm hoping there's a way we can also handle "right-click outside" 🤞

packages/components/src/DataList/components/DataListTags/DataListTags.css
```diff
@@ -7,7 +7,7 @@
 }
 
 .tagCount {
-  --overflow-bg: var(--color-white);
+  --overflow-bg: var(--data-list-item-active-color, var(--color-white));
```
chris-at-jobber, line 10: Just for my own curiousity, what does the comma-separated list of values do here? I know how it works with `background-image` but I've never seen it used for colors!

packages/components/src/DataList/DataList.css
```diff
@@ -50,6 +50,12 @@
   position: relative;
   padding: var(--space-small);
   border-bottom: var(--border-base) solid var(--color-border);
+  transition: all var(--timing-base);
+}
+
+.listItem.active {
+  --data-list-item-active-color: rgb(243, 245, 246);
```
chris-at-jobber, line 57: I just looked up this specific value and it seems _not that far_ from our `surface--background` - is this color just being pulled from the Figma?

packages/components/src/DataList/components/DataListTags/DataListTags.css
darryltec: It's CSS variable specific. It's basically a fallback when that variable doesn't exist. Which it doesn't when the parent isn't active.

packages/components/src/DataList/DataList.css
darryltec: Yup!! I was like, "that does not exist. Chris will comment on it, let's push it" haha

darryltec:
@chris-at-jobber the system context menu showing up is intentional on my side as it's weird to disable that when the menu is already open. The overflow tho, I will fix!

packages/components/src/DataList/components/DataListTags/DataListTags.css
chris-at-jobber: 🌟 The More You Know ☄️  

Specific to this case, we should be using `color-surface` rather than the base `color-white` to define the background of the overflow, right?

darryltec: Oh you right!!

chris-at-jobber: image

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
```diff
@@ -19,41 +23,83 @@ export function DataListItems<T extends DataListObject>({
   mediaMatches,
   data,
 }: DataListItemsProps<T>) {
+  const { itemActionComponent } = useDataListContext();
   const elementData = generateListItemElements(data);
```
ZakaryH, line 27: this is a relatively expensive calculation with a `reduce` nested in a `map` where the `data` we're mapping can keep growing as we load more

due to the mouse enter and mouse leave setting state inside this component, this will get recalculated a lot of times moving your cursor down the list

let's memo-ize this value with `data` as the only dependency to make sure we're only ever recalculating it when we need to!

packages/components/src/DataList/components/DataListLayoutInternal/__tests__/DataListItems.test.tsx
ZakaryH, line 1: what do you think about a test that asserts the context menu is not only present, but on the correct row that we clicked on?

I'm not sure if that will be practical or even possible since it would likely involve looking at some css position styles, and then comparing that value against the index of where we clicked to figure out if it's the right spot or not.

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
```diff
@@ -19,41 +23,83 @@ export function DataListItems<T extends DataListObject>({
...
+      )}
     />
   );
+
+  function handleMouseEnter(item: T) {
+    return () => setActiveID(item.id);
+  }
+
+  function handleMouseLeave() {
+    !contextPosition && setActiveID(undefined);
```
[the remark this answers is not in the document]
darryltec: That was autofixed. I started with Boolean first. haha

image

It probably should be a `!isContextMenuVisible` tho so good catch.

As for `undefined`. That argument is actually new to me since we are using Typescript and Atlantis is purely using `undefined` for things (like Typescript as well). The only use of `null` in Atlantis is from refs.

> bug happened and it's undefined

Can't think of an example where that would happen on logic where you're explicitly setting a state. I can see that on `functions` that should be returning a `string` but then returns `undefined` instead since that's how `void` works. But still, TS would catch that and throw an error about 

> This function is returning undefined but it should be returning a string

I see what you're saying tho since GQL and Apollo does require the difference between null and undefined.

packages/components/src/DataList/components/DataListLayoutInternal/__tests__/DataListItems.test.tsx
darryltec: don't know if jest-dom would actually return the positon. I can fake it tho.

darryltec: image

yup. it don't know about the position. Tho it does include the item whenever you click those bois, so i can fire one of them up and see if it returns the item I'm expecting.

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
ZakaryH: first, thanks for making that change. this is much better now!

second, I do wonder if there's more optimization for us to do here. even with this, we're doing thousands and thousands of iterations as we get to the end of the list with only ~80 items

image

I feel like we should be able to cache or memo-ize even further here, because for each new set of data added we have already gone through the process of generating the elements for everything before. I don't think there's any value in generating them again because there's nothing that would have changed for them.

ideally we leverage the work we've already done, and then just do what's necessary for that new data

that said, I know that's not the point of this PR so I'm not going to block on that but it's something we should decide if we feel is tech debt/worth following up on

packages/components/src/DataList/components/DataListLayoutInternal/__tests__/DataListItems.test.tsx
darryltec: Figured it out! We are good!

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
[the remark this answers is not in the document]
ZakaryH: yeah fair, I mean TS _should_ be protecting us from code paths that could result in that. 

honestly I don't know enough about DataList at the moment to present any super informed cases where I'd reasonably expect an `undefined` could sneak in undetected and cause the vague scenario I mentioned

so I'm fine with not doing anything here

and yes, I agree about `null` vs `undefined` which could become a lot more relevant as we integrate with the backend which might have different, meaningful implications for `undefined` vs `null`

ZakaryH:
@darryltec I'm not sure if this needs design approval first before I give it the official approved status? but from the code side this LGTM

only thing is the bit about potentially optimizing `generateListItemElements` but that does not need to be done in this PR

chris-at-jobber (commented):
Design approved ✅

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
darryltec: yeah, i'm not totally fond of how we did the generation of the items. I think that's a good optimization to do. I'll add a tix to get it looked at.

With that said, it is mostly regenerating the code behind it. Luckily for us, the `key` is preventing everything from being rendered in the DOM.

ZakaryH (approved):
LGTM

PR #1543, opened 2023-09-20
feat(components): Ability to externally clear combobox selection

dlrussel:
## Motivations

We want the user who is implementing the combobox to be able to clear the selected value from outside of the component.

## Changes

Added the a `useEffect` that will update the `setSelectedOption` value whenever the `selected prop is updated.
I also updated this prop so that it can be set to null. So if the selected value were to be updated and the null value was passed in this should trigger a rerender and clear the selected value for the combobox.

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -99,7 +99,10 @@ const ComboboxChip: ComponentStory<typeof Combobox> = args => {
 };
 
 const ComboboxClearSelection: ComponentStory<typeof Combobox> = args => {
-  const [selected, setSelected] = useState<ComboboxOption | null>(null);
+  const [selected, setSelected] = useState<ComboboxOption | null>({
+    id: "1",
+    label: "Bilbo Baggins",
+  });
 
   return (
     <>
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from "react";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
-import { Combobox } from "@jobber/components/Combobox";
+import { Combobox, ComboboxOption } from "@jobber/components/Combobox";
 import { Button as ClearButton } from "@jobber/components/Button";
 
 export default {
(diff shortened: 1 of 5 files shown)


docs/components/Combobox/Web.stories.tsx
```diff
@@ -12,52 +13,61 @@ export default {
...
-      <Combobox.TriggerButton
-        label="Select a teammate"
-        variation="subtle"
+    <>
+      <ClearButton
+        label="Clear Selection"
         type="primary"
-        icon="arrowDown"
-        iconOnRight={true}
+        onClick={() => setSelected(null)}
```
ZakaryH, line 23: typescript isn't happy with this. we'll wanna type the `selected` state as 

`useState<number | null>` and then this will stop complaining

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -41,7 +42,7 @@ interface ComboboxContentProps {
    * @optional
    * @type string
    */
-  readonly selected?: string | number;
+  readonly selected?: string | number | null;
```
ZakaryH, line 45: how do we feel about the whole `string | number | null` type for `selected` vs `ComboboxOption | null`?

or maybe if there's a TS way to accept excess properties so we need the minimum of `label` and `id` but don't care if you give us more? then we don't need to specifically use `ComboboxOption` but I digress...

the reason I ask is because we're a bit mismatched with the input vs output now where we have

`readonly onSelect: (selection: ComboboxOption) => void;`

so the incoming value is just a number or string, but the value we get back will be `{ id: 1, label: "whatever" }` meaning you'd need a setup like this

```
const [ selected, setSelected ] = useState<number | null>(1);

<Combobox>
  <Combobox.Content
   onSelect={selection => setSelected(selection.id)}
   selected={selected}
   />
</Combobox>
```

where it's that `selection.id` that feels a bit weird to me. I think there are pros and cons to either approach, but whichever we choose let's make the input and output match!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -208,6 +209,10 @@ function useComboboxContent(
   const [selectedOption, setSelectedOption] = useState<ComboboxOption | null>(
     defaultValue || null,
   );
+  useEffect(() => {
+    setSelectedOption(defaultValue || null);
+  }, [selected]);
```
ZakaryH, line 214: Darryl had previously suggested scrapping `const [selectedOption, setSelectedOption]` altogether and instead using `props.selected` directly

this could 100% work for single select, and make it simpler for how to manage what is selected and clearing that selection. `selected` becomes mandatory because it is the state now.

my argument was that multiselect will want that internal state, but obviously we're not building multi select right now so that's not the greatest argument

what are your thoughts on that idea?

ZakaryH:
a couple questions/thoughts. aside from that, could we please add a lil test that verifies this use case?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
dlrussel: ok ya good call they should definitely be the same.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: I tried this out to see how hard it would be. it's surprisingly straightforward

image

no more `defaultOption` nor useEffect listening to `props.selected`. the value passed in just is what we use, so if you give us a selection to begin with - cool, that will be selected. if you either give us `null` to begin with or change a value from something to `null` then it's empty. I don't mind it honestly.

is there something you're worried about as far as re rendering on each selection?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: after playing around with this, I feel like `ComboboxOption | null` is preferable as the type for `selected` which then matches the output we currently have

with `selected` as `number | string | null` and applying those same types to the corresponding outgoing `selection` you get from the `onSelect` callback, it becomes a bit inconvenient

say your selected state is `const [ hat, setHat ] = useState<number | null>(2)`

trying to do this will have a problem `onSelect={selection => setHat(selection)}`. we've said `hat` can be either a number or null. however; `selection` can be a number, null or a string. it's that third possibility that causes the issue. you can do some type coercion but that doesn't feel very elegant to me

whereas if we use `ComboboxOption` you can simply pass in a `selected` and `setHat(selection)` with no other steps required. it's very similar to how we handle the `onChange` of `Autocomplete` which uses an `Option`

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
darryltec: Just chiming in about rerendering.

If you're talking about the DOM element rerendering. I can assure you that the majority of the time, React (on web) knows not to rerender an element if it didn't drastically change. Even adding a new element inside another element or adding a simple class name won't trigger a full rerender of that element. 

If you use an index as a `key` and the index has changed then it will rerender the items where the index changed.

You can even test this in the DOM. 

image

If you're talking about the functional component itself rerendering all of the function/const within it, no amount of state would stop that from happening since that's how react keeps on reacting to changes. You can use memoization, but that's overkill and putting in resources where it doesn't need to.

docs/components/Combobox/Web.stories.tsx
```diff
@@ -94,8 +98,70 @@ const ComboboxChip: ComponentStory<typeof Combobox> = args => {
   );
 };
 
+const ComboboxClearSelection: ComponentStory<typeof Combobox> = args => {
+  const [selected, setSelected] = useState<ComboboxOption | null>(null);
```
ZakaryH, line 102: what do you think about having an initial option here so that you can immediately click the button to clear it rather than making a selection and then clearing?

dlrussel: ya good call I'll update that

ZakaryH (approved):
one teeny tiny change with the Combobox clearing web story, but even if we decided not to do that, this still LGTM

happy to take another peek once that's been added

PR #1544, opened 2023-09-21
feat(components): DataList Layout pluggable actions

darryltec:
## Motivations

By default, we have implemented the actions to be triggered by hover. However, we want the freedom to place those actions in any location in the layout. 

For example, we want it to be a hover menu on larger screens. But on smaller screens, we want it to be a part of the layout.

Sep-21-2023 08-15-43

### Added

- Implement the idea of `DataList.LayoutActions` to be used only within `DataList.Layout` to inject it as part of the layout manually.
  - This removes the hover menu whenever you use it on a layout.

### Changed

- Sent the menu to the body

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListLayoutActions/DataListLayoutActions.test.tsx
+++ b/packages/components/src/DataList/components/DataListLayoutActions/DataListLayoutActions.test.tsx
@@ -4,10 +4,12 @@ import { DataListLayoutActions } from "./DataListLayoutActions";
 import { DataListContext, defaultValues } from "../../context/DataListContext";
 import { DataListItemActions } from "../DataListItemActions";
 import { DataListAction } from "../DataListAction";
-import { DataListLayoutContext } from "../../context/DataListLayoutContext";
+import {
+  DataListLayoutContext,
+  defaultValues as layoutDefaultValues,
+} from "../../context/DataListLayoutContext";
 
-const mockMainSetHasInLayoutActions = jest.fn();
-const mockLayoutSetHasInLayoutActions = jest.fn();
+const mockSetHasInLayoutActions = jest.fn();
 const mockItemActionComponent = jest.fn<JSX.Element | undefined, []>(() => (
   <DataListItemActions>
     <DataListAction label="Edit" />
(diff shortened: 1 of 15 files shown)


chris-at-jobber (changes requested):
Bit of a fun one that's only noticeable on-device

When I tap the `more` button, the tooltip remains open alongside the menu, when only the menu should be open. I think it's retaining the faux-hover that can happen on taps?

On desktop, clicking or keyboarding the "more actions" button closes the tooltip and opens the menu as expected.

darryltec:
@chris-at-jobber hmmm it shouldn't since the focus moves to the menu. This is how it works on my phone and simulator

Sep-22-2023 07-47-16

What version of ios is that? Is that chrome? Why is the search bar up top

taylorvnoj:
Would it make sense to make use of the `useLayoutMediaQueries` hook in `DataLiswtItemActionsOverflow.tsx` to conditionally render the tooltip or not?

chris-at-jobber:
@darryltec Chrome on iOS 16.1.2

Also, do you notice how tiny the text on the context menu gets? I thought I was just seeing things at first glance but it's noticeably smaller 🤔

darryltec:
@chris-at-jobber I do actually. could be a classic ios button thing.

darryltec:
@chris-at-jobber @taylorvnoj I still don't see it on mine. I am on ios17 tho. Tried on both chrome and safari. Could just be fixed on that version. It shouldn't really stay open because it is focus and hover dependent and we move those the moment the menu opens

darryltec:
It's also a tooltip issue so i'll open a ticket to fix it separately

packages/components/src/DataList/components/DataListLayoutActions/DataListLayoutActions.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+const mockSetHasInLayoutActions = jest.fn();
+const mockItemActionComponent = jest.fn<JSX.Element | undefined, []>(() => (
+  <DataListItemActions>
+    <DataListAction label="Edit" />
+    <DataListAction label="Email" />
+    <DataListAction label="Delete" />
+  </DataListItemActions>
+));
+
+describe("DataListLayoutActions.test", () => {
```
taylorvnoj, line 21: Not a deal breaker at all, but why the `.test` here

taylorvnoj (approved):
Looks really good - love all the comments you added.  Random question but not necessarily asking for a change.

Approving on the promise that the tooltip ticket is created 😆

packages/components/src/DataList/components/DataListLayoutActions/DataListLayoutActions.test.tsx
darryltec: tbh, bad copy pasting 😆 


```suggestion
describe("DataListLayoutActions", () => {
```

chris-at-jobber:
I'm seeing the hover menu on small screen - any chance I'm somehow on an outdated build?

image

chris-at-jobber (commented):
Design approved ✅ Tested in responsive mode on desktop and on-device in Chrome as well, tooltip issue to be addressed elsewhere

PR #1545, opened 2023-09-25
feat(design): update hover colors for common surfaces [JOB-82570]

chris-at-jobber:
## Motivations

The desired hover indication for our surfaces is that the surface darkens when hovered (comparable to how our buttons behave), rather than changing color to yellow. In addition, there are increasing amounts of designs calling for a hover state on `surface--background` surfaces. This adds a semantic value for those states to improve standardization of such states. There is no associated non-semantic value, as it is lighter than any of `blue` or `greyBlue` existing ranges.

To ensure WCAG AA (4.5:1) contrast is maintained between `text--secondary` and `surface--background`, the new value for `text--secondary` references the base `blue--light` instead of `greyBlue`. This matches the value used by Jobber's Creative Studio for secondary text.

To address this holistically we are also addressing the focus color and associated "focus ring" shadow that we use in lieu of the outline property.

**Examples in components**

`surface--background` in modal/card/drawer header
image

`surface--active` in list item 
image

`-surface--hover` in datalist + tab
image
http://localhost:6005/?path=/docs/components-text-and-typography-heading-docs--page#design--usage-guidelines

Adapted styles in DatePicker to use correct semantic values (no longer using `disabled` colors for interactive elements)
image

New `text--secondary` value in subdued heading variant (mobile)
image

**Focus examples**

image
image
image

**Palette**

image

### Added

- New hover color for `surface--background` referencing `greyBlue--lighter`

### Changed

- Update to `surface--hover` to reference a new rgb value instead of `yellow--lightest`
  - there is no associated "base" palette value as the new value is lighter than any existing values in our base range for blue or greyBlue
- Secondary text now references `blue--lighter` instead of `greyBlue`
- Focus color and shadow style
- Updated components (DatePicker and ListItem) to use new semantic values as needed

## Testing

Run `npm run bootstrap` locally to ensure your `design` package is synced up with your `components` package, or test the [Cloudflare build](https://update-hover-states.atlantis.pages.dev/)

- `surface--hover` changes can be tested by visiting List, Menu, etc and observing that the hover state is now grey
- `surface--background--hover` can be tested by using the color in a component, or visiting the[ color documentation](https://update-hover-states.atlantis.pages.dev/?path=/docs/design-colors--page)

---

Random photo of Atlantis


fagnanm:
@chris-at-jobber two places to take closer look:

Chips
- Hover feels too light and hard to tell if something is hovering. Maybe it's my eyes playing a trick on me.

Date Picker
- The lighter grey mixed with the darker grey for previous/next month days feels off

chris-at-jobber:
@fagnanm pushed up the changes we discussed offline

- introduce blue-tint values for `border`, `surface--hover`, `surface--background`

Maybe it's just the new color, but I think the Chips look fine (and there are the other changes coming down the pipeline for those as well)
image

I'll push up a Datepicker update in a separate commit to keep things more isolated, I agree with you though - we're using `disabled` styling in interactive elements when we already have a visual treatment available for "subdued" surfaces.

fagnanm:
@chris-at-jobber have you checked contrast between the lighter bluegrey and the text colour? Is there enough there to hit AA or AAA?

chris-at-jobber:
@fagnanm good catch - it would fall short (3.88:1). With that said, I can check the components and see where we might run into issues. I'd be surprised if we're using `text--secondary` on top of `surface--background` in _many_ places, although the OFF state of Switch is one I can think of off the top of my head.

chris-at-jobber:
**Blocked note**

We want the corresponding [SG-1](https://github.com/GetJobber/sg-1/pull/1358) and Atlantis changes to go into Jobber simultaneously, so if/when this PR is approved, please do NOT merge immediately

scotttjob (approved):
Ran Atlantis locally and investigated all the changed components. Everything looks great! Very swanky.

PR #1546, opened 2023-09-25
refactor(components): Visual updates to chip

chris-at-jobber:
## Motivations

To better align the Chip for web with our overall visual direction, stylistic changes are being implemented.

**FOR REVIEW/DISCUSSION/To Be Implemented**

- As Chip does not expose the properties of the `InternalChipAffix` publicly, should that be modified to be `InternalChipPrefix` and the current `activeIndicator` div that's been added turned into an `InternalChipSuffix`?
- Need to get the "activeIndicator" to also accept a `cross` icon, which probably means that "activeIndicator" is the wrong name for it (partially why I'm leaning towards the `InternalChipSuffix` approach above)
- Color changes to the hover state will be realized through #1545 

### Added

- a new suffix with a white background that contains an icon, to be used primarily in signifying selection or dismissability

### Changed

- background color of chip for default and hovered states
- text color of chip
- chip no longer changes color when "active", it receives a checkmark icon

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chips/InternalChip.css
+++ b/packages/components/src/Chips/InternalChip.css
@@ -16,13 +16,13 @@
   position: relative;
   max-width: 100%;
   box-sizing: border-box;
-  padding: var(--space-small);
-  border: var(--border-base) solid var(--color-border);
-  border-radius: var(--radius-larger);
-  color: var(--color-text--secondary);
+  padding: var(--space-small) var(--space-base);
+  min-height: calc(var(--space-largest) - var(--space-small));
+  border-radius: calc(var(--space-base) + var(--space-smaller));
+  color: var(--color-heading);
   text-align: left;
   vertical-align: middle;
-  background-color: var(--color-surface);
+  background-color: var(--color-surface--background);
   transition: var(--timing-base) ease-in-out;
   transition-property: background-color, box-shadow, color;
   grid-template-columns: max-content;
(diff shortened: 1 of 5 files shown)


taylorvnoj:
I'm not sure if this PR is fully ready but thought I'd point out the double checkmark on a selected chip in the `Multi Select` example

Screenshot 2023-09-28 at 12 00 53 PM

taylorvnoj:
Yes to `InternalChipPrefix`.  Except would it just be `ChipPrefix` since we're exposing `Chip` and nothing about this is internal anymore?  `ChipPrefix <Icon | Avatar>` like in your doc.

But then what to do with a `ChipSuffix`.  I have more of an understanding now of what you're talking about in the comment above this.  I agree that we want to avoid XOR.  This is where the `( Tags + )` example throws me, because I'm wondering if there will be a reason to have an icon in the suffix area otherwise. 

We have the 'appearance' of chip styled like this in combobox but to your point, it's a custom button.  
What other possible scenarios would someone want an icon on the right rather than the left of the label...
What I'm getting at is, do we _need_ a `suffix` that can be an icon, or can `suffix` only be `selected | dismissible` and stand optioned on that?

Looking through component gallery at other chips - icons are always (9 times out of 10) the prefix and `checkmark or x` is always the suffix. The only other use case is a dropdown, which we're account for with `Combobox`.

chris-at-jobber:
Closing in favour of #1571

PR #1547, opened 2023-09-26
fix(components): Improve performance on the list item creation

darryltec:
## Motivations

With the addition of actions and state management around when it opens, it started reinitializing everything the moment you interact with the list. That's due to how we're tracking the state for the menu from the parent instead of for each list.

This PR changes that by introducing the state per each item so the parent doesn't change which stops each data list item from reinitializing (note: I didn't say rerender since it doesn't rerender the element in the DOM). 

Adding a console.log [here](https://github.com/GetJobber/atlantis/blob/462ba662b8856312e33ab613ba51c50864d1aa96/packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx#L37) shows you that wIthout these changes the list items get reinitialized on load more and interacting with it with a mouse

image

And more if you keep moving your mouse

Compared to this PR where it stays at that number

image

Unfortunately, with how react keeps on reacting to changes, it would still try to reinitialize everytime the data changes. This time it only changes things when it needs to.

### Added

- 

### Changed

- Componentize the list item
  - Moved to its own folder
- Replaced 1 mega state to track which menu is open to a per item state so it only reinitializes the list item that is changing instead of all
- Only run the generate reducer on item render
  - This will pay off with a performance boost once we have a virtualized list and the reducer won't run for all. Just for the visible items,

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Add a console log [here](https://github.com/GetJobber/atlantis/blob/462ba662b8856312e33ab613ba51c50864d1aa96/packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx#L37) and see how many times the component reinitializes compared to master.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListItem/DataListItem.tsx
+++ b/packages/components/src/DataList/components/DataListItem/DataListItem.tsx
@@ -1,10 +1,15 @@
-import React, { Children, MouseEvent, ReactElement, useState } from "react";
+import React, {
+  Children,
+  MouseEvent,
+  ReactElement,
+  useMemo,
+  useState,
+} from "react";
 import { AnimatePresence } from "framer-motion";
 import classNames from "classnames";
 import { useDataListContext } from "@jobber/components/DataList/context/DataListContext";
 import { useDataListLayoutContext } from "@jobber/components/DataList/context/DataListLayoutContext";
 import {
-  DataListItemType,
   DataListLayoutProps,
   DataListObject,
 } from "@jobber/components/DataList/DataList.types";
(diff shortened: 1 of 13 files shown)


packages/components/src/DataList/components/DataListItem/DataListItem.tsx
```diff
@@ -0,0 +1,104 @@
...
+                  item={item}
+                />
+              ))}
+          </DataListActionsMenu>
+        </AnimatePresence>
+      </div>
+    </DataListLayoutActionsContext.Provider>
+  );
+
+  function handleShowMenu() {
```
taylorvnoj, line 87: I might have missed it - but should we test these event handlers?  
Maybe testing mouse enter/leave and checking that handleShowMenu is called, etc

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
```diff
@@ -24,104 +19,32 @@ export function DataListItems<T extends DataListObject>({
   mediaMatches,
   data,
 }: DataListItemsProps<T>) {
-  const { itemActionComponent } = useDataListContext();
-  const elementData = useMemo(() => generateListItemElements(data), [data]);
```
taylorvnoj, line 28: ohhhhh so are you removing the memoization here because now it's being handled in `DataListItem`, which is being mapped over in this file now.  Therefore being handled by each item rather than the whole list of items...

taylorvnoj (commented):
Neat! Huge difference in the console with your changes - that's awesome!  Separating of concerns makes things easier to understand for sure.  Definitely a lot for me to learn here about handing state in components.

Just a couple of small comments about tests (this component is understandably big so let me know if some of these changes are being tested elsewhere or can be tested).

packages/components/src/DataList/components/DataListItem/DataListItem.tsx
darryltec: hmmm thought it was tested in the list items with an s, but it is not. I'll save that baby up on the testing task JOB-78870

packages/components/src/DataList/components/DataListLayoutInternal/DataListItems.tsx
taylorvnoj: Nice! Great point!

taylorvnoj (approved):
LGTM :shipit:

PR #1548, opened 2023-09-27
feat(components): DataList row click and tabbing a11y

darryltec:
## Motivations

We need to allow a row click on the data list. There are 3 classification of an interactable item within DataList (and Jobber)
- normal onClick that is a button
- a link to another page (classic a href tag)
- a route to another page (SPA/PWA/etc)

This PR aims to support those 3.

On top of that, this also adjusts the tabbing experience within the list items.

image

## Out-of-scope
Voice over accessibility. Is not in this PR. Right now, it is good enough but it could be improved. Infortunately, that involves more thinking around how we handle that so to not block this PR further, we can follow up on that task.

### Added

- Tabbing order
- Show actions on focus
- A pretty sweet hover interaction (except for firefox)

### Changed

- Updated the focus trap hook to either focus on the parent if it's focusable or focus on the very first focuseable element.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -128,6 +128,11 @@
 }
 
 .listItem .selectable > :first-child {
+  order: 2;
+}
+
+.listItem .selectable > :last-child {
+  order: 1;
   padding-top: var(--space-smaller);
   opacity: 1;
   transition: opacity var(--transition-properties);
@@ -139,9 +144,9 @@
   }
 }
 
-.listItem:hover .selectable > :first-child,
-.listItem:focus-within .selectable > :first-child,
-.listItem .selectable.selected > :first-child {
+.listItem:hover .selectable > :last-child,
+.listItem:focus-within .selectable > :last-child,
+.listItem .selectable.selected > :last-child {
   opacity: 1;
 }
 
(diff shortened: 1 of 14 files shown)


chris-at-jobber:
One VO notes I've come across them (I know you've split these out)
- row checkbox should read the name of the selected row

packages/components/src/DataList/DataList.css
```diff
@@ -46,18 +50,54 @@
...
+
 .listItem {
   position: relative;
   padding: var(--space-small);
   border-bottom: var(--border-base) solid var(--color-border);
   transition: all var(--timing-base);
 }
 
-.listItem.active {
+.listItem.active,
```
chris-at-jobber, line 64: Should also use this visual state for when the `listItem` itself is focused - almost tripped me up when I was testing the tab order because it wasn't apparent my focus was set anywhere

chris-at-jobber (dismissed):
🤏 just a small change, looking great so far!

packages/components/src/DataList/DataList.css
chris-at-jobber: Update: it looks good now 👻

chris-at-jobber (commented):
Design approved ✅

rodrigoeidelvein (approved):
:shipit:

PR #1549, opened 2023-09-27
feat(components): Allow Bulk actions on DataList

rodrigoeidelvein:
## Motivations

Allow the consumers to add BulkActions on DataList. It uses the same structure as `DataList.Action`, but nothing is passed instead of returning the `item` in the callback. The consumer can we the `selected` prop to know which items are currently selected to trigger their actions.

Things that can be improved, and will be, but I don't know if in this current PR, I don't want to make this one huge.
- Add tests
- Create a component for the fade overflow behavior that we have in filters too, how should we call it?  `DataListFadeOverflow`?
- A11y

### Added

- 

### Changed

- DataList now accepts a `DataList.BulkAction`

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/DataList/components/DataListItemActions/DataListItemActionsOverflow.tsx
```diff
@@ -33,7 +33,7 @@ export function DataListItemActionsOverflow<T extends DataListObject>({
         />
       </Tooltip>
 
-      {item && (
+      {actions && (
```
rodrigoeidelvein, line 36: @darryltec do you see an issue with this? I changed it because bulk actions will not have a specific item to pass here.

packages/components/src/DataList/components/DataListAction/DataListAction.tsx
```diff
@@ -25,7 +25,16 @@ export function InternalDataListAction<T extends DataListObject>({
...
   return (
-    <button className={styles.action} onClick={() => onClick?.(item)}>
+    <button
+      className={styles.action}
+      onClick={() => {
+        if (item) {
+          onClick?.(item);
+        } else {
+          onClick?.();
+        }
```
rodrigoeidelvein, line 35: I'm not really happy with this, open to suggestions

packages/components/src/DataList/DataList.tsx
```diff
@@ -66,6 +67,9 @@ export function DataList<T extends DataListObject>({
   const itemActionComponent = getCompoundComponent<
     DataListItemActionsProps<DataListObject>
   >(props.children, DataListItemActions);
+  const bulkActionsComponent = getCompoundComponent<
+    DataListItemActionsProps<DataListObject>
```
darryltec, line 71: should this be?

```suggestion
    DataListBulkActionsProps<DataListObject>
```

packages/components/src/DataList/DataList.tsx
```diff
@@ -219,3 +224,4 @@ DataList.ItemActions = DataListItemActions;
  * DataListItemActions component.
  */
 DataList.Action = DataListAction;
+DataList.BulkActions = DataListBulkActions;
```
darryltec, line 227: We should explain this like the other things

packages/components/src/DataList/components/DataListAction/DataListAction.tsx
darryltec: first, put it in a handleClick function


```ts
function handleClick() {
  if (item) {
    return () => onClick?.(item)
  }

  return onClick
}
```

packages/components/src/DataList/components/DataListItemActions/DataListItemActionsOverflow.tsx
darryltec: nope!

packages/components/src/DataList/components/DataListBulkActions/DataListBulkActions.tsx
```diff
@@ -0,0 +1,72 @@
...
+    Children.toArray(children).filter<ReactElement>(isValidElement);
+
+  const { sm } = useResponsiveSizing();
+
+  // Collapse all actions under "More actions" when breakpoint is smaller than sm
+  const exposedActions = sm ? getExposedActions(childrenArray, 3) : [];
+  childrenArray.splice(0, exposedActions.length);
+
+  const [leftRef, isLeftVisible] = useInView<HTMLSpanElement>();
+  const [rightRef, isRightVisible] = useInView<HTMLSpanElement>();
```
darryltec, line 33: we should extract this overflowing part out into its own component and use here and on the filters instead of duplicating it

packages/components/src/DataList/DataList.types.ts
```diff
@@ -207,6 +208,14 @@ export interface DataListItemActionsProps<T extends DataListObject> {
   readonly children?: Fragment<ReactElement<DataListActionProps<T>>>;
 }
 
+export interface DataListBulkActionsProps {
+  /**
+   * The actions to render for each item in the DataList. This only accepts the
```
darryltec, line 213: that message seems wrong

packages/components/src/DataList/DataList.types.ts
```diff
@@ -226,10 +235,18 @@ export interface DataListActionProps<T extends DataListObject> {
   /**
    * The callback function when the action is clicked.
    */
-  readonly onClick?: (data: T) => void;
+  readonly onClick?: (data?: T) => void;
+}
+
+export interface DataListBulkActionProps
```
darryltec, line 241: I know this is being used in line 216, but that won't actually change the type of actions

image

still thinks a data might show up there.

One thing we could do tho is to have a `DataList.ItemAction` and `DataList.BatchAction` that both resolves back to just `DataListAction`.


```ts
DataList.ItemAction = DataListAction;
DataList.BatchAction = function DataListBatchAction(props: PropsWithoutThatDataInTheParam) => <DataListAction {...props} />
```

packages/components/src/DataList/DataList.utils.tsx
```diff
@@ -102,3 +102,23 @@ export function sortSizeProp(sizeProp: Breakpoints[]) {
...
+) {
+  const firstNChildren = childrenArray.slice(0, childCount);
+
+  return firstNChildren.reduce((result: typeof childrenArray, child, i) => {
+    const hasIcon = Boolean(child.props.icon);
+
+    if (!hasIcon) return result; // If the child does not have an icon, continue.
+
+    // If it's the first child or if the previous child was added, then add this child.
+    if (i === 0 || (i < childCount && result.length === i)) {
```
darryltec, line 118: that `result.length === i` confused me a bit there haha. mind putting that in a const and name it in a descriptive way

darryltec (changes requested):
First pass. looks pretty good

chris-at-jobber:
@rodrigoeidelvein we need screenreaders to recognize the `n selected` label, otherwise it's just telling me I'm on a selected or unselected checkbox.
checkbox with label 82 selected

chris-at-jobber (dismissed):
2 small changes to the selected actions bar:

1. Font-weight is too heavy (should be `Text` with no emphasis applied)
2. Rather than `justify-content: space-between`, should maintain `space-smaller` gaps between items

image

docs/components/DataList/Web.stories.tsx
```diff
@@ -261,8 +300,12 @@ const Template: ComponentStory<typeof DataList> = args => {
     </DataList>
   );
 
-  function handleActionClick(item: (typeof mappedData)[number]) {
-    alert(`You clicked the action for ${item.label}`);
+  function handleActionClick(item: (typeof mappedData)[number] | undefined) {
```
darryltec, line 303: protip: question mark should suffice

```suggestion
  function handleActionClick(item?: (typeof mappedData)[number]) {
```

packages/components/src/DataList/DataList.css
```diff
@@ -37,13 +37,15 @@
 
 .headerBatchSelect > p {
   font-weight: 500;
+  white-space: nowrap;
```
darryltec, line 40: if we're gonna use the checkbox label, don't need this CSS

packages/components/src/DataList/DataList.types.ts
```diff
@@ -207,6 +208,14 @@ export interface DataListItemActionsProps<T extends DataListObject> {
   readonly children?: Fragment<ReactElement<DataListActionProps<T>>>;
 }
 
+export interface DataListBulkActionsProps {
+  /**
+   * The actions to render on the top of the DataList to make actions to multiple items.
+   * This only accepts the DataList.Action component.
```
darryltec, line 214: missed a spot

```suggestion
   * This only accepts the DataList.BatchAction component.
```

packages/components/src/DataList/components/DataListFilters/DataListFilters.tsx
```diff
@@ -1,9 +1,6 @@
 import React, { ReactElement } from "react";
-import classNames from "classnames";
-import { useInView } from "@jobber/hooks/useInView";
-import styles from "./DataListFilters.css";
-import { CONTAINER_TEST_ID } from "./DataListFilter.const";
 import { DataListSort } from "./components/DataListSort";
+import { DataListOverflowFade } from "./components/DataListOverflowFade/DataListOverflowFade";
```
darryltec, line 3: consistency wise, add an index.ts on this folder and import it via

```suggestion
import { DataListOverflowFade } from "./components/DataListOverflowFade";
```

packages/components/src/DataList/components/DataListFilters/components/DataListOverflowFade/DataListOverflowFade.tsx
darryltec, line 1: I think this component should be 2 levels up and under `DataList/components` instead of `DataList/components/DataListFilters/components/DataListOverflowFade`

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeaderCheckbox.tsx
```diff
@@ -42,14 +51,12 @@ export function DataListHeaderCheckbox({ children }: DataListHeaderCheckbox) {
         initialChild={children}
         switchTo={
           <div className={styles.headerBatchSelect}>
-            <Text>
-              <b>{selected.length} selected</b>
-            </Text>
```
darryltec, line 47: I'd actually keep this here so it's a part of the transition when it shows up. I'd override the `b` tags `font-weight` tho

darryltec: As for how VO would announce the selected, you can add a child with the same text value but style it to [only show for VO](https://webaim.org/techniques/css/invisiblecontent/#techniques:~:text=.sr%2Donly%20%7B%0Aposition%3Aabsolute%3B%0Aleft%3A%2D10000px%3B%0Atop%3Aauto%3B%0Awidth%3A1px%3B%0Aheight%3A1px%3B%0Aoverflow%3Ahidden%3B%0A%7D)

packages/components/src/DataList/components/DataListLayoutInternal/__tests__/DataListHeaderCheckbox.test.tsx
```diff
@@ -112,6 +114,35 @@ describe("DataListHeaderCheckbox", () => {
   });
 
   describe("Select All Checkbox Indeterminate and Checked State", () => {
+    Object.defineProperty(window, "matchMedia", {
```
darryltec, line 117: I believe you have to reset this back to the original. Otherwise it'll be overwritten on other tests files as well.

darryltec (commented):
Okay, we're close!! Just some code structure and organization bits left. It totally works!

On that note @chris-at-jobber I still think we need to let the Deselect scroll as well since on smol screens, you get a smol window. It will be worse on phones with gigantic text settings.
 
Sep-28-2023 11-48-05

packages/components/src/DataList/components/DataListLayoutInternal/DataListHeaderCheckbox.tsx
rodrigoeidelvein: Done!

packages/components/src/DataList/components/DataListBulkActions/DataListBulkActions.tsx
```diff
@@ -0,0 +1,50 @@
...
+              ariaLabel={props.label}
+              onClick={() => props.onClick?.()}
+              type="secondary"
+              variation="subtle"
+            />
+          </Tooltip>
+        );
+      })}
+
+      <DataListItemActionsOverflow actions={childrenArray} />
```
darryltec, line 47: This is _almost_ the exact same code as https://github.com/GetJobber/atlantis/blob/ed7f30d2456b389487281bd0eaeb36ab7101336d/packages/components/src/DataList/components/DataListItemActions/DataListItemActions.tsx#L68-L84

I think we can just create a `DataListActions` component that gives you this code + this code https://github.com/GetJobber/atlantis/blob/ed7f30d2456b389487281bd0eaeb36ab7101336d/packages/components/src/DataList/components/DataListItemActions/DataListItemActions.tsx#L48-L53

and accepts a prop of 
- the list of actions children
- how many to expose (2 or 3)

As for the item, you can grab that from `useDataListLayoutActionsContext` so no need for prop drilling

Could be a follow up tho

darryltec (approved):
Code looks good. There's one that we can make it less duplicated but that can be a follow up.

I'll leave the approval to @chris-at-jobber as I think the selected test is still too bold. Needs about 600 font-weight 
image

chris-at-jobber:
The font-weight for the selection is still too heavy (I think this is what @darryltec pointed out as well). I don't think we should need any custom CSS at all, if we're using `<Text>`?

What it looks like:
image

What it should look like:
image

chris-at-jobber:
For the [tight scrolling window on small screens](https://github.com/GetJobber/atlantis/pull/1549#pullrequestreview-1649387200), I'm inclined to revisit some of the ideas we tried before - either hide the `Deselect` button entirely, or collapse all the actions into the `•••` overflow menu. (Probably the former)

chris-at-jobber:
Just the two notes

Could these be tackled in a smaller followup PR? Probably (especially because the small screen scrolling has a few potential solves).

chris-at-jobber (approved):
It cleared Darryl's approval for some reason

- Update text styling in followup
- Address actions scrolling on very small screens in followup

PR #1550, opened 2023-09-28
feat(components): Refactor/Clean Up Combobox Content

ZakaryH:
## Motivations

ComboboxContent was getting very busy and hard to read. also, we got to a place where we were getting heavy on integration tests (that's probably my fault tbh) so I wanted to see about shifting that towards unit tests where it made sense

### Changed

- extracted the list and search subcomponents into their own files with their own styles
- added unit test files for these components, and moved any tests that were more integration or acceptance style into these where it made sense, to reduce their fragility
- reorganized the contents of `useComboboxAccessibility` and `useComboboxContent` where generic behavior of Combobox is (mostly) in the latter, and things accessibility related in the former 
- moved those hooks into their own folder and files to keep the ComboboxContent smaller and focused
- changed from using state + having it as a dependency in the `useEffect` to avoid stale values in the handler, to using a ref

https://stackoverflow.com/questions/53845595/wrong-react-hooks-behaviour-with-event-listener

## Testing

should be functionally identical. nothing should have changed at all!

I'm pretty confident in our test coverage which appears to be happy with this, but even still - give it a spin in the webstories. make sure it's doing all the things we'd expect including some of the more convoluted flows such as

- type in search to find something low on the list that you'd normally have to scroll to, select it with click/arrow keys and enter or space
- re-open the combobox, see that we're scrolled to a place where we can see our selection
- start typing a new query in search
- move the arrows down, see that it moves to the first position
- tab or click on the clear, see the list go back to normal state
- use the arrows again to move to the top of this new list
- make a selection
- see the selection is there!

**Unit test changes**

are these still valid, valuable, do you have the same level of confidence with them as unit tests? lemme know what you think!

---

Random photo of Atlantis


ZakaryH:
I kinda just plopped all the subcomponents, their stylesheets and tests into the `ComboboxCOntent` directory

I did this because the `components` folder is currently _real_ components eg. Trigger, Action and Content. it felt weird to me somehow to move these subcomponents in there since they are bits and pieces of `Content` only. not totally opposed to moving them because it is busy in there now!

taylorvnoj (approved):
LGTM let's get this bebe in the feature branch

scotttjob:
> I kinda just plopped all the subcomponents, their stylesheets and tests into the `ComboboxCOntent` directory
> 
> I did this because the `components` folder is currently _real_ components eg. Trigger, Action and Content. it felt weird to me somehow to move these subcomponents in there since they are bits and pieces of `Content` only. not totally opposed to moving them because it is busy in there now!
> 
> alternatively we could have the components still in `ComboboxContent` but maybe under their own folders eg. `ComboboxContent/ComboboxSearch`
> 
> looking for feedback on this

Because we already have the pattern in ComboboxTrigger, I'm definitely okay with moving List/Search into their own sub-folders under ComboboxContent. We would just need to prefix them like in ComboboxTrigger (just to keep the pattern consistent).

I would also like to leave a note here, that this is the maximum number of sub-folder levels I would personally be comfortable with, but as always, open to having my mind changed.

`${Root}/src/MainComponent/components/ChildComponent/GrandChildComponent`

If we ever need anything deeper than GrandChildComponent, there's probably a solid argument to be made that we got lost somewhere in the design process. Also the naming convention should follow through like in the Trigger folder so we know right away when viewing a component where it sits in the hierarchy.

PR #1551, opened 2023-09-29
fix(components): Expose Link component props

MichaelParadis:
## Motivations

 The Link component doesn't export the prop type. We should export the type so it can be used by our consumers. This Pull Request fixes that.

### Added

- `LinkProps` type to the exports of the Link components.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Link/Link.tsx
+++ b/packages/components/src/Link/Link.tsx
@@ -1,6 +1,6 @@
 import React, { PropsWithChildren } from "react";
 
-interface LinkProps {
+export interface LinkProps {
   readonly url: string;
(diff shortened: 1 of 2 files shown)


darryltec (approved):
For context, what was the reason why we wanted to do this? Is it for prop drilling and such?

MichaelParadis:
@darryltec It is related to a component that can take `ButtonProps` or a `Link` and I wanted to type the interface correctly.

PR #1554, opened 2023-09-29
fix(components): Always Allow Combobox Gradient to Show

ZakaryH:
## Motivations

our gradients depended on actions being present to show up. this meant that if you had a long list of options, but no actions we wouldn't have that gradient we are using to visually suggest that there are more options to check out

also, the gradient code was in 2 different style files: search and combobox content, all the while it was really related to the list component

### Changed

- 

moved all the gradient styles into `ComboboxList.css` which make it a bit more intuitive hopefully since that's where the gradient visually show up and serve their purpose

since it no longer depends on actions - it will be there regardless of if you have actions or not! 🎉 

lastly I noticed that when using the arrow keys to navigate the list, the yellow focus outline was getting cut off on the left and ride sides due to the padding. to fix this I removed the horizontal padding and gave those list items some margin instead.

then gave the empty/no result messages some padding to make up for that since they just don't have the same problem as they are not focusable! (🤔 should they be? that might be not good for accessibility)

## Testing

just look at it! look how nice!

image

image

image

image

image

**notice there are no actions here**
image

these are all the relevant scenarios. the gradient should be in the same place as before, but it'll be there without actions now. everything else should be the same too, except for that focused state on the options it should not get cut off now.

---

Random photo of Atlantis


chris-at-jobber:
The gradient is present in cases when I don't expect it, and is taking up more space. 

image
(We do have a ticket that addresses this issue so if this PR is intentionally _not_ addressing that, disregard)

image
image
In the case of the list/actions boundary, it's less problematic because there's no clear boundary between the two, but it is noticeable when there are no actions and the gradient appears at the bottom of the popover.

taylorvnoj:
On empty/no result messages being focusable - empty state works nicely with the voice to text. No results doesn't spell it out for you with voice to text but I think the inability to arrow down and focus on anything would possibly indicate that there are no results?

@chris-at-jobber - yes, Zak was really just scratching an itch on Friday and wanted to apply the fade when no actions are present, while leaving fade on scroll for the future ticket

taylorvnoj:
@chris-at-jobber about the padding - I think I'm confused.  
When you open the combobox, there should be 8px padding between the search bar and the first option? Then upon scroll we'll see the gradient and the padding should no longer be present?
Or are we looking to remove that padding after/before

Screenshot 2023-10-02 at 4 03 54 PM

chris-at-jobber:
Unless there's an interaction, the empty state shouldn't be focusable. Screen-readers have other means to navigate non-interactive content so we should be fine there, as long as it's a text element it should be picked up.

chris-at-jobber:
@taylorvnoj I'll have a prototype that hopefully indicates the intended scrolling/gradient up shortly 🔜

chris-at-jobber:
Example here: https://www.figma.com/proto/gQNDrpJYRFAwfbP3jRO0an/DSC-%5B-Combobox-%5D?page-id=2558%3A3417&type=design&node-id=2558-6259&viewport=561%2C757%2C1&t=jwOKnrUDtbHfVfY6-1&scaling=min-zoom&starting-point-node-id=2558%3A6259&mode=design

ZakaryH:
I swear these are identical
image
image


image
image

where the first image in each set is master, and the following one is on this branch

scotttjob:
I'd love to help review this PR and get it closed/merged, but is it supposed to be targeting master? I can't really see what changed in this PR, it's just all of the combobox code. 

I also agree with Zak above that if we need to address larger design issues within combobox that's cool and we definitely should do that, but this PR in specific was about moving the combobox gradient to the list only, and have it not be dependant on actions (which is smart).

If we can update this PR so I only see what has changed, I'd love to review it (or Zak if you want, let's hop on a call and you can just show me visually somehow).

chris-at-jobber:
👍 I think I misunderstood the PR description, disregard my comment (within the scope of this PR)

ZakaryH:
>If we can update this PR so I only see what has changed, I'd love to review it (or Zak if you want, let's hop on a call and you can just show me visually somehow).

@scotttjob that should be good now with the files diff. unfortunately due to our merge process it's still showing all the commits from other work. if we don't love that, I can rebase or just start a new branch since this is 2 commits that are very easy to cherry pick and apply. let me know what you think!

>👍 I think I misunderstood the PR description, disregard my comment (within the scope of this PR)

🎉 🎉

scotttjob (approved):
LGTM!. Does what it says it's doing. :)

PR #1555, opened 2023-10-02
refactor(components): update tooltip styling and reduce bounce [JOB-75936]

chris-at-jobber:
## Motivations

The visual design of the Tooltip has changed, so our coded components should as well.

### Changed

- Tooltip animates in smoothly with no "bounce"
- Tooltip is now dark (`surface--reverse`) with white text (`text--reverse`)
- Padding has slightly changed

## Testing

View the [Tooltip docs](https://visual-updates-to-tooltip.atlantis.pages.dev/?path=/story/components-overlays-tooltip-web--basic) or test it out on Button/other components that use Tooltip.

---

Random photo of Atlantis


packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -65,11 +64,11 @@ export function Tooltip({ message, children }: TooltipProps) {
...
               transition={{
                 type: "spring",
-                damping: 20,
-                stiffness: 300,
+                damping: 50,
+                stiffness: 500,
               }}
             >
-              <Text>{message}</Text>
+              <p className={styles.tooltipMessage}>{message}</p>
```
rodrigoeidelvein, line 71: I wonder if this type of text is something that we will be seeing more in the future and if it's worth making a new variation of text called "reverse".

rodrigoeidelvein (commented):
Just have one thought, other than that, works well and looks good!

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -65,11 +64,11 @@ export function Tooltip({ message, children }: TooltipProps) {
               exit="startOrStop"
               transition={{
                 type: "spring",
```
darryltec, line 66: just need to remove this and it will stop bouncing

packages/components/src/Tooltip/Tooltip.tsx
chris-at-jobber: That actually might be helpful, I think it might need a bit more thought (is it a variant of text, new component, etc.) than the scope of this PR because this classname also handles _font weight_, so does this become a new weight variant of Text? (and then what do we name that property, etc).

The `reverseTheme` part is a Just Do It in my opinion, but I'm not super equipped to wrangle the props in Text on my own 😅

packages/components/src/Tooltip/Tooltip.tsx
chris-at-jobber: I modified the animation so it still _moves_, it's just not so "bouncy"

darryltec: oh, if you remove that, it will fallback to tween, which is the `ease` in css

chris-at-jobber: Do we need to remove it? It's working as I'd expect right now with the new damping and stiffness values

https://github.com/GetJobber/atlantis/assets/39704901/203a9129-bfad-4a75-b29c-4280ae6b1ddf

chris-at-jobber: OK so what is the recommended change? Do I need to delete the entire Framer Motion div, or just change some of these declarations?

darryltec: just need to remove the type

chris-at-jobber: Perfect, will do!

darryltec (approved):
:shipit:

PR #1556, opened 2023-10-02
feat(components): add Flex component for web

mattander:
## Motivations

As we've been working on the [job insights](https://www.figma.com/file/jt6FSbYP0SP8ToSC2rjdAQ/SP-KPI-Revenue%2FProfit-Trends%3A-Design?type=design&node-id=2345-263588&mode=design&t=yKcLWcMQyrFoF0bN-0) feature, the Workday team has run into a few situations where the design has two components aligned on the same line. These have all been easily achieved with a flexbox container, but that has meant using a wrapper and custom class + styles each time. 

## What this component is for

This component is for achieving complex or nuanced layouts inside components and page layouts. Aligning a button next to a heading, spacing two things on opposite ends of a line, etc. 

### Added

- Flex container for web

---

### Picture from Atlanta
image


darryltec (changes requested):
Let's take a step back and narrow down what a Flex component is. 

If you look at the components native one, it only have 3 props on it https://github.com/GetJobber/atlantis/blob/78ec7dc1183f541c32af84cf398ec63777d637aa/packages/components-native/src/Flex/types.ts#L5-L29

Mostly because it isn't trying to encapsulate what a CSS flex is or create the most flexible/powerful component. It's mostly to make it easy to put things side-by-side (row) or potentially create a layout (row/column).

Consider this scenario where you want to put an avatar on the left, a text in the middle, and an icon on the right. The avatar and icon only takes up its own width as its space, while the middle text fills the remaining space

image


Pretty common UI you'd see everywhere. Which would be written as 

```tsx
<Flex template={["shrink", "grow", "shrink"]}>
  <Avatar />
  <Text />
  <Icon />
</Flex>
```

### Let's rotate that

image

A pretty common layout you'd see. Which would be written the same way 

```tsx
<Flex template={["shrink", "grow", "shrink"]} direction="column">
  <TopBar />
  <Content />
  <BottomBar />
</Flex>
```

### Let's rotate it again

image

Another common layout you'd see. Which is pretty much written the same way

```tsx
<Flex template={["shrink", "grow", "shrink"]}>
  <Nav />
  <Content />
  <SideSheet />
</Flex>
```
### And of course, not as common but totally exist

image

```tsx
<Flex template={["grow", "grow", "grow"]}>
  {options.map(option => <Option {...option} />)}
</Flex>
```

### Then there's the nested ones

image

```tsx
<Flex template={["shrink", "grow", "shrink"]}>
  <Avatar />

  <Content>
    <Flex template={["grow", "shrink"]}>
      <Label />
      <Status />
    </Flex>
    <Flex template={["grow", "shrink"]}>
      <Description />
      <Date />
    </Flex>
  </Content>

  <Icon />
</Flex>
```

### With all that said

The best approach to this is to only implement the ones we need. Everything else is easy to add when we do need to support it. 

So the props can start with
- The same 3 from components native.
- `direction` to change the flow

The rest can be removed and talked about if we do need them. 

- Width and height should be done on the parent if you do need to. 
  - Responsive design wise, you should rarely add one of these. 
- The `type` makes sense for content but not for creating layouts since those are either a child or a parent of this component.
- `testId` are either a parent thing or a child thing again. You shouldn't have to add a test id on this component since you have access to everything it does. If you're testing if the `Flex` is doing what it's supposed to, you're essentially testing what is already tested here.
- you should just have `gap` having different ones on each row/column creates a weird flow issue. There will be a case where it needs to be but that's more of an individual component type thing rather than putting everything under flex.
- `padding` should be a wrapping component. If the issue is "I don't want to write CSS" We should have a component for adding that like `Box` since that argues we should have a padding prop on everything (which we shouldn't)

packages/components/src/Flex/types.ts
```diff
@@ -0,0 +1,13 @@
+export type ColumnKeys = "shrink" | "grow" | "auto";
```
mattander, line 1: @darryltec Are you okay with surfacing `auto` as well? Since templates will convert this to a grid container, sometimes these will be necessary to accomplish a layout. 

For ref, grow is `1fr`, shrink is `min-content` and auto is `auto` (obvi).

darryltec: use `max-content` for shrink so it doesn't squish the elements into oblivion

1fr and auto are closely the same so either grow is 1fr or auto

mattander: Would we never want an element to shrink down to its `min-content` to fit in a tight spot?

image

darryltec: that screenshot is max-content, no?

https://codepen.io/notdarryltec/pen/GRPYMda

pretty rare for us to actually allow this 
image

darryltec: i am cool with that

packages/components/src/Content/Content.tsx
```diff
@@ -16,7 +16,7 @@ interface ContentProps {
    * Change the wrapping element to be one of the available
    * semantic tags.
    *
-   * @default 'div'
+   * @default div
```
mattander, line 19: campground cleanup.

docs/components/Flex/Flex.stories.mdx
```diff
@@ -1,12 +1,10 @@
 import { Meta } from "@storybook/addon-docs";
-import { Flex } from "@jobber/components-native";
 
-<Meta title="Components/Layouts and Structure/Flex/Docs" component={Flex} />
```
mattander, line 4: Does removing this do anything? I saw no difference.

packages/components/src/Content/Content.tsx
```diff
@@ -16,7 +16,7 @@ interface ContentProps {
    * Change the wrapping element to be one of the available
    * semantic tags.
    *
-   * @default 'div'
+   * @default "div"
```
mattander, line 19: Cleanup

mattander:
@chris-at-jobber sorry, just getting back to this now. 

This is quite odd. I'm not seeing this at all. 

https://github.com/GetJobber/atlantis/assets/10996915/128d8533-5eb5-4d09-a789-154fb8915df5

Also great suggestion on adding some column keys. I'll do that.

docs/components/Flex/Flex.stories.mdx
darryltec: it used to. not anymore tho because we don't show the props table on mdx

packages/components/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  return (
+    <div
+      className={classnames(
+        template.length ? styles.grid : styles.flex,
+        template.length === 0 && direction === "column" ? styles.column : null,
+        gap ? styles[`${gap}Gap`] : null,
+        align ? styles[`${align}Align`] : null,
+      )}
```
darryltec, line 80: this is okay, but there's a cleaner way to write this with fewer ternaries

```suggestion
      className={classnames(template.length ? styles.grid : styles.flex, {
        [styles.column]: template.length === 0 && direction === "column",
        [styles[`${gap}Gap`]]: Boolean(gap),
        [styles[`${align}Align`]]: Boolean(align),
      })}
```

packages/components/src/Flex/types.ts
darryltec, line 1: We name these things as `Flex.types.ts` now

packages/components/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,86 @@
...
+  children,
+  direction = "row",
+  gap = "base",
+  template = [],
+}: FlexProps) {
+  if (template.length && template.length !== Children.count(children)) {
+    console.warn(
+      "Flex template length does not match children count, this may cause unexpected results.",
+    );
+  }
```
darryltec, line 60: i don't think we need to say this. it's pretty obvious what it's doing when you have 3 child and you only added 2 template

packages/components/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const templateKey =
+    direction === "row" ? "gridTemplateColumns" : "gridTemplateRows";
+
+  const containerStyles: CSSProperties = {};
+
+  if (template.length) {
+    containerStyles[templateKey] = template
+      .map(key => templateValues[key])
+      .join(" ");
+  }
```
darryltec, line 71: i'd put group this on its own function so it's easier to spot that they're all related to each other

darryltec (changes requested):
we're pretty close, just need to narrow down some parts and code cleanup

packages/components/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,86 @@
...
+   *
+   * **Supported keys**
+   * - `"grow"` - Grows to the space available. If all children are set to
+   *   grow, then they'll have equal width.
+   * - `"shrink"` - Shrinks to the smallest size possible. Normally the size of
+   *   the child.
+   *
+   * By default, this will set every children to grow in equal widths.
+   */
+  readonly template?: ColumnKeys[];
```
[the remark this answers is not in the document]
mattander: Both of these decisions stem from the same cause. `display: flex` defaults are great for a non-wrapping row, and don't require anything beyond that original `display: flex` to accomplish it while still supporting align and justify and whatnot. 

However, as soon as you set a template, it becomes harder to use flexbox without having to modify `Children` (wrapping them in `flex: grow shrink basis;` wrappers. React is pretty clear about only messing with children as a last resort and grid makes it really easy to make what looks like a flexbox with child sizing controlled from the parent (with columns/rows). 

So my thinking was that for basic layouts, flexbox does it's thing best with no overrides, but for templates grid quickly becomes the better solution.

packages/components/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,86 @@
+import React, { CSSProperties, Children, ReactNode } from "react";
+// import chunk from "lodash/chunk";
+import classnames from "classnames";
+import { ColumnKeys, Spacing } from "./types";
+import styles from "./Flex.css";
+
+interface FlexProps {
+  children: ReactNode;
+
```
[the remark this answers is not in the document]
mattander: Ah thank you.

packages/components/src/Flex/Flex.tsx
mattander: Thank you! That `Boolean` constructor was the piece I was missing when trying to do this. Didn't even think of it!

packages/components/src/Flex/Flex.test.tsx
```diff
@@ -0,0 +1,69 @@
+import React from "react";
+import { cleanup, render } from "@testing-library/react";
+import { Flex } from ".";
+
+afterEach(cleanup);
+
+it("renders the flexible container", () => {
```
darryltec, line 7: Wrap all of these test in a describe please! Makes it easier for our friends to see who these tests belong to at a glance when it runs with the others

packages/components/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return containerStyles;
+}
+
+export function Flex({
+  align = "center",
+  children,
+  direction = "row",
+  gap = "base",
+  template = [],
```
darryltec, line 62: since this is required, don't need the default

darryltec (changes requested):
Ok one last thing and this is good to go!

packages/components/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,75 @@
...
+  readonly direction?: "row" | "column";
+}
+
+function generateGridStylesFromTemplate(
+  flowDirection: string,
+  layoutTemplate: ColumnKeys[],
+): CSSProperties {
+  const containerStyles: CSSProperties = {};
+  const templateKey =
+    flowDirection === "row" ? "gridTemplateColumns" : "gridTemplateRows";
```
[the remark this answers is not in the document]
mattander: I like this but unfortunately TS seems to find this too abstract and can't infer the key anymore. Even if I specify that `flowDirection` is a union of `"row"` and `"column"` it still says you can't use any string. 
image

Even if I reassure react that those record values in the key map are `CSSProperties` keys, it just says its too complicated lol.
image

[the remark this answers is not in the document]
mattander: I found one way to make TS happy. Gotta use an enum so it's more confident in the possible values. 

```
function generateGridStylesFromTemplate(
  direction: Direction,
  layoutTemplate: ColumnKeys[],
): CSSProperties {
  const containerStyles: CSSProperties = {};
  const templateValues = {
    grow: "1fr",
    shrink: "max-content",
  }
  enum TemplateKeys {
    row = "gridTemplateColumns",
    column = "gridTemplateRows",
  }

  containerStyles[TemplateKeys[direction]] = layoutTemplate
    .map(key => templateValues[key])
    .join(" ");

  return containerStyles;
}
```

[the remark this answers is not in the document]
mattander: UPDATE: darryl taught me about `as const` and now I feel like I've learned some deep magic.

packages/components/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  containerStyles[templateKey] = layoutTemplate
+    .map(key => (key === "grow" ? "1fr" : "max-content"))
```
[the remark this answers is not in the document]
mattander: added back along side the template key change above.

packages/components/src/Flex/Flex.tsx
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * Determine how the children gets laid out
+   *
+   * **Supported keys**
+   * - `"grow"` - Grows to the space available. If all children are set to
+   *   grow, then they'll have equal width.
+   * - `"shrink"` - Shrinks to the smallest size possible. Normally the size of
+   *   the child.
+   *
+   * By default, this will set every children to grow in equal widths.
```
darryltec, line 17: can just delete this, but that's super minor

```suggestion
```

darryltec (approved):
Looks good mah dude! Pending design review tho.

Also, once this is out, we need to add an example for full height layout of app bar, content, bottom bar to show that it can support that as well.

packages/components/src/Flex/Flex.tsx
mattander: Sounds good. Committed.

mattander:
@chris-at-jobber I think we're ready for design review. Note @darryltec's comment [here](https://github.com/GetJobber/atlantis/pull/1556#pullrequestreview-1685070551), that we'll add more examples once we've got this merged.

docs/components/Flex/Flex.stories.mdx
```diff
@@ -25,4 +23,6 @@ further control the layout of the children.
 ## Related Components
```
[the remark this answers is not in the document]
chris-at-jobber: One other docs thought, and I know we're just re-using the same example from mobile, but in the "nested" example even after playing with the props I'm still not 100% sure which elements are actually being adjusted other than I can see the space between the `arrowRight` icon and the rest of the icon changing, or stacking on `direction=column`. Either a simpler example, or some kind of visual indicator (custom background color?) to show which portion is being acted on by the props would make it easier to parse how the props are acting.

image

chris-at-jobber (commented):
Design approved ✅ This component works great, thanks @mattander ! I have a few documentation comments but they're probably more for me/us to tackle as fast-follows unless you'd like to take that on.

docs/components/Flex/Flex.stories.mdx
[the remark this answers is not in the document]
mattander: > Minor and probably a followup, but adding Content as another related component for even simpler layouts might be helpful

Agreed. This would be a great addition as a fast follow.

>Either a simpler example, or some kind of visual indicator (custom background color?) to show which portion is being acted on by the props would make it easier to parse how the props are acting.

Funny you mention that because i considered exactly that. I'm still mulling over the idea of background colours just to explain where stuff is going.

PR #1557, opened 2023-10-02
feat(components): Add multiselect prop to combobox

dlrussel:
## Motivations
This PR adds the prop for the multi select functionality for the combobox component.
This is the first step for adding multi select functionality to the combobox. It will allow us to have a variable to hook off of in order to build the following multi select functionality.

Random photo of Atlantis

### DIFF
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -139,7 +139,7 @@ const ComboboxClearSelection: ComponentStory<typeof Combobox> = args => {
         type="primary"
         onClick={() => setSelected(null)}
       />
-      <Combobox {...args}>
+      <Combobox {...args} multiSelect>
         <Combobox.TriggerButton
           label="Select a teammate"
           variation="subtle"
--- a/packages/components/src/Combobox/Combobox.tsx
+++ b/packages/components/src/Combobox/Combobox.tsx
@@ -29,7 +29,7 @@ export const Combobox = (props: ComboboxProps): JSX.Element => {
     props.children,
   );
   return (
-    <ComboboxContextProvider>
+    <ComboboxContextProvider multiselect={props.multiSelect}>
       {triggerElement}
       {contentElement}
     </ComboboxContextProvider>
--- a/packages/components/src/Combobox/Combobox.tsx
+++ b/packages/components/src/Combobox/Combobox.tsx
@@ -12,10 +12,9 @@ export interface ComboboxProps {
   readonly children: ReactElement | ReactElement[];
 
   /**
-   * multi select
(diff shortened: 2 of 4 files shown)


packages/components/src/Combobox/Combobox.tsx
```diff
@@ -10,6 +10,14 @@ import {
...
 export interface ComboboxProps {
   readonly children: ReactElement | ReactElement[];
+
+  /**
+   * multi select
+   * @default false
+   * @type boolean
+   * @description if true, allows multiple selections
+   */
+  readonly multiSelect?: boolean;
```
taylorvnoj, line 20: Is there a reason why we explicitly set `@description` and `@type`? We should be able to follow our normal props pattern:

```suggestion
  /**
   * When `true`, `Combobox` will allow for multiple selections
   *
   * @default false
   */
  readonly multiSelect?: boolean;
```

taylorvnoj (changes requested):
Looks good - just one suggestion to avoid doubling up on defining type and description

packages/components/src/Combobox/ComboboxProvider.tsx
```diff
@@ -11,16 +12,20 @@ export const ComboboxContext = React.createContext(
 
 export interface ComboboxProviderProps {
   readonly children: React.ReactNode;
+  readonly multiselect?: boolean;
```
ZakaryH, line 15: we've defined the prop for multiselect on the provider, but nothing is passing that value yet. let's add that small change so any instances doing `<Combobox multiselect />` truly have `multiselect` available to them inside `Combobox` via the Provider

packages/components/src/Combobox/Combobox.tsx
dlrussel: ah ok!!!! 👌🏻

docs/components/Combobox/Web.stories.tsx
```diff
@@ -139,7 +139,7 @@ const ComboboxClearSelection: ComponentStory<typeof Combobox> = args => {
         type="primary"
         onClick={() => setSelected(null)}
       />
-      <Combobox {...args}>
+      <Combobox {...args} multiSelect>
```
ZakaryH, line 142: ah whoops this was from me adding it during the screen share. I'm fine with removing it later if you are

this web story is still valid since it's about clearing selections, plus it does nothing functionally at this point. just a little thing we'll need to remember to tidy up later

taylorvnoj (approved):
🚢

docs/components/Combobox/Web.stories.tsx
dlrussel: ok yaaa lets tidy in the next PR!

PR #1560, opened 2023-10-03
refactor(components): reduce the amount of reduce we do in DataList

darryltec:
## Motivations

This started off as fixing that issue with the layout not changing when the screen changes from wide to small (or vice-versa) but it went more than that now. It ended up removing the need to use a reducer and coupling of layout, header, and loading state. Which in turn made (IMO) the code more readable, organized, and reactive to change with fewer maps and reducers.

This also changed how we deal with the loading state on mobile and closely matches the layout now instead of using a glimmer text.

### Added

- Registering layouts to the context
- Hooks for determining which layout to use
- Separate components for list items, list item, and headers

### Changed

- How we handle determining which layout to use

### Deprecated

- 

### Removed

- unused components

### Fixed

- 

### Security

- 

## Testing

Aside from the loading state on smaller screens, there should be no change to how it should show up on the user (sans bugs).

---

Random photo of Atlantis


packages/components/src/DataList/hooks/useActiveLayout.ts
```diff
@@ -0,0 +1,15 @@
+import { useResponsiveSizing } from "./useResponsiveSizing";
+import { useDataListContext } from "../context/DataListContext";
+
+export function useActiveLayout() {
+  const { layoutBreakpoints, layouts } = useDataListContext();
+  const breakpoints = useResponsiveSizing();
+  const reversedLayoutBreakpoints = [...layoutBreakpoints].reverse();
```
rodrigoeidelvein, line 7: Why did you revert it?

darryltec: So it checks the highest breakpoint first and see if it's true

packages/components/src/DataList/components/DataListLayoutActions/DataListLayoutActions.tsx
```diff
@@ -25,6 +26,10 @@ export function DataListLayoutActions() {
     };
   }, []);
 
+  if (loadingState === "initial") {
+    return <Glimmer shape="square" size="large" />;
+  }
+
```
rodrigoeidelvein, line 32: Loved it!

packages/components/src/DataList/DataList.tsx
```diff
@@ -132,33 +151,14 @@ function InternalDataList() {
...
-      {initialLoading && (
-        <DataListLoadingState
-          headers={headers}
-          layouts={layoutComponents}
-          mediaMatches={mediaMatches}
-        />
-      )}
+      {initialLoading && <DataListLoadingState />}
 
       {showEmptyState && <InternalDataListEmptyState />}
```
rodrigoeidelvein, line 159: An improvement that could be made here is to check if we should render these components inside the component itself, as we're doing for DataListHeader, and others.

But, I can also see the point of being more readable this way.

rodrigoeidelvein (approved):
Just a minor comment with a suggestion that might don't make sense actually haha just food for thought

Looks really good, cleaner and works the same way! Have you done any performance test to see if removing the hook with the reduces changed anything?

packages/components/src/DataList/DataList.tsx
darryltec: Hmmm that's definitely something to think about!

darryltec: I'll merge this first tho

darryltec:
> Have you done any performance test to see if removing the hook with the reduces changed anything?

Yup. mostly that i couldn't replicate the issue with the layout not changing anymore compared to before.

PR #1563, opened 2023-10-03
build: Fix warning from the rollup step

rodrigoeidelvein:
## Motivations

We recently started to use `lodash/noop`, `lodash/isEmpty`, and `@jobber/design/foundations` as external dependencies, so rollup was not happy about it. So here I'm telling it that we are using these new external dependencies. 

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Warning in the build phase

### Security

- 

### Before
image

### After
no warning

---

Random photo of Atlantis


packages/components/rollup.config.js
```diff
@@ -65,9 +65,12 @@ export default {
     "lodash/groupBy",
     "lodash/omit",
     "lodash/sortBy",
+    "lodash/noop",
+    "lodash/isEmpty",
```
darryltec, line 69: I wonder if these + the 2 above should just be
```suggestion
    new RegExp("lodash/.*"),
```

darryltec (approved):
can confirm, warning gone!

PR #1564, opened 2023-10-04
fix(components): nested button issue on DataList

darryltec:
## Motivations

When the DataList items are clickable, it becomes a button. Unfortunately, you can also add a button inside it, like the layout actions, so it will fire both the item click and the menu/menu item click. 

So you'll have 2 problems
- HTML would throw an error around a button inside a button not allowed
- The item click fires with the menu click

This PR prevents that from happening.

### Added

- `stopPropagation` on places that need it to prevent event bubbling

### Changed

- The wrapping button to a div with a role of button and acts like a button to avoid the button on button issue

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListActionsMenu/DataListActionsMenu.tsx
+++ b/packages/components/src/DataList/components/DataListActionsMenu/DataListActionsMenu.tsx
@@ -1,4 +1,9 @@
-import React, { CSSProperties, PropsWithChildren, useState } from "react";
+import React, {
+  CSSProperties,
+  MouseEvent,
+  PropsWithChildren,
+  useState,
+} from "react";
 import { AnimatePresence, Variants, motion } from "framer-motion";
 import { useFocusTrap } from "@jobber/hooks/useFocusTrap";
 import { useRefocusOnActivator } from "@jobber/hooks/useRefocusOnActivator";
@@ -37,7 +42,7 @@ export function DataListActionsMenu({
   return createPortal(
     <AnimatePresence>
       {visible && (
-        <div ref={focusTrapRef}>
+        <div ref={focusTrapRef} onClick={handleClick}>
           <motion.div
             role="menu"
             ref={setRef}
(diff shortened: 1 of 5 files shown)


darryltec:
> you end up with button siblings, instead of a parent/child relationship.

@scotttjob That's a classic way of fixing this, but you'd lose out on the button's label when screen readers come into play. 

Right now with how it's set up, it'll read the content of the item. (Not gracefully, but it gets the job done.) It being a child of the item and a sibling of the other action would mean we'd have to give it a meaningful label. The problem is, we don't know what content goes in here so what is a "meaningful label"?

On top of that, it's not always a button. It could be a link, so that has to be a wrapping element since that comes with a lot of accessibility features (cmd + click to open in new tab is one of them). 

That's not to say we couldn't do it, we totally could. But as with everything, it comes with other issues like z-index, tabbing order, styling issues when you programmatically overlay the button/link on top of everything.

scotttjob:
> > you end up with button siblings, instead of a parent/child relationship.
> 
> @scotttjob That's a classic way of fixing this, but you'd lose out on the button's label when screen readers come into play.
> 
> Right now with how it's set up, it'll read the content of the item. (Not gracefully, but it gets the job done.) It being a child of the item and a sibling of the other action would mean we'd have to give it a meaningful label. The problem is, we don't know what content goes in here so what is a "meaningful label"?
> 
> On top of that, it's not always a button. It could be a link, so that has to be a wrapping element since that comes with a lot of accessibility features (cmd + click to open in new tab is one of them).
> 
> That's not to say we couldn't do it, we totally could. But as with everything, it comes with other issues like z-index, tabbing order, styling issues when you programmatically overlay the button/link on top of everything.

Got it, the existing setup/design made it so the stopPropagation call was infinitely easier/faster than refactoring everything for a possibly worse solution (or at best, parity). Makes sense Darryl, thanks! :)

PR #1565, opened 2023-10-04
feat(hooks): create breakpoints hook

darryltec:
## Motivations

Make it easier for devs to use the same breakpoints as the CSS counterpart. Instead of having to add a listener to `window.resize` and matching it up with a random number, devs can just use this hook to check if they should render their component on our supported breakpoints.

### Added

- `useBreakpoints` hook that gives you the sizes that we support
- `mockViewportWidth` to easily test it on your own codebase

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/design/src/ResponsiveBreakpoint.stories.mdx
+++ b/packages/design/src/ResponsiveBreakpoint.stories.mdx
@@ -5,18 +5,16 @@ import { Banner } from "@jobber/components/Banner";
 
 # Responsive Breakpoints
 
-<Banner type="notice" dismissible={false}>
+We have 3 different ways to deal with responsive breakpoints in our components.
 
-To make components responsive depending on the size of its container rather than
-the browser, you should use the
-[useResizeObserver hook](../?path=/docs/hooks-useresizeobserver--use-resize-observer)
-so that the component can be responsive based on its container.
+1. CSS variables (`@custom-media`) that are relative to the browser size.
+2. [`useBreakpoints` hook](..) that is the JS counterpart of the CSS variables.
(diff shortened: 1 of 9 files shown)


MichaelParadis:
I love this idea just going to say shipit without even looking at the code 🤣

MichaelParadis:
@rodrigoeidelvein I may be able to take a look tomorrow. I am super busy with my team's sprint work. Don't hold this up for my review :)

darryltec:
@rodrigoeidelvein realized that the smallAndBelow should actually be extraSmallOnly. Though it didn't clear your review but would be good to get a sanity check on that one

rodrigoeidelvein:
> @rodrigoeidelvein realized that the smallAndBelow should actually be extraSmallOnly. Though it didn't clear your review but would be good to get a sanity check on that one

@darryltec I can see the point of renaming it to `extraSmallOnly`, but should we? Because the CSS variable is called `--small-screens-and-below`, maybe we should keep it similar because we're using those values as reference too.

darryltec:
@rodrigoeidelvein I was okay with it because `only` doesn't really exist on the CSS breakpoint and it's a JS specific thing since it doesn't cascade like CSS does.

Also, might need to update that one on design as well since it is also semantically wrong haha

rodrigoeidelvein:
@darryltec Okay, it makes sense to me. But let's not update the CSS variable now, I don't know if someone is already using those CSS variables, but seems risky.

LGTM

darryltec:
@rodrigoeidelvein agreed!

PR #1569, opened 2023-10-04
feat(components): MultiSelect Combobox

ZakaryH:
## Motivations

we would like to be able to make multiple selections with a Combobox

### Added

aria-multiselectable: pretty self explanatory. helps with accessibility in the multi selectable case.

the `onClose` prop

this is a one or the other type situation with `onSelect`. I can't think of a case where you'd ever want both. feel free to disagree if you think that's valid! to that end, I've added some more XOR 🙀 so you can't have both at least at a TypeScript level

then, I've added some internal state to operate on in the `onClose` case because we explicitly cannot use the external state since that's the whole point of this prop, to not update it until we've closed.

this wasn't too bad honestly, we either use the `selected` we were given, or we initialize the internal state with that and then start working on an internal, possibly mutated version of that!

### Changed

- changed the input and output types of `selected` and `onSelect` to be arrays now, for single select it will be an array of size 1
- modified the logic to make sure we scroll down to the first selected element if there are more than one
- modified selection behavior for multiple where we no longer want to clear the search nor close the listbox

## Testing

**Multi Select general**

with an instance of Combobox having `multiselect` or `multiselect={true}`:

- should be able to make selections without the content closing as you select or un-select
- use the search, make a selection, search should not clear after the selection

**Multi Select onSelect**

same thing but with `onSelect` specifically as the callback mechanism to listen to updates

- selections should immediately be passed onto the consumer via the `onSelect` callback

**Multi Select onClose**
same thing but with `onClose` instead

- selections should NOT immediately be passed to the consumer. instead, make as many selections and edits as you please, only upon closing should you receive the final selections 

**Single Select onClose**

this one is weird, but if we do keep `onClose` for single select (if we don't want that, we'll have to re-jig where some props live but still do-able) then it's going to do the exact same thing as `onSelect`

---

Random photo of Atlantis


packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -115,8 +117,24 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
...
+  }
+}
+
+function handleMultiSelect(
+  selectCallback: (selected: ComboboxOption[]) => void,
+  selected: ComboboxOption[],
+  selection: ComboboxOption,
+) {
+  if (selected.some(s => s.id === selection.id)) {
+    selectCallback(selected.filter(s => s.id !== selection.id));
```
ZakaryH, line 136: thinking about how we might optimize this

`some` is fine because we do need to find if it was already selected, and it short circuits once it finds a result

the `filter` though. is there a smarter way to do that that doesn't involve iterating over all of the items? 🤔

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -47,6 +47,16 @@ handling user interactions, and managing search functionality.
...
 
+#### Callback
+
+`onSelect` is the recommended callback handler that should be used in most cases. This will immediately inform
+the consumer of changes to selections as they happen.
+
+If however immediate changes are not desired, and instead it is preferable to wait until all selections or
+modifications to selections have been made before informing the consumer, `onClose` may be used. The arguments provided to this callback are
+Identical to those of `onSelect`, only the timing differs. Please note this will only
+have a noticeable effect with `multiSelect = true` since the Combobox closes immediately upon selection with single-select.
```
ZakaryH, line 58: I feel conflicted about this

I never like it when two props are incompatible, or one does nothing when the other is present. if they are incompatible I prefer to do something like XOR where you simply can't combine them if we know they aren't valid

in this case though, technically speaking `onClose` isn't incompatible, nor does not do anything. in fact it does the exact same thing as `onSelect` since for single select it closes when you select. so my question is, is this fine? do we need to do anything different? it's just odd that you have two different callbacks that do the same thing for single select

docs/components/Combobox/Web.stories.tsx
```diff
@@ -197,3 +249,6 @@ ClearSelection.args = {};
 
 export const EmptyState = ComboboxEmptyState.bind({});
 EmptyState.args = {};
+
+export const MultiSelect = ComboboxMultiSelection.bind({});
```
ZakaryH, line 253: is this a good enough name if it's demonstrating both multi select AND onClose?

should I just make another example with `onClose`? 

how many examples is too many?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -115,8 +133,28 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
...
+      ? props.onSelect
+      : setInternalSelected;
+
+    if (multiselect) {
+      handleMultiSelect(callbackHandler, selectedOptions, selection);
+    } else {
+      callbackHandler([selection]);
+      setSearchValue("");
+      setOpen(false);
+    }
```
ZakaryH, line 146: I don't know why I made a function for multi select and not single select when they're both like 3 lines. maybe I thought multi was going to be larger than it ended up being

open to changing this in whichever direction

packages/components/src/Combobox/hooks/useComboboxContent.ts
```diff
@@ -1,38 +1,57 @@
...
 
+  const [internalSelected, setInternalSelected] =
+    useState<ComboboxOption[]>(selected);
+  const selectedOptions = onClose ? internalSelected : selected;
+
+  useEffect(() => {
+    if (!open && onClose) {
+      onClose(selectedOptions);
+    }
+  }, [open, onClose]);
```
ZakaryH, line 38: this feels wrong to omit `selectedOptions` but this works fine.... why does that work though?

oh it's because of the `useState` of `internalSelected` that is getting modified and updating this hook. I should probably just add to the dependencies. if we ever moved that `useState` out of here I believe this would stop working, plus someday when we add the exhaustive dependencies eslint rule it's going to yell at us for not having it

ZakaryH:
**TODO**

add tests for these cases
 - multi select -> should allow multiple selections to be made
 - multi select -> should not close after making a selection
 - multi select -> should not clear search after making a selection
 - onClose -> should not update consumer on selections
 - onClose -> should update with modified selections when closed


are there other valuable things to test here? 🤔

ZakaryH:
could `onClose` be a useful hook of the components lifecycle to expose, even if you are using `onSelect`? I'm not so sure about my decision to make them exclusive anymore

you don't absolutely have to use the `selection` argument from `onClose`. you could totally be updating as selections happen with `onSelect` and then ignore that final `onClose` and do something else entirely in that callback

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -247,15 +251,185 @@ describe("Combobox selected value", () => {
...
+      fireEvent.click(option);
+      fireEvent.click(spoder);
+
+      expect(queryByText("Choice: Bilbo Baggins")).not.toBeInTheDocument();
+      expect(queryByText("Choice: Shelob the Spoder")).not.toBeInTheDocument();
+    });
+  });
+});
+
+function MockMultiSelectOnCloseCombobox(props: {
```
ZakaryH, line 371: this got a bit acceptance-y

part of me wants to find ways to make them more unit or integration style tests, but also I feel like due to the nature of how the components work now, where the state is external and our selections depend on that state changing - we kinda have to do something like this

docs/components/Combobox/Web.stories.tsx
dlrussel: I think that functionality is technically demonstrated here so I don't think another example is necessary. I would lean more towards adding more documentation than adding another example if you think there's something else worth calling out!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
dlrussel: kk so I'm thinking about this and I wouldn't say this code is giving me any super big red flags in terms of efficiency. Mainly because we aren't dealing with any extremely large data sets here typically so we should be ok with this. Like one thing I was thinking is that if we really did want to optimize this we could always maintain a separate data structure using `Set` to store the object references to keep track of selected items by their IDs for faster lookups. then we could use the `has` method to check if the ID of the option is present? Let me know if that makes sense but I also think that might just be overkill considering what I mentioned initially.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
scotttjob: toss handleSingleSelect into a function if you're down for it. Might only be 3 lines now, but it could grow in the future and it'll be less anxiety-inducing to add a line in the future than refactor the whole thing.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
scotttjob: There's definitely room for improvement if we need it, but I side with Diane a bit here in that it could be  premature optimization that may never be needed. What we could do though, is mark this as a 'known area for improvement' so that if we do start to see complaints of combobox being slow, we'll know where to start looking.

scotttjob (approved):
I have left some comments here, but most of them are just discussion, some are straight up praise, and the rest aren't a big enough deal to hold up this work. Approved.

Overall, fantastic work on this! LGTM.

ZakaryH:
hey @scotttjob thanks for the review!

there's one point I'm still looking for feedback/thoughts from others on

>could onClose be a useful hook of the components lifecycle to expose, even if you are using onSelect? I'm not so sure about my decision to make them exclusive anymore

>you don't absolutely have to use the selection argument from onClose. you could totally be updating as selections happen with onSelect and then ignore that final onClose arg and do something else entirely in that callback

here's a vague example to illustrate my question

```
const [ favoriteHats, setFavoriteHats ] = useState<ComboboxOption []>([]);
const doSomethingOnClose = () => console.log("hey I closed, maybe we do something else now")

<Combobox>
  <Combobox.Trigger>
  <Combobox.Content
     onSelect={selection => setFavoriteHats(selection)}
     onClose={_ => doSomethingOnClose()}
  />
</Combobox>
```

right now this wouldn't be possible because I've said you can only provide `onSelect` OR `onClose`

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: I agree with both of you, that while there are things we _could_ do here to squeeze as much optimization as possible out of this, we don't currently have any performance problems to speak of

I'm fine leaving it as is, and if someday we start seeing problems around this, we've got some ideas ready to go

docs/components/Combobox/Combobox.stories.mdx
ZakaryH: interesting idea. the thing that wouldn't offer is the flexibility to do something on select, and then also do something on close. it would be a one or the other with the boolean. which is admittedly the boat we're in right now, but I am reconsidering!

looking at some other implementations, they also have an `onChange` and an `onClose`

image

the difference being that their `onClose` doesn't have the selections as an argument, but you're always free to ignore those.

ZakaryH:
@scotttjob ok I've addressed the feedback, and on the note of `onClose` being a one or the other with `onSelect` I'm just not sure what to do here without a real example. looking at other examples of Combobox, these are not mutually exclusive. you can have both

however, with this I think what I'd like to do is start out with the more restrictive iteration. if people come to us and say "hey we need both" then we can open it up fairly easily. probably.

if we do it the other way around where we allow both, maybe no one uses it, maybe it enables something not great. it'll be harder, or it will feel worse to have that functionality stripped away

so let's keep it as is for now!

scotttjob (approved):
Latest changes look good to me, and address my tiny amount of feedback. Great work Zak!

PR #1570, opened 2023-10-05
feat(components): Combobox Header

taylorvnoj:
## Motivations

For `Combobox` multi-select, we want an option to either `Select all` or `Clear` multiple selections, as well as display a counter for the number of options selected.

Screenshot 2023-10-16 at 10 11 19 AM

### Added

There is now a `ComboboxHeader` when `multiSelect = true`.

The label should:
- read `Select` || `Select {subjectNoun}` when no options are selected
- read `{selectedCount} selected` when one or more options is selected

The action button should:
- read `Select all` when no options are selected
- read `Clear` when one or more options is selected

As soon as a selection has been made, the action button will always display `Clear`. This means that you can `Select all` in a filtered state - only if you've made no selections prior to filtering.  For example:

We've filtered to find both Baggins' and can `Select all`:

Screenshot 2023-10-16 at 10 44 37 AM

However if I've previously selected Pippin Took, and then filter for the Baggins', I will need to manually select Bilbo and Frodo.  

Screenshot 2023-10-16 at 10 45 41 AM

This is intended behaviour as filtering should narrow down options to a very select few.

### Changed

`setInternalSelected` is now being used within `optionsSelectionHandler`.  
We need to make sure this new method of option selection within the `ComboboxHeader` will still work when `onSelect` or `onClose` are passed into `ComboboxContent`.

Because that logic (using either `props.onSelect` or `setInternalState`) was also being used in `handleSelection` we decided to extract that ternary and put it in `useComboboxContent` - named `optionsSelectionHandler`.

## Testing

`Trigger Chip` example is no multi-select with a subjectNoun.
`Multi Select` example of course will also display the `ComboboxHeader`

Make sure the `Select all` and `Clear` functionality works, and ensure the counter is displaying the correct amount of options selected.

---

Random photo of Atlantis


packages/components/src/Combobox/ComboboxProvider.tsx
```diff
@@ -7,6 +7,7 @@ export const ComboboxContext = React.createContext(
     open: boolean;
     setOpen: (open: boolean) => void;
     wrapperRef: RefObject<HTMLDivElement>;
+    subjectNoun?: string;
```
ZakaryH, line 10: do we need this?

I'm not seeing it being passed in nor exposed in the `value`

taylorvnoj: Oops! We do not! Thank you!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -45,9 +46,13 @@ interface ComboboxContentProps {
 }
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
-  const { open, setOpen, wrapperRef } = React.useContext(ComboboxContext);
+  const { open, setOpen, wrapperRef, multiselect } =
+    React.useContext(ComboboxContext);
   const optionsExist = props.options.length > 0;
 
+  // Calculate the count of selected options
+  const selectedCount = props.selected ? 1 : 0;
```
ZakaryH, line 54: I'm assuming this is temporary code

a note for later, we'll need to make this look at the `selectedOptions` wrapper that will either be the external state passed in, or the internal state if we have the `onClose` callback

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.css
```diff
@@ -0,0 +1,5 @@
+.header {
+  display: flex;
+  padding: 16px 16px 0 16px;
```
ZakaryH, line 3: 💅 tokens pls and thank u

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
taylorvnoj: Correct - just wanted something to actually change the label within the Header.  Which also means the Header component itself has temporary code based on just 1 selection.

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.css
taylorvnoj: lol let it be known that I know this - but was quickly putting px's instead of tokens. Still......shame!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -82,6 +87,13 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
         setSearchValue={setSearchValue}
       />
 
+      {multiselect && (
+        <ComboboxHeader
+          subjectNoun={props.subjectNoun}
+          selectedCount={selectedCount}
```
ZakaryH, line 93: nice I like this approach where we figure out the logic of the count out here, and `ComboboxHeader` gets to be very naive and only cares about a number

I'm not sure when exactly we'd reuse it, but if we did it makes this component not coupled to our Combobox implementation

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.test.tsx
ZakaryH, line 1: for these `ComboboxHeader` tests, since we are unit testing `ComboboxHeader` which only has props, and never interacts with the Provider - we can actually completely omit that and just do

```
render(<Comboboxheader selectedCount={someNumber}/>)`
```

and then delete the `MockComboboxProvider` function definition in this file entirely

on that, if for some reason we did end up in a scenario where it was necessary to have the `MockComboboxProvider` wrapping these tests, I'd recommend exporting the existing definition and reusing that everywhere so we don't have multiple definitions of it that we need to keep in sync as changes happen!

chris-at-jobber:
This looks really good so far! Seeing it live I'd like to adjust the spacing on the right of the header to have the action visually align with the selection icon ([updated Figma](https://www.figma.com/file/gQNDrpJYRFAwfbP3jRO0an/DSC-%5B-Combobox-%5D?type=design&node-id=1514%3A142971&mode=design&t=ohXz3qq9Wq3mRsQE-1) for reference). 

Playing with it in the code I think the specific adjustment I'm looking for is reducing the `padding-right` to `space-small`
image

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export function ComboboxHeader(props: ComboboxHeaderProps): JSX.Element {
+  const label = props.subjectNoun ? `Select ${props.subjectNoun}` : "Select";
+  const actionText = props.selectedCount > 0 ? "Clear" : "Select all";
+
+  return (
+    <div className={styles.header} data-testid="ATL-Combobox-Header">
+      <Typography textColor="heading" fontWeight="semiBold">
+        {props.selectedCount === 1 ? `${props.selectedCount} selected` : label}
```
ZakaryH, line 18: assuming this is also temp

once it's real, we can probably make a single var of `const hasSelected = props.selectedCount > 0`

and then use that for the actionText and this bit of logic to show the `N selected` or the label

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.test.tsx
taylorvnoj: Ahh ok that makes sense! 👍

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.tsx
taylorvnoj: Exactly, also temp.  Okay I love that and will simplify this nicely

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -99,6 +100,17 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
         setSearchValue={setSearchValue}
       />
 
+      {multiselect && (
+        <ComboboxHeader
+          subjectNoun={props.subjectNoun}
+          selectedCount={selectedOptions.length}
+          onClearAll={() => {
+            setInternalSelected([]);
```
ZakaryH, line 108: just a heads up, this won't work for the case where we were given `onSelect`

the internal state is only used for `onClose` where the passed in state of `selected` can diverge from the internal state

for `onSelect` we'll need to use the function we were given. you can see an example of how this work in `handleSelection` where we either use the `props.onSelect` or the `setInternalState` depending on if it's an `onClose` or not

ZakaryH:
this is looking very nice!

there's a few cases I think would be valuable to have tests for because for one thing they are somewhat obscure flows, so knowing we have coverage on them will be extra valuable compared to flows we'll often see. secondly, the behavior is a bit debatable - so having the tests is a light form of self documentation that lets people know we are aware of the behavior and it is intended

we kinda have these tested already in the scenario where `filteredOptions` just is `options` but we don't have coverage on the scenarios where they diverge!

**Contextual Select All**

- have options A, B, C, AA
- type A in the search
- hit select all
- the selection handler should only have been called A and AA

**Not Contextual Clear**
- have options A, B, C, AA
- either have something already selected, or select it in the test
- type in A
- hit clear
- all selections should be cleared regardless of the search term

then there's on case I'm on the fence about where we pass it `onClose` instead. in this case, we would expect that the handler is not called when we press Select All, but they should be visually selected

this could be done as a pretty thorough integration test.... or we might be able to simply assert that `aria-selected` is true under those conditions

ZakaryH:
couple things here

1) the width of the listbox changes because the text is changing, is that OK?
2) we can't guarantee that the `subjectNoun` won't run right up against the Select All button, is that OK?
Kapture 2023-10-16 at 12 55 28


pretty much the same thing, but taken to an extreme - with an absurdly long subjectNoun, it wraps and makes that same problem even more pronounced. I'm tempted to say this is fine. if you give us a really long word, that's just what happens
Kapture 2023-10-16 at 13 03 25

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.tsx
```diff
@@ -0,0 +1,48 @@
...
+  readonly subjectNoun?: string;
+  readonly selectedCount: number;
+  readonly onClearAll: () => void;
+  readonly onSelectAll: () => void;
+}
+
+export function ComboboxHeader(props: ComboboxHeaderProps): JSX.Element {
+  const hasSelected = props.selectedCount > 0;
+  let label = "Select";
+  let actionLabel = "Select all";
```
ZakaryH, line 16: we've got a few bits of logic here that can be thematically grouped around labels

thinking ahead, as a courtesy to future us (or other developers) let's try to extract this out to a function so that we have as much room as possible before hitting the ol' `Function ComboboxHeader has too many statements (13)` at which point someone will be forced to refactor things

what do you think about this?

```
  const hasSelected = props.selectedCount > 0;
  const actionLabel = hasSelected ? "Clear" : "Select all";
  const label = getLabel(hasSelected, props.selectedCount, props.subjectNoun);
  const handleSelectAll = hasSelected ? props.onClearAll : props.onSelectAll;
```

where `getLabel` is a function defined outside the body of the `ComboboxHeader` to avoid re-making it

```
function getLabel(
  hasSelected: boolean,
  count: number,
  subjectNoun?: string,
): string {
  if (hasSelected) {
    return `${count} selected`;
  }

  if (subjectNoun) {
    return `Select ${subjectNoun}`;
  }

  return "Select";
}
```

ZakaryH: admittedly this is a bit heavy on ternaries, but I am of the opinion these are pretty easy to understand - so they're an appropriate use that keeps the code lean and doesn't sacrifice readability

taylorvnoj:
> the width of the listbox changes because the text is changing, is that OK?

Although it's a bit choppier in this instance, the same thing happens when entering a long search (width grows).  I defer to @chris-at-jobber on if that's ok or not.

> we can't guarantee that the subjectNoun won't run right up against the Select All button, is that OK?

ahhh...I don't think that's ok. I can probably set something so there's always a 16px space between the end of the `subjectNoun` and the beginning of the header button - just like we have when there is a long search term

Screenshot 2023-10-16 at 4 32 30 PM


> with an absurdly long subjectNoun, it wraps

I wonder if this is the lesser or two evils where the other option is that Combobox can expand an entire page, hypothetically.

ZakaryH (changes requested):
a couple questions around when subjectNoun is a long word and if we wanna adjust anything there. could be as simple as some padding, or maybe it's fine! this one is probably more of a @chris-at-jobber question

a couple tests that I think would be valuable to add

then a small suggestion for getting our lines of code down inside `ComboboxHeader` to avoid future refactors for as long as possible!

chris-at-jobber:
I think these are the answers you're looking for...

I think it's outside the scope of this PR, but I don't love the container width dynamically re-sizing based on the length of its' children (ie after first render). For the header label, that should be static right? But can we constrain the search input to not grow/shrink the container as the user types? That one really bothers me.

If we can keep a `space--smaller` gap between the label and the action, I think that's sufficient.

I prefer the label wrapping to the container re-sizing.

ZakaryH:
>I wonder if this is the lesser or two evils where the other option is that Combobox can expand an entire page, hypothetically.

we've got a few options here I feel. we could:
- truncate since the `subjectNoun` should be legible in the search, one could likely fill in the blanks at what the word is having already seen the full version above
- not show the subjectNoun in this section at all :o and keep it "Select` or maybe `Select Option` or `Select Options` for multiselect

taylorvnoj:
@chris-at-jobber Width of the header label being static: yes and no?  In one of @ZakaryH  gifs, if you have a longer `subjectNoun`, once you 'Select all' the content box shrinks again because the label has shrunk down to say "1 selected". So the container width does re-size.

We can either :
- be ok with this possibly rare use case (not saying it's not important, just wondering how often `subjectNoun` will be 1. actually assigned and 2. a very long word)
or
- wrap the header label beyond 9/10 characters for the `subjectNoun` so that there is never any resizing
or
- expand the width of the content to give us a little more wiggle room


Either way I'm going to make a separate ticket to stop the grow/shrink of the container as the user types in the search input.

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.tsx
taylorvnoj: Great call out, thank you Zak! I fixed this here 82d4f425d907227f5b358cd59976fb8afc98563c
I was previously heavy on ternaries in this file so went the opposite way.  Because these ternaries are 1 line I'm totally fine with that for the sake of keeping things lean

taylorvnoj:
@chris-at-jobber `space-smaller` gap added here: 8f9eff3da919358c7d92d5a51db595cb23cf3b41

@ZakaryH let me know what you think about the tests I added here: 3a35cd2fcd3b2930f1d733ae012916e46784d151 - for the `onClear` I would agree I'm on the fence too.  Although I understand that we want it to be clear this Header functionality will be the same regardless of which prop is used, we are documenting that idea in other tests.

taylorvnoj:
@ZakaryH 18e36292590238ba04b4f24a46c0c02dc9bcbef4 onClose is not called at Select All, but are visually selected

ZakaryH:
got one more for ya! given how `select all` works, where it operates on the things that match your filter - if you have no results this does nothing

image

vs the clear in that same scenario, it will do something

image

probably not the worst thing ever but I'm not a huge fan of rendering things that we know will do nothing. I think this comes back to the fact that clear is not contextual whereas select is

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -369,10 +375,75 @@ describe("Combobox multiselect", () => {
       expect(queryByText("Choice: Shelob the Spoder")).not.toBeInTheDocument();
     });
   });
+
+  describe("Select all", () => {
```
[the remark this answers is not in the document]
ZakaryH: my final thoughts are that if this was the only thing to change that prevented an approval, it's fine. I can live with them being in here

however if we are making some changes for the `select all` button when no results, and we'd likely make some tests for that too - then I'd for sure be in favor of moving these!

ZakaryH:
similar, when you have no options at all - we probably shouldn't have the select all button. it just doesn't make sense here imo

image

taylorvnoj:
@chris-at-jobber can we get your opinion on[ this comment](https://github.com/GetJobber/atlantis/pull/1570#issuecomment-1769263787) - when filtering with no results should we still see Select all?

chris-at-jobber:
@taylorvnoj makes sense to hide it if we can!

taylorvnoj:
@ZakaryH  looking for your feedback on [this name](https://github.com/GetJobber/atlantis/pull/1570/files#diff-8a4e1042f4421ead53add52a08e0a84331aa7b33e39019aeeaff0e706ecbfc8dR17) as well.  

At first I only had `searchTermMatches` but then realized when I had a selection, then did a search with no filter results, the Clear button also disappeared.   

Otherwise, added as many tests as I could to self document the possible flows, moved those `Select` and `Clear` tests to `ComboboxContent.test` and obviously created a new subfolder for the header.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentHeader/ComboboxContentHeader.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const hasSelected = props.selectedCount > 0;
+  const actionLabel = hasSelected ? "Clear" : "Select all";
+  const label = getLabel(hasSelected, props.selectedCount, props.subjectNoun);
+  const handleSelectAll = hasSelected ? props.onClearAll : props.onSelectAll;
+
+  const searchTermMatches = props.options.some(option =>
+    option.label.toLowerCase().includes(props.searchValue.toLowerCase()),
+  );
+
+  const showClearButton = hasSelected || searchTermMatches;
```
ZakaryH, line 17: I don't mind the name but what I do think we can do differently here is leverage work we've already done!

in `ComboboxContent` we already have a `filteredOptions` that is doing pretty much this exact operation to figure out if we have any matches

what we could do instead is 

```
 <ComboboxContentHeader
          subjectNoun={props.subjectNoun}
          selectedCount={selectedOptions.length}
          onClearAll={() => {
            optionsSelectionHandler([]);
          }}
          onSelectAll={() => {
            optionsSelectionHandler(filteredOptions);
          }}
          hasResults={filteredOptions.length > 0}
          searchValue={searchValue}
          options={props.options}
        />
```

and then add the prop `readonly hasResults: boolean;` to our types of course

this way we can cut down on iterations, DRY up the code and yeah just take advantage of work we've already done within the component!

and then I suppose to better match the `hasSelected` a name like `hasResults` or `hasSearchResults` might fit just a tiny bit better

ZakaryH: another little benefit from this is that we no longer have to pass in `searchValue` as a prop to this

although I think the odds of reusing this might be low, it's nice to be able to de-couple those concepts. the less we need a component to know about the others, or our specific implementation - the better!

if we thought this was going to be generalized, there's even more we could do - but for now, I think letting it be just a tad bit more naive is a nice improvement

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.tsx
```diff
@@ -16,17 +19,25 @@ export function ComboboxHeader(props: ComboboxHeaderProps): JSX.Element {
   const label = getLabel(hasSelected, props.selectedCount, props.subjectNoun);
   const handleSelectAll = hasSelected ? props.onClearAll : props.onSelectAll;
 
+  const searchTermMatches = props.options.some(option =>
+    option.label.toLowerCase().includes(props.searchValue.toLowerCase()),
+  );
+
+  const showClearButton = hasSelected || searchTermMatches;
```
ZakaryH: a lil nitpick here 💅 

this isn't always going to be a `ClearButton`, so a more flexible name could help communicate that

perhaps..... `showAction`? `showButton`? something like that, that's more agnostic of the function/behavior since that is dynamic

docs/components/Combobox/Web.stories.tsx
```diff
@@ -64,8 +65,9 @@ const ComboboxButton: ComponentStory<typeof Combobox> = args => {
 
 const ComboboxChip: ComponentStory<typeof Combobox> = args => {
   const [selected, setSelected] = useState<ComboboxOption[]>([]);
+
   return (
-    <Combobox {...args}>
+    <Combobox {...args} multiSelect>
```
ZakaryH, line 70: oh I just noticed this. was this intentional to make the chip example a multiselect?

packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -88,55 +88,97 @@ export interface ComboboxSearchProps {
...
+
+  /**
+   * The function to call when the clear all button is clicked.
+   */
+  readonly onClearAll: () => void;
+
+  /**
+   * The function to call when the select all button is clicked.
+   */
+  readonly onSelectAll: () => void;
```
ZakaryH, line 127: in the vein of keeping components simple, arguably the job of the header should be
- conditionally show some text
- conditionally show a button
- do a thing when the button is pressed

an optimization we _could_ make here is to simplify the props and move the logic out of this, and into our Combobox. so out there we'd figure out if the action should be `ClearAll` or `SelectAll` and then the `Header` takes in a single action as a prop. same thing for the label logic. then, the `Header` just gets a single `Heading` prop or something like that

since I don't reasonably expect this to be re-used, I don't mind this setup and am 100% not asking for these changes, just something to consider!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.test.tsx
ZakaryH, line 1: this is fantastic test coverage that will give us so much confidence when we have to make changes in the future!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentHeader/ComboboxContentHeader.test.tsx
```diff
@@ -0,0 +1,214 @@
...
+          ]}
+        />,
+      );
+
+      fireEvent.click(getByText("Clear"));
+
+      expect(onClearAll).toHaveBeenCalledTimes(1);
+    });
+
+    describe("When the search term matches one or more options", () => {
```
ZakaryH, line 174: we have this same `describe` on line `74` in this file, let's move one of these into the other and delete whichever one is now empty

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentHeader/ComboboxContentHeader.test.tsx
```diff
@@ -0,0 +1,214 @@
...
+              { id: "3", label: "Samwise Gamgee" },
+            ]}
+          />,
+        );
+
+        expect(getByText("Clear")).toBeInTheDocument();
+      });
+    });
+
+    describe("When the search term does not match any options", () => {
```
ZakaryH, line 194: same thing here, on line `94` we have `When the search term matches one or more options`

let's group these up!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentHeader/ComboboxContentHeader.test.tsx
```diff
@@ -0,0 +1,214 @@
...
+            { id: "2", label: "Frodo Baggins" },
+            { id: "3", label: "Samwise Gamgee" },
+            { id: "4", label: "Meriadoc Brandybuck" },
+          ]}
+        />,
+      );
+
+      fireEvent.click(getByText("Clear"));
+
+      expect(onClearAll).toHaveBeenCalledTimes(1);
```
ZakaryH, line 171: so I see we have the `onClearAll` and `onSelectAll` defined globally-ish within the scope of this larger `describe` I'm guessing that's to avoid re-defining it each time, but one downside of that is that since it's in that outer scope it will "remember" how many times it has been called, and if we had any other tests in here that made assertions around calling it, those could fail

so imagine after this test we have the inverse assertion of 

`expect(onClearAll).not.toHaveBeenCalled()`

this would fail because the same instance of `onClearAll` was called in a previous test

I'd recommend being a bit more repetitive in this instance and making a fresh, locally scoped version of `onClearAll` and `onSelectAll` for each test to do its assertions against so that it doesn't trip someone up in the future

alternatively, if we want to avoid writing the same thing out over and over, we can always make a helper function that manages the setup, gives us a fresh set of actions and then we just pass in whatever is unique to our test's setup as args. in this case that might be a bit much, but up to you!

docs/components/Combobox/Web.stories.tsx
taylorvnoj: Not really - I can't remember why I added it but I was testing something in an earlier iteration. I'll remove this!

packages/components/src/Combobox/components/ComboboxContent/ComboboxHeader.tsx
taylorvnoj: `showAction` - thank you! Did that here: fda87b77dc2f1061717036b0909a17b0bf0c9c21

docs/components/Combobox/Web.stories.tsx
taylorvnoj: Done here fda87b77dc2f1061717036b0909a17b0bf0c9c21

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentHeader/ComboboxContentHeader.test.tsx
taylorvnoj: re-organized the tests here fda87b77dc2f1061717036b0909a17b0bf0c9c21

ZakaryH (approved):
LGTM pending passing specs

PR #1571, opened 2023-10-06
feat(components): Chip

scotttjob:
Note: As this is my first PR on Atlantis, I expect there to be feedback. Please be honest! :) If you see something that should be better, leave a comment.

## Motivations

As part of DataList/ClientList we needed a new Chip style. We revisited our old Chip style and decided to give everything a bit of a tidy-up. After we talked a few times as a group, we all landed on the following design.

## Changes

1. Added a 500 weight to our font weights to capture the 'Medium' font weight found in the Figma file.
1. Added a new Hook to pull one specific Component out of a list of React Children.

### Added

1. Added a new component: Chip. This is a standalone component that does not need a wrapping parent to function.
2. Added two compound components, Chip.Prefix and Chip.Suffix. These are 'intelligent' in that they only accept a handful of components as valid, and only one child is accepted.
3. Added two ease-of-use components, ChipSelectable and ChipDismissible

### Screenshots
image
image

### Changed

- Existing Chip/Chips are unchanged. In an upcoming sprint, we may upgrade 'Chips' to use `Chip` under the hood, but that has not been finalized.

### Deprecated

- Nothing as of yet.

### Removed

- None

### Fixed

- None

### Security

- Should have no security issues

## Testing

Ran manual tests to ensure all the functionality is working, along with wrote a variety of unit tests to ensure everything works together.

---

Random photo of Atlantis


packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+  background-color: var(--color-greyBlue--lighter);
+}
+
+.invalid,
+.invalid:hover,
+.invalid:focus,
+.subtle.invalid,
+.subtle.invalid:hover,
+.subtle.invalid:focus {
+  border: var(--border-base) solid var(--color-red);
```
chris-at-jobber, line 76:
```suggestion
  border: var(--border-base) solid var(--color-critical);
```

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+  background-color: var(--color-red--lightest);
```
chris-at-jobber, line 77:
```suggestion
  background-color: var(--color-critical--surface);
```

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+.chip .chipBar {
+  width: var(--space-minuscule);
+  height: var(--base-unit);
+  margin: 0 var(--space-small);
+  background-color: var(--color-greyBlue--light);
+}
+
+.chip .selected {
+  border-radius: var(--radius-circle);
+  background-color: var(--color-white);
```
chris-at-jobber, line 58:
```suggestion
  background-color: var(--color-surface);
```

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+}
+
+.base {
+  color: var(--color-blue);
```
chris-at-jobber, line 62:
```suggestion
  color: var(--color-heading);
```

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+.subtle.invalid,
+.subtle.invalid:hover,
+.subtle.invalid:focus {
+  border: var(--border-base) solid var(--color-red);
+  background-color: var(--color-red--lightest);
+}
+
+.subtle {
+  border: var(--border-base) solid var(--color-grey--lighter);
+  background-color: var(--color-white);
```
chris-at-jobber, line 82:
```suggestion
  background-color: var(--color-surface);
```

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+}
+
+.disabled,
+.disabled:hover,
+.disabled:focus {
+  color: var(--color-grey);
+  background-color: var(--color-grey--lighter);
```
chris-at-jobber, line 89:
```suggestion
  background-color: var(--color-disabled--secondary);
```

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+
+.subtle {
+  border: var(--border-base) solid var(--color-grey--lighter);
+  background-color: var(--color-white);
+}
+
+.disabled,
+.disabled:hover,
+.disabled:focus {
+  color: var(--color-grey);
```
chris-at-jobber, line 88:
```suggestion
  color: var(--color-disabled);
```

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+.chip {
+  display: flex;
+  position: relative;
+  height: var(--chip-height);
+  padding: var(--space-small) var(--base-unit);
+  border: none;
+  border-radius: var(--chip-radius);
+  text-align: left;
+  cursor: pointer;
+  transition: all ease-in-out 0.12s;
```
chris-at-jobber, line 15: We do have some standard [timing](https://atlantis.getjobber.com/?path=/docs/design-animation--page#timing) and [easing](https://atlantis.getjobber.com/?path=/docs/design-animation--page#easing) values available

packages/components/src/Chip/Chip.types.d.ts
```diff
@@ -0,0 +1,34 @@
+import React, { PropsWithChildren } from "react";
+
+export interface ChipProps extends PropsWithChildren {
+  /** Accessible Label */
+  ariaLabel?: string;
+  /** Disables Chip */
+  disabled?: boolean;
+  /** Adds a Medium Heading with a Pipe separator */
+  heading?: string;
+  /** Makes Chip red. */
```
chris-at-jobber, line 10: I would be a smidge more semantic in these descriptions so if (when) the styling related to these properties change we don't need to do too much adjusting here. For example, "makes Chip red" could maybe be "Changes Chip styling to inform the user of an error"? This maybe isn't the _best_ example because red is so "error-y" but maybe one day we decide to add "help text" to an invalid Chip that would still leverage this prop... does that make sense?

chris-at-jobber: Heading is another example, we probably don't need to get specific to the font weight, so much as "adds more prominent text to act as a heading"

chris-at-jobber (changes requested):
This looks really great! The comparisons story is amazing for ease-of-review👌 

A few comments on the prop descriptions, not blocking but worth thinking about.

Suggested a few changes in the CSS to use our [semantic colors ](https://job-79340-job-79448-separati.atlantis.pages.dev/?path=/docs/design-colors--page) where possible

Interaction-wise, 

And finally, some visual design stuff:

I am noticing that when the Chip swaps between bordered and un-bordered states (changes variations between subtle and base, or base becomes invalid, for example) the Chip itself re-sizes:
https://github.com/GetJobber/atlantis/assets/39704901/0b0a1a17-b66b-4bf9-a17d-aaa2ba7673ff

Also, a Figma whiff on my part, but the subtle chip should in fact have a hover state! I've updated [the Figma](https://www.figma.com/file/rIIhulZvcp9M82lNOCGv16/branch/S6J5803Pjmy4VvpsQx0o8S/Product%2FOnline?type=design&node-id=18944%3A3858&mode=design&t=6pSlqrkEEa0SSiKX-1). 

The "pipe" separator when `heading` is present should change to `--color-disabled` when Chip is disabled:
image

When disabled, the dismiss, icons, or checkmark indicator in the prefix/suffix should be `--color-disabled`.
image

Text color is black in most of the variations I'm seeing here, where it should use `--color-heading`. It _seems_ to be tied to `subtle` and/or `heading` props?
image

These variants are all invisible! Should we rename this to `extraSubtle` 👻  _for legal reasons, this is a joke 😄_
image

Other than that, one thing I was unclear on was if the dismissible X has an `onClick` - I believe the current Chip does and we'll need to carry that behaviour over.

That's all! Again, this is a fantastic first component and I'm loving how much progress we've made here.

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+  background-color: var(--color-white);
+}
+
+.base {
+  color: var(--color-blue);
+  background-color: var(--color-greyBlue--lightest);
+}
+.base:hover,
+.base:focus {
+  background-color: var(--color-greyBlue--lighter);
```
chris-at-jobber, line 67: One other thought - I think we should set these re-used color variables at the root as I have some semantic color updates in the design pipeline that will allow us to updates these soon.

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+.disabled:hover,
+.disabled:focus {
+  color: var(--color-grey);
+  background-color: var(--color-grey--lighter);
+  cursor: auto;
+}
+
+.invalid.disabled {
+  pointer-events: none;
+  opacity: 0;
```
chris-at-jobber, line 95: I think this is where the invisible variants are coming from!

chris-at-jobber:
_One last_ thing - I was very focused on reviewing Chip for this PR, but I notice that the Chip in Chips are not impacted by this work. Do you see that as a followup PR?

packages/components/src/Chip/Chip.css
scotttjob: I have fixed this (I knew I was gonna get something wrong with the colours, I searched for rgb values in the code and found the root colours, but not the semantic colours). :)

packages/components/src/Chip/Chip.css
scotttjob: Also fixed! :)

packages/components/src/Chip/Chip.css
scotttjob: Also fixed! :)

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,96 @@
...
+  margin-right: calc(-1 * var(--space-small));
+}
+
+.chip .prefix,
+.chip .suffix {
+  display: flex;
+  width: var(--space-large);
+  height: var(--space-large);
+  border-radius: var(--radius-circle);
+  background-color: var(--color-white);
```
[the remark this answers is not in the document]
scotttjob: Also fixed! :)

packages/components/src/Chip/Chip.css
scotttjob: Also fixed! :)

packages/components/src/Chip/Chip.css
scotttjob: Also fixed! :)

packages/components/src/Chip/Chip.css
scotttjob: Also fixed! :)

packages/components/src/Chip/Chip.css
scotttjob: Also fixed! :)

packages/components/src/Chip/Chip.css
scotttjob: Those 20 milliseconds really made the difference though. I have fixed this. ;)

packages/components/src/Chip/Chip.types.d.ts
scotttjob: Totally makes sense in both cases. I did this pretty late last night and I think I was trying to be a bit clever in both cases but it came off lazy instead. 

Also in the habit of having anything written that will be seen by others be double checked by a copywriter, but I need to break myself of that habit. :)

packages/components/src/Chip/Chip.css
scotttjob: Done. I made a new colour at the root of the component to track this.

packages/components/src/Chip/Chip.css
scotttjob: I wasn't 100% what to do here. In your chart, when things are both invalid and disabled at the same time, they don't show up but in your figma they still take up space. I thought that might be weird to just mark them with display:none, but I can definitely do that if you would prefer.

image

scotttjob:
> _One last_ thing - I was very focused on reviewing Chip for this PR, but I notice that the Chip in Chips are not impacted by this work. Do you see that as a followup PR?

100%. I'll spend some time early next week and flip that over.

docs/components/Chip/Comparison.stories.tsx
```diff
@@ -0,0 +1,419 @@
+import React from "react";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { Chip } from "@jobber/components/Chip";
+import { Content } from "@jobber/components/Content";
+import { Typography } from "@jobber/components/Typography";
+import { Avatar } from "@jobber/components/Avatar";
+import { Icon } from "@jobber/components/Icon";
+
+export default {
+  title: "Components/Selections/Chip/Comparisons",
```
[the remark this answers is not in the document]
scotttjob: Done! I'm happy to put things wherever they need to go!

docs/components/Chip/Web.stories.tsx
```diff
@@ -0,0 +1,186 @@
+import React from "react";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { Chip } from "@jobber/components/Chip";
+import { Content } from "@jobber/components/Content";
+import { Icon } from "@jobber/components/Icon";
+
+export default {
+  title: "Components/Selections/Chip/Web",
+  component: Chip,
+  subcomponents: { Chip },
```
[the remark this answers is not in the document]
scotttjob: It's gonezo!

docs/components/Chip/Web.stories.tsx
```diff
@@ -0,0 +1,186 @@
...
+Base.args = {
+  label: "Chip Label",
+  heading: "",
+  ariaLabel: "Accessible Label",
+  role: "button",
+  tabIndex: 0,
+  variation: "base",
+  disabled: false,
+  invalid: false,
+};
```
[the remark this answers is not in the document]
scotttjob: We could do, and I will in an update early next week (I need to move Chips out) I haven't done it yet, just because I'm tight on time today. Great feedback, I've tracked it. :)

packages/components/src/Chip/Chip.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const { getByText } = render(
+      <Chip onKeyDown={onKeyDown} label="Test Chip" />,
+    );
+    fireEvent.keyDown(getByText("Test Chip"), { key: "Enter" });
+    expect(onKeyDown).toHaveBeenCalledTimes(1);
+  });
+
+  it("renders an invalid Chip", () => {
+    const { container } = render(<Chip label="Test Chip" invalid />);
+    expect(container).toMatchSnapshot();
```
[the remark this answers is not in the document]
scotttjob: I was definitely just following another nearby component for what types of tests were written. I can drop the snapshot tests early next week.

packages/hooks/src/useChildComponent/useChildComponent.ts
```diff
@@ -0,0 +1,22 @@
+import React, { ReactNode, useMemo } from "react";
+
+export function useChildComponent(
```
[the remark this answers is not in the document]
scotttjob: Definitely similar. The idea is to pull out the first valid component that matches the matcher, and ignore anything else.

packages/hooks/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from "./useAssert";
+export * from "./useChildComponent";
```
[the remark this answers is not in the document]
scotttjob: I would definitely be down to document it if we feel it's valuable to the larger community (and ourselves).

packages/components/src/Chip/Chip.css
chris-at-jobber: Ah! What that's indicating is that a chip that's both `invalid` and `disabled` shouldn't exist, as it offers the user no recourse to fix whatever issue has led to the invalid state. I do wonder if we need to _technically_ exclude those variations, or just leave it (maybe it just gets the disabled styling?) and if a team designs their way into that state, we can have guidance on why they shouldn't?

scotttjob: We could mark them as `display:none` which does effectively remove them from the DOM screen readers, all of that, but it will give a clue to a developer that their thing _is rendering_ but that it's not showing up because they misconfigured it.

Or we could code it do just flat out not render anything in that situation, but then we run into the "why doesn't my component show up."

I think writing in the documentation to not use invalid/disabled at the same time, and by hiding in the DOM should be enough. I'm not huge on introducing code to print warnings for developers (unless it's potentially destructive) because that runs the risk of something going weird in Prod for a hand-hold.

docs/components/Chip/Web.stories.tsx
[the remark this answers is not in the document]
scotttjob: I lied about next week. I went ahead and did this. ;) It was a good change request. Thanks Rodrigo!

packages/components/src/Chip/Chip.test.tsx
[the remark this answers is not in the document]
scotttjob: Early next week was a lie again! Also I meant to include a ;) in the comment above about copying the idea from a nearby test. These snapshot tests and the snapshots are gone! :)

packages/components/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,59 @@
+import React from "react";
+import { useChildComponent } from "@jobber/hooks/useChildComponent";
+import classNames from "classnames";
+import styles from "./Chip.css";
+import { ChipPrefix } from "./Chip.Prefix";
+import { ChipSuffix } from "./Chip.Suffix";
```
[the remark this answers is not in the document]
scotttjob: Another lie! This is done. :)

chris-at-jobber:
Maybe not critical to tackle in this PR as I think a few of the team members wanted to have a "should we do this" discussion, but the Chips' current behaviour for overflowing text is
- truncate text (rather than wrap)
- reveal a tooltip with the full content on hover/focus

(it appears the tooltip functionality broke at some point but we did have it, and even [iterated on it](https://github.com/GetJobber/atlantis/pull/738) 🤔)

image

Not something to change in this PR as 

- this new Chip is not currently in use
- whether or not it truncates is a styling decision that won't impact the component API

but we should make sure that we've consciously decided to keep or drop this behaviour

packages/components/src/Chip/Chip.types.d.ts
```diff
@@ -0,0 +1,34 @@
...
+  disabled?: boolean;
+  /** Adds more prominent text to act as a heading */
+  heading?: string;
+  /** Changes Chip styling to inform the user of an issue. */
+  invalid?: boolean;
+  /** What do you want Chip to say? */
+  label: string;
+  /** What role does Chip play? */
+  role?: string;
+  /** Defaults to zero. Accessible feature. */
```
rodrigoeidelvein, line 16: I think we could make these descriptions more descriptive. I don't think we have any components using this structure of asking questions to the user. Also, when there's a default value, we can also expose here to the consumer

Example: https://github.com/GetJobber/atlantis/blob/e0183893629b508a397c8a41cb0aed08121db314/packages/components/src/Avatar/Avatar.tsx

packages/components/src/Chip/components/ChipSuffix/Chip.Suffix.tsx
```diff
@@ -0,0 +1,27 @@
...
+import styles from "../../Chip.css";
+import { Icon } from "../../../Icon";
+
+interface ChipSuffixProps extends PropsWithChildren {
+  className?: string;
+}
+
+export function ChipSuffix({ children, className }: ChipSuffixProps) {
+  let singleChild = useChildComponent(children, d => d.type === Icon);
+  if (!["cross", "add", "checkmark"].includes(singleChild?.props?.name)) {
```
rodrigoeidelvein, line 13: Are these allowed icons? Can we extract this arrow into a more readable constant so we know what it is about?

rodrigoeidelvein (changes requested):
Just a few more, we're almost there! 🤞 

Also, I'd to see a review of someone who actively participated in the `Chip` discussions, maybe @dlrussel ?

packages/components/src/Chip/Chip.css
```diff
@@ -55,16 +58,16 @@
...
 .base {
-  color: var(--color-blue);
-  background-color: var(--color-greyBlue--lightest);
+  color: var(--color-heading);
+  background-color: var(--chip-highlighted);
 }
 .base:hover,
 .base:focus {
-  background-color: var(--color-greyBlue--lighter);
+  background-color: var(--chip-highlighted);
```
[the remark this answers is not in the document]
scotttjob: Sorry about this one, I pushed a change before I realized I had messed up the colours. Should be alright now. What happened was I went to move color-greyBlue--light into a semantic variable at the top of the file, and failed to notice there was light, lighter and lightest (so they all got mistakenly updated to the same colour).

scotttjob:
> Maybe not critical to tackle in this PR as I think a few of the team members wanted to have a "should Chip do this" discussion, but the Chips' current behaviour for overflowing/wrapping is
> 
>     * hug contents up to the maximum available width
> 
>     * Once maximum available width is reached:
>       
>       * truncate text (rather than wrap)
>       * reveal a tooltip with the full content on hover/focus
> 
> 
> The use case for this is that since Chips are likely to contain user-generated content, we should have a failsafe of sorts to preserve the layout of their interface but still give them access to the full content if that scenario arises.
> 
> image
> 
> Not something to change in this PR as
> 
>     * this new Chip is not yet currently in use
> 
>     * whether or not it truncates is a styling decision that won't impact the component API
> 
> 
> but we should make sure that we've consciously decided to keep or drop this behaviour.
> 
> (The existing behaviour appears to be buggy when there's multiple truncated Chips)
> Screen.Recording.2023-10-06.at.4.50.35.PM.mov

This is definitely worth a conversation about how we want to handle that. I figured that text-length would come up and how would we handle it because you're right, currently in the new Chip it's just letting you have long text.

My follow-up question though, is how often do we do this pattern. By pattern I mean capping the length of text and inserting a ToolTip to convey the shortened text? I ask because this is the sort of behaviour I would rather capture in another component and just wrap a portion of Chip in it, or at the very least a hook so we don't pollute Chip.

Also while having this enabled by default would make sense for most users, I definitely think we should empower users to disable the ToolTip/shortening functionality, and arguably any max-width as well.

packages/components/src/Chip/Chip.types.d.ts
scotttjob: Done. I've tried my best to be as descriptive as possible without any cheekyness, and pulled the variations out into an exported type.

packages/components/src/Chip/Chip.types.d.ts
```diff
@@ -0,0 +1,35 @@
+import React, { PropsWithChildren } from "react";
+
+export interface ChipProps extends PropsWithChildren {
+  /** Accessible label, which can be different from the primary label. */
+  readonly ariaLabel?: string;
```
taylorvnoj, line 5: Sorry for the last minute comment! Not a deal breaker but would you mind using multiline comments for the types? I personally find it helps with readability and follows our props pattern in other components 😃 

```suggestion
  /**
   * Accessible label, which can be different from the primary label.
   */
  readonly ariaLabel?: string;

```

packages/hooks/src/useChildComponent/useChildComponent.ts
```diff
@@ -0,0 +1,22 @@
...
+  children: ReactNode | undefined,
+  isCorrectComponent: (toCheck: JSX.Element) => boolean,
```
darryltec, line 5: what else would you use this for other than matching with the component?

packages/hooks/src/useChildComponent/useChildComponent.ts
darryltec, line 1: hmmm maybe `useGetComponent`? since it is just looking for 1 and returns it?

packages/components/src/Chip/Chip.types.d.ts
```diff
@@ -0,0 +1,35 @@
...
+  /** Adds more prominent text to act as a heading. Will be displayed on the left with a | separator. */
+  readonly heading?: string;
+  /** Changes Chip styling to inform the user of an issue. */
+  readonly invalid?: boolean;
+  /** The content of the chip. Will be displayed on the right if you include a heading. */
+  readonly label: string;
+  /** The accessible role the Chip is fulfilling. Defaults to 'button' */
+  readonly role?: string;
+  /** Used for accessibility purpopses, specifically using the tab key as navigation. Defaults to 0. */
+  readonly tabIndex?: number;
```
rodrigoeidelvein, line 17: For these defaults, you can use `@default 0` in a new line. You can search for "@default" in our codebase and you will find some examples. 

Here's one of them https://github.com/GetJobber/atlantis/blob/b1a3f289047063a7df19be0681c525145eb89d75/packages/components/src/InputGroup/InputGroup.tsx

darryltec (commented):
Just a few thoughts here and there. Code is looking pretty good tho

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,120 @@
+:root {
+  --chip-height: 40px;
+  --chip-radius: 20px;
+  --chip-disabled--bg: var(--color-grey--lighter);
+  --chip-bar-color: var(--color-greyBlue--light);
+  --chip-base-bg: var(--color-greyBlue--lightest);
+  --chip-base--hover: var(--color-greyBlue--lighter);
+}
```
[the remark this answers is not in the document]
scotttjob: This was a request from Chris to give the values semantic meaning. So we can talk about it as a group, but I'll leave these for now.

packages/components/src/Chip/Chip.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  it("should have a role of button when role not provided", () => {
+    const { getByRole } = render(<Chip label="Test Chip" />);
+    expect(getByRole("button")).toBeInstanceOf(HTMLButtonElement);
+  });
+
+  it("should have a tabIndex of zero if one is not provided", () => {
+    const { container } = render(<Chip label="Test Chip" />);
+    const element = container.querySelector("button");
```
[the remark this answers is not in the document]
scotttjob: Yup, force of habit. I've replaced my style with the Atlantis style of pulling stuff out of the DOM.

packages/components/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,59 @@
...
+  return (
+    <button
+      className={classes}
+      onClick={onClick}
+      tabIndex={tabIndex}
+      onKeyDown={onKeyDown}
+      aria-label={ariaLabel || label}
+      disabled={disabled}
+      role={role}
+    >
```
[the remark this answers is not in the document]
scotttjob: Done.

packages/components/src/Chip/Chip.types.d.ts
```diff
@@ -0,0 +1,35 @@
+import React, { PropsWithChildren } from "react";
+
+export interface ChipProps extends PropsWithChildren {
+  /** Accessible label, which can be different from the primary label. */
```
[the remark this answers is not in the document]
scotttjob: Done. This is how we wrote prop comments at my previous gig so they didn't take up so much screen real-estate, but it'll have to be an adjustment for me to match Atlantis style.

packages/hooks/src/useChildComponent/useChildComponent.ts
```diff
@@ -0,0 +1,22 @@
...
+): JSX.Element | undefined {
+  return useMemo(() => {
+    let foundChildComponent;
+    React.Children.forEach(children, d => {
```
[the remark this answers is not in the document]
scotttjob: Done.

packages/hooks/src/useChildComponent/useChildComponent.ts
```diff
@@ -0,0 +1,22 @@
...
+      if (
+        typeof d !== "undefined" &&
+        typeof d !== "string" &&
+        typeof d !== "number" &&
+        typeof d !== "boolean" &&
```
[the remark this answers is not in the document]
scotttjob: Done.

packages/components/src/Chip/Chip.css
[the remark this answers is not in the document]
scotttjob: Ah got it! I moved them into the .chip.

scotttjob:
> I think this is the last comment from me! The suffix is still looking squishable when the Chip gets condensed. _(I wasn't able to get the checkmark version of the suffix running in the code tab, but I assume it has the same issue)_
> 
> image

I've corrected this. I didn't have a min-width/min-height on the element, easy enough fix!

chris-at-jobber (commented):
Design approved! Thanks again for all your work here @scotttjob 🙏

packages/components/src/Chip/components/ChipDismissible/ChipDismissible.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { Chip } from "../../Chip";
+import { ChipProps } from "../../Chip.types";
+import { Icon } from "../../../Icon";
+
+export function ChipDismissible(props: ChipProps) {
```
darryltec, line 6: Do we need to narrow down that props? I wonder what happens when you add a children on the usage 🤔 . Would it override the line `9`?

darryltec (commented):
There are 7 more comments that got buried here haha

image

packages/components/src/Chip/Chip.css
```diff
@@ -0,0 +1,119 @@
...
+  margin-right: calc(-1 * var(--space-small));
+}
+
+.chip .prefix,
+.chip .suffix {
+  display: flex;
+  width: var(--space-large);
+  height: var(--space-large);
+  min-width: var(--space-large);
+  min-height: var(--space-large);
```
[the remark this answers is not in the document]
scotttjob: Done. Either solution works so I'm happy to use your suggestion.

packages/components/src/Chip/Chip.tsx
```diff
@@ -0,0 +1,58 @@
...
+      disabled={disabled}
+      role={role}
+      type="button"
+    >
+      {prefix}
+      <Typography size="base" fontWeight="medium">
+        {heading}
+      </Typography>
+      {heading && <span className={styles.chipBar}></span>}
+      <span>
```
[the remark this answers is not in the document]
scotttjob: Nope! A leftover from pre Typography being in there (don't like bare text), it's been removed! :)

packages/components/src/Chip/components/ChipDismissible/ChipDismissible.tsx
scotttjob: Screenshot 2023-10-11 at 7 50 47 PM

I was curious myself, turns out it's fine to pass in as-is. My image React treats `children` a bit differently than a normal prop.

scotttjob: Oh also, this prompted me to bring a Dismissible and Selectable chip into the examples to test this. So this comment made the PR even better. :)

scotttjob:
@darryltec

I can't seem to comment directly on two of your comments. 

For the invalid + disabled state, I've just removed the bit of code that was hiding that exact state (to make them invisible). What happens now is if you put a Chip both into the invalid and disabled states at the same time, it ignores invalid and just goes with disabled. I'm personally okay with this natural, 0-effort behaviour because that's how I treated them while coding (probably why it naturally fell back to that), and also `disabled` feels more important than `invalid`. The other option is a `display:none` in that state to properly remove, or adding some guards in the component for that combined state and throwing a warning, but I think this natural falling to `disabled` does make logical sense (at least for now or until we get a complaint).

In the other comment about `value` before `ev` in the onClick, I went ahead and did that. I had to make a few changes for that, but all seems to be working now.

packages/components/src/Chip/components/ChipSelectable/ChipSelectable.tsx
```diff
@@ -0,0 +1,64 @@
...
+  value,
+  variation,
+  onClick,
+  onKeyDown,
+}: ChipSelectableProps) {
+  const [isChipSelected, setIsChipSelected] = useState(selected);
+
+  useEffect(() => {
+    setIsChipSelected(selected);
+  }, [selected]);
```
darryltec, line 25: now that the nitpicks are out of the way, should this be handled internally?

I think this is better handled through the consumer side of things. Keep the components reactive to props rather than it having a state and a side effect for the prop change.

Right now it just works which gives you a false sense of it actually saving the "selected" state only to find that it actually wasn't since the chip is doing something on its own rather than being controlled by the consumer's state.

packages/components/src/Chip/components/ChipSuffix/Chip.Suffix.tsx
```diff
@@ -0,0 +1,29 @@
...
+  let singleChild = useChildComponent(children, d => d.type === Icon);
+  if (!allowedSuffixIcons.includes(singleChild?.props?.name)) {
+    singleChild = undefined;
+  }
+  return (
+    <span
+      className={classNames(
+        styles.suffix,
+        className,
+        !singleChild && styles.empty,
```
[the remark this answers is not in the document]
darryltec: @scotttjob thoughts on this? https://github.com/GetJobber/atlantis/pull/1571/files#r1352557020

darryltec (changes requested):
Ok i did find one show stopping thing with the use of state

packages/components/src/Chip/components/ChipPrefix/Chip.Prefix.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { PropsWithChildren } from "react";
+import { useChildComponent } from "@jobber/hooks/useChildComponent";
+import classNames from "classnames";
+import styles from "../../Chip.css";
```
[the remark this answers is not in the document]
scotttjob: This was a deliberate choice to cut down on duplication of styles because Prefix/Suffix share a few, so they're doubled up/cascading in the main file. If I moved the CSS for Prefix in, I would have to move the CSS for Suffix as well, and then I'd have to duplicate the styles which I wanted to avoid.

packages/components/src/Chip/components/ChipSelectable/ChipSelectable.tsx
scotttjob: These are mainly convenience Chips to cut down on having to repeat patterns, so in this case I wanted to make the component as low stakes as possible to start using. So it will operate on its own without a consumer providing state, but it will also react to state changes in the event that you pass in a prop programatically (due to something else happening in the UI).

So in my head it's the best of both worlds. You can just use the component without having to worry about state management, but if you wanted to provide your own state for any reason, the component handles that as well.

packages/components/src/Chip/components/ChipPrefix/Chip.Prefix.tsx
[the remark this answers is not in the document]
darryltec: that makes sense

packages/components/src/Chip/components/ChipSuffix/Chip.Suffix.tsx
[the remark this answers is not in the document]
scotttjob: This was a conscious choice in this case to render the wrapper because it's a fairly low-stakes component, and the empty element in the DOM gives a consumer a hint that they're using it incorrectly (flat out not showing up in the DOM could be something else entirely).

packages/components/src/Chip/components/ChipSelectable/ChipSelectable.tsx
darryltec: > You can just use the component without having to worry about state

packages/components/src/Chip/components/ChipPrefix/Chip.Prefix.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+describe("Chip.Prefix Component", () => {
+  it("renders Prefix", () => {
+    const { container } = render(
+      <Chip.Prefix>
+        <Avatar initials="" />
+      </Chip.Prefix>,
+    );
+
+    expect(container.querySelector("." + styles.prefix)).toBeInTheDocument();
```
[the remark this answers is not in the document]
scotttjob: I've adjusted this.

packages/components/src/Chip/components/ChipSelectable/ChipSelectable.tsx
darryltec: >  You can just use the component without having to worry about state

In theory, that makes sense. Until you get to a point where you would have to save that state of the chip somewhere so when you come back to it, it'll still have to correct state since it will always start off as false until you set that prop to true. Which then keeps on toggling since you have a state that updates it now.

In the end, there's always a state somewhere being saved that controls this Chip's appearance. So Chip having its own state for selected is redundant and possibly invites misconfiguration that leads to bugs.

It's the same thing that react does for the raw HTML where the moment you put in a `value` prop but don't properly implement the `onChange` to save that input's value in the state, it'll stop working until you've actually implemented it correctly. In react's POV, it's much easier to spot if that component just _reacts_ to what you've specifically told it to do than it doing something on its own until you realize you've made a mistake of not saving its state.

packages/components/src/Chip/components/ChipSelectable/ChipSelectable.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useEffect, useState } from "react";
+import { Chip } from "../../Chip";
+import { ChipSelectableProps } from "../../Chip.types";
+import styles from "../../Chip.css";
+import { Icon } from "../../../Icon";
```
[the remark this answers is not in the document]
scotttjob: This has been fixed. Originally when I started coding this component your update wasn't in place yet so both options generated an error. :)

packages/components/src/Chip/components/ChipSelectable/ChipSelectable.tsx
scotttjob: I disagree with your premises outlined above but in the interest in getting this PR merged, I've made this requested change and we can have a follow-up discussion at some point in the future about this convenience pattern.

darryltec: sounds good. happy to talk about this in detail since there are legit use cases of a component having its own state. Like opening/closing a menu where the consumer doesn't need to worry about controlling that interaction. Compared to a state that needs to be synced between a parent and a child where an internal state might not be needed. It really depends on the use case of that state.

packages/components/src/Chip/components/ChipSuffix/Chip.Suffix.tsx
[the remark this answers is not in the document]
darryltec: hmm. that's fair!

darryltec (approved):
Alright, this LGTM!

rodrigoeidelvein (approved):
LGTM!

PR #1573, opened 2023-10-06
fix(components): Move InputEmail Mobile into Atlantis

scotttjob:
## Motivations

Recently when moving our import locations, we noticed that we missed both apolloFormSubmitPromise and InputEmail. After investigating both pieces, InputEmail was a good candidate for Atlantis, while apolloFormSubmitPromise was better suited to stay in jobber-mobile, but just move to a new location. There is a separate PR to track that work.

## Changes

1. InputEmail moved from jobber-mobile to Atlantis

### Added

- InputEmail

### Changed

- InputEmail was unchanged while copied over.

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- N/A

### Security

- N/A

## Testing

- Testing was minimal as this was a blanket copied component, and was not changed. Impact is minimal since consumers will have to upgrade to the new path first along with the new version of Atlantis.

---

Random photo of Atlantis

### DIFF
--- a/docs/components/InputEmail/Mobile.stories.tsx
+++ b/docs/components/InputEmail/Mobile.stories.tsx
@@ -1,0 +1,28 @@
+import React from "react";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { InputEmail } from "@jobber/components-native";
+
+export default {
+  title: "Components/Forms and Inputs/InputEmail/Mobile",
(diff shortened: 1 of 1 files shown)


rodrigoeidelvein (approved):
I only added the storybook page for InputEmail, I don't think we need somebody else to review it... 

LGTM!

PR #1574, opened 2023-10-06
fix(components): DataList search initial value

darryltec:
## Motivations

In the event that saving a filter and applying it back is required, we need to be able to update the initial value of the search box to reflect that the data you're seeing is indeed programmatically filtered instead of user filtered

### Added

- Initial value for search
  - Updating this value after mount would result in the search input being rerendered.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


rodrigoeidelvein (approved):
Good catch! LGTM!

PR #1575, opened 2023-10-06
chore: enforce a better relative import

darryltec:
## Motivations

We've always had this warning 

image

That was meant to catch people importing between packages via `../` their way in there. That's not good because this is a monorepo and relative paths does not exist after packaged up.

This PR turns that off and replaces it with a package that's made for monorepos and relative packages that throws this error now

image

### Added

- harder linter on reaching to another package through `../`

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,7 +8,8 @@ const packageAliases = [
 ];
 
 module.exports = {
-  extends: ["@jobber/eslint-config"],
+  plugins: ["monorepo-cop"],
+  extends: ["@jobber/eslint-config", "plugin:monorepo-cop/recommended"],
   root: true,
   settings: {
     "import/ignore": ["react-native/index"],
@@ -21,10 +22,10 @@ module.exports = {
   },
   rules: {
     /*
-      Need to figure out a good way to enforce intra vs inter module import
-      rules. For now, warn on these.
+      Atlantis is a monorepo so we need to use `monorepo-cop` to enforce the
+      relative import rule.
      */
-    "import/no-relative-parent-imports": "warn",
+    "import/no-relative-parent-imports": "off",
     "no-restricted-imports": [
       "error",
       {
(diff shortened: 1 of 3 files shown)


rodrigoeidelvein:
How does it affect the first screenshot? Is it going to ask to import from `@jobber/components/DataList` even though it is in the same package?

I know it's kinda ugly to have those `../../../`, but I like that I can cmd + click and it goes directly to the file I want. I miss that when we use `@jobber/components/{component}` in tests for example, but there should be a way to map it too

rodrigoeidelvein:
> How does it affect the first screenshot? Is it going to ask to import from @jobber/components/DataList even though it is in the same package?

Oohh, it deactivates! (sorry I missed this in the PR description). Nice! I also tested it now, just going to make some more tests before approving it, looks nice tho

darryltec:
@rodrigoeidelvein it won't anymore. it won't throw any warning at it and you can still do it within the package as long as it's from the same package folder. Basically you can't go to hooks folder if you're in the component folder and vice versa. But if you ../ to another folder, as long as it's from the same package, you should be okay.

As for test, that's only affecting test, you can still do that on other places. There's a TS setting that we need to figure out first before we can fix that.

rodrigoeidelvein:
For example, in `DataListActions.test.tsx`, there's this error showing up for me
image

I already ran `npm run bootstrap` on root and restarted ESLint and TS, but it keeps showing up. Maybe it would be better to import directly from the folder instead of the package, wouldn't it?

darryltec:
@rodrigoeidelvein yeah, that's different

darryltec:
@rodrigoeidelvein this PR fixes that issue

rodrigoeidelvein (approved):
Good QOL improvement! LGTM

PR #1576, opened 2023-10-06
fix(components): Use button type in Dismissable Chip menu items

TJKruitbosch:
## Motivations

When investigating some flakiness in some internal Jest tests, it was observed that _part_ of the problem was that a page using `Form` with a dismissible `Chips` component was having the Form submit simply by selecting a new item to add to the dismissable chips selection. I'm not sure why we weren't seeing this behaviour in a browser, but the [HTML spec](https://html.spec.whatwg.org/multipage/form-elements.html#attr-button-type) does say that the default type for a `<button>` element is `submit`, and [jsdom matches that behaviour](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLButtonElement-impl.js#L52). It doesn't look like that was intentional though; this component is handling it's own activation logic (the `onClick={() => handleSelectOption(option)}` line below) and conceptually it seems strange that selecting an option from a dropdown menu would submit a form. 

This PR just adds `type="button"` to the buttons used in that chip dropdown menu. That gives these buttons the "client code will handle events on this button, don't submit or reset the form owner" signal as per spec. 

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Buttons in the Dismissable Chip add menu no longer submit ancestor forms

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
@@ -106,6 +106,7 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
               <button
                 key={option.value}
                 role="option"
+                type="button"
                 id={generateDescendantId(i)}
                 className={classNames(styles.menuListOption, {
                   [styles.activeOption]: activeIndex === i,


ZakaryH:
hey @TJKruitbosch thanks for making this PR

this is super weird. I did not know that about buttons.

>When investigating some flakiness in some internal Jest tests

which tests were flaking? if it's a test for a secret feature or something we don't want to mention on an open source repo feel free to DM me instead. 

I'm very curious myself now why it wouldn't happen in browser, and if we need to be setting this on all sorts of other instances of `button` that don't explicitly have a `type` to avoid this exact same scenario

TJKruitbosch:
Hey @ZakaryH; not super-secret, it's the settings page for Online Booking (the area where a user can select which users are bookable). I've got a draft PR #39192 in the main private repo that I'm just filling out the details of because there's more to it than just this if you want to read the small novel that is the description section

ZakaryH (approved):
tried this out with dismissable chips, confirmed they are still dismissing

LGTM!

darryltec:
For future us, added a linter warning for these so we don't forget next time https://github.com/GetJobber/atlantis/pull/1577/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R55

PR #1577, opened 2023-10-06
chore: automate code quality and nitpicks

darryltec:
## Motivations

This is an effort to automate Atlantis' code quality, consistency, and readability based on what we do right now. Opted to keep them as warn to get this in master first. I'll do a follow-up that fixes the violation + making it error out.

### Added

- Ensure code blocks, returns, and functions are spaced out so it's easier to spot
- Since react props are meant to be readonly, we should state them as is, it'll now automatically prefix them with readonly
- Tighten up the use of fragments and we should return `null` if we meant to not return a component instead of an empty fragment.
  - And also remove redundancy

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

run `npm run lint:js -- --fix` to see it work its magic. It's all a warning so it won't really do anything unless you run this script or open a file and save it

---

Random photo of Atlantis


taylorvnoj (approved):
LGTM - really like all these new rules 👌

ZakaryH:
>Ensure code blocks, returns, and functions are spaced out so it's easier to spot

sweet I like that, but how does this work? does it only match on functions defined like

```
function hello() {}
```

vs an arrow function?

```
const hello = () => {}
```

or does that second syntax simply not count, since you'd likely be defining your consts together and it wouldn't get teh spacing rule

> not mentioned, but the thing about button type

also like this because it is very sneaky

> Since react props are meant to be readonly, we should state them as is, it'll now automatically prefix them with readonly

I don't disagree with adding this rule but it is weird. like if everything is readonly, then readonly doesn't mean much? it's just.... always there. 

I wonder if we even have a single instance of something trying to mutate a prop. that seems like a very bad pattern that either the author or the reviewer would catch

>Tighten up the use of fragments and we should return null if we meant to not return a component instead of an empty fragment.

I don't disagree with this either, but I feel like both of these approaches aren't great. if you don't want to render a component, just don't render it at the level above the component! I don't believe components should have a `null` path in them. that said, there are exceptions, and this is probably the better way to handle those than returning `<></>`

darryltec:
@ZakaryH 

> vs an arrow function?

This one lands on the block and block-like. You can write it in 1 line which is technically not a block of code and it's still up to you to group them like a normal const. The moment a bracket comes in, that's when it becomes a block so it adds a space in between.

> I don't disagree with adding this rule but it is weird. like if everything is readonly, then readonly doesn't mean much? it's just.... always there.

It still means something. Especially to TS where it would throw an error on it. [See here](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChkHJQRwAm6IANgJ7JwBcyAzmFKAOYDc+hxZFNZACNGLNiC48CCUaw7cAvrlwIKLZDEyMMWALw4pdRgCI4xgDSGRyY0IuGZNhPaW4A9G+QBRKFHRQmXE10ADo4ZH1jUmMgzBChCJsIGPdPAHkAazhaJlAkZGAwAHImZBB0MCIScipqWNCEROMYYyA). It's not as strict with restructuring right now tho, but when TS does tighten that up, we're ready. 

> I wonder if we even have a single instance of something trying to mutate a prop. that seems like a very bad pattern that either the author or the reviewer would catch

We don't because we caught it. Since the inception of Atlantis, we've caught a few but it's mostly due to not knowing you shouldn't because raw JS doesn't really stop you. It's mostly a precaution on when we are big and someone did something bad and no one caught it. It's the same argument as "it's not scalable when the team get too big to catch these things".

> I don't disagree with this either, but I feel like both of these approaches aren't great. if you don't want to render a component, just don't render it at the level above the component! I don't believe components should have a null path in them. that said, there are exceptions, and this is probably the better way to handle those than returning <></>

Oh yah, but there are cases where an empty fragment is actually used. This one catches it so we have fewer things to catch like that one.

ZakaryH:
@darryltec 

>The moment a bracket comes in, that's when it becomes a block so it adds a space in between.

hmm, so it would auto format this

```
const name = "Bob";
const apiResult = fetchSomething()

const clickHandler = () => {
  console.log("hi")
}

const clickError = "Oh no";
const clickSucceed = "Oh yeah";
```

with those spaces around the `clickHandler`

I'm not sure how I feel about that. you could argue you should have moved `clickError` above with the non function variable definitions, but also it's kinda nice to group variables that are related to the same concept

adding this code warn/fixer is effectively saying that we believe all function definitions should be separate from other variables, which is a fine opinion to have but I feel like we could benefit some additional guidance to accompany this

should all non function variables be defined first so that the spacing feels more, I dunno natural? like

```
const name = "Bob"
const clickError = "Oh no"
const clickSucceed = "oh yeah"

const clickHandler = () => {
  console.log("hi")
}

const failHandler = () => {
 console.log("bye")
}
```

or, is having related logic spatially closer preferred, even if it adds more spaces like I had in that first example. OR is the answer that we shouldn't use arrow functions & variables to hold them like that, and that your functions should instead be 

`function clickHandler() {}`

darryltec:
@ZakaryH That's some good thoughts right there. But yeah, how would you suggest we guide them? 

The goal is still to ensure your code is spaced out so it is easier to catch those blocks. Both of them feel totally normal to me and is easier on the eyes. If I was writing that code i'd definitely use a function instead of the arrow and group them even more if i need/want to.

> I'm not sure how I feel about that. you could argue you should have moved clickError above with the non function variable definitions, but also it's kinda nice to group variables that are related to the same concept

Arguably, yeah. Did you notice that it also starts to make you think and organize your code a bit more and I think that's a win.

ZakaryH:
>That's some good thoughts right there. But yeah, how would you suggest we guide them?

hmm good question. honestly there might not be anything for us to do beyond PR reviews, and suggesting to group variables where it makes sense

automating variable declaration order sounds like a bad time. often the order does matter and cannot easily be shuffled around to be in the preferred order 

>Arguably, yeah. Did you notice that it also starts to make you think and organize your code a bit more and I think that's a win.

yeah I'd agree with that!

overall I'm not terribly worried about the case I mentioned, even in that example - the worst thing about it is that you might end up with extra spaces surrounding your arrow function, but like you've just said that could be seen as a good thing. maybe that prompts you to find another way to better organize your code.

PR #1581, opened 2023-10-11
feat(components): Adjust Combobox Selected Style

ZakaryH:
## Motivations

While playing around with Combobox in the muliselect case, it felt a bit weird to have multiple selections, click one to unselect it and it would still have the same grey background. Sure the checkmark was gone, but overall it still felt like it was selected-ish because selected, focus, and hover all applied that background.
Components___Selections___Combobox___Web_-_Multi_Select_⋅_Storybook_🔊

Thought I'd check out Polaris' implementation. Over there, they only apply the grey background on hover and focus which I like.

Ran this idea by Chris M, and he agreed it looked good.

This approach is great because it's so simple. No complicated logic where we only apply the selected class if it wasn't already selected or anything like that.
### Changed

No more extra selected style!

Also, since that style is gone I've had to update the tests that were looking for the style. Instead, we're looking for the `aria-selected` which is a very easy change that I think is pretty stable. I don't see us removing that anytime soon.

## Testing

Tests should still be passing for a green CI.

Aside from that, the only change should be that now the selected state is only represented by the checkmark. That's not to say you cannot have the grey background and the checkmark, since you can still focus or hover a selected item - but there will only be one now. It won't be a bunch of grey items.

Kapture 2023-10-11 at 15 08 46

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -32,6 +32,7 @@ describe("Combobox validation", () => {
   it("throws an error if there is no Trigger element", () => {
     expect.assertions(1);
     let error;
+
     try {
       render(
         <Combobox>
@@ -52,6 +53,7 @@ describe("Combobox validation", () => {
   it("throws an error if there are multiple of the same Trigger element", () => {
     expect.assertions(1);
     let error;
+
     try {
       render(
         <Combobox>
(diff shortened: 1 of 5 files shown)


chris-at-jobber (commented):
Design approved ✅  Great suggestion, really cleans up the interface especially in multiselect mode 👌

taylorvnoj (approved):
chef's kiss to the highest degree. ship it.

PR #1584, opened 2023-10-13
chore: reduce the amount of mocking a consumer would have to do when using DataList

darryltec:
## Motivations

When using DataList in a jest test environment, the test would immediately fail because `matchMedia` doesn't exist there. In turn, the consumers would have to mock `matchMedia` on their tests for not a lot of gain since breakpoint changes are already tested by DataList itself.

If the consumer does need to test a specific viewport–such as testing layouts of different screens–the can use the `mockViewportWidth` function from `@jobber/components/useBreakpoints`. Arguably, the layouts should be componentized and tested on their own, but if this happens, the consumer has a way.

### Added

- 

### Changed

- guard matchMedia on one of DataList's internal hook

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- use DataList in your project and run any test that uses DataList
- It will throw an error regarding `matchMedia` not existing
- This PR should let it pass

---

Random photo of Atlantis

JOB-80875

### DIFF
--- a/packages/components/src/DataList/hooks/useMediaQuery.ts
+++ b/packages/components/src/DataList/hooks/useMediaQuery.ts
@@ -1,6 +1,19 @@
 import { useEffect, useState } from "react";
 
 export function useMediaQuery(CSSMediaQuery: string) {
+  /**
+   * matchMedia have had full support for browsers since 2012 but jest, being a
+   * lite version of a DOM, doesn't support it.
+   *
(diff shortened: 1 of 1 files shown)


rodrigoeidelvein (approved):
LGTM! Can confirm that tests using DataList can be executed without using the `mockViewportWidth` from `useBreakpoints`!

PR #1585, opened 2023-10-13
fix(components): Avoid rendering DataListItemActions when there's no children

rodrigoeidelvein:
## Motivations

In a case where we want only to use `<DataList.ItemActions onClick={handleActionClick} />`, DataList was still rendering the overlay with the menu of `DataListItemActions`. This PR fixes it.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Avoid rendering `DataListItemActions` when there's no children

### Security

- 

#### Before
Screenshot 2023-10-13 at 17 21 20

#### After
image

---

Random photo of Atlantis


darryltec (approved):
good catch! 🚢

PR #1590, opened 2023-10-17
feat(components-native): add testID property to Button and default testID values to ButtonGroup

andrewsjobber:
## Motivations

As part of the e2e work, we need to modify the Button component to allow passing a testID so the contractors can select elements with language-agnostic unique identifiers. The way it works today, it's using the a11yLabel or label as id for testID and that doesn't work for us because when the app is running the Spanish version the e2e tests written will fail. 

I kept the previous implementation with the addition that now the testID prop will take precedence over the other two fallbacks. I also ended up modifying the ButtonGroup to achieve the same goal with the Form component. I know that we can have multiple ButtonGroup components in the same screen that could have the same testID with the current implementation, but it shouldn’t be a problem because they can always rely on xPath in these scenarios. I'm open to your suggestions too!

### Added

- Optional testID prop to the Button component
- Default testID values to the Buttons inside the ButtonGroup component

## Testing

You should be able to check these new testIDs by using Appium Inspector. I have the following video showing how it works. If you want to test by yourself, please get in touch so that I can help with the configuration.

https://github.com/GetJobber/atlantis/assets/98422073/71583587-a369-4658-9783-934430127861

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/ButtonGroup/ButtonGroup.tsx
+++ b/packages/components-native/src/ButtonGroup/ButtonGroup.tsx
@@ -50,6 +50,7 @@ export function ButtonGroup({
   const { handlePress } = usePreventTapWhenOffline();
   const secondaryActionsRef = useRef<BottomSheetRef>();
   const { primaryActions, secondaryActions } = getActions(children);
+
   return (
     <View style={styles.buttonGroup}>
       {primaryActions.map((action, index) => {
@@ -75,6 +76,7 @@ export function ButtonGroup({
                 fullHeight={true}
                 icon={icon}
                 loading={loading}
+                testID={`primary-action-${index}`}
               />
             )}
           </View>
@@ -88,6 +90,7 @@ export function ButtonGroup({
             accessibilityLabel={t("more")}
             onPress={handlePress(openBottomSheet)}
             fullHeight={true}
+            testID="secondary-action"
           />
         </View>
       )}
--- a/packages/components-native/src/ButtonGroup/ButtonGroup.tsx
+++ b/packages/components-native/src/ButtonGroup/ButtonGroup.tsx
@@ -76,7 +76,7 @@ export function ButtonGroup({
                 fullHeight={true}
                 icon={icon}
                 loading={loading}
-                testID={`primary-action-${index}`}
+                testID={`ATL-ButtonGroup-Primary-Action-${index}`}
               />
             )}
           </View>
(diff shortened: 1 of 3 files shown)


rodrigoeidelvein:
@andrewsjobber When it's testing the app in Spanish, maybe the test could be adapted to get the labels in Spanish instead of adding test IDs?

andrewsjobber:
> @andrewsjobber When it's testing the app in Spanish, maybe the test could be adapted to get the labels in Spanish instead of adding test IDs?

yeah, I had that same discussion with a few people and the problem with that approach is that it's not scalable. The contractors would have to keep inspecting the app in two languages and adding selectors for each language. If we decide to add a third language, the same process would need to be repeated to make all tests work.

We also thought about using the JSONs to be the source of truth, but we have many messages that are built in runtime with the usage of curly braces. Like the example below:
image

We couldn’t come up with a solution for that either.

packages/components-native/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -75,6 +76,7 @@ export function ButtonGroup({
                 fullHeight={true}
                 icon={icon}
                 loading={loading}
+                testID={`primary-action-${index}`}
```
rodrigoeidelvein, line 79: We should update this test id to be in the ATL-* pattern: https://jobber.atlassian.net/wiki/spaces/TEAMATL/pages/2476933462/How+should+we+declare+test+ID+s+in+Atlantis+components

packages/components-native/src/ButtonGroup/ButtonGroup.tsx
```diff
@@ -88,6 +90,7 @@ export function ButtonGroup({
             accessibilityLabel={t("more")}
             onPress={handlePress(openBottomSheet)}
             fullHeight={true}
+            testID="secondary-action"
```
rodrigoeidelvein, line 93: This one too

rodrigoeidelvein (changes requested):
If we have exhausted the possibilities to test these button using the accessibility label, than it's fine to use the test ID. I liked that the description of the prop mentioned that is meant to be used in end-to-end tests, because we should avoid it as much as possible for jest tests.

I just requested a change to update the test ID inside ButtonGroup to be in the ATL-* pattern, other than that, it's good!

rodrigoeidelvein (approved):
LGTM!

PR #1591, opened 2023-10-17
feat(components): Tabs supports caching active tab between page loads

mattander:
## Motivations

Workday has a ticket in an upcoming sprint asking us to add a tabbed view that remembers the last active tab on page load. This PR adds the props required to accomplish this.

I considered just building internal logic that would do a cache in local storage and check on first mount but we'd still need to provide a way to pass in cache keys. This seemed like the most balanced approach.

You'll notice I call the onTabChange callback **after** the active tab is set. I did this purposefully because this is a **callback**, not a hook. We aren't giving devs a hook into the tab change event, we're just letting them react to tab change events. 

## Changes

Updated the whole test file for Tabs because it was missing a `describe` block. Little campground cleanup.

### Added

- `onTabChange` optional prop: callback for when the active tab is changed. It provides new tab and prev tab as arguments. 
- `defaultTab` optional prop: change the active tab on component mount by passing a different tab index (it's usually the first tab, so `0`). If the provided index is out of bounds for how many tabs were passed in, it falls back on `0`. 

## Testing

- Try setting one or both props in the story to see how you can change behaviour
- I added a `describe` block as campground cleanup
---

Random photo of Atlanta


packages/components/src/Tabs/Tabs.test.tsx
```diff
@@ -17,32 +17,86 @@ const omelet = (
...
+          <p>Eggs</p>
+        </Tab>
+        <Tab label="Cheese">
+          <p>🧀</p>
+        </Tab>
+      </Tabs>,
+    );
+
+    expect(queryByText("🍳")).toBeFalsy();
+    expect(queryByText("🧀")).toBeTruthy();
```
darryltec, line 83:
```suggestion
    expect(queryByText("🍳")).not.toBeInTheDocument();
    expect(queryByText("🧀")).toBeInTheDocument();
```

the `.not.toBeInTheDocument` might be a bit weird to think about in terms of queryBy, but it's pretty much the same test in a more descriptive manner

packages/components/src/Tabs/Tabs.test.tsx
```diff
@@ -17,32 +17,86 @@ const omelet = (
...
+        </Tab>
+      </Tabs>,
+    );
 
-  fireEvent.click(getByText("Cheese"));
-  expect(count).toBe(1);
-  fireEvent.click(getByText("Cheese"));
-  expect(count).toBe(2);
+    expect(queryByText("🍳")).toBeTruthy();
+    expect(queryByText("🧀")).toBeFalsy();
```
darryltec, line 100:
```suggestion
    expect(queryByText("🍳")).toBeInTheDocument();
    expect(queryByText("🧀")).not.toBeInTheDocument();
```

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -12,10 +12,24 @@ import { Typography } from "../Typography";
...
+   * Specifies the tab that should be active on mount
+   */
+  readonly defaultTab?: number;
+
+  /**
+   * Callback that fires when the active tab changes
+   * @param newTabIndex
+   * @param prevTabIndex
+   */
+  onTabChange?(newTabIndex: number, prevTabIndex: number): void;
```
darryltec, line 26: oooh what's the useCase for `prevTabIndex` and how confident are we that that's the same value as what would be passed in the `defaultTab`?

thinking about it in this sense

```tsx
const [tab, setTab] = useState(1);

return <Tab defaultTab={tab} onTabChange={setTab}>...</Tab>
```

where the moment that mounts, the `tab` state updates everytime you change the tab therefore making it the previous tab index when `onTabChange` fires 🤔

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -12,10 +12,24 @@ import { Typography } from "../Typography";
...
 }
 
-export function Tabs({ children }: TabsProps) {
-  const [activeTab, setActiveTab] = useState(0);
+export function Tabs({ children, defaultTab, onTabChange }: TabsProps) {
+  const activeTabInitialValue =
+    defaultTab && defaultTab < React.Children.count(children) ? defaultTab : 0;
```
darryltec, line 31: I think we can safely set the defaultTab to 0 if it's not defined

```suggestion
export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
  const activeTabInitialValue =
    defaultTab < React.Children.count(children) ? defaultTab : 0;
```

that way you don't have to guard it

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -12,10 +12,24 @@ import { Typography } from "../Typography";
 
 interface TabsProps {
   readonly children: ReactElement | ReactElement[];
+
+  /**
+   * Specifies the tab that should be active on mount
+   */
```
darryltec, line 18: Would be good to also specify that the number value corresponds to the index

darryltec (changes requested):
This looks really good! just a minor cleanup and we are rock solid!

darryltec:
Oh, probably should add a storybook story for `defaultTab`

packages/components/src/Tabs/Tabs.test.tsx
mattander: This is good to know. I'm working on improving my jest-fu and this helps. Thanks!

packages/components/src/DataTable/Body.tsx
```diff
@@ -5,10 +5,10 @@ import styles from "./DataTable.css";
...
 
 interface BodyProps<T> {
-  table: Table<T>;
-  onRowClick?: (row: Row<T>) => void;
-  emptyState?: ReactNode | ReactNode[];
-  loading: boolean;
+  readonly table: Table<T>;
+  readonly onRowClick?: (row: Row<T>) => void;
+  readonly emptyState?: ReactNode | ReactNode[];
+  readonly loading: boolean;
```
mattander, line 11: FYI for some reason my local branch had a change in this file that accidentally bled into here. I cleaned up the change but this is being auto applied and I can't get rid of it 🤦

packages/components/src/Tabs/Tabs.tsx
mattander: I was thinking about more advanced use-cases like conditional tab logic.

Like "only tabs `1` and `3` would persist. (Why? Who knows?). Or maybe some analytics gathering advanced journey tracking like "users who view tab `2` and then `5` may be trying to find _x_ so present a modal offering to take them there. 

Nothing I've seen so far but it's easy to include and doesn't hurt anything so I figured I'd do it. 

In your example above, `prevTabIndex` would always be the same as the `tab` state at the time the callback is called.

mattander:
image

FYI for anyone 🤔 "wtf"

I had a chance from another branch that for some reason carried over here and I couldn't revert it fully because some of it was auto applied. Since I did this pretty fast, there was no risk to others having pulled the latest changes so it was safe (enough) to hard reset and force push. 😬

mattander:
> Oh, probably should add a storybook story for `defaultTab`

Done. As a bonus I also added one that demonstrates the callback. Let me know if you think that's too much.

packages/components/src/Tabs/Tabs.tsx
darryltec: Analytics gathering might be a good one. But you still have access to it outside of the component so I'm not sure if we should add it now. especially with this statement

> Nothing I've seen so far but it's easy to include and doesn't hurt anything so I figured I'd do it.

It's one of those easy to add but hella hard to remove

mattander: Okay, makes sense to me. 

For my understanding, is the concern here mainly that we're adding something that might not get used, so it makes sense to wait till it's needed? I only included it because I couldn't think of a way it would be used to disrupt user flow. I'm still building my comprehension of what our design system standards are in terms of what we expose to our users.

darryltec: Yeah, pretty much. There are a lot of easy things we could do on a generic component but it comes with a cost of maintaining that feature. If we're not sure what it would be used for, i'd rather not have it in there than someone using it for something it wasn't intended to do (since we don't know).

Same with writing a GQL API. There are some easy things you could add on the API as well but if we don't have a solid enough reason to add it, we shouldn't add it as it is much harder to remove if we realize that we don't need it.

mattander: Okay, I've removed that parameter!

packages/components/src/Tabs/Tabs.test.tsx
darryltec: Nice! unfortunately. can't use that in JO just yet. They're planning to add it so stay tuned.

Also, rotating to Atlantis is a possibility if you want more jest-fu and front-end tricksz 😉

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -12,10 +12,25 @@ import { Typography } from "../Typography";
...
+   * @param newTabIndex
+   */
+  onTabChange?(newTabIndex: number): void;
 }
 
-export function Tabs({ children }: TabsProps) {
-  const [activeTab, setActiveTab] = useState(0);
+export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
+  const activeTabInitialValue =
+    defaultTab < React.Children.count(children) ? defaultTab : 0;
```
darryltec, line 32: Thinking about this more. I wonder if it's better if it goes to the last index rather than reverting back to 0 when the value is higher than the children's length. 

Feels more forgiving when it goes 

> Oh you want 6 but there's only 5 so maybe you want 5

instead of

> Oh you want 6 but there's only 5 so here's 0

what do you think?

mattander: Kind of trying to catch a potential off by one error? 

I wonder if this treads to close to "magic". That's why I fallback on the first tab. Kind of like "you told me one that doesn't exist so I'll just be dumb and revert to non-controlled behaviour"

darryltec: hmm.. that's a good point

darryltec (approved):
:shipit: great work!

PR #1593, opened 2023-10-17
feat(components): Truncate Long Chip Content with Gradient

ZakaryH:
## Motivations

Chip can have a header and a label, which are dynamic content. this means that they could be super long and run out of space, or if you are on a small screen they could even be mediumly long and have the same problem.

so, there's a few things we shouldn't do about this. we shouldn't let it wrap and get very tall that's bad. we also shouldn't let it keep going and going outside its container because this is also not good. instead, what we want is some truncation which is what I'm doing here!

### Added

added the ability for the text to truncate with a fancy lil gradient when it gets too long

also, if it's in that scenario where you can't read the label fully because it got truncated, or the heading, we want to expose that via a tooltip containing the not visible text

### Changed

tooltip was not having a great time with this, so I had to modify it where it's still all rendered as far as setup - but if you give it an empty string it doesn't actually show. 

this is technically a breaking change but you're not losing anything super valuable. if you were rendering a tooltip with an empty string today I don't know why you're doing that, and going forward it will not show. 

### Fixed

changed the render logic slightly on Chip for `heading`. previously it would render an empty `p` tag for the typography of `heading` but I strongly prefer to not render things if they're going to be empty anyway

## Testing

**automated tests**
for right now we're gonna have to rely on the ol' 👀 for most of this which I don't love. gonna see if I can figure out some way to test this though 🤞 

**variations**
since the gradient needs to match the background color, please verify it looks good with all the variations and their states
- subtle
- invalid
- disabled
- base

then their permutations of
- hover
- focus

here's a mostly exhaustive list of the cases and what you should see

**only label, no heading, full label visible**
- tooltip absent
- gradient absent

image

**only label, no heading, full label not visible**
- tooltip has label content only
- gradient present

image

**label and heading, full heading visible, full label not visible**
- tooltip has label content only
- gradient present
- worth noting this is where things can get interesting, it looks like the gradient is on the heading but it counts as being fully visible so it's not included in the tooltip content

image

**label and heading, neither full heading nor full label visible**
- tooltip has heading and label content
- gradient present
- if ever the heading isn't fully visible, the label won't be either due to how this is set up

image

**label and heading, both fully visible**
- tooltip absent
- gradient absent

image

**for your viewing pleasure a few more gradient variations**

**subtle**
image

**subtle disabled**
image

**subtle invalid**
image

---

Random photo of Atlantis


packages/components/src/Chip/Chip.css
```diff
@@ -111,3 +112,76 @@
...
+.chipLabel {
+  position: relative;
+  overflow: hidden;
+}
+
+.chipHeading > p,
+.chipLabel > p {
+  overflow: hidden;
+  white-space: nowrap;
+}
```
ZakaryH, line 130: I don't love this but we need to stop the text from wrapping which it will do by default

if I add `maxNumberOfLines` that stops wrapping, but includes ellipsis truncation behavior that I don't wnt

chris-at-jobber: Maybe eventually we just... do the `maxNumberOfLInes` truncation with gradients as a rule in Text? I see that we're doing it within DataList as well, and as you've seen with Combobox and you can see in the DataList work, that we're starting to really consistently see these designs calling for gradient-based overflow indicators.

But for now yeah this totally makes sense.

packages/components/src/Chip/Chip.css
```diff
@@ -111,3 +112,76 @@
...
+
+.truncateGradient {
+  position: absolute;
+  top: 0;
+  right: 0;
+  width: var(--space-base);
+  height: 100%;
+  background: linear-gradient(
+    to left,
+    var(--chip-base-bg) 25%,
```
chris-at-jobber, line 140: Seeing all these repeated declarations makes me wonder if we can clean up our component-level "tokens"?

If you look at [Button](https://github.com/GetJobber/atlantis/blob/817bc0911d3f0f806a035559d6ab34954f4b5a31/packages/components/src/Button/Button.css#L59-L81) we avoid a lot of repetitive declarations by updating the variables on a state level.

ZakaryH: oh definitely

this is just me trying to get this working at all before optimizing it. there's some weird stuff happening that I need to figure out if we can fix or not with this approach

ZakaryH:
mildly interesting side note, you can have multiple tooltips showing thanks to focus being on one element and hover on another
image

ZakaryH:
oh my. I hadn't even considered this. due to us using `useInView` if for any reason the chip isn't in view, it'll trigger both our gradient which might not be visible thanks to margin, but also the tooltip 😅 

which makes sense. if your viewport is cutting off part of the label, well it's not visible which is exactly how the hook is supposed to work

I don't know if that's the worst? seems like it would be pretty uncommon for your chip to get cut off horizontally unless you're zoomed in or something. but if you are, this could get weird
Kapture 2023-10-18 at 16 41 57

packages/components/src/Chip/Chip.css
ZakaryH: k got this working and followed up with a similar approach with a single gradient definition that looks at a variable that we swap around depending on the variation/state

ZakaryH:
just pushed up a webstory with truncation examples of: only label, heading + label with label not visible, and heading + label with heading not visible

chris-at-jobber:
>if you had it right on the edge of the screen it might trigger?

Given that the [small-screen designs for DataList](https://job-80528-truncate-chips.atlantis.pages.dev/?path=/story/components-lists-and-tables-datalist-web--basic) explicitly call for side-scrolling Chips, seems pretty likely! I also don't think it's the worst, but I'll play with this on a phone to see how sensitive the Chip is to "hover" events on a touch device.

chris-at-jobber:
It is possible to get into a mis-matched gradient state here:

image

(Hover on a focused truncated Tooltip)

Also, the boundary for pointer interactions seems a bit touchy?

https://github.com/GetJobber/atlantis/assets/39704901/a03fc4b3-71d5-4743-a2e8-7996ad193f9d

ZakaryH:
@chris-at-jobber 

>It is possible to get into a mis-matched gradient state here:

nice catch. that's an interesting little intersection of states. I've fixed it by putting the `:hover` styles below `:focus` making it take priority, and added a comment. this is fragile though, so `!important` could be a more brute force, but resilient way to solve this.

>Also, the boundary for pointer interactions seems a bit touchy? Is there some very subtle border-resizing happening on hover that might be causing the boundary to keep resetting itself?

I noticed that too!

looking at the style, it goes from a `border: var(--border-base) solid transparent` to `border: var(--border-base) ${some other color}`

so I don't think it's from borders changing width. it's also not happening on current master, so something in this PR is doing it. if I comment out the gradient it still happens. if I comment out the Tooltip, it stops.

how much of a blocker do you think this is? I can start looking into it in a bit. it might be a quick fix, but it also might not be!

packages/components/src/Chip/Chip.tsx
```diff
@@ -31,29 +33,73 @@ export const Chip = ({
...
-    >
-      {prefix}
-      <Typography size="base" fontWeight="medium">
-        {heading}
-      </Typography>
-      {heading && <span className={styles.chipBar} />}
-      <Typography size="base">{label}</Typography>
-      {suffix}
-    </button>
+    <Tooltip message={tooltipMessage}>
```
scotttjob, line 46: I know this is throwing a wrench in things pretty late, and I know this is a pretty big abstract ask (so feel free to turn it down), but is there a way to pull apart these two pieces and combine them based on a prop? Like to separate the concern of Chip, and then a Chip with a Tooltip because it's too long?

Brainstorming here for ways to accomplish that, but could some of this functionality be pulled into a local hook, which is then enabled via prop so the functionality is close by, but not wrapping everything?

I'm just thinking down the road where maybe we want to repurpose Chip again for something else, and you've already established the super awesome pattern that allows Chip to be conditionally repurposed for greater things. Or, do we hold on to that idea until that situation actually comes up, and then do the work then. That's equally valid.

packages/components/src/Chip/Chip.tsx
```diff
@@ -31,29 +33,73 @@ export const Chip = ({
...
+            />
+          )}
+        </div>
+        {suffix}
+      </button>
+    </Tooltip>
   );
 };
 
+function getTooltipMessage(
```
scotttjob, line 87: Can we either turn this into a useMemo or flip the format to be a single return value? This is probably just my own personal style creeping in here though, so if the multiple return statements are a common pattern here feel free to ignore me.

packages/components/src/Chip/Chip.tsx
```diff
@@ -31,29 +33,73 @@ export const Chip = ({
...
+            </>
+          )}
+          <Typography size="base">
+            {label}
+            <span ref={labelRef} />
+          </Typography>
+          {!labelFullyVisible && (
+            <div
+              className={styles.truncateGradient}
+              data-testid="ATL-Chip-Truncation-Gradient"
```
scotttjob, line 77: Any chance we can bump this to a prop with a default of your value? I doubt anyone needs it today, but I could see someone down the road wanting to override this.

scotttjob (approved):
Great work Zak! I have some comments, but they're entirely optional and one might not even be legit.

packages/components/src/Chip/Chip.tsx
ZakaryH: I hear ya. unfortunately that's how I initially had it with a 

```
function ChipWrapper(props: {children: ReactElement, showTooltip: boolean}):JSX.Element {
if (props.showTooltip) {
  return <Tooltip {...whateverPropsGoHere}>{props.children}</Tooltip>
}

return props.children
}
```

where we invoked it with

```
<ChipWrapper>
  <Chip {...props}/>
</ChipWrapper>
```

the problem was that the tooltip kept getting removed and re-added infinitely, causing a really bad time overall. the fix was to always have the tooltip elements present, but not show them when the message is empty which is why that tooltip change was added in this PR

chris-at-jobber:
@ZakaryH I don't know if I'd want to block the whole PR on it, but 

>if I comment out the Tooltip, it stops

makes me feel like we at least should have a handle on what's happening before we move ahead. 

If the root of the problem is with Tooltip we should land this and fix that separately, but if the problem is inherent to how we're using it with Chip then we can better plan to either tackle it here or in a followup.

packages/components/src/Chip/Chip.tsx
ZakaryH: ideally, yes I would love to have that setup where Chip doesn't always have a Tooltip wrapping it, because I do agree it's a bit goofy to include it when you know full well it has no use. the interactions between `useInView` and `Tooltip` were just not having it :(

packages/components/src/Chip/Chip.tsx
ZakaryH: >turn this into a useMemo

do you mean to memo-ize the value of `tooltipMessage` in the component body where we call this? since this function itself is defined outside the component it's only being made the once

>multiple return statements are a common pattern here feel free to ignore me.

it's a fair question. I'm not sure if that's an established pattern or we have opinions on it in Atlantis. I personally don't mind them because I think they're quite readable, but that's also an opinion

what are you thinking for a single return value? something like

```
function getTooltipMessage() {
 let message = "";
 
  if (heading && !headingFullyVisible) {
     message = `${heading} | ${label}`;
  }

  if (!labelFullyVisible) {
    message = label;
  }

  return message;
}
```

packages/components/src/Chip/Chip.tsx
ZakaryH: I totally get where you're coming from on that. in fact I believe I either suggested or did the same thing at one point and someone gently suggested I not do that as per a guide we have. I'l send it to ya in a moment here

ZakaryH:
@chris-at-jobber 

>makes me feel like we at least should have a handle on what's happening before we move ahead.

yeah that's fair. mkay I will see what I can dig up.

ZakaryH:
@chris-at-jobber 

so this is already happening on other instances of `Tooltip`

image

you just can't see it happening because there's no different hover background style, nor cursor change - but it is happening over and over infinitely if you put your mouse cursor in exactly the right spot. I've actually seen this before with tooltips in other pieces of software, so I'm not terribly surprised

here you can see it a little more clearly if I add some `:hover` styles to the avatar

Kapture 2023-10-24 at 10 18 50

I've narrowed it down to these two lines in the Tooltip

```
        activator.addEventListener("mouseenter", showTooltip);
        activator.addEventListener("mouseleave", hideTooltip);
```

with these commented out the flickering stops, so there's something we'd need to adjust here to fix this issue for not only Chips but everywhere using a Tooltip. though again it's only going to have a visible issue if the activator element has a focus style, otherwise it will silently be toggling infinitely which is still not ideal.

chris-at-jobber:
OK, well knowing that the issue is with Tooltip itself I'm good to move ahead with this Chip update. We can work on a bug ticket for the flickering separately!

chris-at-jobber:
🎉

PR #1594, opened 2023-10-18
feat(components): JOB-79340/JOB-80485 - Separating Chip and Chips

scotttjob:
## Motivations

We needed a standalone Chip for DataList/ClientList/Combobox. Since our existing implementation required a group of chips, an effort began to redesign Chip, implement that new Chip, and then decouple Chip from Chips. We've already redesigned and implemented the new Chip, this is the final step, decoupling Chip from Chips.

This decoupling was tougher than originally anticipated due to the number of tests that were testing internal structure/composition over user behaviour, and due to some 'secret' `Chips` functionality that I initially disregarded as not needed (it was very needed). We've made a special effort to ensure all the original Chips tests are passing with only minor modifications. We've also done a round of integration testing with JO to ensure that this change has no impact on existing functionality.

In the future we may take additional steps to refactor Chips to be simpler, but that may require either a forked component or updating all instances of `<Chips>` within JO to reflect any API changes, so it's not planned for any time soon.

## Changes

Chips no longer has its own internal implementation of a Chip, instead using the decoupled/unopinionated `<Chip>` that is also available to all consumers of Atlantis.

Minor modifications to a few internal Chips tests to reflect updated linting rules (they were throwing errors upon opening the untouched files)

#### Previous Chips Single Select
image

#### New Chips Single Select
image

#### Previous Chips Multi Select
image

#### New Chips Multi Select
image

#### Previous Chips Selection
image

#### New Chips Selection
image

### Added

- No net new development

### Changed

- There should be no noticeable changes by consumers of Atlantis other than the design of Chips is now in line with `Chip`. 

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- N/A

### Security

- N/A

## Testing

1. Ensured all existing functionality in storybook is preserved.
1. Ensured all existing unit tests pass with the changes.
1. Ran a manual integration test with JO to ensure existing `Chips` functionality is preserved where in use.

---

Random photo of Atlantis


chris-at-jobber:
Some design/experience comments / questions

The suffix hover state color should use `color-surface--hover`, currently using `color-disabled` (I think I missed this on the previous PR to create Chip!)

image

The "Selection" chip should only be dismissed by clicking on the X, not the Chip itself

Reference current behaviour for example:
https://github.com/GetJobber/atlantis/assets/39704901/6be5fabc-834e-4a3b-bdef-b91ce39a2177

Single-select and multi-select Chips should not have a `+` indicator (or anything else) in the "suffix" slot when unselected, and the checkmark indicator should use `color-heading` - 
image
[Figma reference](https://www.figma.com/proto/rIIhulZvcp9M82lNOCGv16/branch/S6J5803Pjmy4VvpsQx0o8S/Product%2FOnline?page-id=&type=design&node-id=19126-4306&viewport=595%2C106%2C0.13&t=IzzU2X3FIUiBQbsP-1&scaling=min-zoom&starting-point-node-id=19126%3A4305&show-proto-sidebar=1&mode=design)

scotttjob:
> Some design/experience comments / questions
> 
> The suffix hover state color should use `color-surface--hover`, currently using `color-disabled` (I think I missed this on the previous PR to create Chip!)
> 
> image
> 
> The "Selection" chip should only be dismissed by clicking on the X, not the Chip itself
> 
> Reference current behaviour for example: https://github.com/GetJobber/atlantis/assets/39704901/6be5fabc-834e-4a3b-bdef-b91ce39a2177
> 
> Single-select and multi-select Chips should not have a `+` indicator (or anything else) in the "suffix" slot when unselected, and the checkmark indicator should use `color-heading` image [Figma reference](https://www.figma.com/proto/rIIhulZvcp9M82lNOCGv16/branch/S6J5803Pjmy4VvpsQx0o8S/Product%2FOnline?page-id=&type=design&node-id=19126-4306&viewport=595%2C106%2C0.13&t=IzzU2X3FIUiBQbsP-1&scaling=min-zoom&starting-point-node-id=19126%3A4305&show-proto-sidebar=1&mode=design)

I think I've addressed all of these issues now if you wanted to have another look.

Just for clarity, only the X being clickable is for `Chips` only, whereas in `ChipDismissible` within `Chip` the whole thing is clickable? Or should that pattern be applied to both?

It's a one-line change in ChipDismissible (moving the onClick from `Chip` to `Chip.Suffix`) so big deal if both need to follow that pattern.

taylorvnoj:
@scotttjob I'm still looking through the code, sorry that will take me a little longer.  A non blocker for this PR but something I think should be done, is to add a Docs page for `Chip`. Right now we only have the web example which are great but we should have a Docs page as well.

packages/components/src/Chip/Chip.css
```diff
@@ -23,6 +22,15 @@
   transition: all var(--ease-both) var(--timing-quick);
   align-items: center;
 }
+div.chip {
+  padding: 0 var(--base-unit);
+}
```
taylorvnoj, line 27:
```suggestion
}

```

packages/components/src/Chip/Chip.css
```diff
@@ -23,6 +22,15 @@
   transition: all var(--ease-both) var(--timing-quick);
   align-items: center;
 }
+div.chip {
```
taylorvnoj, line 25: As Darryl would say - let them breathe 😆 

```suggestion

div.chip {
```

packages/components/src/Chip/Chip.css
```diff
@@ -23,6 +22,15 @@
...
+  padding: 0 var(--base-unit);
+}
+button.chip {
+  padding: var(--space-small) var(--base-unit);
+}
```
taylorvnoj, line 30:
```suggestion
}

```

packages/components/src/Chip/Chip.tsx
```diff
@@ -24,6 +26,7 @@ export const Chip = ({
   variation = "base",
 }: ChipProps): JSX.Element => {
   const classes = classnames(styles.chip, {
+    [styles.actingAsDiv]: actAsFormElement,
```
taylorvnoj, line 29: `actingAsDiv` - maybe just `useDiv`?

packages/components/src/Chip/Chip.css
```diff
@@ -136,6 +144,9 @@
 .chip.disabled .suffix {
   background-color: var(--color-disabled);
 }
+.clickableSuffix {
```
taylorvnoj, line 147:
```suggestion

.clickableSuffix {
```

packages/components/src/Chip/Chip.tsx
```diff
@@ -44,17 +48,24 @@ export const Chip = ({
...
-          onClick && onClick(value, ev)
+        onClick={(ev: React.MouseEvent<HTMLButtonElement | HTMLDivElement>) =>
+          onClick &&
+          !actAsFormElement &&
+          onClick(value || label, ev as React.MouseEvent<HTMLButtonElement>)
         }
         tabIndex={tabIndex}
-        onKeyDown={onKeyDown}
+        onKeyDown={(
+          ev: React.KeyboardEvent<HTMLButtonElement | HTMLDivElement>,
```
taylorvnoj, line 60: Cool! I learned something new today - didn't realize you could just use `ev`.  Everywhere else we are using `event` so I'm not sure if there is a broader preference but I think this is neat & I support it!

packages/components/src/Chip/Chip.tsx
```diff
@@ -44,17 +48,24 @@ export const Chip = ({
...
+        ) =>
+          onKeyDown && onKeyDown(ev as React.KeyboardEvent<HTMLButtonElement>)
+        }
         aria-label={ariaLabel || label}
         disabled={disabled}
         role={role}
         type="button"
+        data-testid={dataTestID || "chip-wrapper"}
```
taylorvnoj, line 68: Something else I learned about today is the existence of [this wiki](https://jobber.atlassian.net/wiki/spaces/TEAMATL/pages/2476933462/How+should+we+declare+test+ID+s+in+Atlantis+components)

There are a ton of older components that are writing the testID the way you did but in `Combobox` and `DataList` we are using the convention described in that wiki.  So I guess this should instead be `ATL-Chip`?

packages/components/src/Chip/Chip.types.d.tsx
```diff
@@ -1,11 +1,20 @@
 import React, { PropsWithChildren } from "react";
 
 export interface ChipProps extends PropsWithChildren {
+  /**
+   * If you need Chip to be part of a form element, this makes it not a button, but just a div so you can wrap it in a label.
+   */
+  readonly actAsFormElement?: boolean;
```
taylorvnoj, line 7:
```suggestion
  readonly actAsFormElement?: boolean;
  
```

packages/components/src/Chip/Chip.types.d.tsx
```diff
@@ -1,11 +1,20 @@
...
   /**
    * Accessible label, which can be different from the primary label.
    */
   readonly ariaLabel?: string;
 
+  /**
+   * Enables pass through of a data-testid. This is legacy functionality to allow old tests to pass.
```
taylorvnoj, line 14: > This is legacy functionality to allow old tests to pass

Would we call this legacy? We still use testID's in new tests to make sure we're rendering the right component.

What do you think about a description like this: `Used to locate this view in end-to-end tests`

taylorvnoj (changes requested):
Hey @scotttjob - I haven't gone through all the files yet but feel like this is a good place to post my review comments so far.  

Looking really great, I'm learning new things (which is why I love reviewing so much) and have a few suggestions - mostly non functional ones.  Please feel free to push back if you feel strongly!

packages/components/src/Chip/Chip.tsx
scotttjob: In this particular case it's because I noticed references to `chip-wrapper` in the existing tests, and I thought someone else out there might have done the same, so in this case it's a compatibility thing which is why I kept it as chip-wrapper instead of changing it to the new pattern. If this were a whole new component I would agree to update the ID to be the new pattern.

If we're convinced though that updating this id is non-breaking, I'd be happy to update it both here and in the tests to be inline with Combobox/Datalist.

docs/components/Chips/Web.stories.tsx
```diff
@@ -24,40 +30,66 @@ export default {
...
+                    size={c.size as AvatarSize}
+                  />
+                )}
+                {c.initials && (
+                  <Avatar initials={c.initials} size={c.size as AvatarSize} />
+                )}
+              </Chip.Prefix>
+            </Chip>
+          );
+        })}
```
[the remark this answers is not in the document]
scotttjob: Done! Totally just being lazy here, `c` is obviously not a good variable name and I would have had the same feedback. Thanks Taylor! Great eye here.

docs/components/Chips/Web.stories.tsx
```diff
@@ -69,10 +101,14 @@ const MultiSelectTemplate: ComponentStory<typeof Chips> = args => {
         selected={selected}
         onChange={setSelected}
       >
-        <Chip label="Amazing" value="Amazing" />
-        <Chip label="Wonderful" value="Wonderful" />
-        <Chip label="Brilliant" value="Brilliant" />
-        <Chip label="Magnificent" value="Magnificent" />
+        <Chip label="Amazing" onClick={multiSelectItems} value="Amazing" />
```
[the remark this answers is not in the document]
scotttjob: lol done! But yeah the symmetry is nicer than what I had.

packages/components/src/Chip/Chip.css
scotttjob: I've made all these spacing updates, but I feel like we could probably write a few eslint rules to catch this one in the future. I'll make a note.

packages/components/src/Chip/Chip.css
scotttjob: Done!

packages/components/src/Chip/Chip.css
scotttjob: Also done!

packages/components/src/Chip/Chip.tsx
scotttjob: Done!

packages/components/src/Chip/Chip.css
scotttjob: Done!

packages/components/src/Chip/Chip.tsx
scotttjob: I updated it to be `event` to match the others I don't like standing out from the crowd! 

Really it's just a variable, we can call it anything. I vote for `keyboardEventTaylorsVersion` :)

packages/components/src/Chip/Chip.types.d.tsx
scotttjob: Done!

packages/components/src/Chip/Chip.types.d.tsx
scotttjob: I had initially stripped out the data-testid because I didn't think we needed it, we could just grab the element via the label (since we're setting it) or the role, but a lot of the existing tests were using the data-testid so I put it back in with that comment so the tests I didn't were needed (they were) would pass.

You're right though, the comment about legacy functionality was for us and not consumers. I included a modified version of your comment in my own to reflect the true nature of the prop. Thanks Taylor!

scotttjob:
> @scotttjob I'm still looking through the code, sorry that will take me a little longer. A non blocker for this PR but something I think should be done, is to add a Docs page for `Chip`. Right now we only have the web example which are great but we should have a Docs page as well.

This is a great comment. I've made myself a note to make that happen.

scotttjob:
> Hey @scotttjob - I haven't gone through all the files yet but feel like this is a good place to post my review comments so far.
> 
> Looking really great, I'm learning new things (which is why I love reviewing so much) and have a few suggestions - mostly non functional ones. Please feel free to push back if you feel strongly!

I've now gone through your feedback and addressed everything up to this point (Monday Morning). Even if I didn't make a code change in two cases, I replied to all of your comments. 

I'm not sure how we `resolve conversations` at Jobber, I personally like when the person who made the original comment is the conversation resolver so that they can ensure their requested changes were made, but I also acknowledge we might do things differently here (which is cool)! I'm just not sure of the process so I left them all 'unresolved' for now. If I'm supposed to resolve them instead of typing 'Done!' on them, I can do that too. :)

packages/components/src/Chip/Chip.css
taylorvnoj: @scotttjob that was my thought when doing the suggested change haha. Thanks for making a note - I also haven't really done eslint rule stuff so can put it side desk for myself too - whoever gets there first! 👍

packages/components/src/Chip/Chip.tsx
taylorvnoj: lol! Sounds good to me 😆

packages/components/src/Chip/Chip.tsx
taylorvnoj: Ahh ok good point - nah leave it as is! Maybe one day we can do a cleanup because there's more components **not** using the `ATL` format than there is right now.

packages/components/src/Chip/Chip.types.d.tsx
taylorvnoj: Perfect - love the update!

packages/components/src/Chip/components/ChipDismissible/ChipDismissible.test.tsx
```diff
@@ -21,14 +21,14 @@ describe("Chip Dismissible", () => {
     expect(container.querySelector("svg")).toBeInTheDocument();
   });
 
-  it("calls onClick when clicked", () => {
+  it("does not call onClick when clicked because suffix is clickable", () => {
```
taylorvnoj, line 24: Should we also test the opposite? `calls onClick when the suffix is clicked`

packages/components/src/Chip/components/ChipPrefix/Chip.Prefix.tsx
```diff
@@ -5,15 +5,25 @@ import { Icon } from "@jobber/components/Icon";
...
+        showBG,
+      )}
+    >
       {singleChild}
     </span>
   );
 }
+
+interface ChipPrefixProps extends PropsWithChildren {
+  readonly showBG?: boolean;
```
taylorvnoj, line 28: Is `showBG` show background? I know it's internal but I think the prop name could benefit from having the whole word - what do you think?

packages/components/src/Chip/components/ChipSuffix/Chip.Suffix.tsx
```diff
@@ -26,6 +46,13 @@ export function ChipSuffix({ children, className }: ChipSuffixProps) {
 
 export interface ChipSuffixProps extends PropsWithChildren {
   readonly className?: string;
+  readonly showBG?: boolean;
+  readonly onClick?: (
+    value: string | number | undefined,
```
taylorvnoj, line 51: Should we follow Atlantis pattern and put props before the function? 

Is there any technical benefit to where props are located within the file or is it purely personal preference? Maybe this is something we can talk about in one of our tech strategy meetings or in a retro because I hate to say "this is the way it's always been done" without personally knowing why we always put props on top

packages/components/src/Chips/ChipsTypes.tsx
```diff
@@ -50,19 +50,24 @@ export interface ChipSingleSelectProps extends ChipFoundationProps {
...
 
   /**
-   * Use a custom activator to trigger the Chip option selector
+   * Element that triggers opening the ChipDismissible Menu
    */
-  readonly activator?: ReactElement;
+  activator?: ReactElement;
+
+  /**
+   * Enables pass through of a data-testid. This is legacy functionality to allow old tests to pass.
```
taylorvnoj, line 68: Oh let's match this description with the one we just updated 😄

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
```diff
@@ -28,30 +28,35 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
...
       {sortedVisibleChipOptions.map(chip => (
-        <InternalChip
+        <Chip
           key={chip.value}
           {...chip}
-          onKeyDown={handleChipKeyDown(chip.value)}
-          onClick={handleChipClick(chip.value)}
+          onClick={(
+            inVal: string | number | undefined,
+            ev: React.MouseEvent<HTMLButtonElement>,
```
taylorvnoj, line 36: we've got a couple more `ev`'s in this file 😄

taylorvnoj (changes requested):
Just a few more questions. Really great work Scott, this is a lot of work!

For the `resolved` question - unless it's written somewhere I haven't seen, everyone handles GitHub comments differently.  I agree - I think it makes sense for the original reviewer to resolve a comment once they've seen it done. 👍

packages/components/src/Chip/components/ChipDismissible/ChipDismissible.test.tsx
scotttjob: Lol yup. Thanks for keeping me honest. The functionality changed (comment from Chris) and I was too lazy to go add a testId to the Suffix so I could grab it in the test. I've done that now, and the test is passing. Great piece of feedback Taylor! Good eye.

packages/components/src/Chip/components/ChipPrefix/Chip.Prefix.tsx
scotttjob: It sure does! I've updated it. That's just years of ingrained habit of not wanting to type `background` over and over it's an awkward combination of letters to type. Not nearly as bad as awkward though, phew. :)

packages/components/src/Chip/components/ChipSuffix/Chip.Suffix.tsx
scotttjob: I think it's just personal preference. I like to put the props below the component when they grow large enough to no longer be nicely inlined. When I open a component file I like to see the component, not a bunch of props/types first. But again, that's just personal preference, I think I intended to move those back above after I was done but clearly forgot! :)

packages/components/src/Chips/ChipsTypes.tsx
scotttjob: Donezo.

packages/components/src/Chip/components/ChipDismissible/ChipDismissible.test.tsx
taylorvnoj: love it! thank you!

packages/components/src/Chip/components/ChipSuffix/Chip.Suffix.tsx
taylorvnoj: That reason does make sense! Thanks for moving it, maybe for internal components it's not that big of a deal but something we could still chat about with the team.  Never a bad thing to revisit why we do things a certain way

taylorvnoj:
Sorry Scott - 1 more thing!  In the `Web.stories.tsx` file, those `props` need to change to `args`, it actually comes directly from [Storybook](https://storybook.js.org/docs/react/writing-stories/args)

Screenshot 2023-10-30 at 11 42 50 AM

So once we change from props to args, we can see the component and the code sandbox without error.
Screenshot 2023-10-30 at 11 44 35 AM

This is another type of error that stems from Storybook not understanding our imports:
Screenshot 2023-10-30 at 11 51 10 AM

To fix that, we should add this to `extraImports` at the top of the file:
```
          "@jobber/components/Chip": [
            "Chip",
            "ChipDismissible",
            "ChipSelectable",
          ],
```
Plus whatever other  subcomponents of Chip are needed.
Screenshot 2023-10-30 at 11 52 10 AM

docs/components/Chips/Web.stories.tsx
```diff
@@ -1,10 +1,16 @@
...
 import { Content } from "@jobber/components/Content";
-import { Chip, Chips } from "@jobber/components/Chips";
-import { Icon } from "@jobber/components/Icon";
+import {
+  Chips,
+  InternalChipDismissible,
+  InternalChipMultiSelect,
+  InternalChipSingleSelect,
+} from "@jobber/components/Chips";
+import { Chip } from "@jobber/components/Chip";
```
darryltec, line 10: Should this Chip be the chip from `@jobber/components/Chips` with an s?

darryltec:
Tested in JO and it works in there, however it surface a few issue that is replicable here.

- The `disabled` prop doesn't seem to work if you use the `Chip` component from `@jobber/component/Chip`or `@jobber/component/Chips`. Ran this `git checkout master docs/components/Chips/Web.stories.tsx` since the example from master is what we're using in JO.

- Tabbing experience requires double tabbing to get to the next since the div is tabbable; as well as the inputs behind it. Happens on both single and multi select
  - This also creates the focus jump when you mousedown, the focus shows up, mouseup moves it to the input.

- Prefix icon/avatar seems to have disappeared

packages/components/src/Chip/Chip.types.d.tsx
```diff
@@ -1,11 +1,21 @@
 import React, { PropsWithChildren } from "react";
 
 export interface ChipProps extends PropsWithChildren {
+  /**
+   * If you need Chip to be part of a form element, this makes it not a button, but just a div so you can wrap it in a label.
+   */
+  readonly actAsFormElement?: boolean;
```
darryltec, line 7: Hmmm kinda misleading as it is still technically not a form element when used on its own. I get the need for it, but the prop name is odd.

packages/components/src/Chip/Chip.tsx
```diff
@@ -44,17 +48,25 @@ export const Chip = ({
...
   return (
     <Tooltip message={tooltipMessage}>
-      <button
+      <Tag
         className={classes}
-        onClick={(ev: React.MouseEvent<HTMLButtonElement>) =>
-          onClick && onClick(value, ev)
+        onClick={(ev: React.MouseEvent<HTMLButtonElement | HTMLDivElement>) =>
+          onClick &&
+          !actAsFormElement &&
```
darryltec, line 55: this is interesting tho as we're swallowing the `onClick` when you have `actAsFormElement` as true. A bit of a clashing prop.

why not just skip the extra prop and define the tag logic from onClick existing?

packages/components/src/Chip/Chip.tsx
```diff
@@ -44,17 +48,25 @@ export const Chip = ({
...
+          onClick(value || label, ev as React.MouseEvent<HTMLButtonElement>)
         }
         tabIndex={tabIndex}
-        onKeyDown={onKeyDown}
+        onKeyDown={(
+          event: React.KeyboardEvent<HTMLButtonElement | HTMLDivElement>,
+        ) =>
+          onKeyDown &&
+          onKeyDown(event as React.KeyboardEvent<HTMLButtonElement>)
+        }
```
darryltec, line 64: i don't think you have to narrow this down since it could be a div

chris-at-jobber:
Some interaction issues:

- the cursor when I hover the `Chip` doesn't give a pointer in single-select or multi-select mode

- I can't activate the multi-select Chips with `space` when they're focused; but I can activate them with `space` in between tabs (I assume there's an invisible checkbox somewhere that's receiving focus in the interim steps)

- I can't delete the "selection" Chips by keyboard. I should be able to either 
  - press delete when Chip focused
  - tab into focused Chip to focus the delete button
_Note that with a screen-reader, it still announces that I can delete with `delete` when I have a Chip focused_

- I can't use left/right arrow keys to navigate the "selection" Chips or the single-select Chips

packages/components/src/Chip/Chip.types.d.tsx
```diff
@@ -26,6 +31,11 @@ export interface ChipProps extends PropsWithChildren {
    */
   readonly label: string;
 
+  /**
+   * If you need Chip to be part of a form element, this makes it not a button, but just a div so you can wrap it in a label.
+   */
+  readonly mode?: "button" | "form";
```
chris-at-jobber, line 37: Could this be why we're having issues with the keyboard interactions?

packages/design/assets/icons/starHalf.svg
chris-at-jobber, line 1: What's this change doing here?

chris-at-jobber (changes requested):
Interaction issues noted [here](https://github.com/GetJobber/atlantis/pull/1594#issuecomment-1838924693), then some more question-y comments on the code

scotttjob:
I've started a new branch with a much better PR. I'm closing this version. I really appreciate all the feedback here, it really pushed me to build this better.

PR #1598, opened 2023-10-19
feat(hooks): Hook for managing boolean values

rjdestigter:
## Motivations
Working on a different component (InputDate, DatePicker, and CalendarDatePicker) and using this hook in more than one place to track whether an element has focus or is selected.

#### useBool hook
`useBool` is a simple wrapper around `useState` for managing boolean values. It
provides a simple API for toggling the value, setting it to true, and setting it
to false.

#### Example

```tsx
const [checked, on, off, toggle] = useBool(false);
return (
  <>
    <Checkbox checked={checked} onChange={toggle} />
    <Button onClick={on} label={"On"} />
    <Button onClick={off} label={"Off"} />
  </>
);
```

#### Types
Used branded typing so that hovering over the returned setters doesn show `() => void` but a helpful label:

image

### DIFF
--- a/packages/hooks/src/useBool/useBool.ts
+++ b/packages/hooks/src/useBool/useBool.ts
@@ -1,0 +1,28 @@
+import { useCallback, useState } from "react";
+
+declare const brand: unique symbol;
+
+type Callback = () => void;
+
+export type SetTrue = Callback & {
+  [brand]: "SetTrue";
+};
+
+export type SetFalse = Callback & {
+  [brand]: "SetFalse";
+};
+
+export type Toggle = Callback & {
+  [brand]: "Toggle";
+};
+
+export function useBool(
+  initialState = false,
+): [boolean, SetTrue, SetFalse, Toggle] {
+  const [state, setState] = useState(initialState);
+  const setTrue = useCallback(() => setState(true), []);
+  const setFalse = useCallback(() => setState(false), []);
+  const toggle = useCallback(() => setState(current => !current), []);
+
+  return [state, setTrue as SetTrue, setFalse as SetFalse, toggle as Toggle];
+}
--- a/packages/hooks/README.mdx
+++ b/packages/hooks/README.mdx
@@ -5,6 +5,7 @@ Shared hooks for components in Atlantis.
 ## Hooks
 
 - [useAssert](../?path=/docs/hooks-useassert--page)
+- [useBool](../?path=/docs/hooks-usebool--page)
 - [useCollectionQuery](../?path=/docs/hooks-usecollectionquery--use-collection-query)
 - [useFormState](../?path=/docs/hooks-useformstate--use-form-state)
 - [useIsMounted](../?path=/docs/hooks-useismounted--use-is-mounted)
--- a/packages/hooks/src/index.ts
+++ b/packages/hooks/src/index.ts
@@ -1,4 +1,5 @@
 export * from "./useAssert";
+export * from "./useBool";
 export * from "./useBreakpoints";
 export * from "./useCollectionQuery";
 export * from "./useFocusTrap";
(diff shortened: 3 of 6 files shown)


darryltec:
Ooooh this is sweet! Do you foresee people only wanting toggle and not on, off? That would end up like 

```tsx
const [checked, , , toggle] = useBool(false);
```

yeah?

packages/hooks/src/useBool/useBool.ts
```diff
@@ -0,0 +1,28 @@
+import { useCallback, useState } from "react";
+
+declare const brand: unique symbol;
+
+type Callback = () => void;
+
+export type SetTrue = Callback & {
+  [brand]: "SetTrue";
+};
```
ericchernuka, line 9: I've never seen this style of declaration. Any chance you can explain the rationale?

rjdestigter: In this particular case it's used to make the type opaque so that when inspecting a value of this type (like hovering over it in the code editor) it displays `SetTrue` rather than  `() =>void`

This strategy is also called branding sometimes. It allows you to add specificity to primitive types. Often useful when interacting with the outside world (incoming network responses, user input, etc).

For example, maybe you have a function that takes a number but it has to be a positive number. You can restrict this at a type level:

```ts
// positiveNumber.ts
declare const brand: unique symbol;

type PositiveNumber<T extends number> = T & {
  [brand]: "PositiveNumber";
};

export function isPositiveNumber<T>(number: T): number is PositiveNumber {
  return number >= 0
}
```

```ts
function foobar(n: PositiveNumber) { ... }
```

Rather than having to throw or return an error inside _foobar_ you now let the consumer of foobar deal with the fact that they can't just pass any number. If the consumer has a value of of type _number_ and they'd like to use _foobar_ they will have to make a check themselves:

```ts
// consumer.ts
function myFunc(n: number) {
  if (isPositiveNumber(n) {
     
  }
```

ericchernuka: Thanks for taking the time to write that up. Very cool. Thanks!

rjdestigter: Here's a nice video on the concept as well: https://www.youtube.com/watch?v=rpw59rajUSI

ericchernuka:
@darryltec, that's a great point. In the past, I've created the shape of the following:
```
const [checked, handlers] = useToggle()
```
`handlers` then was an object of `toggle`, `on`, `off` and `reset`

darryltec:
@ericchernuka yeah, talked to John about it last week and that's what I suggested as well. Still pretty torn around how it's easier to rename if it's not in a handler object if you have multiple. Just sucks if you only need `toggle`

rjdestigter:
> @ericchernuka yeah, talked to John about it last week and that's what I suggested as well. Still pretty torn around how it's easier to rename if it's not in a handler object if you have multiple. Just sucks if you only need `toggle`

We could also add `useToggle` as a composition over `useBool`. Have our cake and eat it?

ericchernuka:
@darryltec. I ended up always carrying around `handlers` or whatever contextual name made sense and using it directly instead of destructuring.

rjdestigter:
Noticed a similar pattern was used [here](https://github.com/GetJobber/Jobber/blob/979a42d27417f2cee748844501d8c573a7352b2a/app/javascript/jobber/payments_react/useBooleanToggle.ts).

Are there any change requests for this PR @darryltec ?

darryltec:
oh nope. sorry was swamped with things. I'll take one more looksies right now. As for the toggle, i like that idea for when people starts doing that pattern of `[checked, , , toggle]`

darryltec (approved):
:shipit: Looks good!

PR #1602, opened 2023-10-20
test(components): Upgrade @testing-library/user-events

rjdestigter:
## Motivations

I wanted to make use of the latest API for testing using keyboard interactions for the Calendar Date Picker. See https://github.com/GetJobber/atlantis/pull/1601

## Changes
Upgraded the package to the lastest version and adjusted wherever `userEvent` is used as it now returns promises.

Recommended reading for the team: https://testing-library.com/docs/user-event/intro

When using `jest.useFakeTimers` you have to configure _userEvent_:

```ts
const user = user = userEvent.setup({
  advanceTimers: jest.advanceTimersByTime,
});
...
await user.click(..)
await user.keyboard(..)
```

Tests for _DataListSearch_ are a good example here.

image

### DIFF
--- a/packages/components/src/DataList/components/DataListItemActions/DataListItemActions.test.tsx
+++ b/packages/components/src/DataList/components/DataListItemActions/DataListItemActions.test.tsx
@@ -42,18 +42,18 @@ describe("DataListItemActions", () => {
     expect(screen.getByLabelText("More actions")).toBeInTheDocument();
   });
 
-  it("should only render the first 2 actions", () => {
+  it("should only render the first 2 actions", async () => {
     renderComponent();
     expect(screen.getByLabelText("Edit")).toBeInTheDocument();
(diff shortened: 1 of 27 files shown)


packages/components/src/DataList/components/DataListItemActions/DataListItemActions.test.tsx
```diff
@@ -42,18 +42,18 @@ describe("DataListItemActions", () => {
     expect(screen.getByLabelText("More actions")).toBeInTheDocument();
   });
 
-  it("should only render the first 2 actions", () => {
```
darryltec, line 45: this doesn't seem to have an `await` nor a `userEvent` inside it 🤔

darryltec (approved):
LGTM! Just saw 1 thing but it's harmless enough that I think we can let that go

PR #1606, opened 2023-10-23
fix(components): Fix Combobox Types Exports

ZakaryH:
## Motivations

I messed up the type with the recent cleanup PR 😬 

### Fixed

previously we were exporting more than we needed to. no one needs to import the various combobox prop types, so we can stop exporting them from `index.ts`. the only one people will need to reference outside the component for their usages is `ComboboxOption` which is now the only thing being exported

while I was trying fixes, I changed the type of Combobox from a const arrow function to a normal function which now matches how we've defined all the other sub components of Combobox. additionally now it matches how other compound components like DataList wrt their `.d.ts` files
image

this doesn't really fix or change anything other than it might save someone troubleshooting in the future one rabbit hole to go down because this made no difference. it just looked like it might!

**Import fix**

the real fix here is referencing the `Combobox.types` relatively from within the sub components. this is where things were breaking, unless I'm totally misunderstanding the problem (but this does fix it!)

what was happening is that the types were entirely missing in Jobber, so you'd get no suggestions on the Combobox prop types, and then this error was happening because it lost its types resulting in an `any`
image

though something I don't quite get is when it's safe to do that VS when it's not. sometimes my IDE is suggesting I import from a relative path, and other times it suggests importing from `@jobber/components/Combobox/Combobox.types`

so I'm not sure when to trust that...

ok so this is broken, you can see the inferences from TS are not suggesting anything here
image

this was after updating as per my IDE's suggestion in the TriggerButton to the `@jobber` import
image

so I should not have listened here. it seems to be if it's an internal component use the `@jobber` but if it's user facing, use the relative? 

compared to if I swap this back

image

it's working 🎉 
image

## Testing

make a pre-release (I'll try making one but it was being weird earlier)

or `npm run pack` it yourself, and then install that from Jobber

should see that the TS error in the single `<Combobox>` instance is gone, and when you start typing props to the sub components of Combobox, it should correctly inform you of them and their types

of course the Combobox should also render as expected

---

Random photo of Atlantis


ZakaryH:
@darryltec tagging you on this because I believe you were involved with the path aliasing. at least I think this is related!

packages/components/src/Combobox/index.ts
```diff
@@ -1,3 +1,3 @@
-export { Combobox } from "./Combobox";
+export * from "./Combobox";
```
ZakaryH, line 1: this isn't necessary strictly speaking

while troubleshooting I was narrowing down all possible differences between a working example with DataList and Combobox, where DataList is doing it like this

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/DataList/index.ts#L1

the change here can be reverted, or left as is. it turned out to not be relevant at all.

darryltec (approved):
Good catch. Basically, the main issue is it's not resolving it when packing. could be a rollup issue

https://www.npmjs.com/package/@jobber/components?activeTab=code

PR #1608, opened 2023-10-24
fix(hooks): Fix useCollectionQuery error state when fetching more

rodrigoeidelvein:
## Motivations

Fixing the error state of `useCollectionQuery`. Before, it was just doing a `console.error` when it failed on `fetchMore`. I added a state for the error and updated it when necessary.

I also added a test for this case.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Fixed the error state of `useCollectionQuery` when it fails on `fetchMore`

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/hooks/src/useCollectionQuery/useCollectionQuery.test.tsx
+++ b/packages/hooks/src/useCollectionQuery/useCollectionQuery.test.tsx
@@ -274,6 +274,7 @@ describe("useCollectionQuery", () => {
                 },
                 error: mockError,
               },
+              buildListRequestMockForNextPage(query, responseMock),
             ]),
           });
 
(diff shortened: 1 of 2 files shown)


darryltec:
We should reset that error state when it actually goes through. Right now it keeps it in there even if it successfully does the query the second time.

rodrigoeidelvein:
@darryltec I added a reset before the methods where they can be updated, that it makes sense or should we reset on a successful call?

darryltec:
@rodrigoeidelvein yeah since the error isn't valid at that point, yeah?

rodrigoeidelvein:
@darryltec exactly, it basically cleans up before making the request because we don't know what's going to happen next

darryltec:
Ok, one more ask, let's write a test on the clearing the error part

rodrigoeidelvein:
@darryltec I added one more expect to that same test to make sure it's cleared after a successful fetch

darryltec (approved):
:shipit:

PR #1609, opened 2023-10-24
fix(components): eliminate Tooltip/activator flicker

chris-at-jobber:
## Motivations

In https://github.com/GetJobber/atlantis/pull/1593#issuecomment-1777767499 we uncovered that Tooltip is causing flickers on any activator with a hover state.

https://github.com/GetJobber/atlantis/assets/39704901/f650c880-ae80-4b5b-93fc-5e3208722fd6

### Fixed

- no more flicker at the edge of a Tooltip activator component

https://github.com/GetJobber/atlantis/assets/39704901/4efda7ee-9217-4b89-8ec4-37b38cb8a638

## Testing

Hover around the edges of an element with a Tooltip

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.css
+++ b/packages/components/src/Tooltip/Tooltip.css
@@ -12,6 +12,7 @@
   display: inline-block;
   position: absolute;
   z-index: var(--elevation-tooltip);
+  pointer-events: none;
 }
 
 .tooltip {


ZakaryH (approved):
it fixes the issue! 🎉 

and even better it's such a small change!

so from that perspective, this LGTM

the one thing I am now curious about is _why_ this works. I _think_ I get it, but I'd love to understand it better to know what to look out for in the future

>This example disables pointer events (clicking, dragging, hovering, etc.) on all images.
https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events#disabling_pointer_events_on_all_images

so I assume what this is doing is saying that the wrapper cannot receive clicks, drags, or hovering where hovering is what was causing us grief, the hover away from our trigger and then going back and forth forever

I think that's fine? we should never have content that you need to click in the Tooltip, so not being able to hover it shouldn't be an issue. if you want to see the tooltip, you hover over the trigger - never the tooltip. if someone wanted to say copy text in a tooltip, well they already wouldn't be able to do that today, but now they extra won't be able to

chris-at-jobber:
As for _why_ this works, I believe this being the solution means that the underlying problem is related to the sequencing here:

image

PR #1610, opened 2023-10-24
feat(components): Implement Custom Activator to Combobox

taylorvnoj:
## Motivations

We need to allow consumers of `Combobox` to build their own custom activator, or be given a default `ComboboxTrigger`

### Added

 
 Added a `ComboboxActivator` component which allows consumers to set the trigger of the `Combobox` to be a Button or Chip and it will assign the role of `combobox` to either trigger type.
  
 This way, consumers can customize the appearance of their trigger, as well as handle the output UI as they want to (in a form, separate chips in a card, etc).  If anything other than a `Chip` or `Button` is a child of the `ComboboxActivator`, it will not be given the role of 'combobox'.
 
 Implementation:
 ```
 <Combobox>
      <Combobox.Activator>
        <Button
          label="Select Teammate"
          variation="subtle"
          type="primary"
          icon="arrowDown"
          iconOnRight={true}
        />
      </Combobox.Activator>
      <Combobox.Content
        options={[
          { id: "1", label: "Bilbo Baggins" },
          ...rest of the code
```
 
 -----
Also added a `ComboboxTrigger`.  
When `ComboboxActivator` is not used, the `ComboboxTrigger` will be the default.

```
 <Combobox multiSelect heading="Tags">
      <Combobox.Content
        options={[
          { id: "1", label: "Bilbo Baggins" },
             ...rest of the code
```

`ComboboxTrigger` will:
- render as a Chip with 'add' suffix  
- will also display the selectedOption(s) as the Chip label as they are selected
- when a selection is made, the Chip variation will change from `subtle` to `base`

Once a selection is made in single select, the selected option will become the Chip label.
Once a selection is made in multi select, the Chip will keep the `heading` visible and the `label` will be the list of selectedOptions separated by a comma

Examples of `ComboboxTrigger`:
 
 #### Single select
 
 Kapture 2023-10-26 at 11 40 31
 
 #### Multi select
 
Kapture 2023-10-26 at 11 41 54

### Changed

Update to Chip [here](https://github.com/GetJobber/atlantis/pull/1610/files#diff-af91b758ebbfe83ef19aa7898e882479b307e8178e11e2aa5defaf07fa063240R67); the `chipBar` needs to only be visible when the label is also present

`useComboboxValidation` also had to be updated so that consumers can implement `Combobox` without explicitly setting a `Trigger` (because `ComboboxTrigger` will then be the default).  The verbiage in the errors will change when `ComboboxTriggerChip` & `ComboboxTriggerButton` are officially deleted.

### Deprecated

`ComboboxTriggerButton` and `ComboboxTriggerChip` have been deprecated.

Once this PR is merged, we can double check usages of `Combobox` outside of Atlantis and then **delete**
 these two subcomponents as they are no longer needed.
 

## Testing

Use Storybook to test that `ComboboxTrigger` works as expected:

In the Basic (single select) example:
- select an option and make sure the Chip label changes to that option
- click on the Chip again to reopen the `Combobox` and make a new selection, making sure that new option is now the label of the Chip

In the Multi Select example:
- as you make selections, the options should be immediately load into the Chip label 
- when there is more than one selection, the Chip label should be separating options with a comma
- clicking outside of the `Combobox` or on the trigger will close the element and the Chip still reads `Heading | Option 1, Option 2`

---

Random photo of Atlantis

### CODE REVIEW

File: packages/components/src/Chip/Chip.tsx [hunk 1]
```diff
@@ -48,7 +48,7 @@ export const Chip = ({
       <Typography size="base" fontWeight="medium">
         {heading}
       </Typography>
-      {heading && <span className={styles.chipBar} />}
+      {heading && label && <span className={styles.chipBar} />}
```
taylorvnoj, line 51: I walked through this change with Chris and Scott - only show the chipBar when there is a header and label

chris-at-jobber: 🙏

chris-at-jobber:
Not sure if this is still WIP, but I'm noticing that the trigger's selection behaviour differs for single and multi. I'm _assuming_ this is because we're updating the Chip on _close_, and not actually on _selection_?

✅ Single: Chip updates when selection is made 
❌ Multi: Chip does not update when selection is made ([Figma](https://www.figma.com/file/gQNDrpJYRFAwfbP3jRO0an/DSC-%5B-Combobox-%5D?type=design&node-id=1561%3A98969&mode=design&t=fQr8jCkyKgXKoRcG-1) for intended behaviour)

https://github.com/GetJobber/atlantis/assets/39704901/a9d273b7-fe36-45d3-88ab-f3ae7b065a43

taylorvnoj:
@chris-at-jobber thank you for pointing that out. This is definitely still a WIP but I didn't notice that.
You are right, the multi-select example was using `onClose` so when switching that to `onSelect`, the behaviour is as expected.  Now I'm wondering if that needs to be a documented rule or something that needs to be somehow coded in 🤔 

Another thing that doesn't match the figma is "[Single selections will clear on menu reopening](https://www.figma.com/file/gQNDrpJYRFAwfbP3jRO0an/DSC-%5B-Combobox-%5D?type=design&node-id=1515%3A111782&mode=dev).".  Right now, when I make a selection in single select and reopen the Combobox, the selected option is still what is visible in the chip.  Just want to confirm that we should also update that behaviour to match the figma?

scotttjob (approved):
LGTM! I pulled it down, ran everything, Chip looks great in situ everything worked as I expected it to.

Code looks good as well, I have some overall thoughts on testing but they're Atlantis/JO wide, not specific to this PR so I'll hold those until I can show the whole team. :)

The only tiny question I had is probably a known thing. Is this button supposed to be so big? Does it just fill the available container? I know it's not even related to Combobox though, so it's just a question definitely not blocking in any way.

image

chris-at-jobber:
> Is this button supposed to be so big? Does it just fill the available container

That's a "for docs purposes only" button but it would definitely look better if we remove the `fullWidth` prop on it 😄

chris-at-jobber:
I actually wonder what we need it for? In single-select, clicking on the selected thing should clear selection, and in multi-select we have a "clear" action in the comboboxheader

chris-at-jobber:
A few comments/questions, but I don't think any are directly tied to this PR!

chris-at-jobber:
I have _one_ comment and it's non-blocking, but we should have a story for this so teams can actually reference the example

ZakaryH:
>I'm seeing that clicking a selected item does not clear the combobox. Why is that? We should be effectively mirroring the experience of a checkbox with these options.

popping in here because I needed a break from my task

for multiselect that's true it's checkbox-y, but for single select it's more of a radio button.

that said, we can make it do whatever we want! I've seen single select combobox implementations that un-select on click if it was already selected, and others that do nothing if you click on something that was already selected so we've just gotta decide what we want ours to do

taylorvnoj:
> is the point of that story to show that you can clear the combobox from outside the combobox

yup!

> In single-select, clicking on the selected thing should clear selection

> I'm seeing that clicking a selected item does not clear the combobox. Why is that?

I think I need more detail on this @chris-at-jobber. Was there an update from [this comment](https://github.com/GetJobber/atlantis/pull/1610#issuecomment-1779512575) where we decided _clear single selection on opening_ wasn't the right path?

In single-select, when you open the Combobox and make a selection, the Combobox will close because there is nothing else to be done.  If you reopen, you can choose another selection or clear the selection by calling `setSelected[]`. 
If I'm interpreting this right, I think you want the behaviour of clicking a selected item to match what it does in multi-select?

Open Combobox, make selection. Combobox closes. Re-open Combobox, click same selection.  Option is now unselected and Chip goes back to original state?

Kapture 2023-10-27 at 15 46 18

> we should have a story for this so teams can actually reference the example

Just to clarify, are you talking about the Clear Selection story? Or adding another story?

> it would definitely look better if we remove the fullWidth prop on it

There is no `fullWidth` prop on it 😆 But I can put a Content around it to shrink it

@scotttjob @chris-at-jobber

taylorvnoj:
nvm I fixed the full width problem with a Storybook decorator and then added a div around the Button and Combobox to make them a row again.
Screenshot 2023-10-27 at 4 00 23 PM

done here: 37c6bd800494051284f5d8bda3dc649605e19eac
@scotttjob

taylorvnoj:
I'm going to merge this to unblock a few other tasks - @chris-at-jobber  lets chat about next steps elsewhere 😄

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -11,8 +11,13 @@ import { useComboboxAccessibility } from "../../hooks/useComboboxAccessibility";
 import { ComboboxContentProps, ComboboxOption } from "../../Combobox.types";
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
-  const { open, setOpen, wrapperRef, multiselect } =
+  const { open, setOpen, setSelected, wrapperRef, multiselect } =
     React.useContext(ComboboxContext);
+
+  useEffect(() => {
+    setSelected(props.selected);
+  }, [props.selected]);
```
ZakaryH, line 19: hmmm I'm trying to think of a way to avoid making another copy of `props.selected` here

I worry this is getting pretty confusing where if it's `onSelect` you are operating on `props.selected` which is now `selected` from the provider

but also we are operating on `props.selected` via `selectedOptions` which already figures out for you if we are either using `props.selected` or the internal state

I feel like there's gotta be a way we can leverage the props/state we already have to track this

---------

if all we're doing is making a copy of `props.selected` to put into the provider to make it available to the trigger and we only ever use `setSelected` once, I'm tempted to just pass that as a prop to the provider when we initialize it instead

chris-at-jobber:
Posting here for transparency/visibility: [this](https://www.figma.com/proto/rIIhulZvcp9M82lNOCGv16/branch/BTne4kcNajZPeaJNvVirtM/Product%2FOnline?page-id=19147%3A2774&type=design&node-id=19375-3357&viewport=-81%2C2086%2C0.5&t=jraxuHywCQaC59II-1&scaling=scale-down&starting-point-node-id=19375%3A3357&show-proto-sidebar=1&mode=design) is what I was expecting to see happen when I clicked on an already-selected single-select item. I do think we can decide on what we collectively expect this to do because the radio analogy is also fair!

_(Ignore the outdated Chip styling, that Figma branch has yet to merge)_

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
darryltec: yeah, once we do the new API, we'll wack this out since we don't need to. I'd actually do that on my PR since I have access to the selected now and can pass it in

ZakaryH: haha I ended up doing the same thing in my PR too, just passing it in to the provider

PR #1613, opened 2023-10-26
feat(components-native): add custom testID to Select component

andrewsjobber:
## Motivations

Some screens in the app has more than one select and we don’t have way to uniquely identify these components. So following the[ testID recommendations document](https://jobber.atlassian.net/wiki/spaces/TEAMATL/pages/2476933462/How+should+we+declare+test+ID+s+in+Atlantis+components#Exposing-test-ID-as-a-prop), I added a custom testID for the Select.

### Added

- new testID property to the Select component

## Testing

You can pass a testID to any component in the app using this version and it should be present with the new pattern: ATL-{customId}-Select. If not present, should just be as it's.

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/Select/Select.test.tsx
+++ b/packages/components-native/src/Select/Select.test.tsx
@@ -25,7 +25,6 @@ afterEach(() => {
 
 const defaultPlaceholder = "Select an option";
 
-// eslint-disable-next-line max-statements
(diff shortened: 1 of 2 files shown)


taylorvnoj (approved):
LGTM - thanks for adding this!

PR #1614, opened 2023-10-26
feat(components): Refactor ComboboxContent to ComboboxOption

ZakaryH:
## Motivations

we swapped to the API for options being

```
 <Combobox.Option id={1} label={"yurp"} />

</Combobox>
```

this meant a couple things
- now `Combobox.Option` is the element that needs to access certain values via Context because we don't want to expose everything via props given it is now user facing
- ComboboxContent is flipped to be the private one

with that, much of the architecture was no longer sensible. we could move a lot of responsibilities out of the provider, up out of the Content and into the top level Combobox. then of course some things had the inverse experience, where they needed to be re-organized to be passed to Option but not through props. this refactor is based on that.

**Removing Triggers and Content**
then we wanted to fully remove the client facing components of `ComboboxContent` `TriggerButton` and `TriggerChip` where the latter two are fully gone, and `ComboboxContent` still exists but it's now only internally used.

**onClose**
lastly, we decided to revisit our decision around `onClose` being an XOR prop with `onSelect` because it was getting awkward both for a client to use, and for us to code around. it still exists, but serves a different purpose now and provides no `selection` arg.

the same flow can be achieved but with different surrounding code. now, your state will be immediately updated as selections are made, but you have an `onClose` callback that allows you to decide to wait to perform a potentially expensive operation at that point with the state that is already up to date.

### Added

- 

### Changed

- onClose & onSelect are no longer an XOR
- onClose remains a prop, but is now just a callback for when it closes - no args for selections provided
- with this onClose, consumers can wait to trigger their potentially expensive operation
- context and props reorganized since content is now a private component
- different pieces of state put into context since Option is now "separate"
- re-worked how we find the first selected element and scroll to it
- changed validation now that Content is not required, nor Options, nor Trigger
- updated webstories
- updated/relocated/deleted tests (some were no longer valid, some were checking things that were not their responsibility, some were redundant)
- label (chip heading) now shows up regardless of if it's single select, multiselect, or if you have a selection

### Removed

- the whole wait until close behavior is gone
- Fully removed TriggerButton and TriggerChip, and all code related to them
- removed some tests asserting stuff around Content & Triggers

### Fixed

technically fixed some bugs around `onClose` firing on re-render, but it's only fixed because that behavior is totally gone 🙃 
## Testing

test coverage should be solid. there might be a couple things that slipped thru the cracks while moving most of the sub component tests into more unit-y tests. if you think of any or notice any, just holler!

aside from that, there's so much to test. it's pretty much everything that has been touched. I'll try to make a list but really just play around with it in every way imaginable with every prop permutation you can think of

---

Random photo of Atlantis


ZakaryH:
oh no. something I just realized that is going to be a nuisance is mapping over data to create `Combobox.Option`s

React doesn't like this because it doesn't have a single parent, and if we give it a fragment, well now that's the parent and we need to check for fragments with options as children. possible not the worst, but it doesn't just work out of the box

ZakaryH:
misc thoughts:

- having 2 distinct hover appearances - 1 for the chip, 1 for the suffix - is odd. I feel like I expect it to do something different but it doesn't
- not being able to select-all after making a selection because now it's only "clear" feels not great

packages/components/src/Combobox/hooks/useComboboxValidation.ts
```diff
@@ -1,58 +1,70 @@
...
+): boolean {
+  return activators.length > 1 ? true : false;
 }
 
-function isContentElement(child: ReactNode): boolean {
-  return React.isValidElement(child) && child.type === ComboboxContent;
+/**
+ * Return all instances child component that matches the `type` provided
+ */
+export function getCompoundComponents<T>(
```
ZakaryH, line 59: originally I was borrowing this from DataListUtils

then we had to update the types to allow for mapping with the whole `ComboboxNode` meaning the `getCompoundComponents` now needed to accept `ComboboxNode` as a type for `children` which is waay to tied to Combobox

with that, I've copy/pasted a variation of it here. I don't like this very much, but I haven't had the time to think about how to make this generic & uncoupled from `ComboxboxNode` which is the type of `children` we must deal with

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -1,58 +1,91 @@
...
+    <ComboboxContextProvider
+      multiselect={props.multiSelect}
+      selected={selectedOptions}
+      selectionHandler={handleSelection}
+      open={open}
+      setOpen={setOpen}
+      handleClose={handleClose}
+      shouldScroll={shouldScroll}
+    >
+      <div ref={wrapperRef}>
```
darryltec, line 58: since buttons and chips are inline elements, gotta style this div to be inline-block instead of block, it'll always take up the whole row since it is wrapped in a block element

darryltec (commented):
This looks hella good to me! Just have that 1 show stopping comment about the div but all in all it is solid with the testing i did and matching it up with what's in master. Also tested if we put it in a modal to see if it renders above it and it's all good!


Did find an issue on multiselect, If I select the 3rd option and press down, the focus jumps to the first one on the option. I expected it to go to the 4th option. This is happening on master tho so could just log a bug haha.

ZakaryH:
woah the github comment UI changed and I was very lost for a moment

>the focus jumps to the first one on the option. I expected it to go to the 4th option. This is happening on master tho so could just log a bug haha.

@darryltec yeah that's known, and we went back and forth on what should happen there. 

I think your expectation is valid, and one we thought about. I believe we have all the state and whatnot needed to achieve what you're describing since we know the first selected element, and we are setting where the focus goes on arrow presses , it would just be a matter of hooking it all up!

ZakaryH:
last few commits do a couple things

**label/heading**
label (chip heading) now shows up regardless of if it's single select, multiselect, or if you have a selection

this is because with single select it would be pretty non obvious what the heck this UI element is for. it could just say `All` potentially, whereas with multiselect it would say `Teammates | George, Ben` and it's clear what those things mean

so now it will say `Dinner Choice | Steak` rather than just `Steak` for the single case

I do think it's debatable to use `All` as an option for a single select when multi select offers that same behavior in a much more flexible way, but with the use case we have in mind - this is necessary because multiselect won't work

**goodbye multiselect**

since `ComboboxTrigger` no longer has to ask "hey am I multiselect"? no one else needs that info, at least not from the provider. it can be accessed as a prop in `ComboboxContent` and its children. with that, I've removed it as a prop/value from the provider

**updated tests**

as you'd expect, some tests failed on this change so I've updated them and added a new assertion as well!

darryltec (approved):
LGTM!

PR #1616, opened 2023-10-27
feat(components): Implement option and deprecate content

darryltec:
## Motivations

In an effort to simplify the API of Combobox and ensure JO gets a some buffer time before we pull the rug under it this PR is an intermediary state when we do remove the Triggers and Content components within Combobox.

I left the test as is as it is still technically testing what the internal ComboboxContent could do until the refactor comes.

### Added

- Added a Combobox.Option and ensure we only have either the Option or the Content being rendered
- Internalize the logic on building the options

### Changed

- Deprecate Combobox.Content

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -228,41 +228,39 @@ const ComboboxMultiSelection: ComponentStory<typeof Combobox> = args => {
   const [selected, setSelected] = useState<ComboboxOption[]>([]);
 
   return (
-    <Combobox {...args} multiSelect heading="Teammates">
-      <Combobox.Content
-        options={[
-          { id: "1", label: "Bilbo Baggins" },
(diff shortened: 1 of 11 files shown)


taylorvnoj (approved):
Looks good, tested the validation which looks good but when I click on the Code tab I'm still seeing this error. Strange Storybook thing or? I've done a clean and rebuild but that sentence isn't even in the codebase anymore 🤔 

But I did see the right validation error message when actually building Combobox with Content, etc

Screenshot 2023-10-30 at 5 22 55 PM

darryltec:
> But I did see the right validation error message when actually building Combobox with Content, etc

Unfortunately, Code is dependent on production where that error exist. Should go away once this is released

PR #1617, opened 2023-10-30
chore(components): Update the Combobox test to follow the new API

darryltec:
## Motivations

This rewrites the test to follow the new API and assumes the old compound components are no more. It would be easier to review this via editor mode (press `.`) and look for the file there as the file has completely changed.

### Added

- 

### Changed

- Combobox test to reflect the new API and tests more of the user interaction

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -1,532 +1,339 @@
 import React from "react";
-import { cleanup, fireEvent, render, screen } from "@testing-library/react";
-import { ComboboxOption } from "./Combobox.types";
+import { render, screen } from "@testing-library/react";
+import { userEvent } from "@testing-library/user-event";
 import { Combobox } from "./Combobox";
+import { ComboboxOption } from "./Combobox.types";
 import {
   COMBOBOX_OPTION_AND_CONTENT_EXISTS_ERROR,
   COMBOBOX_TRIGGER_COUNT_ERROR_MESSAGE,
 } from "./hooks/useComboboxValidation";
-import { Chip } from "../Chip";
+import { Button } from "../Button";
 
 // jsdom is missing this implementation
 const scrollIntoViewMock = jest.fn();
 window.HTMLElement.prototype.scrollIntoView = scrollIntoViewMock;
 
-afterEach(cleanup);
+const MENU_TEST_ID = "ATL-Combobox-Content";
+const OVERLAY_TEST_ID = "ATL-Combobox-Overlay";
 
-describe("Combobox validation", () => {
-  it("renders without error if the correct count and composition of elements are present", () => {
-    expect(() => {
-      render(
-        <Combobox>
-          <Combobox.TriggerButton label="Button" />
-          <Combobox.Content
-            options={[]}
-            onSelect={jest.fn()}
-            selected={[]}
-          ></Combobox.Content>
-        </Combobox>,
-      );
-    }).not.toThrow();
+const activatorLabel = "Select a Baggins";
+const handleAction = jest.fn();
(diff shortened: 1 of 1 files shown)


packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -1,530 +1,339 @@
...
-            <Chip variation="subtle" label="Teammates" />
-          </Combobox.Activator>
-          <Combobox.Content
-            options={[]}
-            onSelect={jest.fn()}
-            selected={[]}
-          ></Combobox.Content>
-        </Combobox>,
-      );
+const activatorLabel = "Select a Baggins";
```
ZakaryH, line 19: so this does work, but what do you think about instead of having a static heading + getByText, we do `getByRole("combobox")` which will target the activator button in a more generic way

this will help with translations and all that sort of stuff down the road!

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -1,530 +1,339 @@
...
-    fireEvent.click(button);
-
-    expect(getByTestId("ATL-Combobox-Content")).not.toHaveClass("hidden");
-
-    const option = getByText("Bilbo Baggins");
-    fireEvent.click(option);
-
-    expect(getByTestId("ATL-Combobox-Content")).toHaveClass("hidden");
+  it("should not show the heading when there's a selection", () => {
+    expect(screen.queryByText(activatorLabel)).not.toBeInTheDocument();
```
ZakaryH, line 103: hmmm in this case though, we do want that activator label. so I guess we do need to keep it around for assertions like this, but maybe we use the `getByRole` for finding the button VS checking the button text like this

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -1,530 +1,339 @@
...
 
-      expect(onClose).toHaveBeenCalledWith([
+    it("should call onClose with selections when the content is closed", async () => {
+      await userEvent.click(screen.getByText(activatorLabel));
+      await userEvent.click(screen.getByText("Bilbo Baggins"));
+      await userEvent.click(screen.getByText("Frodo Baggins"));
+      await userEvent.click(screen.getByTestId(OVERLAY_TEST_ID));
+
+      // This should be 1. If this errors out and expects the value to be 1,
+      // then you've fixed the bug! Please to change the value to 1.
```
ZakaryH, line 242: couple things with this test wrt the big ol refactor PR

1) it won't really exist soon because `onClose` doesn't have args, it's just a lifecycle hook type thing
2) it _should_ fix this problem but yeah we can verify that once these live together

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -1,530 +1,339 @@
...
 
-    it("should not update consumer as selections are made", () => {
-      const { getByText, queryByText } = render(
-        <MockMultiSelectOnCloseCombobox />,
-      );
-
-      const button = getByText("Click Me");
-      const option = getByText("Bilbo Baggins");
-      const spoder = getByText("Shelob the Spoder");
+    it("should not update consumer as selections are made", async () => {
```
ZakaryH, line 250: this one will also be gone soon, but whatever given the current state this is valid

ZakaryH (approved):
one suggestion around using `getByRole`

then the rest are really just me thinking out loud/planning for when this merges

it's probably fine to merge this, then I pull it into my PR and adjust the tests that don't need to be asserted anymore

also I'm fine with addressing the `getByRole` thing in my PR since I know I have to make adjustments in this file no matter what

PR #1618, opened 2023-10-30
feat(components): Make Combobox heading prop optional

taylorvnoj:
## Motivations

`heading` prop in `<Combobox>` needs to be optional for flexibility when using either a `Combobox.Activator` or the default `ComboboxTrigger`.

### Changed

 
`heading` prop on `ComboboxProps` is now optional.

When a consumer of `Combobox` is using using a `Combobox.Activator` as the trigger, they will be able to set the label of the `Button` or `Chip` directly within that component.  This use case does not have a need for a `Combobox` `heading` so to avoid errors, `heading` needs to be optional.

When a consumer does not use a custom activator, `Combobox` will default to using `ComboboxTrigger` which is a Chip.  It is in this implementation, that a `heading` should be set.  We decided to set a default for this Chip heading (`Select`) and then the consumer can use the `heading` prop to change that as they need to.

If a `heading` prop is put on a `Combobox` that is using a `Combobox.Activator`, the label of the activator will remain.  I have decided to include this in the documentation in a later PR.  In my opinion, there doesn't need to be any safe guarding here but if there is a different take on that, I'm happy to hear it!

`<Combobox heading="Teammates">`

Screenshot 2023-10-30 at 3 04 56 PM

`<Combobox>`

Screenshot 2023-10-30 at 3 05 44 PM

## Testing

Make sure `<Combobox>` is not giving an error if there is no `heading`. 
In the Basic Storybook example, remove the `heading="Teammates"` on line 24 to see the Chip heading read `Select` instead.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/Combobox.types.ts
+++ b/packages/components/src/Combobox/Combobox.types.ts
@@ -14,9 +14,9 @@ export interface ComboboxProps
   readonly multiSelect?: boolean;
 
   /**
-   * The Chip heading for the trigger
+   * The label for the Combobox trigger
    */
-  readonly heading?: string;
+  readonly label?: string;
 
   readonly onSelect?: (selection: ComboboxOption[]) => void;
   readonly onClose?: (selection: ComboboxOption[]) => void;
--- a/packages/components/src/Combobox/Combobox.types.ts
+++ b/packages/components/src/Combobox/Combobox.types.ts
@@ -15,7 +15,7 @@ export interface ComboboxProps {
   /**
    * The Chip heading for the trigger
    */
-  readonly heading: string;
+  readonly heading?: string;
 }
 
 export interface ComboboxTriggerProps {
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -25,7 +25,7 @@ const BasicCombobox: ComponentStory<typeof Combobox> = args => {
       {...args}
       onSelect={setSelected}
       selected={selected}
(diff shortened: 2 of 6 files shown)


packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -15,7 +15,7 @@ export interface ComboboxProps {
   /**
    * The Chip heading for the trigger
    */
-  readonly heading: string;
+  readonly heading?: string;
```
darryltec, line 18: Now that i'm looking at this closely. should this be called heading? I get that it is called heading on the chip but in the event (future proofing) that we changed our mind and say it's not a chip anymore, does heading still make sense as a combobox prop? Basically decoupling what is exposed to the consumer and what it is doing behind the scenes.

If we look at all the _input_-driven things we have. Inputs have placeholders, Checkbox/Radios have labels, should this just be called label? (Again, decoupling it from what is being implemented behind the scenes and what would the consumer reach for if they do need to add it)

taylorvnoj: Yup good point. I've been going back and forth about using `label` or `placeholder` since future iterations will likely include Selects and Inputs, etc.  
Let's go with `label` since a placeholder could be an example of an expected value, and that's not what this prop is doing.

darryltec: Sounds good! This looks good to me so just need that 1 change and we solid

darryltec (approved):
:shipit:

PR #1619, opened 2023-10-30
feat(components): add missing role to modal component

ericchernuka:
## Motivations
When writing integration tests, it's helpful to be able to wait for the Modal to appear and then use the [@testing-library within helper](https://testing-library.com/docs/dom-testing-library/api-within/) to scope additional queries. The issue is that currently, Modal does not have a role and is, therefore, not easy to lookup using the accessible queries. To avoid that problem, we can add `role` to the root modal div to ease this burden.

### Changed
- Adds `role="dialog"` to root Modal element for accessibility and testing concerns.

## Testing
- Using Storybook, navigate to the Modal component. 
- Select the element using the dev tools to see the appropriate `role="dialog"` on the root element for the Modal component.


packages/components/src/Modal/Modal.test.tsx
```diff
@@ -5,6 +5,11 @@ import styles from "./Modal.css";
 
 afterEach(cleanup);
 
+test('modal contains aria role of "dialog"', async () => {
+  const { findByRole } = render(<Modal open>Content</Modal>);
+  expect(await findByRole("dialog")).toBeDefined();
```
darryltec, line 10:
```suggestion
  expect(await findByRole("dialog")).toBeInTheDocument();
```

exists here in atlantis 😉

darryltec (approved):
:shipit:

PR #1620, opened 2023-10-30
chore: remove testing library cleanup

ericchernuka:
## Motivations
The need to manually cleanup `@testing-library/react` has long passed, so good riddance.

Reference: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-cleanup
Reference 2: https://testing-library.com/docs/react-testing-library/api/#cleanup

### Changed
- Removed manual `cleanup` import from `@testing-library/react` from all test files.
- Some linting changes also came in upon making changes but were limited to adding missing `readonly` types within test files.


darryltec (approved):
Test passed so i guess were good! Thanks for updating the generator as well

PR #1621, opened 2023-10-31
docs: Update Combobox documentation for Activator and new API

taylorvnoj:
## Motivations

With the recent API change, we need to update documentation in Storybook.

### Changed

- updated order of stories in the `web` file just for better readability 
- added a story for `Combobox.Activator` with multiple examples using `Button` and `Chip`
- updated the `Trigger` and `Options` section of the documentation to reference the new subcomponents

---

Random photo of Atlantis


docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -111,10 +126,6 @@ be used in the
 [empty state](../?path=/story/components-selections-combobox-web--empty-state)
 of the Combobox, as well as the noun within the search placeholder.
 
-Follow InputText content guidelines for the text input.
-
-Follow Button content guidelines for the actions.
-
```
taylorvnoj, line 117: This seemed redundant to me, considering we are linking to the product vocabulary page just a few sentences before this.

docs/components/Combobox/Web.stories.tsx
```diff
@@ -265,20 +252,17 @@ const ComboboxMultiSelection: ComponentStory<typeof Combobox> = args => {
...
+CustomActivator.args = {};
+
 export const EmptyState = ComboboxEmptyState.bind({});
 EmptyState.args = {};
 
 export const MultiSelect = ComboboxMultiSelection.bind({});
 MultiSelect.args = {};
+
+export const SingleSelect = ComboboxSingleSelection.bind({});
+SingleSelect.args = {};
```
taylorvnoj, line 268: This and a few other changes are just be organizing the order of the stories and templates in this file

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -28,19 +28,37 @@ Combobox is composed of three main parts:
...
+
+The custom activator should be used when the Combobox trigger is a
+[Button](../?path=/docs/components-actions-button-docs--page) or an alternate
+variation of Chip. Button or Chip should be a child of `<Combobox.Activator />`
+and will be rendered in place of the default trigger. Examples of using a custom
+activator can be found in the
+[custom activator](../?path=/code/components-selections-combobox-web--custom-activator)
+story.
 
 #### Content
```
taylorvnoj, line 55: Should this still be called Content?  I think so - Combobox still has a content/popover and I think we can continue to reference it like that even though `Combobox.Content` is no longer used by the consumer.

taylorvnoj:
@ZakaryH can I get your eyes to double check that I included all the changes made? I feel like other areas of this documentation are possibly out of date?

docs/components/Combobox/Combobox.stories.mdx
ZakaryH: yeah agreed. I need to read the rest of the actual text content under this heading, but referring to the stuff that pops open as "Content" is still reasonable even if we're no longer exposing that literal component

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -28,19 +28,37 @@ Combobox is composed of three main parts:
...
-user experience requirements.
-
-There are two options for trigger elements of the Combobox,
-[TriggerButton](../?path=/story/components-selections-combobox-web--trigger-button)
+initiate option selection. When implementing Combobox, you may rely on the
+default trigger or, you may pass in a custom activator.
+
+The default trigger is a
+[Chip](../?path=/story/components-selections-chip-web--subtle) containing a
+label that reads "Select" and an `add` Icon suffix. The label can be customized
```
ZakaryH, line 36: I wonder if we could clarify things with some more specificity

it's actually a Chip `heading` that gets the "Select" value. I know that's a bit pedantic, but since Chip literally does have a `label` prop it might be confusing to be using these terms interchangeably

I get that we call it label because the prop to the `Combobox` is `label`... but maybe we change that? 

also we don't do anything with `label` if they have a custom activator and we don't really communicate the fact that it only shows up if you use the default ComboboxTrigger. some naming or prop documentation might be able to make that clearer

-------

**aside**

the fact that `label` can be an empty string still doesn't sit right with me. it results in this DOM

image

where we are rendering an empty `span`

the reason it happens is because of this `label={hasSelection ? selectedLabel : ""}`

and we have to do that because Chip has `label` as required so we're skirting around that by using an empty string. feels like we're misusing Chip here, either that or Chip shouldn't be this way. maybe instead Chip says I need at least a `heading` or a `label` - I don't care which but it can't be neither

sorry this has gone off the rails, but I do feel like it's worth considering now while we have both this and Chip in progress!

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -28,19 +28,37 @@ Combobox is composed of three main parts:
...
+by adding the optional `label` prop to better reflect the type of data being
```
ZakaryH, line 37: I mentioned it above but either via sub heading in this section, or adding some more context let's emphasize that the optional `label` prop to `Combobox` only affects the default trigger

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -28,19 +28,37 @@ Combobox is composed of three main parts:
...
+selected. Examples of using a default trigger can be found in the
+[single-select](../?path=/code/components-selections-combobox-web--single-select)
 and
-[TriggerChip](../?path=/story/components-selections-combobox-web--trigger-chip).
-You must have one trigger component.
+[multi-select](../?path=/code/components-selections-combobox-web--multi-select)
+stories.
+
+Notice that omitting a trigger subcomponent will result in the default trigger
```
ZakaryH, line 44: nitpick: Trigger was our old wording, now it's the Activator which might be a better word to use here

though I do get that we're under a heading called `Trigger` and discussing that topic, so maybe it's fine?

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -28,19 +28,37 @@ Combobox is composed of three main parts:
...
+story.
 
 #### Content
 
-This component is responsible for displaying the list of selectable options,
-handling user interactions, and managing search functionality.
+The content of the Combobox is responsible for displaying the list of selectable
+options, handling user interactions, and managing search functionality.
+
+`<Combobox.Option />` is a wrapper component that should be used to wrap each
```
ZakaryH, line 60: I'd change the wording here since at this point in time, `Combobox.Option` does not support children (lol what a sentence)

all of its functionality comes solely from its props, and we don't need people to wrap anything with the `Option`

one day when we support headers or custom format that might change, but let's wait til we cross that bridge!

ZakaryH:
it's a nitpick and not even related to the changes here but I noticed we have

`Single-select is ComboBox's` where we have an uppercase B vs lowercase everywhere else

if we're already in here adjusting things, let's swap that!

ZakaryH:
again, not from these changes but overall

>When clicking "Select all", all options will be selected.

let's either call out that it selects all filtered options here, or add another bit that mentions this behavior

ZakaryH:
hmm the Selection Callbacks section just needs to be re-written. most of that isn't true anymore

I'm happy to do that, or I can talk you through the changes. lmk which you'd prefer!

ZakaryH:
as I read teh `subjectNoun` stuff I'm wondering about all the props we have now around text strings

we've got

- label
- subjectNoun
- searchPlaceholder

is that getting to be a bit much? do we really need to allow all these to be customized?

oh nice. we don't even need `searchPlaceholder` anymore. it does nothing since it has been replaced with `subjectNoun`. so yeah we can remove that and then it's just two concepts which have very distinct purposes

docs/components/Combobox/Combobox.stories.mdx
taylorvnoj: I always disliked the word trigger anyway, so everything is changed to say activator. Technically the `ComboboxTrigger` does still exist as the default but that word won't surface to the user.

docs/components/Combobox/Combobox.stories.mdx
taylorvnoj: Ok - I created subheadings here: a027f48c8f8ec63183f53a816e2d9d30e70e126d
I made some of the verbiage around `label` more clear

docs/components/Combobox/Combobox.stories.mdx
taylorvnoj: I spoke with Scott briefly this morning about all this.
What do we think about changing `label` in Combobox to be called something entirely different.  

Maybe I'm making way for a use case that will never come up but what if down the line, Combobox's default trigger is no longer a `Chip`.  If it's a `Button`, `label` still does make sense.  If we rename Combobox's `label` prop to something entirely different and specific to Combobox, does that avoid confusion between Combobox props and Chip props?

docs/components/Combobox/Combobox.stories.mdx
taylorvnoj: Ok I kept it very surface level for now, but changed this sentence here e6db88b642dc292eb549e6a52b3de510d8a599ca

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -25,22 +25,50 @@ adding a new item to the list of options or managing a selection.
...
+_Note:_
+
+- The optional `label` prop on Combobox will only affect the default activator.
+  A custom activator label will overwrite the Combobox label.
+- Omitting a custom activator subcomponent will result in the default activator
+  being used.
+
+##### Custom activator
+
+The custom activator should be used when the Combobox activator is a
```
chris-at-jobber, line 56: Similar to above, I know we have some actual restrictions on this so we might have to be a bit more verbose, but can we embed an example so we can do more showing, less telling?

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -25,22 +25,50 @@ adding a new item to the list of options or managing a selection.
...
-user experience requirements.
+The activator is the user-facing interface to open and close the Combobox and
+initiate option selection. When implementing Combobox, you may rely on the
+default activator or, you may pass in a custom activator.
 
-There are two options for trigger elements of the Combobox,
-[TriggerButton](../?path=/story/components-selections-combobox-web--trigger-button)
+##### Default activator
+
+The default activator is a
```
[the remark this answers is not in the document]
taylorvnoj: This is totally fair however putting `Combobox` inside a Canvas tag and getting it to render is surprisingly difficult, at least for me right now. I can try with a different set of eyes tomorrow.  
That being said, we were also trying to steer clear of too many live examples within the docs page (because otherwise why have examples as well).  This is getting into a larger conversation about what our docs should look like and we have yet to return to that conversation (me + you + survey results).
So if I can get a Canvas in there relatively quickly tomorrow I will - if not I'll put this into the backlog because I do agree, having a live example here makes sense.

docs/components/Combobox/Combobox.stories.mdx
taylorvnoj: Ok I actually changed the wording in these sections completely, to not reference Chip or Button and just keep it high level enough that if the default activator changed, it wouldn't break our docs.

Your points about the prop name `label` and what Chip is doing is valid, lets bring that convo out of this PR

taylorvnoj:
I deprecated `searchPlaceholder` here: d1c1c87faeb28a1370ae07148b59ca2fc939c635 - it is actually being used outside of Atlantis (in `ClientsIndex`) so I didn't fully remove it yet.

In an effort to get this closed so we can re-focus on our other priority, here are some takeaways:
- think more about `label` prop name (chat with Zak and decide steps forward)
- `label` can = `" "` which isn't great (chat with Zak and decide steps forward)
- try to get `Combobox` examples in a `<Canvas>` to show default v. custom activator (I have a ticket for this; JOB-83185)

@ZakaryH if you could pretty please write the `Selection Callbacks` section for me, I would forever owe you.  I feel like it would be much faster if you rewrote that part

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -25,22 +25,50 @@ adding a new item to the list of options or managing a selection.
...
 Combobox is composed of three main parts:
 
-#### Trigger
+#### Activator
 
-The trigger is the user-facing interface to open and close the Combobox and
-initiate option selection. The choice between them depends on the design and
-user experience requirements.
+The activator is the user-facing interface to open and close the Combobox and
+initiate option selection. When implementing Combobox, you may rely on the
```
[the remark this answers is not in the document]
ZakaryH: >why else would I be here? 😄

because reading is fun and knowledge is power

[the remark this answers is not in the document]
scotttjob: image

ZakaryH:
>I deprecated searchPlaceholder here: https://github.com/GetJobber/atlantis/commit/d1c1c87faeb28a1370ae07148b59ca2fc939c635 - it is actually being used outside of Atlantis (in ClientsIndex) so I didn't fully remove it yet.

@taylorvnoj where are you seeing that?

I do see a result for `searchPlaceholder` in `ClientsIndex` but that's a happy coincidence that the variable name matches. they probably used it because they liked our name so much!

```
          <DataList.Search
            placeholder={labels.searchPlaceholder}
            onSearch={setSearchTerm}
            initialValue={initialSearchTermRef.current}
          />
```

as for the selection callbacks section, yeah no worries I'll type that up and commit it here

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -62,17 +84,17 @@ takes an explicit action to close it such as clicking outside the Combobox or
 pressing the ESC key.
 
 When `multiSelect` is enabled, a Combobox header will be displayed below the
-search input.
+search input. When no selections have been made, the header label will read
+"Select" and the button will read "Select all". When one or more options have
+been selected, the header label will show the number of selected items, like a
+counter. The button will then read "Clear".
```
ZakaryH, line 90: similar to other feedback, I wonder if we can be less specific and leverage the examples to show off what it looks like exactly. that way if we change the words, nothing or at least less, needs to be updated in the docs

so perhaps instead of describing each button state's wording, we say

".... a Combobox header will be displayed below the search input. This header provides details about the selection count, and a button to either select all or deselect all."

though to really get the most out of this abstracted approach we'd probably want to remove the specific references to the wording in the following paragraphs. I don't mind if we leave it like it.

taylorvnoj: updated here 2c68e306ab2157618230bac7788580542b4848d2

I found a way to be less specific while still explaining the behaviour.

As far as examples, Combobox wasn't playing nicely in a Canvas so I've put that ticket in the backlog. For this section, I don't believe we need an example on the docs page (at least in this current structure of the docs) since we have a Multi Select story.

taylorvnoj:
> > I deprecated searchPlaceholder here: [d1c1c87](https://github.com/GetJobber/atlantis/commit/d1c1c87faeb28a1370ae07148b59ca2fc939c635) - it is actually being used outside of Atlantis (in ClientsIndex) so I didn't fully remove it yet.
> 
> @taylorvnoj where are you seeing that?
> 
> I do see a result for `searchPlaceholder` in `ClientsIndex` but that's a happy coincidence that the variable name matches. they probably used it because they liked our name so much!
> 
> ```
>           <DataList.Search
>             placeholder={labels.searchPlaceholder}
>             onSearch={setSearchTerm}
>             initialValue={initialSearchTermRef.current}
>           />
> ```
> 
> as for the selection callbacks section, yeah no worries I'll type that up and commit it here

Oops you're right, I didn't look too deeply at the JO reference.  I deleted those props completely.

Thank you so so much for updating the other section!

ZakaryH:
so this LGTM but I did just make changes myself so I feel weird about approving it

at this point only @chris-at-jobber has not actively committed to this branch. any chance I could bother you for an approval?

chris-at-jobber (approved):
I'm good with the docs, probably unqualified to approve the code changes, but I trust that you two are both extremely qualified to review those 😄

PR #1623, opened 2023-11-01
feat: Hackathon

taylorvnoj:
# HACKATHON - DO NOT MERGE


scotttjob:
We did some great work! We had a lot of fun! Unfortunately we didn't win, which means this code isn't for merging. :(

We're going to close the PR, but we'll leave the branch alive just in case we ever want to revisit our incredible work here. 

Great job everyone!

PR #1625, opened 2023-11-03
fix(components): remove useragent styling from Combobox search [JOB-82086]

chris-at-jobber:
## Motivations

Atlantis has resets in the docs site that strip out styling on `input type="search"`, but all sites do not have this.

So, Combobox looks like this in Jobber:

image

## Changes

Add styling to remove browser defaults

### Added

- New styling rules for search input
- new styling rules for dismiss button as the `cross` icon could be moved off-center by external styling

### Fixed

- Unwanted browser-provided styling that conflicted with our own

## Testing

Pull pre-release into Jobber

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.css
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.css
@@ -24,6 +24,7 @@
 }
 
 .clearSearch {
+  display: flex;
   position: absolute;
   top: 50%;
   right: var(--space-base);
@@ -36,6 +37,8 @@
   background-color: var(--color-surface--background);
   cursor: pointer;
   transform: translateY(-50%);
+  justify-content: center;
+  align-items: center;
 }
 
 .clearSearch:focus {
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.css
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.css
@@ -12,6 +12,17 @@
   font-size: var(--typography--fontSize-base);
 }
 
+.searchInput::-webkit-search-decoration,
+.searchInput::-webkit-search-cancel-button,
+.searchInput::-webkit-search-results-button,
+.searchInput::-webkit-search-results-decoration {
+  -webkit-appearance: none;
+}
+
+.searchInput:focus {
+  outline: none;
+}
+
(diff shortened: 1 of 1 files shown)


chris-at-jobber:
For discussion: is this the best way to clear user agent styling? I'm not certain, but it seems to work alright. Open to other solutions!

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.css
```diff
@@ -12,7 +12,19 @@
   font-size: var(--typography--fontSize-base);
 }
 
+.searchInput::-webkit-search-decoration,
+.searchInput::-webkit-search-cancel-button,
+.searchInput::-webkit-search-results-button,
+.searchInput::-webkit-search-results-decoration {
+  -webkit-appearance: none;
```
chris-at-jobber, line 19: It looks like Storybook just uses `appearance: none` but this [solution I found](https://stackoverflow.com/questions/9421551/how-do-i-remove-all-default-webkit-search-field-styling) uses the webkit approach - should I use both? Will our post-processors add any webkit prefixes that are needed?

taylorvnoj:
I haven't put this into JO yet but I noticed that on the code tab within Storybook, I can see this happening 🤔 

Screenshot 2023-11-06 at 5 39 42 PM

chris-at-jobber:
@taylorvnoj after a great deal of personal frustration... I _think_ it's because the Code panel pulls from `@jobber/components` and not the local working copy. To test, I changed the background of the input to green and the bottom border to 28px red, and neither appeared in Code, but did in Canvas.

To verify, I did pull the pre-release in Jobber:

image

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.css
scotttjob: Further down the page on your link they also suggest an `appearance` approach in 2023, but I did want to try your solution in all three browsers and it looks like it's working just fine as-is.

I think what might be making it 'work' better though is your `outline:none` below which in other circumstances would _technically_ interfere with accessibility, but we do our own `focus` state so we can safely disable the built-in browser solution without affecting anything negatively. 

Even React themselves supports that decision:
image

scotttjob (approved):
LGTM! Does what it says it does (I spun up a 'clean' project to double check outside of storybook). We `might` not need the webkit specific stuff with the outline:0 but I'm fine with doubling up to cover all the bases.

PR #1626, opened 2023-11-06
feat(components): Optional DataList action

darryltec:
## Motivations

The DataList was initially built to have all of the actions or none of it. This enables a way to hide some actions when it's not valid. For example, you shouldn't have an archive action on an already archived item.

Since DataList actions gets built outside of the template, the `visible` prop needs to be function that has the item it's attached to as the parameter so the consumers can easily determine if it should should up based on the existing data.

### Added

- `visible` prop on DataList item actions

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/components/DataList/Web.stories.tsx
+++ b/docs/components/DataList/Web.stories.tsx
@@ -162,6 +162,7 @@ const Template: ComponentStory<typeof DataList> = args => {
 
       <DataList.ItemActions onClick={handleActionClick}>
         <DataList.ItemAction
+          visible={item => item.species !== "Droid"}
           icon="edit"
           label="Edit"
           onClick={handleActionClick}
--- a/packages/components/src/DataList/DataList.types.ts
+++ b/packages/components/src/DataList/DataList.types.ts
@@ -315,7 +315,7 @@ export interface DataListActionsProps<T extends DataListObject> {
 }
 
 export interface DataListBulkActionProps
-  extends DataListActionProps<DataListObject> {
+  extends Omit<DataListActionProps<DataListObject>, "visible"> {
   /**
    * The callback function when the action is clicked.
    */
--- a/packages/components/src/DataList/DataList.types.ts
+++ b/packages/components/src/DataList/DataList.types.ts
@@ -290,6 +290,11 @@ export interface DataListActionProps<T extends DataListObject> {
   readonly destructive?: boolean;
 
   /**
+   * Determine if the action is visible for a given item.
+   */
+  readonly visible?: (item: T) => boolean;
+
+  /**
    * The callback function when the action is clicked.
    */
   readonly onClick?: (data: T) => void;
--- a/packages/components/src/DataList/components/DataListActions/DataListActions.tsx
+++ b/packages/components/src/DataList/components/DataListActions/DataListActions.tsx
@@ -27,6 +27,7 @@ export function DataListActions<T extends DataListObject>({
   return (
     <DataListOverflowFade>
       {exposedActions.map(({ props }) => {
+        if (props.visible && !props.visible(activeItem)) return null;
         if (!props.icon) return null;
 
         return (
--- a/docs/components/DataList/DataList.stories.mdx
+++ b/docs/components/DataList/DataList.stories.mdx
@@ -104,6 +104,18 @@ action that is related to the item.
 </DataList>
 ```
 
+##### Hiding an action
+
+Hide an action by using the `visible` prop. It accepts a function that has a
+parameter of the item it's related to.
+
+```tsx
+<DataList.ItemAction
+  visible={item => item.status === "Archived"}
+  label="Delete"
+/>
+```
+
 <Tabs>
   <Tab label="DataList.ItemActions">
     <ArgsTable of={DataList.ItemActions} />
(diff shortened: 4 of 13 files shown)


packages/components/src/DataList/components/DataListActions/DataListActions.tsx
```diff
@@ -27,6 +27,7 @@ export function DataListActions<T extends DataListObject>({
   return (
     <DataListOverflowFade>
       {exposedActions.map(({ props }) => {
+        if (props.visible && !props.visible(activeItem)) return null;
```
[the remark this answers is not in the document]
darryltec: > Mean you could get rid of the check for props.visible existing?

yeah, wrote it that way first but I also just want it to intentionally always show up unless otherwise stated not to.

> Also I think if icon is required [here](https://github.com/GetJobber/atlantis/pull/1626/files#diff-e48b68917ca593858e13de32ffc2f9ed3557c7a8572ae29e71fc120da08f9a7cR285) in the action prop definition the if statement below could be removed needed.

That's more of a TS precautionary shit. Either I start casting or just write it like it's not there as TS thinks it's still not there.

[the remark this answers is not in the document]
MichaelParadis: Ahh fair. TS can be weird sometimes so that makes a ton of sense

packages/components/src/DataList/DataList.types.ts
```diff
@@ -310,7 +315,7 @@ export interface DataListActionsProps<T extends DataListObject> {
 }
 
 export interface DataListBulkActionProps
-  extends DataListActionProps<DataListObject> {
+  extends Omit<DataListActionProps<DataListObject>, "visible"> {
```
MichaelParadis, line 318: Not sure if this should be a future thing to address or deal with it here. 
I wonder how it would handle bulk action when the individual action is not visible. For example bulk archiving things that include an already archived items. 
Would we want to filter out the items that should be filtered or have the consumer deal with it?

darryltec: we'd most likely tell them that we've only archived 20 out of the 21 since the 1 is already archived. It would be weird to start removing actions because of the things you've selected because you'd eventually have no actions at that point

docs/components/DataList/Web.stories.tsx
```diff
@@ -162,6 +162,7 @@ const Template: ComponentStory<typeof DataList> = args => {
 
       <DataList.ItemActions onClick={handleActionClick}>
         <DataList.ItemAction
+          visible={item => item.species !== "Droid"}
```
MichaelParadis, line 165: Something I noticed while testing. If all the ItemActions are not visible it still shows the more icon that opens an empty menu. In this case all ItemActions had this visible prop added

https://github.com/GetJobber/atlantis/assets/17303106/f5035f83-cb93-4114-a2ab-51838dc3cd8d

darryltec: oh right! gooood catch

packages/components/src/DataList/DataList.types.ts
MichaelParadis: Sounds good

packages/components/src/DataList/DataList.types.ts
```diff
@@ -289,6 +289,11 @@ export interface DataListActionProps<T extends DataListObject> {
    */
   readonly destructive?: boolean;
 
+  /**
+   * Determine if the action is visible for a given item.
+   */
+  readonly visible?: (item: T) => boolean;
```
MichaelParadis, line 295: What would are thoughts be on allowing the row action to not be visible for specific rows? I think it would be nice but I think there may be some UX considerations.

packages/components/src/DataList/DataList.types.ts
MichaelParadis: I think maybe add something to the documentation that would specify the consumer dealing with the items when handling the bulk action that they may not want to be visible for that item

packages/components/src/DataList/DataList.types.ts
darryltec: There are a lot of factors in hiding an action on a single item/row. In Jobber's sense, we have statuses. You can archive pretty much any state except for something that is already archived. So you would want an archive action for anything that isn't in an archived state.

MichaelParadis: I was mainly thinking about hiding [this](https://github.com/GetJobber/atlantis/blob/f07dac102920d9a611f43dbc89c340dacc2c00fc/packages/components/src/DataList/DataList.types.ts#L232)
So it doesn't have the cursor if the onClick wouldn't do anything.

As I think more I don't really think it has value with what we need the DataList to do right now. I think this is something to consider in the future _if_ teams ask for it and it makes sense

darryltec: OH, like the whole thing. With the amount of actions we're thinking of doing, I don't think we need that yet

MichaelParadis: Agreed

MichaelParadis (changes requested):
Only change is the bug fix. Not sure if you had a chance to work on it yesterday but feel free to re-request once it is fixed

packages/components/src/DataList/DataList.types.ts
darryltec: yeah, i'll hold off on that until we solidify how we actually do batch

docs/components/DataList/Web.stories.tsx
darryltec: It's good now!

MichaelParadis (approved):
LGTM!

PR #1627, opened 2023-11-07
fix(components): Fix Chip gradient on Safari

rodrigoeidelvein:
## Motivations

While doing some tests on Safari, I noticed that the gradient on Chip was always showing up, even if the label was fully visible. To fix it, I added one pixel to the ref element that is being passed to the hook `useInView` to it can understand the size of the element to handle the intersections

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Fixed the gradient of Chip on Safari

### Security

- 

### Before
image

### After
image

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chip/Chip.css
+++ b/packages/components/src/Chip/Chip.css
@@ -174,3 +174,8 @@
     rgba(255, 255, 255, 0)
   );
 }
+
+.chipLabelRef {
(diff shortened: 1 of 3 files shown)


darryltec (approved):
can confirm, works!

PR #1630, opened 2023-11-09
fix: remove @use-it/event-listener dependency and use regular event handler attachment api.

timoteialbu:
## Motivations

During the vite migration work we noticed that there was some errors coming from globals that weren't being set. After doing some digging into this we realized that this was related to an issue with the `@use-it/event-listener` dependency.

## Changes

- The `@use-it/event-listener` dependency was unnecessary and causing issues. It's better to remove unnecessary dependencies.

### Removed
- We have removed the `@use-it/event-listener` dependency and have updated the code to use the regular event handler attachment api. We also updated the uses to use a shared hook to reduce unnecessary code duplication.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Autocomplete/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu.tsx
@@ -1,6 +1,5 @@
 import React, { RefObject, useEffect, useLayoutEffect, useState } from "react";
 import classnames from "classnames";
-import useEventListener from "@use-it/event-listener";
 import { createPortal } from "react-dom";
 import { usePopper } from "react-popper";
 import { AnyOption, Option } from "./Option";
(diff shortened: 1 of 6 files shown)


taylorvnoj (approved):
This all looks good to me. Tested `Menu` & `Autocomplete` in both Storybook and JO and everything works as it was before. Great job finding that fix Diane!

packages/components/src/Autocomplete/Menu.tsx
```diff
@@ -157,20 +162,24 @@ function useOnKeyDown(
   keyName: string,
   handler: (event: KeyboardEvent) => boolean | void,
 ) {
+  const newHandler = (event: KeyboardEvent): void => {
```
timoteialbu: should this be wrapped in a useCallback? I am slightly concerned about re-creating this all the time when the component that uses it re-renders.

PR #1633, opened 2023-11-13
fix(components): Remove transition from DataList's ListItem and Tags

rodrigoeidelvein:
## Motivations

There's a delay between the `background-color` animation of `listItem` on focus and tags gradient. I think the best thing is to remove the animation altogether to avoid this behaviour.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Remove transition from DataList's ListItem and Tags

### Security

- 

#### Before
DataList-before

#### After
Datalist-after

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -70,7 +70,6 @@
   position: relative;
   padding: var(--space-small);
   border-bottom: var(--border-base) solid var(--color-border);
-  transition: all var(--timing-base);
 }
 
 .listItem,
--- a/packages/components/src/DataList/components/DataListTags/DataListTags.css
+++ b/packages/components/src/DataList/components/DataListTags/DataListTags.css
@@ -23,5 +23,4 @@
     var(--overflow-bg) 100%
   );
   align-items: center;
-  transition: all var(--timing-base);
 }


chris-at-jobber:
Can we try adjusting the timing before we get rid of the transition altogether? If there's a delay on the transition being passed through to the tags, maybe something like `(--timing-fast)` for the tags?

chris-at-jobber:
Or, instead of using a solid gradient pseudo-element to create the illusion of masking, can we use a [gradient clip](https://stackoverflow.com/questions/9525215/using-css-can-you-apply-a-gradient-mask-to-fade-to-the-background-over-text)?

rodrigoeidelvein:
@chris-at-jobber I tried a lot of combinations and they all seem to be a bit unsynchronized. Using `var(--timing-quick)` on the tags feels closer, but it's still a little bit off. Or I might be seeing it wrong too 🤣 Do you want to pair later to troubleshoot it together?

chris-at-jobber:
Chatted "offline" but the branch `resolve-transition-flicker-on-datalist-tags` uses `mask-image` to resolve the flickering without dropping the transitions if you want to try that here, or I can put that up as a new PR!

chris-at-jobber:
Closing in favour of https://github.com/GetJobber/atlantis/pull/1637

PR #1636, opened 2023-11-15
fix(components): fix Chip, Combobox, and DataList focus states [JOB-82715]

chris-at-jobber:
## Motivations

Upon pulling the DataList and Combobox into Jobber, we found useragent styling on some of our custom-styled `button` and other HTML elements that was being masked by Storybook.

**Chip**

image

**Combobox**
image
image
image

**DataList**
image

It's possible there's a "systemic" fix to this like including a reset for HTML `button` elements in our Atlantis foundations stylesheet, but in service of solving the immediate problem this is my fix.

### Changed

- Changed existing focus background and shadow styling to apply to `focus-visible` contexts for a smarter treatment of focus states: keyboard users always get the full focus treatment, pointer users receive what the browser heuristics determine is appropriate.
- Outline has been removed in contexts where we are already adding a focus style

### Fixed

- useragent styling is replaced with Atlantis styling for focus states

## Testing

- pull prerelease into Jobber or a "sandbox" project

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chip/Chip.css
+++ b/packages/components/src/Chip/Chip.css
@@ -24,7 +24,12 @@
   align-items: center;
 }
 
-.chip:focus {
+.chip:focus,
+.chip:focus-visible {
+  outline: none;
+}
+
+.chip:focus-visible {
   box-shadow: var(--shadow-focus);
 }
 
@@ -48,11 +53,6 @@
   background-color: var(--color-surface);
   align-items: center;
   justify-content: center;
-}
-
-.chip .suffix:hover,
-.chip .prefix:hover {
-  background-color: var(--color-disabled);
 }
 
 .chip .prefix.empty,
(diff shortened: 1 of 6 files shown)


packages/components/src/Chip/Chip.css
```diff
@@ -50,11 +55,6 @@
   justify-content: center;
 }
 
-.chip .suffix:hover,
```
chris-at-jobber, line 53: Chip prefix and suffix do not currently have the ability to receive interactions and should not have a hover state.

chris-at-jobber:
In Jobber, you can verify that the focus styles are as intended, and observe the correct focus behaviour for both keyboard and pointer interactions.

https://github.com/GetJobber/atlantis/assets/39704901/8fdc769e-08fb-478a-97cf-9cfa57ddf454

packages/components/src/Chip/Chip.css
```diff
@@ -24,7 +24,12 @@
   align-items: center;
 }
 
-.chip:focus {
+.chip:focus,
+.chip:focus-visible {
+  outline: none;
```
chris-at-jobber, line 29: General note for this PR: I was wondering if just applying `:focus-visible` styling is enough, but in testing at a [basic level](https://codepen.io/murhaus/pen/KKJyJjQ) it _looks like_ there needs to be something telling the element not to receive a focus ring in the `:focus` state even if `:focus-visible` is declared.

chris-at-jobber: image

scotttjob (approved):
LGTM, does what it says it does on the tin.

chris-at-jobber:
Thanks @scotttjob !

PR #1637, opened 2023-11-15
fix(components): resolve transition flicker on DataList "tag count" element

chris-at-jobber:
## Motivations

When hovering the datalist row, the tag count's background gradient has a delay in receiving the new color value, so its' transition is not synced to the background of the row.

To resolve this, we'll get rid of the gradient background and apply a `mask-image` to the tags themselves.

### Changed

- tag container is visually truncated by a mask on the container instead of an element with a gradient being positioned above it
- modified DOM structure of the tags + count to add a wrapper and allow the count to sit above the masked container

### Fixed

- tag count does not flicker when the Datalist row is hovered

## Testing

- Hover on a clickable DataList with overflowing tags

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListTags/DataListTags.css
+++ b/packages/components/src/DataList/components/DataListTags/DataListTags.css
@@ -1,27 +1,27 @@
+.tagWrapper {
+  position: relative;
+}
+
 .tags {
   display: flex;
-  position: relative;
   z-index: var(--elevation-default);
   overflow: hidden;
   gap: var(--space-small);
   white-space: nowrap;
+  mask-image: linear-gradient(
+    to left,
+    transparent,
+    transparent var(--space-large),
+    rgba(0, 0, 0, 1) var(--space-largest),
+    rgba(0, 0, 0, 1)
+  );
 }
 
 .tagCount {
-  --overflow-bg: var(--data-list-item-active-color, var(--color-surface));
-
   display: flex;
   position: absolute;
   top: 0;
   right: 0;
   height: 100%;
-  padding-left: var(--space-base);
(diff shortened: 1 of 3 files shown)


packages/components/src/DataList/components/DataListTags/DataListTags.css
```diff
@@ -1,27 +1,27 @@
...
   display: flex;
-  position: relative;
   z-index: var(--elevation-default);
   overflow: hidden;
   gap: var(--space-small);
   white-space: nowrap;
+  mask-image: linear-gradient(
+    to left,
+    transparent,
+    transparent var(--space-large),
```
chris-at-jobber, line 14: Setting the stops using space values instead of the more common "%" because we want to ensure that the tags are fully clipped beneath the tag count. Same approach as the previous gradient with some adjustments.

packages/components/src/DataList/components/DataListTags/DataListTags.css
```diff
@@ -1,27 +1,27 @@
+.tagWrapper {
+  position: relative;
+}
+
 .tags {
   display: flex;
```
chris-at-jobber, line 6: position moves up to the wrapper because it was used to set the position of the tag count, which now lives outside of the `tags` div

rodrigoeidelvein (changes requested):
Veeery nice solution!

But we should only apply it tags that overflow with the count, for example:
image

before it was like this:
image

chris-at-jobber:
@rodrigoeidelvein that makes sense! What's the best way to apply conditional styling like that?

rodrigoeidelvein (approved):
LGTM!

PR #1642, opened 2023-11-16
fix: breakpoints and viewport mock

darryltec:
## Motivations

Related to https://github.com/GetJobber/atlantis/pull/1639 so to ensure breakpointsand mocking work, we need to update this

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


rodrigoeidelvein (approved):
LGTM!

Breakpoints are equivalent but using a different syntax and they work as expected. Tested it on the useBreakpoints page on Atlantis docs and also on CampaignsList where it is being used in several places! All working as expected and the same way as before.

And it also fixes the test that was breaking there.

PR #1646, opened 2023-11-20
fix(components): DataList filters gap

darryltec:
## Motivations

There was an issue where we're drilling down via `nth-child` to add a margin between items but if your child was in a `display: contents` it fails as it doesn't skip that element anymore but also invalidates other CSS properties gets added to it. To mitigate that we need to make the gap bits more resilient by using gap.

### Added

- added 1 more div to gap those actions out

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- alignment issue with search

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
+++ b/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
@@ -1,14 +1,22 @@
+.wrapper {
+  display: flex;
+  align-items: center;
+}
+
 .searchInput {
   --transition-properties: var(--timing-base) ease-in-out;
-  --button-offset: calc(var(--space-largest) - var(--space-smallest));
+  --button-offset: calc(var(--space-largest) - var(--space-smaller));
 
   position: absolute;
   /* Inputs are off by 1 when put beside a button */
-  top: calc(var(--space-small) + 1px);
+  top: 50%;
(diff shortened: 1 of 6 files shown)


packages/components/src/DataList/components/DataListSearch/DataListSearch.css
```diff
@@ -1,14 +1,22 @@
...
+  align-items: center;
+}
+
 .searchInput {
   --transition-properties: var(--timing-base) ease-in-out;
-  --button-offset: calc(var(--space-largest) - var(--space-smallest));
+  --button-offset: calc(var(--space-largest) - var(--space-smaller));
 
   position: absolute;
   /* Inputs are off by 1 when put beside a button */
```
chris-at-jobber, line 11: Non-blocking - can this comment go away now since we're clearing the "1px" adjustment?

chris-at-jobber (approved):
Nice cleanup, appreciate the fix on the small-screen search while you're in here 👌

packages/components/src/DataList/components/DataListSearch/DataListSearch.css
darryltec: Oh shoot, you right!

PR #1648, opened 2023-11-21
fix(components): fix Chip transition state

chris-at-jobber:
## Motivations

Chip was calling an easing token that didn't exist, so the intended transition was not being applied.

Why was a token that doesn't exist being used? Probably because we had docs for it - despite a search of the codebase returning no matches for _any_ easing tokens.

image

image

### Fixed

- Transition state of Chip works as expected

### Removed

- Documentation of tokens that don't seem to exist anywhere in Atlantis

## Testing

Hover or focus Chip

---

Random photo of Atlantis

### DIFF
--- a/packages/design/src/Animation.stories.mdx
+++ b/packages/design/src/Animation.stories.mdx
@@ -73,9 +73,3 @@ side nav, or accordion.
 ### Linear
 
 Use for color and opacity changes (things that don’t move).
-
-| Name          | Value                           |
-| :------------ | :------------------------------ |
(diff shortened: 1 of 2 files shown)


packages/design/src/Animation.stories.mdx
```diff
@@ -73,9 +73,3 @@ side nav, or accordion.
 ### Linear
 
 Use for color and opacity changes (things that don’t move).
-
```
chris-at-jobber, line 76: If you, dear reviewer, can find this token being defined somewhere in the codebase, by all means these docs can stay 😄

scotttjob (approved):
LGTM

PR #1649, opened 2023-11-21
fix(components-native): add padding horizontal on content overlay title

darryltec:
## Motivations

We were told that long titles go edge to edge and it shouldn't be. Found out that there's no padding on it. This puts that padding in there.

### Added

- Padding horizontal on ContentOverlay title
- Allow title to wrap and get bigger than the set height
- If no dismiss, center align that text so when it wraps it looks center aligned

#### Before

image

#### After

image

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/ContentOverlay/ContentOverlay.style.ts
+++ b/packages/components-native/src/ContentOverlay/ContentOverlay.style.ts
@@ -34,6 +34,7 @@ export const styles = StyleSheet.create({
     zIndex: tokens["elevation-base"],
     borderTopLeftRadius: modalBorderRadius,
     borderTopRightRadius: modalBorderRadius,
+    minHeight: tokens["space-extravagant"],
   },
 
   headerShadow: {
--- a/packages/components-native/src/ContentOverlay/ContentOverlay.style.ts
+++ b/packages/components-native/src/ContentOverlay/ContentOverlay.style.ts
@@ -58,7 +58,7 @@ export const styles = StyleSheet.create({
     flexDirection: "row",
     justifyContent: "center",
     paddingTop: tokens["space-base"],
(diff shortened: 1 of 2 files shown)


packages/components-native/src/ContentOverlay/ContentOverlay.style.ts
```diff
@@ -54,18 +54,18 @@ export const styles = StyleSheet.create({
...
     paddingTop: tokens["space-base"],
+    paddingHorizontal: tokens["space-base"],
   },
 
   titleWithDismiss: {
     flex: 1,
     justifyContent: "center",
     paddingLeft: tokens["space-base"],
     paddingRight: tokens["space-smaller"],
-    flexGrow: 1,
```
darryltec, line 69: `flex: 1` is already `flexGrow: 1` and `flexShrink: 1` adding one more is not only redundant but also makes the styling unpredictable as iOS seems to reset back the `flexShrink` to `0` if you specify `flexGrow`

packages/components-native/src/ContentOverlay/ContentOverlay.style.ts
```diff
@@ -34,7 +34,7 @@ export const styles = StyleSheet.create({
     zIndex: tokens["elevation-base"],
     borderTopLeftRadius: modalBorderRadius,
     borderTopRightRadius: modalBorderRadius,
-    height: tokens["space-extravagant"],
```
darryltec, line 37: NEVER ADD A HEIGHT! IF YOU NEED TO, USE MIN-HEIGHT!

rodrigoeidelvein (approved):
LGTM!

PR #1650, opened 2023-11-22
feat: DataList batch selection for select all with unknown IDs

darryltec:
## Motivations

Right now, the way the DataList would show you that you have selected all and the checkbox are checked, including the ones that aren’t loaded yet, is to give the DataList all of the ID’s of the selected items.

A new pattern emerged on Batch Action where if you select all, we don’t need to know/query for all the ID’s as that would get expensive the more data you have. Instead, we need to tell the DataList that everything on that list is selected except for the ones that aren’t. In doing so, we don’t need to query for all the ID’s and we would only be tracking the ones that are unselected.

### Added

- the selected prop now supports either an array of id's `[1, 2, 3]` or an object `{ totalCount: 100000, unselected: [] }`
  - instead of using the `totalCount` prop to use as "how many was selected" we're allowing the consumer to pass in any number they want. In theory, it will be the same number but, there are cases where there are more items now than it was before.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/DataList/__tests__/DataList.selectable.test.tsx
```diff
@@ -0,0 +1,91 @@
+import { screen, waitFor } from "@testing-library/react";
+import * as pom from "../__POM__/DataList.selectable.pom";
```
rodrigoeidelvein, line 2: I've never seen this, but it's really interesting. Is it something that we should consider using more often in our tests?

rodrigoeidelvein (approved):
Code overall looks good, there's only one behavior that I want to confirm that is the intended, maybe @chris-at-jobber can help here too

For example, in gmail, if I first single-select an item and then click on the "Select all" checkbox, it will first unselect everything. It will only selecting everything if there's nothing selected.
gmail_selection

In our DataList, it will select all even though I already have one selected
datalist_selection

This makes me question if we actually need the "Deselect all" button

rodrigoeidelvein:
Polaris' [Index Table](https://polaris.shopify.com/components/tables/index-table) behaves in the same way as DataList. Take that, Google! 

But they also don't have the "Deselect All" button

chris-at-jobber:
@rodrigoeidelvein Our checkboxes across Jobber web and mobile follow the pattern of

indeterminate > selected > unselected

so we should keep that consistent. As far as whether the "Deselect All" button is essential, I definitely see the point that the checkbox manages that, but the answer isn't essential to this PR. I will follow up on that though!

rodrigoeidelvein:
> @rodrigoeidelvein Our checkboxes across Jobber web and mobile follow the pattern of
> 
> indeterminate > selected > unselected
> 
> so we should keep that consistent. As far as whether the "Deselect All" button is essential, I definitely see the point that the checkbox manages that, but the answer isn't essential to this PR. I will follow up on that though!

Perfect! Thanks, @chris-at-jobber !

PR #1652, opened 2023-11-23
fix(components-native): avoid InputText re-renders

jessedoyle:
## Motivations

The InputText native component automatically trims whitespace from values.

Currently, the `onChangeText` callback is invoked when the input is blurred regardless of if the value actually changes. This can lead to additional renders, which cause further state-related issues down the line (i.e. when displaying contextual menus, etc.).

### Fixed

- The InputText `onChangeText` callback is no longer invoked on blur when the value does not change.

## Testing

1. Use the native `InputText` in a react component.

i.e.

```tsx
import React from "react";
import { InputText } from "@jobber/components-native";

const MyComponent = () => {
  const value = "Testing 🧪";
  
  const onChangeText = (text: string) => {
    console.log(`Text changed to: ${text}`);
  }
  
  return <InputText value={value} onChangeText={onChangeText} />;
};
```

2. Blur the input.

Previously the callback would be called because of whitespace trimming.

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/InputCurrency/InputCurrency.test.tsx
+++ b/packages/components-native/src/InputCurrency/InputCurrency.test.tsx
@@ -82,8 +82,6 @@ describe.each([{ includeATLContext: true }, { includeATLContext: false }])(
       });
 
       fireEvent.changeText(getByLabelText(placeHolder), `${value}`);
-      fireEvent(getByLabelText("Price"), "blur");
-      fireEvent(getByLabelText("Price"), "blur");
 
       await waitFor(() => {
         expect(getByDisplayValue("123.45912")).toBeDefined();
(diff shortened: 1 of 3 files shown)


packages/components-native/src/InputCurrency/InputCurrency.test.tsx
```diff
@@ -82,8 +82,6 @@ describe.each([{ includeATLContext: true }, { includeATLContext: false }])(
       });
 
       fireEvent.changeText(getByLabelText(placeHolder), `${value}`);
-      fireEvent(getByLabelText("Price"), "blur");
-      fireEvent(getByLabelText("Price"), "blur");
```
jessedoyle, line 86: Note to reviewer - this `blur` triggers [this callback](https://github.com/GetJobber/atlantis/blob/fc326c01fb124414c9a84198655d88476c40e7ae/packages/components-native/src/InputCurrency/InputCurrency.tsx#L193-L201). The onBlur callback references the `field` value from react-hook-form, though the test setup doesn't wrap the component in a form controller/provider. This will cause the input to render a display value of `0`.

Simple fix is to avoid triggering blur events in specs, but also happy to reconsider alternative solutions!

darryltec (approved):
issue seems to be still there on JM but this is a good update so i approve!

PR #1653, opened 2023-11-24
feat(components): Add stepped variation of progress bar

jjwlee94:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/ProgressBar/ProgressBar.css
+++ b/packages/components/src/ProgressBar/ProgressBar.css
@@ -2,7 +2,8 @@
   --progress-bar--height: var(--space-base);
 }
 
-progress.ProgressBar[value] {
+progress.ProgressBar[value],
+progress.SteppedProgressBar[value] {
   width: 100%;
   height: var(--progress-bar--height);
   border: none;
@@ -10,14 +11,24 @@ progress.ProgressBar[value] {
   appearance: none;
 }
 
+progress.SteppedProgressBar[value] {
+  margin-right: var(--space-small);
+}
+
 progress.ProgressBar[value],
-progress.ProgressBar[value]::-webkit-progress-bar {
+progress.ProgressBar[value]::-webkit-progress-bar,
+progress.SteppedProgressBar[value],
+progress.SteppedProgressBar[value]::-webkit-progress-bar {
   box-shadow: inset 0px 2px 4px rgba(0, 0, 0, 0.05);
   border-radius: calc(var(--progress-bar--height) / 2);
 }
 
 progress.ProgressBar[value]::-webkit-progress-bar {
   background-color: var(--color-surface--background);
+}
+
+progress.SteppedProgressBar[value]::-webkit-progress-bar {
+  background-color: var(--color-greyBlue--lightest);
 }
 
 progress.ProgressBar[value]::-webkit-progress-value {
@@ -30,4 +41,20 @@ progress.ProgressBar[value]::-moz-progress-bar {
   border-radius: calc(var(--progress-bar--height) / 2);
   background-color: var(--color-interactive);
   transition: all var(--timing-slow) ease-out;
+}
+
+progress.SteppedProgressBar[value]::-webkit-progress-value {
+  border-radius: calc(var(--progress-bar--height) / 2);
+  background-color: var(--color-interactive);
+  transition: all var(--timing-quick) ease-out;
+}
+
+progress.SteppedProgressBar[value]::-moz-progress-bar {
+  border-radius: calc(var(--progress-bar--height) / 2);
+  background-color: var(--color-interactive);
+}
+
+.SteppedProgressBar {
+  display: flex;
+  flex-direction: row;
 }
--- a/packages/components/src/ProgressBar/ProgressBar.test.tsx
+++ b/packages/components/src/ProgressBar/ProgressBar.test.tsx
@@ -16,4 +16,11 @@ describe("with props", () => {
     );
     expect(container).toMatchSnapshot();
   });
+
+  it("accept stepped variation", () => {
+    const { container } = render(
+      <ProgressBar currentStep={2} totalSteps={3} stepped />,
+    );
+    expect(container).toMatchSnapshot();
+  });
 });
--- a/packages/components/src/ProgressBar/ProgressBar.tsx
+++ b/packages/components/src/ProgressBar/ProgressBar.tsx
@@ -19,15 +19,53 @@ interface ProgressBarProps {
    * @default base
    */
   readonly size?: keyof typeof sizes;
+
+  /**
+   * Set the variation to the stepped progress bar
+   * @default false
+   */
+  readonly stepped?: boolean;
 }
 
 export function ProgressBar({
   currentStep,
   totalSteps,
   size = "base",
+  stepped = false,
 }: ProgressBarProps) {
   const percentage = (currentStep / totalSteps) * 100;
   const progressBarClassName = classnames(styles.ProgressBar, sizes[size]);
+  const steppedProgressBarClassName = classnames(
+    styles.SteppedProgressBar,
+    sizes[size],
+  );
+
+  function steppedVariation() {
+    return (
+      <div className={steppedProgressBarClassName}>
+        {Array.from({ length: totalSteps }).map((_, index) => {
+          const step = index + 1;
+          const value = step <= currentStep ? 100 : 0;
+
+          return (
+            <progress
+              key={step}
+              className={steppedProgressBarClassName}
+              max={100}
+              value={value}
+              aria-valuenow={value}
+              aria-valuemin={0}
+              aria-valuemax={100}
+            >
+              {value}%
+            </progress>
+          );
+        })}
+      </div>
+    );
+  }
+
+  if (stepped) return steppedVariation();
 
   return (
     <progress
--- a/packages/components/src/ProgressBar/ProgressBar.css.d.ts
+++ b/packages/components/src/ProgressBar/ProgressBar.css.d.ts
@@ -1,5 +1,6 @@
 declare const styles: {
   readonly "ProgressBar": string;
+  readonly "SteppedProgressBar": string;
 };
 export = styles;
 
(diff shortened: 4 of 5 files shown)


packages/components/src/ProgressBar/ProgressBar.css
```diff
@@ -20,6 +27,10 @@ progress.ProgressBar[value]::-webkit-progress-bar {
   background-color: var(--color-surface--background);
 }
 
+progress.SteppedProgressBar[value]::-webkit-progress-bar {
+  background-color: var(--color-greyBlue--lightest);
```
chris-at-jobber, line 31: Both variations should use the same background color, so we don't need this extra declaration

packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -19,15 +19,53 @@ interface ProgressBarProps {
    * @default base
    */
   readonly size?: keyof typeof sizes;
+
+  /**
+   * Set the variation to the stepped progress bar
+   * @default false
+   */
+  readonly stepped?: boolean;
```
chris-at-jobber, line 27: Do we want this to be boolean? Referencing @irmalcol comment [here](https://getjobber.slack.com/archives/C03BJHV3PMG/p1700666198355529?thread_ts=1698257310.303899&cid=C03BJHV3PMG), it sounds like we have a bit more flexibility with this property if we treat it as a string 

Would something like 

```
  @default progress
  variation?: progress | stepped; 
```

work here?

packages/components/src/ProgressBar/ProgressBar.css
jjwlee94: For context, we had adjusted the inactive state background colour to be a bit darker than what Atlantis version had because it was a bit harder to see in the stepped variation. If we want to maintain consistency between the two variations I can change this back!

`var(--color-surface--background)`:
image

`var(--color-greyBlue--lightest)`:
image

chris-at-jobber: Fair! There is a PR against our [design package](https://github.com/GetJobber/atlantis/pull/1545) that will update `surface--background` to map to `greyBlue--lightest` in the next few-ish days so let's use the existing design `surface--background` token and let it absorb the updates when that lands.

packages/components/src/ProgressBar/ProgressBar.tsx
jjwlee94: Oops, this slipped my mind. Agree that having this as a string would be this more flexible for potential additions in the future! Changed in [`bce3e13`](https://github.com/GetJobber/atlantis/pull/1653/commits/bce3e13933b1f7fe6343fa71eba7647e6a575b83).

packages/components/src/ProgressBar/ProgressBar.css
jjwlee94: Oh good to know! Sounds good :+1: Fixed in [`cb7b493`](https://github.com/GetJobber/atlantis/pull/1653/commits/cb7b493612510ae26de4f8e87bba85472c8c6e3b).

chris-at-jobber (commented):
Design looks great! Just gave it a test with VoiceOver and we'll need to make some adjustments - I walk through it in a video [here](https://us02web.zoom.us/clips/share/rraO6L3qxh-syPQMRQvgtMK1F0lWapK1aJSrfEUA78dpeXM-ESrAW2UR4Y7CuZE_0bm1m6XbxooPo5pJOdzJ6tFwtg.3cZzYT20pGSTNrFZ).

TLDR the current approach tells VoiceOver that there are 5 separate progress bars, all either 100% or 0% complete. This is a confusing way to convey the information, so we need to find a way to communicate the total progress state to the user in a single element.

In the example with 3/5 filled, we currently read:

"progress 100%, progress 100%, progress 100%, progress 0%, progress 0%"

How it should read:

"progress, 60%" OR "progress, 3 out of 5"

packages/components/src/ProgressBar/ProgressBar.tsx
```diff
@@ -19,15 +19,53 @@ interface ProgressBarProps {
...
+    return (
+      <div className={steppedProgressBarClassName}>
+        {Array.from({ length: totalSteps }).map((_, index) => {
+          const step = index + 1;
+          const value = step <= currentStep ? 100 : 0;
+          const ariaValue = currentStep / totalSteps;
+          const ariaLabel = `progress, ${step} out of ${totalSteps}`;
+
+          return (
+            <progress
```
darryltec, line 53: to chris' point about voiceover, this is gonna get read out separately and since this is a non-standard way of showing progress, i'd use a `progressbar` role https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role along with the other `aria`s to read out the correct information.

packages/components/src/ProgressBar/ProgressBar.test.tsx
```diff
@@ -16,4 +16,11 @@ describe("with props", () => {
     );
     expect(container).toMatchSnapshot();
   });
+
+  it("accept stepped variation", () => {
+    const { container } = render(
+      <ProgressBar currentStep={2} totalSteps={3} variation={"stepped"} />,
+    );
+    expect(container).toMatchSnapshot();
```
darryltec, line 24: let's avoid writing more snapshot tests here. I'd assert the specific things that the user would see.

- there should be 3 steps
- 2 out of the 3 are filled
- 1 out of 3 is not filled

packages/components/src/ProgressBar/ProgressBar.test.tsx
```diff
@@ -16,4 +16,25 @@ describe("with props", () => {
     );
     expect(container).toMatchSnapshot();
   });
+
+  describe("stepped variation", () => {
+    it("renders correctly", () => {
```
jjwlee94, line 21: I'm not sure if this is too generic; I could separate these out into separate assertions if that's preferred.

darryltec: yeah, should separate them out.  Basically, the description should tell you what it's testing. That helps in understanding the test better. And also help you write them haha.

jjwlee94: Applied in [`ecc624d`](https://github.com/GetJobber/atlantis/pull/1653/commits/ecc624d5423292ffdb6c1dc9346b313145e15492)!

darryltec (commented):
last few things then this solid!

chris-at-jobber:
Just wanted to pop in with a fun wrinkle that @Kristen-Haring and I came across when updating the Figma component today!

Our `base` size is 16px high, and the `small` size is 4px high, but the intended design of this new bar is 8px high.

Our proposed solution is
- add a new `size` variant named `smaller` that uses the 4px height
- update `small` to use 8px

An added bonus point is that it maps to the `spacing` [variable values](https://atlantis.getjobber.com/?path=/docs/design-spacing--page#values) as well (base = 16, small = 8, smaller = 4)

It would be a _visually_ breaking change, but looking at usages in Jobber, a relatively straightforward fix to update the 6-ish existing usages of `small` to `smaller`.

This might be better-suited as a follow-up PR but wanted to make sure it was captured. Apologies for this entering the discussion later in the game than we'd have liked!

chris-at-jobber:
The last item in the stepped bar has a margin-right that will throw off alignment:
image

Can we use `gap` on the wrapper instead of `margin-right` on the child elements to keep the spacing only where it needs to be?

Kristen-Haring:
Hellllllo - just checking out the PR and noticed that the spacing for progress bar when we get really wide (images above) looks too tight. Is there a way to indicate a percentage-based spacing vs straight pixels to give it more breathing room at wider implementations?

jjwlee94:
@chris-at-jobber I've adjusted size small to 8px in height and smaller to 4px in height as standardized. I wasn't sure what you meant by your comment below, so let me know if there's anything else required here!

> An added bonus point is that it maps to the spacing [variable values](https://atlantis.getjobber.com/?path=/docs/design-spacing--page#values) as well (base = 16, small = 8, smaller = 4)

I've also changed the styling to use `gap` on the wrapper instead of `margin-right`; what do we think of using `var(--space-base)` for the spacing? (Currently it's set to `var(--space-small)` @Kristen-Haring 

| Small | Base |
| ----- | ----- |
| image | image |
| image | image |

Also, what are your thoughts on a consistent transition speed between the two variations? The default variation uses `var(--timing-slow)` while the stepped variation uses `var(--timing-quick)`.

Kristen-Haring:
Thanks so much for the screens, @jjwlee94! After referring back to the current implementation on prod in our signup flow, I prefer the `var(--space-small)` so we can keep it as is if it has to be static.

The `quick` animation is my preference since that works well for the `stepped` variant - does it seem like a good clip for the base progress bar version? I trust you with that decision, Jeewon.

docs/components/ProgressBar/Web.stories.tsx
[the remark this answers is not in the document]
jjwlee94: Added in [`66d3356`](https://github.com/GetJobber/atlantis/pull/1653/commits/66d3356674f861ed8d73e944bdab8a66f92a6fec); let me know if what I have is :ok_hand:

packages/components/src/ProgressBar/ProgressBar.css
```diff
@@ -31,3 +39,19 @@ progress.ProgressBar[value]::-moz-progress-bar {
...
+progress.SteppedProgressBar[value]::-webkit-progress-value {
+  border-radius: calc(var(--progress-bar--height) / 2);
+  background-color: var(--color-interactive);
+  transition: all var(--timing-quick) ease-out;
+}
+
+progress.SteppedProgressBar[value]::-moz-progress-bar {
+  border-radius: calc(var(--progress-bar--height) / 2);
+  background-color: var(--color-interactive);
+}
```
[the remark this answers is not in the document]
jjwlee94: Cleaned up this file! We do have a difference in transition speed between the default vs. stepped variation (see [comment](https://github.com/GetJobber/atlantis/pull/1653#issuecomment-1828901849)); if your team has a preference on keeping this consistent, I can remove this.

darryltec (approved):
I am good with this! @chris-at-jobber is design good with this?

chris-at-jobber (commented):
Design approved! Thanks for handling that screen-reader functionality @jjwlee94 🙏

PR #1665, opened 2023-11-30
fix(components): multiselect preventing space bar from being recognized

darryltec:
## Motivations

Multiselect listens to spacebar, enter, up/down, and in doing so prevented it from being recognized globally. That is bad since the user won't be able to type spaces on inputs. [See here](https://codesandbox.io/p/sandbox/wispy-shape-ykfqnx?file=/Example.tsx)

### Added

- 

### Changed

- Instead of using the keydown hook that adds a global listener, we'll use the built-in react listener called `onKeyDown` which gives you exactly what you need at a more scoped down to that element only
- Removed the space bar listener on dropdown as that happens by default with checkboxes

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Add an input beside the multiselect
```diff
diff --git a/docs/components/MultiSelect/Web.stories.tsx b/docs/components/MultiSelect/Web.stories.tsx
index d9691558..74acf54b 100644
--- a/docs/components/MultiSelect/Web.stories.tsx
+++ b/docs/components/MultiSelect/Web.stories.tsx
@@ -21,7 +21,7 @@ const BasicTemplate: ComponentStory<typeof MultiSelect> = args => {
     { label: "Ignored", checked: true },
   ]);
   return (
-    <MultiSelect {...args} options={options} onOptionsChange={setOptions} />
+    <><input type="text"/><MultiSelect {...args} options={options} onOptionsChange={setOptions} /></>
   );
 };
```

- You should be able to type a space on the input
- focusing on the multiselect, you should be able to open it with space or enter; escape should close it
- you should be able to select an option through enter, or spacebar
- you should be able to move the focus on the next checkbox on up or down

---

Random photo of Atlantis


scotttjob (approved):
Tried it out, works good!

PR #1673, opened 2023-12-04
feat(components): expose search term on combobox action

darryltec:
## Motivations

There are instances where the action would need to
- Only show up when there's a search term
- Base the label via the search term
- Know the search term on click

This PR enables all of that.

### Added

- Added a `visible` prop
- Exposes the search term on actions

| Empty search term | With search term |
|--|--|
| image | image |

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.tsx
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.tsx
@@ -30,7 +30,7 @@ export function ComboboxAction(props: ComboboxActionProps) {
         <Typography
           element="span"
           size="base"
-          textColor="green"
+          textColor="interactive"
           fontWeight="semiBold"
         >
           {computedLabel}
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.tsx
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.tsx
@@ -4,18 +4,27 @@ import { Typography } from "../../../Typography";
 import { ComboboxActionProps } from "../../Combobox.types";
 import { ComboboxContext } from "../../ComboboxProvider";
 
-export function ComboboxAction(props: ComboboxActionProps): JSX.Element {
+export function ComboboxAction(props: ComboboxActionProps) {
   const { searchValue } = useContext(ComboboxContext);
+
+  if (props.visible) {
+    const isVisible =
+      typeof props.visible === "function" && !props.visible({ searchValue });
+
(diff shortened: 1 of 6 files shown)


packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.tsx
```diff
@@ -1,24 +1,39 @@
...
         className={styles.actionButton}
-        onClick={props.onClick}
+        onClick={e => props.onClick(e, options)}
         type="button"
-        aria-label={props.label}
       >
         <Typography
           element="span"
           size="base"
           textColor="green"
```
chris-at-jobber, line 33:
```suggestion
          textColor="interactive"
```

😉

taylorvnoj: It actually has to be `green`? `interactive` is on mobile Typography, not web @chris-at-jobber

chris-at-jobber: 🤯 OK separate task... "Allow Typography to consume semantic colors" 😄

packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.tsx
```diff
@@ -1,24 +1,39 @@
...
 
-export function ComboboxAction(props: ComboboxActionProps): JSX.Element {
   return (
     <div className={styles.actionContainer}>
       <button
         className={styles.actionButton}
-        onClick={props.onClick}
+        onClick={e => props.onClick(e, options)}
         type="button"
-        aria-label={props.label}
```
taylorvnoj, line 13: No more `aria-label`?

What used to show up - under ARIA Attributes in the inspector:
Screenshot 2023-12-04 at 5 03 21 PM

What we have now:
Screenshot 2023-12-04 at 5 03 43 PM

taylorvnoj (commented):
Ok I _love_ this!  Just a pondering about tests and a question about the aria-label

packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.tsx
darryltec: Aria's are needed for when it's different from the label (or lack of it). it looks like they're always the same so aria's aren't really needed. Forgot to add a comment on it haha

PR #1675, opened 2023-12-05
feature(hooks): Move useShowClear to @jobber/hooks

rodrigoeidelvein:
## Motivations

As we want to implement the same feature of clearing inputs on the web, we should turn this hook into a shared one and reuse the same logic on the web we already had on mobile.

It used to be a hook from `InputFieldWrapper`, then I updated the imports on `@jobber/components-native`.

I didn't want docs for it because it should be an internal hook, so I didn't expose it in our docs.

### Added

- 

### Changed

- `useShowClear` is now a shared hook.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


darryltec (approved):
This is an interesting one tho since it is under hooks for sharing between package and not for being a "this will be helpful for jobber in general" type deal. No objections on it tho and it's good that we don't document this for the public haha

PR #1677, opened 2023-12-06
feat(components): Add clearable prop to inputs

rodrigoeidelvein:
## Motivations

Adding the clearable prop to inputs, except for `InputDate,` because I hit a blocker where `react-datepicker` requires the event when changing the input value for some reason, and I can't mimic it when setting the value directly to an empty string.

I tried to follow the same names and conventions we have on `@jobber/components` name.

### Added

- Added the `clearable` prop to inputs, except for `InputDate`.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -9,13 +9,13 @@ import React, {
 } from "react";
 import { v1 as uuidv1 } from "uuid";
 import { Controller, useForm, useFormContext } from "react-hook-form";
+import { useShowClear } from "@jobber/hooks";
 import { FormFieldProps } from "./FormFieldTypes";
 import styles from "./FormField.css";
 import { FormFieldWrapper } from "./FormFieldWrapper";
 import { FormFieldPostFix } from "./FormFieldPostFix";
 
-// Added 13th statement to accommodate getErrorMessage function
-/*eslint max-statements: ["error", 13]*/
+// eslint-disable-next-line max-statements
 export function FormField(props: FormFieldProps) {
   const {
     actionsRef,
@@ -42,6 +42,7 @@ export function FormField(props: FormFieldProps) {
     onFocus,
     onBlur,
     onValidation,
+    clearable = "never",
   } = props;
 
   const {
@@ -56,6 +57,7 @@ export function FormField(props: FormFieldProps) {
 
   const [identifier] = useState(uuidv1());
   const [descriptionIdentifier] = useState(`descriptionUUID--${uuidv1()}`);
+  const [focused, setFocused] = useState(false);
   /**
    * Generate a name if one is not supplied, this is the name
    * that will be used for react-hook-form and not neccessarily
@@ -71,6 +73,8 @@ export function FormField(props: FormFieldProps) {
     }
   }, [value, watch(controlledName)]);
 
+  const fieldValue = watch(controlledName);
+
   useImperativeHandle(actionsRef, () => ({
     setValue: newValue => {
       setValue(controlledName, newValue, { shouldValidate: true });
@@ -80,6 +84,14 @@ export function FormField(props: FormFieldProps) {
   const message = errors[controlledName]?.message;
   const error = getErrorMessage();
   useEffect(() => handleValidation(), [error]);
+
+  const showClear = useShowClear({
+    clearable,
+    multiline: false,
+    focused,
+    hasValue: Boolean(fieldValue),
+    disabled,
+  });
 
   return (
     <Controller
@@ -122,6 +134,8 @@ export function FormField(props: FormFieldProps) {
             error={error}
             identifier={identifier}
             descriptionIdentifier={descriptionIdentifier}
+            showClear={showClear}
+            onClear={handleClear}
           >
             {renderField()}
           </FormFieldWrapper>
@@ -163,6 +177,12 @@ export function FormField(props: FormFieldProps) {
           }
         }
 
+        function handleClear() {
+          handleBlur();
+          setValue(controlledName, undefined, { shouldValidate: true });
+          onChange && onChange("");
+        }
+
         function handleChange(
           event: ChangeEvent<
             HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement
@@ -199,12 +219,15 @@ export function FormField(props: FormFieldProps) {
           if ((target as HTMLInputElement).select) {
             setTimeout(() => readonly && (target as HTMLInputElement).select());
           }
+
+          setFocused(true);
           onFocus && onFocus();
         }
 
         function handleBlur() {
           onBlur && onBlur();
           onControllerBlur();
+          setFocused(false);
         }
       }}
     />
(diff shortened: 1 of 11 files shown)


packages/components/src/FormField/components/ClearAction.css
```diff
@@ -0,0 +1,25 @@
...
+  padding: 0;
+  border: none;
+  border-radius: var(--radius-circle);
+  background-color: var(--color-surface--background);
+  cursor: pointer;
+  justify-content: center;
+  align-items: center;
+}
+
+.addedMargin {
```
chris-at-jobber, line 23: What do we need so much margin for?

packages/components/src/FormField/FormFieldTypes.ts
```diff
@@ -113,6 +114,15 @@ export interface CommonFormFieldProps {
    * Set the component to the given value.
    */
   readonly value?: string | number | Date;
+
+  /**
+   * Add a clear action on the input that clears the value.
+   *
+   * You should always use `while-editing` if you want the input to be
+   * clearable. if the input value isn't editable (i.e. `InputDateTime`) you can
```
chris-at-jobber, line 122: This comment is referencing a mobile-only component, but I guess the example is still valid?

packages/components/src/FormField/components/ClearAction.css
```diff
@@ -0,0 +1,25 @@
...
+  width: var(--space-large);
+  height: var(--space-large);
+  padding: 0;
+  border: none;
+  border-radius: var(--radius-circle);
+  background-color: var(--color-surface--background);
+  cursor: pointer;
+  justify-content: center;
+  align-items: center;
+}
```
chris-at-jobber, line 21: We need focus states here, can we write them using `:focus-visible` (see [ComboboxClearAction](https://github.com/GetJobber/atlantis/blob/979b53e34b1e16cbbb7de53184c1d9eb42dba8d7/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.css#L21-L33) for reference)

chris-at-jobber: Just need to make sure we're clearing the `outline` property and setting the `box-shadow: var(--shadow-focus)`

packages/components/src/FormField/components/ClearAction.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Icon } from "../../Icon";
+
+interface ClearActionProps {
+  /**
+   * Click handler
+   */
+  readonly onClick: () => void;
+}
+
+export function ClearAction({ onClick }: ClearActionProps): JSX.Element {
```
chris-at-jobber, line 12: As a followup to this PR it would be nice to swap this component into [ComboboxContentSearch](https://github.com/GetJobber/atlantis/blob/da2879c62e6740a1d37263b1ac37444f903e8a9e/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.tsx)

chris-at-jobber (dismissed):
This looks good and works well! Just two small changes on my end

image

- we need to add focus state (see Combobox example above)
- there's an [unused CSS class](https://github.com/GetJobber/atlantis/pull/1677/files#r1418101731) lying around

chris-at-jobber:
You can reference my [PR here](https://github.com/GetJobber/atlantis/pull/1679/files#diff-51eef1ce8accb5a864d38c1c20a27c9aca2aaa8accf7266e5106c5f10cbcfb41R44-R49) where I'm making Combobox clearing have a better focus state if you like!

packages/components/src/FormField/components/ClearAction.css
rodrigoeidelvein: oh yeah, I added it before because I thought it would be the clear button on the left side of the timer/number picker. but it's a good catch, it's not being used anymore.

packages/components/src/FormField/FormFieldTypes.ts
rodrigoeidelvein: Good point, I can change it to `InputTime`. Even though we can edit it, the timer picker is probably more used than typing on it

taylorvnoj:
I'm pretty sure we can't remove the clock icon right, because that is a browser thing? I think having the icon and clearable present at the same time is inevitable but wanted to double check...

Screenshot 2023-12-08 at 4 18 11 PM

chris-at-jobber:
I'm seeing very inconsistent behaviour around when the clearable button appears and how it interacts with the keyboard when I am set to "while-editing"

- Clear action seems to only appear after I perform an on-scroll while focusing the input. It should be present whenever a value is present and the input is focused

- When I `tab`, it removes the clear action instead of focusing it (when set to always, this is not a problem)

https://github.com/GetJobber/atlantis/assets/39704901/bfae9ba5-97a5-4600-ae23-fd8470e9ba85

packages/components/src/FormField/components/ClearAction.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <div className={styles.clearContainer}>
+      <button
+        className={styles.clearInput}
+        // Prevent the input from losing focus when the clear button is clicked
+        onMouseDown={event => event.preventDefault()}
```
taylorvnoj, line 18: I could be misunderstanding - but does this mean that once we mousedown on the clearable, the entire input should still be focused? After I interact with the clearable, no part of the component is in focus anymore.

Kapture 2023-12-08 at 17 01 00

rodrigoeidelvein: Good catch, I added a focus back to the input on `handleClear`

chris-at-jobber:
The behaviour still appears unpredictable:

https://github.com/GetJobber/atlantis/assets/39704901/b17d1cde-1f70-4dce-8374-73d85aba88aa

Sometimes the clear button appears, sometimes it hides, sometimes it appears to be triggered by a re-render, other times not

packages/components/src/FormField/components/ClearAction.tsx
taylorvnoj: Perfect - works really well now!

packages/components/src/FormField/FormField.tsx
```diff
@@ -14,8 +14,7 @@ import styles from "./FormField.css";
 import { FormFieldWrapper } from "./FormFieldWrapper";
 import { FormFieldPostFix } from "./FormFieldPostFix";
 
-// Added 13th statement to accommodate getErrorMessage function
-/*eslint max-statements: ["error", 13]*/
+// eslint-disable-next-line max-statements
```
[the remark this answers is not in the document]
rodrigoeidelvein: Good call, we don't need it anymore because I moved the logic to `InputFieldWrapper`. I can add it back

chris-at-jobber (commented):
Design approved on the `always` state!

PR #1678, opened 2023-12-06
fix(components): remove gradient on scroll from top and bottom of Combobox [JOB-79443]

tyler-vincent:
## Motivations

When scrolling to the top or bottom of a Combobox list the gradients used as a visual cue that the combo list can be inline scrolled doesn't disappear, which can be misleading and generally doesn't look good. 

#### Before 
beforefixedblur

#### After
fixedblur

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContentList.css
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContentList.css
@@ -18,34 +18,42 @@
   padding: var(--space-small) var(--space-base);
 }
 
+.container::before,
 .container::after {
   content: "";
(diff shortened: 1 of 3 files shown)


chris-at-jobber (commented):
Design approved! Thanks for the fix @tyler-vincent

scotttjob (approved):
LGTM! Pulled everything down, ran it locally everything works as you would expect.

The only nit I could possibly have is the useEffect block you wrote would work well as a custom hook, so it could exist in a reusable fashion but we'll do that extraction if we need it a second time. No need to prematurely optimize. 

Thanks for the contribution!

PR #1679, opened 2023-12-06
fix(components): Clean up interactive state colors in combobox

chris-at-jobber:
## Motivations

Combobox was using `surface--background` as proxy for our new hover color before that new hover color existed.

Now it just uses the new hover color!

image
image

### Changed

- using `hover` color for hover and focus states instead of `surface--background` color

### Fixed

- useragent styling leaking into combobox clear-search action

## Testing

View in Storybook

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.css
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.css
@@ -42,5 +42,9 @@
 }
 
 .clearSearch:focus {
+  outline: transparent;
+}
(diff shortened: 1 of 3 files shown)


packages/components/src/Combobox/components/ComboboxContent/ComboboxContentSearch/ComboboxContentSearch.css
```diff
@@ -42,5 +42,9 @@
 }
 
 .clearSearch:focus {
+  outline: transparent;
```
chris-at-jobber, line 45: Resolves the extra outline we get from the browser
image

PR #1682, opened 2023-12-08
fix(components): Always show sort icons on Datalist [JOB-82902]

tyler-vincent:
## Motivations

Working theory: not surfacing the sort arrows until the user has clicked a sortable heading is a hard way for the user to learn that they need to click that heading. It’s also hard to tell which columns are in fact sortable. This is also inconsistent with DataTable, which surfaces sort permanently.

This PR shows the sort arrows permanently for any column that can be sorted on Datalist. It also adds a subtle rollover effect to the sorting arrows.

## Changes

Screenshot 2023-12-08 at 2 09 18 PM

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
+++ b/packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
@@ -30,9 +30,11 @@ export function DataListHeaderTile<T extends DataListObject>({
       onClick={handleOnClick}
     >
       <Text maxLines="single">{headers[headerKey]}</Text>
-      {sortingState?.key === headerKey && (
+      {sortingState?.key === headerKey ? (
         <DataListSortingArrows order={sortingState.order} />
-      )}
+      ) : sortingState?.key !== headerKey && isSortable ? (
+        <DataListSortingArrows order="none" />
+      ) : null}
     </Tag>
   );
 
--- a/packages/components/src/DataList/components/DataListHeaderTile/DataListSortingArrows.css
+++ b/packages/components/src/DataList/components/DataListHeaderTile/DataListSortingArrows.css
@@ -16,3 +16,8 @@
 .sortIcon path.inactive {
   fill: var(--color-inactive--surface);
 }
+
+.sortIcon:hover path {
+  fill: var(--color-interactive--subtle);
+  transition: all var(--timing-base) ease;
+}
--- a/packages/components/src/DataList/DataList.test.tsx
+++ b/packages/components/src/DataList/DataList.test.tsx
@@ -13,11 +13,7 @@ import {
   DATA_LOAD_MORE_TEST_ID,
   EMPTY_FILTER_RESULTS_MESSAGE,
 } from "./DataList.const";
-import {
-  DataListItemType,
-  DataListProps,
-  DataListSorting,
-} from "./DataList.types";
+import { DataListItemType, DataListProps } from "./DataList.types";
 import { DATALIST_TOTALCOUNT_TEST_ID } from "./components/DataListTotalCount";
 import {
   DATALIST_LOADINGSTATE_ROW_TEST_ID,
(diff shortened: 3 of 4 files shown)


packages/components/src/DataList/components/DataListHeaderTile/DataListSortingArrows.css
```diff
@@ -16,3 +16,8 @@
 .sortIcon path.inactive {
   fill: var(--color-inactive--surface);
 }
+
+.sortIcon:hover path {
+  filter: brightness(85%);
```
chris-at-jobber, line 21: Let's use `var(--color-interactive--subtle)` here; our preference for hover interactions is to darken, and to use set values rather than filters/tints/opacities

chris-at-jobber (changes requested):
Looks great @tyler-vincent ! One suuuper small change and then design's good to go 😃

packages/components/src/DataList/components/DataListHeaderTile/DataListSortingArrows.css
chris-at-jobber: Also I lied, one other small change - can we match the alignment of the right-aligned sort icon to the column's contents? Reference [here](https://www.figma.com/file/TuoXq9l7Mg1Xfusj6qnx31/%5BJobber-2025-System%5D-List-Views?type=design&node-id=4514%3A63926&mode=design&t=m6hsxzxVr8mU9hRr-1)
image

chris-at-jobber: (If the alignment is proving finicky to resolve, the color change would be sufficient within the scope of this PR)

tyler-vincent: @scotttjob can attest I really tried to fix that alignment, but ultimately failed when the clock ran out. Ended up deciding to just get this in for now. The tip on `var(--color-interactive--subtle)` is appreciated, will push that change up soon.

packages/components/src/DataList/components/DataListHeaderTile/DataListSortingArrows.css
```diff
@@ -16,3 +16,8 @@
 .sortIcon path.inactive {
   fill: var(--color-inactive--surface);
 }
+
+.sortIcon:hover path {
+  filter: var(--color-interactive--subtle);
```
chris-at-jobber, line 21: This needs to be `fill:` instead of `filter:`, then we should be good to go!

chris-at-jobber:
```suggestion
  fill: var(--color-interactive--subtle);
```

chris-at-jobber (changes requested):
Very small one 🤏

packages/components/src/DataList/components/DataListHeaderTile/DataListSortingArrows.css
tyler-vincent: Ah shoot, good catch, totally missed fill vs filter. Fixed!

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -30,9 +30,11 @@ export function DataListHeaderTile<T extends DataListObject>({
       onClick={handleOnClick}
     >
       <Text maxLines="single">{headers[headerKey]}</Text>
-      {sortingState?.key === headerKey && (
+      {sortingState?.key === headerKey ? (
         <DataListSortingArrows order={sortingState.order} />
-      )}
+      ) : sortingState?.key !== headerKey && isSortable ? (
```
scotttjob, line 35: This **feels** like it could be neater, but I also understand what's happening just fine so I'm going to allow it. :)

scotttjob (approved):
Pulled this down, ran everything. LGTM! Does what it says it does! 

Looks great in action, and I definitely think this is a nice modification.

Thank you for the contribution! :)

PR #1687, opened 2023-12-13
fix(components): update combobox list to use overflow: auto rather than scroll

gregorijobber:
## Motivations

We're seeing some strange behaviors of the Combobox list trying to use scrollbars even we don't need them (as per image below)

image

### Added

- 

### Changed

- Update Combobox List styling to use `overflow: auto` rather than `overflow: scroll`

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.css
@@ -6,8 +6,7 @@
   margin-top: var(--space-small);
   border: var(--border-base) solid var(--color-border);
   border-radius: var(--radius-larger);
-  overflow: scroll;
-  overflow-y: auto;
+  overflow: auto;
   background: var(--color-surface);
 }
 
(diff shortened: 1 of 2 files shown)


scotttjob (commented):
From a coding perspective this fine to me. I ran the code and confirmed the scrollbars are hidden so all good there.

I haven't actually approved it though, as I'd like our designer to weigh in just to make sure there wasn't a design decision for the scrollbars to be present before we take them out (in which case we can probably move this option to a prop).

gregorijobber:
> From a coding perspective this fine to me. I ran the code and confirmed the scrollbars are hidden so all good there.
> 
> I haven't actually approved it though, as I'd like our designer to weigh in just to make sure there wasn't a design decision for the scrollbars to be present before we take them out (in which case we can probably move this option to a prop).

As this change will not hide the scrollbars if they're needed (keep in mind that I've added `auto` rather than `hidden`) I don't think we need to worry about this @scotttjob but yes let's wait for a design review! Check [this slack 🧵](https://getjobber.slack.com/archives/C03BJHV3PMG/p1702415065654259?thread_ts=1702414572.886209&cid=C03BJHV3PMG) too!

gregorijobber:
@chris-at-jobber we'll wait for your 👀 on this one! Thanks in advance!

gregorijobber:
Good catch @fagnanm! Fixed!

scotttjob (approved):
LGTM! Saw the design approval, and it looks/works well. Thanks for the contribution!

PR #1691, opened 2023-12-15
refactor(components): update checkbox styling with re-theme flag [JOB-84551]

chris-at-jobber:
## Motivations

Allow checkbox to be re-styled when a given class is present

### Changed

- Enable checkbox to be re-themed and create alternate styling for re-theme

## Testing

Apply the `jobber-retheme` class to the Storybook iframe `body`

- [ ] With `jobber-retheme`, should look different
- [ ] Without `jobber-retheme`, should look the same

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Checkbox/Checkbox.css
+++ b/packages/components/src/Checkbox/Checkbox.css
@@ -1,5 +1,10 @@
 .checkbox {
+  --checkbox-size: var(--space-base);
   display: inline-block;
+}
+
+:global(.jobber-retheme) .checkbox {
+  --checkbox-size: calc(var(--space-base) + var(--space-smaller));
 }
 
 .wrapper {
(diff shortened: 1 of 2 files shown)


packages/components/src/Checkbox/Checkbox.css
```diff
@@ -45,6 +50,15 @@
   border-color: var(--color-disabled--secondary);
 }
 
+:global(.jobber-retheme) .checkBox {
+  border-color: var(--color-border);
+  border-radius: calc(var(--radius-base) / 2);
+}
+
+:global(.jobber-retheme) .checkBox:hover {
```
fagnanm, line 58: @chris-at-jobber can we apply this to the label as well

packages/components/src/Checkbox/Checkbox.css
```diff
@@ -45,6 +50,15 @@
   border-color: var(--color-disabled--secondary);
 }
 
+:global(.jobber-retheme) .checkBox {
+  border-color: var(--color-border);
+  border-radius: calc(var(--radius-base) / 2);
```
fagnanm, line 55: @chris-at-jobber should we do this? Or just set to radius-small and then bump small up to 4 and base up to 8?

fagnanm (changes requested):
Mostly looks good, one question and an update to make so that the hover state applies to labels as well

packages/components/src/Checkbox/Checkbox.css
chris-at-jobber: Yeah that's cleaner, I'll do that

chris-at-jobber: Just realized I also need to update the design package to add a radius-small but we need to do that anyways so why not now 😄

chris-at-jobber:
@fagnanm radius and hover state updated

fagnanm (approved):
Changes look good. TIL about focus-visible!

PR #1693, opened 2023-12-16
refactor(components): update Button for retheme [JOB-84545]

chris-at-jobber:
## Motivations

Allow Button to be re-styled when a given class is present

### Changed

- Enable Button to be re-themed and create alternate styling for re-theme
- For both modes, set focus to be on `:focus-visible` to follow browser heuristics on when to show focus ring

## Testing

Apply the `jobber-retheme` class to the Storybook iframe `body`

- [ ] With `jobber-retheme`, should look different
- [ ] Without `jobber-retheme`, should look the same except for focus no longer persists after you've clicked the button with a pointer

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -134,6 +134,13 @@ a.button:active {
   fill: var(--button--color-variation--hover);
 }
 
+:global(.jobber-retheme) .secondary:hover,
+:global(.jobber-retheme) .subtle.primary:hover,
+:global(.jobber-retheme) .secondary:focus-visible,
+:global(.jobber-retheme) .subtle.primary:focus-visible {
+  border-color: var(--button--color-variation--hover);
+}
+
 .subtle.secondary {
   background-color: var(--button--color-surface);
 }
@@ -157,6 +164,17 @@ a.tertiary:focus-visible,
 a.subtle.secondary:hover,
 a.subtle.secondary:focus-visible {
   border-color: transparent;
+}
+
+:global(.jobber-retheme) .tertiary:hover,
+:global(.jobber-retheme) .tertiary:focus-visible,
+:global(.jobber-retheme) a.tertiary:hover,
+:global(.jobber-retheme) a.tertiary:focus-visible,
+:global(.jobber-retheme) .subtle.secondary:hover,
+:global(.jobber-retheme) .subtle.secondary:focus-visible,
+:global(.jobber-retheme) a.subtle.secondary:hover,
+:global(.jobber-retheme) a.subtle.secondary:focus-visible {
+  border-color: var(--button--color-variation--hover);
 }
 
 .subtle.tertiary {
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -23,7 +23,7 @@
   vertical-align: middle;
   background-color: var(--button--color-surface);
   cursor: pointer;
-  transition: all var(--timing-base);
+  transition: all var(--timing-base) ease-out;
   appearance: none;
   fill: var(--button--color-variation);
   align-items: center;
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -43,7 +43,7 @@
 :global(.jobber-retheme) .button * {
   font-weight: 600 !important;
   line-height: 1;
-  letter-spacing: none;
+  letter-spacing: 0;
   /* post-retheme this can move into the TSX on L177 */
 }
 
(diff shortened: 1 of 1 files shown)


packages/components/src/Button/Button.css
```diff
@@ -30,16 +30,26 @@
   justify-content: center;
 }
 
+:global(.jobber-retheme) .button {
+  min-height: calc(var(--base-unit) * 2.5);
```
chris-at-jobber, line 34: thinking about absorbing these height values into a var because they get re-used in icons but for now we can leave it, I think

packages/components/src/Button/Button.css
```diff
@@ -48,12 +58,12 @@ a.button:active {
...
 
-.button:focus,
-a.button:focus,
+.button:focus-visible,
+a.button:focus-visible,
 .button:active,
 a.button:active {
   box-shadow: var(--shadow-focus);
-  outline: none;
+  outline: transparent;
```
chris-at-jobber, line 66: Allows users with high-contrast assistive tech to override our outline with a more powerful one of their choosing

packages/components/src/Button/Button.css
```diff
@@ -171,22 +181,37 @@ a.button.disabled:hover {
...
+  padding: 0 calc(var(--space-base) - var(--space-smaller));
   letter-spacing: 0;
 }
 
+:global(.jobber-retheme) .small {
+  padding: calc(var(--space-base) - var(--space-smaller));
+}
+
 .base {
-  min-height: calc(var(--base-unit) * 2.25);
```
chris-at-jobber, line 179: this min-height is actually covered by `.button`

packages/components/src/Button/Button.css
```diff
@@ -171,22 +181,37 @@ a.button.disabled:hover {
...
 }
 
+:global(.jobber-retheme) .large {
+  padding: 0 calc(var(--space-base) + var(--space-smaller));
+}
+
 /* Icon */
 
+.button svg {
+  flex-shrink: 0;
```
chris-at-jobber, line 212: This protects the icon from collapsing and is useful even in non-re-theme state

darryltec:
There's a slight difference between sg1 and atlantis 🤔 

image

chris-at-jobber:
Ooh yes, I need to override that letter-spacing that's coming from the `Typography`. It also looks like the spacing between the icon and the label is smaller in SG-1, I'll see if I can get that to match.

darryltec:
Seems to be because it has a negative margin 
image

darryltec:
not sure who's wrong on this one tho haha

image

darryltec:
Hmm... what's the plan on the decision for "tertiary looks like secondary now" 

This is SG1
image

darryltec (changes requested):
Aside from the letter spacing and icon spacing this looks good. Had some questions tho

chris-at-jobber:
@darryltec made the fixes, discussed the "question" ones [here](https://getjobber.slack.com/archives/GBKR5RA7K/p1702925871571829) - I don't think we've fully resolved the `surface--hover` vs no `surface--hover`, but you should be able to review the fixes and the "secondary = tertiary" change.

packages/components/src/Button/Button.css
```diff
@@ -200,29 +256,42 @@ a.button.disabled:hover {
...
 }
 
+:global(.jobber-retheme) .hasIconAndLabel > *:first-child {
+  margin-left: calc(var(--space-smaller) * -1);
+}
+
+:global(.jobber-retheme) .iconOnRight > *:first-child {
+  margin-left: 0;
+  margin-right: calc(var(--space-smaller) * -1);
+}
```
darryltec, line 293: :chefs-kiss:

packages/components/src/Button/Button.css
```diff
@@ -30,16 +30,27 @@
...
 .button * {
   color: inherit;
   fill: inherit !important;
   /* required to over-ride very specific fill from <Icon> */
 }
 
+:global(.jobber-retheme) .button * {
+  font-weight: 600 !important;
+  line-height: 1;
+  letter-spacing: none;
```
darryltec, line 46: i believe this has to be a unit and be 0

darryltec (changes requested):
Just need to fix that letter spacing and we solid!

packages/components/src/Button/Button.css
chris-at-jobber: Say no more!

darryltec (approved):
Approveled, as Scott would say

chris-at-jobber:
@darryltec updating the [sg1 hover state](https://github.com/GetJobber/sg-1/pull/1390) to match what we did here

PR #1694, opened 2023-12-18
refactor(components): Separating Chip and Chips (for real this time)

scotttjob:
## Motivations

We needed a standalone Chip for DataList/ClientList/Combobox. Since our existing implementation required a group of chips, an effort began to redesign Chip, implement that new Chip, and then decouple Chip from Chips. We've already redesigned and implemented the new Chip, this is the final step, decoupling Chip from Chips.

The original version of this PR made a critical mistake early on, deleting the InternalChip file instead of using it to properly integrate the new Chip component. This version of the PR corrects that mistake, and as a result this is a drastically simpler PR, with less breaking changes.

## Changes

Chips no longer has its own internal implementation of a Chip, instead using the decoupled/unopinionated `<Chip>` that is also available to all consumers of Atlantis.

#### Previous Chips Single Select
image

#### New Chips Single Select
image

#### Previous Chips Multi Select
image

#### New Chips Multi Select
image

#### Previous Chips Selection
image

#### New Chips Selection
image

### Added

- No net new development

### Changed

- There should be no noticeable changes by consumers of Atlantis other than the design of Chips is now in line with `Chip`. 

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- N/A

### Security

- N/A

## Testing

1. Ensured all existing functionality in storybook is preserved.
1. Ensured all existing unit tests pass with the changes.

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Chips/Web.stories.tsx
+++ b/docs/components/Chips/Web.stories.tsx
@@ -2,9 +2,7 @@ import React, { useState } from "react";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
 import { Content } from "@jobber/components/Content";
 import { Chip, Chips } from "@jobber/components/Chips";
-import { Icon } from "@jobber/components/Icon";
 import { Text } from "@jobber/components/Text";
-import { Avatar } from "@jobber/components/Avatar";
 import { useFakeOptionQuery } from "./utils/storyUtils";
 
 export default {
@@ -32,22 +30,15 @@ const BasicTemplate: ComponentStory<typeof Chips> = args => {
       <Text>
         You are <u>{selected ? selected : "_______"}</u>
       </Text>
-      <Chips {...args} selected={selected} onChange={setSelected}>
-        <Chip
-          prefix={<Avatar initials="AZ" />}
-          label="Amazing"
-          value="Amazing"
-        />
-        <Chip
-          prefix={<Icon name="video" />}
-          label="Wonderful"
-          value="Wonderful"
-        />
-        <Chip
-          prefix={<Icon name="starFill" />}
-          label="Brilliant"
-          value="Brilliant"
-        />
+      <Chips
+        {...args}
+        selected={selected}
+        onChange={setSelected}
+        type="singleselect"
+      >
+        <Chip label="Amazing" value="Amazing" />
+        <Chip label="Wonderful" value="Wonderful" />
+        <Chip label="Brilliant" value="Brilliant" />
         <Chip label="Magnificent" value="Magnificent" />
       </Chips>
     </Content>
@@ -97,6 +88,7 @@ const SelectionTemplate: ComponentStory<typeof Chips> = args => {
   return (
     <Chips
       {...args}
+      type="dismissible"
       selected={selected}
       onChange={handleSelect}
       onCustomAdd={handleCustomAdd}
--- a/packages/components/src/Chip/Chip.css
+++ b/packages/components/src/Chip/Chip.css
@@ -8,11 +8,11 @@
   --chip-radius: 20px;
   --chip-base-bg: var(--color-surface--background);
   --chip-base--hover: var(--color-surface--background--hover);
-  display: flex;
+  display: inline-flex;
   position: relative;
   height: var(--chip-height);
   min-width: 0;
-  padding: var(--space-small) var(--base-unit);
+  padding: 0 var(--base-unit);
   border: var(--border-base) solid transparent;
   border-radius: var(--chip-radius);
   color: var(--color-heading);
@@ -77,8 +77,7 @@
   --chip--gradient-color-variation: var(--chip-base-bg);
 }
 
-.base:hover,
-.base:focus-visible {
+.base:hover {
   background-color: var(--chip-base--hover);
   --chip--gradient-color-variation--hover: var(--chip-base--hover);
 }
--- a/packages/components/src/Chip/Chip.test.tsx
+++ b/packages/components/src/Chip/Chip.test.tsx
@@ -49,13 +49,7 @@ describe("Chip", () => {
 
   it("should have a role of button when role not provided", () => {
     const { getByRole } = render(<Chip label="Test Chip" />);
-    expect(getByRole("button")).toBeInstanceOf(HTMLButtonElement);
-  });
-
-  it("should have a tabIndex of zero if one is not provided", () => {
-    const { getByRole } = render(<Chip label="Test Chip" />);
-
-    expect(getByRole("button")).toHaveAttribute("tabIndex", "0");
+    expect(getByRole("button")).toBeInstanceOf(HTMLDivElement);
   });
 
   it("should set aria-label if one is provided", () => {
@@ -64,12 +58,6 @@ describe("Chip", () => {
     );
 
     expect(getByRole("button")).toHaveAttribute("aria-label", "Test Label");
-  });
-
-  it("should default aria-label to provided label if not provided", () => {
-    const { getByRole } = render(<Chip label="Test Chip" />);
-
-    expect(getByRole("button")).toHaveAttribute("aria-label", "Test Chip");
   });
 });
 
--- a/packages/components/src/Chip/Chip.tsx
+++ b/packages/components/src/Chip/Chip.tsx
@@ -16,11 +16,12 @@ export const Chip = ({
   invalid,
   label,
   value,
+  dataTestID,
   onClick,
   onKeyDown,
   children,
   role = "button",
-  tabIndex = 0,
+  tabIndex,
   variation = "base",
 }: ChipProps): JSX.Element => {
   const classes = classnames(styles.chip, {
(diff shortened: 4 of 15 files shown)


docs/components/Chips/Web.stories.tsx
```diff
@@ -26,27 +24,21 @@ export default {
...
-          prefix={<Icon name="video" />}
-          label="Wonderful"
-          value="Wonderful"
-        />
-        <Chip
-          prefix={<Icon name="starFill" />}
-          label="Brilliant"
-          value="Brilliant"
-        />
+      <Chips
```
scotttjob, line 33: New design for this doesn't have the prefixes, so I've removed them for the test.

docs/components/Chips/Web.stories.tsx
```diff
@@ -91,9 +84,11 @@ const SelectionTemplate: ComponentStory<typeof Chips> = args => {
     handleSelect,
     handleCustomAdd,
   } = useFakeOptionQuery();
+
   return (
     <Chips
       {...args}
+      type="dismissible"
```
scotttjob, line 91: I had issues with XOR in this branch. I have a feeling a recent update to ts-xor caused issues with the original implementation of the prop splitting for single/multiple/select. 

I updated the XOR usage to the typescript unary operator and made the usage of the different chipp types explicit and there's no longer errors about the activator being undefined like in the x-or implementation (which it should have been able to be for the single-select elsewhere in this test file)

packages/components/src/Chip/Chip.test.tsx
```diff
@@ -49,13 +49,7 @@ describe("Chip", () => {
 
   it("should have a role of button when role not provided", () => {
     const { getByRole } = render(<Chip label="Test Chip" />);
-    expect(getByRole("button")).toBeInstanceOf(HTMLButtonElement);
-  });
-
-  it("should have a tabIndex of zero if one is not provided", () => {
```
scotttjob, line 55: Chip no longer defaults to zero if you don't pass in a value, so this test is no longer valid.

packages/components/src/Chip/Chip.test.tsx
```diff
@@ -65,12 +59,6 @@ describe("Chip", () => {
 
     expect(getByRole("button")).toHaveAttribute("aria-label", "Test Label");
   });
-
```
scotttjob, line 68: Chip no longer defaults the aria-label to the passed in label, so this test is no longer valid (was causing the double up errors in testing because two labels were present on the screen)

packages/components/src/Chip/Chip.tsx
```diff
@@ -16,11 +16,12 @@ export const Chip = ({
   invalid,
   label,
   value,
+  dataTestID,
   onClick,
   onKeyDown,
   children,
   role = "button",
-  tabIndex = 0,
+  tabIndex,
```
scotttjob, line 24: The original Chip implementation didn't do this, so we shouldn't either.

packages/components/src/Chip/Chip.tsx
```diff
@@ -41,19 +42,21 @@ export const Chip = ({
     label,
     heading,
   );
+  const Tag = onClick ? "button" : "div";
```
scotttjob, line 45: Returning this functionality from the original chip.

packages/components/src/Chip/Chip.tsx
```diff
@@ -41,19 +42,21 @@ export const Chip = ({
...
 
   return (
-    <Tooltip message={tooltipMessage}>
-      <button
+    <Tooltip message={tooltipMessage} setTabIndex={false}>
```
scotttjob, line 48: Tooltip always setting the tabIndex to zero which was causing issues with the tabbing/keyboard controls. So I needed a way to selectively disable that functionality in certain scenarios (when used within a Chips)

packages/components/src/Chip/Chip.types.d.ts
```diff
@@ -6,6 +6,11 @@ export interface ChipProps extends PropsWithChildren {
    */
   readonly ariaLabel?: string;
 
+  /**
+   * The testing id for the chip if necessary. Defaults to `chip-wrapper`.
+   */
+  dataTestID?: string;
```
scotttjob, line 12: This needs to be a simple pass-through to support Chips (without changing the tests too much). If we wanted to lock down this usage to something specific, we'd have to update all the `chip-wrapper` usages in the tests.

packages/components/src/Chip/components/ChipSuffix/Chip.Suffix.tsx
```diff
@@ -1,13 +1,19 @@
...
 import { Icon } from "@jobber/components/Icon";
+import { InternalChipButton } from "@jobber/components/Chips/InternalChipButton";
 import { useChildComponent } from "../../hooks";
 import styles from "../../Chip.css";
 
 export function ChipSuffix({ children, className }: ChipSuffixProps) {
-  let singleChild = useChildComponent(children, d => d.type === Icon);
+  let singleChild = useChildComponent(
+    children,
+    d => d.type === Icon || d.type == InternalChipButton,
```
scotttjob, line 11: Adding support for the dismissible button. This is a bit too much of a 'tight-couple' for me to be comfortable, but since it's just checking a type and not actually using functionality I was okay with it.

packages/components/src/Chips/ChipsTypes.tsx
```diff
@@ -96,7 +95,6 @@ export interface ChipDismissibleProps extends ChipFoundationProps {
   onLoadMore?(searchValue: string): void;
 }
 
-export type ChipsProps = XOR<
-  ChipSingleSelectProps,
-  XOR<ChipMultiSelectProps, ChipDismissibleProps>
->;
+export type ChipsProps =
```
scotttjob, line 98: I didn't want to edit this, but with the XOR in place, any usage of Chips in the test files causes an error around the activator not being allowed to be undefined (even though we clearly said it was).

So instead of XOR I just used the Typescript unary operator, and explicitly set usage in Chips to single/multi/dismissible and was able to get proper type checking on my props.

packages/components/src/Chips/InternalChip.css
```diff
@@ -59,11 +59,16 @@
  */
 
 .chip:focus,
-.input:focus ~ .chip {
+.input:focus-visible ~ div[role="button"] {
```
scotttjob, line 62: Wasn't getting the focus ring without this change, which we need! :)

packages/components/src/Chips/InternalChip.tsx
```diff
@@ -18,42 +13,20 @@ export function InternalChip({
...
-      </Typography>
-      <InternalChipAffix affix={suffix} {...affixProps} />
-    </Tag>
+      {prefix && <Chip.Prefix>{prefix}</Chip.Prefix>}
+      {suffix && <Chip.Suffix>{suffix}</Chip.Suffix>}
+    </Chip>
   );
-
-  return isTruncated() ? <Tooltip message={label}>{chip}</Tooltip> : chip;
-
```
scotttjob, line 53: This is handled now in the Chip itself, so we're okay to remove this double usage.

packages/components/src/Chips/InternalChip.tsx
```diff
@@ -1,14 +1,9 @@
-import React, { useState } from "react";
-import classnames from "classnames";
-import styles from "./InternalChip.css";
+import React from "react";
```
scotttjob, line 1: Kept this file around this time so we could use it to configure Chip easily. Really cut down on the amount of changes compared to the other branch.

packages/components/src/Chips/InternalChipMultiSelect.tsx
```diff
@@ -32,7 +33,9 @@ export function InternalChipMultiSelect({
             <InternalChip
               {...chip.props}
               active={isChipActive}
-              suffix={checkmarkIcon(isChipActive)}
+              {...(isChipActive
```
scotttjob, line 36: We changed the functionality of InternalChip a bit to not show the suffix if we don't provide it, so this just changes the prop passing a bit to reflect that.

packages/components/src/Chips/tests/InternalChip.test.tsx
```diff
@@ -22,33 +22,9 @@ it("should fire the callback when it's clicked", async () => {
 });
 
 describe("Chip icon colors depending on state", () => {
-  it("should be red when it's invalid", () => {
```
scotttjob, line 25: Along with the skipped tests at the bottom of this file (that I removed), there is something up with trying to check for the fill color on these icons. Easily the longest time I spent on this PR was trying to get these few tests to pass. I checked the render in the browser, and they render properly there, but for some reason don't render properly in the tests, at least not being able to pull the colours out via toHaveStyle (everything is the base style, it never picks up the cascaded styles). Since everything renders well elsewhere, and because of the skipped tests at the bottom, I'm okay with pulling these tests out.

I took the exact code out, rendered it in the browser, and everything is disabled properly or red properly, but when you render in these tests, everything always comes back as though it's the original render without disabled/invalid.

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -21,9 +21,15 @@ interface TooltipProps {
    * Tooltip text
    */
   readonly message: string;
+
+  readonly setTabIndex?: boolean;
```
scotttjob, line 25: Forcibly having a tabIndex of zero on the top level of the Chip was interfering with tab and keyboard based controls for Chips so we had to make setting the tabIndex configurable.

packages/components/src/Chip/Chip.css
```diff
@@ -8,11 +8,11 @@
   --chip-radius: 20px;
   --chip-base-bg: var(--color-surface--background);
   --chip-base--hover: var(--color-surface--background--hover);
-  display: flex;
+  display: inline-flex;
```
scotttjob, line 11: regular flex causes fun height issues when the button becomes a div

chris-at-jobber:
@ZakaryH we can close this one now right?

ZakaryH:
goodnight sweet prince (implemented in another PR)

PR #1697, opened 2023-12-19
feat(components): Add clearable to DataList Search

rodrigoeidelvein:
## Motivations

Making it easier for SPs to clear the search on DataList. 

I had to add a `max-width` to the input search container because it increased the width when the clear button was being shown. I set it to `base * 13` (208px) because it's closer to the current width that was being rendered (214px). I'm open to suggestions if you know a better solution.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
+++ b/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
@@ -32,7 +32,6 @@
     position: static;
     visibility: visible;
     width: auto;
-    max-width: calc(var(--base-unit) * 13);
     padding: 0;
     background-color: transparent;
     opacity: 1;
--- a/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
+++ b/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
@@ -32,6 +32,7 @@
     position: static;
     visibility: visible;
     width: auto;
+    max-width: calc(var(--base-unit) * 13);
     padding: 0;
     background-color: transparent;
     opacity: 1;
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -29,7 +29,7 @@
   display: grid;
   padding: var(--space-small) 0;
   gap: var(--space-small);
-  grid-template-columns: auto max-content;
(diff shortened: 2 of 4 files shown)


packages/components/src/DataList/components/DataListSearch/DataListSearch.css
```diff
@@ -32,6 +32,7 @@
     position: static;
     visibility: visible;
     width: auto;
+    max-width: calc(var(--base-unit) * 13);
```
chris-at-jobber, line 35: Isn't the row that holds the search using `<Grid>`? I would expect this search input to take up the prescribed column value rather than be responsible for its' own width

rodrigoeidelvein: We're not using `Grid` here. But we are using `display: grid;` [here](https://github.com/GetJobber/atlantis/blob/080c07cb01bf6842de106c51f167120ec2a168a9/packages/components/src/DataList/DataList.css#L32), where I could set `grid-template-columns: auto fit-content(208px);` and it would work too.

chris-at-jobber: Oh OK - I'm not super opinionated about how we do it, but it does seem like we might want to put the layout responsibilities outside of this component.

chris-at-jobber:
Is it possible to get the clear button closer to the right of the input when it's small?

image

Also, I just went back and looked at the original mockups for this list, and I think this should be using `base` size Input:

image

rodrigoeidelvein:
@chris-at-jobber I changed the input to `base`, now it shows the mini label when there's a value. Is that okay? I also moved the clear button a little bit to the left of the input and handled the max-height in the grid container

chris-at-jobber:
Looks good! Just need a smidgen more padding when it collapses for small screens:
image

Currently it's `small`, if we go up to something like `small` + `smaller` we should be good:
image
image

chris-at-jobber (changes requested):
Just this change [here](https://github.com/GetJobber/atlantis/pull/1697#issuecomment-1865260006) and we're good

chris-at-jobber (approved):

PR #1700, opened 2023-12-20
feat: Input time auto fill

darryltec:
## Motivations

Native Input Time lacks the ability to fill in the rest of the time values automatically. This created a friction with our users having to type in everything and that takes too much time if you're always setting it by the hour or NN:00`. Which our users do.

This aims to add a bit of optimization to our users day by automatically predicting the full time the user wants depending on the time of day and what number they type.

### For users in a 12 hour format

- We have to predict 10, 11, 12, or 1 when the user types 1.
  - We wait for a few milliseconds for another number in case they want to type another number.
- Typing any other number would immediately set it to that hour. I.e., `2 -> 2:00PM`.
- Typing anything after 5 would result in the type-ahead setting the time to AM instead of PM. 
  - I.e., `5 -> 5:00PM` `6 -> 6:00AM` since our data suggests that

### For 24 hours

- This one is pretty straight-forward as 1 would predict from 10-19 and 2 would predict 20-23 and any other number would be set to their respective hour (`9 -> 9:00`) immidiately
- If the user doesn't the another number after 1 or 2, it'll be set to the next hour (if it's 13:00 right now and you type only 1, it'll predict that you want 14:00)

Once the values are set, the type-ahead stops and the native experience takes over, which is exactly the UX we want at that point.

## Known Quirk

- You can fill out HH, MM, AM/PM but the number you type is always gonna be the hour

### Added

- time predictor hook
- only available for controlled components (for now)

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- 12 hours is easy to test as our machines are in a 12 hour format.
- 24 hours needs 2 things. 
  - Set your browser language to en-GB so it thinks you're in location with a 24 hour format
  - Go to your System preferences > General > Date and Time > Toggle 24 hours
  - Restart your browser

---

Random photo of Atlantis

JOB-86009


jlelford:
> always go to AM if you set a number after 7 based on the (unvalidated) assumption that most SPs don’t work after 7pm

Tried out the newer build, I think you have the right idea here I think we should be more aggressive with it, I think if the number is after 5 we should default to AM. I think a 6 or 7 am job is more likely than a 6 or 7 pm, but you're right it should be validated. I'll get someone to check start times of jobs and see if our assumptions line up.

Otherwise this is a great improvement Derl!

packages/components/src/InputTime/InputTime.tsx
```diff
@@ -27,7 +26,18 @@ export function InputTime({
...
-    return <FormField type="time" {...fieldProps} />;
+    return (
+      <FormField
+        inputRef={ref}
+        type="time"
+        {...fieldProps}
+        onBlur={handleBlur}
+        onKeyUp={e => {
+          fieldProps.onKeyUp?.(e);
+          !isNaN(parseInt(e.key, 10)) && setTypedTime(prev => prev + e.key);
```
scotttjob, line 37: Total nit so feel free to ignore and this is something I only learned recently myself, but the `Number()` typescript function is a 'better' way to parse numbers. I put better in quotes because it's subjective and parseInt still has a place when you specifically want an integer and not a float. You don't have to provide the radix argument for Number(), it's faster because it straight up doesn't try to parse garbage, it's either a number or it isnt (123abc is NaN instead of 123). 

Not something for this PR, but something I wanted to let you know of! :)

Also as a bit of an actual nit, these two lines are slightly clunky. Maybe an if statement instead of the inline `&&`? But maybe it's fine and the rest of this PR is just so clean that this stands out a tiny bit. Like, a super tiny amount. I have to comment on SOMETHING. :)

packages/components/src/InputTime/hooks/__test__/useTimePredict.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    beforeEach(() => {
+      jest.spyOn(Intl, "DateTimeFormat").mockReturnValue({
+        // @ts-expect-error - We don't need to mock all the methods
+        resolvedOptions: () => ({ hour12: true }),
+      });
+    });
+
+    it.each`
+      currentTime | typedTime | expected
+      ${"8:24"}   | ${"1"}    | ${"10:00"}
```
scotttjob, line 35: Okay so hear me out. This rocks. This is a great test.

Is there any value in providing this to the team that's asking for it, and getting them to double check the expected values? It might be a really good way to verify it's what they're asking for.

Kind of like, alright this is what we're testing for. Does this meet your needs exactly? I'm sure you're already talking to them, so even if this comment is irrelevant because you've already done this work, this kind of test is the sort of thing we can put in front of non-tech people and say "does this line up with what you want."

Either way, great test. Great work.

packages/components/src/InputTime/hooks/useTimePredict.ts
```diff
@@ -0,0 +1,167 @@
+import debounce from "lodash/debounce";
+import { useCallback, useEffect, useRef, useState } from "react";
+import { InputTimeProps } from "../InputTimeProps";
+
+interface UseTimePredictProps extends Pick<InputTimeProps, "value"> {
+  readonly handleChange: (value: string) => void;
+}
+
+const DEBOUNCE_TIME = 300;
```
scotttjob, line 9: Is there any value in exposing this as something consumers can override?

packages/components/src/InputTime/hooks/useTimePredict.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export function useTimePredict({ value, handleChange }: UseTimePredictProps) {
```
scotttjob, line 11: Just a general comment that I really like this hook. Clean code, lots of whitespace, self-documented but you also included real comments to help out future maintainers.

Is there any value to putting a comment at the top of the file to just generally explain motivation, purpose and scope as well? 

This is some really great code and I'm worried in the future we'll start adding little tweaks to it here and there and eventually lose sight of why we did it, so having a bit of a block explaining why we did this, what the limits are (intentionally) and what it's 'not' doing. That way if we ever get down a rabbit hole in the future with it we'll know what it's 'supposed' to be doing.

packages/components/src/InputTime/InputTime.test.tsx
```diff
@@ -1,185 +1,37 @@
 import React from "react";
-import { fireEvent, render } from "@testing-library/react";
+import { fireEvent, render, screen } from "@testing-library/react";
 import { CivilTime } from "@std-proposal/temporal";
 import { InputTime } from ".";
 
-it("renders a InputTime", () => {
```
scotttjob, line 6: I like seeing all the deleted inline snapshots. I think they were the items giving me issues in the other stream of work, so this might make my life easier! :)

scotttjob (approved):
Code LGTM! 

I ran the Storybook examples, everything looked and rendered fine. I caught a very small bug that I let Darryl know about out of band and he's already fixing so I don't want to hold this PR up with a request changes.

The rest of my comments are either praise, or the tiniest of nits that we can absolutely go to prod without completing so I'm comfortable approving this.

darryltec:
Will do a follow up on those comments since i agree with them. :shipit:

PR #1701, opened 2023-12-20
chore: prep banner for retheme

darryltec:
## Motivations

Banners are being redesigned but it's in experiment so hide it behind a flag.

image

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

add this anywhere on preview.tsx as the class needs to exist before the component renders

```ts
setTimeout(() => {
  document.body.classList.add("jobber-retheme");
}, 0);
```

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Banner/Banner.tsx
+++ b/packages/components/src/Banner/Banner.tsx
@@ -5,11 +5,11 @@ import { useResizeObserver } from "@jobber/hooks/useResizeObserver";
 import styles from "./Banner.css";
 import types from "./notificationTypes.css";
 import { BannerIcon } from "./components/BannerIcon";
+import { BannerType } from "./Banner.types";
 import { Icon } from "../Icon";
 import { Text } from "../Text";
 import { Button, ButtonProps } from "../Button";
-
-export type BannerType = "notice" | "success" | "warning" | "error";
+import { getAtlantisConfig } from "../utils/getAtlantisConfig";
 
 interface BannerProps {
   readonly children: ReactNode;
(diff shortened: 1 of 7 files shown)


packages/components/src/Banner/Banner.tsx
```diff
@@ -78,37 +79,38 @@ export function Banner({
...
+        )}
+      </div>
+
+      {dismissible && (
+        <button
+          className={styles.closeButton}
+          onClick={handleClose}
+          aria-label="Close this notification"
+        >
+          <Icon name="cross" color={iconColors[type]} />
```
chris-at-jobber, line 110: Icon should use `interactive--subtle` on all options now

chris-at-jobber: Post re-theme we could probably just drop in `ButtonDismiss` here

darryltec: only for retheme, yeah?

chris-at-jobber: For now, yeah

packages/components/src/Banner/Banner.tsx
```diff
@@ -117,6 +123,22 @@ export function Banner({
   }
 }
 
+function getBannerIcon(type: BannerType): IconNames | undefined {
```
chris-at-jobber, line 126: Is this only meant to override the icon if the user hasn't already set an icon? When I cycle between banner types the Icon doesn't update

chris-at-jobber: (Which I guess is a safe way to approach it, if someone has specified a preferred icon for a given use case, we leave that to them - otherwise use a sensible default)

chris-at-jobber (approved):

packages/components/src/Banner/Banner.tsx
darryltec: yes! you got it haha

PR #1703, opened 2023-12-21
fix(components): Fix th cell padding in DataTable

mattander:
## Motivations

Our team has a design debt ticket that led us to discover this.

In DataTable all of the table body cells (`tbody > tr > td`) have space-base for the cell padding on all sides. However, for table headers (`thead > tr > th`) we have an inline style that is overriding `padding-right`. This is in place to override to `0` if we are manually sorting the columns (to make room for the sorting control). However, if manual sorting is falsy our fallback is `padding-right: inherit` which I believe was meant to fallback on the default padding set on the th. 

Unfortunately that's not quite how inherit works. Inherit instead tells the cascade to fallback on the defined default in the **parent** element or the browser default if no parent default is found. So we end up with `padding-right: 0` either way. In order to fix this, I moved the inline style override into an object and then if we are manually sorting I add the padding right override. This way when we aren't manually sorting, there's no inline style to contend with. 

Before:
Screenshot 2023-12-21 at 7 32 54 AM
Screenshot 2023-12-21 at 7 33 03 AM

After:
Screenshot 2023-12-21 at 7 24 34 AM
Screenshot 2023-12-21 at 7 24 38 AM
Screenshot 2023-12-21 at 7 25 50 AM
Screenshot 2023-12-21 at 7 25 59 AM

### Fixed

- Incorrect table heading cell padding

## Testing

Check padding without manual sorting. Check padding wtih manual sorting.

---

image

### DIFF
--- a/packages/components/src/DataTable/Header.tsx
+++ b/packages/components/src/DataTable/Header.tsx
@@ -6,10 +6,10 @@ import styles from "./DataTable.css";
 import { SortingType } from "./types";
 
 interface HeaderProps<T> {
-  table: Table<T>;
-  stickyHeader?: boolean;
-  sorting?: SortingType;
-  onRowClick?: (row: Row<T>) => void;
+  readonly table: Table<T>;
+  readonly stickyHeader?: boolean;
+  readonly sorting?: SortingType;
+  readonly onRowClick?: (row: Row<T>) => void;
 }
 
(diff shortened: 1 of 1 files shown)


rodrigoeidelvein (approved):
Nice catch!

LGTM!

PR #1704, opened 2023-12-21
fix(components): prevent multiple breakpoints being registered within DataList

ericchernuka:
## Motivations
When using DataList within an app that has StrictMode enabled any useEffect's will be run at least twice on mount. Due to the way DataList registers layouts for determining the which layout to render based on the active breakpoint. When this registration happens multiple times it caused multiple breakpoints of the same name to be registered `["xs", "xs", "md", "md"]`.  There was an issue with the logic where it decided if the DataListLayout should render based on whether the breakpoint is current and the next available is not present, but since there were 2 values for the same breakpoint, it caused the component not to render the layout.

### Fixed

To prevent needless re-renders and state, I've reduced down to only registering the layout once since `children` will cause the useEffect to fire on every render. Then, since we stored the layout as an object, I memoized the breakpoints as an array from that collection.

Lastly, moving DataListLayout to utilize the useActiveLayout hook allows for consolidation of the logic of the currently visible breakpoint so that the logic is a simple equality check.

### Security

- 

## Testing

In building out this fix I had to enable StrictMode within storybook, but that caused a whole slew of issues with the entire Storybook. To get the fix ready I had to do the following:
```
.storybook/preview.js
// Comment out the decorators export as they seem to be the source of the "gave too many hooks and a previous render" exceptions.

.storybook/main.js
// Add the following configuration to the config
const config = {
  reactOptions: {
    strictMode: true,
  },
  stories: [
```

---

Random photo of Atlantis


darryltec (approved):
can confirm. still works as expected

PR #1706, opened 2024-01-04
fix(components): Attemping to fix typing issues for Vue-based Atlantis consumption

scotttjob:
## Motivations

We've had reports that users building with Vue are seeing props typing issues with Atlantis components that generate acceptable prop values from css files.

The solution presented in this PR is a bit crude, but it does solve the issue in local testing with no changes to the core build process + local dev experience (the css.d.ts files are currently generated files and removing that step from the build process opened up a small can of worms). 

This PR will have a preview build generated from it so Vue consumers can verify that our local testing success can be replicated in other environments.

## Changes

- Updated library build process to copy specific css.d.ts files to the `dist` folder

### Fixed

- Vue-based Atlantis consumers should no longer receive errors when attempting to use specific props on certain components, and additionally gain autocomplete on the previously errored props.

## Testing

- Consume Atlantis in a Vue based environment, and attempt to use the `accent` prop on the `Card` component and you should see a full list of available accent colours instead of an error.

---

Random photo of Atlantis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13050,6 +13050,15 @@
       "integrity": "sha512-hzzTS+X9EqDhx4vwdch/DOZci/bfh5J6Nyz8lqvyfBg2ROx2fPafX+LpDfpVgSvQKj0EYkwTYpBO3z2etwbkOw==",
       "dev": true
     },
+    "node_modules/@types/fs-extra": {
+      "version": "8.1.5",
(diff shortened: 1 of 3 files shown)


timoteialbu (approved):
LGTM - Thanks for making this change. It's fixed the issues we were seeing with types.

PR #1708, opened 2024-01-04
feat: Add support for providing your own of acceptable file types to upload for InputFile

ajpeng:
## Motivations

feat(components): I want to be able to provide a list of acceptable MIME types such that I can restrict what the user should be able to upload.  

### Added
- Added another option for `allowedTypes` called `custom`. When this option is selected it will look at `acceptedMIMETypes` as the source of what MIME types the user can upload. 
- An additional prop called `acceptedMIMETypes`. This is prop expects a list of MIMETypes which will be passed down to the `useDropzone` hook. 

## Testing
Here we are setting the MIME type as `image/heic`. For a list of MIME types we can refer to either the [MDN web docs])https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) or [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml)

image

This alerts the file selector that we are looking for a HEIF image and everything else is greyed out.
image

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputFile/InputFile.test.tsx
+++ b/packages/components/src/InputFile/InputFile.test.tsx
@@ -42,8 +42,7 @@ describe("Post Requests", () => {
   it("renders an InputFile with custom accepted MIME types", () => {
     const { container } = render(
       <InputFile
-        allowedTypes="custom"
-        acceptedMIMETypes={["image/png", "image/jpg", "application/pdf"]}
+        allowedTypes={["image/png", "image/jpg", "application/pdf"]}
         getUploadParams={fetchUploadParams}
       />,
     );
--- a/packages/components/src/InputFile/InputFile.test.tsx
+++ b/packages/components/src/InputFile/InputFile.test.tsx
@@ -39,6 +39,17 @@ describe("Post Requests", () => {
     expect(container).toMatchSnapshot();
   });
 
+  it("renders an InputFile with custom accepted MIME types", () => {
+    const { container } = render(
+      <InputFile
+        allowedTypes="custom"
+        acceptedMIMETypes={["image/png", "image/jpg", "application/pdf"]}
+        getUploadParams={fetchUploadParams}
+      />,
+    );
+    expect(container).toMatchSnapshot();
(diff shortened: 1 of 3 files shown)


packages/components/src/InputFile/InputFile.tsx
```diff
@@ -103,10 +103,18 @@ interface InputFileProps {
    *
    * @param "images" - only accepts all types of image
    * @param "basicImages" - only accepts png, jpg and jpeg
+   * @param "custom" - uses `acceptedMIMETypes` as the list of allowed types
    *
    * @default "all"
    */
-  readonly allowedTypes?: "all" | "images" | "basicImages";
+  readonly allowedTypes?: "all" | "images" | "basicImages" | "custom";
```
darryltec, line 110: What if instead of custom, you allow the string[] that way, you won't need the other type

ajpeng: That also works. I will update the implementation

jessedoyle: Looks like this was addressed in https://github.com/GetJobber/atlantis/pull/1708/commits/1957cef5bce8df14081fb3350b3fbf178755c6fe!

packages/components/src/InputFile/InputFile.test.tsx
```diff
@@ -39,6 +39,16 @@ describe("Post Requests", () => {
...
   });
 
+  it("renders an InputFile with custom accepted MIME types", () => {
+    const { container } = render(
+      <InputFile
+        allowedTypes={["image/png", "image/jpg", "application/pdf"]}
+        getUploadParams={fetchUploadParams}
+      />,
+    );
+    expect(container).toMatchSnapshot();
```
darryltec, line 49: I'd prefer if we write an assertion test for this. can just find an input and then `toHaveAttribute` it.

darryltec (approved):
LGTM. Can do my last comment as fast follow tho

PR #1709, opened 2024-01-05
fix(components): InputFile update the useCallback dependency array [85395]

tyler-santos:
## Motivations

Currently the useCallback hook is being called without any dependencies. This means that the hook will never know when there have been changes, and not update accordingly. 

## Changes

This PR adds those missing depenedencies.

### Fixed

- This now has the correct dependencies.

## Testing

To test this is working properly, try updating state in a component that uses InputFile before and after these changes and notice that state will not update correctly without these changes.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -168,7 +168,12 @@ export function InputFile({
 }: InputFileProps) {
   const options: DropzoneOptions = {
     multiple: allowMultiple,
-    onDrop: useCallback(handleDrop, []),
+    onDrop: useCallback(handleDrop, [
+      getUploadParams,
+      onUploadStart,
+      onUploadProgress,
+      onUploadComplete,
+    ]),
   };
 
   if (allowedTypes === "images") {
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -168,12 +168,7 @@ export function InputFile({
 }: InputFileProps) {
   const options: DropzoneOptions = {
     multiple: allowMultiple,
-    onDrop: useCallback(handleDrop, [
-      getUploadParams,
-      onUploadStart,
-      onUploadProgress,
-      onUploadComplete,
-    ]),
+    onDrop: useCallback(handleDrop, [uploadFile]),
   };
 
   if (allowedTypes === "images") {


ZakaryH:
>try updating state in a component that uses InputFile before and after these changes and notice that state will not update correctly without these changes.

@tyler-santos could you expand on that a bit? I think I get what's happening and what the consequences might be but an example would help out

**My Understanding**

since `onDrop` is `handleDrop` from a  `useCallback` with no dependencies, we're never going to rebuild that function

`handleDrop` itself doesn't have a whole lot going on, it's really just a wrapper to do `uploadFIle` for each of its args. with that, I could see an argument for `uploadFIle` being the dependency since if that changes we wanna make sure we're using the latest version of it

the problem I would imagine is the following scenario. I have `function A` as `onUploadStart`/`onUploadProgress`/`onUploadComplete`

`handleDrop` says ok I am what I am and have no plans to change, I see what `uploadFIle` wants to do and that means using `function A`

for whatever reason, in the parent component we change any of those callback props to now be `function B`. the new props cause `InputFile` to re-render, however `handleDrop` doesn't rebuild and going forward when it gets called, the `uploadFile` invocation that comes from `uploadFile` will still be using `function A` despite us now having and wanting to use `function B`

I made a little proof of concept to understand this and you can see my logging that I'm passing in as a `onUploadStart` has diverged

image

do I have that right?

if yes, my follow up question is going to be - can we reduce the dependencies to just `uploadFIle` because I'm pretty sure that will get rebuilt when we re-render due to new props, and then we _should_ get the same desired behavior. I'll test this out while you read my little novel and respond!

ZakaryH (commented):
pretty sure I get this but wanna make sure and then have one follow up question assuming I do get it!

tyler-santos:
> since onDrop is handleDrop from a  useCallback with no dependencies, we're never going to rebuild that function

Yes! 

> handleDrop itself doesn't have a whole lot going on, it's really just a wrapper to do uploadFIle for each of its args. with that, I could see an argument for uploadFIle being the dependency since if that changes we wanna make sure we're using the latest version of it

Yes! 

> for whatever reason, in the parent component we change any of those callback props to now be function B. the new props cause InputFile to re-render, however handleDrop doesn't rebuild and going forward when it gets called, the uploadFile invocation that comes from uploadFile will still be using function A despite us now having and wanting to use function B

Yes! This is the main issue we are seeing. One small technical thing is that its the updating of state in the parent which causes the child to re-render, not the change in props but that's not really relevant to the problem. 

> if yes, my follow up question is going to be - can we reduce the dependencies to just uploadFIle because I'm pretty sure that will get rebuilt when we re-render due to new props, and then we should get the same desired behavior. I'll test this out while you read my little novel and respond!

I think you are right about this as well - since upload file encapsulates all of the other dependencies - changes to the others should cause a change in uploadFile to be registered and the function to be re-evaluated. I will do one little testing double check with that change and then push it up! 

Thanks so much for taking the time to review this, we appreciate it!

ZakaryH (approved):
I've looked at the existing instances of `InputFile` and I'm not seeing anything concerning where the callback functions are obviously changing in such a way that this change might break them. admittedly it would be a little strange for something to be relying on this cached callback behavior.

that said, I wouldn't mind increasing that confidence as much as possible by giving a couple of them a quick try. I was only able to successfully upload in 1/3 places I tried though the other 2 that weren't working for me on `master` either so that doesn't really tell me much of anything. if you know about some of these instances that'd be great, otherwise I might reach out before this goes into Jobber for real

image

PR #1710, opened 2024-01-05
feat: Added Error State and onErrorCallback for InputFile

ajpeng:
## Motivations
I notice that it was possible for the input file to fail in two places. The first place is when the component requests for upload params and the second place. when the file is being uploaded. The lack of having an onError callbacks makes it difficult for the consumer of the component to know if something went wrong.

We also allowed the option to pass down a validator for the files to be rejected. When the file is rejected, the UI will update with a validation error message that was provided by the validator method. 

## Changes
- Added an onUploadError callback which will be trigger when the component fails to fetch the upload params or when the upload process fails.
- Added error state to the component
- 
## Testing

For testing the validator method I used the following function and assigned it to the validator property.
```
function pngFileValidator(file: File) {
  if (!file.name.endsWith(".png")) {
    return {
      code: "wrong-file-type",
      message: "Only .png files are allowed",
    };
  }
  return null;
}
```

image

---

Random photo of Atlantis

### DIFF
--- a/docs/components/InputFile/InputFile.stories.mdx
+++ b/docs/components/InputFile/InputFile.stories.mdx
@@ -66,22 +66,25 @@ reject or accept file uploads based on your needs. The value of the function
 must accept a File and return `null` when it is accepted or an error object when
 rejected.
 
-For example, the code below only accepts .PNG files that are less than 2MB in
-size.
+For example, the code below only accepts files that are less than 2MB in size.
 
 ```ts
 function pngFileValidator(file: File) {
-  if (!file.name.endsWith(".png")) {
-    return {
-      code: "wrong-file-type",
-      message: "Only .png files are allowed",
-    };
-  } else if (file.size > 2000000) {
+  if (file.size > 2000000) {
     return {
       code: "file-too-big",
       message: "The file is too big. Try using a smaller image",
     };
   }
+
   return null;
 }
 ```
+
+#### Notes
+
+If you would like to restrict specific file types, it is advised that you use
+the `allowedTypes` property instead of the `validator` property. Using
+`allowTypes` will allow the OS know what type of files the uploader is expecting
+before the user can upload their file. The invalid files will be shown as
+greyed-out as a result of this behaviour.
--- a/docs/components/InputFile/InputFile.stories.mdx
+++ b/docs/components/InputFile/InputFile.stories.mdx
@@ -69,7 +69,7 @@ rejected.
 For example, the code below only accepts files that are less than 2MB in size.
 
 ```ts
-function pngFileValidator(file: File) {
+function fileSizeValidator(file: File) {
   if (file.size > 2000000) {
     return {
       code: "file-too-big",
--- a/docs/components/InputFile/InputFile.stories.mdx
+++ b/docs/components/InputFile/InputFile.stories.mdx
@@ -58,3 +58,30 @@ updating an array of `FileUpload`s based on their `key` as uploads progress.
  */
 export function updateFiles(updatedFile: FileUpload, files: FileUpload[]);
 ```
+
+### validator
+
+InputFile can accept a `validator` function as a prop. This will allow you to
+reject or accept file uploads based on your needs. The value of the function
+must accept a File and return `null` when it is accepted or an error object when
+rejected.
+
+For example, the code below only accepts .PNG files that are less than 2MB in
+size.
+
+```ts
+function pngFileValidator(file: File) {
+  if (!file.name.endsWith(".png")) {
+    return {
+      code: "wrong-file-type",
+      message: "Only .png files are allowed",
+    };
+  } else if (file.size > 2000000) {
+    return {
+      code: "file-too-big",
+      message: "The file is too big. Try using a smaller image",
+    };
+  }
+  return null;
+}
+```
--- a/docs/components/InputFile/InputFile.stories.mdx
+++ b/docs/components/InputFile/InputFile.stories.mdx
@@ -84,7 +84,4 @@ function fileSizeValidator(file: File) {
 #### Notes
 
 If you would like to restrict specific file types, it is advised that you use
-the `allowedTypes` property instead of the `validator` property. Using
-`allowTypes` will allow the OS know what type of files the uploader is expecting
-before the user can upload their file. The invalid files will be shown as
-greyed-out as a result of this behaviour.
+the `allowedTypes` property instead of the `validator` property.
--- a/packages/components/src/InputAvatar/InputAvatar.tsx
+++ b/packages/components/src/InputAvatar/InputAvatar.tsx
@@ -13,7 +13,7 @@ interface InputAvatarProps extends Omit<AvatarProps, "size"> {
    * More info is available at:
    * https://atlantis.getjobber.com/?path=/docs/components-forms-and-inputs-inputfile-docs--page#getuploadparams
    */
-  getUploadParams(file: File): Promise<UploadParams>;
+  getUploadParams(file: File): UploadParams | Promise<UploadParams>;
 
   /**
    * Triggered when an image is changed.
--- a/packages/components/src/InputAvatar/InputAvatar.tsx
+++ b/packages/components/src/InputAvatar/InputAvatar.tsx
@@ -13,7 +13,7 @@ interface InputAvatarProps extends Omit<AvatarProps, "size"> {
    * More info is available at:
    * https://atlantis.getjobber.com/?path=/docs/components-forms-and-inputs-inputfile-docs--page#getuploadparams
    */
-  getUploadParams(file: File): UploadParams | Promise<UploadParams>;
+  getUploadParams(file: File): Promise<UploadParams>;
 
   /**
    * Triggered when an image is changed.
--- a/packages/components/src/InputFile/InputFile.css
+++ b/packages/components/src/InputFile/InputFile.css
@@ -23,6 +23,10 @@
   background-color: var(--color-surface--hover);
 }
 
+.dropZone.error {
+  border: var(--color-critical) dashed var(--border-thick);
+}
+
 .validationErrors {
   display: flex;
   justify-content: center;
--- a/packages/components/src/InputFile/InputFile.css
+++ b/packages/components/src/InputFile/InputFile.css
@@ -29,6 +29,5 @@
 
 .validationErrors {
   display: flex;
-  justify-content: center;
   flex-direction: column;
 }
--- a/packages/components/src/InputFile/InputFile.css
+++ b/packages/components/src/InputFile/InputFile.css
@@ -26,4 +26,5 @@
 .validationErrors {
   display: flex;
   justify-content: center;
+  flex-direction: column;
 }
--- a/packages/components/src/InputFile/InputFile.css
+++ b/packages/components/src/InputFile/InputFile.css
@@ -24,7 +24,7 @@
 }
 
 .dropZone.error {
-  border: var(--color-critical) dashed var(--border-thick);
+  border-color: var(--color-critical);
 }
 
 .validationErrors {
--- a/packages/components/src/InputFile/InputFile.css
+++ b/packages/components/src/InputFile/InputFile.css
@@ -22,3 +22,8 @@
   border-color: var(--color-focus);
   background-color: var(--color-surface--hover);
 }
+
+.validationErrors {
+  display: flex;
+  justify-content: center;
+}
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -138,6 +138,11 @@ interface InputFileProps {
    * Upload event handler. Triggered on upload completion.
    */
   onUploadComplete?(file: FileUpload): void;
+
+  /**
+   *  Upload event handler. Triggered on upload error.
+   */
+  onUploadError?(error: Error): void;
 }
 
 interface CreateAxiosConfigParams extends Omit<UploadParams, "key"> {
@@ -165,6 +170,7 @@ export function InputFile({
   onUploadStart,
   onUploadProgress,
   onUploadComplete,
+  onUploadError,
 }: InputFileProps) {
   const options: DropzoneOptions = {
     multiple: allowMultiple,
@@ -229,7 +235,9 @@ export function InputFile({
       key = uuidv1(),
       fields = {},
       httpMethod = "POST",
-    } = await getUploadParams(file);
+    } = await getUploadParams(file).catch((e: Error) => {
+      onUploadError && onUploadError(e);
+    });
 
     const fileUpload = getFileUpload(file, key, url);
     onUploadStart && onUploadStart({ ...fileUpload });
@@ -254,7 +262,7 @@ export function InputFile({
       file,
       handleUploadProgress,
     });
-    axios.request(axiosConfig).then(handleUploadComplete);
+    axios.request(axiosConfig).then(handleUploadComplete).catch(onUploadError);
   }
 }
 
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -122,7 +122,7 @@ interface InputFileProps {
    * More info is available at:
    * https://atlantis.getjobber.com/?path=/docs/components-forms-and-inputs-inputfile-docs--page#getuploadparams
    */
-  getUploadParams(file: File): UploadParams | Promise<UploadParams>;
+  getUploadParams(file: File): Promise<UploadParams>;
 
   /**
    * Upload event handler. Triggered on upload start.
@@ -230,14 +230,17 @@ export function InputFile({
   }
 
   async function uploadFile(file: File) {
-    const {
-      url,
-      key = uuidv1(),
-      fields = {},
-      httpMethod = "POST",
-    } = await getUploadParams(file).catch((e: Error) => {
-      onUploadError && onUploadError(e);
-    });
+    let params;
+
+    try {
+      params = await getUploadParams(file);
+    } catch {
+      onUploadError && onUploadError(new Error("Failed to get upload params"));
+
+      return;
+    }
+
+    const { url, key = uuidv1(), fields = {}, httpMethod = "POST" } = params;
 
     const fileUpload = getFileUpload(file, key, url);
     onUploadStart && onUploadStart({ ...fileUpload });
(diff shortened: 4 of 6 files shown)


packages/components/src/InputFile/InputFile.tsx
```diff
@@ -123,7 +124,7 @@ interface InputFileProps {
    * More info is available at:
    * https://atlantis.getjobber.com/?path=/docs/components-forms-and-inputs-inputfile-docs--page#getuploadparams
    */
-  getUploadParams(file: File): UploadParams | Promise<UploadParams>;
```
ZakaryH, line 126: while I agree that it's odd to have a non promise here, and truthfully don't quite understand the use case for it - is it necessary to remove it as a possible type?

ZakaryH:
also tagging @chris-at-jobber on here since there are some UI changes with the validation/error messaging

packages/components/src/InputAvatar/InputAvatar.tsx
```diff
@@ -13,7 +13,7 @@ interface InputAvatarProps extends Omit<AvatarProps, "size"> {
    * More info is available at:
    * https://atlantis.getjobber.com/?path=/docs/components-forms-and-inputs-inputfile-docs--page#getuploadparams
    */
-  getUploadParams(file: File): UploadParams | Promise<UploadParams>;
```
ZakaryH, line 16: same thought here, is it possible to maintain the old behavior?

we haven't deprecated the type at all prior to this, so in the even someone was making use of it - it would start to fail going forward

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -166,10 +179,13 @@ export function InputFile({
   onUploadStart,
   onUploadProgress,
   onUploadComplete,
+  onUploadError,
+  validator,
 }: InputFileProps) {
   const options: DropzoneOptions = {
     multiple: allowMultiple,
-    onDrop: useCallback(handleDrop, [uploadFile]),
```
ZakaryH, line 172: this `uploadFile` dependency was a recent addition from another PR. was this removal intentional?

packages/components/src/InputAvatar/InputAvatar.tsx
ajpeng: I'm ok with keeping the existing behaviour. I will add back UploadParams as one of the valid return types.

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -180,7 +196,18 @@ export function InputFile({
...
   }
 
-  const { getRootProps, getInputProps, isDragActive } = useDropzone(options);
+  const { getRootProps, getInputProps, isDragActive, fileRejections } =
+    useDropzone(options);
+  const validationErrors = fileRejections?.map(({ file, errors }) => {
+    return errors.map(error => {
+      return (
+        <InputValidation
+          message={file.name + "\n" + error.message}
```
ZakaryH, line 205: not sure how I feel about adding a newline character as the message to achieve the layout we want

I'll circle back on this

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -207,6 +234,9 @@ export function InputFile({
               {hintText}
             </Typography>
           )}
+          {fileRejections?.length > 0 && (
+            <div className={styles.validationErrors}>{validationErrors}</div>
```
ZakaryH, line 238: curious what would happen here if `validationErrors` contained multiple elements

is it going to have a problem with not having a single fragment/parent?

packages/components/src/InputFile/InputFile.tsx
ajpeng: No that was unintentional. I missed this change during a merge

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -228,12 +258,17 @@ export function InputFile({
...
-      key = uuidv1(),
-      fields = {},
-      httpMethod = "POST",
-    } = await getUploadParams(file);
+    let params;
+
+    try {
+      params = await getUploadParams(file);
+    } catch {
+      onUploadError && onUploadError(new Error("Failed to get upload params"));
```
ZakaryH, line 266: thinking about if the error messages are OK to live in here or if they should be extracted either to a `messages.ts` or simply to a constant

ajpeng: I can extract them to a messages.ts

ZakaryH: before you do gimme one moment, I'm not seeing any other instances of `messages.ts` in Atlantis - so that actually might not be a pattern we're set up to use

ZakaryH: ok nevermind that's my bad we don't have that set up for Atlantis

I think a constant in this file will suffice. when we eventually do make such a change it'll be slightly easier to find and replace the user facing string

packages/components/src/InputFile/InputFile.tsx
ZakaryH: wait, it's not showing up as a newline. what is this doing? just adding a space?

image

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -139,6 +140,18 @@ interface InputFileProps {
...
+
+  /**
+   *  Upload event handler. Triggered on upload error.
+   */
+  onUploadError?(error: Error): void;
+
+  /**
+   * Pass a custom validator function that will be called when a file is dropped.
+   */
+  readonly validator?: <T extends File>(
```
ZakaryH, line 152: tempted to add an example validator to the docs so people can see how to implement their own

packages/components/src/InputFile/InputFile.tsx
ZakaryH: hmm so the layout is a bit iffy, I don't love how this shows up

image

also since we're doing

```
  const validationErrors = fileRejections?.map(({ file, errors }) => {
    return errors.map(error => {
      return (
        <InputValidation
          message={file.name + "\n" + error.message}
          key={file.name}
        />
      );
```

we get the multiple identical key console error

```
react-dom.development.js:86 Warning: Encountered two children with the same key, `Additional_Tools_for_Xcode_14.1.dmg`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -180,7 +196,18 @@ export function InputFile({
     options.accept = allowedTypes.join(",");
   }
 
-  const { getRootProps, getInputProps, isDragActive } = useDropzone(options);
+  const { getRootProps, getInputProps, isDragActive, fileRejections } =
```
ZakaryH, line 199: I'll preface this with the fact that I know I'm asking to add some more (design) scope, so feel free to push back

if we have the `fileRejections` and know if something didn't work out, I wonder if we should be adding a red border or something to the overall dropzone

definitely more of a @chris-at-jobber question/solution though

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -258,7 +293,12 @@ export function InputFile({
       file,
       handleUploadProgress,
     });
-    axios.request(axiosConfig).then(handleUploadComplete);
+    axios
+      .request(axiosConfig)
+      .then(handleUploadComplete)
+      .catch(() => {
+        onUploadError && onUploadError(new Error("Failed to upload file"));
```
ZakaryH, line 300: thinking out loud here, might not require any changes

we have the `validator` which will display messages when a file is rejected, and that's done inside this component

then we have the `onUploadError` that will trigger if the upload fails either from getting the upload params or if the actual request fails. however neither of these have any built in visual communication that something went wrong, it would be left up to each consumer how to communicate that - if at all

I feel conflicted about that. I'm not opposed to the consumer being informed and being able to do whatever makes sense for them, but I wonder about having a consistent baseline experience when the upload fails that at least lets the user know something went wrong

packages/components/src/InputFile/InputFile.tsx
ajpeng: image
I think it should be fine. This is an example of what can happen

packages/components/src/InputFile/InputFile.tsx
ZakaryH: or phrased another way, should visual error states be a concern of this component in the same way the validations are?

packages/components/src/InputFile/InputFile.tsx
ajpeng: I think it would look cleaner if it had `flex-direction:column; justify-content:flex-start`

image

ZakaryH: right but in your case you don't have two violations for a single file which is not that farfetched if we have say a type limit, a size limit and a quantity limit

you'd hit that non unique key problem

I'm curious why yours are stacking and mine aren't. is that just a screen size thing? if so, I think the one I ended up with is pretty not great/not that readable and we should do something to ensure they always stack.

ajpeng: I'm updating the css class to be 
```
.validationErrors {
  display: flex;
  justify-content: center;
  flex-direction: column;
}
```

packages/components/src/InputFile/InputFile.tsx
ajpeng: I think so I didn't dig into why that was the case though

packages/components/src/InputFile/InputFile.tsx
ajpeng: >> you'd hit that non unique key problem
I can update the key to be `{file.name}{error.code}` that should fix the unique issue

packages/components/src/InputFile/InputFile.tsx
ZakaryH: gotcha, alright if it's not doing the thing we'd expect it to do (newlining) - and we're ok with this result, meaning we don't need to find some other way to actually make it break to two lines - then let's remove the `"\n"` altogether and simply use a backticked string with some js vars to achieve the string format we want

packages/components/src/InputFile/InputFile.tsx
ajpeng: I think that's more of a design decision for @chris-at-jobber. One thing I'm thinking about is that when the upload params and request fail it's more of a system error. In the case of the user uploading the wrong file type, that is something more in their control. I can add props and methods to keep track of that those errors but I'm a bit worried that it might blow up in scope.

ZakaryH: agreed on both a design decision and scope adding

I'd say hold off on adding anything until we have a concrete decision of "yeah we need that". again this was some thinking out loud, and not a demand/request!

packages/components/src/InputFile/InputFile.tsx
ajpeng: I can add a section for validations

chris-at-jobber:
Could we do this @ajpeng @ZakaryH ? Follows the general pattern from our inputs of the validation appearing below + left-aligned.

[Figma](https://www.figma.com/file/rIIhulZvcp9M82lNOCGv16/branch/vOCZXzxhXtp2ND2fHvqunV/Product%2FOnline?type=design&node-id=12831%3A94146&mode=design&t=mPhdH0rCJR9ukpiu-1)

ajpeng:
> Could we do this @ajpeng @ZakaryH ? Follows the general pattern from our inputs of the validation appearing below + left-aligned.
> 
> [Figma](https://www.figma.com/file/rIIhulZvcp9M82lNOCGv16/branch/vOCZXzxhXtp2ND2fHvqunV/Product%2FOnline?type=design&node-id=12831%3A94146&mode=design&t=mPhdH0rCJR9ukpiu-1)

That works with me. I'll update the component

packages/components/src/InputFile/InputFile.css
```diff
@@ -22,3 +22,13 @@
   border-color: var(--color-focus);
   background-color: var(--color-surface--hover);
 }
+
+.dropZone.error {
+  border: var(--color-critical) dashed var(--border-thick);
```
chris-at-jobber, line 27: Only need to specificy border-color here:

```suggestion
  border-color: var(--color-critical);
```

chris-at-jobber (commented):
Design approved ✅ 

I did have one non-blocking comment about some CSS we can simplify and one question about another bit of CSS, but I can't figure out how to get the error state to show up in Storybook (maybe a good fast-follow would be to add a story for that?) so it's just a question at this point.

packages/components/src/InputFile/InputFile.css
```diff
@@ -22,3 +22,13 @@
...
+}
+
+.validationErrors {
+  display: flex;
+  justify-content: center;
```
chris-at-jobber, line 32: What is the `justify-content` doing here?

ajpeng: That doesn't need to be there anymore. It was used to get the validations to be in the center of the component.

ajpeng:
> Design approved ✅
> 
> I did have one non-blocking comment about some CSS we can simplify and one question about another bit of CSS, but I can't figure out how to get the error state to show up in Storybook (maybe a good fast-follow would be to add a story for that?) so it's just a question at this point.

I also had trouble getting this error state to show up in Storybook. What I did was I assigned a default method to the validator prop. 'm wondering if it was possible to do that in the canvas. I didn't see any examples of this in the codebase.

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -190,35 +217,41 @@ export function InputFile({
...
+          <Button
+            label={buttonLabel}
+            size={size}
+            type="secondary"
+            fullWidth={true}
+          />
+        )}
+      </div>
+      {fileRejections?.length > 0 && (
+        <div className={styles.validationErrors}>{validationErrors}</div>
```
ZakaryH, line 252: I was curious what the accessibility UX would be like here with the rejections in their own div

image

it's not great - you don't get the error message read out to you at all but that's because of the `InputValidation` component itself, not this implementation

ZakaryH: the screenshot is of the InputValidation component example and not the one you get on `InputFile` to demonstrate that it happens everywhere

docs/components/InputFile/InputFile.stories.mdx
```diff
@@ -58,3 +58,30 @@ updating an array of `FileUpload`s based on their `key` as uploads progress.
...
 export function updateFiles(updatedFile: FileUpload, files: FileUpload[]);
 ```
+
+### validator
+
+InputFile can accept a `validator` function as a prop. This will allow you to
+reject or accept file uploads based on your needs. The value of the function
+must accept a File and return `null` when it is accepted or an error object when
+rejected.
+
```
ZakaryH, line 68: tempted to mention this behavior or interaction with `allowedTypes`

image

I don't have a `validator` passed in, but I get the error message showing due to the `options.allow` only being set to images, and I've bypassed that with the sneaky little "show all options" in the native picker dialog

though I suppose these are slightly different?

the `allowedTypes` will attempt to restrict what you can select in that native dialog, and if you skirt it - it will show a rejection message based on that

VS the `validator` won't do anything to restrict what you can pick, it will only tell you after you've submitted if it was invalid with a custom message

that does mean you can end up with something like this though

image

ZakaryH: you know what, let's just call it out anyway

>Note that restricting accepted file types with `allowedTypes` will provide a default error message if an invalid file is selected.

And then for the example below, let's remove the PNG validation since it's a bit confusing to have multiple ways to validate a PNG. let's keep the example limited to validating on attributes other than the file type which you should probably be doing thru `allowedTypes`

ZakaryH (commented):
@ajpeng k one last little documentation update since there are now _almost_ 2 ways of doing the same thing

packages/components/src/InputFile/InputFile.tsx
chris-at-jobber: @ZakaryH how do we do it with InputText? I was pretty sure in that one we actually announce the error when you focus the input (I might be thinking of mobile specifically because I _know_ we solved it there).

ZakaryH: hmm I'm not 100% sure, I'd need to dig into it more

it seems like we're adding some kind of aria label or something to be read out on the input itself?

Kapture 2024-01-15 at 11 16 33

it's kinda hard to see but after I move off the input without a value, it reads out "you have to enter a name" without me tabbing to the error itself

once I tab to the error it says it's an image with no title and I can't get to the error message

docs/components/InputFile/InputFile.stories.mdx
ajpeng: that makes sense to me to me. I will update the example to just check for file size.

docs/components/InputFile/InputFile.stories.mdx
```diff
@@ -58,3 +58,33 @@ updating an array of `FileUpload`s based on their `key` as uploads progress.
...
+For example, the code below only accepts files that are less than 2MB in size.
+
+```ts
+function pngFileValidator(file: File) {
```
ZakaryH, line 72: this function name is still referencing the png validation, let's update that!

ajpeng: renamed

docs/components/InputFile/InputFile.stories.mdx
```diff
@@ -58,3 +58,33 @@ updating an array of `FileUpload`s based on their `key` as uploads progress.
...
+
+  return null;
+}
+```
+
+#### Notes
+
+If you would like to restrict specific file types, it is advised that you use
+the `allowedTypes` property instead of the `validator` property. Using
+`allowTypes` will allow the OS know what type of files the uploader is expecting
```
ZakaryH, line 88: `allowedTypes` rather than `allowTypes`

I think there's a word missing here

>will allow the OS know what

but you know what I think we can just leave it at 

>If you would like to restrict specific file types, it is advised that you use
the `allowedTypes` property instead of the `validator` property.

that gets at the main thing we want to communicate. describing all the behavior of the `allowedTypes` would be better suited either in its own section or on the prop JS docs

ZakaryH (approved):
LGTM

thank you for bearing with me through these last few nitpicks! really appreciate it

PR #1712, opened 2024-01-08
fix(components): static thumbnails for HEIC in all browsers but safari (fixes forever loading thumbnail for HEIC) [84773]

tyler-santos:
## Motivations

Currently when you upload an HEIC image to Jobber using any browser but safari, the InputFile component will not render a thumbnail but stay in a loading state forever. 

### Added

- Updates the basic icon for images to be image instead of camera
- Detects the browser being used and either serves a real thumbnail if it is safari or a static icon thumbnail if it is not safari
- Tests for the component as there were no specs previously
- A library for checking the client device _**Disclaimer**_ it is 100% possible to do this with the following code instead according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator):
```
function getBrowserName(userAgent) {
  // The order matters here, and this may report false positives for unlisted browsers.

  if (userAgent.includes("Firefox")) {
    // "Mozilla/5.0 (X11; Linux i686; rv:104.0) Gecko/20100101 Firefox/104.0"
    return "Mozilla Firefox";
  } else if (userAgent.includes("SamsungBrowser")) {
    // "Mozilla/5.0 (Linux; Android 9; SAMSUNG SM-G955F Build/PPR1.180610.011) AppleWebKit/537.36 (KHTML, like Gecko) SamsungBrowser/9.4 Chrome/67.0.3396.87 Mobile Safari/537.36"
    return "Samsung Internet";
  } else if (userAgent.includes("Opera") || userAgent.includes("OPR")) {
    // "Mozilla/5.0 (Macintosh; Intel Mac OS X 12_5_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36 OPR/90.0.4480.54"
    return "Opera";
  } else if (userAgent.includes("Edge")) {
    // "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36 Edge/16.16299"
    return "Microsoft Edge (Legacy)";
  } else if (userAgent.includes("Edg")) {
    // "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36 Edg/104.0.1293.70"
    return "Microsoft Edge (Chromium)";
  } else if (userAgent.includes("Chrome")) {
    // "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36"
    return "Google Chrome or Chromium";
  } else if (userAgent.includes("Safari")) {
    // "Mozilla/5.0 (iPhone; CPU iPhone OS 15_6_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.6 Mobile/15E148 Safari/604.1"
    return "Apple Safari";
  } else {
    return "unknown";
  }
}

const browserName = getBrowserName(navigator.userAgent);
console.log(`You are using: ${browserName}`);
```

_However_, I found this to be not so elegant, and you _**must**_ use the whole thing in that order or it breaks. Alternatively this library should always be up to date, is simple to use, and a small package size. **With that being said I dont know what auditing process we go through with atlantis, external libraries and security**

React Device Detect:

https://github.com/duskload/react-device-detect#readme

### Removed

- The camera icon for image

### Fixed

- Fixes the forever loading state for HEIC uploads

### Security

- Need to audit the React Device Detect library?  

### Why only safari?

Safari is currently the only browser that suppors HEIC.
Screenshot 2024-01-10 at 4 07 33 PM

## Testing

Test that images upload with a thumbnail in safari but chrome shows the static icon as shown below:

Old default icon:
Screenshot 2024-01-08 at 10 22 34 AM

New default icon and how a HEIC should look in any browser but safari:
Screenshot 2024-01-08 at 10 24 58 AM

In safari:
Screenshot 2024-01-10 at 4 05 52 PM
---

Random photo of Atlantis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37984,6 +37984,40 @@
         "react-dom": "^16.9.0 || ^17 || ^18"
       }
     },
+    "node_modules/react-device-detect": {
+      "version": "2.2.3",
+      "resolved": "https://registry.npmjs.org/react-device-detect/-/react-device-detect-2.2.3.tgz",
+      "integrity": "sha512-buYY3qrCnQVlIFHrC5UcUoAj7iANs/+srdkwsnNjI7anr3Tt7UY6MqNxtMLlr0tMBied0O49UZVK8XKs3ZIiPw==",
+      "dependencies": {
+        "ua-parser-js": "^1.0.33"
+      },
+      "peerDependencies": {
+        "react": ">= 0.14.0",
+        "react-dom": ">= 0.14.0"
+      }
+    },
+    "node_modules/react-device-detect/node_modules/ua-parser-js": {
+      "version": "1.0.37",
+      "resolved": "https://registry.npmjs.org/ua-parser-js/-/ua-parser-js-1.0.37.tgz",
+      "integrity": "sha512-bhTyI94tZofjo+Dn8SN6Zv8nBDvyXTymAdM3LDI/0IboIUwTu1rEhW7v2TfiVsoYWgkQ4kOVqnI8APUFbIQIFQ==",
+      "funding": [
+        {
+          "type": "opencollective",
+          "url": "https://opencollective.com/ua-parser-js"
+        },
+        {
+          "type": "paypal",
+          "url": "https://paypal.me/faisalman"
+        },
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/faisalman"
+        }
+      ],
+      "engines": {
+        "node": "*"
+      }
+    },
     "node_modules/react-devtools-core": {
       "version": "4.28.4",
       "resolved": "https://registry.npmjs.org/react-devtools-core/-/react-devtools-core-4.28.4.tgz",
@@ -45655,6 +45689,7 @@
         "lodash": "^4.17.21",
         "react-countdown": "^2.3.2",
         "react-datepicker": "~4.10.0",
+        "react-device-detect": "^2.2.3",
         "react-dropzone": "^11.0.2",
         "react-hook-form": "^7.43.7",
         "react-markdown": "^8.0.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,8 @@
       ],
       "dependencies": {
         "@std-proposal/temporal": "0.0.1",
-        "jest-environment-jsdom": "^29.5.0"
+        "jest-environment-jsdom": "^29.5.0",
+        "react-device-detect": "^2.2.3"
       },
       "devDependencies": {
         "@apollo/client": "^3.7.10",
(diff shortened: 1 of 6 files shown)


ZakaryH:
regarding the package vs MDN code snippet question, let's err on the side of avoiding packages when possible! while that code is a bit verbose due to having to go over each case in that order, I think it's still preferable to adding an entire package

this is because the package has more things we might not need but also because....

**Remix + making it safe for the server**

something we're going to be working on very soon is making our code safe for Remix's server side rendering

when rendering on the server, we won't have a browser/window/document to talk to which means we have to add guards around anything trying to access something like `window.navigator.userAgent` so that it won't have a bad time

all that is to say let's do the following
- use the MDN snippet
- let's add a util file in `packages/components/src/utils` 
- add a modification that will first check for `typeof document === "undefined"` before we try to access `navigator.userAgent` and let it fallback to the default on the server, when it renders on the client it'll get the right answer
- https://remix.run/docs/en/main/guides/gotchas#typeof-window-checks

ZakaryH:
>Updates the basic icon for images to be image instead of camera

do you happen to have a screenshot of this change? I'm going to try to get it myself in the meantime, but if you have it handy it'd be nice to document on the PR

ZakaryH (changes requested):
let's go with the MDN snippet + the Remix guards please and thank you

once we're on that I can give this a QA pass

also appreciate the tests!

tyler-santos:
@ZakaryH Thanks so much for reviewing this! I think I have addressed all of the feedback and created some extra utilities to get this to behave similarly to that library (functions for isSafari, isChrome, etc).

ZakaryH:
tried a few uploads on Chrome
- jpg
- svg
- HEIC

image

even though my file was `blah.HEIC` it was still correctly identified so good to know it's not case sensitive

also the thumbnail is correctly showing as the image icon rather than camera

then on Safari

image

HEIC thumbnail showing as intended

interesting that I am allowed to upload HEIC files in Safari without modifying the selection dialog options like I had to do in Chrome to upload that HEIC

Firefox

image

and Edge
image

ZakaryH:
tried making the `userAgent` undefined/empty string and that correctly falls back to the default! 🎉

image

chris-at-jobber (commented):
Design approved on the icon change, makes more sense tbh

ZakaryH (approved):
LGTM

PR #1713, opened 2024-01-09
refactor(components): Radius adjustments for retheme

chris-at-jobber:
## Motivations

When global tokens like radius change in the retheme, it exposes some un-needed radii being used in our components.

Why do these options need to be rounded?
image

### Changed

- AutoComplete option never needed a radius, now it has none
- StatusLabel is 1px rounder (otherwise it becomes a full on circle when larger radius tokens are applied)

## Testing

- View components in Storybook
- You can also go into the inspector and override the `--radius-base` property and see that nothing wild happens to the Autocomplete options, and the StatusLabel stays the same roundness.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/StatusLabel/StatusLabel.css
+++ b/packages/components/src/StatusLabel/StatusLabel.css
@@ -14,7 +14,7 @@
   width: var(--statusLabel-icon-diameter);
   height: var(--statusLabel-icon-diameter);
   margin-top: var(--space-minuscule);
-  border-radius: var(--radius-base);
+  border-radius: 3px;
   background-color: var(--color-success);
   flex-shrink: 0;
 }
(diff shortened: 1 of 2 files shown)


packages/components/src/StatusLabel/StatusLabel.css
```diff
@@ -14,7 +14,7 @@
   width: var(--statusLabel-icon-diameter);
   height: var(--statusLabel-icon-diameter);
```
fagnanm, line 15: @chris-at-jobber Can we also make this 10px instead of 12px so it matches SG-1?

fagnanm (changes requested):
@chris-at-jobber border radius change is good, but can we update the status label dot size while we are at it?

packages/components/src/StatusLabel/StatusLabel.css
chris-at-jobber: ✅

fagnanm (approved):
LGTM

PR #1718, opened 2024-01-10
fix: align center banner icon on retheme

darryltec:
## Motivations

Noticed the banner icons aren't aligned properly on the retheme styling

image

This fixes that

image

### Added

- align self center on the banner icon

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

add this on preview.tsx

```
document.body.classList.add("jobber-retheme");
```

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Banner/Banner.tsx
+++ b/packages/components/src/Banner/Banner.tsx
@@ -104,6 +104,7 @@ export function Banner({
 
       {dismissible && (
         <button
+          type="button"
           className={styles.closeButton}
           onClick={handleClose}
           aria-label="Close this notification"
--- a/packages/components/src/Banner/components/BannerIcon/BannerIcon.css
+++ b/packages/components/src/Banner/components/BannerIcon/BannerIcon.css
@@ -1,14 +1,3 @@
-.wrapper {
-  display: inline-flex;
-  padding: var(--space-smaller);
-  border-radius: var(--radius-circle);
-  background-color: var(--color-surface--background);
-}
-
-:global(.jobber-retheme) .wrapper {
-  align-self: center;
-}
-
 .success {
   background-color: var(--color-success);
 }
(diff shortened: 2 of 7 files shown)


packages/components/src/Banner/components/BannerIcon/BannerIcon.css
```diff
@@ -5,6 +5,10 @@
   background-color: var(--color-surface--background);
 }
 
+:global(.jobber-retheme) .wrapper {
```
chris-at-jobber, line 8: When the text wraps this leaves it centered in the banner, doesn't it? (definitely the easiest but I think we want to keep it aligned to that first row of text)

darryltec: gonna be a weird one tho because 1st line of text is also aligned self to the middle until that button goes down below

chris-at-jobber: I know it's messy but is there a way that we could change the alignment at different sizes? I know we manipulate other elements in Banner based on component width. Something like

```
.wrapper.medium {
  align-self: center;
}
```

? Or would that not work because this is technically a subcomponent?

darryltec: it would work if i move all the CSS into 1 file haha

packages/components/src/Banner/Banner.tsx
```diff
@@ -104,6 +104,7 @@ export function Banner({
 
       {dismissible && (
         <button
+          type="button"
```
chris-at-jobber, line 107: 😅

chris-at-jobber (approved):
Nice 👌

packages/components/src/Banner/Banner.tsx
darryltec: yeah, we have a linter now that reminds us to do that in case it goes in a form haha

PR #1720, opened 2024-01-11
fix(components): Fix DataList Small Breakpoint DataListHeader Visibility

ZakaryH:
## Motivations

The batch/bulk actions header isn't showing up on mobile when we would like it to

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

fix the header not showing when **all** selected. previously it would only show if you had more than zero selected, but less than **all** which is treated differently

now it will show even if you have all selected

**BEFORE**

image

**AFTOR**

image

## Testing

you'll need a DataList with everything including a DataListHeader plus batch setup, and then view it on a small resolution screen

this might be a lot to set up, I've been testing it on the new Client index page via `npm run pack @jobber/components` in Atlantis to make the local package that has this version and then installing that in Jobber and trying it out

you should be able to use the generated preview version from this PR though, I think

medium/large screen should remain the same since it is `visible` and thus won't meet the second `&&`'ed condition paired with this fix

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListHeader/DataListHeader.tsx
+++ b/packages/components/src/DataList/components/DataListHeader/DataListHeader.tsx
@@ -18,15 +18,14 @@ export function DataListHeader() {
     headers,
     layoutBreakpoints,
   } = useDataListContext();
-  const { selectedIDs } = useBatchSelect();
+  const { hasAtLeastOneSelected } = useBatchSelect();
 
   const size = getVisibleSize();
   const { layout } = useActiveLayout();
 
   const visible = headerVisibility[size];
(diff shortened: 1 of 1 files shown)


ZakaryH:
@chris-at-jobber here's the fix, I know you gave it a thumbs up in a thread already but this has a gif to show it off

chris-at-jobber (approved):
No more disappearing when you select all! Very good 👏 Pulled down and tested in Atlantis locally and in Cloudflare.

I'm comfortable reviewing the code change if you're OK with my approving it, nice and readable 📖

PR #1721, opened 2024-01-11
refactor(components): replace react-markdown component with markdown-to-jsx — (Minor) Feature Release

scotttjob:
## Motivations

We're replacing the `react-markdown` library with `markdown-to-jsx` to push Atlantis closer to being SSR compatible. Effort has been taken to configure `markdown-to-jsx` to look and act like `react-markdown` (which was the old library in use, and one that does not support SSR rendering)

## Changes

Refactored the <Markdown> component to use `markdown-to-jsx` under the hood instead of `react-markdown`

### Removed

- react-markdown dependency

### Added
- markdown-to-jsx dependency 

### Fixed

- SSR rendering will no longer break on this component.

## Testing

Render Markdown with the previous implementation and the new one. Compare the differences. We've done this ourselves via snapshots. We've only had to update two snapshots, and in both cases the only differences are spacing (and arguably markdown-to-jsx is more 'correct' in regards to the indentation & spacing it provides in the final result)

---

Random photo of Atlantis


scotttjob:
Closed this stream of work. We've decided to take another direction than updating the Markdown component wholesale.

PR #1723, opened 2024-01-11
feat(components): Add ability to have multiple sort options on DataList

taylorvnoj:
### The plan after this PR is merged:

1. This PR will be a breaking change since we are updating the API.  I will update the current uses in JO while I'm bumping the Atlantis version immediately after this merges. See a Draft PR here fixing breaking changes with the pre-release: https://github.com/GetJobber/Jobber/pull/42293
2. Update logic for headers with toggle sort; when headers with toggle sort get pushed into the Sort Combobox on small screens, the labels are hardcoded to say `(A-Z)` which obviously might not be the unit for categories like Last Activity which are numerical. 
3. Write new tests
4. Code clean up
5. Update documentation

## Motivations

Because `DataList` has the potential for columns to contain multiple types of data depending on the implementation, `DataList` needs a sorting pattern where one column can have multiple sorting options.

Consumers of the component need to be able to have:
1. Columns that have default asc/desc sort
6. Columns that have an array of options in a sorting list dropdown to sort against multiple different criteria

Kapture 2024-01-17 at 16 46 58

### Added

Created a subcomponent for the sorting dropdown called `DataListSortingOptions` since it consists of a `ul/li` which displays the additional sorting options.

When `DataList` is viewed on small screens, all sort labels are now pushed into the Sort `Combobox`

### Changed
 
`sortable` has been updated to become an array of objects, taking a `key` with optional `options`:

Before: `sortable: ["label", "home", "lastActivity"],`

After:
```
        sortable: [
          {
            key: "label",
            options: [
              { label: "First name (A-Z)", order: "asc" },
              { label: "First name (Z-A)", order: "desc" },
              { label: "Last name (A-Z)", order: "asc" },
              { label: "Last name (Z-A)", order: "desc" },
            ],
          },
          {
            key: "home",
            options: [
              { label: "Home world (A-Z)", order: "asc" },
              { label: "Home world (Z-A)", order: "desc" },
              { label: "Population (A-Z)", order: "asc" },
              { label: "Population (Z-A)", order: "desc" },
            ],
          },
          { key: "lastActivity" },
        ],
```

The backend logic for this sorting is of course up to the consumer of the component and they should put that logic in the `onSort` which takes a callback already.

## Testing

For now, take a look at the Storybook example.

Because the API list we are using in Storybook does not have sorting, we should probably consider putting this into ClientList and making sure sort still behaves as expected - [see here](https://github.com/GetJobber/Jobber/pull/42293)

---

Random photo of Atlantis


packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -1,23 +1,41 @@
+/* eslint-disable max-statements */
```
taylorvnoj, line 1: I'm at 14 but struggling to find an easy way to bring this down atm

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.css
```diff
@@ -1,5 +1,6 @@
 .headerLabel {
   display: flex;
+  position: relative;
```
taylorvnoj, line 3: This is keeping the sorting dropdown in the right spot. Otherwise you get this:

Screenshot 2024-01-17 at 5 17 22 PM

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
```diff
@@ -0,0 +1,44 @@
...
+  list-style: none;
+  cursor: pointer;
+  align-items: center;
+  justify-content: space-between;
+}
+
+.option:hover,
+.option:focus {
+  border-radius: var(--radius-large);
+  background-color: var(--color-surface--hover);
```
chris-at-jobber, line 35: - background-color can also move to the `:focus-visible` declaration

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
```diff
@@ -0,0 +1,44 @@
...
+}
+
+.option:focus-visible {
+  box-shadow: var(--shadow-focus);
```
chris-at-jobber, line 39: also need `outline: transparent` on this one

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
```diff
@@ -0,0 +1,44 @@
+.optionsList {
```
chris-at-jobber, line 1: There's a lot of shared styling between this menu and [DataListActionsMenu](https://github.com/GetJobber/atlantis/blob/7c2af0111b3ee141b35861872f7fdf7f1aca3c57/packages/components/src/DataList/components/DataListActionsMenu/DataListActionsMenu.css#L1-L28) - would it make sense to share the code where we have overlap?

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
chris-at-jobber: Actually never mind - because we're not using buttons, there's no default browser styling to override

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
chris-at-jobber: The border radius should just be set on the `.option`

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
```diff
@@ -0,0 +1,44 @@
...
+  box-shadow: var(--shadow-base);
+  margin-top: var(--space-smaller);
+  padding: var(--space-small) 0;
+  border: var(--border-base) solid var(--color-border);
+  border-radius: var(--radius-larger);
+  background: var(--color-surface);
+  align-items: flex-start;
+}
+
+.option {
```
chris-at-jobber, line 15: option styling is again quite similar to an existing element, this time [DataListAction](https://github.com/GetJobber/atlantis/blob/7c2af0111b3ee141b35861872f7fdf7f1aca3c57/packages/components/src/DataList/components/DataListAction/DataListAction.css#L1-L18)

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
```diff
@@ -0,0 +1,44 @@
...
+.option:focus {
+  border-radius: var(--radius-large);
+  background-color: var(--color-surface--hover);
+}
+
+.option:focus-visible {
+  box-shadow: var(--shadow-focus);
+}
+
+.optionSelected {
```
chris-at-jobber, line 42: Do we need a separate class for this? I don't _think_ there's any problem with always declaring the gap on the `option`

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
chris-at-jobber: Might even be identical to [ComboboxOption](https://github.com/GetJobber/atlantis/blob/7c2af0111b3ee141b35861872f7fdf7f1aca3c57/packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.css#L1-L14)

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
```diff
@@ -0,0 +1,44 @@
...
+  white-space: nowrap;
+  list-style: none;
+  cursor: pointer;
+  align-items: center;
+  justify-content: space-between;
+}
+
+.option:hover,
+.option:focus {
+  border-radius: var(--radius-large);
```
chris-at-jobber, line 34: As a favour can I please get the retheme radius specified on this one as well (example [here](https://github.com/GetJobber/atlantis/blob/7c2af0111b3ee141b35861872f7fdf7f1aca3c57/packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.css#L15-L18))

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
```diff
@@ -0,0 +1,44 @@
+.optionsList {
+  position: absolute;
+  top: 100%;
+  left: 0;
+  max-width: var(--popover--width);
+  box-shadow: var(--shadow-base);
+  margin-top: var(--space-smaller);
+  padding: var(--space-small) 0;
+  border: var(--border-base) solid var(--color-border);
+  border-radius: var(--radius-larger);
```
chris-at-jobber, line 10: Another ask for [retheme radius](https://github.com/GetJobber/atlantis/blob/7c2af0111b3ee141b35861872f7fdf7f1aca3c57/packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.css#L15-L18) to be specified independently

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
```diff
@@ -0,0 +1,44 @@
...
+  background: var(--color-surface);
+  align-items: flex-start;
+}
+
+.option {
+  display: flex;
+  min-width: calc(var(--space-largest) * 3.5);
+  min-height: calc((var(--space-largest) - var(--space-smaller)));
+  box-sizing: border-box;
+  margin: 0 var(--space-small);
```
chris-at-jobber, line 20: I've seen this in a few places now where we're specifying left and right margins on the options, and only top/bottom padding on the container. Is there a reason we're not just providing padding all around on the container?

Current
image
image

Just using padding on the parent
image

chris-at-jobber (commented):
Design approved ✅ 

I left some Non-blocking™ CSS suggestions and questions where I think we can clean up or simplify a bit

chris-at-jobber:
One accessibility note that I've come across before but have yet to document:

The sortable headers are announced as buttons ✅ but there's no indication of what those buttons will do (ie sort) 😢 or in the case of the new dropdown, that the button will render a menu 😢  

Similarly, the list items in the new multi-sort dropdown are announced to screen-readers ✅ but there's no indication that they make a selection, or which one _is_ selected once a selection is made 😢 

I'll make a ticket to resolve those issues separately if it's outside the scope of this work (and some of it is pre-existing so I imagine that is the case)

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.tsx
```diff
@@ -0,0 +1,77 @@
...
+interface DataListSortingOptionsProps {
+  readonly options: SortableOptions[];
+  readonly selectedOption: SortableOptions | null;
+  readonly onSelectChange: (selectedOption: SortableOptions) => void;
+  readonly onClose: () => void;
+  readonly optionsListRef: React.RefObject<HTMLUListElement>;
+  readonly dataListHeaderTileRef: React.RefObject<HTMLElement>;
+}
+
+export function DataListSortingOptions({
```
ericchernuka, line 17: Just wondering if there's a reason this couldn't just be a generic dropdown for re-use? Most of this is what you'd get in a headless dropdown and could then be used across all components that need dropdown functionality.

I'm including this as an example, not as a suggestion, but Radix gives this all as primitives that you then style/configure to the needs. https://www.radix-ui.com/primitives/docs/components/dropdown-menu

I worry that every time we need the same functionality that looks slightly differently, we're re-implementing the behaviour.

taylorvnoj: Hey @ericchernuka ! This is a great talking point and question.  Apologies in advance for a bit of a long winded response but I feel that some context is relevant.

Part of our discovery in September when building `Combobox` was to look at our existing selection type components (`MultiSelect`, `Select`, `Menu` etc) and evaluate if any of those could be used to build up our `ComboboxContent` - we have the components already so why not? However, the look and functionality that `Combobox` needed, meant the best path forward was to build our own dropdown within Combobox.  

In quick succession we then were faced with building an actions dropdown in `DataList` and now this sorting dropdown - where yes, you have a 'trigger' with a dropdown menu (the purpose of `Combobox`).  The snag there is that a `Combobox` trigger is either an Atlantis `Chip` or `Button`.  Given that `DataListSortingOptions` is triggered by `Text`, right now we cannot use `Combobox`.  

tldr: Yes, Atlantis needs to evaluate our current selection components, gather each of their requirements and then build one reusable dropdown. Unfortunately that is not something we have capacity for immediately however it's on the horizon.

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
taylorvnoj: You're right! Deleted `optionSelected` and moved gap to `option` here: c4c6e3b43c9a2539695caff930c3983508b83a84

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
taylorvnoj: We moved `border-radius` to option here: ee1423098f7ad2d4250e87076edf50478938b89c and set the retheme radius here: cd2b3022ed9844e16ec123391e6a68a12dd6380b

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
taylorvnoj: Done here: cd2b3022ed9844e16ec123391e6a68a12dd6380b

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.css
taylorvnoj: When you say share the code, do you mean have 1 file that merges `optionsList`/`menu` into one variable?  

It makes sense, but I wonder what that would look like in the code.  For example, I'd probably want to put that css in `DataList.css` to easily share between both files.  Let me know if I'm way off base here but...

I wonder two things:
1. Would we have to move the other variables from `DataListActionsMenu` and `DataListSortingOptions` into `DataList.css` and then delete those other css files and rely on `DataList.css`?  I think the answer is yes, which again I'm ok with - just wondering if we have a restriction on how long css files are.
2. When it comes to the case of `DataListSortingOptions` and `ComboboxOption` having identical `options` css -  in this case, I wonder where we should put the shared css. I don't _think_ we have a component that reaches outside of it's own file to grab css from another component (calling `ComboboxOption` inside `DataList` feels strange?).

I think what this comes down to is the conversation Eric and I are having below - we are now building the same UI for different selection components (that all have slightly different functionality) and eventually we'll need to have a component that fits the needs of `ComboboxContent`/`Sorting`/`ActionsMenu`, etc.  Let me know your thoughts or if I took this in a completely different direction then you intended 😆

taylorvnoj:
Thank you for making that accessibility ticket @chris-at-jobber! There will likely be changes to this PR on eng side so I think it's a good idea to work on the accessibility once this is shipped.

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -38,24 +67,41 @@ export function DataListHeaderTile<T extends DataListObject>({
...
+    if (sortableItem?.options) {
+      setIsDropDownOpen(!isDropDownOpen);
+    } else {
+      toggleSorting(headerKey);
+    }
+  }
+
+  function handleSelectChange(selectedOption: SortableOptions) {
+    if (sortableItem) {
+      toggleSorting(sortableItem.key, selectedOption.order);
```
rodrigoeidelvein, line 101: I think we should add the `sortableItem.label` to this call, too. For example, in the current DataList story, if you select "First name (A-Z)" or "Last name (A-Z)" the output of `sorting` will be the same: `{key: 'label', order: 'desc'}`.

rodrigoeidelvein (changes requested):
still taking a look, but I think we can get started on those. We can pair to discuss it if you want, too. let me know

packages/components/src/DataList/DataList.types.ts
```diff
@@ -49,6 +49,12 @@ export type DataListHeader<T extends DataListObject> = {
 export interface DataListSorting {
   readonly key: string;
   readonly order: "asc" | "desc";
+  readonly label: string | undefined;
+}
+
+export interface SortableOptions {
+  readonly label: string;
+  readonly order: "asc" | "desc";
```
taylorvnoj, line 57: This is on my radar

packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
```diff
@@ -44,17 +45,30 @@ export function DataListSort() {
...
 
+        acc.push({
+          label: `${label} (A-Z)`,
+          value: `${sort.key},asc,${label}`,
+        });
+        acc.push({
+          label: `${label} (Z-A)`,
+          value: `${sort.key},desc,${label}`,
+        });
+
```
taylorvnoj, line 71: After chatting with Chris and Darryl yesterday,  we have plans for a new way to handle these cases - where there is no dropdown custom sort but when the DataList is in mobile web, all sorting labels go to the Sort Combobox.

This will be addressed in a later PR (within this sprint)

packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
```diff
@@ -44,15 +44,21 @@ export function DataListSort() {
   function getSortByOptions() {
     const options = sortable.reduce(
       (acc: Record<"label" | "value", string>[], sort) => {
-        const label = headers[sort];
+        const label = headers[sort.key];
         if (!label) return acc;
 
         if (label === "Last activity") {
```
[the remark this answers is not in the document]
taylorvnoj: That was done here 8d28274470e494cf158f75a3183715965e3b1a01

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
taylorvnoj: Also done here 8d28274470e494cf158f75a3183715965e3b1a01

packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
rodrigoeidelvein: I don't know if I understood how it will be done, but I trust you!

but maybe before cleaning it up, we should update the implementation on JO because we are introducing breaking changes, right?

rodrigoeidelvein (approved):
LGTM!

packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
taylorvnoj: Absolutely - I'm going to merge this and then update JO and fix the breaking changes asap

PR #1725, opened 2024-01-12
fix(hooks): Guard useAssert to work with SSR

rodrigoeidelvein:
## Motivations

Currently, we are checking the `process.env` directly, which is not available at Remix by default. This change adds the optional chaining to check if `process.env` is available.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- `useAssert` can be used in Remix

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/hooks/src/useAssert/useAssert.tsx
+++ b/packages/hooks/src/useAssert/useAssert.tsx
@@ -9,7 +9,7 @@ export function useAssert(
   message: string,
   options?: Options,
 ) {
-  if (process.env.NODE_ENV !== "production" && shouldShow) {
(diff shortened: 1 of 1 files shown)


scotttjob (approved):
LGTM!

Double checked with Eric that the question mark based guarding works just fine in 2024. My SSR knowledge is a bit out of date so while `typeof process !== 'undefined'`still works just fine and is still the way to do it in ES5 based environments, it's a lot more verbose than just a `?` question mark so that's going to be our preferred way to guard in 2024.

Great work!

I have an extremely minor comment that the second question mark is technically not necessary (if process is defined, env is most likely going to be as well), but I like the idea of being extra careful here. :)

PR #1726, opened 2024-01-13
feat(components): more robust empty state for "truly" empty DataList

chris-at-jobber:
## Motivations

For cases where the DataList is truly empty, we should have a more engaging empty state. We've tested a pattern that has proven more effective for this use case, so we should bake it into the DataList as a sensible default.

This is my first stab, primarily to gather feedback on approach and solicit ideas from engineers.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.css
+++ b/packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.css
@@ -47,6 +47,10 @@
   background-color: var(--color-surface--background);
 }
 
+.fakeDataCell:nth-last-of-type(-n + 2) {
+  display: none;
+}
+
 @media only screen and (min-width: 768px) {
   .fakeDataRow {
     grid-template-columns: 1fr 1fr 1fr 1fr;
--- a/packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.css
+++ b/packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.css
@@ -1,5 +1,6 @@
 .emptyStateWrapper {
   display: flex;
+  position: relative;
   box-sizing: border-box;
   padding: var(--space-base);
   justify-content: center;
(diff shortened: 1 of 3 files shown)


packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.css
```diff
@@ -8,3 +9,50 @@
   align-items: center;
   flex: 1;
 }
+
+.emptyStateWrapper h3 {
+  text-align: center;
+}
+
+.fakeData {
```
chris-at-jobber, line 17: Referencing the DOM/CSS from existing empty states in Jobber. I think we can probably move what I have in the TSX out to a subcomponent or something - I don't love that I have to get into negative z-indexing here to get the fake data beneath the message and action, but I think the alternative would be adding more wrappers around those and absolutely positioning them.

packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.css
```diff
@@ -8,3 +9,50 @@
...
+  opacity: 0.2;
+}
+
+.fakeDataCell {
+  height: var(--data-list--minimum-item-height);
+  border-radius: var(--radius-base);
+  background-color: var(--color-surface--background);
+}
+
+@media only screen and (min-width: 768px) {
```
chris-at-jobber, line 50: Use Atlantis breakpoints

packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.tsx
```diff
@@ -20,12 +21,41 @@ export function InternalDataListEmptyState() {
...
+  if (filtered) {
+    return (
+      <div className={styles.emptyStateWrapper}>
+        <Text align="center">{message}</Text>
+        {renderButton(action)}
+      </div>
+    );
+  } else {
+    return (
+      <div className={styles.emptyStateWrapper}>
```
chris-at-jobber, line 33: As noted elsewhere, this could probably be moved out to its own component

packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.css
```diff
@@ -8,3 +9,50 @@
   align-items: center;
   flex: 1;
 }
+
+.emptyStateWrapper h3 {
+  text-align: center;
```
chris-at-jobber, line 14: Couldn't figure out how to expose the `align` from `Typography` on `Heading` so here we are.

packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.tsx
chris-at-jobber: Alternatively, could these actually live in the columns of the `DataList.Layout` so you consistently get a more representative empty state?

packages/components/src/DataList/components/DataListEmptyState/DataListEmptyState.tsx
```diff
@@ -20,12 +21,41 @@ export function InternalDataListEmptyState() {
...
+            <div className={styles.fakeDataCell}></div>
+          </div>
+          <div className={styles.fakeDataRow}>
+            <div className={styles.fakeDataCell}></div>
+            <div className={styles.fakeDataCell}></div>
+            <div className={styles.fakeDataCell}></div>
+            <div className={styles.fakeDataCell}></div>
+          </div>
+        </div>
+        <Heading level={3}>{message}</Heading>
```
chris-at-jobber, line 54: Question - because Heading accepts children, could `message` include <Emphasis>?

PR #1727, opened 2024-01-16
feat(components): onSearch Callback Prop for Combobox

ZakaryH:
## Motivations

we have a known use case where we want to load data as it is searched for rather than having it all in the list up front. it's a tremendous amount of options where you simply wouldn't want to scroll thru them all, nor select them all

with that, we need to expose the ability for a consumer to hook into the search input's value as it changes, to receive that value and go grab whatever data makes sense for that. once they have the data, they can simply iterate over it as the new `Combobox.Option`s

this does present some notable differences from the current experience where we are able to make many more assumptions about the state and behavior of things, so some additional props were needed 

### Added
- onSearch prop, allowing consumer to receive the search term as it changes (debounced)
- a debounce prop to configure the debounce time on said callback, defaults to 300ms
- loading prop to manually display a glimmer while loading, only useful for onSearch flow where we are likely fetching additional data with some sort of API request
- added the corresponding Glimmer and some simple styles to give it some spacing

**Consumer Responsibilites**
when using `onSearch` we stop doing a handful of things for you. you are now responsible for implementing the search, filtering, fetching data, setting/unsetting loading states etc.

the consumer is also responsible handling the clear search, which can be done with the clear button or simply deleting, and the search term is `""` again - at which point it would be up to the consumer to either restore a cached initial list of options, or perhaps simply fire a query again with no search terms to fetch what would hopefully be the original data

### Changed

- moved some scroll code into a hook
- adjusted options logic & var naming, moving it up and simply passing it down to the sub components who don't really care if they are filtered options or the initial options - we can make the same decisions without knowing that detail
- fixed/updated some tests
- if you are searching and have no results, we simply show the `No results for {blah}` no matter what. if you had no options initially or not makes no difference. Search message > Empty message as far as priority now.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

I have an example in the webstory for Combobox now, you can check that out or make your own if you want. it's a bit verbose but simple example of how you might implement this yourself and what would be needed

- existing tests should pass
- debounce should work on the passed in `onSearch` meaning if you do "fsadfadsfs" really fast, you'd only receive one call for that and not however many characters that is
- the non `onSearch` flow should be unchanged

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -464,6 +464,7 @@ describe("Combobox Custom onSearch", () => {
 
     await user.type(screen.getByPlaceholderText("Search"), "Val");
     jest.advanceTimersByTime(300);
+    expect(mockOnSearch).toHaveBeenLastCalledWith("Val");
     await user.click(screen.getByTestId("ATL-Combobox-Content-Search-Clear"));
     jest.advanceTimersByTime(300);
 
@@ -491,9 +492,6 @@ describe("Combobox Custom onSearch", () => {
 
     expect(screen.queryAllByTestId("ATL-Glimmer")).toHaveLength(0);
   });
-
-  // implement me once bug is fixed
-  // it("should show the correct header when searching and no results found", () => {});
 
   it("should show the correct message when searching, and no options present", async () => {
     renderCustomOnSearchCombobox(false, true);
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { render, screen } from "@testing-library/react";
-import { userEvent } from "@testing-library/user-event";
+import { UserEvent, userEvent } from "@testing-library/user-event";
 import { Combobox } from "./Combobox";
 import { ComboboxOption } from "./Combobox.types";
 import { COMBOBOX_TRIGGER_COUNT_ERROR_MESSAGE } from "./hooks/useComboboxValidation";
@@ -18,6 +18,9 @@ const handleAction = jest.fn();
 const handleSelect = jest.fn();
 const mockSelectedValue = jest.fn<ComboboxOption[], []>().mockReturnValue([]);
 const mockMultiSelectValue = jest.fn().mockReturnValue(false);
+const mockOnSearch = jest.fn();
+
+let user: UserEvent;
 
 afterEach(() => {
   handleAction.mockClear();
@@ -430,6 +433,122 @@ describe("Combobox Compound Component Validation", () => {
   });
 });
 
+describe("Combobox Custom onSearch", () => {
+  beforeEach(() => {
+    jest.useFakeTimers();
+    user = userEvent.setup({
+      advanceTimers: jest.advanceTimersByTime,
+    });
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+    mockOnSearch.mockClear();
+  });
+  it("should only call the debounced onSearch one, with the correct value", async () => {
+    renderCustomOnSearchCombobox(false, true);
+
+    await user.type(screen.getByPlaceholderText("Search"), "V");
+    await user.type(screen.getByPlaceholderText("Search"), "a");
+    await user.type(screen.getByPlaceholderText("Search"), "l");
+
+    jest.advanceTimersByTime(200);
+    expect(mockOnSearch).not.toHaveBeenCalled();
+    jest.advanceTimersByTime(300);
+    expect(mockOnSearch).toHaveBeenCalledTimes(1);
+    expect(mockOnSearch).toHaveBeenCalledWith("Val");
+  });
+
+  it("should call the debounced onSearch with an empty string when cleared with the clear button", async () => {
+    renderCustomOnSearchCombobox(false, true);
+
+    await user.type(screen.getByPlaceholderText("Search"), "Val");
+    jest.advanceTimersByTime(300);
+    await user.click(screen.getByTestId("ATL-Combobox-Content-Search-Clear"));
+    jest.advanceTimersByTime(300);
+
+    expect(mockOnSearch).toHaveBeenLastCalledWith("");
+  });
+
+  it("should not have option filtering behavior out of the box like the non custom onSearch version", async () => {
+    renderCustomOnSearchCombobox(false, true);
+
+    await user.type(screen.getByPlaceholderText("Search"), "Value 1");
+    jest.advanceTimersByTime(300);
+
+    expect(screen.queryByText("API Value 1")).toBeInTheDocument();
+    expect(screen.queryByText("API Value 2")).toBeInTheDocument();
+  });
+
+  it("should show the correct amount of loading glimmers when loading is true", () => {
+    renderCustomOnSearchCombobox(true, true);
+
+    expect(screen.getAllByTestId("ATL-Glimmer")).toHaveLength(5);
+  });
+
+  it("should not show the loading glimmers when loading is false", () => {
+    renderCustomOnSearchCombobox(false, true);
+
+    expect(screen.queryAllByTestId("ATL-Glimmer")).toHaveLength(0);
+  });
+
+  // implement me once bug is fixed
+  // it("should show the correct header when searching and no results found", () => {});
+
+  it("should show the correct message when searching, no results found and opitions existed", async () => {
+    renderCustomOnSearchCombobox(false, true, true);
+
+    await user.type(screen.getByPlaceholderText("Search"), "Value 4");
+    jest.advanceTimersByTime(300);
+
+    // heads up these quotes are tricky and cause the test to not pass if they don't match exactly
+    expect(screen.getByText("No results for “Value 4”")).toBeInTheDocument();
+  });
+
+  it("should show the correct message when no options provided and searching", async () => {
+    renderCustomOnSearchCombobox(false, false);
+
+    await user.type(screen.getByPlaceholderText("Search"), "Value 1");
+    jest.advanceTimersByTime(300);
+
+    expect(screen.getByText("No options yet")).toBeInTheDocument();
+  });
+
+  it("should show the correct message when no options provided and not searching", () => {
+    renderCustomOnSearchCombobox(false, false, true);
+
+    expect(screen.getByText("No options yet")).toBeInTheDocument();
+  });
+});
+
+function renderCustomOnSearchCombobox(
+  loading: boolean,
+  hadInitalOptions: boolean,
+  renderWithoutOptions = false,
+) {
+  const options = renderWithoutOptions
+    ? []
+    : [
+        { id: "1", label: "API Value 1" },
+        { id: "2", label: "API Value 2" },
+      ];
+
+  return render(
+    <Combobox
+      multiSelect
+      selected={mockSelectedValue()}
+      onSelect={handleSelect}
+      onSearch={mockOnSearch}
+      loading={loading}
+      hadInitalOptions={hadInitalOptions}
+    >
+      {options.map(option => (
+        <Combobox.Option id={option.id} label={option.label} key={option.id} />
+      ))}
+    </Combobox>,
+  );
+}
+
 function renderCombobox() {
   return render(
     <Combobox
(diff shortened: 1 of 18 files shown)


packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -437,6 +437,7 @@ function renderCombobox() {
       multiSelect={mockMultiSelectValue()}
       selected={mockSelectedValue()}
       onSelect={handleSelect}
+      debounce={0}
```
ZakaryH, line 440: debatable if this is the best way to handle this or not

it does make tests happy and avoid the need to do any kind of timer manipulation because otherwise tests will make their assertions too soon

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -74,7 +78,8 @@ export function Combobox(props: ComboboxProps): JSX.Element {
           wrapperRef={wrapperRef}
           open={open}
           setOpen={setOpen}
-          options={options}
+          options={filteredOptions}
```
ZakaryH, line 81: decided that the sub components really don't need to know or care that the options they are receiving are filtered or not

just render them if you get them

opted to add one more prop of `hadInitialOptions` since there is a different flow we want to happen if you're searching and got nothing as a result of filtering VS if you never had anything to begin with

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.test.tsx
```diff
@@ -83,6 +92,7 @@ function renderComboboxContent(
       shouldScroll={{ current: false }}
       open={open}
       selected={selected}
+      searchValue=""
```
ZakaryH, line 95: hmm I probably should have made this `searchValue={searchValue}` rather than hardcoding it to an empty string

it doesn't really matter, the tests still pass, but it is silly

packages/components/src/Combobox/hooks/useCombobox.ts
```diff
@@ -14,18 +22,51 @@ type UseComboboxReturn = {
...
   onClose?: () => void,
   multiSelect?: boolean,
+  onSearchChange?: (searchValue: string) => void,
+  debounceTime: number = 300,
 ): UseComboboxReturn {
   const wrapperRef = useRef<HTMLDivElement>(null);
   const shouldScroll = useRef<boolean>(false);
   const [open, setOpen] = useState<boolean>(false);
   const [searchValue, setSearchValue] = useState<string>("");
+  const [filteredOptions, setFilteredOptions] =
```
ZakaryH, line 41: `filteredOptions` is a bit of a weird name. it fits really well when you are searching locally

but if you never search, or you provided an `onSearchChange` then it's possibly a misnomer. it might just be some options that aren't filtered, we can't really know - nor do we particularly care

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContentList.tsx
```diff
@@ -69,6 +71,11 @@ export function ComboboxContentList(props: ComboboxListProps): JSX.Element {
                 />
               );
             })}
+          {props.loading && (
+            <div className={styles.loadingContainer}>
+              <Glimmer size="larger" />
```
ZakaryH, line 76: not sure how many of these to display, we won't know how many results to expect until it's done and then you don't need a glimmer

we could show a glimmer for each of the options we had before you started searching? we might get back more or less after the search though, tough to say!

packages/components/src/Combobox/hooks/useCombobox.ts
ZakaryH: I don't think I knew it when I made this comment but my intuition was vaguely correct

this gets weird and causes a bug when you're not using the internal filtering

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContentList.tsx
```diff
@@ -1,49 +1,21 @@
...
+import {
+  ComboboxListProps,
+  ComboboxOption as ComboboxOptionType,
+} from "../../../Combobox.types";
 import { ComboboxOption } from "../../ComboboxOption/ComboboxOption";
 
 export function ComboboxContentList(props: ComboboxListProps): JSX.Element {
-  const [listScrollState, setlistScrollState] = useState("");
-
-  useEffect(() => {
```
ZakaryH, line 11: moving all this into a local hook isn't necessary

I was in here and felt like the effect hook was taking up a considerable amount of line real estate in this component and wanted to tidy that up a bit

packages/components/src/Combobox/hooks/useCombobox.ts
```diff
@@ -14,18 +21,44 @@ type UseComboboxReturn = {
...
+  );
+
+  const debouncedFilterOptions = useCallback(
+    debounce((value: string) => {
+      const filtered = options.filter(option => {
+        return option.label.toLowerCase().includes(value.toLowerCase());
+      });
+
+      setInternalFilteredOptions(filtered);
+    }, debounceTime),
```
ZakaryH, line 59: is it really worth debouncing the internal version?

filtering a local set of data isn't very expensive, and there are certain interactions that feel laggy or unresponsive as a result, for example it'll delay a moment before updating the filtered list but since we immediately update the input value and that's tied to the "no results" message you can have a moment where it'll say "no results for ''" rather than immediately going back to the unfiltered state

of course it's possible to fix that, it just means we need to either debounce everything that happens as a result of typing in the input, or we move everything into a single debounced handler

ZakaryH: Kapture 2024-01-17 at 08 36 58

packages/components/src/Combobox/Combobox.test.tsx
darryltec: can use fake timers (`jest.useFakeTimers()`) and `await jest.runAllTimersAsync();` to advance it past the debounce

ZakaryH:
TODO: write some tests for the new functionality

I can see in my gif that there's a bug where the "select" is showing up in the header when I have no search results. that should not be there.

so test for things like that

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContentList.tsx
chris-at-jobber: > a glimmer for each of the options we had before you started searching

This is my preference if we can. Playing with it right now, the wrapper collapsing when it has multiple options feels a bit shaky.

Also, how does it look if we use `Glimmer.Text` here? (just trying to approximate the elements we're replacing as closely as possible)

packages/components/src/Combobox/hooks/useCombobox.ts
chris-at-jobber: If we can avoid doing loading (or loading-esque, which is how I think of debouncing) delays in scenarios where they're not needed, it should feel much better. _(Is `scenarios where they're not needed` hard to know?)_

packages/components/src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContentList.tsx
ZakaryH: Kapture 2024-01-17 at 13 23 13

here's what it looks like with the `Glimmer.Text`

hmm ok I can see about having a glimmer for each of the initial options. trying to think through if there's the potential for that going sideways where you have 0 initial options, and then search + load and that could be strange

if I can't get that working, is there a static number you'd be happy with?

ZakaryH: image

chris-at-jobber: Nice, thank you 🙏

chris-at-jobber:
@ZakaryH I'm not entirely sure what the solution is here but it feels like the results are in a "loading" state too often/frequently? As I type, I am guessing it's related to the debounce timing, but you can see it basically blinks in and out of the loading state repeatedly which feels not great

https://github.com/GetJobber/atlantis/assets/39704901/7dc3c1fb-98e6-4e34-b2ce-c2376f1c9633

ZakaryH:
@chris-at-jobber 

>^ Also - is my search query modifying the option labels?

haha for this example yes it is. that's just how I put together the example, I thought it would feel more natural if the results being returned actually did match what you queried for in the search - so to achieve that I'm simply tacking on whatever you searched for onto the static set of first names

not a real thing, just a this-particular-webstory-example thing!

ZakaryH:
>I'm not entirely sure what the solution is here but it feels like the results are in a "loading" state too often/frequently? As I type, I am guessing it's related to the debounce timing, but you can see it basically blinks in and out of the loading state repeatedly which feels not great

yeah we could bump up the default debounce timing to be much higher and more forgiving, that way we'll be pretty darn sure you're done typing before we trigger the search callback.it's also possible to customize the debounce to tailor it to whatever makes the most sense for a given usage

ZakaryH:
**Search > Empty State**

the last commit does a few things that "fix" the bug as a side effect

previously if you had no options in your list and were searching we made a distinction between if that was because you had no options to begin with, if that was the case you'd always see

`No options yet`

compared to if you had no options visible because of a search, you'd see

`No results for {blah}`

now, we're just going to say ok if you are searching - we don't care if you had options or not originally, if you are searching and we received 0 options then we're going to show the `No results for` message. 

this simplifies the logic quite a bit, removes the need for the `hadInitialOptions` prop, removes some states to test and a couple other nice little benefits

I've also killed the `showEmptyState` prop on the `ComboboxContentList` component since we're already receiving the `options` from which we can derive that same information seeing as all it was was `options.length > 0`. it was actually being used in some redundant ways where it would be doing checks like `condition && condition` where if the first one was true, the second must also be true given they were different ways of asking the exact same question 🙃

packages/components/src/Combobox/hooks/useCombobox.ts
ZakaryH: stopped debouncing the internal callback to search/filter since it only led to a feeling of unresponsiveness and it's not at all necessary

packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -34,10 +34,25 @@ export interface ComboboxProps {
...
+
   /**
    * The Chip heading for the trigger
    */
   readonly label?: string;
+
+  /**
+   * Should the Combobox display the loading state, only needed when providing custom onSearch that has async loading behavior.
+   */
+  readonly loading?: boolean;
```
darryltec, line 55: > only needed when providing custom onSearch that has async loading behavior.

I think this shouldn't be tied to onSearch as initially loading is a valid case. Meaning, there is a query behind the first load of options and it is loading.

Real example in JO is it's loading the initial tags options in clients index.

darryltec (approved):
Looks good to me. I have 2 ask tho and one of them is important.

Important ask: Don't restrict loading to only be for onSearch as it is valid for cases where the options initially loads.

Not so important ask: To get as close as possible to what the devs would write for the search example, we can use [Rick and Morty GQL Test API](https://studio.apollographql.com/public/rick-and-morty-a3b90u/variant/current/explorer) which has a search filter on it.

<details>
  <summary>
    Click here to find out
  </summary>

```ts
import { ApolloClient, InMemoryCache, gql } from "@apollo/client";

export interface ListQueryType {
  characters: {
    results: {
      name: string;
    }[];
  };
}

export const LIST_QUERY = gql`
  query ListQuery($filter: FilterCharacter) {
    characters(filter: $filter) {
      results {
        name
      }
    }
  }
`;

export const apolloClient = new ApolloClient({
  uri: "https://rickandmortyapi.com/graphql",
  cache: new InMemoryCache(),
});
```

```tsx
const ComboboxCustomSearch: ComponentStory<typeof Combobox> = args => {
  const [selected, setSelected] = useState<ComboboxOption[]>([]);
  const [searchTerm, setSearchTerm] = useState("");
  const { data, loading } = useQuery<ListQueryType>(LIST_QUERY, {
    variables: {
      filter: { name: searchTerm },
    },
    fetchPolicy: "network-only",
    nextFetchPolicy: "cache-first",
    client: apolloClient,
  });

  return (
    <Combobox
      {...args}
      onSelect={setSelected}
      selected={selected}
      label="Teammates"
      multiSelect
      onSearch={setSearchTerm}
      loading={loading}
    >
      {data?.characters.results?.map(({ name }) => (
        <Combobox.Option key={name} id={name} label={name} />
      ))}
    </Combobox>
  );
};
```

</details>

Both could be a follow up as it works for what it is intended to do.

PR #1730, opened 2024-01-18
feat(components): Expose all sizing tokens in web Content

taylorvnoj:
## Motivations

Let's support all sizing tokens on Content, for flexibility when using the component

### Added

Exposed `minuscule`, `smallest`, `smaller`, `larger`, `largest` and `extravagant` as props on Content `spacing` prop

## Testing

Make sure all options on `spacing` prop changes the vertical spacing between children in Content story.

---

Random photo of Atlantis


taylorvnoj:
@chris-at-jobber my only question is - we also have `Paddings.css` in the Content folder but it seemingly is doing nothing? It's not being called anywhere. Unless you have additional context, I'm thinking of just deleting that paddings file

chris-at-jobber:
@taylorvnoj that's an old @darryltec file from the looks of it - there _are_ some non-obvious internal bits to Content so there's a chance it _could_ be doing something, somewhere. I'd leave it for now unless we're 100% confident it doesn't do anything? My first guess was it was at least intended to allow for Content's padding to be controlled (the same way we do on the mobile component), but I don't think we ever exposed that ability 🤔

chris-at-jobber (approved):
This looks and works great! Approved, pending what we want to do with that possibly useless `Padding` file

chris-at-jobber:
Also, semantically I'd argue this PR is a `feat()` PR as it does introduce new functionality.

darryltec:
Doesn't seem to be used anywhere. If you delete it and it still bootstraps, you should be good

PR #1732, opened 2024-01-23
docs(components): Improve Combobox Custom Search Example

ZakaryH:
## Motivations

per some feedback on the initial `onSearch` PR, it would be preferable to have more realistic code as an example for how to implement a custom search with the `onSearch` prop talking to an api to fetch results based on the query

## Changes

swapped out the scuffed code to use Apollo stuff to talk to a Richard and Mortimer character API

make a lil util file with the necessary code because if you have it in the webstory bad things happen, plus it's tidier this way

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

just play around with it! see that you have some options loaded initially

search for some characters not available in the initial set of data, see that it fetches them

undo the search term, get your initial options back!

I will say it's very hard to see the loading state now. I'm not sure this setup is actually ever setting loading to be `true` long enough to be perceptible...

it is definitely toggling the value, but it's difficult to see the loading state at all

image

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -1,12 +1,11 @@
 import React, { useState } from "react";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
-import { useQuery } from "@apollo/client";
 import { Combobox, ComboboxOption } from "@jobber/components/Combobox";
 import { Button } from "@jobber/components/Button";
 import { Typography } from "@jobber/components/Typography";
 import { Chip } from "@jobber/components/Chip";
 import { Icon } from "@jobber/components/Icon";
-import { LIST_QUERY, ListQueryType, apolloClient } from "./storyUtils";
+import { useFakeQuery } from "./storyUtils";
 
 export default {
   title: "Components/Selections/Combobox/Web",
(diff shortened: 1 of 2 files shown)


darryltec (changes requested):
Looks good! Would need to inject the files to the code sandbox for it to work there like this 

https://github.com/GetJobber/atlantis/blob/03135427c5c110b2922e22328fa0469dcede77a1/docs/components/Chips/Web.stories.tsx#L123-L126

and then import it

https://github.com/GetJobber/atlantis/blob/03135427c5c110b2922e22328fa0469dcede77a1/docs/components/Chips/Web.stories.tsx#L120-L122

darryltec:
as for testing the loading state, can test it with throttling your internet through dev tools

ZakaryH:
updated and imported, it is working on the code tab now

for my own knowledge, since I copied this pattern from the DataList - I'm assuming we have no need to do this `extraImport` stuff there since the code tab is hidden for DataList?

darryltec:
> for my own knowledge, since I copied this pattern from the DataList - I'm assuming we have no need to do this extraImport stuff there since the code tab is hidden for DataList?

yup. for now

darryltec (approved):
🚢

PR #1733, opened 2024-01-24
feat(components): Animated presence

darryltec:
## Motivations

Make it easy to add an enter/exit animation on any element you want be it a conditional element or in a map/list.

### Basic/Conditional 

[Live demo](https://animated-presence.atlantis.pages.dev/?path=/story/components-utilities-animatedpresence-web--basic)

Jan-26-2024 16-15-15

### List

[Live demo](https://animated-presence.atlantis.pages.dev/?path=/story/components-utilities-animatedpresence-web--list)

Jan-26-2024 16-16-12

### Stepper

[Live demo](https://animated-presence.atlantis.pages.dev/?path=/story/components-utilities-animatedpresence-web--stepper)

Jan-26-2024 16-16-42

### Added

- Added an animated presence component to handle enter/exit animations on components

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


packages/components/src/AnimatedPresence/AnimatedPresence.tsx
```diff
@@ -0,0 +1,111 @@
...
+   * unexpected behavior.
+   */
+  readonly exitBeforeEnter?: boolean;
+}
+
+export function AnimatedPresence(props: AnimatedPresenceProps) {
+  const shouldReduceMotion = useReducedMotion();
+
+  if (shouldReduceMotion) {
+    return <>{props.children}</>;
```
scotttjob, line 56: I like this because you know it's going to be the first question asked about this feature "How do I turn it off?" And the response is "we respect browser settings" which is perfect. Great stuff.

docs/components/AnimatedPresence/Web.stories.tsx
```diff
@@ -0,0 +1,198 @@
...
+export default {
+  title: "Components/Utilities/AnimatedPresence/Web",
+  component: AnimatedPresence,
+  parameters: {
+    viewMode: "story",
+    previewTabs: { code: { hidden: false } },
+  },
+} as ComponentMeta<typeof AnimatedPresence>;
+
+const BasicTemplate: ComponentStory<typeof AnimatedPresence> = args => {
```
scotttjob, line 25: Overall this looks great, I have very little feedback. That said, there's something in the storybook stories that I want to call out. It might only exist there, but I'm not 100% sure. Might in the real code too.

So it looks like on the Basic story (might exist elsewhere too) there's a margin separating each item. And it looks like that margin is not part of the animation.

This is a nitpick, but I've spent so many years looking at animations that it jumped out to me.

Because the margin is not animated, there appears to be a bit of a 'jump' at the end of the transition. I took a video of it in motion and broke it down to two frames to show the jump happening. The first frame is two frames back from the jump, where you can see sort of see the text as it's taken out via opacity/height. Then the second frame shows right before the margin is removed (the div is taken out of the DOM) and then the final frame where it's out.

We can chat on a call if you want to see it better. :)

image
image
image

scotttjob (changes requested):

docs/components/AnimatedPresence/Web.stories.tsx
darryltec: yeah, noticed that too. Unfortunately, the solution for that might just be allowing `AnimatedPresence` to set a spacing instead of a wrapping `Content` since framer needs to know about those values

packages/components/src/AnimatedPresence/AnimatedPresence.tsx
darryltec: yeah, also a good example of componentizing too because i could've just gone and do it in 1 component but doing it this way saves the client side a bit more logic and useEffect since those would be rendered too even if you don't want them. This way it's not gonna get to that code at all.

chris-at-jobber: Should add an accessibility section to the docs (under the template it would be an h2 near the bottom) to note that it respects `prefers-reduced-motion`, as I think designers will probably ask about this more than anyone

darryltec: oh shit you right!

docs/components/AnimatedPresence/Web.stories.tsx
darryltec: ok, latest version of framer motion actually fixes that and we're due for an update anyway so I'll do that first haha

darryltec: Fixed now!

scotttjob (approved):
LGTM! Nice utility component, and a good way to get consistent (easy) motion into an app.

A piece of follow-up feedback I could see us getting (but we can deal with it if it comes up, no need to slow down this great PR any further) is a way to speed up a list style animation with more than a handful of initial items. If you bump it up to 20 or more the wait for the animation to finish becomes noticeable. 

It might be enough to just provide a handle to generateDelayTime so a consumer can provide their own function, but again a change for the future.

Great work! :)

PR #1734, opened 2024-01-26
feat(components): Add data-value to sortable options

taylorvnoj:
## Motivations

Adding an `id` to the `li` of our sortable options means that mapping sorting logic from the backend to the `DataList` component will:
1. be more straightforward
2. avoid mapping against the label string (ex "Last Name (A-Z)") which is extremely brittle and if that string changes, good luck tracking down why sort is broken 😅 

## Changes

`SortableOptions` type now has `id`:

```
export interface SortableOptions {
  readonly id: string;
  readonly label: string;
  readonly order: "asc" | "desc";
}
```

### Added

Everyone gets an `id` 

## Testing desktop screen size:

- [ ] Each custom sort `option` selected will return an `id`

NOTE: <s> Toggling on Last Activity will NOT return an `id` unless you on a small screen. </s> JUST KIDDING I did [this](https://github.com/GetJobber/atlantis/pull/1734/files#diff-9ee4883bf9251e8d0c7ebf791241f261dd3abfe95e9fd3eb337cdd4b7da70c47R106) so now toggling on Last Activity will return the `headerKey` itself. 

Kapture 2024-01-25 at 20 11 59

## Testing mobile screen size:

- [ ] Each `option` selected in the `Combobox` will return a `id`

`id` needed to be pushed into `DataListSort` as well.  Because the `sortable` `options` are now being pushed into a `Combobox`, we needed to add `id` to the `value`  The reason we have an `id` for Last Activity here, even when the Storybook props do not have an `options` array, it because we are mapping `id` to be `sort.key` for now. Again, when each `sortable` header has it's own `options` (in a future ticket this sprint), we won't need to push these `key` only values to the `Combobox`, because `key` only sort values will no longer exist. If that doesn't make sense, let me know I will try to explain better in person.

Kapture 2024-01-25 at 20 24 37

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.types.ts
+++ b/packages/components/src/DataList/DataList.types.ts
@@ -51,9 +51,9 @@ export interface DataListSorting extends SortableOptions {
 }
 
 export interface SortableOptions {
+  readonly id: string;
   readonly label: string;
   readonly order: "asc" | "desc";
-  readonly dataValue: string;
 }
 
 export interface DataListProps<T extends DataListObject> {
--- a/packages/components/src/DataList/DataList.types.ts
+++ b/packages/components/src/DataList/DataList.types.ts
@@ -53,6 +53,7 @@ export interface DataListSorting extends SortableOptions {
 export interface SortableOptions {
   readonly label: string;
   readonly order: "asc" | "desc";
+  readonly dataValue: string;
 }
 
 export interface DataListProps<T extends DataListObject> {
--- a/packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
+++ b/packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
@@ -18,7 +18,7 @@ export function DataListSort() {
       onSelect={selection => handleKeyChange(selection[0].id.toString())}
       selected={[
         {
-          id: `${state?.key},${state?.order},${state?.label}`,
+          id: `${state?.key},${state?.order},${state?.label},${state?.dataValue}`,
           label: state?.order || "",
         },
       ]}
@@ -53,20 +53,19 @@ export function DataListSort() {
           customOptions.forEach(option => {
             acc.push({
               label: option.label || "",
-              value: `${sort.key},${option.order},${option.label}`,
+              value: `${sort.key},${option.order},${option.label},${option.dataValue}`,
             });
           });
 
           return acc;
         }
-
         acc.push({
           label: `${label} (A-Z)`,
-          value: `${sort.key},asc,${label}`,
+          value: `${sort.key},asc,${label},${sort.key}`,
         });
         acc.push({
           label: `${label} (Z-A)`,
-          value: `${sort.key},desc,${label}`,
+          value: `${sort.key},desc,${label},${sort.key}`,
         });
 
         return acc;
@@ -83,7 +82,8 @@ export function DataListSort() {
   function getButtonLabel() {
     const selectedOption = sortByOptions.find(
       option =>
-        option.value === `${state?.key},${state?.order},${state?.label}`,
+        option.value ===
+        `${state?.key},${state?.order},${state?.label},${state?.dataValue}`,
     );
 
     return selectedOption?.label || "";
@@ -91,8 +91,8 @@ export function DataListSort() {
 
   function handleKeyChange(value?: string) {
     if (value && value !== "none") {
-      const [key, order, label] = value.split(",");
-      onSort({ key, order: order as "asc" | "desc", label });
+      const [key, order, label, dataValue] = value.split(",");
+      onSort({ key, order: order as "asc" | "desc", label, dataValue });
 
       return;
     }
--- a/packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
+++ b/packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
@@ -13,14 +13,12 @@ export function DataListSort() {
 
   const sortByOptions = getSortByOptions();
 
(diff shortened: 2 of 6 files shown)


packages/components/src/DataList/DataList.types.ts
```diff
@@ -53,6 +53,7 @@ export interface DataListSorting extends SortableOptions {
 export interface SortableOptions {
   readonly label: string;
   readonly order: "asc" | "desc";
+  readonly dataValue: string;
```
darryltec, line 56: i wonder if this should also just be called key... or id

packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
```diff
@@ -13,12 +13,14 @@ export function DataListSort() {
 
   const sortByOptions = getSortByOptions();
 
+  const selectedSortID = `${state?.key},${state?.order},${state?.label},${state?.dataValue}`;
```
darryltec, line 16: ooh what happens if someone puts a comma on their label?

packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
```diff
@@ -53,20 +55,19 @@ export function DataListSort() {
           customOptions.forEach(option => {
             acc.push({
               label: option.label || "",
-              value: `${sort.key},${option.order},${option.label}`,
+              value: `${sort.key},${option.order},${option.label},${option.dataValue}`,
```
darryltec, line 58: maybe put this in a function so you can have 1 source of truth

darryltec (commented):
early comments, testing it now

packages/components/src/DataList/DataList.types.ts
taylorvnoj: I'm down to call it `id`. IMO it we shouldn't call it `key` since we already have a `key` in the same object

Screenshot 2024-01-26 at 5 12 05 PM

darryltec: id it is!

packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
taylorvnoj: much sad happens 😢 

Screenshot 2024-01-26 at 5 32 33 PM

It's also broken on master if you put a comma in there - the selection doesn't stick and we loose the full label 😭 

Screenshot 2024-01-26 at 5 36 28 PM

darryltec: i wonder if you can use an emoji 😆

taylorvnoj: I could `JSON.stringify` everything and that does seem to fix it 🤔  Am I going in the right direction?

darryltec: could be. could also look at that sortID and see if the label needs to be there. You have an extra ID on it now, looks like you could just use the ID 🤔

taylorvnoj: Okay I fixed the comma issue here c315056c2f8d04ffe2ef9a821c51a2504f623ddd 

I'm pretty sure label needs to be there so it goes back into the Combobox chip once a selection is made but if I'm misunderstanding what you're thinking, let me know.

packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.tsx
taylorvnoj: I've split out some logic where we were setting these values in multiple places - but this file needs a good clean up so I'll put more effort into that in my cleanup PR

taylorvnoj:
@darryltec fixed the state issue like we talked about here b2e25cb1d5b9b5b393cae39a1ebbff1c944abc61 and did some renaming in the `handleSelectChange` - ready for another look when you have a moment!

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
```diff
@@ -40,6 +38,10 @@ export function DataListHeaderTile<T extends DataListObject>({
 
   const Tag = isSortable ? "button" : "div";
 
+  const selectedOption = sorting?.state
+    ? ({ ...sorting?.state } as SortableOptions)
+    : null;
```
darryltec, line 43: don't need to spread it and can actually take advantage of the `||` operator here. 

```suggestion
  const selectedOption: SortableOptions | null = sorting?.state || null;
```

adding the type instead of casting it is also safer

darryltec (approved):
Just noticed 2 minor things. Can fix it now or on your cleanup. Up to you, it doesn't really change how this works haha.

Also, I'm guessing you have another ticket for documentation?

packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx
taylorvnoj: Fair - just pushed this change

packages/components/src/DataList/DataList.types.ts
```diff
@@ -51,6 +51,7 @@ export interface DataListSorting extends SortableOptions {
 }
 
 export interface SortableOptions {
+  readonly id: string;
```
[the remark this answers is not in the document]
taylorvnoj: Sounds good - I'll do this in clean up and yes, I've got a documentation subtask 👍

PR #1736, opened 2024-01-26
fix(components): Updating InputTime to remove InputTimeSafari for SSR compatibility (and because we didn't need it anymore)

scotttjob:
## Motivations

We're moving to support SSR rendering with Atlantis, and we had a polyfill in use here that was blocking rendering. It turns out the InputTimeSafari package is no longer needed, which is where the polyfill was in use (to check for the state where Safari needs special attention).

So we removed InputTimeSafari, which meant we could remove the polyfill, which means this component now renders fine in an SSR environment.

Testing seems good across the board.

## Changes

- Removed InputTimeSafari and the associated supportsTime polyfill.

### Removed

- InputTimeSafari

### Fixed

- SSR Support for InputTime

## Testing

InputTime component should render well in all browsers (including Safari)

---

Random photo of Atlantis


darryltec (approved):
LGTM. Still works the same on 
- modern browsers
- works on safari 14.1 and up
- with the autofill, it became useable on 14 and down tho usage is still below 0 so not a big deal image

PR #1737, opened 2024-01-26
fix(components): Removing uuid in web and replacing with useId

scotttjob:
## Motivations

We have an upcoming need to support SSR, and our usage of uuid does not play well with that kind of rendering. We need the ids to match from server to client.

## Changes

- Removed uuid usage in internal components. Should not affect external functionality.

## Testing

- All form elements or anything previously using uuid, should render and operate as normal.

---

Random photo of Atlantis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45649,7 +45649,6 @@
         "@types/react-datepicker": "^4.8.0",
         "@types/react-router": "5.1.7",
         "@types/react-router-dom": "^5.3.3",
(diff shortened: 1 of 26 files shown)


dlrussel (approved):
Checked for errors in jobber online and jobber frontend. This looks good to me but I also made commits to the branch so I will make sure it gets one more approval.

rodrigoeidelvein (approved):
LGTM!

PR #1740, opened 2024-01-30
fix(components): Add SortableOptions Partial to DataListSorting

taylorvnoj:
### Changed

`DataListSorting` type needs to have a Partial extend for `SortableOptions` in order to continue allowing for sorting with or without options (`id`, `label` & `order`)

## Testing

Darryl and I have been working on implementing changes needed [here](https://github.com/GetJobber/Jobber/pull/42605)

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.types.ts
+++ b/packages/components/src/DataList/DataList.types.ts
@@ -51,9 +51,9 @@ export interface DataListSorting extends SortableOptions {
 }
 
 export interface SortableOptions {
-  readonly id: string | undefined;
-  readonly label: string | undefined;
-  readonly order: "asc" | "desc";
+  readonly id?: string;
+  readonly label?: string;
+  readonly order?: "asc" | "desc";
 }
 
 export interface DataListProps<T extends DataListObject> {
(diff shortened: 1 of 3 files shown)


packages/components/src/DataList/DataList.types.ts
```diff
@@ -51,9 +51,9 @@ export interface DataListSorting extends SortableOptions {
 }
 
 export interface SortableOptions {
-  readonly id: string;
-  readonly label: string;
-  readonly order: "asc" | "desc";
+  readonly id?: string;
+  readonly label?: string;
+  readonly order?: "asc" | "desc";
```
darryltec, line 56: you need all 3 no?

also, it's already optional 
image

taylorvnoj: But once I put it in JO, the setSort is telling me I'm missing `id` and `label`

taylorvnoj: Resolving because we avoided this by adding the partial

darryltec (approved):

PR #1743, opened 2024-01-30
fix(components): Update handling of sorting state

taylorvnoj:
## Motivations

Last Activity header was not sorting in the correct order

### Added
 
- added `setSorting` function to better handle sorting state updates; and share that function between `toggleSorting` as well as `handleSelectChange` (sorting on key only vs. sorting with options)

### Changed

- updated the check for order being `undefined` in `toggleSorting`
- update variable names 

## Testing

Update sorting state in the web stories to make sure that on load, Last activity header has desc sort indicated. Another click on the header returns no sorting state. Clicking a third time means the sorting state is asc.

```
  const [sortingState, setSortingState] = useState<DataListSorting>({
    key: "lastActivity",
    order: "desc",
  });
```

---

Random photo of Atlantis


darryltec (approved):
:shipit:

PR #1750, opened 2024-02-05
fix(hooks): support add and remove listener on mockViewportWidth

darryltec:
## Motivations

Some packages, like framer-motion, supports the older match media function of add and remove event listener. That's all good, but we mock the match media when mocking the viewport width without the add and remove event listener. That would make other tests that include framer motion fail.

### Added

- support for deprecated add and remove event listener

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


taylorvnoj (approved):
LGTM

PR #1753, opened 2024-02-08
feat(components-native): JOB-88274 Prep inputs for retheme

MichaelParadis:
## Motivations

Updating the base styling of `components-native`'s Input to match our new designs

### Added

- Added focused state to InputPressable

### Changed

- Changed theming of Inputs to prepare for the retheme
   - Added full borders around our Inputs
   - Updated the size of the Inputs to match the new designs.
   - Added padding to Inputs to match the new designs

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing
 - Use the Jobber Mobile builds [here](https://github.com/GetJobber/jobber-mobile/pull/8688) and verify the styling of Inputs matches the designs

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/Select/Select.tsx
+++ b/packages/components-native/src/Select/Select.tsx
@@ -151,15 +151,13 @@ export function Select({
           <View
             style={[styles.container, (invalid || !!error) && styles.invalid]}
           >
-            {label && (
-              <Text
-                level="textSupporting"
-                variation={textVariation}
-                hideFromScreenReader={true}
-              >
-                {label}
-              </Text>
-            )}
+            <Text
+              level="textSupporting"
+              variation={textVariation}
+              hideFromScreenReader={true}
+            >
+              {label}
+            </Text>
 
             <View style={styles.input}>
               <View style={styles.value}>
--- a/packages/components-native/src/Select/Select.tsx
+++ b/packages/components-native/src/Select/Select.tsx
@@ -131,7 +131,7 @@ export function Select({
       invalid={invalid || !!error}
       hasValue={hasValue}
       styleOverride={{
-        container: { borderBottomWidth: undefined },
+        container: { paddingLeft: undefined },
       }}
     >
       <View
--- a/packages/components-native/src/InputDate/InputDate.tsx
+++ b/packages/components-native/src/InputDate/InputDate.tsx
@@ -174,6 +174,7 @@ function InternalInputDate({
   return (
     <>
       <InputPressable
+        focused={showPicker}
         clearable={canClearDate}
         disabled={disabled}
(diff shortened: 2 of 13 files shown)


taylorvnoj:
I'm looking through Storybook before I look at the code and found something curious.  InputTime disabled has 4px border radius top right and left, while the bottom is 2px

Screenshot 2024-02-13 at 5 03 10 PM

taylorvnoj:
> I'm looking through Storybook before I look at the code and found something curious. InputTime disabled has 4px border radius top right and left, while the bottom is 2px
> 
> Screenshot 2024-02-13 at 5 03 10 PM

Looks like it's on each disabled version of any input 🤔

taylorvnoj:
We'll have to update this to avoid the 4px borders on disabled state:

https://github.com/GetJobber/atlantis/pull/1753/files#diff-8581ea025e6443fa75d988e94b7ef53130ec45442ad3c3886d19c0a5397919e3R47-R51

packages/components-native/src/Select/Select.tsx
```diff
@@ -151,15 +151,13 @@ export function Select({
...
-                {label}
-              </Text>
-            )}
+            <Text
+              level="textSupporting"
+              variation={textVariation}
+              hideFromScreenReader={true}
+            >
+              {label}
+            </Text>
```
taylorvnoj, line 160: Screenshot 2024-02-13 at 10 52 32 PM

I think we need to adjust a `flexDirection` somewhere - Selects are looking a bit wonky 🤔 

If I remove the `flexDirection: row` it looks better.  And I'm wondering if we're able to vertically center that chevron a bit more, to match... `InputPassword` for example has that eye icon vertically center

Screenshot 2024-02-13 at 10 54 19 PM

taylorvnoj (changes requested):
Thanks for making that disabled state change!

I looked quickly at the build and so far things look good - enough for me to approve changes here and then do a more in depth look in the mobile PR.

Other than that - just one more thing I noticed with the `Select` component.

packages/components-native/src/Select/Select.tsx
MichaelParadis: The issues with Select are largely due to them having been made before InputPressable so it has weird styles from before. I think we should focus more on how it actually renders in the mobile device. Especially because the existing Select for mobile renders incorrectly as well
image

taylorvnoj: Ahh okay that's fair - Select's look good on my device so let's leave it at that!

taylorvnoj (approved):
Great work! LGTM

PR #1754, opened 2024-02-08
fix: ensure loading spinner shows up on combobox when actions are there

darryltec:
## Motivations

Noticed that when you set the loading to true, the spinner doesn't show up when an action is available. It's because it's being rendered under behind the actions 
image

This fixes it

image

### Added

- 

### Changed

- from fixed to sticky so it stays the the bottom but keeps itself visible when scrolled

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContentList.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContentList.tsx
@@ -3,6 +3,7 @@ import classnames from "classnames";
 import { Text } from "@jobber/components/Text";
 import { Glimmer } from "@jobber/components/Glimmer";
 import { Spinner } from "@jobber/components/Spinner";
+import { AnimatedPresence } from "@jobber/components/AnimatedPresence";
 import styles from "./ComboboxContentList.css";
 import {
   ComboboxListProps,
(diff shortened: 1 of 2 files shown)


packages/components/src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContentList.tsx
```diff
@@ -43,16 +44,15 @@ export function ComboboxContentList(props: ComboboxListProps): JSX.Element {
...
-            </div>
-          )}
+          <div
+            className={classnames([styles.loadingContainer, styles.hasOptions])}
+          >
+            <AnimatedPresence transition="fromBottom">
+              {Boolean(props.loading && optionsExist) && (
+                <Spinner size="small" />
+              )}
+            </AnimatedPresence>
```
darryltec, line 54: also added some pizzazz when it shows and hides. It slides from the bottom

mattander (approved):
LGTM in QA.

Makes sense to me as well. Behaviour is the same since the user has to scroll bottom to trigger anyways. 

The only case covered by `fixed` but not by this is the `onSearch` loading will be hidden. But I don't think that's worth blocking on.

MichaelParadis (approved):
LGTM!

PR #1756, opened 2024-02-12
fix(components): clearing search term in Combobox after clicking on clear button 

gregorijobber:
## Motivations
There was no way of the Combobox clear search term while closing it

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- When closing the Combobox it will clear the search term as well

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -46,6 +46,8 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
           selectedCount={props.selected.length}
           onClearAll={() => {
             props.selectedStateSetter([]);
+            props.setSearchValue("");
+            props.handleSearchChange("");
           }}
           onSelectAll={() => {
             props.selectedStateSetter(props.options);
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -46,8 +46,6 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
           selectedCount={props.selected.length}
           onClearAll={() => {
             props.selectedStateSetter([]);
-            props.setSearchValue("");
-            props.handleSearchChange("");
           }}
           onSelectAll={() => {
             props.selectedStateSetter(props.options);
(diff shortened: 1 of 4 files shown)


packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -46,6 +46,8 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
           selectedCount={props.selected.length}
           onClearAll={() => {
             props.selectedStateSetter([]);
+            props.setSearchValue("");
+            props.handleSearchChange("");
```
ZakaryH, line 50: hey sorry I wasn't part of the slack thread on this but I don't agree with these changes

the clear all button should be able to be used based on the context provided by the search term, but it shouldn't be able to change it

the select/clear all is more about the selected state and what items we have, not the current search term

ZakaryH:
this looks great. my only request is could we please add a simple test case for this?

probably in the `describe("Combobox Custom onSearch", () => {` block

a test that asserts our `onSearch` was called with the arg we expect when the Combobox is closed

gregorijobber:
> this looks great. my only request is could we please add a simple test case for this?
> 
> probably in the `describe("Combobox Custom onSearch", () => {` block
> 
> a test that asserts our `onSearch` was called with the arg we expect when the Combobox is closed

I've added the test in the onClose context as from my understandment this is a callback from the onClose. Lemme know if you think this is the wrong place!

ZakaryH:
>I've added the test in the onClose context as from my understandment this is a callback from the onClose. Lemme know if you think this is the wrong place!

so the `onClose` context is more about the `onClose` prop which is its own separate callback. my first thought was to put the test in there too actually so I see where you're coming from

I think the `onSearch` block of tests is just a bit better since that's closer to the behavior we're looking to test in this scenario. that when we provide an `onSearch`, it gets called at all the right times. specifically when we close the Combobox.

I don't think it's a big deal though, and I do have a small cleanup PR related to Combobox in progress. if I really want to move it, I might just do it then. I don't want to hold this PR up on such a small thing, it is providing a test case for the scenario we want!

ZakaryH:
tested this out with a custom `onSearch`

verified that I am getting `""` as an arg when it closes via
- ESC press
- clicking outside the open Combobox area

whereas before the internal search would clear but we were not informing the `onSearch` of this update 

thanks for fixing this!

ZakaryH (approved):
LGTM

PR #1759, opened 2024-02-14
refactor(components): CJS -> ESM exploration

scotttjob:
## Motivations

Vite is happier with ESM builds than it is with CJS builds, plus this is something we've been meaning to do for a while just don't have the time (but now I'm here!)

## Changes

Attempts to convert Atlantis from a CJS build to an ESM build (as well as jobber/design).

### Added

- Nothing new, except for converting CJS code to ESM code.

### Changed

- Hopefully nothing!

### Deprecated

- CJS build

### Removed

- CJS build

### Fixed

- ESM build

### Security

- Hopefully nothing!

## Testing

Make sure that this branch works when a builder is expecting an ESM build.

---

Random photo of Atlantis


kingstonfung:
An observation: Components are producing index files that `require()` from `./dist/` folder, which causes "cannot resolve" problems when brought into another place. Perhaps just some strings / folder management stuffs, taking away the `./dist/` path makes the component consumer-env happy again.

For example:

```
Over at /node_modules/@jobber/components/Table.js


... 
var Table = require("./dist/Table");
...
✘ [ERROR] Could not resolve "./dist/Table"
```

I am taking a look into this as well to see if I can provide any help during my available time.

Thanks!

kingstonfung:
(me dropping + sharing notes)

What I found so far:
- I noticed we have this file: scripts/entryPoints.ts that would generate the entrypoints for each components
- We can likely tweak these entrypoint generations to see if we can thread the needle, to get the linkage properly
- If successful, then (maybe?) we could look further into the repurposing this script, or simply avoid more aliasing if we just bring the components one level up

More poking to come...... that's it for now.

scotttjob:
Closing this PR for now, as we're not going to do the work right away and it'll be easier to pick up the changes fresh off main when we're ready.

PR #1764, opened 2024-02-15
fix(components-native): Apply input transform in InputText before trim [OPS-20241]

davidr64:
## Motivations

InputNumber was providing string values when text ended with a space.

## Changes

InputText applies the transform before treating the form field value as a string.

## Testing

Use InputNumber finish the number with a space like "13 " and see that InputNumber returns the number 13 instead of the string "13" **after blur**

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/InputText/InputText.test.tsx
+++ b/packages/components-native/src/InputText/InputText.test.tsx
@@ -582,8 +582,6 @@ describe("Transform", () => {
       const form = useForm();
       useEffect(() => {
         return form.watch((value, { name, type }) => {
-          console.log(value, name, type);
-          console.log("value", value);
           onFormValueUpdate(value);
         }).unsubscribe;
       }, [form]);
--- a/packages/components-native/src/InputText/InputText.test.tsx
+++ b/packages/components-native/src/InputText/InputText.test.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import {
   RenderAPI,
   fireEvent,
@@ -6,6 +6,7 @@ import {
   waitFor,
 } from "@testing-library/react-native";
 import { Platform, TextStyle } from "react-native";
+import { FormProvider, useForm } from "react-hook-form";
 import { InputText, InputTextProps } from "./InputText";
 import { InputAccessoriesProvider } from "./context";
 import {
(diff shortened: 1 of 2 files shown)


packages/components-native/src/InputText/InputText.tsx
```diff
@@ -444,25 +437,33 @@ function InputTextInternal(
...
+  }
 }
 
 function trimWhitespace(
-  field: ControllerRenderProps<FieldValues, string>,
-  onChangeText?: (newValue: string) => void,
+  inputValue: string | undefined,
+  onChangeText: (newValue: string) => void,
 ) {
-  if (!field.value || !field.value.trim) {
```
davidr64, line 453: Previously field.value could have been a transformed value (most likely a number).

packages/components-native/src/InputText/InputText.test.tsx
```diff
@@ -528,3 +529,104 @@ describe("InputText", () => {
...
+  describe("When working with uncontrolled components", () => {
+    const onFormValueUpdate = jest.fn();
+    beforeEach(jest.clearAllMocks);
+
+    function FormWrapper({ children }: { readonly children: React.ReactNode }) {
+      const form = useForm();
+      useEffect(() => {
+        return form.watch((value, { name, type }) => {
+          console.log(value, name, type);
+          console.log("value", value);
```
rodrigoeidelvein, line 586:
```suggestion
```

We can remove these

rodrigoeidelvein (changes requested):
Just a minor cleanup on those console.logs and we're good!

I tested with the pre-release branch, and also applying the changes on the deprecated `atlantis/InputText`, it looks good! Thanks, @davidr64

rodrigoeidelvein (approved):
LGTM!

packages/components-native/src/InputText/InputText.test.tsx
```diff
@@ -528,3 +529,102 @@ describe("InputText", () => {
...
+  });
+
+  describe("When working with uncontrolled components", () => {
+    const onFormValueUpdate = jest.fn();
+    beforeEach(jest.clearAllMocks);
+
+    function FormWrapper({ children }: { readonly children: React.ReactNode }) {
+      const form = useForm();
+      useEffect(() => {
+        return form.watch((value, { name, type }) => {
```
rodrigoeidelvein, line 584:
```suggestion
        return form.watch(value => {
```

rodrigoeidelvein (changes requested):
These params are not being used anymore, that's why it failed

packages/components-native/src/InputText/InputText.test.tsx
davidr64: Ah thanks I just pulled to see why

davidr64: I blame me not paying attention and the linting failure output being obtuse.

rodrigoeidelvein (approved):
Linting error is gone! :shipit:

PR #1766, opened 2024-02-16
refactor(components): Increase coverage of DataListHeaderTile

rodrigoeidelvein:
## Motivations

Increasing the coverage of the DataListHeaderTile component. I added a test for the new sorting options, testing the `onSort` callback and the visibility of the options.

I used the same `Header` section of the `DataList.test.tsx` file.

### Added

- 

### Changed

- Added tests for DataListHeaderTile

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.test.tsx
+++ b/packages/components/src/DataList/DataList.test.tsx
@@ -394,6 +394,65 @@ describe("DataList", () => {
       expect(screen.queryByText(mockHeaders.name)).not.toBeInTheDocument();
       expect(screen.queryByText(mockHeaders.email)).not.toBeInTheDocument();
     });
+
+    it("should render the sorting arrows when sorting is specified", () => {
+      renderLayout(undefined, {
+        sortable: [{ key: "name" }],
+        onSort: jest.fn(),
+        state: undefined,
+      });
+      expect(screen.queryByTestId(SORTING_ICON_TEST_ID)).toBeInTheDocument();
+    });
+
+    it("should not render the sorting arrows when sorting is not specified", () => {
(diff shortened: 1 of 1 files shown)


packages/components/src/DataList/DataList.test.tsx
```diff
@@ -394,6 +394,65 @@ describe("DataList", () => {
       expect(screen.queryByText(mockHeaders.name)).not.toBeInTheDocument();
       expect(screen.queryByText(mockHeaders.email)).not.toBeInTheDocument();
     });
+
+    it("should render the sorting arrows when sorting is specified", () => {
```
taylorvnoj, line 398: We do already have this specific test farther down in the file [here](https://github.com/GetJobber/atlantis/blob/master/packages/components/src/DataList/DataList.test.tsx#L471) - but let's delete that one because I like your `it` statement better.

And actually....below that test there is one to `render a title when it's provided` which seems to be tested in the `Header` describe already. We can probably delete that test also

taylorvnoj (changes requested):
Thank you Rodrigo!  Just some clean up to delete the `Sorting` describe since we're covering it in the `Header` describe now. And since sorting is in the header I'm good with that 👌

packages/components/src/DataList/DataList.test.tsx
rodrigoeidelvein: Good catch! I didn't notice them. Just deleted them

taylorvnoj (approved):
LGTM!

PR #1768, opened 2024-02-16
fix(components): Always show clear when there are selected options in ComboBox

ryanb-jobber:
## Motivations

A [bug ticket](https://jobber.atlassian.net/browse/OPS-20251) was reported whereby the client list is blank due to a tag filter, and the filter cannot be cleared because there are currently no client tags in your account, which prevents the clear button from appearing. 

We worked around this issue in https://github.com/GetJobber/Jobber/pull/43181 by clearing selections if there are no tags in the account. However, it seems that having the "clear" button always visible when there are "selections" provided would be valuable. This change ensures the header is alway shown in cases where there is an active selection, regardless of options being supplied. 

### Changed

- ComboBoxHeader always appears when there are selections provided, to ensure the clear operation can always be accessed. 

Screenshot 2024-02-16 at 12 20 26 PM

## Testing

This can be seen in the client list using a few edge case scenarios:

- Switching between franchise accounts after applying a tag filter, and the second franchise does not have any tags
- Having a single tag on some clients, filtering on that tag, and then deleting all those clients

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -39,7 +39,7 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
         handleSearchChange={props.handleSearchChange}
       />
 
-      {((props.multiselect && optionsExist) || props.selected.length > 0) && (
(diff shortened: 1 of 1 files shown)


darryltec (approved):
:shipit:

Issue #1767, opened 2024-02-16
Would like to be able to set the key for data elements on a DescriptionList


type-face:
I'm using the [`DescriptionList` component](https://atlantis.getjobber.com/?path=/story/components-lists-and-tables-descriptionlist-web--basic) to display a list of payments, but am running into problems because the string values that I'm passing to the tuple is not always unique. This causes a warning:
```
Warning: Encountered two children with the same key, `Feb 08, 2024`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```
It also causes duplicate rendering issues if I modify the array that I passed to the `DescriptionList`.

If I add a unique identifier to the string (see second screenshot) this fixes the problem. However, I do not want to show that unique identifier to the end user.

Would it be possible to allow this component to accept data values that have an optional argument for the `key`?

**Preferred**
<img width="477" alt="Screenshot 2024-02-16 at 9 37 53 AM" src="https://github.com/GetJobber/atlantis/assets/6773149/2a3a5d85-78e3-4f85-a6cd-bec4eb60241d">

**Workaround**
<img width="480" alt="Screenshot 2024-02-16 at 9 31 41 AM" src="https://github.com/GetJobber/atlantis/assets/6773149/e25caf45-d013-4e27-8bca-a76ccd5088ee">

type-face:
Happy to take a stab at this myself, just not sure if it's something we want.

PR #1771, opened 2024-02-20
fix(components): Allow duplicated terms on DescriptionList data tuple [1767]

type-face:
## Motivations

I wanted to use a description list but have rows with duplicate "terms".

e.g.
image

 The existing component caused a React unique key error when using the same term multiple times. Additionally, any attempt to set state using this data caused a rendering error (duplicate rendering of each element in the data set) due to the duplicate keys.
Screenshot 2024-02-20 at 12 57 30 PM

## Changes
This change allows the `data` prop for the `DescriptionList` to accept a tuple that also has an optional key value. If the key value is not included, the `key` will be set to the `term` value, as was the existing behaviour of the component. This will prevent breaking any existing uses of the component. However, if the `key` prop is passed in, this will be used in place of the `term` when setting the key for each element.

e.g.
```
    <DescriptionList
      data={[
        ["Issued", "2018-12-08", "1"],
        ["Issued", "2019-01-06", "2"],
      ]}
    />
```

### Fixed

Fixes https://github.com/GetJobber/atlantis/issues/1767

## Testing

`npm test packages/components/src/DescriptionList/DescriptionList.test.tsx`. EXPECT: no warning messages

---

Random photo of Atlantis


type-face:
@GetJobber/atlantis Just curious what the turnaround time is for PR review. I have a feature I'm working on that depends on this change. Hoping to find out soon if this is a viable approach so that I can pivot if needed. Thanks!

packages/components/src/DescriptionList/DescriptionList.tsx
```diff
@@ -4,17 +4,18 @@ import { Typography } from "../Typography";
...
+  readonly data: [string, string | ReactNode, string?][];
 }
 
 export function DescriptionList({ data }: DescriptionListProps) {
   return (
     <dl className={styles.descriptionList}>
-      {data.map(([term, description]) => (
-        <div key={term} className={styles.termGroup}>
+      {data.map(([term, description, key]) => (
+        <div key={key || term} className={styles.termGroup}>
```
darryltec, line 18: Hi, apologize for the delay, we were heads down on the retheme.

thoughts on using term + index instead of adding another value on the tuple? I understand the side effects of it but this is also a super simple list that it's okay to do.

type-face: Thanks for the suggestion, @darryltec! I'll make that change.

packages/components/src/DescriptionList/DescriptionList.test.tsx
```diff
@@ -14,6 +14,18 @@ it("renders an object as a list of key value pairs", () => {
...
+  const { container } = render(
+    <DescriptionList
+      data={[
+        ["Issued", "2018-12-08"],
+        ["Issued", "2019-01-06"],
+      ]}
+    />,
+  );
+  expect(container).toMatchSnapshot();
+});
```
darryltec, line 27: quick question, is this meant to test the key problem? Or just that you can put duplicate terms?

packages/components/src/DescriptionList/DescriptionList.tsx
```diff
@@ -4,17 +4,18 @@ import { Typography } from "../Typography";
 
 interface DescriptionListProps {
   /**
-   * A tuple where the first item is the string to display as the term
-   * and the second value is the string to display as the definition
+   * A tuple where the first item is the string to display as the term,
+   * the second value is the string to display as the definition,
+   * and the third value is an optional key to use as the React key in scenarios where the display value is not unique for all elements in the data set.
```
darryltec, line 9: no more 3rd term

darryltec (changes requested):
Just an issue with documentation

packages/components/src/DescriptionList/DescriptionList.test.tsx
type-face: Both, in a way. This is just to check that there are no warnings when duplicate terms are used, which was what happened before my change. However, this test wouldn't have explicitly failed before, either. It just would have caused a warning message. I added a check to ensure that the console isn't issuing a warning. Does that make sense? Is this a good approach? Sorry, still ramping up my React skills.

packages/components/src/DescriptionList/DescriptionList.tsx
```diff
@@ -5,16 +5,16 @@ import { Typography } from "../Typography";
 interface DescriptionListProps {
   /**
    * A tuple where the first item is the string to display as the term
-   * and the second value is the string to display as the definition
+   * and the second value is the string to display as the definition,
```
darryltec, line 8:
```suggestion
   * and the second value is the string to display as the definition
```

packages/components/src/DescriptionList/DescriptionList.test.tsx
```diff
@@ -14,6 +14,21 @@ it("renders an object as a list of key value pairs", () => {
   expect(container).toMatchSnapshot();
 });
 
+it("allows duplicate terms", () => {
```
darryltec, line 17: if this is testing both allowing and not throwing the key error, i'd like that to be reflected here

packages/components/src/DescriptionList/DescriptionList.test.tsx
```diff
@@ -14,6 +14,21 @@ it("renders an object as a list of key value pairs", () => {
...
+  const { container } = render(
+    <DescriptionList
+      data={[
+        ["Issued", "2018-12-08"],
+        ["Issued", "2019-01-06"],
+      ]}
+    />,
+  );
+  expect(container).toMatchSnapshot();
+  expect(logSpy).toHaveBeenCalledTimes(0);
```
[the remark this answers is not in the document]
type-face: Hi @darryltec. Do you mean that I should do something like?:
`  expect(logSpy).toHaveBeenCalledWith(
    'Warning: Encountered two children with the same key, `%s`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.%s", "Issued", ',
  );`
That would be more specific but also doesn't seem great as I expect that warning message might change at some point. Would be an issue for an upgrade, though.

Sorry, still ramping up my React skills and not sure the best practice for testing these kind of warnings.

packages/components/src/DescriptionList/DescriptionList.test.tsx
```diff
@@ -14,6 +14,21 @@ it("renders an object as a list of key value pairs", () => {
...
+
+  const { container } = render(
+    <DescriptionList
+      data={[
+        ["Issued", "2018-12-08"],
+        ["Issued", "2019-01-06"],
+      ]}
+    />,
+  );
+  expect(container).toMatchSnapshot();
```
darryltec, line 28: while you're updating this, i'd like to not add more snapshots so if we can 

```ts
expect(getAllByText("Issued")).toHaveLength(2)
```

that would be great!

packages/components/src/DescriptionList/DescriptionList.test.tsx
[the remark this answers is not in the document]
darryltec: we could just check for partial text like `Warning: Encountered two children with the same key` I'd rather be specific and it failing on upgrade than it being very generic and failing randomly. Especially locally on dev as we have some pre-commit hooks that would fail.

packages/components/src/DescriptionList/DescriptionList.test.tsx
```diff
@@ -14,6 +14,38 @@ it("renders an object as a list of key value pairs", () => {
...
+  render(
+    <DescriptionList
+      data={[
+        ["Issued", "2018-12-08"],
+        ["Issued", "2019-01-06"],
+      ]}
+    />,
+  );
+  expect(logSpy).not.toHaveBeenCalledWith(
+    expect.stringContaining(
```
type-face, line 41: @darryltec This is what I came up with. This seems like a strange test to me, though. However, it does protect against a regression. Let me know if this is what you were thinking.

darryltec: Perfection!

darryltec (approved):
:shipit:

PR #1781, opened 2024-02-26
feat(components): Adds a role of status to the Glimmer component 

davidr64:
## Motivations

This will hopefully improve accessibility but honestly the reason for this change is to make it easier to grab glimmers during testing. Rather than relying on glimmer test IDs.  We can use [getByRole](https://testing-library.com/docs/queries/byrole/#busy) to find the glimmers on the page. 

`getAllByRole('status', { busy: true })` will return a non empty array if there are glimmers on the page. Without this change the same line would work however it would be `getAllByRole('generic', { busy: true})` which doesn't feel as correct. 

https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html

>4. After a user activates a process, an icon symbolizing 'busy' appears on the screen. The screen reader announces "application busy".

## Changes

Simply adding the status role to the glimmer component. 

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Glimmer/Glimmer.tsx
+++ b/packages/components/src/Glimmer/Glimmer.tsx
@@ -76,6 +76,7 @@ export function Glimmer({
   return (
     <div
       aria-busy="true"
+      role="status"
       className={className}
       data-testid={GLIMMER_TEST_ID}
       style={{ width }}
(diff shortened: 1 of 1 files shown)


taylorvnoj (approved):
LGTM - thanks for adding this David!

PR #1788, opened 2024-03-04
feat(components): Allow a prefix element on ComboBox.Option. Promote internal StatusLabel to StatusIndicator, top level component 

felgamedev:
## Motivations

We want to be able to show status of options in a combobox list in the form of a small icon. To do so, we want to reuse the existing StatusIcon in a higher level while maintaining the option to pass any other React element as a prefix (such as text, or an icon)

### Added

StatusIcon[Internal component in StatusLabel] -> StatusIndicator [Component]

### Changed

- Combobox.Option now has optional prefix prop

## Testing

- Use the prefix on a combobox option, passing in any valid React element
- Create a StatusIndicator component

---

Random photo of Atlantis


packages/components/src/StatusIndicator/StatusIndicator.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+interface StatusIndicatorProps {
+  readonly status: StatusType;
+}
+
+export function StatusIndicator({ status }: StatusIndicatorProps) {
+  return (
+    <div
+      style={{ backgroundColor: `var(--color-${status}` }}
+      className={styles.statusIndicator}
```
ericchernuka, line 19: Should this be a `<span />` or is it intended to always be `display: block`?

packages/components/src/StatusIndicator/StatusIndicator.tsx
```diff
@@ -0,0 +1,22 @@
...
+  | "informative";
+
+interface StatusIndicatorProps {
+  readonly status: StatusType;
+}
+
+export function StatusIndicator({ status }: StatusIndicatorProps) {
+  return (
+    <div
+      style={{ backgroundColor: `var(--color-${status}` }}
```
ericchernuka, line 18: What's our stance on style attributes being used instead of classes? Just wondering as I've typically seen the styles given a name based on the status.

docs/components/StatusIndicator/StatusIndicator.stories.mdx
```diff
@@ -0,0 +1,48 @@
...
+import { Content } from "@jobber/components/Content";
+
+<Meta title="Components/Status and Feedback/StatusIndicator/Docs" />
+
+# Status Indicator
+
+The StatusIndicator component is a purely visual component that indicates a
+status (e.g., "Active", "Overdue").
+
+## Usage guidelines
```
darryltec, line 12: I think we should add another sentence in between about "wanting a label? use StatusLabel"

packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.tsx
```diff
@@ -29,7 +31,12 @@ export function ComboboxOption(props: ComboboxOptionProps) {
       }
       className={classnames(styles.option)}
     >
-      {props.label}
+      <Flex template={["shrink", "grow"]}>
+        {props.prefix}
+
+        {props.label}
+      </Flex>
```
darryltec, line 38: this is a bit problematic on long labels when you don't have a prefix 

image

for reference, it does this when you have a prefix

image

packages/components/src/StatusIndicator/StatusIndicator.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { render } from "@testing-library/react";
+import React from "react";
+import { StatusIndicator } from "./StatusIndicator";
+
+it("renders a success status indicator", () => {
+  const { container } = render(<StatusIndicator status="success" />);
+  expect(container).toMatchSnapshot();
```
darryltec, line 7: i wouldn't use a snapshot for this one. I'd actually just assert that. Also, from the looks of it, the snapshot seems wrong too

packages/components/src/StatusIndicator/StatusIndicator.tsx
darryltec: yeah we normally create a CSS file for it and then reference that + use the .d.ts to generate the types we need.

https://github.com/GetJobber/atlantis/blob/a5704355cdc4ffef2919c12a4dbd028c675df909/packages/components/src/ProgressBar/ProgressBarStepped.tsx#L10

https://github.com/GetJobber/atlantis/blob/a5704355cdc4ffef2919c12a4dbd028c675df909/packages/components/src/ProgressBar/Sizes.css#L1-L11

Mostly so we can take advantage of CSS goodness when we need to.

With that said, you mostly just moved this to its own component so i don't mind it.

packages/components/src/StatusIndicator/StatusIndicator.tsx
darryltec: could be a span tho i'd add `display: inline-block` on it tho so it properly flows with other inline elements

packages/components/src/StatusIndicator/StatusIndicator.css
```diff
@@ -0,0 +1,8 @@
+.statusIndicator {
+  width: var(--statusLabel-icon-diameter);
+  height: var(--statusLabel-icon-diameter);
+  margin-top: var(--space-minuscule);
```
darryltec, line 4: i'd actually take this margin out as that would create so much pain for us when you try and use this component where a parent is vertically aligning to the middle and this is offset by 1px down.

I get why this is here tho because of status label so i'd find a way to preserving old styling and alignment on StatusLabel while keeping StatusIndicator free of opinions about spacing.

packages/components/src/StatusLabel/StatusType.ts
darryltec, line 1: Since this is more so a StatusIndicator thing, i'd move this file there (and rename to `StatusIndicator.type.ts`)

darryltec (changes requested):
found an issue with the flex usage in combobox. All-in-all this is pretty good. Just need a bit of a code reorg and cleanup

packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.tsx
felgamedev: Found a way to ensure the rules are applied, not sure how pretty it is though!

docs/components/StatusIndicator/StatusIndicator.stories.mdx
```diff
@@ -0,0 +1,53 @@
...
+| <StatusIndicator status="success" />     | `success`     |
+| <StatusIndicator status="critical" />    | `critical`    |
+| <StatusIndicator status="informative" /> | `informative` |
+| <StatusIndicator status="warning" />     | `warning`     |
+| <StatusIndicator status="inactive" />    | `inactive`    |
+
+## Related components
+
+If you are looking to use a StatusIndicator with a label, consider using
+[StatusLabel](../?path=/docs/components-status-and-feedback-statuslabel-docs--page).
```
darryltec, line 53: We should be stern on this one and go with "you should use" instead of consider

darryltec (changes requested):
just found 1 little tiny thing and we should be good to go!

darryltec (approved):
:shipit: thanks for your contribution!

PR #1789, opened 2024-03-05
fix(components): Fix button label and icon alignment when inside a modal

fagnanm:
## Motivations
When using actions on a modal we allow the passing of icons into the buttons. Turns out our CSS for ordering of the actions also affects the children within the action (labels and icons). This causes whenever you add an icon to a button the right section of a modal the icon will also swap places with the label.

## Changes
Made the global selector more specific so it doesn't touch every element within the left/right sides, just the first element (the buttons themselves)

### Before
image

### After
image

## Testing

Add an icon to a modal action and see what happens

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Modal/Modal.css
+++ b/packages/components/src/Modal/Modal.css
@@ -131,10 +131,10 @@
   margin-left: var(--space-small);
 }
 
-.rightAction *:first-child {
+.rightAction > *:first-child {
   order: 2; /* 1 */
 }
 
-.rightAction *:nth-child(2) {
+.rightAction > *:nth-child(2) {
   order: 1; /* 1 */
 }


ZakaryH:
so I was slightly worried this would be a widespread change of the default icon positioning on `Modal`s going from right to left and wanted to see if that resulted in any bad scenarios but after looking through the product, I actually can't find an instance of a Modal with actions that have an icon

believe it or not, the request that spawned this ticket might be the first instance of a Modal with an action having an icon in 5 years!

ZakaryH:
tested this out with some icons with `iconOnRight: true` and then omitting it entirely to leave it on the left as the default behavior

buttons are all still in the correct places
image

icons are being put in the right place too

lastly, while I was searching through the product I did find some instances of Modals that only have a `primaryAction` and `tertiaryAction`, no secondary so thought I'd double check that. also fine

image

ZakaryH (approved):
QA'd it, looks great

checked out the initial PR that added this code, couldn't see anything that it was trying to accomplish that this won't also cover

LGTM

PR #1790, opened 2024-03-06
fix(components): Fix DataListLayout not rerendering

rodrigoeidelvein:
## Motivations

`DataList` was not rerending if the `children` in the same layout changed because it was not being reregistered with the new `children`.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- DataList not rerendering when the `children` of the same layout changed.

### Security

- 

## Testing

`git checkout 7098d47` and go to the `DataList` page and try to toggle the loading using the button on the top.

---

Random photo of Atlantis


darryltec (approved):
i'd like to think about how we can improve this better but to unblock you, since this was added recently, this LGTM

PR #1792, opened 2024-03-07
fix(components): Fix an improper console mock in a test

mattander:
## Motivations

I mocked out console error in one test but forgot that the scope would leak since it was on global scope 🤦‍♂️ 

I've fixed it with a [method from Kent C Dodds](https://github.com/jestjs/jest/pull/5267#issuecomment-356605468). This is to keep the console clean in tests with expected errors. 

## Testing
Just drop a console error in a test following this one and make sure you see it. 
---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -431,9 +431,9 @@ describe("Combobox Compound Component Validation", () => {
 
   it("throws an error when there are multiple Combobox Activators present", () => {
     // This keeps the testing console clean
-    jest.spyOn(console, "error").mockImplementation(() => {
-      return;
-    });
+    const consoleErrorSpy = jest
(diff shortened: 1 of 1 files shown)


packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -431,7 +431,9 @@ describe("Combobox Compound Component Validation", () => {
 
   it("throws an error when there are multiple Combobox Activators present", () => {
     // This keeps the testing console clean
-    console.error = jest.fn();
+    jest.spyOn(console, "error").mockImplementation(() => {
```
ericchernuka, line 434:
```suggestion
const consoleErrorSpy = jest.spyOn(console, "error").mockImplementation(() => ({}))`

... test

consoleErrorSpy.mockRestore()
```

PR #1793, opened 2024-03-07
fix: change bottom sheet background to surface background and reduce radius

darryltec:
## Motivations

This is the retheme style now

### Added

add surface background
image

### Changed

- to smaller radius for the bottom sheet and content overlay

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


chris-at-jobber:
These are more questions for @fagnanm, but I'm unclear why we're making the changes here.

- I thought the plan was to re-introduce dark overlays - this would mean we can keep the sheets using `surface`
- Why are we making the BottomSheet and ContentOverlay less round and changing the background colors?

Design discussion aside, I think this is a much less pressing change than the others so I'm going to jump to the larger mobile one.

PR #1799, opened 2024-03-11
feat(components): Add ESM Build

scotttjob:
## Motivations

Besides it being a good, modern thing to do. We need this to land for Vite reasons.

## Changes

- Added an ESM Build
- Updated CJS Build to use Index file instead of multi-file approach.
- Still using the Barrel files to generate the exports.

### Added

- ESM Build

### Changed

- CJS Build Style, should be 1:1 compatible regardless.

## Testing

- Nothing existing should break with this change. Types + Code is still available the same way for existing imports.

---

Random photo of Atlantis


scotttjob:
Went nuclear on this branch to learn. I learned what I needed to, but we would never merge this level of a blown up repo.

PR #1800, opened 2024-03-11
refactor(components): Deprecate to prop for Button

dlrussel:
## Motivations

We need to decouple the atlantis button from react router in order to support the Remix work and support SSR.
In order to do this we need to remove the to prop from our Atlantis button component.

### Deprecated
Add deprecation warning to the Button `to` prop.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -53,7 +53,9 @@ interface ButtonAnchorProps extends ButtonFoundationProps {
 
 interface ButtonLinkProps<S = unknown> extends ButtonFoundationProps {
   /**
-   * Used for client side routing. Only use when inside a routed component.
+   * **Deprecated**: to will be removed in the next major version
+   * Use shared ButtonLink component from Jobber instead
+   * @deprecated
    */
   readonly to?: LinkProps<S>["to"];
 }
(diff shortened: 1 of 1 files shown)


packages/components/src/Button/Button.tsx
```diff
@@ -53,7 +53,9 @@ interface ButtonAnchorProps extends ButtonFoundationProps {
 
 interface ButtonLinkProps<S = unknown> extends ButtonFoundationProps {
   /**
-   * Used for client side routing. Only use when inside a routed component.
+   * **Deprecated**: to will be removed in the next major version
+   * Use shared ButtonLink component from Jobber instead
+   * @deprecated
```
scotttjob, line 58: Okay so the idea here is to deprecate the prop until we issue the next major version change? 

I think that might line up well with the ESM change I'm prepping now, so consumers might not have a ton of warning, but we could absolutely write a small documentation piece that suggests how to fix this (either use the ButtonLink in JO, or implement it yourselves in your own project).

My only suggestion about the copy/text above is recommending consumers `use shared ButtonLink component from Jobber instead`. Since Atlantis is an open-source repo and we do have consumers outside the JO ecosystem, we might need to remove that line, maybe opt for a link to that short bit of documentation we'll need to write?

scotttjob (changes requested):
Small change request to the text for the comment.

Instead of my documentation suggestion for now, we could just delete the reference to JO for now and circle back later if we decide to write the guidance documentation.

scotttjob (approved):

PR #1802, opened 2024-03-13
feat(components): Restyle web statusLabel

fagnanm:
## Changes
- StatusLabels look different, they have a background and rounded corners with matching coloured text
- StatusIndicator becomes a circle instead of a rounded square
- InputValidation (and any place use we use `Typography textColor="critical"`) gets a darker shade of red
- Moved a root variable to statusIndicator from statusLabel as it is no longer used in statusIndicator

### Before
image

image

### After
image

Input validation text changes to using a darker red as Typography critical variant was updated
image

## Comments
StatusLabel and inlineLabel (which was deprecated but not really because people still use it) now look pretty similar. We could continue that way, or potentially combine the two components and make the indicator a boolean. Will leave that decision to later.

## Testing
Atlantis import can be tested in this Rails branch `restyle-status-labels`. Note the colours will be slightly different as I've updated the tokens that we'll change to after the re-theme. 

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/StatusIndicator/StatusIndicator.css
+++ b/packages/components/src/StatusIndicator/StatusIndicator.css
@@ -1,8 +1,12 @@
+:root {
+  --statusLabel-icon-diameter: 8px;
+}
+
 .statusIndicator {
   display: inline-block;
   width: var(--statusLabel-icon-diameter);
   height: var(--statusLabel-icon-diameter);
-  border-radius: 3px;
+  border-radius: 50%;
   background-color: var(--color-success);
   flex-shrink: 0;
 }
--- a/packages/components/src/StatusLabel/StatusLabel.css
+++ b/packages/components/src/StatusLabel/StatusLabel.css
@@ -1,20 +1,58 @@
-:root {
-  --statusLabel-icon-diameter: 10px;
-}
-
 .statusLabelRow {
   display: flex;
+  width: fit-content;
+  padding: 6px 10px 6px 8px;
+  border-radius: 12px;
   flex-direction: row;
   flex-wrap: nowrap;
-  gap: var(--space-small);
-  width: fit-content;
+  gap: 6px;
+}
+
+/* Reset the <Text> line height so we can reliably get 24px tall */
+.statusLabelRow p {
+  line-height: 1;
 }
 
 .labelTextEndAligned {
   flex-direction: row-reverse;
 }
 
-.statusIndicator {
-  display: inline-flex;
-  padding-top: var(--space-minuscule);
+.statusIndicatorWrapper {
+  display: flex;
+  padding-top: 2px;
+}
+
+.success {
+  background-color: var(--color-success--surface);
+  p {
+    color: var(--color-success--onSurface);
+  }
+}
+
+.warning {
+  background-color: var(--color-warning--surface);
+  p {
+    color: var(--color-warning--onSurface);
+  }
+}
+
+.critical {
+  background-color: var(--color-critical--surface);
+  p {
+    color: var(--color-critical--onSurface);
+  }
+}
+
+.inactive {
+  background-color: var(--color-inactive--surface);
+  p {
+    color: var(--color-inactive--onSurface);
+  }
+}
+
+.informative {
+  background-color: var(--color-informative--surface);
+  p {
+    color: var(--color-informative--onSurface);
+  }
 }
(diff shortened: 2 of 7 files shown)


packages/components/src/StatusIndicator/StatusIndicator.css
```diff
@@ -1,8 +1,12 @@
+:root {
```
fagnanm, line 1: Moved this from statusLabel to the indicator as it's no longer referenced in that CSS just this one.

packages/components/src/StatusLabel/StatusLabel.css
```diff
@@ -1,20 +1,58 @@
...
+  border-radius: 12px;
   flex-direction: row;
   flex-wrap: nowrap;
-  gap: var(--space-small);
-  width: fit-content;
+  gap: 6px;
+}
+
+/* Reset the <Text> line height so we can reliably get 24px tall */
+.statusLabelRow p {
```
fagnanm, line 12: Is there a better way to do this using Typography rather than Text?

packages/components/src/StatusLabel/StatusLabel.css
```diff
@@ -1,20 +1,58 @@
-:root {
-  --statusLabel-icon-diameter: 10px;
-}
-
 .statusLabelRow {
   display: flex;
+  width: fit-content;
+  padding: 6px 10px 6px 8px;
```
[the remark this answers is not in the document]
MichaelParadis: These won't be converted from what I see in my dev tools based on the output of the build
image

packages/components/src/StatusLabel/StatusLabel.css
MichaelParadis: I found setting the height of `.statusLabelRow` to `12px` (24px - paddingTop - paddingBottom) and removing the lineHeight got the height of the status label to reliably get to 24px as well. Up to you on that though (I didn't do extensive testing)

packages/components/src/StatusIndicator/StatusIndicator.css
MichaelParadis: Makes sense

packages/components/src/StatusLabel/StatusLabel.css
fagnanm: @MichaelParadis yeah challenge there will be in some cases the labels will wrap onto two lines which will throw the height off.

MichaelParadis: That makes sense. In that case I think the lineHeight stuff makes the most sense

packages/components/src/StatusLabel/StatusLabel.css
```diff
@@ -1,20 +1,48 @@
 :root {
-  --statusLabel-icon-diameter: 10px;
+  --labelBackgroundColor: var(--color-success--surface);
+}
+
+.success {
```
fagnanm, line 5: Not sure the best way to do this. Success is the default, but something was getting mad at me because success was being passed as a class as a status but was included.

packages/components/src/StatusLabel/StatusLabel.css
```diff
@@ -1,20 +1,48 @@
...
 .labelTextEndAligned {
   flex-direction: row-reverse;
 }
 
-.statusIndicator {
-  display: inline-flex;
-  padding-top: var(--space-minuscule);
+.statusIndicatorWrapper {
+  display: flex;
+  padding-top: 0.125rem;
```
taylorvnoj, line 47: 0.125rem should = 2px

```suggestion
  padding-top: var(--space-smallest);
```

taylorvnoj: Unless we're using `rem` for any potential dimension changes?

packages/components/src/StatusLabel/StatusLabel.css
taylorvnoj: I think this is great! I don't see anything getting mad?

taylorvnoj (commented):
Mostly a question about using a token in one spot - They look great!

packages/components/src/StatusLabel/StatusLabel.css
fagnanm: @taylorvnoj reason I've kept it as rems is that if we ever update space-small there is a very small chance that it would still align the indicator with the center of the text label unless the text size was changed in the same proportions.

taylorvnoj: That makes sense!

PR #1807, opened 2024-03-14
fix(components): Use unique item key in DataListSortingOptions

ryanb-jobber:
## Motivations

The clients index in Jobber supports directional sorting on names. Because we have two sorting options with the same id, but different direction, we see the following error in the browser console:

```
Warning: Encountered two children with the same key, `firstName`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

This error also floods the console when unit testing.

### Changed

Construct element `key` based on both the option `id` and `order`

Random photo of Atlantis


taylorvnoj (approved):
LGTM - tested via pre-release and console errors are gone. Thanks @ryanb-jobber

PR #1810, opened 2024-03-18
fix(components): InternalThumbnailImage to support SSR

scotttjob:
## Motivations

We would like Atlantis to support SSR, and currently the InternalThumbnailImage sets its imageSource as part of the initial render, instead of in a side effect.

So we updated this initial call to a blank useEffect, which does the same thing as the original code, in a marginally better way.

## Changes

- Wrapping the setImageSource in a useEffect inside InternalThumbnailImage of the FormatFile component.

### Changed

- FormatFile will now set its source image after the first render vs on the first render.

### Fixed

- SSR support for the FormatFile component.

## Testing

FormatFile should be unchanged in a browser environment.
FormatFile should now render without error in a server environment.

---

Random photo of Atlantis


chris-at-jobber:
@scotttjob does this change impact the user's perception of the loading state of this component? I'm reading it as "no" based on the description and the preview build, but just want to confirm.

scotttjob:
> @scotttjob does this change impact the user's perception of the loading state of this component? I'm reading it as "no" based on the description and the preview build, but just want to confirm.

Nope it should be transparent to the user. It happens a single frame later than normal, but there was still a loading in process with the `then` callback in play.

rodrigoeidelvein (approved):
I did some testing with the `useSafeLayoutEffect` too and I didn't notice any difference, I don't think it's something that end-users will notice.

LGTM!

PR #1812, opened 2024-03-20
TEST: Do not merge directly

scotttjob:
## Motivations

This is a combination of a few different PRs that are actively in flight to give us the ability to test these changes as a whole in context before they land.

This PR will never leave DRAFT and will be eventually closed.

## Changes

- ESM Build + useMediaQuery update + InternalThumbnailState update.

---

Random photo of Atlantis


.gitignore
```diff
@@ -8,3 +8,4 @@ node_modules
 .DS_Store
 .idea
 storybook-static
+.rollup.cache
```
scotttjob, line 11: With the new changes to the build, there was a subfolder for .rollup.cache which we don't want as part of the repo so I've ignored it. It's a build artifact, the whole folder.

packages/components/rollup.config.mjs
```diff
@@ -1,35 +1,63 @@
 /* eslint-disable import/no-default-export */
-import multiInput from "rollup-plugin-multi-input";
+import path from "path";
```
scotttjob, line 2: In general the changes in this file are as follows:

Update to be an ESM based file with the update to MJS.

Updated all 'require' calls to 'import' calls

Included modern _filename,__dirname variables as they are not available in MJS files.

Updated the glob to include all files under the `src` directory including a base level `index`

Included the nodePolyfills project so that `process` is correctly polyfilled for the `react-markdown` package.

Included a new `alias` entry for jobber/hooks. This means we no longer have to include the excludes workaround and can package jobber/hooks directly with jobber/components. This is necessary for Vite + PNPM support.

Updated the CSS Scheme to Extract instead of Inject. You will now need to include either `@jobber/components/styles` in an ESM environment or `@jobber/components/dist/styles.css` in a CJS environment in order to see Atlantis Component Styles.

Added a new ESM build to compliment the existing CJS build.

Moved the copy step until after the bundle has completed building. This ensures the index.d.ts file is present.

Removed a variety of packages from the external configuration section. In general, packages in this array should ALSO be in peerDependencies in package.json. The files being excluded from the build, was causing an issue with Vite + PNPM due to how strict they are.

packages/components/src/index.tsx
```diff
@@ -0,0 +1,73 @@
+export * from "./AnimatedPresence";
```
scotttjob, line 1: This file is required for ESM support. When creating new components in the future you will have to update this file, along with the package.json file to correctly export your component. I will update the documentation to reflect this change.

packages/components/tsconfig.json
```diff
@@ -11,7 +11,7 @@
     "outDir": "dist",
     "paths": {
       "@jobber/hooks/*": [
-        "../../packages/hooks/src/*"
```
scotttjob, line 14: If we look in the src directory it means we have to parse the typescript twice, which was causing a lot of issues in Rollup. Looking at the dist folder means we look at built files instead. I believe this was the root of why we needed to do our old jobber/hooks juggling in order to get it loading.

packages/design/rollup.config.cjs
```diff
@@ -1,14 +1,14 @@
 /* eslint-disable import/no-default-export */
-import multiInput from "rollup-plugin-multi-input";
```
scotttjob, line 2: Updating this file to be cjs instead of js just means that rollup treats this build as a CJS build, which it still is.

.prettierignore
```diff
@@ -1,3 +1,4 @@
 packages/generators/templates
+packages/components/dist/styles.css
```
scotttjob, line 2: CI breaks without this. CSS was not part of our component build previously.

.stylelintignore
```diff
@@ -1,3 +1,4 @@
 packages/design/foundation.css
+packages/components/dist/styles.css
```
scotttjob, line 2: CI breaks without this. CSS was not part of our component build previously.

packages/components/src/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+export * from "./Checkbox";
+export * from "./Chip";
+export { Chips } from "./Chips";
+export * from "./Combobox";
+export * from "./ConfirmationModal";
+export * from "./Content";
+export * from "./Countdown";
+export * from "./DataDump";
+export * from "./DataList";
+export { DataTable, createColumnHelper } from "./DataTable";
```
scotttjob, line 18: There are some component name collisions if * is exported from the items in this list with only named exports.

packages/components/rollup.config.mjs
```diff
@@ -1,35 +1,63 @@
...
+import presetenv from "postcss-preset-env";
+import multiInput from "rollup-plugin-multi-input";
+import nodePolyfills from "rollup-plugin-polyfill-node";
+import alias from "@rollup/plugin-alias";
 
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
 export default {
-  input: `src/*/index.{ts,tsx}`,
+  input: `src/**/index.{ts,tsx}`,
```
scotttjob, line 21: It doesn't find the root index.tsx file without this change.

packages/components/rollup.config.mjs
```diff
@@ -1,35 +1,63 @@
...
   plugins: [
-    multiInput(),
+    nodePolyfills(),
+    alias({
+      entries: [
+        {
+          find: /@jobber\/hooks\/(.*)/,
```
scotttjob, line 27: So that imports such as `import { useAssert } from '@jobber/hooks/useAssert` work properly.

packages/components/rollup.config.mjs
```diff
@@ -1,35 +1,63 @@
...
+          replacement: (_, p1) =>
+            path.resolve(
+              __dirname,
+              `../../node_modules/@jobber/hooks/dist/${p1}/${p1}.js`,
+            ),
+        },
+      ],
+    }),
+    nodeResolve(),
```
scotttjob, line 36: Makes anything trying to resolve in the 'node' way possible.

packages/components/rollup.config.mjs
```diff
@@ -92,40 +120,45 @@ export default {
           src: "src/Combobox/components/ComboboxContent/ComboboxContentList/ComboboxContent.css.d.ts",
           dest: "dist/Combobox/components/ComboboxContent/ComboboxContentList",
         },
+        {
```
scotttjob, line 123: publint prefers if an mts types file is present for mjs/esm builds, so that's why it gets copied and pasted to a new extension.

packages/components/rollup.config.mjs
```diff
@@ -92,40 +120,45 @@ export default {
...
+          src: "dist/index.d.ts",
+          dest: "dist",
+          rename: "index.d.mts",
+        },
       ],
+      hook: "writeBundle",
```
scotttjob, line 129: Copy everything at the end (that way all the files are actually present)

rodrigoeidelvein:
I'm still testing/reviewing it, but I just thought of something... with this CSS change, do you think we should update the installation instructions [here](https://github.com/GetJobber/atlantis/blob/master/packages/components/README.md) too?

packages/components/rollup.config.mjs
```diff
@@ -92,40 +120,45 @@ export default {
...
     "classnames",
-    new RegExp("lodash/.*"),
     "@std-proposal/temporal",
     "@jobber/design",
     "@jobber/design/foundation",
     "@jobber/formatters",
-    new RegExp("@jobber/hooks/.*"),
     "zxcvbn",
-    "use-resize-observer/polyfilled",
-    "@apollo/client",
```
rodrigoeidelvein, line 128: Why were some of these removed and others not?

scotttjob: Good question!

These packages caused issues with jobber-frontend when we packaged them this way after getting the ESM piece working there.

I think we're slightly misusing the 'external' configuration option for Rollup, because if we're making something as external, I believe we also have to mark it as a peerDep inside the package.json.

I think we're getting away with it because some of the packages we've marked as external (so rollup doesn't bundle them) just happen to be present in node_modules because someone else manually included that package (say inside JO), or because they were included by a 3rd party library.

I wanted to go the full mile and either mark everything in here as a peerDep inside package.json, or remove them from here entirely but I didn't want to overload this PR with unrelated changes.

I definitely think we should circle back here when we get some tech debt time, to clean up out we're bundling/exporting/peerdepping 3rd party libraries. I think if we moved `react-hook-forms` in here specifically into a peerDep inside package.json, we could fix the issues people were having with trying to roll their own react-hook-forms context.

scotttjob:
> I'm still testing/reviewing it, but I just thought of something... with this CSS change, do you think we should update the installation instructions [here](https://github.com/GetJobber/atlantis/blob/master/packages/components/README.md) too?

Yeah that's a great idea. I was going to circle back to docs after we land the changes, but it's probably a good idea to update them in tandem with the code.

rodrigoeidelvein (approved):
Awesome, Scott! 

I tested it in jobber-frontend, jobber-online, new Remix/Next/vite projects and it all worked as expected. I just need we might need to write a small heads up in the [breaking changes](https://atlantis.getjobber.com/?path=/docs/changelog-upcoming-breaking-changes--page) page so people know they need to include the `import "@jobber/components/styles";` too.

Cool thing that we can now import multiple components in a single `import` statement :)

`import { Switch, Card } from "@jobber/components";`

LGTM!

PR #1814, opened 2024-03-20
fix(components): Make useMediaQuery SSR Friendly

dlrussel:
Note that we originally merged this but reverted because we thought there might have been a last minute breaking change but we have assured that everything works as expected.

## Motivations
Our useMediaQuery implementation currently isn't SSR friendly because it makes use of DOM specific objects.

In order to prepare atlantis for the remix work we need to ensure that our library is compatible with server side rendering. With SSR the application will be rendered on the server and the server doesn't recognize the window and therefore our currently implementation won't work.

## Changes
Update useMediaQuery to make use of useSyncExternalStore from the react library that enables this hook to be rendered on the server instead of just on the client via a guard.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/hooks/useMediaQuery.ts
+++ b/packages/components/src/DataList/hooks/useMediaQuery.ts
@@ -1,6 +1,22 @@
-import { useEffect, useState } from "react";
+import { useCallback, useSyncExternalStore } from "react";
 
-export function useMediaQuery(CSSMediaQuery: string) {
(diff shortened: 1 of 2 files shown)


scotttjob (approved):
LGTM.

The implementation largely matches some internal dupcliation uses we have.

So we'll be able to update those to use this, once this lands. 

Great work!

LGTM!

PR #1820, opened 2024-03-21
fix(components): Fixing issue with InputDate causing an additional render

scotttjob:
## Motivations

The first time a field adjacent to an InputDate causes a re-render, it triggers a console error about an incorrect setState.

We tracked it down to InputDate setting a ref reference directly on Render. I've moved it to a useEffect to mark it as a sideEffect. Tests are passing locally, want to see if the same is true in JO.

## Changes

- Fixed a bug in InputDate where an extra render was being fired on every state update.

## Testing

InputDate should operate as intended, and consumers of Atlantis should no longer see the console error in their tests or locally.

---

Random photo of Atlantis


dlrussel (approved):
Checked this pre-release in JO and I didn't see any errors! The fix looks good to me! 🚢

PR #1830, opened 2024-03-26
feat(components-native): BREAKING CHANGE: Move react-native-svg to peerDependencies

krakitine:
## Motivations

When upgrading `react-native-svg` in a consumer of `@jobber/components-native` to another major version, a runtime error is encountered: 
```
 ERROR  Invariant Violation: Tried to register two views with the same name RNSVGRect
```

This was due to the consumer (jobber-mobile) having v15 and components-native on v13

## Changes

- BREAKING CHANGE: Moved `react-native-svg` from `dependencies` to `peerDepencencies`

## Testing

I checked the https://github.com/software-mansion/react-native-svg/releases to see that v13 ➤ v14 had some breaking changes related to imports. Good news, we don't use those imports.

When updating jobber-mobile to v15, I noticed that snapshots started to fail with 
```
      <RNSVGSvgView
        align="xMidYMid"
-       bbHeight="24"
-       bbWidth="24"
+       bbHeight={24}
+       bbWidth={24}
        focusable={false}
        meetOrSlice={0}
        minX={0}
        minY={0}
        style={
```
I was surprised not to see it here, when doing an 
```
npm ls react-native-svg               
atlantis@0.0.1 /Users/konstantin/dev/atlantis
└─┬ @jobber/components-native@0.58.3 -> ./packages/components-native
  └── react-native-svg@12.4.4
```
why is it on v12?

---

Random photo of Atlantis


taylorvnoj:
@krakitine Thank you for doing this - I think we'll also want to add `react-native-svg` as a dev dependency on the root level `package.json`, much like our other peer dependencies in `components-native/package.json`

krakitine:
> @krakitine Thank you for doing this - I think we'll also want to add `react-native-svg` as a dev dependency on the root level `package.json`, much like our other peer dependencies in `components-native/package.json`

I added `react-native-svg@15` as a devDependency 
I had to go into the package-lock and remove reference to v12 and then update the snapshots for Icon

taylorvnoj (approved):
LGTM - everything plays nicely in atlantis and then back into consumers of `@jobber/components-native`.  Thanks again @krakitine

PR #1834, opened 2024-03-28
feat(components): Heading can accept an element prop to override the numbered heading element

mattander:
## Motivations

Our designers often use headings in their designs for the styles without intending for them to be that semantic heading. However, there's no way for devs to do this without custom CSS. So our codebase is filled with invalid HTML where we have heading structures like h1 > h4 > h4 > h2 > h4. This is an HTML violation and an accessibility violation. 

## Changes

You can now cast a heading as an element different than the level you provided. So you can have a span styled like an `<h2>` or an `<h4>` that looks like an `<h1>`. This separates the intended styles from the semantic meaning of the element and allows devs to accomplish designs without disrupting document structure.

## Testing

Try it out!
---

## This is a great gif of Atlantis
amazing gif of atlantis

### DIFF
--- a/packages/components/src/Heading/Heading.tsx
+++ b/packages/components/src/Heading/Heading.tsx
@@ -11,7 +11,7 @@ interface HeadingProps {
   readonly level: HeadingLevel;
   readonly children: ReactNode;
   /**
-   * @default "p"
+   * Allows overriding of the element rendered. Defaults to the heading specified with level.
    */
   readonly element?: "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "p" | "span";
 }
--- a/packages/components/src/Heading/Heading.tsx
+++ b/packages/components/src/Heading/Heading.tsx
@@ -10,11 +10,15 @@ interface HeadingProps {
    */
   readonly level: HeadingLevel;
   readonly children: ReactNode;
+  /**
+   * @default "p"
+   */
+  readonly element?: "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "p" | "span";
 }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Heading/Heading.tsx
```diff
@@ -10,11 +10,15 @@ interface HeadingProps {
    */
   readonly level: HeadingLevel;
   readonly children: ReactNode;
+  /**
+   * @default "p"
```
MichaelParadis, line 14: I think this default is incorrect. Looking at the code it would actually be `h<level>`

MichaelParadis: Or maybe not specifying a default here. It could be something like
```
Allows overriding of the element rendered. If one isn't specified it will be a heading element matching the specified level
```

mattander: 🤦‍♂️  I forgot to delete that when I grabbed it from Typography. Yes, let me resolve this.

MichaelParadis: We have all been there 😂

mattander: Okay! Should be good now.

mattander:
@MichaelParadis do you want me to also take this chance to migrate the test away from snapshots?

MichaelParadis (approved):
LGTM!
@mattander regarding your comment about snapshot cleanup. I don't think that is the best idea in this case because we are looking to verify exactly what is being rendered and a snapshot allows this. In the future we will look into visual regression testing and remove the snapshot tests

PR #1840, opened 2024-04-01
refactor(components): apply new theme to all components by default [JOB-87500]

chris-at-jobber:
## Motivations

As the new theme is being widely rolled out, moving retheme-scoped styling to be the only styling for components.

### Changed

- Updated snapshots
- Updated docs in one or two cases where relevant

### Removed

- Removed `:global(.jobber-retheme)` selectors and moved the respective styling onto the main un-scoped selectors
- Removed `inRetheme` checks from several TSX files

## Testing

View components without `jobber-retheme` class and verify they look the same as on `retheme-feature-branch` with the `jobber-retheme` class applied

For code review, meaningful changes are easier to get to if you apply these filters to remove generated files:
image

---

Random photo of Atlantis


packages/components/src/Button/Button.css
```diff
@@ -6,18 +6,20 @@
...
 
   --button--duration-loading: var(--timing-loading--extended);
+
+  --button--size: 40px;
 }
 
 .button {
   display: inline-flex;
-  min-height: calc(var(--base-unit) * 2.25);
+  min-height: var(--button--size);
```
chris-at-jobber, line 15: bye-bye mystery calcs

packages/components/src/Button/Button.css
```diff
@@ -30,23 +32,12 @@
...
-}
-
 .button * {
   color: inherit;
   fill: inherit !important;
   /* required to over-ride very specific fill from <Icon> */
 }
 
-:global(.jobber-retheme) .button * {
-  font-weight: 600 !important;
```
chris-at-jobber, line 44: Weight and letter-spacing now handled in the Typography component

packages/components/src/Button/Button.css
```diff
@@ -89,13 +76,6 @@ a.button:active {
   --button--color-variation--hover: var(--color-destructive--hover);
 }
```
chris-at-jobber, line 78: Learning and subtle now use the same underlying color

packages/components/src/Button/Button.css
```diff
@@ -159,21 +134,17 @@ a.button:active {
...
 
-:global(.jobber-retheme) .tertiary:hover,
-:global(.jobber-retheme) .tertiary:focus-visible,
-:global(.jobber-retheme) a.tertiary:hover,
-:global(.jobber-retheme) a.tertiary:focus-visible {
+.tertiary:hover,
+.tertiary:focus-visible,
+a.tertiary:hover,
+a.tertiary:focus-visible {
```
chris-at-jobber, line 147: @fagnanm do you want to revisit the "tertiary also have borders" decision in the scope of this? I could simplify this by grouping these selectors with the `secondary` styles defined above

packages/components/src/Button/Button.css
```diff
@@ -248,54 +200,29 @@ a.button.disabled:hover {
 }
 
 .onlyIcon {
-  width: calc(var(--base-unit) * 2.25);
+  width: var(--button--size);
   padding: 0;
 }
 
-.onlyIcon.small {
```
chris-at-jobber, line 255: Because we're re-setting the button size variable on the `.small` and `.large` classes, we don't need to redefine te width values here

packages/components/src/Button/Button.tsx
```diff
@@ -188,7 +188,7 @@ function ButtonInternals({ label, icon, size = "base" }: ButtonProps) {
 function getTypeSizes(size: string) {
```
chris-at-jobber, line 188: Since 2 of the three cases now return "base" do we even need this function anymore? Feels like a check on the Typography component could now suffice.

packages/components/src/Card/Card.css
```diff
@@ -64,17 +60,9 @@
 .header {
   display: flex;
   padding: var(--card--base-padding);
+  padding-bottom: 0;
   background-color: var(--card--header-background);
```
chris-at-jobber, line 64: We could in theory just delete this declaration and the variable for the card header background now.

packages/components/src/Checkbox/Checkbox.css
```diff
@@ -1,12 +1,11 @@
-.checkbox {
-  --checkbox-size: var(--space-base);
+.checkBoxParent {
+  --checkbox--size: 20px;
+  --checkbox--border--checked: var(--color-interactive);
```
chris-at-jobber, line 3: Setting these as variables gets rid of a ton of messy selector-nesting to manage state

packages/components/src/Checkbox/Checkbox.css
```diff
@@ -1,12 +1,11 @@
-.checkbox {
-  --checkbox-size: var(--space-base);
+.checkBoxParent {
```
chris-at-jobber, line 1: Renamed this class because reading things like `.checkbox .checkBox` was seriously hard to parse

packages/components/src/Checkbox/Checkbox.css
```diff
@@ -1,12 +1,11 @@
...
+  --checkbox--size: 20px;
```
chris-at-jobber, line 2: Bye bye mystery calcs

packages/components/src/Checkbox/Checkbox.css
chris-at-jobber: It doesn't necessarily show up in the diff because I didn't commit alternatives, but a lot of similar-looking states have varying levels of specificity, so I had found myself either declaring the same value multiple times for different selectors throughout the file, or disable linting and over-riding specificity warnings.

packages/components/src/Button/Button.css
fagnanm: @chris-at-jobber yes, I wrote a ticket up here for it https://jobber.atlassian.net/browse/JOB-92981

packages/components/src/Menu/Menu.tsx
```diff
@@ -284,12 +283,12 @@ function Action({
       onClick={onClick}
       ref={actionButtonRef}
     >
-      {icon && (
-        <span className={styles.icon}>
-          <Icon name={icon} />
-        </span>
-      )}
-      <Typography element="span" size="base" textColor="text">
+      {icon && <Icon name={icon} color="interactiveSubtle" />}
```
taylorvnoj, line 286: Do we want this color here?  What about those[ instances](https://jobber.atlassian.net/browse/JOB-87494) where we want the icon color to come through? With this set here, we won't be able to see those different icon colors in the menu.

taylorvnoj (commented):
I'm about half way - just wanted to put these comments up for now and I'll continue on...

packages/components/src/Banner/Banner.css
```diff
@@ -2,18 +2,13 @@
...
-}
-
-:global(.jobber-retheme) .banner {
   padding: calc(var(--space-base) - var(--space-smaller)) var(--space-base);
   border: var(--border-base) solid var(--color-border);
   border-radius: var(--radius-base);
+  color: var(--color-text);
   background-color: var(--color-surface);
+  gap: var(--space-small);
+  align-items: center;
```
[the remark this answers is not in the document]
chris-at-jobber: That is a great catch, Banner has a lot of React-side styling happening. Will get this cleaned up!

packages/components/src/Menu/Menu.tsx
chris-at-jobber: Matt actually split that out into a separate ticket but I might as well do it here!

packages/components/src/Drawer/Drawer.css
```diff
@@ -15,60 +11,49 @@
...
+  rather than using the breakpoint custom properties.
+  DO NOT COPY THIS - USE THE BREAKPOINTS DEFINED IN THE @jobber/design package!
   */
   @media (width >= 640px), print {
     position: relative;
     z-index: auto;
     width: var(--drawer-width);
+    border: var(--border-base) solid var(--color-border);
+    border-bottom-left-radius: var(--radius-base);
+    border-top-left-radius: var(--radius-base);
```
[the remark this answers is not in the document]
chris-at-jobber: There is a Figma but the best "source of truth" would be to do one of...
- turn on the retheme in the `useAtlantisConfig` file on master
- find the corresponding component in Jobber

I'll double check Drawer!

chris-at-jobber:
@taylorvnoj for your Checkbox example, this branch is correct. In some places components that had `radius-base` applied when it was 2px, the new 8px `radius-base` now looks like a circle, which we don't want. Or, more generally, some things got TOO ROUND with the new overall roundness, so we've had to dial them back (see Combobox and DataList for a few similar examples)

packages/components/src/Banner/Banner.css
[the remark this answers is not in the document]
chris-at-jobber: Fixed [here](https://github.com/GetJobber/atlantis/pull/1840/commits/5b9eb492051090f58a004afbe80d380a7f5012dd)

taylorvnoj:
@chris-at-jobber ohhhh that's right. Okay thank you for clarifying!

packages/components/src/Drawer/Drawer.css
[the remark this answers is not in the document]
taylorvnoj: Just checked Jobber, this is correct. It makes sense to not have the radius on the right as it is flush with the browser on the right side

packages/components/src/Banner/Banner.test.tsx
```diff
@@ -7,9 +7,11 @@ it("renders a success banner", () => {
   expect(container).toMatchSnapshot();
 });
 
-it("adds an error class to error banners", () => {
```
chris-at-jobber, line 10: We're no longer using classnames to do any changes to banner , so replaced this test with a test that does the same thing as the other tests. It's still a snapshot, which I know we don't love, but it seemed like a reasonable way to keep a similar level of coverage.

packages/components/src/Banner/Banner.tsx
```diff
@@ -103,17 +89,10 @@ export function Banner({
       </div>
 
       {dismissible && (
-        <button
-          type="button"
-          className={styles.closeButton}
+        <ButtonDismiss
```
chris-at-jobber, line 92: Since icon is no longer changing color, we can use the standard ButtonDismiss here now.

packages/components/src/Banner/Banner.css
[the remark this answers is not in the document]
taylorvnoj: We don't have to act on it because this conversation is out of scope for this PR, but I'm just noticing that the icon in the web banner is center aligned when there's multiple lines of text, and in the mobile banner the icon is at the top.  

Screenshot 2024-04-10 at 4 32 25 PM
Screenshot 2024-04-10 at 4 32 35 PM

packages/components/src/Drawer/Drawer.css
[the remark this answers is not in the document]
chris-at-jobber: Drawer did need a tiny tweak (same update to headings as Card and Modal) - updated [here](https://github.com/GetJobber/atlantis/pull/1840/commits/7e3ae6edb36a838a104ef936872df5c23f49ffee)

packages/components/src/Menu/Menu.tsx
chris-at-jobber: image

packages/components/src/Button/Button.css
```diff
@@ -248,54 +200,29 @@ a.button.disabled:hover {
...
 }
 
 .iconOnRight > *:first-child {
   order: 1;
+  margin-left: 0;
+  margin-right: calc(var(--space-smaller) * -1);
 }
 
-:global(.jobber-retheme) .hasIconAndLabel > *:first-child {
+.hasIconAndLabel > *:first-child {
```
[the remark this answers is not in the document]
chris-at-jobber: Should be [fixed now](https://github.com/GetJobber/atlantis/pull/1840/commits/707f43450b4d2651a8635636e560d2c79ea6cfca):
image

packages/components/src/Menu/Menu.tsx
chris-at-jobber: https://github.com/GetJobber/atlantis/pull/1840/commits/cb6de7f15a683f2ffa70924916ea3f3b30c44aaf

packages/components/src/Button/Button.tsx
chris-at-jobber: I was thinking something more like changing line 180 to

`size={"large" ? "large" : "base"}`

but that doesn't really change anything and leaving this how it is does give us a better outlet for managing label size going forward, so I think I will in fact leave this alone.

chris-at-jobber:
@taylorvnoj thanks for the review! I've either made changes or left comments explaining lack of changes on some of the more "discussion"-y points.

packages/components/src/Menu/Menu.tsx
taylorvnoj: Brilliant! Thank you for doing that here

packages/components/src/Button/Button.tsx
taylorvnoj: Perfect - sounds good to me!

taylorvnoj (approved):
LGTM - thanks for all this work Chris 🙏  Since this is going into a feature branch, we can do a bigger QA with a pre-release when that branch becomes a PR

PR #1843, opened 2024-04-02
fix(components): Update modal close button to stay fixed

MichaelParadis:
### Added

- 

### Changed

- components: Change modal dismiss button to not move when the header has multiple lines

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Verify the Close Button doesn't move when the header has multiple lines
image

---

Random photo of Atlantis


packages/components/src/Modal/Modal.css
```diff
@@ -95,11 +95,7 @@
 }
 
 .closeButton {
-  padding: 0;
-  border: none;
-  background-color: transparent;
-  cursor: pointer;
-  appearance: none;
+  align-self: flex-start;
```
MichaelParadis, line 98: This class wasn't used so I reused it for this

chris-at-jobber: Could you delete this class (and the wrapper in the TSX) and change the `align-items` [on the .header](https://github.com/GetJobber/atlantis/pull/1843/files#diff-155f6ca6b54ece4c9503ddeaa6a6d780bffb49359c5eb972356b1ce3a96da238R74) to `start`?

chris-at-jobber: ooooh I see why you can't, never mind me 😄

fagnanm:
@chris-at-jobber you okay with the title/close being slightly misaligned on modals until the re-theme lands for everyone? 2 weeks.

chris-at-jobber:
Is the misalignment in the current theme caused by the specific pixel values being used? The flex solution that was in the earlier commits of this PR seemed to look fine in both themes.

fagnanm:
> Is the misalignment in the current theme caused by the specific pixel values being used? The flex solution that was in the earlier commits of this PR seemed to look fine in both themes.

@chris-at-jobber can make both alignments work but with the retheme official launch next week wondering if it is worth it? Or just roll forward with it being misaligned on old until the retheme is fully launched.

chris-at-jobber:
It's probably fine for that ~1week, I'm just wondering what the problem was with the initial flex solution because it seemed to look as intended in both themes

fagnanm:
> It's probably fine for that ~1week, I'm just wondering what the problem was with the initial flex solution because it seemed to look as intended in both themes

@chris-at-jobber wasn't vertically aligned with the text. Wasn't center-aligned with the text anymore but then just sat above the x-height.

chris-at-jobber:
Ah I see what you're saying, yeah this is fine. I think with the [ESM rollout](https://github.com/GetJobber/atlantis/pull/1849) happening today this change likely won't land in Jobber until tomorrow anyways so one less day of misalignment for the few accounts still on the old theme.

chris-at-jobber (approved):

PR #1844, opened 2024-04-02
feat(components): adds ability to provide "preferredPlacement" for Tooltip

ericchernuka:
## Motivations
There are instances where being able to influence the placement of a tooltip is desired such as on a navigation elements within the shell of a UI. By adding a `preferredPlacement` consumers are able to influence the placement depending on their contextual needs. By default it will use `top` like it always has.

### Added
- Adds ability to provide `preferredPlacement` to Tooltip component.

---

Random photo of Atlantis


darryltec (approved):
:shipit:

PR #1845, opened 2024-04-03
feat(components-native): Adjust menu visibility on iOS

jdeichert:
## Motivations

On iOS, when clicking a `<Menu>` button (or `customActivator`), the menu's position is wrong if the keyboard happens to be visible. This leads to the menu being placed _behind_ the keyboard in some cases.

The solution we came up with is to dismiss the keyboard, wait for it to become fully hidden, and then show the menu. This allows the menu position to be calculated correctly. 

Notably, when opening the menu on android, the menu's position correctly takes the keyboard height into account and therefore doesn't need any changes.

Also, 2 design changes are included (as requested): adjusted the left/right positioning and the menu's radius.

### Changed

- When opening the menu on iOS we dismiss the keyboard, wait for the `keyboardDidHide` event, and then show the menu
- Adjusted menu left/right positioning logic to remove padding
- Adjusted Menu radius

## Testing

> See comments for screenshots/videos of changes.

Because this is the first official usage of `<Menu>` in JM, you'll have to manually create it in some view to verify its behaviour. Testing in storybook will not represent the platform-specific (iOS) behaviour.

1. Build the atlantis package, link to it from JM
2. Create a new `<Menu>` in some view
3. Try the menu in various positions: float left, float right, in top half of screen, in bottom half of screen
4. Click the menu activator/button to open it, verify that the menu position makes sense
    * On both iOS and Android, the menu should open immediately because the keyboard should not be visible
5. Do the same process, but focus on an `<InputText>` (to open keyboard) before you click the menu button.
    * iOS: the keyboard should dismiss, become fully hidden, and then the menu should open
    * Android: the menu should become visible immediately

---

Random photo of Atlantis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45714,10 +45714,10 @@
     },
     "packages/components": {
       "name": "@jobber/components",
-      "version": "4.90.5",
+      "version": "4.90.6",
       "license": "MIT",
       "dependencies": {
-        "@jobber/design": "^0.56.3",
+        "@jobber/design": "^0.56.4",
         "@jobber/formatters": "*",
         "@jobber/hooks": "^2.9.4",
         "@popperjs/core": "^2.0.6",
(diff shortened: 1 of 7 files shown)


chris-at-jobber (commented):
Design approved ✅ 

As a question, do we need to add the loading state to the activator _here_, or is that an in-product implementation detail?

JeffMarvin (approved):
Nice stuff, LGTM!

PR #1849, opened 2024-04-03
feat(components): BREAKING CHANGE: Updating Atlantis to support ESM + Injected CSS

scotttjob:
## Motivations

This is the ESM + CSS Update for SSR/Vite/PNPM support that has been talked about for a while now.

We've combined two impactful build changes into one, and marking this as a major version update to denote the change.

Moving forward, to import Atlantis Styles you will have to import `@jobber/components/styles` or `@jobber/components/dist/styles.css` into your project depending on if you're in an ESM or CJS environment respectively.

In addition, we've included an ESM build of Atlantis for the first time. This is now provided in addition to our CJS build. A tremendous amount of effort has been taken to ensure that this change results in as little disruption as possible, but issues are always possible. If something breaks in your repo as a result of this change, please let us know.

The only notable changes that were found during testing

1. if you were previously accessing a file under the 'dist' folder that is not explicitly provided in the package.json exports, that import will fail. `@jobber/components/dist/Some/Random/File` is no longer possible. If you are attempting to use a file not exported by the system, please open a PR with the additional export as a change instead.
1. If you were previously using double slashes (//) to get around an eslint rule preventing importing a Private Atlantis package, that will no longer work for the same reason as above. Files have to be present in the package.json exports array now to be properly accessed.

## Changes

- ESM Build
- Styles are now Extracted instead of Injected.
---

Random photo of Atlantis


MichaelParadis (approved):
LGTM!
All the changes from the PR that was reverted are present.

PR #1855, opened 2024-04-04
docs(components-native): Mark mobile Menu component as alpha

jdeichert:
## Motivations

We have determined that the mobile Menu component is not ready for prime time and needs work to handle various overflow cases. We've decided to mark it as _alpha_ until we can invest more time in it.

### Changed

- Mark the mobile Menu component as alpha

## Testing

Open storybook and you should see this little alpha tag:

Screenshot 2024-04-04 at 1 41 11 PM


chris-at-jobber (approved):
Looks good! For future I would love to revisit where Menu actually sits (this is not a navigation component) but that is a question for another day 😄

PR #1856, opened 2024-04-04
feat(components): Add Toolbar to InputText

ZakaryH:
## Motivations

we want to enable a place to slot in some content inside of an InputText

## Changes

- modified how we position the clear button & loading indicator on InputText
- modified the textarea (multiline) to that the resizable element is actually the wrapping container, not the text area itself
- styles and some resizing logic adjusted to account for the above

### Added

`toolbar` and `toolbarVisibility` props that can render a React component (node) inside the InputText. almost like a footer.

Multiline usage with placeholder
Kapture 2024-04-05 at 16 02 56

Single line usage with placeholder and clear button
Kapture 2024-04-05 at 16 06 09

### Fixed

- 

### Security

- 

## Testing

unfortunately these changes are touching the core `FormField` `input` case, so please do give most `Input____` components a quick try. even if the tests are asserting that everything is working as intended, many of the things I had to address were based on being in a bad spot visually so we'll need to inspect most of it ourselves to make sure it looks good!

**things to look out for**
- double scroll bars (unless you've minimized it to a tiny, not that useful size)
- gradient showing up on elements it should not, or obstructing new text as you type it in a multiline textarea
- text area growing ONLY if you have provided the `rows: { min: 1, max: 6 }` prop, and it should STOP growing once you hit that maximum*
- position of clearable button, position of loading indicator
- position of placeholder text & the mini label once you type in the input
- resize handle isn't too close to a button if you provide right aligned buttons to the `toolbar`
- resize handle isn't obscured on Firefox
- single line input feels & looks good
- focus behavior feels good when using toolbar
- resizing the content & then attempting to scroll should always work

*there's some existing behavior here that I don't necessarily love where if you manually drag the container to be bigger, and then type in it - it will snap back to a smaller size

---

Random photo of Atlantis


packages/components/src/FormField/FormFieldWrapper.tsx
```diff
@@ -58,6 +59,7 @@ export function FormFieldWrapper({
         (placeholder && type === "select") ||
         // Naively assume that if the the type is tel, it is the InputPhoneNumber
         (placeholder && type === "tel"),
+      [styles.text]: type === "textarea" || type === "text",
```
ZakaryH, line 62: moved the label styles that now apply to both text and text area into a new style of `text`

ZakaryH:
last commit adds `relative` to the wrapping fragment (now div) around the `input` so that the loading indicator which translates 50% goes to the middle properly since it'll be relative to the correct element now regardless if the toolbar is present, making the wrapping container much taller

the clearable button was being positioned in a different way that was part of the document flow, and resulted in pushing the toolbar content over when it was present. it was also getting centered on the entire container and as such was in the wrong spot when toolbar present. I've changed how that is positioned to be absolute + translateY, so that it can be positioned with that same `relative` around the input

this unfortunately does touch every single input implementation and so had to update many snapshots

ZakaryH:
one thing I noticed is that the tab order when you have a clearable inputtext + a toolbar is not ideal

it'll cycle thru the buttons first, and then get to the clear button

packages/components/src/FormField/FormField.css
```diff
@@ -44,7 +44,6 @@
 
   display: flex;
   width: 100%;
-  height: var(--field--height);
```
ZakaryH, line 47: this was restricting the height of the container that can now be a lot more than just the input field itself

I'm pretty sure this is safe to remove, but I'm really not too sure what the initial need was for it since presumably you'd be able to let the content determine how tall it needs to be from the get go

ZakaryH:
hmm clicking the toolbar itself counts as not focusing the input. I'm not sure that's right. it definitely doesn't feel great if you misclick

actually, even if you click the button itself that also loses focus. I don't think that's right.

ZakaryH:
not totally sold on the animation. on the way in it's fine, but on the way out it's hitching a bit

packages/components/src/FormField/FormFieldWrapper.tsx
```diff
@@ -125,7 +135,31 @@ export function FormFieldWrapper({
...
+          <div className={styles.childrenWrapper}>
+            {children}
+            <AnimatePresence>
+              {isToolbarVisible && (
+                <motion.div
+                  key="toolbar"
+                  initial={animationInitial}
+                  animate={{
+                    opacity: 1,
+                    height: "50px",
```
ZakaryH, line 148: this was for testing, give it a proper value

ZakaryH:
these `tabIndex={-1}` on `div` elements is to prevent clicks on them from collapsing the toolbar because we're focused on another element

I think it's from event bubbling and the fact that they can't accept focus themselves, but when those elements are clicked - it's registering the body as the `activeElement` rather than them - which is of course not contained in the wrapper div, and so it was setting focused to false

docs/components/InputText/Web.stories.tsx
```diff
@@ -27,6 +28,31 @@ Multiline.args = {
...
+    >
+      <Button label="Rewrite" size="small" icon="sparkles" fullWidth={false} />
+      <Button
+        ariaLabel="Undo"
+        size="small"
+        icon="redo"
+        type="tertiary"
+        fullWidth={false}
+      />
+    </div>
```
darryltec, line 52: i think we can just put them side by side instead of very far away. UX wise, that's better

packages/components/src/FormField/FormField.css
```diff
@@ -326,3 +337,22 @@
 .description {
   margin-top: var(--space-smaller);
 }
+
+.toolbar {
+  display: flex;
+  position: relative;
+  margin: var(--space-small) var(--field--padding-right) var(--space-small)
```
darryltec, line 344: Ah, this is the reason why that animation looks a bit choppy. it doesn't account for the margin. The `height: 0` on he animation doesn't remove the margin from that element.

packages/components/src/FormField/FormField.tsx
```diff
@@ -150,7 +150,7 @@ export function FormField(props: FormFieldProps) {
               );
             default:
               return (
-                <>
+                <div style={{ position: "relative" }}>
```
darryltec, line 153: what's this one for?

packages/components/src/FormField/FormFieldWrapper.tsx
```diff
@@ -125,7 +134,32 @@ export function FormFieldWrapper({
           )}
 
           {prefix?.label && <AffixLabel {...prefix} labelRef={prefixRef} />}
-          <div className={styles.childrenWrapper}>{children}</div>
+
+          <div className={styles.childrenWrapper} tabIndex={-1}>
+            {children}
+            <AnimatePresence>
+              {isToolbarVisible && (
+                <motion.div
```
darryltec, line 142: Oh i just realized something. This toolbar shouldn't be here actually. It should be on below everything else as it will mess with the positioning of prefix and suffix.

Moving it down like components-native should fix the tabbing order tho

Web

image


Components native

image

packages/components/src/FormField/FormFieldWrapper.tsx
```diff
@@ -125,7 +134,32 @@ export function FormFieldWrapper({
...
+                  animate={{
+                    opacity: 1,
+                    height: "auto",
+                  }}
+                  exit={{ opacity: 0, height: 0 }}
+                  transition={{
+                    duration: tokens["timing-base"] / 1000,
+                    ease: "easeInOut",
+                  }}
+                  tabIndex={-1}
```
darryltec, line 154: weird that you had to tabindex -1 it when divs aren't tabbable be default 🤔

packages/components/src/FormField/hooks/useToolbar.ts
```diff
@@ -0,0 +1,25 @@
...
+  isToolbarVisible: boolean;
+  animationInitial: boolean | { opacity: number; height: number };
+}
+
+export function useToolbar(props: UseToolBarProps): UseToolbar {
+  const isToolbarVisible =
+    props.toolbar !== undefined &&
+    (props.toolbarVisibility === "always" || props.focused);
+  const animationInitial =
+    props.toolbarVisibility === "always" ? false : { opacity: 0, height: 0 };
```
darryltec, line 19: would passing in `initial={true || false}` in the animated presence do what this is trying to do?

darryltec (changes requested):
bit of an issue with prefix prop that i think we could solve with another DOM structure

docs/components/InputText/Web.stories.tsx
ZakaryH: fair, but I wanted to make sure to have an example that had a button in the bottom right because that's a design we've seen, and I had worries about how close it would end up being to the resize handler

so having an example with that layout lets us try it out

darryltec: oh, that makes sense!

packages/components/src/FormField/FormField.tsx
ZakaryH: I meant to move this into a style rather than doing it inline, so it shouldn't be here but as for why it's there at all:

https://github.com/GetJobber/atlantis/pull/1856#issuecomment-2038391680

packages/components/src/FormField/FormFieldWrapper.tsx
ZakaryH: when you say "everything else" what do you mean? there are a few different levels 

- outside the entire `containerClasses` div
- outside the `wrapperClasses` div
- outside the `inputWrapper` div?

packages/components/src/FormField/FormFieldWrapper.tsx
ZakaryH: agreed. I'm not 100% sold on this being the best fix, and it barely even makes sense - but it does fix the problem

without this, clicks on these elements combined with the focus event listener would unset `focused` because the clicks would seemingly go thru the elements and end up on the body somehow

packages/components/src/FormField/hooks/useToolbar.ts
ZakaryH: I'm not sure I follow. `true || false`  that's just always going to be `true` no?

also, are you asking to swap to `AnimatedPresence`? because this is using `AnimatePresence` from Framer, not the Atlantis component

on that note, those names are incredibly similar

darryltec: oh no, animate presence. just saying there's a prop to take out the initial animations on any motion children haha

ZakaryH: ah gotcha. I'll take a look, this is my first time using the API so definitely not sure what the best way to manage some of this stuff is

packages/components/src/FormField/FormFieldWrapper.tsx
darryltec: thinking somewhere inside the wrapperClasses

packages/components/src/FormField/FormField.css
ZakaryH: ah nice catch. removing that does make it feel smoother 100%

however then I need to add the margin in the animation, which is giving some interesting results of jumping around in a new way!
Kapture 2024-04-09 at 11 13 44

darryltec: i wonder if the margin just needs to not be on the animating element. would adding another wrapping div help?

packages/components/src/FormField/FormField.css
```diff
@@ -282,19 +308,6 @@
...
-     * disabling property-no-unknown here as this is a real property
-     * that stylelint just doesn't know about. We may be able to remove
-     * this when we update our stylelint to not care about IE11.
-     */
-    /* stylelint-disable-next-line property-no-unknown */
-    aspect-ratio: 1/1;
-    padding: 0;
-  }
-}
-
```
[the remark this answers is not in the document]
ZakaryH: Kapture 2024-04-10 at 14 42 53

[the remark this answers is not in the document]
ZakaryH: this is the visual difference on Chrome between the two approaches

**without the aspect ratio**
image


**with the aspect ratio**
image

different sizes might have different appearances. I'll have to try that out.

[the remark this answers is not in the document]
ZakaryH: looking at those two images. I can't really tell the difference. I'm unclear as to why we'd do it two different ways if they both do the same thing. personally I think consistency and simplicity are preferable, assuming I'm not missing something it was providing

packages/components/src/FormField/FormField.css
ZakaryH: looking back at this comment now is funny

how little I knew about how important that height was

packages/components/src/FormField/FormField.css
ZakaryH: moved the margin to a separate div and let the animation be only responsible for height/opacity. feels much better now.

packages/components/src/FormField/FormFieldWrapper.tsx
ZakaryH: this was a wild ride but I think it's done with the prefixes not totally busted (for single line, multiline is already borked)

packages/components/src/FormField/FormField.css
```diff
@@ -326,3 +339,21 @@
...
+  display: flex;
+  position: relative;
+  margin: var(--space-small) var(--space-base) var(--space-base) var(--space-base);
+}
+
+.toolbar::before {
+  content: "";
+  display: block;
+  position: absolute;
+  top: calc(-1.25 * var(--space-base));
```
ZakaryH, line 353: this 1.25 is a bit magical

I don't even remember why I chose that value anymore. could have simply been this is when it started showing up

packages/components/src/FormField/FormField.css
```diff
@@ -66,6 +66,13 @@
   box-shadow: var(--shadow-focus);
 }
 
+.horizontalWrapper {
```
ZakaryH, line 69: I don't love this name. it's not very descriptive

container, wrapper, inputWrapper, childrenWrapper and a couple other names were already taken was the problem but there's gotta be something better I can come up with

packages/components/src/FormField/FormField.tsx
ZakaryH: we no longer need this because we're not trying to place the toolbar in the same container as the input, and so we can rely on the `relative` provided by the inputWrapper I believe

ZakaryH:
nice. tests are passing locally but not on CI 🙃 

image

packages/components/src/FormField/components/ClearAction.css
```diff
@@ -1,12 +1,6 @@
-.clearContainer {
-  display: flex;
-  align-items: center;
-  padding-left: var(--space-small);
-}
-
 .clearInput {
   display: flex;
-  position: relative;
+  position: absolute;
```
MichaelParadis, line 3: I noticed something with clearable on long text in a single line input:
On this branch:
image

On master:
image


I also noticed the Clearable will overlap with the loading if they are both present (not this happens even if the toolbar prop isn't used).
image

MichaelParadis: I also noticed this when the InputText is wrapped in a div with a max width of 50%. I found changing this absolute to relative "fixed" it but I haven't fully caught up on the context of why that change was made so there is probably other things to consider
image

MichaelParadis: More for me to test afterwards but this also happens with InputPassword's that are clearable and have visibility enabled
image

InputNumber and the browser default stepper on firefox (may be the same fix in general)
image

MichaelParadis: Also focused with a suffix can overlap (the same as InputPassword but more general. This also happens with Icon
image

MichaelParadis (changes requested):
The only real issue I saw was relating to the ClearButton rendering.
See: https://github.com/GetJobber/atlantis/pull/1856/files#r1568863248 for more information

packages/components/src/FormField/FormField.css
```diff
@@ -85,11 +92,17 @@
   --field--padding-top: calc(var(--space-large) + var(--space-smaller));
 }
 
-.wrapper.textarea {
+.wrapper.text {
   --field--placeholder-offset: 0;
   --field--placeholder-transform: 0;
-  height: auto;
+  /* flex-direction: column; */
```
MichaelParadis, line 98: Minor thing but I notice some of the CSS is commented out. It is probably worth removing these unused styles

ZakaryH: good catch. yeah I went through many iterations on the CSS and it kinda shows. I'll clean those up!

packages/components/src/FormField/components/ClearAction.css
ZakaryH: ugh yeah I see what you mean

clearable being positioned absolutely is because previously it would center itself vertically, and that worked because it was only ever as tall as the single line input. now that the single input can be taller depending on whatever you've given it as a tool bar, we want it to be top right - not centered on the overall height

worth nothing clearable + multiline is broken on master right now. that and a bunch of other props (prefix, suffix, loading) don't play very well with multiline. unfortunately the underlying reason they don't work is because of height and toolbar adds height 😭 even to the single line

that long input thing happens today with loading, because it's absolutely positioned so yeah that tracks that it now happens with clearable

image

ZakaryH: overall the problem is that so many of these things (suffix, affix, loading, clearable, password eye thing) are all built assuming a set height

once that height changes, like it does with the toolbar, they break.

they shouldn't be centered, they should be top right/top left -ish and that's where the absolute comes in, but yeah as we're seeing that has a lot of side effects

ZakaryH: master, InputNumber + loading

you can barely see it but yeah that same overlapping control problem

image

packages/components/src/InputText/InputText.tsx
```diff
@@ -30,6 +30,8 @@ interface BaseProps
       | "defaultValue"
       | "prefix"
       | "suffix"
+      | "toolbar"
+      | "toolbarVisibility"
```
ZakaryH, line 34: do these make sense here? like yeah it makes it easy to pass them down to the underlying FormField because it knows about them, but these are input text specific

slightly worried that if another component directly uses `FormFieldProps` without `Pick`ing, then it'll inherit these things that won't make sense

chris-at-jobber: the pattern for at least the existing Input{Thing} is to `Pick` so I think you're good?

assuming 
- if someone makes a new component that pulls FormField in, they'd go "no toolbar for me please" when picking in that usage
- the `Pick` pattern is something we can either reliably catch in reviews, or have tooling around?

ZakaryH:
@MichaelParadis ok I think swapping to `relative` fixed everything. the whole thing I mentioned about `absolute` wasn't actually correct. I think it might have worked like that at one point, but we're no longer using that approach and this seems to be good now in all the scenarios you mentioned

image

image

this is just with a non `multiline` since many props don't work with that

packages/components/src/InputText/InputText.tsx
MichaelParadis: One thought we could have is a `FormFieldInternalProps` that we don't export when publishing the package. We could then expose all the other FormField props we currently do. I don't think this is something we should do because `FormField` is very much an internal component that you should know what you are doing if you are going to use it outside of Atlantis. For example it is marked private
image

chris-at-jobber:
@ZakaryH I will have a design review for ya by EOD

chris-at-jobber:
Any additional notes to follow, but recording as I come across things:

---

Non-blocking because there is already a lot happening here but I would appreciate if the left side of the  toolbar respected the size property to keep any left-aligned actions aligned when the input is large or small

Base
image

Small
image

Large
image

chris-at-jobber:
Is it intended that even when there is _no_ max set on rows, that the input does not grow? Side by side of this implementation vs what we've already implemented:

In Atlantis, with and without toolbar
https://github.com/GetJobber/atlantis/assets/39704901/8c235af3-84c6-44e3-9d9c-c58bdb773ea3

In Jobber with toolbar
https://github.com/GetJobber/atlantis/assets/39704901/fbaac3e0-7d56-4772-8e31-25920b8365ae

ZakaryH:
>Is it intended that even when there is no max set on rows, that the input does not grow? Side by side of this implementation vs what we've already implemented:

@chris-at-jobber  I am really not too sure. there are a few things about that prop which behave, in my opinion, oddly 

reading the documentation, it suggests that only having a `min` is not valid, and that it should contain a min & max. it just doesn't seem to be enforced. I think I know what's happening code-wise, and yeah it's not setup to work with only 1 present. min likely causing more issues than a max because it will try to compare them.

>Specifies the visible height of a long answer form field. Can be in the form of a single number to set a static height, or an object with a min and max keys indicating the minimum number of visible rows, and the maximum number of visible rows.

this same behavior can be seen on master right now. if you only give it a min, it won't grow. similarly if you don't give it a `rows` at all, it also won't grow. 

is this right? I'm not sure, but that is how it works today. so in practice, setting a min/max also enables an auto grow behavior that isn't really mentioned.

that's interesting that your other example without a max rows is growing. I haven't seen that happening in Atlantis. I'll have to see what's going on there.

chris-at-jobber (commented):
Design approved ✅ 

There is some funkiness around how the resizing works, but discussed "offline" with @ZakaryH and determined that our existing implementations in Jobber will be compatible with the approach in this PR.

ZakaryH:
@chris-at-jobber pushed up a change to address the `size` prop thing

@MichaelParadis up to you if you feel it's worth a re-approval from the dev side. simply used the spacing variables that get set by the size classes

MichaelParadis:
@ZakaryH I just looked at the change and it makes sense to me. Github didn't dismiss my review but if you need an approval let me know

chris-at-jobber:
Beauiful, thanks @ZakaryH ! Design approval still stands, and has grown only more solid.

PR #1857, opened 2024-04-08
feat(components): Update Headings to match retheme design

MichaelParadis:
### Added

- 

### Changed

- Updated Heading styling to match the new retheme styles

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Import pre-release into Jobber Online. While in the retheme verify the Heading styling matches for each of the levels.
Open a modal and verify that the heading on the modal also matches.
You can use code from the Heading  story and Modal web story to quickly get all the components on the same view

Note: Modal should be an H2 with H2 stylings. See: https://github.com/GetJobber/atlantis/pull/1857#discussion_r1556071137

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Modal/Modal.css
+++ b/packages/components/src/Modal/Modal.css
@@ -81,7 +81,7 @@
 
 :global(.jobber-retheme) .header h3 {
   font-family: var(--typography--fontFamily-normal);
-  font-size: var(--typography--fontSize-largest);
+  font-size: var(--typography--fontSize-larger);
 }
 
 /**
--- a/packages/components/src/Modal/Modal.css
+++ b/packages/components/src/Modal/Modal.css
@@ -79,11 +79,6 @@
   background-color: transparent;
 }
 
-:global(.jobber-retheme) .header h3 {
-  font-family: var(--typography--fontFamily-normal);
-  font-size: var(--typography--fontSize-larger);
-}
-
 /**
  * Ensure there's no extra padding top on the next element. This mostly negates
  * the <Content /> padding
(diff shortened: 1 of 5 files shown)


packages/components/src/Modal/Modal.css
```diff
@@ -81,7 +81,7 @@
 
 :global(.jobber-retheme) .header h3 {
   font-family: var(--typography--fontFamily-normal);
-  font-size: var(--typography--fontSize-largest);
+  font-size: var(--typography--fontSize-larger);
```
chris-at-jobber, line 84: Just delete this declaration and replace the Heading in the Modal tsx with level 2!

chris-at-jobber: I have this done [here](https://github.com/GetJobber/atlantis/pull/1840/files#diff-2a52591e6f55264bf76c5b2f7ace34e83339e786010d5ec3800069f2208e8f3cR117) if you want to reference

MichaelParadis: H2 would have a larger font size, however I don't think this is needed anymore since styling of the base header should match. I am going to verify quickly before pushing the change

MichaelParadis: Based on this I think that should be a level 2
image

packages/components/src/Modal/Modal.css
```diff
@@ -79,11 +79,6 @@
   background-color: transparent;
 }
 
-:global(.jobber-retheme) .header h3 {
```
MichaelParadis, line 82: Uses the base H3 sizing

packages/components/src/Modal/Modal.css
chris-at-jobber: Yes so I think we want `Heading level={2}` to be used here, instead of specifying a `Heading level={3}` and then injecting styling here to make it _look like_ a `Heading level={2}`

MichaelParadis: I see, I will test my change locally and push it up so semantically it will be an H2 with H3 styling

MichaelParadis: Sounds good, I will update it

chris-at-jobber (approved):
Looks good, thanks for the setup assistance with the config file!

image

image

PR #1859, opened 2024-04-10
feat(components): Prevent displaying dropdown when only 2 sorting options for Datalist

jiama843:
## Motivations

We want to be able to change the naming of our options (without defaulting to A-Z), however we don't want to have a sorting dropdown if there are only 2 custom elements.

### Added

- 

### Changed

- When we only have 2 sorting options for a data list header, we don't display the dropdown

https://github.com/GetJobber/atlantis/assets/24597281/fb3de6c7-68ea-40e8-93e7-925de1a18a99

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Check that the data list with only 2 options doesn't display the dropdown.

Can test locally by removing a couple of sortable options from `docs/components/DataList/Web.stories.tsx`

---

Random photo of Atlantis


MichaelParadis:
I am noticing that the onSort call back isn't being called when the drop down isn't displayed because there are only two options. Is looks like the test doesn't check this either
https://github.com/GetJobber/atlantis/pull/1859/files#diff-68e76e139eaf82eaf2c574e854ba953077924fb6325c38d8f2c31fc3e74ecefbR446


https://github.com/GetJobber/atlantis/assets/17303106/c607bd1a-dbb1-4572-8fd4-3817ea462b86

@jiama843

MichaelParadis:
@jiama843 the issue seems to be related to the drop down not being present but it is trying to open in. I think not trying to open the dropdown if there are only 2 options will resolve the issue
https://github.com/GetJobber/atlantis/blob/3ffcbf303b2f64242160a02f545d11fbc78897c7/packages/components/src/DataList/components/DataListHeaderTile/DataListHeaderTile.tsx#L104-L109

packages/components/src/DataList/DataList.test.tsx
```diff
@@ -411,46 +414,83 @@ describe("DataList", () => {
...
+              ],
+            },
+          ],
+          onSort: mockOnSort,
+          state: undefined,
+        });
+
+        fireEvent.click(screen.getByText(mockHeaders.name));
+        expect(screen.queryByText("Ascending")).not.toBeInTheDocument();
+        expect(screen.queryByText("Descending")).not.toBeInTheDocument();
```
MichaelParadis, line 446: This should probably test that the `mockOnSort` is called

MichaelParadis:
@jiama843 this was implemented in https://github.com/GetJobber/atlantis/pull/1883 and JO has been updated with this change. I am closing this since it is no longer needed. Thanks again for your help with this :)

PR #1866, opened 2024-04-16
fix(components): SSR Fixes

scotttjob:
## Motivations

We have some SSR related issues.

A recent update to Tooltip treated a Ref as a Boolean, so instead of flipping between true/false it was permanently true (which breaks in SSR environments)

In addition, it was noted that the Toast component when loaded via the ESM barrel file broke client-side functionality in certain framework environments. This is because the Toast component currently creates a React Root when the `showToast` function is imported, for compatibility in environments that aren't full SPAs. That createRoot call was conflicting with framework `hydrateRoot` calls and preventing all client-side interaction.

Some effort was taken to see if we could update the Toast component in a way that would support both island based loading and full SPA loading. Working solutions ran into issues with local (and remote) tests, and the timebox expired to get the tests in a passing state. A follow-up issue has been created to find a solution that fits into both worlds and keeps our tests happy as well.

So the solution for now is to remove the Toast from the ESM barrel file, and only import it directly (via a dynamic import in certain framework environments). This solves the immediate issue while leaving more room for a holistic, toasty solution in the future.

## Changes

- Removing Toast from the ESM Barrel File (follow-up issue created to refactor Toast)
- Modifying a recent change to Tooltip to account for the Ref in useIsMounted.

### Changed

- Toast
- Tooltip

### Fixed

- SSR Issues (document not available in non-browser environments, Tooltip causing hydration issues)

## Testing

- Should clear CI with no issues.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Toast/showToast.tsx
+++ b/packages/components/src/Toast/showToast.tsx
@@ -15,23 +15,29 @@ import { Toast, ToastProps, ToastRef } from "./Toast";
 import styles from "./Toast.css";
 
 const targetId = "atlantis-toast-element";
-let target = document.querySelector(`#${targetId}`);
+let target =
+  typeof document !== "undefined" && document.querySelector(`#${targetId}`);
 
-if (!target) {
+if (typeof document !== "undefined" && !target) {
   target = document.createElement("div");
   target.id = targetId;
   target.classList.add(styles.wrapper);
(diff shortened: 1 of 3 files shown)


darryltec (approved):
can confirm, ESM don't get angry anymore and still fires on CJS

PR #1868, opened 2024-04-17
fix(components): remove Chip prefix background [JOB-92493]

chris-at-jobber:
## Motivations

In initial development of Chip, we missed that only suffix should have the `surface` background. This leads to unexpected results when using in-product.

image

### Fixed

-  white (surface) background color from prefix, now only applied to suffix

## Testing

See the Prefix story

---

Random photo of Atlantis


taylorvnoj (approved):
🙏 thank you! LGTM!

PR #1869, opened 2024-04-19
feat(components): Add autofocus support for InputText (JOB-94181)

sabidhasan:
## Motivations
We want our input text to have autofocus as part of design requirements, for the chat drawer. The Native component supports the autofocus prop, but the HTML web version does not. This PR adds support for this prop for the web component.

### Added
- Added `autoFocus` to `InputText`

## Testing
Added tests for `InputText` to assert that focus is given when the new prop is used.

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -44,7 +44,7 @@ export function FormField(props: FormFieldProps) {
     onValidation,
     onKeyUp,
     clearable = "never",
-    autoFocus,
+    autofocus,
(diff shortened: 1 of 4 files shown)


NickBoers-Jobber (approved):
LGTM.

ZakaryH:
nice

I was thinking we could have this apply to more than InputText since basically every other kind of input could have a similar use case, but I'm fine with having this scoped down to just what you need for now. extending it won't be much work if we need it elsewhere.

PR #1871, opened 2024-04-23
fix(components): update small button sizing

chris-at-jobber:
## Motivations

The Atlantis button updates in the retheme looked OK for small buttons _with a label_, but were TOO SMALL when just an icon:
image

This is considerably smaller than the [Figma spec](https://www.figma.com/file/rIIhulZvcp9M82lNOCGv16/Product%2FOnline?type=design&node-id=14531%3A17&mode=design&t=ij64MZJIw7uMAvcE-1) calls for.

### Fixed

- Small buttons are now 32 by 32px when they only have an Icon, and are 32px high when they have a label

## Testing

Set up a small button with an icon and no label in Storybook

---

image

### DIFF
--- a/packages/components/src/Button/Button.css
+++ b/packages/components/src/Button/Button.css
@@ -180,8 +180,8 @@ a.button.disabled:hover {
 /* Sizes */
 
 .small {
-  --button--size: 24px;
-  padding: var(--space-small) calc(var(--space-base) - var(--space-smaller));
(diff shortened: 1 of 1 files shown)


MichaelParadis (approved):
LGTM! The size of the button is changed as expected. I do think is possible that a few usages may have layouts affected but `size="small"` isn't used often let alone small with only an icon

PR #1873, opened 2024-04-24
feat(design): overhaul the base color palette

chris-at-jobber:
## Motivations

As we refine the systemic impacts of the new theme, it's become noticeable that we're reaching outside of our base palette or adding one-off values in ways that aren't particularly systemic, to account for 
- limited per-"color" ranges (lightest, lighter, light, base, dark)
- lack of full ranges for new colors such as taupe and "request", "quote", "invoice"
- subtle differences at the low-contrast end of ranges

We also have palette ranges with considerable overlap. Notably, `grey`, `greyBlue`, and `blue` have several tones that could be indistinguishable from one another in isolation.

image

In addition, our existing palette is somewhat warped in scale on a per-color basis as the "base" value that the others are based on varies wildly in lightness/contrast. For example, `blue` and `yellow` are both the base values for their respective ranges, but do not have comparable contrast with the general application's `surface` value. 

image

Finally, usage of the base palette should be easy to flag in production. Instead, it requires careful manual inspection such as "search for `--color-` and look at all of the instances you find to see if they use, for example, `--color-green` instead of `--color-interactive`".

### Added

- New base value range
- `-base-` token name modifier to indicate that the value is from the base palette

### Changed

- Underlying values to hsl for easier parsing of relative lightness and saturation within a given hue range

### Deprecated

- Pointed current base palette values to their new counterparts

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis


chris-at-jobber:
## Thoughts for feedback

- Does the `--color-base-` modifier make sense? `Base` is a concept we use to refer to props in terms of the "default" so meaning-wise this might be confusing. The industry standard is generally "primitives" to my knowledge

- Any concerns moving to HSL? This is primarily for my own preferences in terms of readability and ability to tweak lightness in an intuitive manner, but it seems like HSL usage is growing more prevalent across the board over RGB

- Is the deprecation model for the existing base values sensible? I feel like it should be a straightforward process to find and replace the deprecated values once the new ones are in place, and this gives us a "compatibility layer", but open to being wrong there (either there's a better way to deprecate, or we should be aggressive and go breaking change off the bat?)

packages/design/src/colors.css
```diff
@@ -83,20 +83,20 @@
...
-  --color-lime: rgb(153, 194, 71);
-  --color-lime--light: rgb(184, 212, 126);
-  --color-lime--lighter: rgb(214, 230, 181);
-  --color-lime--lightest: rgb(240, 246, 227);
-  --color-lime--dark: rgb(99, 125, 46);
+  --color-lime: var(--color-base-lime--500);
+  --color-lime--light: var(--color-base-lime--400);
+  --color-lime--lighter: var(--color-base-lime--300);
+  --color-lime--lightest: var(--color-base-lime--200);
+  --color-lime--dark: var(--color-base-lime--700);
```
chris-at-jobber, line 99: @jdeichert this is what I meant by "deprecation layer" / "compatibility mode" - pointing the existing "{color}/dark/light/lighter/lightest" allows us to land this change without breaking any existing usages of the base tokens (including in `@components`) so we can do a migration pass in consuming apps.

.storybook/main.js
```diff
@@ -119,6 +119,9 @@ const config = {
                 require("postcss-preset-env")({
                   stage: 1,
                   preserve: true,
+                  features: {
```
chris-at-jobber, line 122: Storybook was checking for ["functional notation" of rgb and hsl values](https://cssdb.org/#color-functional-notation), which is interesting because the hsl values in this PR are not written in that format.

chris-at-jobber: What it was checking for:

`hsl(100deg 50% 50%)`
`hsla(100deg 50% 50% / 0.75)`

How we're writing the values:

`hsl(100deg, 50%, 50%)`
`hsla(100deg, 50%, 50%, 0.75)`

packages/design/src/colors.css
```diff
@@ -2,56 +2,56 @@
...
+  --color-inactive: var(--color-base-blue--700);
+  --color-inactive--surface: var(--color-base-blue--200);
+  --color-inactive--onSurface: var(--color-base-blue--900);
 
   /* Typography + Iconography */
-  --color-heading: var(--color-blue);
-  --color-text: var(--color-greyBlue--dark);
-  --color-text--secondary: var(--color-blue--light);
+  --color-heading: var(--color-base-blue--900);
+  --color-text: var(--color-base-taupe--800);
```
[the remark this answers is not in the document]
chris-at-jobber: tweaked saturation on `blue--800` and it feels much better

packages/design/src/colors.css
```diff
@@ -2,120 +2,120 @@
...
+  --color-inactive: var(--color-base-blue--700);
+  --color-inactive--surface: var(--color-base-blue--200);
+  --color-inactive--onSurface: var(--color-base-blue--900);
 
   /* Typography + Iconography */
-  --color-heading: var(--color-blue);
-  --color-text: var(--color-greyBlue--dark);
-  --color-text--secondary: var(--color-blue--light);
+  --color-heading: var(--color-base-blue--900);
+  --color-text: var(--color-base-blue--800);
```
chris-at-jobber, line 41: image

packages/components/src/FormField/FormField.css
```diff
@@ -8,9 +8,9 @@
   --public-field--bottom-left-radius: var(--radius-base);
   --public-field--bottom-right-radius: var(--radius-base);
 
-  --field--placeholder-color: var(--color-greyBlue--light);
+  --field--placeholder-color: var(--color-base-blue--600);
```
chris-at-jobber, line 11: Considering moving these 3 component tokens up to the main semantic layer to make them more apparent for theming, but also very open to not doing that here

packages/design/src/colors.css
chris-at-jobber: image

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -150,7 +150,7 @@ export function Checkbox({
                   onFocus={onFocus}
                 />
                 <span className={styles.checkBox}>
-                  <Icon name={iconName} size="small" color="white" />
+                  <Icon name={iconName} color="surface" />
```
chris-at-jobber, line 153: so that checkbox doesn't show a white checkmark in dark mode

(this might also merit just hiding the checkmark altogether when unselected, but this change has less... surface... area)

packages/components/src/Glimmer/Glimmer.css
```diff
@@ -5,6 +5,11 @@
   --duration-base: calc(var(--timing-slowest) * 4); /* 2s */
   --duration-fast: calc(var(--duration-base) / 2); /* 1s */
 
+  @media screen and (prefers-color-scheme: dark) {
+    --glimmer-background: rgba(var(--color-white--rgb), 10%);
```
chris-at-jobber, line 9: Looking at this again I think I want to just do this with semantic colors instead of media queries

chris-at-jobber: Not a great way to do this, so I'm going to remove this change for now

scotttjob (approved):
The base concept makes sense to me, I've seen that term on and off for years in the agency world so it never struck me as odd. If you do decide to go with primitive though to stick closer to standards, I've been looking for a word in my component work concepts and primitive is a pretty good one.

Zero issues with moving to hsl. It's so much easier to reason about as a person without having to visualize it or really think about it in your head. It's also been in the spec for close to 10 years now.

The deprecation plan makes sense. Support the old, roll this out, target the old and replace, remove the compatibility layer.

And your comment about removing unused tokens? Yes please! Whenever we get the chance let's pull stuff out if it's unused (and old).

PR #1875, opened 2024-04-24
fix(components): Allow InputText Default Value to be Clearable

ZakaryH:
## Motivations

currently, if you provide a `defaultValue` to an `InputText` and make that clearable, you cannot actually clear that value. it will always remain.

this was exposed via DataList's search, if you give that an `initialValue` it is passed into an `InputText` as a `defaultValue` which is where the real problem is

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

if you provide a `defaultValue` to an `InputText` along with `clearable` and it will work as intended

(sorta) fixed a type issue where we were only accepting "always" and "never" as values for `clearable` when there is code to deal with `while-editing` but there was no way to hit that code, at least if you listened to the types

### Security

- 

## Testing

**DataList**
give the `DataList.Search` an `initialValue`, see that you can clear the value when pressing the X

**InputText**
give an `InputText` a `defaultValue` and make it `clearable`, see that you can clear the value when pressing the X

**While-editing clear type**

give an `InputText` a `while-editing` `clearable` prop, see that once you type in the input and maintain focus, it is clearable. if you keep the value and lose focus on the input, the X to clear is no longer present*

*other Input types are not handling focus properly and so this ain't working for them

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -173,7 +173,7 @@ export function FormField(props: FormFieldProps) {
 
         function handleClear() {
           handleBlur();
-          setValue(controlledName, undefined, { shouldValidate: true });
+          setValue(controlledName, "", { shouldValidate: true });
           onChange && onChange("");
           inputRef?.current?.focus();
         }
--- a/packages/components/src/FormField/FormFieldTypes.ts
+++ b/packages/components/src/FormField/FormFieldTypes.ts
@@ -1,6 +1,7 @@
 import { ChangeEvent, ReactNode, RefObject } from "react";
 import { RegisterOptions } from "react-hook-form";
 import { XOR } from "ts-xor";
+import { Clearable } from "@jobber/hooks/src/useShowClear";
 import { IconNames } from "../Icon";
 
 export type FormFieldTypes =
@@ -121,7 +122,7 @@ export interface CommonFormFieldProps {
    * clearable. if the input value isn't editable (i.e. `InputTime`) you can
    * set it to `always`.
    */
-  readonly clearable?: "never" | "always";
+  readonly clearable?: Clearable;
 }
 
 export interface FormFieldProps extends CommonFormFieldProps {
--- a/packages/components/src/FormField/FormFieldWrapper.tsx
+++ b/packages/components/src/FormField/FormFieldWrapper.tsx
@@ -6,7 +6,7 @@ import React, {
   useState,
 } from "react";
 import classnames from "classnames";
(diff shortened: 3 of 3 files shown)


packages/components/src/FormField/FormField.tsx
```diff
@@ -173,7 +173,7 @@ export function FormField(props: FormFieldProps) {
 
         function handleClear() {
           handleBlur();
-          setValue(controlledName, undefined, { shouldValidate: true });
```
ZakaryH, line 176: image
https://react-hook-form.com/docs/useform/setvalue


while that doesn't directly recommend you to use an empty string, there's some related functionality around the default value that says to use an empty string or null

image


null does not work, so went with an empty string

ZakaryH, line 176: image
https://react-hook-form.com/docs/useform/setvalue


while that doesn't directly recommend you to use an empty string, there's some related functionality around the default value that says to use an empty string or null

image


null does not work, so went with an empty string

packages/components/src/FormField/FormFieldTypes.ts
```diff
@@ -121,7 +122,7 @@ export interface CommonFormFieldProps {
    * clearable. if the input value isn't editable (i.e. `InputTime`) you can
    * set it to `always`.
    */
-  readonly clearable?: "never" | "always";
+  readonly clearable?: Clearable;
```
ZakaryH, line 125: further down, we accept 3 values
https://github.com/GetJobber/atlantis/blob/master/packages/hooks/src/useShowClear/useShowClear.ts#L1

and we have code to handle that case and do something different

https://github.com/GetJobber/atlantis/blob/master/packages/hooks/src/useShowClear/useShowClear.ts#L28-L29

we just weren't leveraging that same type and instead had a more restrictive type here

ZakaryH: oh no. something is not working with `focused` on other input types other than `InputText` 😬 quite possibly something I broke with my recent PR

that concept is only used in conjunction with `clearable` `while-editing`

I can revert those type changes so it's back to only `always | never` OR I can keep it like this which works on `InputText` but not others. technically we'd be slightly better off as far as functionality since now 1 type can use `while-editing` rather than 0. it's just that it's more obvious that the prop isn't working on other input types

MichaelParadis (approved):
LGTM!

MichaelParadis (approved):
LGTM! I wanted to test one thing I thought of after my first approval

PR #1876, opened 2024-04-26
fix(components): Input Text Safari Resize Handle

ZakaryH:
## Motivations

On Safari, you can't resize a text area 😱

it only happens on multiline `InputText`s that **don't** have a toolbar prop.

this is because the toolbar will use margins to make sure its content doesn't cover the handle. however the non toolbar version will fill the entire area, and I guess Safari treats this different from other browsers 🤷 
image

image

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

made it so that the resize handle that is only present on multiline InputTexts is not obstructed and can be used on all browsers, especially Safari which was not having a great time

we put some padding around the entire textarea anyway, so I've just implemented that same amount of spacing but with margin on the container that would otherwise obstruct the resize handle. making sure to only do this for the multiline/textarea case because 

A) I don't want to interfere with the spacing of all the other input types. there are so many variations that have right aligned elements
B) many of the right aligned props are broken or not supported with multiline anyway :^)

`clearable` isn't supported, `loading` is lost, `suffix` icon & label are lost etc.

### Security

- 

## Testing

**on Chrome, Safari and Firefox**
head to the storybook for InputText and make it `multiline`. you should be able to drag the resize handle to extend/shorten it.

also try some other props such as `size` and `rows`. with a `small` `size` it should still have enough margin to allow you to grab and use the resize handle

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -70,9 +70,12 @@
 
 .horizontalWrapper {
   display: flex;
-  width: 100%;
   height: var(--field--height);
   flex: 1;
+}
+
+.textarea .horizontalWrapper {
+  margin-right: var(--field--padding-right);
 }
 
 .timeInputLabel:not(:focus-within) input {
@@ -216,6 +219,7 @@
   flex: 1;
   resize: none;
   scroll-padding-bottom: var(--space-base);
+  padding-right: 0;
 }
 
 .select .input {
--- a/packages/components/src/FormField/FormField.css.d.ts
+++ b/packages/components/src/FormField/FormField.css.d.ts
@@ -2,11 +2,11 @@ declare const styles: {
   readonly "container": string;
   readonly "wrapper": string;
   readonly "horizontalWrapper": string;
+  readonly "textarea": string;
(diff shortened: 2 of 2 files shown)


ZakaryH:
this does double up the spacing when you have a suffix icon or label
image

but yeah, it already looks pretty bad and need to be revisited. also I checked in product and we have exactly 0 instances of someone using `multiline` + `suffix`.

ZakaryH:
@chris-at-jobber this does introduce some double spacing in one scenario that no one appears to be using in product

tagging you on it just in case

chris-at-jobber:
@ZakaryH good catch, but yes as you noted nobody uses these (no real use case, which is why we haven't optimized for it) so, as a wise man once said, "I'm not worried about it"

MichaelParadis (approved):
I was able to resize multiline inputs of all sizes with all browsers with and without the toolbar


I saw this existing issue with small single line inputs that are clearable but this isn't the result of this PR. I should note we have no instances of a clearable input that is small

image

PR #1877, opened 2024-04-30
fix(components): Reduce Scope of Safari Scrollbar Issue

ZakaryH:
## Motivations

the quick fix for the not being able to grab the resize handle has broader effects than I'd like. namely when you have content that is greater than the area, your scrollbar is not right aligned. the margin pushes it off a bit
**Before (Chrome)**
image

**Before (Firefox)**
image

I really tried to do this in a more elegant way, and got like 90% of the way there but more issues keep on popping up and I need to move on with my life, so here we are.

**After (Chrome, then Firefox)**
image

image

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

only going to apply the margin replacing padding approach on Safari where it is problematic. the others, firefox & chrome, have no need for the margins and it can make them look bad with certain scrollbar settings.

### Security

- 

## Testing

**on Firefox, Chrome and Safari**

with a multiline InputText, try using the resize handle

try typing enough content that it wants to scroll, should still be able to use the resize handle. on Safari it will look a bit off :/

toolbar should all be fine. that was never the problematic case even with Safari since it has something in place to avoid overlapping with the resize handle

also, try this out with different macOS scrollbar settings. Taylor caught the last issue because we had different optiosn selected.

Appearance

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -106,7 +106,7 @@
 
 .wrapper.textarea.safari {
   overflow: unset;
-  overflow-x: scroll;
+  overflow-y: scroll;
 }
 
 .invalid,
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -74,10 +74,6 @@
   flex: 1;
 }
 
-.textarea .horizontalWrapper {
-  margin-right: var(--field--padding-right);
-}
-
 .timeInputLabel:not(:focus-within) input {
   color: var(--field--background-color);
   -webkit-text-fill-color: var(--field--background-color);
@@ -106,6 +102,11 @@
 .wrapper.textarea {
   resize: vertical;
   overflow: auto;
+}
+
+.wrapper.textarea.safari {
+  overflow: unset;
+  overflow-x: scroll;
 }
 
 .invalid,
@@ -219,7 +220,6 @@
   flex: 1;
   resize: none;
   scroll-padding-bottom: var(--space-base);
-  padding-right: 0;
 }
 
 .select .input {
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -74,6 +74,10 @@
   flex: 1;
 }
 
+.textarea.safari .horizontalWrapper {
+  margin-right: var(--field--padding-right);
+}
+
 .timeInputLabel:not(:focus-within) input {
   color: var(--field--background-color);
   -webkit-text-fill-color: var(--field--background-color);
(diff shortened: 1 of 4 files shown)


packages/components/src/FormField/FormField.css
```diff
@@ -108,6 +104,11 @@
   overflow: auto;
 }
 
+.wrapper.textarea.safari {
+  overflow: unset;
+  overflow-y: scroll;
```
ZakaryH, line 109: using the Y axis because it feels less intrusive or disruptive

if I use the X, then it reserves the space for the horizontal scroll, and make the input feel much more imbalanced

image

while it's still a similar problem on the Y axis, it's much harder to notice. it's only when things go all the way to the edge that you might see that it stops earlier due to the extra spacing, or the situation where you have the inner scrollbar like mentioned somewhere on this PR

image

packages/components/src/FormField/FormFieldWrapper.tsx
```diff
@@ -31,6 +32,7 @@ interface LabelPadding {
   paddingRight: number | string | undefined;
 }
 
+// eslint-disable-next-line max-statements
```
ZakaryH, line 35: I hate to do this and I know it encourages more bad behavior but I really do not wish to refactor a bunch of this for 1 added line. feel free to disagree with me though!

packages/components/src/FormField/hooks/useIsSafari.ts
```diff
@@ -0,0 +1,5 @@
+export function useIsSafari(): boolean {
+  return globalThis?.navigator
+    ? /^((?!chrome|android).)*safari/i.test(navigator.userAgent)
+    : false;
+}
```
ZakaryH, line 5: guarding this with a check to make it safe for SSR

taylorvnoj:
Okay, the good news is that Chrome and Firefox look/work great.

Bad news - Safari is giving me trouble. I cannot resize at all (locally and via the Cloudflare build). Were you able to get it working on your end?  Here you'll see that on localhost I cannot resize:

Kapture 2024-05-01 at 17 26 24

I wonder if [this](https://github.com/GetJobber/atlantis/pull/1877/files#diff-a241f863addccad1f3bf4f9b8f770b0b391031775b7f1dd532972daf63f76e4fR221) `resise: none` is overwriting the safari targeted css you just added?

I tried to add `-webkit-resize: vertical;` to further target Safari and that at least allowed me to see the resize indicator but I still couldn't resize the input 🤔  I'm hoping I'm somehow missing something

ZakaryH:
good catch!

it's this setting. I had mine set to "always" and that's why it worked. if I toggle it to something else, it stops working because in the initial state it has no reason to scroll, and so it doesn't leave any space for the scrollbar which is effectively the resize handle too

image

taylorvnoj (approved):
Perfect - I can now resize on Safari also. Great work, Zak 🚀

PR #1881, opened 2024-05-02
fix(components): Portal Menu Content

ZakaryH:
## Motivations

When you open a `Menu` the little dialog portion that pop open is a descendant of the button or activator. This means that if your activator or `Menu` component is inside a container that has an `overflow: hidden` then the menu will get cut off. Classic.

The fix is pretty standard. Use a portal to place it higher in the DOM, namely the `document.body`. That's the main purpose of this PR.

## Changes
**Position + Portal**
Prior to this, we were doing the positioning ourselves with CSS. Checking to see if it should go above, below, left or right and all that good stuff. Additionally we did something totally different for small screens. Instead providing a bottom sheet-like experience.

Now, because we want to portal the menu to the body, we cannot use those same styles that rely on relative positions. Why not use popper like we do elsewhere. It's quite simple to use it in [conjunction with portals](https://popper.js.org/react-popper/v2/react-portals/). 

Popper is handling the non small screen positioning, and we are handling the small screen ourselves with the same CSS but inverted where we'd apply styles to anything bigger than the small and the small was the default.

------
**`useFocusTrap`**

previously when tabbing on the open menu, it would not close and you'd make your way to the next/previous elements. Not the best, not the worst. However now that it is on the body, we tab to things that are nowhere near where we were. Not good. So now we're going to trap focus which is a best practice for pop up dialogs/modals and this is arguably one of those.

-------

**InputText + Toolbar focus**

A problem that come up when using `Menu` in this context is that the toolbar (when `whileEditing` `toolbarVisibility` is chosen) will not show when the input isn't focused. The way we were determining if the input was focused when it now had other elements inside of it was to see if they were descendants. This doesn't work with the portalled content.

I came up with a way to deal with that. It works and is pretty lightweight but very open to discussion on how to better handle this. It's a pretty interesting problem.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Security

- 

## Testing

Use a Menu normally. It should behave and look just like it did before. Moving above and below when the available space dictates it is necessary. I did see some code about moving right and left, though I'm not totally clear on how you'd cause that to trigger.

Use a Menu within a container that has `overflow: hidden`. Should be able to see the open menu.

Use a Menu within an InputText as the `toolbar`. See that focusing the input shows the toolbar, then click on the exposed button to launch the menu and see that the open menu is both not cut off visually but also the entire toolbar and menu remain open and usable!

**Rewrite Button**

Please note that this will NOT work in product with the Rewrite button. There is some additional code in that component managing focus states that needs to be removed. If you want to try a prerelease build with that, comment out or delete all the focus managing `useEffect` code  in the React magic input component.

---

Random photo of Atlantis

### DIFF
--- a/docs/components/InputText/Web.stories.tsx
+++ b/docs/components/InputText/Web.stories.tsx
@@ -13,8 +13,6 @@ export default {
 } as ComponentMeta<typeof InputText>;
 
 const BasicTemplate: ComponentStory<typeof InputText> = args => {
-  const [show, setShow] = React.useState(false);
-
   return (
     <InputText
       {...args}
--- a/docs/components/InputText/Web.stories.tsx
+++ b/docs/components/InputText/Web.stories.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
 import { InputText } from "@jobber/components/InputText";
 import { Button } from "@jobber/components/Button";
-import { Menu } from "@jobber/components/Menu";
 
 export default {
   title: "Components/Forms and Inputs/InputText/Web",
@@ -13,47 +12,7 @@ export default {
 } as ComponentMeta<typeof InputText>;
 
 const BasicTemplate: ComponentStory<typeof InputText> = args => {
-  return (
-    <InputText
-      {...args}
-      toolbar={
-        <Menu
-          items={[
-            {
-              actions: [
-                {
-                  label: "Edit",
-                  icon: "edit",
-                  onClick: () => {
-                    alert("✏️");
-                  },
-                },
-              ],
-            },
-            {
-              header: "Send as...",
-              actions: [
-                {
-                  label: "Text Message",
-                  icon: "sms",
-                  onClick: () => {
-                    alert("📱");
-                  },
-                },
-                {
-                  label: "Email",
-                  icon: "email",
-                  onClick: () => {
-                    alert("📨");
-                  },
-                },
-              ],
-            },
-          ]}
-        />
-      }
-    />
-  );
+  return <InputText {...args} />;
 };
 
 export const Basic = BasicTemplate.bind({});
--- a/docs/components/InputText/Web.stories.tsx
+++ b/docs/components/InputText/Web.stories.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
 import { InputText } from "@jobber/components/InputText";
 import { Button } from "@jobber/components/Button";
+import { Menu } from "@jobber/components/Menu";
 
 export default {
   title: "Components/Forms and Inputs/InputText/Web",
@@ -12,7 +13,49 @@ export default {
 } as ComponentMeta<typeof InputText>;
 
 const BasicTemplate: ComponentStory<typeof InputText> = args => {
-  return <InputText {...args} />;
+  const [show, setShow] = React.useState(false);
+
+  return (
+    <InputText
+      {...args}
+      toolbar={
+        <Menu
+          items={[
+            {
+              actions: [
+                {
+                  label: "Edit",
+                  icon: "edit",
+                  onClick: () => {
+                    alert("✏️");
+                  },
+                },
+              ],
+            },
+            {
+              header: "Send as...",
+              actions: [
+                {
+                  label: "Text Message",
+                  icon: "sms",
+                  onClick: () => {
+                    alert("📱");
+                  },
+                },
+                {
+                  label: "Email",
+                  icon: "email",
+                  onClick: () => {
+                    alert("📨");
+                  },
+                },
+              ],
+            },
+          ]}
+        />
+      }
+    />
+  );
 };
 
 export const Basic = BasicTemplate.bind({});
--- a/docs/components/Menu/Web.stories.tsx
+++ b/docs/components/Menu/Web.stories.tsx
@@ -51,10 +51,7 @@ Horizontal.args = {
 };
 
 const CustomActivatorTemplate: ComponentStory<typeof Menu> = args => (
-  <>
-    <div style={{ height: "600px" }}></div>
-    <Menu {...args} activator={<Button label="My Fancy Menu" />} />
-  </>
+  <Menu {...args} activator={<Button label="My Fancy Menu" />} />
 );
 
 export const CustomActivator = CustomActivatorTemplate.bind({});
--- a/docs/components/Menu/Web.stories.tsx
+++ b/docs/components/Menu/Web.stories.tsx
@@ -51,9 +51,7 @@ Horizontal.args = {
 };
 
 const CustomActivatorTemplate: ComponentStory<typeof Menu> = args => (
-  <div style={{ width: "200px", height: "100px", overflow: "hidden" }}>
-    <Menu {...args} activator={<Button label="My Fancy Menu" />} />
-  </div>
+  <Menu {...args} activator={<Button label="My Fancy Menu" />} />
 );
 
 export const CustomActivator = CustomActivatorTemplate.bind({});
--- a/docs/components/Menu/Web.stories.tsx
+++ b/docs/components/Menu/Web.stories.tsx
@@ -51,7 +51,9 @@ Horizontal.args = {
 };
 
 const CustomActivatorTemplate: ComponentStory<typeof Menu> = args => (
-  <Menu {...args} activator={<Button label="My Fancy Menu" />} />
+  <div style={{ width: "200px", height: "100px", overflow: "hidden" }}>
+    <Menu {...args} activator={<Button label="My Fancy Menu" />} />
+  </div>
 );
 
 export const CustomActivator = CustomActivatorTemplate.bind({});
--- a/docs/components/Menu/Web.stories.tsx
+++ b/docs/components/Menu/Web.stories.tsx
@@ -51,7 +51,10 @@ Horizontal.args = {
 };
 
 const CustomActivatorTemplate: ComponentStory<typeof Menu> = args => (
-  <Menu {...args} activator={<Button label="My Fancy Menu" />} />
+  <>
+    <div style={{ height: "600px" }}></div>
+    <Menu {...args} activator={<Button label="My Fancy Menu" />} />
+  </>
 );
 
 export const CustomActivator = CustomActivatorTemplate.bind({});
--- a/packages/components/src/FormField/hooks/useFormFieldFocus.ts
+++ b/packages/components/src/FormField/hooks/useFormFieldFocus.ts
@@ -9,10 +9,10 @@ interface UseFormFieldFocusProps {
   wrapperRef?: React.RefObject<HTMLDivElement>;
 }
 
-const FOCUS_DATA_ATTRIBUTE = "data-atl-maintain-portal-focus";
+const PORTAL_FOCUS_ATTRIBUTE_NAME = "data-atl-maintain-portal-focus";
 
-export const focusAttribute = {
-  [FOCUS_DATA_ATTRIBUTE]: "true",
+export const formFieldFocusAttribute = {
+  [PORTAL_FOCUS_ATTRIBUTE_NAME]: "true",
 };
 
 export function useFormFieldFocus(
@@ -36,7 +36,7 @@ export function useFormFieldFocus(
         document.activeElement,
       );
       const focusException = activeElementRef.current?.closest(
-        `[${FOCUS_DATA_ATTRIBUTE}='true']`,
+        `[${PORTAL_FOCUS_ATTRIBUTE_NAME}='true']`,
       );
 
       if (!focusedElementWithinWrapper && !focusException) {
--- a/packages/components/src/FormField/hooks/useFormFieldFocus.ts
+++ b/packages/components/src/FormField/hooks/useFormFieldFocus.ts
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 
 interface UseFormFieldFocus {
   focused: boolean;
@@ -12,23 +12,33 @@ export function useFormFieldFocus(
   props: UseFormFieldFocusProps,
 ): UseFormFieldFocus {
   const [focused, setFocused] = useState(false);
+  const activeElement = useActiveElement();
+  const activeElementRef = useRef(activeElement);
 
   useEffect(() => {
-    function handleFocusIn() {
-      setFocused(true);
-    }
+    activeElementRef.current = activeElement;
+  }, [activeElement]);
 
-    function handleFocusOut() {
-      setTimeout(() => {
-        const focusedElementWithinWrapper = props.wrapperRef?.current?.contains(
-          document.activeElement,
-        );
+  function handleFocusIn() {
+    setFocused(true);
+  }
 
-        if (!focusedElementWithinWrapper) {
-          setFocused(false);
-        }
-      }, 1);
-    }
+  function handleFocusOut() {
+    setTimeout(() => {
+      const focusedElementWithinWrapper = props.wrapperRef?.current?.contains(
+        document.activeElement,
+      );
+      const focusException = activeElementRef.current?.closest(
+        "[data-atl-maintain-portal-focus='true']",
+      );
+
+      if (!focusedElementWithinWrapper && !focusException) {
+        setFocused(false);
+      }
+    }, 1);
+  }
+
+  useEffect(() => {
     props.wrapperRef?.current?.addEventListener("focusin", handleFocusIn);
     props.wrapperRef?.current?.addEventListener("focusout", handleFocusOut);
 
@@ -45,3 +55,21 @@ export function useFormFieldFocus(
     focused,
   };
 }
+
+const useActiveElement = () => {
+  const [active, setActive] = useState(document.activeElement);
+
+  const handleFocusIn = () => {
+    setActive(document.activeElement);
+  };
+
+  useEffect(() => {
+    document.addEventListener("focusin", handleFocusIn);
+
+    return () => {
+      document.removeEventListener("focusin", handleFocusIn);
+    };
+  }, []);
+
+  return active;
+};
--- a/packages/components/src/FormField/hooks/useFormFieldFocus.ts
+++ b/packages/components/src/FormField/hooks/useFormFieldFocus.ts
@@ -8,6 +8,12 @@ interface UseFormFieldFocusProps {
   wrapperRef?: React.RefObject<HTMLDivElement>;
 }
 
+const FOCUS_DATA_ATTRIBUTE = "data-atl-maintain-portal-focus";
+
+export const focusAttribute = {
+  [FOCUS_DATA_ATTRIBUTE]: "true",
+};
+
 export function useFormFieldFocus(
   props: UseFormFieldFocusProps,
 ): UseFormFieldFocus {
@@ -29,7 +35,7 @@ export function useFormFieldFocus(
         document.activeElement,
       );
       const focusException = activeElementRef.current?.closest(
-        "[data-atl-maintain-portal-focus='true']",
+        `[${FOCUS_DATA_ATTRIBUTE}='true']`,
       );
 
       if (!focusedElementWithinWrapper && !focusException) {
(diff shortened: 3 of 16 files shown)


packages/components/src/Menu/Menu.tsx
```diff
@@ -12,12 +11,16 @@ import classnames from "classnames";
...
 import { IconNames } from "@jobber/design";
-import { useSafeLayoutEffect } from "@jobber/hooks/useSafeLayoutEffect";
+import { usePopper } from "react-popper";
+import { useIsMounted } from "@jobber/hooks/useIsMounted";
+import ReactDOM from "react-dom";
 import styles from "./Menu.css";
 import { Button } from "../Button";
 import { Typography } from "../Typography";
 import { Icon } from "../Icon";
+import { focusAttribute } from "../FormField/hooks/useFormFieldFocus";
```
ZakaryH, line 23: this import location feels awkward.... pulling something from form field into `Menu` 

the alternative is moving it up to somewhere more generic and less tied to a specific component like `hooks`. I don't know if I like that either though because it is directly, and currently only used in the form field because that's currently the only place that has this problem

packages/components/src/Menu/Menu.tsx
```diff
@@ -106,80 +113,81 @@ export function Menu({ activator, items }: MenuProps) {
...
+                transition={{
+                  type: "tween",
+                  duration: 0.15,
+                }}
+              />
+              <div
+                ref={setPopperElement}
+                className={styles.popperContainer}
+                {...positionAttributes}
+                {...focusAttribute}
```
ZakaryH, line 153: this turns into `data-atl-maintain-portal-focus: "true"` on this wrapping div

I'm not super in love with this approach but it's the lightestweight solution I can think of that lends itself to expanding to other components that run into this same problem

packages/components/src/FormField/hooks/useFormFieldFocus.ts
```diff
@@ -8,27 +8,43 @@ interface UseFormFieldFocusProps {
...
+  }
+
+  function handleFocusOut() {
+    setTimeout(() => {
+      const focusedElementWithinWrapper = props.wrapperRef?.current?.contains(
+        document.activeElement,
+      );
+      const focusException = activeElementRef.current?.closest(
+        `[${FOCUS_DATA_ATTRIBUTE}='true']`,
+      );
```
ZakaryH, line 39: this works but it's a little weird

the problem I'm solving is that the toolbar hides when the `toolbarVisibility` is `whileEditing`. the way it hides is it checks if the descendants of the toolbar are focused or not, and then sets some state depending on that

now that the Menu list is being portalled, it is no longer a descendant. it's just some element on the `body`, and so we need some way to disambiguate that element on the body from any other element on the body

then we can whitelist it and say please don't lose focus if the current active element has a parent somewhere up the tree that has `"data-atl-maintain-portal-focus"`

we will have to be careful about using the attribute. it must only be applied to somethign that is portalled to the body and can never be a parent of a very broad tree - otherwise if we give focus to some other element that is a descendant that should cause the toolbar to lose focus, it would  not

docs/components/InputText/Web.stories.tsx
```diff
@@ -12,7 +13,49 @@ export default {
 } as ComponentMeta<typeof InputText>;
 
 const BasicTemplate: ComponentStory<typeof InputText> = args => {
-  return <InputText {...args} />;
+  const [show, setShow] = React.useState(false);
```
ZakaryH, line 16: some example code that isn't needed. either remove entirely or make an example with this.

docs/components/Menu/Web.stories.tsx
```diff
@@ -51,7 +51,9 @@ Horizontal.args = {
 };
 
 const CustomActivatorTemplate: ComponentStory<typeof Menu> = args => (
-  <Menu {...args} activator={<Button label="My Fancy Menu" />} />
+  <div style={{ width: "200px", height: "100px", overflow: "hidden" }}>
```
ZakaryH, line 54: more code that can be removed. only exists to illustrate the overflow case ain't broken no more.

packages/components/src/FormField/hooks/useFormFieldFocus.ts
```diff
@@ -45,3 +61,21 @@ export function useFormFieldFocus(
     focused,
   };
 }
+
+const useActiveElement = () => {
```
ZakaryH, line 65: probably should be moved into its own hook

packages/hooks/src/useWindowDimensions/useWindowDimensions.ts
```diff
@@ -0,0 +1,28 @@
...
+function getWindowDimensions() {
+  const { innerWidth: width, innerHeight: height } = window;
+
+  return {
+    width,
+    height,
+  };
+}
+
+export function useWindowDimensions() {
```
ZakaryH, line 12: I made this a "real" hook because it's generic enough. it's literally a base hook from React Native so I feel pretty safe in that line of thought

however maybe we aren't quite ready for it to be at this level, since it's literally being used in exactly 1 place right now

I'm open to demoting it down to be a hook within the `Menu/hooks` directory

packages/hooks/src/useWindowDimensions/useWindowDimensions.stories.mdx
```diff
@@ -0,0 +1,26 @@
...
+```
+
+<Canvas withToolbar>
+  <Story name="useWindowDimensions">
+    {() => {
+      const { width } = hooks.useWindowDimensions();
+      useEffect(() => {
+        console.log({ width });
+      }, [width]);
+      return <h1>useWindowDimensions</h1>;
```
ZakaryH, line 23: make this example better. it requires the console to be open to see anything.

packages/components/src/Menu/Menu.tsx
```diff
@@ -64,35 +63,46 @@ export interface SectionProps {
...
+      ],
+    },
+  );
+  const positionAttributes =
+    width > 489
+      ? {
+          ...attributes.popper,
+          style: popperStyles.popper,
+        }
+      : {};
```
ZakaryH, line 104: this is a little awkward

we're now using Popper to position the menu when it's a medium+ screen

but when it's a small screen, we don't want to use any popper logic at all. we know it should simply be position fixed at the bottom of the screen, spanning the full width. to that end, we don't want to apply any popper styles or attributes in the small screen case

I don't love that `489` is a magic number, and there isn't a super great way to share it with the media query variable value

it's also kind of meh that the position is handled by JS + CSS for biggers screens and then just CSS for the small screens

packages/components/src/Menu/Menu.tsx
```diff
@@ -12,12 +11,17 @@ import classnames from "classnames";
...
+import { useIsMounted } from "@jobber/hooks/useIsMounted";
+import ReactDOM from "react-dom";
+import { useFocusTrap } from "@jobber/hooks/useFocusTrap";
 import styles from "./Menu.css";
 import { Button } from "../Button";
 import { Typography } from "../Typography";
 import { Icon } from "../Icon";
+import { focusAttribute } from "../FormField/hooks/useFormFieldFocus";
 
 const variation = {
```
ZakaryH, line 26: I can't tell if this is still needed or not

I've tested out the show/hide animation on different screen sizes and it looks to be behaving identically to current master but there's definitely some things in here that no longer exist


```
  startOrStop: (position: string) => {
    let y = 10;

    if (position === "below") y *= -1;
    if (window.innerWidth < 640) y = 150;
```

that's looking for a `position === "below"` and I removed the `custom={position}` which I figure must be how it gets the `position.vertical` which would be `below` or `above` so I thought something wold break but it hasn't

we do get back a below/above from popper's `state` if this is still relevant. gonna look into it a bit more.

ZakaryH: ah this doesn't currently work on master

image


it used to only be a string, and this comparison might pass or fail - but then it was changed to what it is now and it's an object, so it'll never be the string "below"

https://github.com/GetJobber/atlantis/commit/4ba4981c90c998ff20fa63971b9721fdf54c54ab#diff-9912300449143379721456414d95c1dcf3414a8f0fb096d4d641b95ec0382f41L55-R63

ZakaryH: the `y` offset doesn't depend on the `position` at all. that's just looking at the window width, and from what I can tell it is telling the animation where to start and end

with a smaller screen we want it to move further for the animation. that will still work with this new setup since popper is only responsible for the resting position, not the animation!

packages/components/src/Menu/Menu.tsx
```diff
@@ -106,80 +121,76 @@ export function Menu({ activator, items }: MenuProps) {
...
+                  duration: 0.15,
+                }}
+              />
+              <div
+                ref={setPopperElement}
+                className={styles.popperContainer}
+                {...positionAttributes}
+                {...focusAttribute}
+              >
+                {items.length > 0 && (
```
ZakaryH, line 156: since we have nothing that says you must provide a non empty array, now that I'm using the focus trapping on the menu - it looks for the first focusable element to call `.focus()` on

however, if `items` is an empty array, it blows up 😱 

I don't think it's valid to have an empty menu. the UI before this was a silly looking sliver of a menu. we should probably not allow that, but for now this at least makes tests that were passing in empty `items` pass - and it guards against rendering that sliver

packages/components/src/Menu/Menu.tsx
```diff
@@ -63,38 +63,53 @@ export interface SectionProps {
...
+        {
+          name: "flip",
+          options: {
+            flipVariations: true,
+          },
+        },
+        {
+          name: "offset",
+          options: {
+            offset: [0, 2],
```
ZakaryH, line 99: 2 is a bit magic, but it matches the old layout (right) exactly
image

noticing the width of these is different though. have to look into that.

packages/components/src/Menu/Menu.css
```diff
@@ -97,7 +76,7 @@
 }
 
 .action:hover,
-.action:focus-visible {
+.action:focus {
```
ZakaryH, line 79: uncertain about this change and if it is correct. it was recently changed from focus -> focus-visible

the problem is that when the menu opens and the first element is programmatically focused by the `useFocusTrap` it was not visually focusing

with this style rule it is. not sure if maybe it should be both rules rather than just one or the other.

packages/components/src/Menu/Menu.tsx
```diff
@@ -251,29 +260,11 @@ export interface ActionProps {
...
-}: ActionProps) {
+function Action({ label, sectionLabel, icon, onClick }: ActionProps) {
   const actionButtonRef = useRef() as RefObject<HTMLButtonElement>;
 
-  useEffect(() => {
-    if (shouldFocus) {
-      // Focus on the next tick to allow useRefocusOnActivator to initialize
-      setTimeout(() => actionButtonRef.current?.focus(), 0);
-    }
-  }, [shouldFocus]);
```
ZakaryH, line 275: we don't need this anymore. at least I'm pretty confident. `useFocusTrap` will both focus the first focusable element, and keep focus trapped

`useFocusTrap` does have a short list of focusable elements, but `button` is one of them, and that's what we're rendering as an `Action`

```
const focusables = [
    "button",
    "[href]",
    "input",
    "select",
    "textarea",
    '[tabindex]:not([tabindex="-1"])',
  ];
```

ZakaryH: I'm somewhat open to the argument that I didn't need to add `useFocusTrap` for the scope of this ticket. 

I could be convinced, but I will say it feels very bad in practice without it. the focus runs off to the `document.body` because that's where our popper'd menu is now

packages/components/src/Menu/Menu.tsx
ZakaryH: also I'm always putting this in here. might it be worth having more control over?

does get kind of weird exposing a prop like `shouldMenusMenuKeepFocusWhenYouLaunchItFromInputTextsToolbar` that is deeply connected to another component's implementation. I suppose we're doing that either way, but doing it secretly is better?

packages/components/src/Menu/Menu.css
chris-at-jobber: `focus-visible` is intended to only show the focus "ring" when the user is operating the UI via keyboard/non-mouse input. I think in this particular case, the `background-color` should be present on both hover and focus, but we should move the outline styling to a separate `focus-visible` declaration.

packages/components/src/Menu/Menu.tsx
chris-at-jobber: @ZakaryH isn't this function checking which viewport edge the activator is closest to, so that it can position itself above, below, right, or left, accordingly?

packages/components/src/Menu/Menu.tsx
chris-at-jobber: 2 questions

- can we make the offset 6? it's still a bit magic, but our [Figma mocks](https://www.figma.com/file/rIIhulZvcp9M82lNOCGv16/Product%2FOnline?type=design&node-id=14110%3A0&mode=design&t=dN7Zpe2AQMGHYuxn-1) have it set as 6

ZakaryH: yup. most definitely. this is how that would look

image


haha that's actually what that `motion.div` code I was asking you about was trying to do. it's been broken for a few years now, but once upon a time that's what was happening. I have an idea of how to fix it. I'll give that a shot.

packages/components/src/Menu/Menu.tsx
ZakaryH: @chris-at-jobber negative. that logic was all part of the `setPosition` code. instead of us doing that ourselves, we're delegating that to Popper :)

though the result of that code would set `position` to be something like `{ vertical: "below", horizontal: "right" }` and that `position` would be passed to this `startOrStop(position: string)`

as noted, that is currently not working because the comparison would always fail. however, back when it was working - this function would potentially do 2 things

- if it's "below" flip the offset so that the starting/ending position of the menu's animation is the opposite, much like you were asking about on another comment
- if the screen is below 640px, then make the ending offset position 50px greater, not entirely sure why - if we wanted a higher speed (assuming it takes the same time) or just like how it looks moving a greater distance?

packages/components/src/Menu/Menu.tsx
ZakaryH: ahh I remember this. `useFocusTrap` and `useRefocusOnActivator` don't play well together

one tries to remember the last `activeElement` assuming it is the original button, and then the other tries to set the activeElement to be the first item in the newly opened dialog

then, when we close it the `useRefocusOnActivator` tries to set the focus back to the `activeElement` which is the now unmounted button inside the dialog 🥴 

fixing the interaction between these two might be fun

@chris-at-jobber if you had to choose one or the other which is preferable?

- refocusing properly when closed, however tabbing out of the Menu will NOT close it, and now that it is on the body you'll be tabbing to stuff miles away from where you were
- tabbing the opened Menu will stay locked in the dialog, but when you close it the focus gets lost and the focus is now on the body 😅 

I will call out that the former is the current experience

packages/components/src/Menu/Menu.css
ZakaryH: ok I've split that into its own declaration, but as far as I can tell it isn't doing anything. I've forced the `focus-visible` state and it's getting the styles, but even with an aggressive red colour - nada
Components___Forms_and_Inputs___InputText___Web_-_Multiline_⋅_Storybook

packages/components/src/Menu/Menu.tsx
chris-at-jobber: They work together in Modal don't they? If we have to choose, I would (for now) leave the experience as is with `refocusOnActivator` working and the tabs running out of the menu, and we can revisit this later. We should _actually_ be operating the menu options with arrow keys instead of tabs anyways, which is a separate long-standing issue.

ZakaryH: oh! that's a very good point. I forgot about Modal

ok scratch that. I'll see what it is doing to make it work.

packages/components/src/Menu/Menu.css
chris-at-jobber: In Chrome/ium, I think you need to provide [full outline styles](https://developer.mozilla.org/en-US/docs/Web/CSS/outline) for anything to happen there, while Safari provides the style and size out of the box.

By setting the outline-color to transparent and supplying our `--shadow-focus` on `focus-visible`, we should get the results we want across browsers.

packages/components/src/Menu/Menu.tsx
ZakaryH: should be working now where it'll feel like it is coming out of the button and back in based on the direction

chris-at-jobber: Nice! I'll give it a try when the build is up

packages/components/src/Menu/Menu.css
ZakaryH: ah gotcha. so if the styles were already trying to do that without the additional styles for Chrome - then it's currently not working and like you said, we can follow up on that later?

chris-at-jobber: Correct, probably easier to do in a more intentional/smaller followup

packages/components/src/Menu/Menu.tsx
ZakaryH: lol, I remember now

you have to call the refocus hook first, before the the focus trap hook or else it won't work

thinking about how we can better signal this temporal requirement beyond my comment. maybe we combine the two into a single hook that always calls them in the right order? trying to think if there's ever a time where you only want one of them and not both

anyways that's something for another day. this should work nicely now!

chris-at-jobber: Hey! It works nicely now, just like you said. 🎉

chris-at-jobber (commented):
Design approved! This will definitely need an eng. review as the hooks and Typescript work is beyond me, but this is a very nice improvement from a UX perspective.

packages/components/src/Menu/Menu.tsx
```diff
@@ -12,31 +10,34 @@ import classnames from "classnames";
...
+const MENU_OFFSET = 6;
 
 const variation = {
   overlayStartStop: { opacity: 0 },
-  startOrStop: (position: string) => {
+  startOrStop: (placement: string | undefined) => {
     let y = 10;
 
-    if (position === "below") y *= -1;
+    if (placement?.indexOf("bottom") !== -1) y *= -1;
```
scotttjob, line 33: This is very much a personal style, both achieve the same goal (so feel free to ignore), but `includes` instead of `indexOf` is a bit tidier because it returns a boolean vs having to check for -1 with indexOf.

packages/components/src/Menu/Menu.tsx
```diff
@@ -297,3 +293,13 @@ function Action({
     </button>
   );
 }
+
+function MenuPortal({ children }: { readonly children: React.ReactElement }) {
+  const mounted = useIsMounted();
+
+  if (!mounted?.current) {
```
scotttjob, line 300: I appreciate the SSR consideration, thanks!

scotttjob (approved):
I left a few comments. Zak and I talked a bit about this yesterday 1:1, and we agreed that it's a fine solution to a relatively convoluted issue (passing refs around might have been nearly impossible to mentally map in the future). 

I think this PR is a prime example of the value in the Form Field component work we want to do to remove our over-reliance on a centralized component, this work might have been much simpler if everything wasn't so tightly coupled together.

I did run this code locally, all the test cases marked in the PR worked. I was able to use a menu normally, within a height restricted wrapper, and as a toolbar (even with a height restriction on the toolbar component): 
image

I really appreciated the inline comments for some of the code.

I have a general piece of feedback that I won't hold up this work with, in that this PR is a bit large and might have benefited from being broken down into smaller pieces. The custom hook for instance, or even the new FormField hook could have been either done as atomic PRs (introduce the code with test suites, but don't hook it up yet), or in a feature branch pattern where the PRs just target a feature branch and you merge it all at once at the end.

All that said though, great work at tracking down a really tricky problem and solving it with some re-usable pieces. LGTM!

packages/components/src/Menu/Menu.tsx
ZakaryH: that's fair, I'd agree that's a nicer check. I'll update that.

packages/components/src/Menu/Menu.tsx
ZakaryH: no problemo!

there are a few things in this PR that are attempting to access either the `document` or `window` and I've put in some checks to hopefully address them - that said, I didn't explicitly test this on a Remix setup

PR #1883, opened 2024-05-08
fix(components): Remove hardcoded labels in DataListSort

taylorvnoj:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/docs/components/DataList/Web.stories.tsx
+++ b/docs/components/DataList/Web.stories.tsx
@@ -110,6 +110,7 @@ const Template: ComponentStory<typeof DataList> = args => {
         sortable: [
           {
             key: "label",
+            sortType: "dropdown",
             options: [
               {
                 id: "firstName",
@@ -131,6 +132,7 @@ const Template: ComponentStory<typeof DataList> = args => {
           },
           {
             key: "home",
+            sortType: "dropdown",
             options: [
               {
                 id: "homeWorld",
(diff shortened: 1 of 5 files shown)


docs/components/DataList/DataList.stories.mdx
```diff
@@ -304,3 +304,53 @@ want to inform the user about something related to the Datalist.
...
+
+The `sorting` prop is an object that configures the sorting functionality of the
+DataList.
+
+The `state` property is an object that holds the current sorting state of the
+DataList. The `onSort` property is a function that is called when the user
+clicks on a sortable column. Lastly, `sortable` is an array that holds
+information about which columns are sortable.
+
+Each column is identified by a `key` and will have a `sortType` property that
```
MichaelParadis, line 318: One comment that can be handled by docs is that `toggle` doesn't update the arrow / work for all options when the toggle sort has more than 2 options

MichaelParadis: ex. changing the story to have `label` as a `toggle` instead of dropdown

MichaelParadis (changes requested):
Minor comment on a documentation change. See: https://github.com/GetJobber/atlantis/pull/1883/files#r1596975546

docs/components/DataList/DataList.stories.mdx
taylorvnoj: Hmm... I wonder if that is something we should be putting in the docs? I feel like we could/should fix that in the future and I can write a ticket for it. The component is also still technically in Alpha so I hesitate to point out behaviour that may change later on.
How strongly do you feel about including this @MichaelParadis ?

MichaelParadis: I am good with just approving. I have a slight concern about people having problems due to misuse but I think it is extremely unlikely (toggles are to switch between 2 states so why use it for more than 2)

MichaelParadis (approved):
LGTM! based on the component still being in Alpha and it may change in the future. we can update the docs if issues arise and we don't change it

docs/components/DataList/DataList.stories.mdx
taylorvnoj: Exactly - ok I agree. Thanks Michael!

PR #1884, opened 2024-05-08
feat(components): add textCase=none to Typography

chris-at-jobber:
## Motivations

Because menu's subheading uses an `element={h6}` semantically, it was picking up styling from in-app CSS classes in Jobber that target `h6` making it uppercase. We still want it to be a semantic h6, so I added a `none` to `textCase` from `Typography` to give us the locked-down specificity we need for that to not happen.

### Added

- New textcase of `none` to Typography

## Testing

Set a style of `h6 { text-transform: uppercase; }` (and other adjustments like font-weight, if you like) in the inspector stylesheet and observe that the casing of the Menu's subheader is preserved

image

---

Random photo of Atlantis


ZakaryH (approved):
LGTM

PR #1886, opened 2024-05-10
feat(components): Add name and value props to Button

jjwlee94:
## Motivations

- Context: https://github.com/GetJobber/jobber-frontend/pull/312#discussion_r1594908049

### Added

- 

### Changed

- Added name and value props to Button.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -26,6 +26,7 @@ interface ButtonFoundationProps {
   readonly id?: string;
   readonly label?: string;
   readonly loading?: boolean;
+  readonly name?: string;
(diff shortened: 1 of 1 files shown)


scotttjob (approved):
LGTM!

I made some adjustments, namely moving name and value into SubmitActionProps which forces them to coincide with 'submit'

Which is how the <button> component works under the hood so that lines up. Any usages of <button> without type="submit" will not include name or value as FormData.

If you're downstream of this change, and are running into typing issues in your own code based on destructured props, ensure that any usage of 'name' or 'value' coincides with providing the 'submit' prop as 'true'.

PR #1888, opened 2024-05-10
feat(components): Add support for link click listener for Markdown component (JOB-96041)

sabidhasan:
## Motivations
The markdown component works well for handling markdown text, but our team has run into a use case where we need to know when a link was clicked. This PR extends the Markdown component to expose the underlying React Markdown's `component` prop. This way, we can use a custom component and attach whatever events we want.

### Added
Added an optional `components` prop to the Markdown component. This object is spread to the React Markdown component, ensuring that existing values can be overridden if desired.

Also added tests for this functionality.

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FeatureSwitch/__snapshots__/FeatureSwitch.test.tsx.snap
+++ b/packages/components/src/FeatureSwitch/__snapshots__/FeatureSwitch.test.tsx.snap
@@ -89,6 +89,7 @@ exports[`renders a full FeatureSwitch 1`] = `
             Send a 
             <a
               href="www.fakeurl.com"
+              rel="noreferrer"
               target="_blank"
             >
               notification
--- a/packages/components/src/Markdown/Markdown.test.tsx
+++ b/packages/components/src/Markdown/Markdown.test.tsx
@@ -1,4 +1,4 @@
-import { render } from "@testing-library/react";
+import { fireEvent, render } from "@testing-library/react";
 import React from "react";
 import { Markdown } from ".";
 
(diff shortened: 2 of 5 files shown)


leticiawanderley (commented):
I'm seeing a weird behaviour in the storybook: even when I use the same examples as the ones in your tests, the custom components aren't applied.
Screenshot 2024-05-13 at 9 39 32 AM
Screenshot 2024-05-13 at 9 40 05 AM

Is there anything else I should be doing to interact with these changes?

sabidhasan:
> I'm seeing a weird behaviour in the storybook: even when I use the same examples as the ones in your tests, the custom components aren't applied. Screenshot 2024-05-13 at 9 39 32 AM Screenshot 2024-05-13 at 9 40 05 AM
> 
> Is there anything else I should be doing to interact with these changes?

@leticiawanderley I think sometimes the storybook code editor has issues with live updating. Two ways to get around this:

1. You can try testing by installing this version of Atlantis in JO (or another demo repo) and testing that way to see your changes are being applied.
2. Another option is to update the default code in the `BasicTemplate` in `Web.stories.tsx`, because that gets updated and picked up by Webpack via hot reload. After applying this diff and running Webpack again, I can see the change in Storybook:

```diff
diff --git a/docs/components/Markdown/Web.stories.tsx b/docs/components/Markdown/Web.stories.tsx
index 7a50d869..978c6a23 100644
--- a/docs/components/Markdown/Web.stories.tsx
+++ b/docs/components/Markdown/Web.stories.tsx
@@ -16,9 +16,10 @@ const BasicTemplate: ComponentStory<typeof Markdown> = args => {
 
   return (
     <Markdown
-      content={content}
+      content={"# Foobar"}
       basicUsage={args.basicUsage}
       externalLink={args.externalLink}
+      components={{ h1: () => <h2>Test</h2> }}
     />
   );
 };
```

image

leticiawanderley (approved):
Thanks, Abid!
LGTM

leticiawanderley (approved):
LGTM!

JeffMarvin (approved):
Yeah, this should work. Thanks for setting this up.

PR #1890, opened 2024-05-15
feat(components): Box Component

scotttjob:
## Motivations

We've had a need for a generic layout component for a while. This component was rolled elsewhere, and we're moving it into Atlantis so everyone can benefit from its greatness.

The Box component is for all your wrapping component needs. You can put borders around your box, or padding, or margins, or border radiuses. You can set flex direction, gap size, all sorts of fun things.

There is basic documentation, but could be expanded with additional examples.

## Changes

* Added the Box component

### Added

- The Box Component

## Testing

The box component should render well in the provided storybook stories.

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Box/Box.tsx
+++ b/packages/components/src/Box/Box.tsx
@@ -1,0 +1,75 @@
+import React from "react";
+import classnames from "classnames";
+import { getBorderClassNames, getBorderVars } from "./utils/getBoxConfig";
+import {
+  getHeightClassName,
+  getHeightVars,
+  getWidthClassName,
+  getWidthVars,
+} from "./utils/getBoxDimension";
+import {
+  getMarginClassNames,
+  getMarginVars,
+  getPaddingClassNames,
+  getPaddingVars,
+} from "./utils/getBoxSpaces";
+import type { BoxProps } from "./Box.types";
+import styles from "./styles/Box.css";
+import gapStyles from "./styles/BoxGap.css";
+import radiusStyles from "./styles/BoxRadius.css";
+
+export function Box({
+  alignItems,
+  alignSelf,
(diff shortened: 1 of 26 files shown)


packages/components/src/Box/Box.tsx
```diff
@@ -0,0 +1,75 @@
...
+        radius && radiusStyles[`radius-${radius}`],
+        gap && gapStyles[`gap-${gap}`],
+      )}
+      style={{
+        ...getPaddingVars(padding),
+        ...getMarginVars(margin),
+        ...getHeightVars(height),
+        ...getWidthVars(width),
+        ...getBorderVars(border),
+        ...(background && { backgroundColor: `var(--color-${background})` }),
```
chris-at-jobber, line 62: Is it possible to restrict the usages to semantic tokens? With the new `--color-base` prefix on the "base" tokens that might make it easier to do automatically, but basically ensuring that "if you use an Atlantis component, it responds to theme/mode overrides" is something we can continue to confidently promise

chris-at-jobber: (Same note applies everywhere that we reference color in this file)

darryltec (commented):
This LGTM. Tho @chris-at-jobber have a good question about background colours and it totally is possible, just need to write them manually now instead of inferring from all colour tokens

chris-at-jobber:
In the name of getting this in the hands of people, I'm OK saying the semantic restrictions can be a fast-follow, but I'd hope it would be a _very_ fast follow as we have very semantic-token-reliant product work coming down the pipeline 😄

darryltec (approved):
In that case, ship it!

darryltec (approved):
go forth

PR #1891, opened 2024-05-15
revert: Revert "fix(components): Portal Menu Content"

taylorvnoj:
Reverts GetJobber/atlantis#1881

PR bringing this into JO has many visual inconsistencies as well as test failures that are surprising: https://github.com/GetJobber/Jobber/pull/46355

To avoid blocking Atlantis latest from being imported into JO, we need to revert this and revisit when we have time.

See this visual inconsistency from Business Insights https://getjobber.slack.com/archives/C034RUPSGG1/p1715740906272669


JeffMarvin (approved):
Revert looks sane and as expected.

PR #1894, opened 2024-05-15
fix(components): Allow Button variation prop combined with submit prop

jdeichert:
## Motivations

When a Button had `submit`, TS disallowed `variation` because we were extending ButtonFoundationProps instead of BaseActionProps. Confirmed with Chris, allowing `variation` combined with `submit` makes sense.

### Changed

- Updated Button prop types to allow variation + submit

---

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -70,7 +70,7 @@ interface DestructiveActionProps extends ButtonFoundationProps {
 }
 
 interface SubmitActionProps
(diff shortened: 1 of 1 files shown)


taylorvnoj (approved):
Nice fix! I added `variation="subtle"` to a button via pre-release and it:
- had no ts errors
- looked as it should with the added variation
- still behaved normally

There are [many failed snapshot tests](https://github.com/GetJobber/Jobber/pull/46826) so heads up - I don't see a problem with these. Might want to consider when is the best time to merge this PR (maybe after Menu is done). But that's up to you.

PR #1898, opened 2024-05-21
fix(components): Portalled Menu Take 2

ZakaryH:
## Motivations

This reverts commit ad35140afe6ac6756cf5167e29d9f94ef4b68286.

I want to bring back those changes, but with a couple extra little fixes

## Changes

- using `fixed` placement strategy to avoid absolute sometimes pushing content when incorrectly transformed/translated
- only apply extra y offset for small screen

for the y offset thing, the small UI achieved thru CSS used to also look at 640px as the breakpoint but it was updated in the CSS at some point however the JS part was not and that's how we got here. I think this might have been related to the issue I was seeing with the transform being incorrect after opening in the small screen state

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

pushing content

menu displacement way too much on medium screens. that code should only apply for the little bottom sheet like UX that we do thru CSS on small screens (sub 490px)

### Security

- 

## Testing

refer to the previous PR please and thank you. all those steps are still valid

the only additional thing to look out for is on screens > 490 and < 640 the menu animation should be the same as screens > 640

it shouldn't move an additional amount in that range like it was previously. only the small screen (< 490) should get that but it'll look very different since we display a totally distinct UI in that case

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/hooks/useActiveElement.ts
+++ b/packages/components/src/FormField/hooks/useActiveElement.ts
@@ -1,0 +1,19 @@
+import { useEffect, useState } from "react";
+
+export function useActiveElement() {
+  const [active, setActive] = useState(document?.activeElement);
+
+  const handleFocusIn = () => {
+    setActive(document?.activeElement);
+  };
+
+  useEffect(() => {
+    document?.addEventListener("focusin", handleFocusIn);
+
+    return () => {
+      document?.removeEventListener("focusin", handleFocusIn);
+    };
+  }, []);
+
+  return active;
+}
--- a/packages/components/src/FormField/hooks/useFormFieldFocus.ts
+++ b/packages/components/src/FormField/hooks/useFormFieldFocus.ts
@@ -1,4 +1,5 @@
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
+import { useActiveElement } from "./useActiveElement";
 
 interface UseFormFieldFocus {
   focused: boolean;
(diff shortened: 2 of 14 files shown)


ZakaryH:
so there was an issue happening where the opened Menu was pushing content down in some instances. I've been unable to reproduce it where it was discovered, however in a convoluted scenario in Storybook I was able to replicate similar behavior

the cause was that even though the opened dialog is absolutely positioned, unless the parent (body in this case) has an `overflow: hidden` then it will still try to make room for the content. the `transform` from the popper attributes/styles was getting incorrect values and then placing the content below the button like this

image

now the `body` hasn't actually grown to accommodate this as you can see. it's pushing it, and when I close the dialog it jumps back to its normal position. I **think** this is what was happening, though for me to get this was really weird. I had to trigger it in the small screen size to get the little UI, and then expand to a bigger size and trigger it again. I will say, now that I've changed the y offset logic to no longer apply on medium screens, even that setup isn't working anymore so I'm suspicious of that animation causing popper to get confused about how much space it had and translating it below

`fixed` placement doesn't have this same issue so that's why I'm using that placement strategy instead
image

this does mean that it would get cut off as shown, but popper _should_ be smart enough to know to place that above rather than below in the first place

ZakaryH:
I was finally able to hit the scenario where the Menu was pushing content all the time without strange workaround

when this is happening it is **not** actually using the new code. I don't know why this is happening. I had to do some aggressive resetting/cache clearing with`npm ci` and restarting webpack/vite for it to pick up the proper changes

I don't really get why it's not using the new code sometimes, and what it is even doing in that broken case - but what happens is that it is appending the Menu content to the same element and that's moving things. whereas the new code, the whole point of it, is portalling to the `body` and when it's doing that, especially with a `fixed` placement I'm never getting the pushed content as I'd expected

with that, I'm confident this works when the correct code is being run that is.

PR #1899, opened 2024-05-22
refactor(components): Update Styling of Tabs

MichaelParadis:
## Motivations

With the Retheme we need to update the styling of Tabs to match our new designs

### Changed

- Changed styling of Tabs to match retheme

## Testing

- Verify Tabs render correctly in the normal case and when in a modal
- Verify Tabs Overflow works in the normal case and in a modal.
- Use `git revert d6339fb1d2fcf12865680246b2e4e38527f01aaf` to get the modal story to use tabs
- Verify that this works in all browsers (Safari, Firefox and Chrome)

### After

Modal with overflow

image

Modal no overflow
image

Normal tabs
image

Normal Tabs with overflow
image

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Tabs/Tabs.css
+++ b/packages/components/src/Tabs/Tabs.css
@@ -20,6 +20,7 @@
   position: absolute;
   top: 0;
   bottom: 0;
+  z-index: var(--elevation-base);
   width: 24px;
 }
 
@@ -37,8 +38,11 @@
   display: flex;
   margin: 0;
   margin-bottom: calc(-1 * var(--border-base));
+  padding: 0;
+  border-bottom: solid var(--border-base) var(--color-border);
   overflow-x: auto;
   -webkit-overflow-scrolling: touch;
+  gap: var(--space-large);
 }
 
 ul.tabRow {
@@ -47,35 +51,43 @@ ul.tabRow {
 
 .tabRow li {
   display: flex;
+  position: relative;
   list-style: none;
 }
 
 .tab {
-  flex: 0 0 auto;
   height: var(--tab--height);
   margin: 0;
-  padding-left: var(--public-tab--inset);
-  padding-right: var(--public-tab--inset);
-  border: solid var(--border-base) var(--color-border);
-  border-left-width: 0;
+  padding: var(--space-smaller) var(--space-small);
+  border: none;
   outline: none;
+  color: var(--color-text--secondary);
   background: var(--color-surface);
   cursor: pointer;
   transition: all var(--timing-base);
+  flex: 0 0 auto;
 }
 
-.tab:hover,
-.tab:focus {
-  background-color: var(--color-surface--hover);
+.tab.selected,
+.tab:not(.selected):hover,
+.tab:not(.selected):focus {
+  color: var(--color-heading);
 }
 
-.tab:focus span {
-  box-shadow: var(--shadow-focus);
+.tab:not(.selected):focus-within {
+  outline: none;
+  background: var(--color-surface--hover);
 }
 
-.selected {
-  border-top-color: var(--color-interactive);
-  border-bottom-color: var(--color-surface);
+.selected::after {
+  content: " ";
+  display: block;
+  position: absolute;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  height: var(--space-smaller);
+  background-color: var(--color-interactive);
 }
 
 .selected:focus {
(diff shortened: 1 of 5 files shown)


packages/components/src/Tabs/Tabs.tsx
```diff
@@ -82,16 +82,14 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
...
-            {React.Children.map(children, (tab, index) => (
-              <InternalTab
-                label={tab.props.label}
-                selected={activeTab === index}
-                activateTab={activateTab(index)}
-                onClick={tab.props.onClick}
-              />
-            ))}
-          </li>
+          {React.Children.map(children, (tab, index) => (
```
MichaelParadis, line 85: Moving the `li` into `InternalTab` allows `gap` to space the tabs out evenly like there are in SG-1

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -131,22 +129,23 @@ export function InternalTab({
...
+        type="button"
+        role="tab"
+        id={label}
+        className={className}
+        onClick={event => {
+          activateTab();
+          onClick(event);
+        }}
+      >
+        <Typography element="span" size="large" fontWeight="semiBold">
```
MichaelParadis, line 145: I removed the `textColor` prop because it allows for `hover` and `focus` states to style it easily with CSS

packages/components/src/Tabs/Tabs.css
```diff
@@ -47,35 +51,43 @@ ul.tabRow {
...
 
 .tab {
-  flex: 0 0 auto;
   height: var(--tab--height);
   margin: 0;
-  padding-left: var(--public-tab--inset);
-  padding-right: var(--public-tab--inset);
-  border: solid var(--border-base) var(--color-border);
-  border-left-width: 0;
+  padding: var(--space-smaller) var(--space-small);
```
MichaelParadis, line 61: Updated padding from SG-1.
TODO: Determine if the `public-tab-inset` needs to be updated for the content

packages/components/src/Tabs/Tabs.css
```diff
@@ -47,35 +51,43 @@ ul.tabRow {
 
 .tabRow li {
   display: flex;
+  position: relative;
   list-style: none;
 }
 
 .tab {
-  flex: 0 0 auto;
```
MichaelParadis, line 54: stylelint formatted this to be the last css property on line 68

packages/components/src/Tabs/Tabs.css
MichaelParadis: Regressions were fixed in https://github.com/GetJobber/atlantis/pull/1899/commits/fcd477b82174c76fdff8398931eaf360ab57bb19

packages/components/src/Tabs/Tabs.css
```diff
@@ -11,7 +11,8 @@
 
 .overflow {
   position: relative;
-  border-bottom: solid var(--border-base) var(--color-border);
+  padding-left: var(--public-tab--inset);
```
MichaelParadis, line 14: removed the extra border because it is also on `.tabRow`. I did this so that the borderBottom wouldn't extend too far out.

packages/components/src/Tabs/Tabs.css
```diff
@@ -20,25 +21,29 @@
...
 
 .overflowRight:after {
-  right: 0;
+  right: var(--public-tab--inset);
   box-shadow: inset -16px 0 16px -16px rgba(var(--color-black--rgb), 0.25);
 }
 
 .overflowLeft:before {
-  left: 0;
+  left: var(--public-tab--inset);
```
MichaelParadis, line 34: Due to the padding on `.overFlow` the need to offset the box-shadow or else it will render in the wrong position and look incorrect

JeffMarvin:
@MichaelParadis In Arc (Chromium) I am seeing the shading on the right even after the tabs are full width/not overflowing... 
image

packages/components/src/Tabs/Tabs.tsx
JeffMarvin: Nice.

JeffMarvin (approved):
Ok, approving this... I see that the comment I made about the overflow shading now showing up after changing the viewport size is  an issue with the current component on `master` so that would be out of scope. Checked all three browsers, looks good.

PR #1901, opened 2024-05-27
fix(components): fixes a bug where we call setState inside of the react render cycle

cameron-jobber:
## Motivations

- discovered a bug where we were calling `setState` _inside_ the react render cycle (in this case inside the reducer function)
- this bug manifested with an ugly console error in both the browser and in tests
```
console.error
    Warning: Cannot update a component (`QuotesIndex`) while rendering a different component (`ForwardRef(ConfirmationModalInternal)`). To locate the bad setState() call inside `ForwardRef(ConfirmationModalInternal)`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
```

### Fixed

- moved calling of ConfirmationModal callbacks out of the reducer
- this removed a react error from the console

## Testing

- this was a pure refactor to fix the bug
- added a regression test that would fail without this change
- here is what the error would look like in the test:
```
"Warning: Cannot update a component (`%s`) while rendering a different component (`%s`). To locate the bad setState() call inside `%s`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render%s", "StatefulWrapper", "ForwardRef(ConfirmationModalInternal)", "ForwardRef(ConfirmationModalInternal)", "
        at title (/Users/cameron/workspace/atlantis/packages/components/src/ConfirmationModal/ConfirmationModal.tsx:173:5)
        at useState (/Users/cameron/workspace/atlantis/packages/components/src/ConfirmationModal/tests/interaction-ConfirmationModal.test.tsx:61:33)"
```

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
+++ b/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
@@ -57,16 +57,12 @@ function confirmationModalReducer(
       };
 
     case "confirm":
-      state.onConfirm && state.onConfirm();
-
       return {
         ...state,
         open: false,
       };
 
     case "cancel":
-      state.onCancel && state.onCancel();
-
       return {
         ...state,
         open: false,
@@ -261,6 +257,12 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
 
   function handleAction(type: "confirm" | "cancel") {
     return () => {
+      if (type === "confirm") {
+        state.onConfirm && state.onConfirm();
+      } else if (type === "cancel") {
+        state.onCancel && state.onCancel();
+      }
+
       dispatch({ type });
       onRequestClose && onRequestClose();
     };
--- a/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
+++ b/packages/components/src/ConfirmationModal/ConfirmationModal.tsx
@@ -258,9 +258,9 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
   function handleAction(type: "confirm" | "cancel") {
     return () => {
       if (type === "confirm") {
-        state.onConfirm && state.onConfirm();
+        state.onConfirm?.();
       } else if (type === "cancel") {
-        state.onCancel && state.onCancel();
+        state.onCancel?.();
       }
 
       dispatch({ type });
--- a/packages/components/src/ConfirmationModal/tests/interaction-ConfirmationModal.test.tsx
+++ b/packages/components/src/ConfirmationModal/tests/interaction-ConfirmationModal.test.tsx
@@ -54,7 +54,7 @@ test("simple ConfirmationModal should cancel", () => {
   expect(requestCloseHandler).toHaveBeenCalled();
 });
 
-test("confirm action should not happen during a render cycle", () => {
+test("confirm action should not happen during a render cycle and cause a React warning", () => {
   jest.spyOn(console, "error");
 
   function StatefulWrapper() {
--- a/packages/components/src/ConfirmationModal/tests/interaction-ConfirmationModal.test.tsx
+++ b/packages/components/src/ConfirmationModal/tests/interaction-ConfirmationModal.test.tsx
@@ -54,6 +54,35 @@ test("simple ConfirmationModal should cancel", () => {
   expect(requestCloseHandler).toHaveBeenCalled();
 });
 
+test("confirm action should not happen during a render cycle", () => {
+  jest.spyOn(console, "error");
+
+  function StatefulWrapper() {
+    const [foo, setFoo] = React.useState(true);
+
+    return (
+      <>
+        {foo ? "FOO TRUE" : "FOO FALSE"}
+        <ConfirmationModal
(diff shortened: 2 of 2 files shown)


packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -261,6 +257,12 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
 
   function handleAction(type: "confirm" | "cancel") {
     return () => {
+      if (type === "confirm") {
+        state.onConfirm && state.onConfirm();
+      } else if (type === "cancel") {
+        state.onCancel && state.onCancel();
+      }
+
```
JeffMarvin, line 265: 👍🏻 
Makes sense, thanks for catching this!

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -261,6 +257,12 @@ export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
 
   function handleAction(type: "confirm" | "cancel") {
     return () => {
+      if (type === "confirm") {
+        state.onConfirm && state.onConfirm();
+      } else if (type === "cancel") {
+        state.onCancel && state.onCancel();
+      }
```
ericchernuka, line 264:
```
      if (type === "confirm") {
        state.onConfirm?.();
      } else if (type === "cancel") {
        state.onCancel?.();
      }
```
We can simplify this with optional chaining.

packages/components/src/ConfirmationModal/tests/interaction-ConfirmationModal.test.tsx
```diff
@@ -54,6 +54,35 @@ test("simple ConfirmationModal should cancel", () => {
...
+      </>
+    );
+  }
+
+  const { getByText } = render(<StatefulWrapper />);
+
+  fireEvent.click(getByText("We Shall"));
+  expect(getByText("FOO FALSE")).toBeInTheDocument();
+  expect(console.error).not.toHaveBeenCalled();
+});
```
JeffMarvin, line 84: Nice to have a test to prevent future regressions!

Pontificating just to understand your perspective here... As written, the test strikes me as a bit blunt... effectively we're checking that NO error is logged to the console. That said, it's equally or more brittle to try to detect that a particular error string is not logged... in addition to the overall suspect nature of tests expecting something to NOT happen... Likely these issues have already occurred to you, so mostly curious to hear your reasoning behind this current solution. 

I do wonder if the test name could be more explicit to indicate what is being expected... 
```
test("confirm action should not happen during a render cycle and cause a React warning"
```

Or perhaps a comment above the last line to explain that a React error shouldn't happen? 

All just suggestions... let me know what you think.

JeffMarvin (commented):
Nice catch, thanks for putting up a fix! 
Just had a question about your test, but overall no problem with your solution.

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
cameron-jobber: Simplified in b5a4b40eb7cf754abf35c536ae2ddc6856e0b05a

packages/components/src/ConfirmationModal/tests/interaction-ConfirmationModal.test.tsx
cameron-jobber: Hey @JeffMarvin,

My perspective is that this test would be characterized as a 'regression test'. i.e. a test that exists to catch a very specific bug. As such it's a little bit odd. Had we not introduced this bug in the first place, we certainly wouldn't have written the test. You could imagine infinite _hypothetical_ regressions that we could write tests for for things that do NOT happen, as you say. Instead we just write tests for the regressions we find. The 'write a failing test that exhibits the bug, and then fix it so the test passes'.

Honestly, I haven't written so many regression tests that I have a ready pattern for it. In this case, I also contemplated
- making a new file `regressions-ConfirmationModal.test.tsx`
- adding a describe block `describe('regressions', ...)`

In this case, the problem is not the React warning. The React warning is merely the manifestation of the problem. The problem is calling `setState` inside the render cycle. But adding a note about the warning certainly makes it a bit clearer.

And of course, now any future devs who are curious can just look up this PR and get this discussion, too. 😅 

I've updated the PR with your suggestion for the `it` message in b65222ab4cd445e17b0b0e9c4e27d2d42607c0e0

Happy to discuss this more! Maybe there's a better pattern we can identify for this kind of regression test.

JeffMarvin: That all makes sense -- appreciate the additional context. I think trying to put a patten in for regression tests is a problem we could solve _after_ we achieve the institutional behavior in for adding regression tests when we find them.

JeffMarvin (approved):
LGTM - thanks for the contribution and the test! 

Once I merge this, a new release of the `@jobber/components` package will be ready for you to bring in to your application. Happy to assist with any questions you have on that front!

PR #1909, opened 2024-05-31
feat(components): add side drawer

darryltec:
## Motivations

We've been recreating side drawers on our repos so it's time to make it an official atlantis component

### Added

- Side drawer component

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


chris-at-jobber:
@darryltec few initial thoughts/questions

- this removes the `variation` prop from the in-usage SideDrawer, any reason to exclude that?
- this adds a border to the bottom of the header that isn't present in "in-Jobber" usages (in fairness, many instances do have a border between the content and the title, but that's added by the child content in those instances)
- thoughts on adding a `SideDrawer.Back` action to reduce some of the inconsistency in arrow usage you pointed out last week? (can also be a "design side" solve)

chris-at-jobber:
Also, I think we'll need a **Related components** section that mentions Drawer and Modal (if you need some content on when/why to use each of these, I can work on that)

docs/components/SideDrawer/SideDrawer.stories.mdx
```diff
@@ -0,0 +1,76 @@
+import { Meta, ArgsTable } from "@storybook/addon-docs";
+import { SideDrawer } from "@jobber/components/SideDrawer";
+
+<Meta title="Components/Overlays/SideDrawer/Docs" />
+
+# Side Drawer
+
+SideDrawers are overlays that allow users to view, edit, or show information
```
chris-at-jobber, line 8: >view, edit, or show

view/show are the same, right?

darryltec:
@chris-at-jobber 

> this removes the variation prop from the in-usage SideDrawer, any reason to exclude that?

No reason other than I moved the one in the FE repo than the "in-jobber" one 😆 , I'll add it!

> this adds a border to the bottom of the header

yeah, almost all of them seems to use a border that i just added it in. I wanted to shadow it up when scrolling tho but got lazy. What do you think about removing the border and add a shadow on scroll/is scrolled?

> thoughts on adding a SideDrawer.Back action to reduce some of the inconsistency in arrow usage

Gotta admit, i did not like that pattern 😆 like, if it's a deeper page, open another side sheet on top of each other or like this 
image

or the ios style where it's just peeking 😆 that said, if that's the pattern then yeah, I'm cool adding it. Just weird when you use it to close the drawer when it isn't opened on another drawer. Ya feel me?

docs/components/SideDrawer/SideDrawer.stories.mdx
darryltec: gotta get my word count up yo 😆

chris-at-jobber:
>removing the border and add a shadow on scroll/is scrolled?

That would be very nice!

chris-at-jobber:
Couple more thoughts

- any thoughts on what element to use semantically for this? The "inline" `Drawer` uses an aside but this feels more like a `dialog`, as the behaviour is virtually identical to a Modal
- does this have focus trapping and "return focus to activator" set up? (again, similar to Modal, feels like it should)

darryltec:
@chris-at-jobber should be a dialog as it blocks interaction on anything outside of it as it traps the focus once opened. Also puts back focus to activator, yes. Gotchu fam!

scotttjob (approved):
Approving to enable. I don't see anything scary in here.

My only concern is if this is SSR-proof, so double check that, otherwise it should be good to go (at least for v1)

chris-at-jobber:
Open to doing this as an improvement in a separate PR, but it would be really nice if we could handle the back button l[ike this](https://www.figma.com/proto/899biwQC5nAo0e6uICeltn/SideDrawer?page-id=66%3A275&node-id=66-283&viewport=1056%2C788%2C1&t=UcR2Dos84E54SPY9-1&scaling=min-zoom&starting-point-node-id=66%3A395)?

darryltec:
@chris-at-jobber ah yeah that be an improvement haha 😆 i have a few ideas to do that, nothing concrete yet tho

PR #1910, opened 2024-05-31
feat(components): Add url to DataList.ItemAction

taylorvnoj:
## Motivations

`DataList.ItemAction` can be a url 

Kapture 2024-06-03 at 18 56 50

### Added

- new optional `actionUrl` prop on `DataListActionProps` that is an `<a>` tag
- tests and small documentation update

### Security

- using `rel="noopener noreferrer"`in the `<a>` for safety reasons 🦺 

## Testing

- A `DataList.ActionItem` can be a url 
- it appears and behaves like the other actions in that same menu (focus, keyboard events, etc)
- Changes can be tested here: https://github.com/GetJobber/Jobber/pull/47034

---

Random photo of Atlantis


packages/components/src/DataList/components/DataListAction/DataListAction.tsx
```diff
@@ -34,6 +35,22 @@ export function DataListAction<T extends DataListObject>({
     }
   }
 
+  if (actionUrl) {
+    return (
+      <a
+        href={actionUrl}
+        target="_blank"
```
chris-at-jobber, line 42: Opening in a new tab without the user's intent is an anti-pattern and if it happens by surprise can make the experience seem broken as it appears in the current tab that nothing is happening. I know in the demo we got from David the case was to allow multiple workflows to be kicked off without revisiting the DataList, but that was with him cmd+clicking.

packages/components/src/DataList/components/DataListAction/DataListAction.tsx
```diff
@@ -34,6 +35,21 @@ export function DataListAction<T extends DataListObject>({
     }
   }
 
+  if (actionUrl) {
+    return (
+      <a
```
chris-at-jobber, line 40: Thank you 😄  🙏

packages/components/src/DataList/components/DataListAction/DataListAction.tsx
taylorvnoj: Ahhhh ok! I read "new tab" in the ticket and went with it, but now I understand why we shouldn't do that. Thanks Chris!

packages/components/src/DataList/DataList.types.ts
```diff
@@ -350,6 +350,11 @@ export interface DataListActionProps<T extends DataListObject> {
    * The callback function when the action is clicked.
    */
   readonly onClick?: (data: T) => void;
+
+  /**
+   * The URL to navigate to when the action is clicked.
+   */
+  readonly actionUrl?: string;
 }
```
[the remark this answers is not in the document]
taylorvnoj: Oh that's a great idea! Yes we do have some _undesirable_ XOR happenings 😆  but I fully agree that this would be a great use of it.  I can absolutely see confusion happening around the `onClick` & `actionUrl` trying to be used at the same time.  I'll update this!

chris-at-jobber:
Is there a case where we actually do/might need both? (I'm just thinking of a team wanting to add onClick analytics tracking on a url action - is that valid?)

chris-at-jobber:
To my comment above, we talked Offline™ and resolved this. No further questions!

PR #1911, opened 2024-05-31
docs(components): Added more description about how Form onSubmit works with throwing errors. 

suark:
## Motivations

We had a bug during development about 2 months ago. It resulted when an error was handled during the `onSubmit` function of the form, preventing it from bubbling up. This didn't allow the form library to catch the error and trigger the `onSubmitError` function.

### Changed

Added some description to the inline comment for form's `onSubmit` function and added some sentences to the form's story document page. 

## Testing

n/a

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/Form/types.ts
+++ b/packages/components-native/src/Form/types.ts
@@ -73,7 +73,10 @@ export interface FormProps<T extends FieldValues, SubmitResponseType> {
   onBeforeSubmit?: (data: FormValues<T>) => Promise<boolean>;
 
   /**
-   * A callback function that handles the submission of form data
+   * A callback function that handles the submission of form data.
+   * If errors happen during submission, the error must not be caught and handled without throwing it again.
+   * If no error is thrown, the onSubmitSuccess callback will be called.
+   * If an error is thrown, the onSubmitError callback will be called.
    */
   onSubmit: (data: FormValues<T>) => Promise<SubmitResponseType>;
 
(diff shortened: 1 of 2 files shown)


packages/components-native/src/Form/types.ts
```diff
@@ -73,7 +73,10 @@ export interface FormProps<T extends FieldValues, SubmitResponseType> {
   onBeforeSubmit?: (data: FormValues<T>) => Promise<boolean>;
 
   /**
-   * A callback function that handles the submission of form data
+   * A callback function that handles the submission of form data.
+   * If errors happen during submission, the error must not be caught and handled without throwing it again.
+   * If no error is thrown, the onSubmitSuccess callback will be called.
+   * If an error is thrown, the onSubmitError callback will be called.
```
taylorvnoj, line 79: What do you think about just a _slight_ change in wording?

```suggestion
   * A callback function that handles the submission of form data.
   * If an error occurs during submission, it should not be caught and handled silently; the error must be thrown again.
   * If the submission is successful and no error is thrown, the `onSubmitSuccess` callback will be called.
   * If an error is thrown, the `onSubmitError` callback will be called.
```

suark: Thanks for the suggestion! It looks great.

taylorvnoj (approved):
LGTM - thanks again!

PR #1912, opened 2024-06-03
fix(components): Fix Menu Focus Logic

ZakaryH:
## Motivations

While this worked in Jest/React/Storybook and some other React contexts, with a particular setup in product it was failing.

In a breakpoint you could see 3 different values for `activeElement` `document.activeElement` and `activeElementRef`.

Only `document.activeElement` was correct at that moment and as a consequence, the check we were doing in `useFormFieldFocus` that looks to see if the current active element has a parent with the special focus data attributes on it was failing and causing the focus to go elsewhere and close the menu instantly

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Somewhat interesting since this was always working in Storybook. Make sure it's still good here, and by that I mean
- inputText that is `clearable: while-editing`, the clear button should correctly show when you have text - then hide when the input loses focus
- inputText that has a `toolbar` and `toolbarVisibility: while-editing` - same idea. the toolbar should only show when either the input or the toolbar content is focused

- the most complicated version is where you have a `multiline` `InputText` with `toolbar` where the markup for the toolbar is a `Menu`. opening the Menu and having the focus correctly moved to the first item in that menu shouldn't cause the toolbar to close

then, in product try it out on Quote create/edit where we were seeing the problems in the first place. on both small and large screens it should work.

Changes can be

---

Random photo of Atlantis


packages/components/src/FormField/hooks/useFormFieldFocus.ts
```diff
@@ -19,12 +18,6 @@ export function useFormFieldFocus(
   props: UseFormFieldFocusProps,
 ): UseFormFieldFocus {
   const [focused, setFocused] = useState(false);
-  const activeElement = useActiveElement();
-  const activeElementRef = useRef(activeElement);
-
-  useEffect(() => {
-    activeElementRef.current = activeElement;
```
ZakaryH, line 26: previously we were trying to use refs to make sure that we weren't getting a stale version of `activeElement` since our handlers are not getting re-built

however since `document.activeElement` is outside React anyway, we can just use that directly and not have to over complicate things trying to ensure we're not using stale state

packages/components/src/FormField/hooks/useFormFieldFocus.ts
```diff
@@ -56,7 +49,7 @@ export function useFormFieldFocus(
         handleFocusOut,
       );
     };
-  }, [props.wrapperRef?.current]);
```
ZakaryH, line 59: this wasn't doing anything since ref updates don't cause re-renders. we should look at getting the exhaustive deps linting rule in here. I believe it would have flagged this.

taylorvnoj: seems like `exhaustive-deps` rule would be relatively simple to add.  Maybe something to put side desk or write a ticket for

taylorvnoj (approved):
This is great! All testing scenarios you listed look/work great. I also used the pre-release to test in product and the previous issues are no longer happening 🎉  Thanks for all the hard work on this Zak

PR #1915, opened 2024-06-05
feat(components): improve positioning of DataList actions [JOB-99926]

chris-at-jobber:
## Motivations

When the content in the last column of a DataList row is full-width or right-aligned, the "hover" actions on the row can be obscured. This violates the ["recognition over recall" principle](https://www.nngroup.com/articles/recognition-and-recall/).

### Changed

- reduced the size of the DataList hover actions and changed their offset to avoid covering contents of the rightmost column

## Testing

Hover over a row

Changes can be

---

Random photo of Atlantis


packages/components/src/DataList/components/DataListItemActions/DataListItemActions.tsx
```diff
@@ -21,8 +21,8 @@ export function DataListItemActions<T extends DataListObject>(
 }
 
 const variants: Variants = {
-  hidden: { opacity: 0, y: 10 },
-  visible: { opacity: 1, y: 0 },
+  hidden: { opacity: 0, y: 0 },
```
chris-at-jobber, line 24: Adjusts the animation to map better to the new position

packages/components/src/DataList/components/DataListActions/DataListActions.tsx
```diff
@@ -54,7 +54,7 @@ export function DataListActions<T extends DataListObject>({
                   (props.onClick as () => void)?.();
                 }
               }}
-              type="secondary"
+              type="tertiary"
```
chris-at-jobber, line 57: Tertiary button in the "individual" actions

packages/components/src/DataList/components/DataListItemActionsOverflow/DataListItemActionsOverflow.tsx
```diff
@@ -22,7 +22,7 @@ export function DataListItemActionsOverflow<T extends DataListObject>({
         <Button
           icon="more"
           ariaLabel="More actions"
-          type="secondary"
+          type="tertiary"
```
chris-at-jobber, line 25: tertiary button in the "more" actions

unclear to me at this point why this action is different from the [DataList actions](https://github.com/GetJobber/atlantis/pull/1915/files#r1764033275) but here we are

packages/components/src/DataList/components/DataListActions/DataListActions.tsx
```diff
@@ -25,7 +24,7 @@ export function DataListActions<T extends DataListObject>({
   childrenArray.splice(0, exposedActions.length);
 
   return (
-    <DataListOverflowFade>
+    <>
```
chris-at-jobber, line 27: This impacts both the "row" actions and the "bulk" actions, but has no negative impact on either instance:

image

image
image

chris-at-jobber: At one point the design direction was to have the bulk actions be scrollable, which is likely why this was implemented, but you can see that they collapse down to the ••• menu once the viewport is narrowed so the gradient fade support added by `DataListOverFlowFade` was not giving us anything, and was in turn:

- coupling the spacing of these actions to the spacing in the filters bar
- clipping the focus shadow of the actions

So two birds with one stone, as they say

packages/components/src/DataList/components/DataListItemActions/DataListItemActions.css
```diff
@@ -1,13 +1,14 @@
 .menu {
   display: flex;
   position: absolute;
-  top: calc(var(--space-small) * -1);
+  top: -26px;
```
chris-at-jobber, line 4: Yes this number is relatively magic

packages/components/src/DataList/components/DataListActions/DataListActions.tsx
```diff
@@ -25,7 +24,7 @@ export function DataListActions<T extends DataListObject>({
   childrenArray.splice(0, exposedActions.length);
 
   return (
-    <DataListOverflowFade>
+    <span data-testid="ATL-DataListActions">
```
chris-at-jobber, line 27: DataListOverFlowFade came with a testid that was being referenced in BatchActions test - this preserves the test

packages/components/src/DataList/components/DataListItemActions/DataListItemActions.css
```diff
@@ -1,13 +1,12 @@
...
   right: 0;
   z-index: var(--elevation-menu);
   box-shadow: var(--shadow-base);
-  padding: var(--space-smaller);
```
chris-at-jobber, line 8: Since padding is now 0, no need to declare it

packages/components/src/DataList/components/DataListItemActions/DataListItemActions.css
```diff
@@ -1,13 +1,12 @@
...
   border: var(--border-base) solid var(--color-border);
   border-radius: var(--radius-base);
   background-color: var(--color-surface);
-  gap: var(--space-smaller);
+  gap: var(--space-smallest);
```
chris-at-jobber, line 11: Keeps a pinch of space between the actions so the focus states don't encroach too heavily on sibling buttons

packages/components/src/DataList/components/DataListOverflowFade/DataListOverflowFade.css
```diff
@@ -1,6 +1,5 @@
 .fadeContainer {
-  --overflow-fade--offset: var(--space-smaller);
-  --overflow-fade--negative-offset: calc(var(--overflow-fade--offset) * -1);
```
chris-at-jobber, line 3: These values are now the same and can be applied accordingly with one variable

I considered removing the margin and padding declarations, but since the positions of the gradients are coupled to those values, felt worth keeping them in sync

packages/components/src/DataList/components/DataListActions/DataListActions.tsx
taylorvnoj: hmmm, we have a file, `packages/components/src/DataList/components/DataListBulkActions/DataListBulkActions.const.ts` that holds

```
export const BULK_ACTIONS_CONTAINER_TEST_ID =
  "ATL-DataListBulkActions-Container";
```

Which actually isn't being used anywhere. We should use `BULK_ACTIONS_CONTAINER_TEST_ID` instead, but not in this file.

We should:
- remove the span in `DataListActions.tsx`, it just becomes empty fragment
-  `DataListBulkActions.tsx` would now have the bulk actions test id:

```tsx
import React from "react";
import styles from "./DataListBulkActions.css";
import { BULK_ACTIONS_CONTAINER_TEST_ID } from "./DataListBulkActions.const";
import { DataListBulkActionsProps } from "../../DataList.types";
import { useDataListContext } from "../../context/DataListContext";
import { DataListActions } from "../DataListActions";
import { useResponsiveSizing } from "../../hooks/useResponsiveSizing";

// This component is meant to capture the props of the DataList.BulkActions
export function DataListBulkActions(
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  _: DataListBulkActionsProps,
) {
  return null;
}

export function InternalDataListBulkActions() {
  const { bulkActionsComponent } = useDataListContext();
  if (!bulkActionsComponent) return null;

  const { children } = bulkActionsComponent.props;

  const { sm } = useResponsiveSizing();

  // Collapse all actions under "More actions" when breakpoint is smaller than sm
  const itemsToExpose = sm ? 3 : 0;

  return (
    <div
      className={styles.bulkActions}
      data-testid={BULK_ACTIONS_CONTAINER_TEST_ID}
    >
      <DataListActions itemsToExpose={itemsToExpose}>
        {children}
      </DataListActions>
    </div>
  );
}
```

- then in `DataListBulkActions.test.tsx` you put that testid in there instead:
```tsx
import { BULK_ACTIONS_CONTAINER_TEST_ID } from "./DataListBulkActions.const";

describe("DataListBulkActions", () => {
  it("should render the 4 buttons", () => {
    renderComponent();
    const overflowContainer = screen.getByTestId(
      BULK_ACTIONS_CONTAINER_TEST_ID,
    );

    expect(within(overflowContainer).getAllByRole("button")).toHaveLength(4);
  });
  ```
  
  This way, we're using an unused testid, which is more correct to what we're testing. Let me know if I'm not communicating this well enough we can hop on a call

File: packages/components/src/DataList/components/DataListItemActionsOverflow/DataListItemActionsOverflow.tsx (hunk 3 above)
Comment (taylorvnoj) [reply packages/components/src/DataList/components/DataListItemActionsOverflow/DataListItemActionsOverflow.tsx:25]: It's a good separation of concerns.
`DataListActions` are rendering the primary actions we see on hover. `DataListItemActionsOverflow` is that ••• menu button and needs to handle state for opening/closing the menu

I think we could also pretty easily extract & simplify `DataListItemActionsOverflow` to share with other menus that would have the same behaviour. Obviously not something we'll do now but, hope that helps clear things a little

**Review by taylorvnoj — CHANGES_REQUESTED** [review]
Looking really really good - we're no longer blocking any info 👌 

Just that one change with the testid and this will be good to go!

Screenshot 2024-09-18 at 5 53 44 PM

**Comment by taylorvnoj** [comment]
@chris-at-jobber for the failed `lint_javascript`, to see the errors, run:
```
npm run lint:js 
```

These two have to be fixed:

Screenshot 2024-09-18 at 7 11 06 PM

PR #1917, opened 2024-06-05
fix(components): Handle dynamic number of Tab children

CarlyHarly:
## Motivations

SC Engagement wants to dynamically change the number of Tab children in a Tabs component based on some selection criteria. (See /reviews page -> Manage Drawer -> _Automatically ask for a review ..._ . When the delivery method changes, we want to show the associated tab or tabs.)

Problem: When the `activeTab` is greater than that new number of Tab children then we see the white-screen-of-death because we are dereferencing nil objects. 

Repro steps (for this [PR](https://github.com/GetJobber/Jobber/pull/47139))
1. Select Email and SMS delivery method and see two tabs, select SMS tab.
2. Select SMS deliver method
Expected: SMS tab is displayed (see video below)
Actual: white-screen-of-death

https://github.com/GetJobber/atlantis/assets/85175844/b8e5c291-e492-4ba1-bf72-1db39ac614c8

## Changes

- Listen for changes to the number of children Tab and set the `activeTab` to `activeTabInitialValue`

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Tabs/Tabs.tsx
+++ b/packages/components/src/Tabs/Tabs.tsx
@@ -52,7 +52,7 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
 
   const activeTabProps = (React.Children.toArray(children) as ReactElement[])[
     activeTab
-  ].props;
+  ]?.props;
 
   const handleOverflowing = () => {
     if (tabRow.current) {
@@ -78,6 +78,12 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
     };
   });
 
+  useEffect(() => {
+    if (activeTabInitialValue !== activeTab) {
+      setActiveTab(activeTabInitialValue);
+    }
+  }, [React.Children.count(children)]);
+
   return (
     <div className={styles.tabs}>
       <div className={overflowClassNames}>
@@ -95,9 +101,9 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
       <section
         role="tabpanel"
         className={styles.tabContent}
-        aria-label={activeTabProps.label}
+        aria-label={activeTabProps?.label}
       >
-        {activeTabProps.children}
+        {activeTabProps?.children}
       </section>
     </div>
   );
--- a/packages/components/src/Tabs/Tabs.tsx
+++ b/packages/components/src/Tabs/Tabs.tsx
@@ -1,13 +1,7 @@
-import React, {
-  MutableRefObject,
-  ReactElement,
-  ReactNode,
-  useEffect,
-  useRef,
-  useState,
-} from "react";
+import React, { ReactElement, ReactNode, useEffect, useState } from "react";
 import classnames from "classnames";
 import styles from "./Tabs.css";
+import { useTabsOverflow } from "./hooks/useTabsOverflow";
 import { Typography } from "../Typography";
 
 interface TabsProps {
@@ -31,10 +25,7 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
   const activeTabInitialValue =
     defaultTab < React.Children.count(children) ? defaultTab : 0;
   const [activeTab, setActiveTab] = useState(activeTabInitialValue);
-  const [overflowRight, setOverflowRight] = useState(false);
-  const [overflowLeft, setOverflowLeft] = useState(false);
-  const tabRow = useRef() as MutableRefObject<HTMLUListElement>;
-
+  const { overflowRight, overflowLeft, tabRow } = useTabsOverflow();
   const overflowClassNames = classnames(styles.overflow, {
     [styles.overflowRight]: overflowRight,
     [styles.overflowLeft]: overflowLeft,
@@ -53,30 +44,6 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
   const activeTabProps = (React.Children.toArray(children) as ReactElement[])[
     activeTab
   ]?.props;
-
-  const handleOverflowing = () => {
-    if (tabRow.current) {
-      const scrollWidth = tabRow.current.scrollWidth;
-      const clientWidth = tabRow.current.clientWidth;
-      const maxScroll = scrollWidth - clientWidth;
-      const scrollPos = tabRow.current.scrollLeft;
-
-      if (scrollWidth > clientWidth) {
-        setOverflowRight(scrollPos >= 0 && scrollPos != maxScroll);
-        setOverflowLeft(scrollPos > 0 && scrollPos < scrollWidth);
-      }
-    }
-  };
-
-  useEffect(() => {
-    handleOverflowing();
-    tabRow.current &&
-      tabRow.current.addEventListener("scroll", handleOverflowing);
-
-    return () => {
-      window.removeEventListener("scroll", handleOverflowing);
-    };
-  });
 
   useEffect(() => {
     if (activeTab > (React.Children.count(children) - 1)) {
--- a/packages/components/src/Tabs/Tabs.tsx
+++ b/packages/components/src/Tabs/Tabs.tsx
@@ -79,7 +79,7 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
   });
 
   useEffect(() => {
-    if (activeTabInitialValue !== activeTab) {
+    if (activeTab > (React.Children.count(children) - 1)) {
       setActiveTab(activeTabInitialValue);
     }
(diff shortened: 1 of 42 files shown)


CarlyHarly:
General question: how should we be handling dynamic tabs? Instead, should we unmount the Tab component and remount with new tabs? Or something else?

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -78,6 +78,12 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
     };
   });
 
+  useEffect(() => {
```
CarlyHarly, line 81: Is there a better approach than this? Will this mess up any other scenarios?

CarlyHarly: I didn't run or add unit tests yet. I can do that once we are aligned on the solution or guidance. Thanks!

ZakaryH:
haven't fully come to a conclusion about whether this approach or another might be best, but I did notice that we'd need to account for updating the overflow logic so just gonna make note of that

right now it seems to be set once based on measurements/Children(count) and then not updated with the count changing
Kapture 2024-06-05 at 15 53 51

CarlyHarly:
> right now it seems to be set once based on measurements/Children(count) and then not updated with the count changing

I don't really follow, does this mean on master or in the PR? And what seems to be set?

ZakaryH:
>I don't really follow, does this mean on master or in the PR? And what seems to be set?

technically both, but on master that's the correct behavior since it's not expecting the count to ever change whereas now we're making changes to support a dynamic count of Tabs and it's not updating based on that


```
  const handleOverflowing = () => {
    if (tabRow.current) {
      const scrollWidth = tabRow.current.scrollWidth;
      const clientWidth = tabRow.current.clientWidth;
      const maxScroll = scrollWidth - clientWidth;
      const scrollPos = tabRow.current.scrollLeft;

      if (scrollWidth > clientWidth) {
        setOverflowRight(scrollPos >= 0 && scrollPos != maxScroll);
        setOverflowLeft(scrollPos > 0 && scrollPos < scrollWidth);
      }
    }
  };

  useEffect(() => {
    handleOverflowing();
    tabRow.current &&
      tabRow.current.addEventListener("scroll", handleOverflowing);

    return () => {
      window.removeEventListener("scroll", handleOverflowing);
    };
  });
```

it'll have to be somewhere in here that is either not getting called again, or has stale values that make it believe that it is overflowing despite the fact that it may now have different dimensions with less (or more!) tabs than it did on first render

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -78,6 +78,12 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
...
+    if (activeTabInitialValue !== activeTab) {
```
CarlyHarly, line 82: Carly reconsider this case: the activeTab > React.Children.count(children)

CarlyHarly: Updated to only call setActiveTab if activeTab is greater than the number of new tabs.

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -52,7 +52,7 @@ export function Tabs({ children, defaultTab = 0, onTabChange }: TabsProps) {
 
   const activeTabProps = (React.Children.toArray(children) as ReactElement[])[
     activeTab
-  ].props;
+  ]?.props;
 
   const handleOverflowing = () => {
     if (tabRow.current) {
```
CarlyHarly, line 58: @ZakaryH Where does tabRow.current get set? I see it gets passed to a `ul` element.

ZakaryH: that's actually where it is being set!

```
<ul ... ref={tabRow}>
```

is basically shorthand for

```
<ul ... ref={(ulRef) => tabRow.current = ulRef}>
```

ZakaryH:
looked into why the overflow state isn't getting updated, it's because of the condition around the setters

```

  const handleOverflowing = () => {
    if (tabRow.current) {
      const scrollWidth = tabRow.current.scrollWidth;
      const clientWidth = tabRow.current.clientWidth;
      const maxScroll = scrollWidth - clientWidth;
      const scrollPos = tabRow.current.scrollLeft;

      if (scrollWidth > clientWidth) {
        setOverflowRight(scrollPos >= 0 && scrollPos != maxScroll);
        setOverflowLeft(scrollPos > 0 && scrollPos < scrollWidth);
      }
    }
  };
```

if I remove `scrollWidth > clientWidth` this correctly recognizes that the dimensions have changed with a new set of tabs, and correctly calls `setOverflowRight(true)`

trying to think of why we would put that assignment behind a conditional. seems like it's configured to `setOverflowRight` and `setOverflowLeft` to the correct values

might simply be to prevent calling that when we know it's not necessary, and saving ourselves the re-render that would be caused by the state update?

packages/components/src/Tabs/Tabs.tsx
CarlyHarly: Ah neato! Thanks!

CarlyHarly:
> hopefully something more elegant than that, but that would be the general idea. this would prevent the Tabs from getting confused by having a different amount of Tabs.
> do you think that would work at all, or is there something that is impractical/undesirable if we did that? 

I think for our use case where we are alternating between 1 and 2 tabs, it's a reasonable approach (although not as elegant).

ZakaryH:
last commit might look like a lot but it's really just

- fixing `window.removeEventListener("scroll", handleOverflowing);` to remove the event listener from the ref rather than the `window`
- adding a test for overflowRight
- extracting overflow logic into a hook because the file was hitting the maximum length rule
- removing the conditions in the scroll handler that would cause the `setOverflow` methods to not get fired even though we are hitting the scrollhandler - this allows the overflow classes to be correctly modified as the tabs are dynamically added/removed
- throttled the handler because it was firing quite a few times, and sometime it will call those state setters & cause re-renders

ZakaryH:
also if you want to play around with my weird Storybook example code here's what I've been using

<details>
  <summary>Tabs webstory dummy code</summary>

  ```

const BasicTemplate: ComponentStory<typeof Tab> = args => {
  const [tabs, setTabs] = useState(getTabs(9));
  console.log(tabs);

  return (
    <>
      <select
        onChange={e => setTabs(getTabs(parseInt(e.currentTarget.value, 10)))}
      >
        {Array.from({ length: 10 }, (_, i) => i).map(option => (
          <option key={option} value={option} selected={9 === option}>
            {option}
          </option>
        ))}
      </select>
      <Tabs defaultTab={1}>{tabs.map(tab => tab)}</Tabs>
    </>
  );
};

function getTabs(count: number): any[] {
  const tabbers = [
    <Tab key="eggs" label="Eggs">
      eggs
    </Tab>,
    <Tab key="cheese" label="cheese">
      Cheese
    </Tab>,
    <Tab key="oats" label="oats">
      oats
    </Tab>,
    <Tab key="yogurt" label="yogurt">
      yogurt
    </Tab>,
    <Tab key="apples" label="apples">
      apples
    </Tab>,
    <Tab key="oranges" label="oranges">
      oranges
    </Tab>,
    <Tab key="cherries" label="cherries">
      cherries
    </Tab>,
    <Tab key="grapes" label="grapes">
      grapes
    </Tab>,
    <Tab key="bananas" label="bananas">
      bananas
    </Tab>,
    <Tab key="pears" label="pears">
      pears
    </Tab>,
  ];

  return tabbers.slice(0, count);
}
```
</details>

ZakaryH:
so that should be accounting for the overflow now. I think the only other things I'd like to see on this PR would be a test for adding/removing a tab dynamically and ensuring things behave as expected there as far as which tab is selected and the defaultTab prop being applied correctly!

CarlyHarly:
>  I think the only other things I'd like to see on this PR would be a test for adding/removing a tab dynamically and ensuring things behave as expected there as far as which tab is selected and the defaultTab prop being applied correctly!

@GetJobber/atlantis Updated this PR with specs for adding/removing a tab dynamically. PR is ready for 👀

taylorvnoj (approved):
LGTM! Apologies for the delay in getting this approved!
I tested in Storybook and that works really well. I also ran a pre-release (via GitHub Actions) to run this change through CircleCi in product and there were no blockers there.

It seems like you have a work around for this already, but if you need to bump product with the latest `@jobber/components` version once this PR is published (~15 mins) and bring this change in, you're welcome to tag me as reviewer on that. 

:shipit:

PR #1918, opened 2024-06-05
fix(components): Fix DataList Sticky Header Extra Width [JOB-96126]

ZakaryH:
## Motivations

Right now there's an issue on medium and below screens where you can scroll horizontally because of some extra unintended width

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

Even though we're setting a `width: 0` on this parent element

image
the nested FormField elements still have a width
image
image

these are causing the layout to have horizontal scrolling
Kapture 2024-06-05 at 14 11 41

with that `width: 0` we're clearly trying to make them take up no space, and have them be functionally invisible. so, instead of doing it with a `width` I'm doing it with `display: none`

I tried doing some conditionals in the JSX via `visible && (....` however `visible` is `false` on big screens which isn't right. in that case it is definitely visible, we're just not accounting for that fact
### Security

- 

## Testing
this can all be done in Storybook

head on over to the DataList basic example. swap to a mobile view, see that you can pull it back and forth horizontally on master

see that you can no longer do that on this PR

additionally
- on large screens, the search should always be there - no button needed to toggle visibility
- on medium and below screen, the search button should be there and when you press it the search input shows AND it is auto focused then when you press the X to hide the search input, it goes away again
- small/mobile screens same thing as medium and below

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
+++ b/packages/components/src/DataList/components/DataListSearch/DataListSearch.css
@@ -17,9 +17,9 @@
   background-color: var(--color-surface);
   opacity: 0;
   transform: translateY(-50%);
-  transition: 
+  transition:
     opacity var(--transition-properties),
-    width var(--transition-properties), 
+    width var(--transition-properties),
     visibility var(--transition-properties);
 
   @media (--medium-screens-and-up) {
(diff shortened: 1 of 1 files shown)


ZakaryH:
shoot, one thing I noticed is that this changes the animation considerably where previously it would feel like it is growing and animating nicely

with `display` we won't get that same effect at all. might need to change the approach a bit to maintain that animation in a different way.

ZakaryH:
solved this another way in a different PR

closing

PR #1919, opened 2024-06-07
fix(components): use semantic value for Datepicker background

chris-at-jobber:
## Motivations

The wrapper for the `react-datepicker` used in Datepicker is hard-coded to #fff, meaning when you define a new `surface` color, the component doesn't respond and you get "show-through" between the cells

image

### Fixed

- Datepicker respondes to semantic updates

## Testing

Override `--color-surface` token in inspector stylesheet

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.css
+++ b/packages/components/src/DatePicker/DatePicker.css
@@ -15,6 +15,7 @@
   border-color: var(--color-border);
   border-radius: var(--radius-base);
   font-family: var(--typography--fontFamily-normal);
+  background-color: var(--color-surface);
 }
 
 .datePicker.inline {


ZakaryH:
noticed there's some hard coded borders around the whole thing and border bottom on the header that won't toggle with a theme change

Components___Selections___DatePicker___Web_-_Basic_⋅_Storybook

wondering if we should update those too while we're at it so that it won't always have a white-ish border

chris-at-jobber:
@ZakaryH we could be more explicit about it, but that border color is already being over-written by the `border-color` here:
image

(See when you swap it for red, or the outlandish color of your choosing)
image

ZakaryH (approved):
LGTM

PR #1920, opened 2024-06-07
feat(components): modify the Page component's title prop to accept a ReactNode 

MattDekinder:
## Motivations

For Campaigns designs we want to include a status label after the title text.  
image

## Changes

The page component title is now a ReactNode rather than a mere string. A string value will still format in the same way

### Changed

- The page component title prop is now a ReactNode

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Page/Web.stories.tsx
+++ b/docs/components/Page/Web.stories.tsx
@@ -3,6 +3,9 @@ import { ComponentMeta, ComponentStory } from "@storybook/react";
 import { Page } from "@jobber/components/Page";
 import { Content } from "@jobber/components/Content";
 import { Text } from "@jobber/components/Text";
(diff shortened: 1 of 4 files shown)


MattDekinder:
@GetJobber/atlantis Hi Atlantis, I'd love if we could open the page title prop up to components and I see that we have similar flexibility in new Atlantis components (eg [SideDrawer](https://github.com/GetJobber/atlantis/pull/1909)). I haven't used the same technique to pass the component as a child. Is that a problem?

I also note that Storybook is unable to render changes with a ReactNode `string` example input, again, is this a concern? Just let me know if you'd prefer that I go a different way with the changes here

MattDekinder:
We're going to put up a new PR for this

PR #1924, opened 2024-06-10
feat(components): update interactive borders and avatar [JOB-97530]

chris-at-jobber:
## Motivations

We need to be able to target the borders of interactive elements for theming, and for future improvements to accessibility of interactive elements.

Also want to be able to do this iteratively, so we can implement the "hook" before we use the "hook".

### Added

- 

### Changed

- Underlying semantic value for borders of interactive elements
- Also made the text color in Avatar _non-semantic_ as there are computations being done in that component to determine whether the text should be light or dark based on background color.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Avatar/Avatar.css
+++ b/packages/components/src/Avatar/Avatar.css
@@ -15,7 +15,7 @@
   color: var(--color-heading);
   font-size: var(--avatar-size);
   -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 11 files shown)


ZakaryH (approved):
LGTM

PR #1925, opened 2024-06-10
fix(components): Update InternalThumbnailImage to retry getting the thumbnail if there is a 404 

tyler-santos:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormatFile/InternalThumbnailImage.tsx
+++ b/packages/components/src/FormatFile/InternalThumbnailImage.tsx
@@ -4,6 +4,7 @@ import styles from "./InternalThumbnailImage.css";
 import { FileUpload } from "../InputFile";
 import { Glimmer } from "../Glimmer";
 
+const RETRY_COUNT = 4;
 interface InternalThumbnailImageProps {
   readonly file: FileUpload;
 }
@@ -12,6 +13,7 @@ export function InternalThumbnailImage({ file }: InternalThumbnailImageProps) {
   const { name, src } = file;
   const [imageSource, setImageSource] = useState<string>();
   const [imageLoaded, setImageLoaded] = useState<boolean>(false);
+  const [retryCount, setRetryCount] = useState<number>(0);
 
   useEffect(() => {
     src().then(url => setImageSource(url));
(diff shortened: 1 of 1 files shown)


scotttjob (changes requested):
A 404 from a web server indicates that the file could not be found. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404.

The PR description for this has not been filled in, so I'm not sure why we need to retry an arbitrary 4 times (if this change is correct & needed, I would like retry_count to be a prop with a default of 4).

In addition, we're retrying on any image error, not specifically a 404 error so the PR title and code don't match up.

If this is a situation where the file is being uploaded, and we want this component to patiently wait until the upload is complete before loading the URL, I would rather us modify this component to react to changes to the file prop instead of the solution presented in this PR. Or, we change how the promise on `src` works to only resolve when the file is actually present at the provided url.

tyler-santos:
> A 404 from a web server indicates that the file could not be found. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404.
> 
> The PR description for this has not been filled in, so I'm not sure why we need to retry an arbitrary 4 times (if this change is correct & needed, I would like retry_count to be a prop with a default of 4).
> 
> In addition, we're retrying on any image error, not specifically a 404 error so the PR title and code don't match up.
> 
> If this is a situation where the file is being uploaded, and we want this component to patiently wait until the upload is complete before loading the URL, I would rather us modify this component to react to changes to the file prop instead of the solution presented in this PR. Or, we change how the promise on `src` works to only resolve when the file is actually present at the provided url.

Hey @scotttjob  - my apologies I forgot that the atlantis PRs trigger an automatic review label, and I didn't remove it. This is not ready for review yet, it was something @darryltec and myself were working on together, and we wanted to get up into a PR so we could each continue to work off of it independently.

With that being said I appreciate the feedback, and more specifically this feedback:

>  I would rather us modify this component to react to changes to the file prop instead of the solution presented in this PR. Or, we change how the promise on src works to only resolve when the file is actually present at the provided url.

The file URL is not changing, which is why we can't react to changes in it, I would like that much better if we could. Changing how the promise on src works to only resolve when the file is actually present sounds like a potential path though that would be interesting to explore and definitley something for Darryl and myself to chat about. 

Thanks again for the feedback and sorry again for the lack of inidicator that this wasnt ready for review.

PR #1926, opened 2024-06-11
feat: side drawer improvements

darryltec:
### Added

- reverse scrolling prop
- animate back button going in and out as requested by our great designer
- a mechanism to register components

### Changed

- where the scroll is added because the reverse scrolling messes with shadows and sticky header/footer.
  - scroll is now on the content instead of the whole sidesheet
- share more CSS
- hookify slot ids

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


packages/components/src/SideDrawer/SideDrawer.tsx
```diff
@@ -155,6 +182,49 @@ export function SideDrawer({
     }
   }
 }
+const transition = {
+  duration: tokens["timing-base"] / 1000,
+  ease: "easeInOut",
+} as const;
+const BUTTON_WIDTH = 40;
```
scotttjob, line 189: Maybe make this a prop with a default?

packages/components/src/SideDrawer/SideDrawer.tsx
```diff
@@ -70,6 +80,11 @@ export function SideDrawer({
         toolbarPortal: ref?.querySelector(toolbar.selector),
         backPortal: ref?.querySelector(backButton.selector),
         footerPortal: ref?.querySelector(footer.selector),
+        components,
+        registerComponent: key =>
```
scotttjob, line 84: I'm approving to unstick, but I will admit this feels like a whole lot of code/system in order to set the internal component state to 'there is a back button present'. However I can see you've set this up to be extended easily when adding new components to the pattern, and the design of that system is on a solid foundation. 

Also, I'm not huge on patterns being setup when there's only one usable example so far (backButton). In your defense though, I imagine you do this sort of premature optimization because if you don't, you run the risk of showing up to the code in 6 weeks to find 3 new props after `backButtonPresent` for  things like `headerPresent` or `toobarPresentSoPleaseLoadMyFeatureSpecificChange` instead of a single line of `registerComponent('MyNewComponent')`. So long way of saying, it's a lot, but I understand why it's a lot.

packages/components/src/SideDrawer/SideDrawer.tsx
```diff
@@ -155,6 +182,49 @@ export function SideDrawer({
...
+} as const;
+const BUTTON_WIDTH = 40;
+
+function BackButtonContainer(props: HTMLAttributes<HTMLDivElement>) {
+  const { components } = useSideDrawerContext();
+  const [scope, animate] = useAnimate<HTMLDivElement>();
+
+  useEffect(() => {
+    if (components.backButton) {
+      animate(
```
scotttjob, line 197: Could we have built this with CSS Animations instead and applying classes? This feels like a lot of work for what might have been a ~5 line keyframe animation.

scotttjob (approved):
Left a few comments, nothing major, and nothing blocking.

LGTM!

packages/components/src/SideDrawer/SideDrawer.tsx
darryltec: actually, since i'm animating between known values, probs should be actually. lemme try that

packages/components/src/SideDrawer/SideDrawer.tsx
```diff
@@ -80,12 +95,14 @@ export function SideDrawer({
...
+
       <AnimatePresence initial={false}>
         {open && (
           <motion.div
-            className={styles.drawer}
+            className={classNames(styles.drawer, {
+              [styles.reverseScroll]: scrollDirection === "reverse",
+            })}
             ref={setRef}
-            data-elevation={"elevated"}
```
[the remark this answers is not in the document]
darryltec: oh shit, you right

packages/components/src/SideDrawer/SideDrawer.tsx
darryltec: yeah, it's getting there, i'm hoping this is the last of it with this compound pattern

packages/components/src/SideDrawer/SideDrawer.tsx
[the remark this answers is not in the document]
darryltec: tis back!

packages/components/src/SideDrawer/SideDrawer.css
```diff
@@ -124,3 +127,35 @@
...
+  }
+}
+
+.backButton {
+  --transition: var(--timing-base) ease-in-out;
+
+  width: 0;
+  opacity: 0;
+  transform: translateX(-8px);
+  transition:
```
scotttjob, line 151: didn't even need a keyframe! 👌

scotttjob (approved):
LGTM!

PR #1927, opened 2024-06-12
feat(components): Add testID prop to components Icon

taylorvnoj:
## Motivations

Allow a pass-through of a `testID` prop 

### Added

- new `testID` prop on `components` `Icon`
- tests for: using the `testID` prop & the situation where someone has used `name` as `data-testid` instead.

_Note:_ `name` was previously being used as a way to grab `data-testid` and there are two instances of that being used in jest tests in product. **_The same applies to `components-native` `Icon`_.** That's why I decided to use the OR operator and keep `name` as an option with `data-testid`.

#### Keeping `testID || name` 
In components where we have `suffix`, `prefix` or components like `Button` where we have `icon` as a prop:

```
 ...
  readonly icon?: IconNames;
```

We're bringing in `IconNames` and teams are using _that_ to test that an `Icon` is being used in these components as prefixes, etc.  So, only having `testID` here is not only more of a breaking change than I thought, I don't think we should expect consumers of `Button` to put a `testID` on a nested `Icon` that's being rendered as a prop.

## Testing

`testID` prop successfully being used here: https://github.com/GetJobber/Jobber/pull/47421

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Icon/Icon.test.tsx
+++ b/packages/components/src/Icon/Icon.test.tsx
@@ -36,3 +36,13 @@ it("renders star icon with custom color", () => {
   const { container } = render(<Icon name="star" customColor="#f33323" />);
   expect(container).toMatchSnapshot();
 });
+
+it("applies testID prop to svg element", () => {
+  const { getByTestId } = render(<Icon name="star" testID="star-icon" />);
+  expect(getByTestId("star-icon")).toBeInTheDocument();
+});
+
+it("applies name prop as testID when testID prop is not provided", () => {
+  const { getByTestId } = render(<Icon name="star" />);
+  expect(getByTestId("star")).toBeInTheDocument();
+});
--- a/packages/components/src/Icon/Icon.tsx
+++ b/packages/components/src/Icon/Icon.tsx
@@ -24,9 +24,20 @@ export interface IconProps {
    * Sets a custom color for the icon. Can be a rgb() or hex value.
    */
   readonly customColor?: string;
+
+  /**
+   * Used to locate this view in end-to-end tests
+   */
+  readonly testID?: string;
 }
 
-export function Icon({ name, color, customColor, size = "base" }: IconProps) {
+export function Icon({
+  name,
+  color,
+  customColor,
+  size = "base",
+  testID,
+}: IconProps) {
   let icon;
   const { svgClassNames, pathClassNames, paths, viewBox } = getIcon({
     name,
@@ -47,7 +58,7 @@ export function Icon({ name, color, customColor, size = "base" }: IconProps) {
       xmlns="http://www.w3.org/2000/svg"
       viewBox={viewBox}
       className={svgClassNames}
-      data-testid={name}
+      data-testid={testID || name}
     >
       {icon}
     </svg>


packages/components/src/Icon/Icon.tsx
```diff
@@ -47,7 +58,7 @@ export function Icon({ name, color, customColor, size = "base" }: IconProps) {
       xmlns="http://www.w3.org/2000/svg"
       viewBox={viewBox}
       className={svgClassNames}
-      data-testid={name}
+      data-testid={testID || name}
```
taylorvnoj, line 61: There are two external instances where we are using `name` in jest tests. I personally do not see the harm in continuing to support that - but I'd also be happy to remove `name` from this `data-testid` and update the tests in Jobber if we think that's best.

packages/components/src/Icon/Icon.test.tsx
```diff
@@ -36,3 +36,13 @@ it("renders star icon with custom color", () => {
   const { container } = render(<Icon name="star" customColor="#f33323" />);
   expect(container).toMatchSnapshot();
 });
+
+it("applies testID prop to svg element", () => {
+  const { getByTestId } = render(<Icon name="star" testID="star-icon" />);
+  expect(getByTestId("star-icon")).toBeInTheDocument();
+});
```
JeffMarvin, line 43: Consider also adding a test for the id `star` not being in the document here...

packages/components/src/Icon/Icon.tsx
JeffMarvin: Thinking long term and about trying to implement this prop predictably across all of our library components, I am inclined to fix those JO tests and not support the `name` property here also being the testID. We won't always have a `name` prop on other components, so it wouldn't be a "standard" behavior and might get confusing.

JeffMarvin (commented):
This looks good -- but I have one question regarding some internal discussion we've had around auto-naming these. In the document linked in the AC of the JIRA ticket for this, we thought we might use some kind of format like `ATL-${testID}-Icon`

Also, just noticed... and hopefully we have this as an Atlantis standard practice... should this be `testId`? (I prefer `testID` but I think we generally want `testId`)

scotttjob:
> This looks good -- but I have one question regarding some internal discussion we've had around auto-naming these. In the document linked in the AC of the JIRA ticket for this, we thought we might use some kind of format like `ATL-${testID}-Icon`
> 
> Also, just noticed... and hopefully we have this as an Atlantis standard practice... should this be `testId`? (I prefer `testID` but I think we generally want `testId`)

That's a good callout about wrapping the ID in some additional Atlantis specific context. Should we build a utility function that accepts the testId, and the component context (Icon in this case) and returns the id String? Could even build a `useTestId` hook that builds it automatically and manages the state a bit better so it doesn't regenerate on every render.

In terms of testID versus testId, I'm personally not concerned with either/or as long as we're consistent. So I'm happy with testID or testId or superCoolTestId (as long as they all have superCoolTestId)

packages/components/src/Icon/Icon.tsx
scotttjob: Agree with Jeff. If there ends up being a tiny bit of cleanup in JO for these changes, I'm okay with that. It's better for everything in the long-run to be consistent.

If it ever turns out to be a HUGE change though, we can talk about potentially pushing some of that work on teams (especially if we can accommodate in the short-term with a migration-friendly build for a while)

taylorvnoj: Perfect! There will be cleanup in mobile for that too (because the same thing is happening there for this component) but I can pre-emptively make sure I have a draft PR up making those changes so we have a fast follow after merging into Atlantis.

I'll make the change

taylorvnoj:
> should this be testId? (I prefer testID but I think we generally want testId)

10 Atlantis components currently have a `testID` prop already and they're all `components-native`. React Native's _out of the box_ `testID`'s are formatted that way, so I'm assuming we copied that.  For consistency we should keep them all the same with `testID`.

> In the document linked in the AC of the JIRA ticket for this, we thought we might use some kind of format like ATL-${testID}-Icon

@JeffMarvin @scotttjob I agree that the `ATL-` format should be used but imo, this changes the scope of all these tickets (which is completely fine).

My thinking:

Take `..src/DataList/components/DataListLoadingState/DataListLoadingState.tsx` in Atlantis:

```
...
export const DATALIST_LOADINGSTATE_ROW_TEST_ID =
  "ATL-DataList-LoadingState-Row";
...
...
return (
    <div className={styles.wrapper}>
      {loadingData.map((child, i) => (
        <div
          key={i}
          className={styles.loadingItem}
          data-testid={DATALIST_LOADINGSTATE_ROW_TEST_ID}
        >
          {layout(child)}
        </div>
      ))}
    </div>
  );
...
```

We're using `data-testid` to set an `ATL-` `testId` that is exported and used in product.  The same can be said for `ATL-Glimmer`, etc. 

I wholeheartedly agree with using the `ATL-` format, I think that changes the scope of this work.  Instead of adding a `testID` prop that the consumer can do whatever they want with, do we define & export a properly formatted `ATL-` testId on each Atlantis component?  It's a pattern that's already happening.

Let me know if I'm misunderstanding - might have to take it offline and have a conversation.  Might also have to double check that we could have a `data-testid` on each component...

packages/components/src/Icon/Icon.tsx
taylorvnoj: ACTUALLY 👀 I found a good reason to keep `data-testid={testID || name}`.

In components where we have `suffix`, `prefix` or components like `Button` where we have `icon` as a prop:

```
 ...
  readonly icon?: IconNames;
```

We're bringing in `IconNames` and teams are using _that_ to test that an `Icon` is being used in these components as prefixes, etc.  So, only having `testID` here is not only more of a breaking change than I thought, I don't think we should expect consumers of `Button` to put a `testID` on a nested `Icon` that's being rendered as a prop.

packages/components/src/Icon/Icon.test.tsx
taylorvnoj: Because of my comment [here](https://github.com/GetJobber/atlantis/pull/1927#discussion_r1647809883) we're going to be leaving `data-testid={testID || name}` as is, so I'll continue to test that `star` & `star-icon` can both be used as `testid`'s

taylorvnoj:
Talked to @scotttjob about the decision to keep `data-testid={testID || name}` and updated this PR description with details.  Putting back up for review 😄

scotttjob (approved):
Same thing as the mobile side! This component is harmonious. LGTM

PR #1931, opened 2024-06-14
fix: Add css to style href in link component

emilylim04:
## Motivations

- I noticed when migrating one of our team components to the new remix home, when I used the Link component it wasn't being styled properly. In jobber online we style our a tags but in the new home we can't rely on that anymore. 

## Changes

- Added css styling to a tag 

### Added

- Css styling to a tag in link component

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Link/Link.css
+++ b/packages/components/src/Link/Link.css
@@ -1,0 +1,4 @@
+.link {
+  color: var(--color-interactive);
+  text-decoration: underline;
+}
--- a/packages/components/src/Link/Link.tsx
+++ b/packages/components/src/Link/Link.tsx
@@ -1,5 +1,5 @@
 import React, { PropsWithChildren } from "react";
-import { Typography } from "../Typography";
+import styles from "./Link.css";
 
 export interface LinkProps {
   readonly url: string;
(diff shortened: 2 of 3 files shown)


emilylim04:
Sorry still in draft!

packages/components/src/Link/Link.css
```diff
@@ -0,0 +1,4 @@
+.link {
```
emilylim04, line 1: @darryltec I'm not sure if Atlantis follows a class name convention? Happy to update if this is completely wrong haha.

packages/components/src/Link/Link.tsx
```diff
@@ -20,6 +21,7 @@ export function Link({
       {...(external && { target: "_blank" })}
       {...(ariaLabel && { "aria-label": ariaLabel })}
       {...(ariaExpanded && { "aria-expanded": ariaExpanded })}
+      className={styles.link}
```
emilylim04, line 24: I went with the css style route vs using the typography component so I could add the underline. But if you think I should go with the typography component I can switch it up!

packages/components/src/Link/Link.css
darryltec: no strict rule on this one, link is a good addition

packages/components/src/Link/Link.tsx
darryltec: this works too!

darryltec (approved):
:shipit:

PR #1932, opened 2024-06-14
feat: Add button variation option to actions buttons in the Menu component (JOB-98306)

JobberNate:
## Motivations

Feature enhancement - Needed an option to style an action button in the menu component for delete options

### Added

- Action Buttons for the Menu component can be either styled destructively or default

### Changed

- added optional "buttonVariation" (default or destructive) parameter 

## Testing

Changes can be

---

Random photo of Atlantis


packages/components/src/Menu/Menu.tsx
```diff
@@ -264,20 +264,34 @@ export interface ActionProps {
    */
   readonly icon?: IconNames;
 
+  /**
+   * Visual style for the action button
+   */
+  readonly buttonVariation?: "default" | "destructive";
```
scotttjob, line 270: Can we update this to match our style in a few [other spots](https://github.com/GetJobber/atlantis/blob/17a49de3c78e2db92c4f775bc954b362ba8527ce/packages/components/src/DataList/DataList.types.ts#L342)? Just to be a boolean instead of a typed string.

packages/components/src/Menu/Menu.tsx
```diff
@@ -264,20 +264,34 @@ export interface ActionProps {
...
-function Action({ label, sectionLabel, icon, onClick }: ActionProps) {
+function Action({
+  label,
+  sectionLabel,
+  icon,
+  buttonVariation,
+  onClick,
+}: ActionProps) {
   const actionButtonRef = useRef() as RefObject<HTMLButtonElement>;
+  const buttonClasses = classnames(styles.action, {
```
scotttjob, line 286: Some of our icons have built in styles, such as trash defaulting to color-destructive (which is why the screenshot looks okay in the PR).

Because of the color defaulting on trash can we also set the [Icon Color manually](https://github.com/GetJobber/atlantis/blob/17a49de3c78e2db92c4f775bc954b362ba8527ce/packages/components/src/DataList/components/DataListAction/DataListAction.tsx#L26) when the `destructive` boolean is present, to ensure they line up in the case an Atlantis consumer chooses an icon other than `destructive`?

scotttjob (changes requested):
Thank you for the submission! I've requested two small changes, please let me know if you have any questions about either change request and I'd be happy to help.

Other than those two minor issues, this looks great. Thanks again!

packages/components/src/Menu/Menu.tsx
JobberNate: For sure! The original intention of the typed string was since we have other button variations - this would enable the door to be opened for that. If you feel this isn't needed, I can go ahead with the boolean!

packages/components/src/Menu/Menu.tsx
JobberNate: For sure! Great callout. I'll make that change!

scotttjob (approved):
LGTM! Thanks Nate!

PR #1933, opened 2024-06-14
feat: allow thumbnails on gallery

darryltec:
## Motivations

When using the gallery component, the thumbnail is the source image. It's all well and good until you pass in a large image that the browser needs to download. Furthermore, the user could be downloading a large image that they don't intend to view as they could've viewed it before and don't care about it.

This PR adds the ability to load a smaller thumbnail image on the gallery view so we keep the initial load small and thus speed up the wait that user would have to do

### Added

- allow files in gallery to pass in thumbnail urls

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Gallery/Web.stories.tsx
+++ b/docs/components/Gallery/Web.stories.tsx
@@ -22,7 +22,8 @@ const files = [
     type: "image/png",
     size: 213402324,
     progress: 1,
-    src: "https://picsum.photos/250",
+    thumbnailSrc: "https://picsum.photos/250",
+    src: "https://picsum.photos/550",
   },
   {
     key: "def",
@@ -30,7 +31,8 @@ const files = [
     type: "image/png",
     size: 124525234,
     progress: 1,
-    src: "https://picsum.photos/250",
+    thumbnailSrc: "https://picsum.photos/250",
+    src: "https://picsum.photos/550",
   },
   {
     key: "efg",
@@ -38,7 +40,8 @@ const files = [
     type: "image/png",
     size: 233411234,
     progress: 1,
-    src: "https://picsum.photos/250",
+    thumbnailSrc: "https://picsum.photos/250",
+    src: "https://picsum.photos/550",
   },
   {
     key: "jkl",
@@ -46,7 +49,8 @@ const files = [
     type: "image/png",
     size: 233411234,
     progress: 1,
-    src: "https://picsum.photos/250",
+    thumbnailSrc: "https://picsum.photos/250",
+    src: "https://picsum.photos/550",
   },
   {
     key: "mno",
@@ -54,7 +58,8 @@ const files = [
     type: "image/png",
     size: 233411234,
     progress: 1,
-    src: "https://picsum.photos/250",
+    thumbnailSrc: "https://picsum.photos/250",
+    src: "https://picsum.photos/550",
   },
   {
     key: "pqr",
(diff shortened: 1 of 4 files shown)


packages/components/src/Gallery/GalleryTypes.ts
```diff
@@ -35,6 +35,11 @@ export interface GalleryProps {
 
 export interface File
   extends Pick<FileUpload, "key" | "name" | "type" | "size" | "progress"> {
+  /**
+   * The thumbnail url of the file.
+   */
+  readonly thumbnailSrc?: string;
```
darryltec, line 41: now that i think about it, this should probs come from format file

darryltec: oh, nvm, it should be here as the format file isn't really gonna open the image and source there is the thumbnail size

docs/components/Gallery/Web.stories.tsx
```diff
@@ -22,71 +22,80 @@ const files = [
     type: "image/png",
     size: 213402324,
     progress: 1,
-    src: "https://picsum.photos/250",
+    thumbnailSrc: "https://picsum.photos/250",
```
tyler-santos, line 25: Strangly in story book, under canvas I see the correct behaviour where the thumbnail image and the actual image loaded dont match but when I go to the code section, I get the same photo for both despite the code saying its two seperate URL's. I dont think there is a problem other than it not reflecting what is really being used.

darryltec: the code tab is loading prod version of atlantis. we really need to throw a banner there 😆

tyler-santos: Ohhhh okay haha thanks I did not know this!

packages/components/src/Gallery/Gallery.test.tsx
```diff
@@ -167,3 +172,33 @@ describe("when a non-image is clicked", () => {
     });
   });
 });
+
+describe("Thumbnails", () => {
+  it.each(files.map(file => [file.name, file.thumbnailSrc]))(
+    "should display the thumbnailSrc as thumbnails for %s",
```
tyler-santos, line 178: Super small nitpick, I was confused when I read the test title, do you think this might be a little clearer? 

"should use the thumbnailSrc as the image source for %s",

packages/components/src/Gallery/Gallery.test.tsx
```diff
@@ -167,3 +172,33 @@ describe("when a non-image is clicked", () => {
...
+      const thumbnailImage = getByAltText(fileName);
+
+      await waitFor(() => {
+        expect(thumbnailImage).toHaveAttribute("src", src);
+      });
+    },
+  );
+
+  it.each(files.map(file => [file.name, file.src]))(
+    "should display the src as thumbnails for %s when thumbnailSrc is not provided",
```
tyler-santos, line 190: and then for this one: "should use the src as image source for %s when thumbnailSrc is not provided",

packages/components/src/Gallery/Gallery.test.tsx
darryltec: that makes sense

tyler-santos (approved):
Everything looks good to me - this should make a pretty big performance improvement. Thanks Darryl for doing this so quickly 🥳

PR #1935, opened 2024-06-17
fix(components): make `Select` options readable in dark mode

chris-at-jobber:
## Motivations

In Windows, when the color theme changes, the select `option` elements were not picking up changes in theme. This has the potential to render the options unreadable:
image

### Changed

- Added a more explicit selector to ensure the `option` elements pick up a semantically-themed background-color.

### Deprecated

- 

### Removed

- 

### Fixed

- Select is readable in other color modes (ie dark mode)
image

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Select/Select.css
+++ b/packages/components/src/Select/Select.css
@@ -23,6 +23,10 @@
   appearance: none;
 }
 
+.select option {
+  background-color: var(--field--background-color);
+}
+
 /**
  * Hides arrow in IE10 and up, the arrow is replaced by a custom font later on
  */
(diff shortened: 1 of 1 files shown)


scotttjob (approved):
Re-approving this.

Chris and I confirmed that this css does solve the problem at hand, but there's something wrong with his build process -> browserstack that's not letting us confirm with 100% certainty.

So we're going to roll this out anyway, and Chris will monitor the feedback. I'm going to take a separate track to see why Chris can't see this particular small CSS change in his browser stack environment.

PR #1940, opened 2024-06-18
fix: reduce gap on input suffixes

darryltec:
## Motivations

the gap between suffix and input shouldn't rely on the outer padding that changes depending on the input size

Before | After
:--:|:--:
| image | image |
| image | image |
| image | image |

### Added

- 

### Changed

- gap between the input and the suffix has been reduced and is static

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


packages/components/src/FormField/FormField.css
```diff
@@ -329,6 +329,7 @@
 
 .affixIcon.suffix {
   margin: 0 0 0 calc(var(--field--padding-right) * -1);
+  padding-left: var(--space-small);
```
chris-at-jobber, line 332: Is there a reason we _wouldn't_ do the same thing for the padding-right on the default `affixIcon`?

chris-at-jobber (approved):
Outside of the date selection input that triggered this discussion, I found only one other icon suffix usage that I could quickly verify.

## Before
image

## After
image

I did explore adding to the prefix and while I like the tighter grouping it gives the prefix with the icon and the label/value, it does create a bit of imbalance on the leading side of the input so I can see the case for leaving it as-is for prefix.

image

packages/components/src/FormField/FormField.css
chris-at-jobber: Played with this in product, feels like it needs more work to look good in the prefix context

PR #1943, opened 2024-06-21
feat(components-native): Add testID prop to components-native Icon

taylorvnoj:
## Motivations

Allow a pass-through of a testID prop

### Added

- new `testID` prop on `components-native` `Icon`
- tests for: using the `testID` prop & the situation where someone has used name as `data-testid` instead.

#### Keeping `testID || name` 
In `components-native` where we have `suffix`, `prefix` or components like `Button` where we have `icon` as a prop:

```
 ...
  readonly icon?: IconNames;
```

We're bringing in `IconNames` and teams are using _that_ to test that an `Icon` is being used in these components as prefixes, etc.  So, only having `testID` here is not only more of a breaking change than I originally thought. I don't think we should expect consumers of `Button` to put a `testID` on a nested `Icon` that's being rendered as a prop.

## Testing

Nothing fails in external mobile repo: https://github.com/GetJobber/jobber-mobile/pull/9365

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/Icon/Icon.test.tsx
+++ b/packages/components-native/src/Icon/Icon.test.tsx
@@ -37,3 +37,13 @@ it("renders quote icon with themed color", () => {
   const tree = render(<Icon name="quote" color="brand" />).toJSON();
   expect(tree).toMatchSnapshot();
 });
+
+it("applies testID prop to svg element", () => {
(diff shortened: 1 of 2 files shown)


scotttjob (approved):
LGTM!

PR #1946, opened 2024-06-22
refactor(components): move non-public component tokens off of `:root` [JOB-98605]

chris-at-jobber:
## Motivations

We've identified an unintentional fragility in exposing component-level variables on `:root`. Because they're not scoped to the component, they could be over-ridden accidentally by another component.

### Changed

- Component-level CSS variables are moved off of `:root` and onto the top-level parent of their respective components
- In cases where other styles are declared on the same classname, I have added a newline between variables and regular declarations for readability
- In cases where the top-level classname was not the same as the component name, I added a new classname and applied it at the top-level of the component for intuitive-ness(?)
- In cases where a variable is not used or used only once, I generally removed the variable and moved the definition to the source, excepting height/width definitions where it may be useful to reveal the intent of an otherwise "magic" number

### Removed

- The ability to over-ride these variables outside of the component

### Fixed

- FormatFile also now uses some semantic tokens so it's not broken in dark mode

image

## Testing

- Check that any removed variables are not called in Jobber (outside of the Atlantis `dist` styles)
- Interact with the components and see that they look the same as on production

Changes can be

---

Random photo of Atlantis


packages/components/src/Popover/Popover.css
```diff
@@ -1,9 +1,19 @@
-:root {
-  --card--base-padding: var(--space-base);
```
chris-at-jobber, line 2: this is not referenced anywhere in this file or in Jobber. There is a variable for this in Card, actually highlighting the fragility of these living on `:root`.

packages/components/src/Popover/Popover.css
```diff
@@ -1,9 +1,19 @@
...
+.popover {
   --popover--padding: var(--space-base);
+  --popover-content--base-padding: var(--space-base);
```
chris-at-jobber, line 3: This replaces the `--card-` variable

packages/components/src/ProgressBar/Sizes.css
```diff
@@ -1,11 +1,11 @@
 .smaller {
-  --progress-bar--height: var(--space-smaller);
+  --progress-bar--height: 4px;
```
chris-at-jobber, line 2: Just following our desired pattern of not defining height/width dimensions with `space` tokens because if we increased "spacing" we would not want progress bars to resize

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -1,10 +1,6 @@
-:root {
-  --file-card--height: calc(var(--space-base) * 3.5);
-  --file-card--background-color: var(--color-white);
-  --file-card--header-background-color: var(--color-greyBlue--lighter);
-  --file-card--content-background-color: var(--color-white);
+.formatFile {
+  --file-card--background-color: var(--color-surface);
```
chris-at-jobber, line 2: A few unused variables in here removed (and one that was used only once which doesn't actually merit using a var?)

packages/components/src/FormatFile/FormatFile.css
```diff
@@ -32,7 +28,7 @@
   padding: 0;
   border: none;
   text-align: left;
-  background-color: var(--color-transparent);
+  background-color: transparent;
```
chris-at-jobber, line 31: This token was doing nothing, I figure replacing it is more "safe" but also... maybe it doesn't actually need this declaration.

chris-at-jobber: Would rather follow up on that 🤷

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -48,7 +48,7 @@ export function FormatFile({
 }: FormatFileProps) {
   const isComplete = file.progress >= 1;
   const fileSize = getHumanReadableFileSize(file.size);
-  const wrapperClassNames = classnames(styles[display], {
+  const wrapperClassNames = classnames(styles[display], styles.formatFile, {
```
chris-at-jobber, line 51: created this new class as the others being applied at this layer are specific to certain variations of the component (size/presentation)

packages/components/src/Switch/Switch.tsx
```diff
@@ -39,7 +39,7 @@ export function Switch({
     }
   };
 
-  const className = classnames(styles.track, {
+  const className = classnames(styles.track, styles.switch, {
```
chris-at-jobber, line 42: technically I could have done the styling on `track` but I feel like defining the variables on a class named the same as the component name is a more intuitive pattern.

packages/components/src/Avatar/Avatar.css
```diff
@@ -1,9 +1,7 @@
-:root {
-  --avatar-size: calc(var(--base-unit) * 2);
+.avatar {
+  --avatar-size: 32px;
```
ZakaryH, line 2: I do see some overrides in product around this. be good to double check they still work as expected.

packages/components/src/Button/Button.css
```diff
@@ -1,16 +1,12 @@
-:root {
+.button {
```
ZakaryH, line 1: also see some overrides in product for `--button--` variables`

probably fine, but worth verifying quick

packages/components/src/Card/Card.css
```diff
@@ -1,10 +1,8 @@
-:root {
+.card {
   --card--accent-color: var(--color-grey);
   --card--base-padding: var(--space-base);
   --card--border: var(--border-base) solid var(--color-border);
-}
```
ZakaryH, line 5: also for cards. might be the case that once we see they continue to work for one we can reasonably assume the rest will be fine.

packages/components/src/Popover/Popover.css
ZakaryH: that's very strange

I wonder why we were using card variables in popover when it definitely doesn't contain cards by default. gonna take a look at the PR where we added that. maybe cards were doing something weird in Popovers?

packages/components/src/Button/Button.css
chris-at-jobber: Interesting, this one seems to have the most examples so I might kick this one out into its' own PR. A lot of the usages are people seemingly attempting to rebuild an Atlantis button from scratch, which is theoretically as simple as dropping in the semantic tokens they actually, but there are a few explicit overrides that appear to be happening.

chris-at-jobber: OK changed the usages [in Jobber](https://github.com/GetJobber/Jobber/pull/48119) - this will need a tweak to the one remaining usage of these public values to consume from `--public--` but I think that's safe to do on the import of these changes?

packages/components/src/Avatar/Avatar.css
```diff
@@ -1,9 +1,11 @@
 :root {
-  --avatar-size: calc(var(--base-unit) * 2);
-  --avatar-border-size: var(--border-base);
+  --public--avatar-size: 32px;
```
chris-at-jobber, line 2: Will need to update this usage when it comes in to Jobber

packages/components/src/Drawer/Drawer.css
```diff
@@ -1,9 +1,7 @@
-:root {
+.container {
```
chris-at-jobber, line 1: No usages in Jobber

packages/components/src/Card/Card.css
chris-at-jobber: PR to [remove Card usages](https://github.com/GetJobber/Jobber/pull/48156)

packages/components/src/FormatFile/FormatFile.css
chris-at-jobber: No usages in Jobber

packages/components/src/Popover/Popover.css
chris-at-jobber: only usages in Jobber are in a copy/paste, unaffected by these changes

packages/components/src/ProgressBar/ProgressBar.css
```diff
@@ -1,5 +1,5 @@
-:root {
-  --progress-bar--height: var(--space-base);
+.ProgressBar {
+  --progress-bar--height: 16px;
```
chris-at-jobber, line 2: No usages in Jobber

packages/components/src/SideDrawer/SideDrawer.css
```diff
@@ -1,9 +1,7 @@
-:root {
+.drawer {
   --sideDrawer--base-padding: var(--space-base);
```
chris-at-jobber, line 2: No usages in Jobber

packages/components/src/StatusIndicator/StatusIndicator.css
```diff
@@ -1,8 +1,6 @@
-:root {
+.statusIndicator {
   --statusLabel-icon-diameter: 8px;
```
chris-at-jobber, line 2: No usages in Jobber

packages/components/src/StatusLabel/StatusLabel.css
```diff
@@ -1,5 +1,14 @@
-:root {
+.statusLabelRow {
   --labelBackgroundColor: var(--color-success--surface);
```
chris-at-jobber, line 2: No usages in Jobber

packages/components/src/Switch/Switch.css
```diff
@@ -1,4 +1,4 @@
-:root {
+.switch {
```
chris-at-jobber, line 1: No usages in Jobber

packages/components/src/Tabs/Tabs.css
```diff
@@ -1,10 +1,11 @@
 :root {
-  --tab--height: calc(var(--space-larger) + var(--space-small));
-  --tab--inset: var(--space-base);
   --public-tab--inset: var(--tab--inset);
 }
 
 .tabs {
+  --tab--height: 40px;
```
chris-at-jobber, line 6: No usages in Jobber

packages/components/src/Tooltip/Tooltip.css
```diff
@@ -1,14 +1,12 @@
...
-  --tooltip--offset: calc(-1 * var(--space-smaller));
-  --tooltip--surface: var(--color-surface--reverse);
-}
-
 .shadowActivator {
   display: none;
 }
 
 .tooltipWrapper {
+  --tooltip--arrow-size: 8px;
```
chris-at-jobber, line 6: No usages in Jobber

ZakaryH:
@chris-at-jobber another conflict

also we'll need the reverted Storybook version that's in master now in order to be able to import this and try it out in product

in the meantime I'll test it out in Atlantis

packages/components/src/Button/Button.css
```diff
@@ -1,16 +1,15 @@
 :root {
-  --button--color-variation: var(--color-interactive);
-  --button--color-variation--hover: var(--color-interactive--hover);
+  --public--button--color-variation: var(--color-interactive);
+  --public--button--color-variation--hover: var(--color-interactive--hover);
```
ZakaryH, line 3: I think we're missing a step here if I'm not mistaken.

if we only changing the variable name from `--button--color-variation` to `--public--button--color-variation` but we don't update the places that are consuming it via `var(--button--color-variation)` then the public version isn't being used.

if I throw a random button into a page somewhere I see this

image

where it's getting those variables defined by

```
/* Variation */

.qkFQwI0PKfw- {
  --button--color-variation: var(--color-interactive);
  --button--color-variation--hover: var(--color-interactive--hover);
}
```

which is the unique-ified version of this rule

```
/* Variation */

.work {
  --button--color-variation: var(--color-interactive);
  --button--color-variation--hover: var(--color-interactive--hover);
}
```

chris-at-jobber: Oh of course, very good catch

packages/components/src/FormatFile/FormatFile.css
ZakaryH: I see below we replaced `var(--file-card--border-color)` with `var(--color-border)`

that appears to have been the only instance of `--file-card--border-color`

if that's how we'd prefer to define the border color, then we can get rid of the variable definition altogether since it's not being used!

packages/components/src/FormatFile/FormatFile.css
ZakaryH: I think it was _somehow_ getting the correct value???

you're right, that var doesn't exist
image

and yet, the computed value is....

image

which is the exact same value we get with your fix!?
image

I don't know if there's some string matching magic going on where it saw the word "transparent" and ran with it

somehow these changes are equivalent. this is just a much more correct and understandable way of achieving it.

packages/components/src/Tabs/Tabs.css
```diff
@@ -1,10 +1,11 @@
 :root {
-  --tab--height: calc(var(--space-larger) + var(--space-small));
-  --tab--inset: var(--space-base);
   --public-tab--inset: var(--tab--inset);
```
ZakaryH, line 2: I'm not too familiar with css variables when it comes to using them before they are defined

is this going to have a value via some sort of hoisting? or I wonder if it's going to say "I don't know who that is" since we define `--tab--inset` below this line. gonna take a peek.

ZakaryH: ah, yeah it does **not** do any kind of hoisting and so these end up being undefined and we lose the padding right/left. it's tough to see but if you inspect the DOM the highlighting will show ya

image

image

VS the old way

image

packages/components/src/Popover/Popover.css
```diff
@@ -20,20 +30,8 @@
...
-  border: var(--border-base) solid var(--color-border);
-  border-radius: var(--radius-base);
-  font-size: var(--typography--fontSize-base);
-  line-height: normal;
-  background: var(--color-surface);
-}
-
 .popover > * {
-  --public-content--padding: var(--card--base-padding);
+  --public-content--padding: var(--popover-content--base-padding);
```
ZakaryH, line 34: this is a strange css rule. I think I get what it's doing but there's gotta be a more elegant way to do this.

don't need to do it in this PR. just wanted to mention is because it feels like something that can be improved.

packages/components/src/Tabs/Tabs.css
chris-at-jobber: https://github.com/GetJobber/atlantis/pull/1946/commits/043e747ae904ffd6064b987a261bccd68eb16d56

ZakaryH: this is an interesting case. the quickest fix is to move the `:root` definition down below `.tabs` though that is kind of gross and breaks the pattern set by every other `:root` definition's placement in the css file

ZakaryH: nice. ok I think that's a good fix.

I don't quite understand why we want to allow the vertical padding and horizontal padding to diverge but it's already enabling that today so 🤷

ZakaryH (approved):
LGTM. great stuff!

there's a couple things I wanna double check in product with this PR, but I'd rather do that when we import it and swap over those variables to use the new `--public--` format anyway

packages/components/src/Tabs/Tabs.css
chris-at-jobber: It's because that public inset is allowing common layout / container components to manipulate whether the tabs's children sit flush or indented to the left/right, which is decoupled from whether the content itself should be flush.

Arguably the tab component should have less of an opinion about that altogether, but I think this is was an early "convenience" solve that's pretty embedded in product usage by this point.

chris-at-jobber:
@ZakaryH Thanks! [This one](https://github.com/GetJobber/Jobber/pull/48119) should probably have a review before we bring this one in so I'm going to hold off on merging until then

PR #1948, opened 2024-06-25
feat(components): SSR support for toast

MichaelParadis:
## Motivations

During our SSR support updates we missed adding support for toasts. This PR fixes that by preventing references to `document` if it is undefined

### Added

- Added SSR support for toasts

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing
Verify Toasts work in the referenced PRs

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Toast/Toast.test.tsx
+++ b/packages/components/src/Toast/Toast.test.tsx
@@ -22,7 +22,7 @@ beforeEach(() => {
 });
 
 afterEach(() => {
-  jest.runOnlyPendingTimers();
+  act(() => jest.runOnlyPendingTimers());
   jest.useRealTimers();
 });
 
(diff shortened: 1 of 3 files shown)


packages/components/src/Toast/Toast.test.tsx
```diff
@@ -96,15 +103,15 @@ it("stops and starts the timer when the item is hover toggled", async () => {
 
   fireEvent.mouseEnter(getByText("Bland Toast"));
 
-  await act(() => {
+  act(() => {
```
MichaelParadis, line 106: These awaits were not actually needed.

darryltec:
Code looks good. Mind testing it in remix where you render it in the server. I wonder if that shows up on the client after mount or not at all until you fire it in the client again

MichaelParadis:
@darryltec I have a referenced PR. I had to fire it again on the client since I wasn't able to find a way to modify the document while hydrating (I could have easily missed something)

scotttjob:
> should also export this in here `packages/components/src/index.tsx` so you can import from `@jobber/components` directlly

Only if this has been tested elsewhere to make sure it doesn't break existing implementations. That's why it was left out to begin with.

MichaelParadis:
@scotttjob I assume it was left out initially because toast would break on SSR usages. However since this fixes that it should be good. I will do another pre-release and verify

scotttjob:
> @scotttjob I assume it was left out initially because toast would break on SSR usages. However since this fixes that it should be good. I will do another pre-release and verify

That's correct. However when I fixed the SSR issues in a last minute effort when I realized they were broken, it caused issues in other repos. So just make sure it's gravy everywhere and we should be good to go.

MichaelParadis:
>That's correct. However when I fixed the SSR issues in a last minute effort when I realized they were broken, it caused issues in other repos. So just make sure it's gravy everywhere and we should be good to go.

I was able to verify that this worked in the other places 😉

MichaelParadis:
@darryltec updated in 4b12bdf465f6d1e8abc0ef645c19dc71b2b9ad27

darryltec (approved):
:shipit:

PR #1950, opened 2024-06-26
fix(components): Update FormField to avoid useId StrictMode problem

jdeichert:
## Motivations

Eric found that React's built-in `useId` is not stable across renders when within a `StrictMode` tree (only affects local development). I confirmed this using his [branch](https://github.com/GetJobber/atlantis/compare/master...ec/rhf-stable-id-fix) for testing.

**Why is this a problem?**

Because we rely on `useId()` to generate a unique stable identifier when the parent component chooses not to pass a `name` prop to FormField. This id/name is passed to react-hook-form's controller and if it's not stable between renders, this breaks the form's validation logic.

**Why is this a _new_ problem?**

A recent [PR](https://github.com/GetJobber/atlantis/pull/1942/files) merged that updated `FormField` to use `useController` from `react-hook-form` instead of `<Controller>`. `useController` is more flexible and allows pulling out/storing state where `<Controller>` can't.

This change revealed this problem in one of our forms because:
* The inputs didn't have a `name` prop being passed to them, so they generated ids internally
* It's within a `StrictMode` tree 
* The old `<Controller>` component seems to have internal caching outside of React's state so the old way didn't run into this.

Eric also came up with a solution that works great! See [comment](https://github.com/GetJobber/atlantis/pull/1950#discussion_r1655452592) for more info.

### Fixed

- `FormField` problems in `StrictMode` when `name` prop isn't supplied

## Testing

I tried to add a unit test for this but unfortunately there isn't a way to validate the expectation. RHF's controller logic isn't exposed by `FormField` so we can't easily inspect that internal state. I did attempt to verify the `name` prop doesn't change in `StrictMode`... but that doesn't work as both cases still result in the same final generated id... the only difference is that RHF's internal state is now correct when it wasn't before.

To manually verify this:

1. Edit `docs/components/InputText/Web.stories.tsx` and paste the below script
2. Add a log to `FormField` around line 68: `console.log('RHF', control._fields);`
3. Visit http://localhost:6005/?path=/story/components-forms-and-inputs-inputtext-web--multiple
4. Observe that there are only 2 field names: `firstName` and `generatedName--:r3:`
    * If there are 3 names (2 being generated) this is the bug!

```tsx
export const Multiple = () => {
  return (
    <StrictMode>
      <Form>
        <InputText
          name="firstName"
          validations={{
            required: "Required",
          }}
          placeholder="First Name"
        />
        <InputText
          placeholder="Last Name"
          validations={{
            required: "Required",
          }}
        />

        <button type="submit">Submit</button>
      </Form>
    </StrictMode>
  );
};
```

### DIFF
--- a/packages/components/src/FormField/FormField.tsx
+++ b/packages/components/src/FormField/FormField.tsx
@@ -14,8 +14,21 @@ import styles from "./FormField.css";
 import { FormFieldWrapper } from "./FormFieldWrapper";
 import { FormFieldPostFix } from "./FormFieldPostFix";
 
-// eslint-disable-next-line max-statements
 export function FormField(props: FormFieldProps) {
+  // Warning: do not move useId into FormFieldInternal. This must be here to avoid
+  // a problem where useId isn't stable across multiple StrictMode renders.
+  // https://github.com/facebook/react/issues/27103
+  const id = useId();
+
+  return <FormFieldInternal {...props} id={id} />;
+}
+
+type FormFieldInternalProps = FormFieldProps & {
+  readonly id: string;
+};
+
+// eslint-disable-next-line max-statements
+function FormFieldInternal(props: FormFieldInternalProps) {
   const {
     actionsRef,
     autocomplete = true,
@@ -23,6 +36,7 @@ export function FormField(props: FormFieldProps) {
     defaultValue,
     description,
     disabled,
+    id,
     inputRef,
     inline,
     keyboard,
@@ -52,16 +66,13 @@ export function FormField(props: FormFieldProps) {
       : // If there isn't a Form Context being provided, get a form for this field.
         useForm({ mode: "onTouched" });
 
-  const [identifier] = useState(useId());
-  const [descriptionIdentifier] = useState(`descriptionUUID--${useId()}`);
+  const descriptionIdentifier = `descriptionUUID--${id}`;
   /**
    * Generate a name if one is not supplied, this is the name
    * that will be used for react-hook-form and not neccessarily
    * attached to the DOM
    */
-  const [controlledName] = useState(
-    name ? name : `generatedName--${identifier}`,
-  );
+  const [controlledName] = useState(name ? name : `generatedName--${id}`);
 
   useEffect(() => {
     if (value != undefined) {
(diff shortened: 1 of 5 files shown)


packages/components/src/FormField/FormField.tsx
```diff
@@ -52,16 +66,13 @@ export function FormField(props: FormFieldProps) {
...
+  const descriptionIdentifier = `descriptionUUID--${id}`;
   /**
    * Generate a name if one is not supplied, this is the name
    * that will be used for react-hook-form and not neccessarily
    * attached to the DOM
    */
-  const [controlledName] = useState(
-    name ? name : `generatedName--${identifier}`,
-  );
+  const [controlledName] = useState(name ? name : `generatedName--${id}`);
```
ericchernuka, line 75: Another use of state we can remove. This can just be derived. No need to persist in state.

packages/components/src/FormField/FormField.tsx
```diff
@@ -95,7 +105,7 @@ export function FormField(props: FormFieldProps) {
 
   const fieldProps = {
     ...rest,
-    id: identifier,
+    id,
     className: styles.input,
     name: (validations || name) && controllerName,
```
ericchernuka, line 110: Only a nit, we seem to have `name`, `controlledName`, and `controllerName`. Just wondering if simplifying all these names would make it easier. I don't even think we need to extract `controllerName` from `useController` since we're spreading `...rest` onto `fieldProps`. So there's a few optimizations, but we may just want to rename `name: nameProp` in the destructuring, then we can use:
```
const name = nameProp ?? `generatedName--${id}`
```

Then we can replace this here with:
```suggestion
    name: (validations || nameProp) && name,
```

ericchernuka: Fair enough. Seemed like only 3 places so I figured it was pretty lo-risk since it keeps most of the logic contained rather than referring to different values as it goes down the file which actually makes it more bug prone.

ericchernuka (commented):
Looks great!

I haven't had a chance to run it, but overall the code looks great.

taylorvnoj (approved):
The pre-release in product does result in 2 failed tests (https://github.com/GetJobber/Jobber/pull/47985). Seems like we're assigning a `:r7:` id, similar to the `Autocomplete` snap you commented on earlier. 

I'm approving because everything looks great via testing the code - but we should have a plan for what we can do with those tests before we merge

PR #1952, opened 2024-07-05
fix(components): Loosen React Dependency Version Requirement

ZakaryH:
## Motivations

I'm having trouble installing latest `master` in Jobber. it's yelling about React version dependency resolution. I suspect it's because we have `"react": "18.2.0"` which is quite specific and when it doesn't get exactly that, it can't resolve.

## Changes

"18.2.0" -> "^18.2.0"

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -121,7 +121,7 @@ You will be able to select the platform you want to create the component.
 ? Component Name: ExampleComponent
 ? Generate for:
 > Web
-> React native
+> React Native
 > Both
 
 ✔  +! 5 files added
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
     "postcss-preset-env": "^8.3.0",
     "prettier": "^2.4.1",
     "prop-types": "^15.7.2",
-    "react": "^18.2.0",
+    "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-ga4": "^2.1.0",
     "react-hook-form": "^7.30.0",
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
     "postcss-preset-env": "^8.3.0",
     "prettier": "^2.4.1",
     "prop-types": "^15.7.2",
-    "react": "18.2.0",
+    "react": "^18.2.0",
     "react-dom": "18.2.0",
     "react-ga4": "^2.1.0",
     "react-hook-form": "^7.30.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -104,7 +104,7 @@
         "postcss-preset-env": "^8.3.0",
         "prettier": "^2.4.1",
         "prop-types": "^15.7.2",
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -129,7 +129,7 @@
     "postcss-preset-env": "^8.3.0",
     "prettier": "^2.4.1",
     "prop-types": "^15.7.2",
-    "react": "18.2.0",
+    "react": "^18.2.0",
```
ZakaryH, line 132: this one might be fine? it's in dev dependencies whereas the others are in peer dependencies which is what I think the problem is. pretty sure you're allowed to be as specific as you want with your own dev dependencies.

ZakaryH:
noticing a few things that aren't quite right. I'll double check that they weren't caused by this, and if not probably make a ticket or two for 'em

**after clicking to open the Mobile/Basic ContentOverlay I get an error**
image

**InputNumber/Mobile/Basic can't type or clear it**
**InputSearch/Mobile/Basic can't clear**

ZakaryH:
this ain't right. however it's totally unrelated to this change. seems like our flex style is causing the first child to be full width despite the fact that it's a span and should be inline block

image

package.json
ZakaryH: undoing it because I don't have a good reason to change it

ZakaryH: on 2nd/3rd thought..... looking at all the historic values, we've always opted to use `^` as a prefix here

my same logic of sticking to what we were doing before without a good reason to change it might instead suggest that we do go with `^18.2.0` here. I could be convinced otherwise but as far as I know we don't have a need for this to be so specific.

ZakaryH:
prerelease is failing because package.json changes are ignored and so it's breaking on the thing that this PR is trying to fix

we'll need to make a commit in README or some `src` tracked file for it to trigger and pick up the change

docs/README.md
```diff
@@ -121,7 +121,7 @@ You will be able to select the platform you want to create the component.
 ? Component Name: ExampleComponent
 ? Generate for:
 > Web
-> React native
+> React Native
```
ZakaryH, line 124: since it's the name of the product we should treat it as a proper noun.

I hope my English teacher is proud of me.

MichaelParadis:
@ZakaryH RE: the content overlay. That mobile version does mention that web might be different (it looks like it isn't displaying now that we don't show the Docs page in the new storybook version because it isn't rendering on the canvas
https://github.com/GetJobber/atlantis/blob/JOB-81443/loosen-react-dependency-req/docs/components/ContentOverlay/Mobile.stories.tsx

MichaelParadis (approved):
LGTM! I was able to verify that installing in JO and JM worked with these changes

PR #1953, opened 2024-07-05
fix(components): Fix Rollup Config File Names

ZakaryH:
## Motivations

types were getting very confused when being consumed outside this project

specifically `Card` `Modal` and `ProgressBar`

types derived from styles in this way
`Card`
```
import colors from "./cardcolors.css";

  readonly accent?: keyof typeof colors;
```

would not correctly find a type that had keys of actual color values. instead it would seemingly fallback to a definition that says any css import is a string, and therefore `keyof string` would give string prototype values like `match` or `indexOf`

image

Glimmer wasn't experiencing this problem and uses the same pattern of `keyof typeof size` so I went to see what was different. found the rollup config and noticed the file names in there for Card, Modal and ProgressBar didn't match their actual file names. maybe we updated them at one point and forgot to update the rollup value.

fixing the names results in the built bundle/package/whatever having the correct types
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

updated the rollup config to use the correct names

### Security

- 

## Testing

you'll need to build & import this branch into another project. preferably `jobber` since it has tons of invocations of these components.

you'll also need my other PR to do this
https://github.com/GetJobber/atlantis/pull/1952

otherwise it will yell at you about dependencies

once that's been installed, everything should just be happy. instances of `Card` `Modal` and `ProgressBar` should NOT be upset about incorrect values being given to them from existing implementations! if you want, you can inspect the type definitions and yeah they should be the expected values and not `String` methods/keys

image

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/rollup.config.mjs
+++ b/packages/components/rollup.config.mjs
@@ -64,12 +64,15 @@ export default {
     }),
     copy({
       targets: [
-        { src: "src/Card/colors.css.d.ts", dest: "dist/Card" },
+        { src: "src/Card/cardcolors.css.d.ts", dest: "dist/Card" },
         { src: "src/Content/Spacing.css.d.ts", dest: "dist/Content" },
         { src: "src/Gallery/Gallery.css.d.ts", dest: "dist/Gallery" },
         { src: "src/Grid/GridAlign.css.d.ts", dest: "dist/Grid" },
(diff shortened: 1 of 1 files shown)


MichaelParadis (approved):
LGTM!

PR #1957, opened 2024-07-12
fix: Storybookv7 update broken links

MichaelParadis:
## Motivations

Due to use removing the `Docs` tab we had to update the titles for the `Docs` page to remove the trailing `/Docs` and update links to those pages.

One thing to note is that some of our pages now have stories attached to them which adds them to the Sidebar. See screenshot below. The stories were added to the Canvases because we had some extra code snippets needed for them but in Storybook 7 you could no longer do this so we used a `<Story/>` to work around this. In the future we are revamping storybook and the goal will be to cleanup the organization after the version bump.

image

### Added

- 

### Changed

- Updated links to the storybook docs pages to account for the removed `Docs` tab

### Deprecated

- 

### Removed

- 

### Fixed

- fixed ColorSwatches on the `Colors` section in Storybook

### Security

- 

## Testing

 - Verify that the `Docs` sections render correctly.
 - Verify that links on docs pages take you to the correct location

Changes can be

---

image


taylorvnoj:
In `docs/README.md` we have broken links in:
- `Design System` section linking to Components/Design/Hooks
- `Tooling and configuration` section linking to EsLint & StyleLint configuration pages
- `Publishing a failed release to NPM` there's a link to the changelog that's broken (although that's currently broken on master, still going to the wrong place)

In `docs/CONTRIBUTING.md` there's a broken link:
- line 199 to `Card` is broken
ACTUALLY - the URL seems to be correct but locally, clicking the link won't work for me? Can you double check that locally?

MichaelParadis:
@taylorvnoj sure thing. I think I missed the Md files and only updated the MDX files based on your comments. I will update it shortly

docs/CONTRIBUTING.md
```diff
@@ -128,9 +128,9 @@ component once your proposal is approved.
 ### Implement a proposal
 
 All components should start with a proposal. One of our
-[Guiding Principles](#guiding-principles) is to start all components with
-documentation. Once your proposal has been approved, proceed to contribute code
-to Atlantis to make it a reality!
+[Guiding Principles](https://atlantis.getjobber.com/?path=/docs/contributing--docs#guiding-principles)
```
MichaelParadis, line 131: This link wasn't working from my testing. I updated this to point to the location on the atlantis domain because it was only opening in the iframe or not scrolling when I changed it to match the others

MichaelParadis:
@taylorvnoj RE: Code sandbox, I have found that links to the code tab from our storybook are flakey (even on the main branch). My gut makes me think it is related to this: https://github.com/GetJobber/atlantis/blob/07da251e539be4a8d1c15785ecb6abaaf2a39624/.storybook/manager.js#L32-L56

I can update them to point to the canvas so it works and they can switch the tab manually

taylorvnoj:
> @taylorvnoj RE: Code sandbox, I have found that links to the code tab from our storybook are flakey (even on the main branch). My gut makes me think it is related to this:

Ahhh ok - I'm good with that!

docs/CONTRIBUTING.md
taylorvnoj: hmmm I'm still seeing some weirdness with the link. I'm ok with just removing this link since it's referring to something directly below, if we can't get it working as intended.
https://github.com/user-attachments/assets/55944442-459a-4b97-a8c6-e4b0c19d2675

MichaelParadis: Sure thing, I noticed that our Table of contents doesn't use default browser behavior (it does a javascript scroll). I will update it right now because I am pretty sure it is a storybook router issue

MichaelParadis: Removed in 464ff26e

taylorvnoj: The `InputDate` story has a 'see related components' link that works 🤔

MichaelParadis: It is an link to another page where as Guiding Principles is a link to the same page but a different section

MichaelParadis: ohh I see let me see. My assumption is mdx vs MD

taylorvnoj: Same thing for the link to `InputText` on `InputPassword` story - but this should be able to be like all the other links?

https://github.com/user-attachments/assets/872a856c-abb2-42b3-9971-d506d7fc8411

MichaelParadis: So looking at it the internal links on component pages have a `sb-docs` class but the contributing.md link doesn't 🤔 
The InputPassword is a fun one (it links to the live atlantis docs page directly instead of locally 😭  )

taylorvnoj (changes requested):
These last few and then we're good - oh and also on the [Hooks page](http://localhost:6005/?path=/docs/packages-hooks--docs) the link to `useResizeObserver` isn't working.

taylorvnoj (approved):
Thank you for all those fixes! SHIP IT!

PR #1962, opened 2024-07-18
docs(components): update Box and Card component docs

chris-at-jobber:
## Motivations

Box is a relatively new component but has some overlap with Card, so clearing up when to use each is helpful.

It will also give Triton something a bit meatier to train on so it can be more robust in answering this question!
image

### Added

- "Related components" to Box and Card

### Changed

- How we present the alpha state of Box

## Testing

Read the docs in the build!

---

Random photo of Atlantis


darryltec (approved):
:shipit:

PR #1966, opened 2024-07-24
feat(components): SegmentedControl

scotttjob:
## Motivations

There are a few one off versions of `SegmentedControl` in product, so we should promote it to Atlantis!

Kapture 2024-10-30 at 22 58 55

### Added

- New `SegmentedControl` component
- Docs page and component examples

Checklist for AC:
- [ ] options are customizable children, however options are intended to be text or icons
- [ ] SegmentedControl can accept up to 5 options (any more than that and the focus will not work - this should be documented in docs section)
- [ ] SegmentedControl is keyboard- and screenreader-operable
- tab key to enter and exit the component
- arrow right and left to switch options
- screen reader reads out the `ariaLabel`, which option is currently selected & indicates that the options are related to each other (3 of 5)
- [ ] All options take up the same width
- [ ] Options should reflow/wrap internally in a fixed width parent (not advised to wrap text - this is communicated in the docs)
Screenshot 2024-10-30 at 10 56 15 PM

---

Random photo of Atlantis


packages/components/src/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,47 @@
...
+  useEffect(() => {
+    const length = Children.count(children);
+
+    if (container.current) {
+      container.current.style.setProperty(
+        "--segmentedControl--option-count",
+        String(length),
+      );
+    }
+  }, [children]);
```
ericchernuka, line 32: Children are technically always considered unstable so you may just want to remove either the dep array or just count the length and make it the dep array value.

packages/components/src/SegmentedControl/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,47 @@
...
+  checked,
+  children,
+  label,
+  onChange,
+}: SegmentedControlOptionProps) {
+  const { selectedOption, handleChange } = useSegmentedControl();
+  const localChecked = useMemo(() => {
+    return checked ?? selectedOption === value;
+  }, [selectedOption, checked]);
+  const inputId = `${value.toString()}_${useId()}`;
```
ericchernuka, line 28: Should be safe only needing useId

packages/components/src/SegmentedControl/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * The value of the option.
+   */
+  readonly value: string | number;
+  /**
+   * The label to appear beside the radio button.
+   */
+  readonly checked?: boolean;
+  readonly label?: string;
+  readonly onChange?: (valueIn: React.ChangeEvent<HTMLInputElement>) => void;
```
ericchernuka, line 14: Nit: argument name is actually the native event so it's a bit misleading. Do we want to expose the native event or should we just have a `onSelected` callback that fires.

packages/components/src/SegmentedControl/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+interface SegmentedControlOptionProps extends PropsWithChildren {
+  /**
+   * The value of the option.
+   */
+  readonly value: string | number;
+  /**
+   * The label to appear beside the radio button.
+   */
+  readonly checked?: boolean;
```
ericchernuka, line 12: This seems like an odd behaviour. Seems like checked should always come from the provider rather than enabling a situation where the top level value can reflect a different checked state than the option.

packages/components/src/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,47 @@
...
+  defaultOption,
+  children,
+}: SegmentedControlProps<T>) {
+  const container = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    const length = Children.count(children);
+
+    if (container.current) {
+      container.current.style.setProperty(
+        "--segmentedControl--option-count",
```
ericchernuka, line 28: You could avoid the useEffect all together by leveraging data attributes to reflect the count.

```html
<div data-option-count={React.Children.count(children).length}
```

```css
.container {
  --segmentedCpntrol--option-count: attr(option-count, 1)
}
```

packages/components/src/SegmentedControl/SegmentedControlProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  return { handleChange, selectedOption, setSelectedOption };
+};
+
+export function SegmentedControlProvider<T>({
+  children,
+  onSelectOption,
+  defaultOption,
+}: SegmentedControlProviderProps<T>) {
+  const [selectedOption, setSelectedOption] = useState<T>(defaultOption);
```
ericchernuka, line 40: Anyway we can change this to be controlled and uncontrolled? Even lifting the concept from headless libraries. https://github.com/radix-ui/primitives/blob/main/packages/react/use-controllable-state/src/useControllableState.tsx

packages/components/src/SegmentedControl/SegmentedControlProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+    setSelectedOption(event.target.value as T);
+    onSelectOption(event.target.value as T);
+  };
+
+  return (
+    <SegmentedControlContext.Provider
+      value={{
+        handleChange,
+        selectedOption,
+        setSelectedOption: () => {
```
ericchernuka, line 52: Should be wrapped in a useCallback

packages/components/src/SegmentedControl/SegmentedControlProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export function SegmentedControlProvider<T>({
+  children,
+  onSelectOption,
+  defaultOption,
+}: SegmentedControlProviderProps<T>) {
+  const [selectedOption, setSelectedOption] = useState<T>(defaultOption);
+
+  const handleChange: ChangeEventHandler<HTMLInputElement> = event => {
```
ericchernuka, line 42: Should be wrapped in a useCallback for stability.

packages/components/src/SegmentedControl/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { PropsWithChildren, useId, useMemo } from "react";
+import { useSegmentedControl } from "./SegmentedControlProvider";
+
+interface SegmentedControlOptionProps extends PropsWithChildren {
```
ericchernuka, line 4: Ability to pass a name for the input would be a nice perk so it could be leveraged inside forms but also saved as value of the form.

packages/components/src/SegmentedControl/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * The value of the option.
+   */
+  readonly value: string | number;
```
ericchernuka, line 8: Should probably be a generic so types could be inferred by the ones passed in rather than needing to coerce it.

packages/components/src/SegmentedControl/SegmentedControlOption.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <>
+      <input
+        type="radio"
+        id={inputId}
+        checked={localChecked}
+        value={value}
+        onChange={handleChangeLocal}
+      />
+      <label htmlFor={inputId}>{children || label}</label>
```
ericchernuka, line 39: If this is an icon only we'll need devs to add visually hidden text to add accessibility. Conversely we could say children can be whatever but an optional `accessibilityLabel` prop can be used when it's just an icon as a fast track to setting an `aria-label.`

scotttjob: Appreciate all the feedback Eric! The PR is very much a draft, and I communicated internally about all the work this would still need to go through before it was ready. I wanted to get a V0.01 over the fence.

I'll be sure that whomever picks up this torch makes sure your feedback is addressed. :)

packages/components/src/SegmentedControl/SegmentedControl.module.css
```diff
@@ -0,0 +1,62 @@
...
+
+.container input[type="radio"] {
+  position: absolute;
+  left: -999vw;
+  cursor: pointer;
+}
+
+.container label {
+  z-index: 1;
+  padding: var(--space-smaller) 12px;
```
taylorvnoj, line 20: @chris-at-jobber I see in figma the padding is `--space-smaller` 
 `12px`; let me know how this feels to you.

packages/components/src/SegmentedControl/SegmentedControl.module.css
```diff
@@ -0,0 +1,62 @@
...
+  cursor: pointer;
+  align-items: center;
+}
+
+.container span {
+  position: absolute;
+  left: 0;
+  width: calc(100% / var(--segmentedControl--option-count));
+  height: 100%;
+  box-shadow: var(--shadow-low);
```
chris-at-jobber, line 34: This is my bad but can we leave out the border -> shadow change for now? I was playing around a bit with the Triton one as a "safe sandbox" example but I think leveraging shadows in our UI elements is a change that I need to look at more systemically vs introducing here in isolation.
```suggestion
  border: var(--color-border--interactive);
```

packages/components/src/SegmentedControl/SegmentedControl.module.css
```diff
@@ -0,0 +1,62 @@
...
+  background-color: var(--color-surface);
+  transition: var(--timing-base) left ease;
+}
+
+.container input[type="radio"]:focus-visible ~ span {
+  box-shadow: var(--shadow-focus);
+}
+
+.container input[type="radio"]:nth-of-type(1):checked ~ span {
+  left: 0;
```
chris-at-jobber, line 45: This is a huge nitpick but setting this to 0 leaves us with a "double border" at the left edge of the component, so can we use

```suggestion
  left: -1px;
```

here? _(OR if there's some other modification we should make to box-sizing, etc that would get us the same result, ultimately we just want to avoid the double-border)_

**Current**
image

**What we want**
image

packages/components/src/SegmentedControl/SegmentedControl.module.css
chris-at-jobber: I think there's something iffy about the lineheight (in Figma vs code) because in Figma, the output is a 32px high element (which matches the height of a `small` Button):

image

Thinking more about the vertical padding... I'm noticing that even some of the example usages we see SegmentedControl placed in line with both small and base Buttons, so maybe we need to leverage similar sizes? 

Something like `smallest` padding, but a min-height for a `small` and `base` SegmentedControl that lines up with Button? (32px small, 40 base, etc) 

image
image
image11

Or if there's something we can do CSS-wise that defaults to SegmentedControl taking up the same height as its' container, but maybe that's too generic and/or prone to breaking 🤔

chris-at-jobber: (Configuring the size can be a follow-up if need be, just realizing that we see different sizes in use already)

packages/components/src/SegmentedControl/SegmentedControl.module.css
taylorvnoj: Not your bad! Updated 👍🏻

packages/components/src/SegmentedControl/SegmentedControl.module.css
```diff
@@ -3,6 +3,8 @@
   grid-template-columns: repeat(var(--segmentedControl--option-count), 1fr);
   align-items: center;
   position: relative;
+  height: var(--space-larger);
```
chris-at-jobber, line 6:
```suggestion
  min-height: 32px;
```

We (semi-)recently aligned on [moving away from using spacing tokens for sizing](https://atlantis.getjobber.com/?path=/docs/guides-frontend-styleguide--docs#size-width-and-height) - you can reference Button as an example of what that looks like in practice:

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Button/Button.module.css#L11

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Button/Button.module.css#L184

packages/components/src/SegmentedControl/SegmentedControlOption.tsx
taylorvnoj: We could, but I'm leaning toward keeping the `value` for better readability (`pizza_:r0:` vs. `:r0:`). I've had to dig into test outputs before that only referenced the `useId` dozens of times and it was _horrible_ to find what I wanted 😅  So for readability, debugging and uniqueness, I'd prefer to keep `value`.

packages/components/src/SegmentedControl/SegmentedControl.tsx
ericchernuka: This way will also avoid the issue if someone dynamically changes the contents. Rare, but shouldn't assume its not going to occur.

ericchernuka: Forgot I had a better solution below. https://github.com/GetJobber/atlantis/pull/1966#discussion_r1690678615

packages/components/src/SegmentedControl/SegmentedControl.tsx
taylorvnoj: ok with that approach, first I was getting `Property 'length' does not exist on type 'number'`.
That also that broke (I think) `grid-template-columns` because the options then stacked vertically.

Screenshot 2024-10-23 at 2 31 00 PM

What do you think about setting the variable in `style`. It works and gets rid of the `useEffect`

```
<div 
        style={
          {
            "--segmentedControl--option-count": Children.count(children),
          } as CSSProperties
        }
```

packages/components/src/SegmentedControl/SegmentedControlOption.tsx
taylorvnoj: Removed the `checked` prop here bfac31d85bc847ce0fc080890cc9818bea9f5838
Seems like selection state is already managed using `selectedOption` from `SegmentedControlContext`

packages/components/src/SegmentedControl/SegmentedControl.tsx
ericchernuka: Oh, that's purely due to my dumb brain forgetting that `.count(children)` returns you the number.

The attr check was then failing because the attribute lookup didn't contain anything and fell back to 1?

packages/components/src/SegmentedControl/SegmentedControlProvider.tsx
taylorvnoj: Something like this?

```
 const resetToDefaultOption = useCallback(() => {
    setSelectedOption(defaultOption);
  }, [defaultOption]);

  return (
    <SegmentedControlContext.Provider
      value={{
        handleChange,
        selectedOption,
        setSelectedOption: resetToDefaultOption,
      }}
    >
 ```

Comment (ericchernuka) [reply packages/components/src/SegmentedControl/SegmentedControlProvider.tsx:52]: Ya, but I wonder why we even need this method?

File: packages/components/src/SegmentedControl/SegmentedControl.tsx (hunk 5 above)
Comment (taylorvnoj) [reply packages/components/src/SegmentedControl/SegmentedControl.tsx:39]: Yeah that's my understanding.  I _think_ that setting this directly in the style prop is still ok and still helping us avoid re renders?

File: packages/components/src/SegmentedControl/SegmentedControl.module.css (hunk 12 above)
Comment (taylorvnoj) [reply packages/components/src/SegmentedControl/SegmentedControl.module.css]: I've added a subtask as a fast follow to add an optional size prop for `base` and `small` - really shouldn't take much of a lift and I can branch off this branch while it's being reviewed 👍🏻

File: packages/components/src/SegmentedControl/SegmentedControl.module.css (hunk 14 above)
Comment (taylorvnoj) [reply packages/components/src/SegmentedControl/SegmentedControl.module.css:49]: I love a _nit by chris_ ™️ 
Used `box-sizing` to fix this - setting `-1px` actually cut off some of the left border

Screenshot 2024-10-23 at 5 20 01 PM

File: packages/components/src/SegmentedControl/SegmentedControl.module.css (hunk 15 above)
Comment (taylorvnoj) [reply packages/components/src/SegmentedControl/SegmentedControl.module.css]: Shoulda remembered this! Done here fdc3bd61a2c24f8cb3185930cedcd94fade6fd6e

File: packages/components/src/SegmentedControl/SegmentedControl.module.css (hunk 14 above)
Comment (chris-at-jobber) [reply packages/components/src/SegmentedControl/SegmentedControl.module.css:49]: Oh nice 👌

File: packages/components/src/SegmentedControl/SegmentedControlProvider.tsx (hunk 8 above)
Comment (taylorvnoj) [reply packages/components/src/SegmentedControl/SegmentedControlProvider.tsx]: For now, `handleChange` and `setSelectedOption` are wrapped in `useCallback`

File: packages/components/src/SegmentedControl/SegmentedControlOption.tsx (hunk 9 above)
Comment (taylorvnoj) [reply packages/components/src/SegmentedControl/SegmentedControlOption.tsx:4]: Considering this as a future TODO as discussed, but will not be done in this PR

File: packages/components/src/SegmentedControl/SegmentedControlOption.tsx (hunk 11 above)
Comment (taylorvnoj) [reply packages/components/src/SegmentedControl/SegmentedControlOption.tsx]: Good call out - `aria-label` prop added here 82b04af7bb42b8c1841d789323edbbf2e5350c97

Comment (ericchernuka) [reply packages/components/src/SegmentedControl/SegmentedControlOption.tsx]: Might suggest limiting it to `children`. `label` would be redundant

File: docs/components/SegmentedControl/Web.stories.tsx [hunk 16]
```diff
@@ -0,0 +1,88 @@
...
+    >
+      {options.map(option => (
+        <SegmentedControl.Option key={option.value} value={option.value}>
+          {option.label}
+        </SegmentedControl.Option>
+      ))}
+    </SegmentedControl>
+  );
+};
+
```
taylorvnoj, line 43: I also want to note that a more type safe implementation can be written, if that is preferred. 

```
type ControlledOptions = "pizza" | "pizza2" | "pizza3" | "pizza4";

export const Controlled: Story = () => {
  const [activeOption, setActiveOption] = useState<ControlledOptions>("pizza");

  return (
    <SegmentedControl<ControlledOptions>
      selectedValue={activeOption}
      onSelectValue={setActiveOption}
    >
      <SegmentedControl.Option<ControlledOptions> value="pizza">
        Pizza
      </SegmentedControl.Option>
      <SegmentedControl.Option<ControlledOptions> value="pizza2">
        Ultra Pizza
      </SegmentedControl.Option>
      <SegmentedControl.Option<ControlledOptions> value="pizza3">
        Mega Pizza
      </SegmentedControl.Option>
      <SegmentedControl.Option<ControlledOptions> value="pizza4">
        Giga Pizza
      </SegmentedControl.Option>
    </SegmentedControl>
  );
};
```

`SegmentedControl` and `SegmentedControlOption` can be used with the generic type parameter (`ControlledOptions`) which ensures type safety and restricts the `value` to a specific set of options.

docs/components/SegmentedControl/SegmentedControl.stories.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+Use a
+[controlled](../?path=/story/components-selections-segmentedcontrol-web--basic-controlled)
+SegmentedControl when you need to manage the state of the selected option from a
+parent component. This is useful when the selected option needs to be
+synchronized with other state or when you need to perform side effects when the
+selected option changes.
+
+Use an
+[uncontrolled](../?path=/story/components-selections-segmentedcontrol-web--basic-uncontrolled)
```
Aiden-Brine, line 79: Nit: I think this is the more default, simpler, out of the box option and usually those are listed first in the docs so I would move this before controlled. I think as you read through the docs you should encounter more options for customization

packages/components/src/SegmentedControl/SegmentedControl.tsx
```diff
@@ -0,0 +1,54 @@
...
+   */
+  readonly defaultValue?: T;
+
+  /**
+   * A unique name for the SegmentedControl, that links the group of
+   * options together.
+   *
+   * @default useId()
+   */
+  readonly name?: string;
```
taylorvnoj, line 28: @Aiden-Brine I played around with having the `name` here or in the option, but decided to have `name` in the parent and pass it down through the provider. This centralizes management of the `name` prop and the user doesn't have to set `name` on each option, which can be a headache/redundant/leave room for error if `name` isn't consistent between options, like a typo or something.

docs/components/SegmentedControl/SegmentedControl.stories.mdx
taylorvnoj: good point - done here 3326df62e26c0e05d2154d358770f80412824a33

docs/components/SegmentedControl/SegmentedControl.stories.mdx
```diff
@@ -0,0 +1,111 @@
...
+version when the SegmentedControl can manage it's own state. This is useful when
+the selected option does not need to be synchronized with other state or when
+you need a simple implementation.
```
Aiden-Brine: This is a good explanation of when to use a controlled vs uncontrolled segmented control. It doesn't really explain how to actually make it controlled vs uncontrolled. If I was unfamiliar with the concept I don't think I would know how to go about it. Also, looking at `SegmentedControlProvider` `isControlled` is set to true if both `selectedValue` and `onSelectedValue` are set. So what happens if I only set one? If this leads to weird or unexpected behaviour I think it should be documented. 

Sometimes we can rely on the prop descriptions but looking here:
Screenshot 2024-11-13 at 11 09 20 AM
If I didn't know anything about the component I wouldn't know how these three props connect to controlled vs uncontrolled

docs/components/SegmentedControl/Web.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  { value: "tacos", label: "Tacos" },
+  { value: "sushi", label: "Sushi" },
+  { value: "burgers", label: "Burgers" },
+];
+
+export const Controlled: Story = () => {
+  const [activeOption, setActiveOption] = useState("tacos");
+
+  return (
+    <SegmentedControl
```
Aiden-Brine, line 31: If I don't look in the underlying code and only look on Storybook these two examples are the same right? Can we modify them so that the controlled and uncontrolled behaviour is clear while only looking in Storybook?

packages/components/src/SegmentedControl/SegmentedControl.tsx
Aiden-Brine: I agree that this is a better location for it 👍

packages/components/src/SegmentedControl/SegmentedControl.tsx
```diff
@@ -18,21 +18,32 @@ interface SegmentedControlProps<T> extends PropsWithChildren {
...
 
 export function SegmentedControl<T>({
   selectedValue,
   onSelectValue,
   defaultValue,
   children,
+  name = useId(),
 }: SegmentedControlProps<T>) {
   const container = useRef<HTMLDivElement>(null);
+  console.log("name", name);
```
Aiden-Brine, line 39: There are a couple of `console.log` calls to remove 😄

Aiden-Brine (changes requested):

docs/components/SegmentedControl/SegmentedControl.stories.mdx
```diff
@@ -0,0 +1,111 @@
...
+using [Menu](/?path=/docs/components-navigation-menu--docs) as an alternative
+single-selection method for modifying the view. Otherwise, truncation will
+prevent the user from reading the label and a label with wrapped text will
+create layout inconsistencies.
+
+## Controlled and Uncontrolled
+
+Use an
+[uncontrolled](../?path=/story/components-selections-segmentedcontrol-web--basic-uncontrolled)
+version when the SegmentedControl can manage it's own state. This is useful when
```
chris-at-jobber, line 73: Given my surface-level understanding, this appears to be explaining the concepts of controlled and uncontrolled components, not how they relate to this component. Is there anything specific to how SegmentedControl can be used in either fashion that users would need to know?

docs/components/SegmentedControl/SegmentedControl.stories.mdx
```diff
@@ -0,0 +1,111 @@
...
+If you do not have enough horizontal room to allow for readable labels, consider
+using [Menu](/?path=/docs/components-navigation-menu--docs) as an alternative
+single-selection method for modifying the view. Otherwise, truncation will
+prevent the user from reading the label and a label with wrapped text will
+create layout inconsistencies.
+
+## Controlled and Uncontrolled
+
+Use a
+[controlled](../?path=/story/components-selections-segmentedcontrol-web--basic-controlled)
```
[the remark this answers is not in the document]
taylorvnoj: Ouf good catch, I renamed the stories. Done here 018798faf05fb6b3f43218b89714477df86be4cc

packages/components/src/SegmentedControl/SegmentedControl.tsx
taylorvnoj: classic! removed them here 022a54257396fbcbe86a1e4793737bab2085f7cb

docs/components/SegmentedControl/SegmentedControl.stories.mdx
taylorvnoj: Great points!
I added more info on the Docs page about how to implement controlled vs uncontrolled. I created code examples and described what props to use for each variation. Updated here 2caca5d47ba6fec9e469e27187d5f5e5e97d48bf
I hear what you're saying about explaining the unexpected behaviour if you deviate from that, however I prefer to demonstrate/explain the happy path, rather than add more documentation on what will go wrong if you don't follow the documentation. If you feel strongly about that lets chat. I personally feel like it may confuse the reader but I'm open for more conversation on that!

The prop descriptions were leaving a lot to be desired 😆  I added more detail here 39d330bc2bf949ff8cb99a2a404044c46f640cfb

docs/components/SegmentedControl/SegmentedControl.stories.mdx
taylorvnoj: Great question - similar to Aiden's above where he pointed out I wasn't explaining how to implement controlled vs uncontrolled for this component specifically. I updated that doc section here 2caca5d47ba6fec9e469e27187d5f5e5e97d48bf

chris-at-jobber (dismissed):
Small visual tweak - the selected "segment" white background should have a border as well ([Figma for reference](https://www.figma.com/design/rIIhulZvcp9M82lNOCGv16/Product%2FOnline?m=auto&node-id=21472-363&t=b8A9mVv2AWprfTub-1))

**Current**
image

**Figma**
image

taylorvnoj:
@chris-at-jobber after some struggling (double border being brought back, it looking horrific back in product lol, strange behaviour in Storybook, etc) I got a great tip from Becca to use outline and we're looking much better

Screenshot 2024-11-14 at 10 59 41 AM

Screenshot 2024-11-14 at 10 59 48 AM

Aiden-Brine (approved):
Thanks for making those changes and great job with this!

chris-at-jobber (commented):
Design approved!

PR #1967, opened 2024-07-25
fix(components): Mock popper to prevent act warnings in test suites

jdeichert:
## Motivations

Popper internally calls `forceUpdate` which leads to a re-render outside of React's normal lifecycle. This causes testing-library to rightfully throw act warnings in various test suites.

After digging into this, the common recommendation is to mock popper in tests.

NOTE: The `Tooltip` test suite has certain expectations around popper behaviour so I did not mock it there.

### Fixed

* Mock popper in various test suites
* Modernized test suites with userEvent and screen queries

## Before / After

Before this change, when running this Combobox test case, there is an act warning in the console. After, there are no act warnings.

Screenshot 2024-07-25 at 11 27 25 AM

Screenshot 2024-07-25 at 11 27 17 AM

## Testing

1. Run the combobox test suite and observe zero act warnings in the console.
2. Disable the mock at the top of that suite and you'll see 1 act warning.

Feel free to verify the other suites I updated all contain zero act warnings.

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.test.tsx
+++ b/packages/components/src/Tooltip/Tooltip.test.tsx
@@ -3,6 +3,8 @@ import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { Tooltip } from ".";
 
+// We automatically mock popper (__mocks__/@popperjs/core.js), but in this test suite
+// we are verifying certain behaviours that require popper to be active.
 jest.unmock("@popperjs/core");
 
 it("shouldn't show the tooltip", async () => {


packages/components/src/Autocomplete/Autocomplete.test.tsx
```diff
@@ -4,6 +4,15 @@ import userEvent from "@testing-library/user-event";
...
 
+// Mock popper to avoid forceUpdate causing act warnings with testing-library.
+jest.mock("@popperjs/core", () => ({
+  createPopper: () => ({
+    destroy: jest.fn(),
+    forceUpdate: jest.fn(),
+    update: jest.fn(),
+  }),
+}));
+
```
JeffMarvin, line 15: Just curious if you chose this style over Manual Mocks -- i.e. placing a mock for this module in a global `__mocks__` folder, a la https://jestjs.io/docs/manual-mocks

JeffMarvin: Give it a try. We don't do it commonly and I'm no Jest expert, so I can't tell if there are pitfalls... but it seems to be in line with what you're trying to achieve in this PR.

JeffMarvin (changes requested):
Recommended exploring the manual mock approach for the module.

packages/components/src/Tooltip/Tooltip.test.tsx
```diff
@@ -3,6 +3,8 @@ import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { Tooltip } from ".";
 
+jest.unmock("@popperjs/core");
```
JeffMarvin, line 6: 🎉 
Since the library mocks can be a bit hidden, consider adding a comment here to explain why we're unmocking the popper library.

JeffMarvin (approved):
Nice stuff. All of the tests changed here are passing locally, and (compared to master) do not trigger the act warnings.

PR #1971, opened 2024-07-30
fix(components): Change label FomField label to have full width

nad182:
## Motivations

The bug with label not taking full text area width was originally reported [here](https://getjobber.slack.com/archives/C03BJHV3PMG/p1720545457085239).

### Changed

- CSS of textarea:
  - Gave label the width of 100%
  - Changed the paddings of input to accommodate label
  - This causes the scroll bar to appear differently: it starts the top of the input (underneath the label), instead of from the very top of the `FormField` (which included the label).

### WAS
image

### IS
image

Changes can be

---

Random photo of Atlantis


ZakaryH:
if we are going the full width with no padding, there's one scenario where things will look different

image

- placeholder is unreasonably long
- the userAgent is **not** reported back as Safari and so we don't get the extra margin

now our placeholder content goes to the very edge of the input rather than having some spacing on the right. I do think this is very unlikely to happen.

whereas Safari reporting userAgents will handle this fine due to the extra margin right
image

ZakaryH:
ah another thing to consider is that with a padding on the top, we won't always have a label. so without a label, and that padding top we go from
Kapture 2024-07-30 at 11 00 38

to

Kapture 2024-07-30 at 11 01 17

where the text that scrolls upwards now disappears into an empty white space compared to it feeling like it went into/above the textarea

also the scrollbar doesn't go the full height. that one is pretty minor IMO. I don't know how many people would notice that, plus there's some other strange behavior with scrollbars in this code to begin with

nad182:
> ah another thing to consider is that with a padding on the top, we won't always have a label. so without a label, and that padding top we go from Kapture 2024-07-30 at 11 00 38 [ Kapture 2024-07-30 at 11 00 38 ](https://private-user-images.githubusercontent.com/11843215/353565959-f6c5806f-9574-4d76-a528-f9dd15489a5a.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjIzNjMwMTEsIm5iZiI6MTcyMjM2MjcxMSwicGF0aCI6Ii8xMTg0MzIxNS8zNTM1NjU5NTktZjZjNTgwNmYtOTU3NC00ZDc2LWE1MjgtZjlkZDE1NDg5YTVhLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA3MzAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNzMwVDE4MDUxMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWYxMDcyMDQ1NmRiZjNlMTU1NGQ3ZjY5MWQyODE0ODIzMTU3Y2JiYmE2OTViYjc3YWI3YmZhMjY5MWNkNDNkNzEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.u51z3QFVwIKqvWI4T_YBL4uuvZLH0CM0i_ppsJHoAaA) [ ](https://private-user-images.githubusercontent.com/11843215/353565959-f6c5806f-9574-4d76-a528-f9dd15489a5a.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjIzNjMwMTEsIm5iZiI6MTcyMjM2MjcxMSwicGF0aCI6Ii8xMTg0MzIxNS8zNTM1NjU5NTktZjZjNTgwNmYtOTU3NC00ZDc2LWE1MjgtZjlkZDE1NDg5YTVhLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA3MzAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNzMwVDE4MDUxMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWYxMDcyMDQ1NmRiZjNlMTU1NGQ3ZjY5MWQyODE0ODIzMTU3Y2JiYmE2OTViYjc3YWI3YmZhMjY5MWNkNDNkNzEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.u51z3QFVwIKqvWI4T_YBL4uuvZLH0CM0i_ppsJHoAaA)
> 
> to
> 
> Kapture 2024-07-30 at 11 01 17 [ Kapture 2024-07-30 at 11 01 17 ](https://private-user-images.githubusercontent.com/11843215/353566126-27af9474-0756-49fb-bc9d-349879dc7174.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjIzNjMwMTEsIm5iZiI6MTcyMjM2MjcxMSwicGF0aCI6Ii8xMTg0MzIxNS8zNTM1NjYxMjYtMjdhZjk0NzQtMDc1Ni00OWZiLWJjOWQtMzQ5ODc5ZGM3MTc0LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA3MzAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNzMwVDE4MDUxMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQwZmE4MTk0NGZmOWM3OWE5YTVlNTMzMTk5MDRiY2Y3ZjkyODI0OTAxZDEzNWYxZjc5OGVmMWJlODliYjlhNzEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.L62_8w3mV9AflJP8txuc1yJKHZVHcgmCcrQ2J2Y9J-c) [ ](https://private-user-images.githubusercontent.com/11843215/353566126-27af9474-0756-49fb-bc9d-349879dc7174.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjIzNjMwMTEsIm5iZiI6MTcyMjM2MjcxMSwicGF0aCI6Ii8xMTg0MzIxNS8zNTM1NjYxMjYtMjdhZjk0NzQtMDc1Ni00OWZiLWJjOWQtMzQ5ODc5ZGM3MTc0LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA3MzAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNzMwVDE4MDUxMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQwZmE4MTk0NGZmOWM3OWE5YTVlNTMzMTk5MDRiY2Y3ZjkyODI0OTAxZDEzNWYxZjc5OGVmMWJlODliYjlhNzEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.L62_8w3mV9AflJP8txuc1yJKHZVHcgmCcrQ2J2Y9J-c)
> 
> where the text that scrolls upwards now disappears into an empty white space compared to it feeling like it went into/above the textarea
> 
> also the scrollbar doesn't go the full height. that one is pretty minor IMO. I don't know how many people would notice that, plus there's some other strange behavior with scrollbars in this code to begin with

That one should be relatively straightforward to fix (I think). I can just make it so that the padding is added ONLY if there is a label. That should fix the scrollbar starting from the top as well.

ZakaryH:
for context I remember what all the regex/Safari stuff was about now

when you use an iphone/apple device in Chrome's devtools for a mobile view - it replicates the userAgent that would be reported. on apple devices, it gets identified (correctly) as Safari since I believe that's the underlying engine or whatever that influences how the browser will behave

that's how you get these letters appearing in a bad place
image

that's because it's using margin and not padding in this scenario, which is intentional to leave room to interact with the draggable handle on Safari. other browsers handle the zIndex in a nice way where you can still interact with the handle despite it technically being "behind" another element. in order for Safari to access it, we've gotta add that margin.

image


but since it's margin, we're basically already accounting for the space it takes up. then the width - padding value is applied, which isn't needed in this specific case and it removes padding again, leaving that space for characters to show up

your fix will address this, we'd just need to sort out those couple extra scenarios. alternatively, if we wanted to try a different approach we could make that width - padding rule conditional, and not apply it when we're on a Safari instance. that might end up being easier? I haven't fully thought through that case, but based on my understanding I could see that being all that's required.

nad182:
@ZakaryH , I pushed the fix to add the padding on the right, since the label is occupying the full width with my change in the previous commit. This should fix the case when the label is overly long.
I've also added some logic to add and remove top paddings to the input (and the wrapper) based on whether the label is provided (present) or not. Please, let me know what you think.

PS. This also fixes the scrollbar to start from the top in case there is no label.

packages/components/src/FormField/FormField.css
```diff
@@ -215,10 +215,16 @@
   flex-direction: column;
 }
 
+/* If there's a label, add padding-top to wrapper to account for "absolute" positioned label */
+.inputWrapper:has(.label) .childrenWrapper {
```
ZakaryH, line 219: so we're removing the padding-top from `.input` and then re-adding it to `.childrenWrapper` I'd prefer we try to put the padding back on the same element if possible. makes things easier to follow and track down imo

image
image

vs how it works on current master

image

packages/components/src/FormField/FormField.css
```diff
@@ -215,10 +215,16 @@
   flex-direction: column;
 }
 
+/* If there's a label, add padding-top to wrapper to account for "absolute" positioned label */
```
ZakaryH, line 218: this is largely a personal preference so I could be convinced otherwise but I usually want a fairly compelling reason to add comments

I'd say the code communicates the first half of this comment - if there's a label, add padding

so it's really only the "why" around the absolutely positioned label that the comment adds. I guess my question would be, did you find that part unintuitive or hard to understand to the point where you think a comment explaining the padding would save time or effort?

packages/components/src/FormField/FormField.css
```diff
@@ -232,6 +238,13 @@
   -webkit-box-shadow: 0 0 0 30px var(--color-surface) inset !important;
 }
 
+/* If there's no label, add back padding-top to input */
```
ZakaryH, line 241: personally I think the code already does a solid job of communicating this. `:not(:has(.label` and then the corresponding rule is adding the padding

is there something you think the comment adds or better explains?

ZakaryH:
ah, we're starting to get into the fun of FormField

the "basic" or non `multiline` variant of `InputText` doesn't work so well with the `childWrapper` padding

image

vs `master`

image


it's because the input (not textarea) has its own padding that handles things AND it has a label, so now it's double dipping on padding 😭 

you're welcome to keep going with this approach but I do want to mention another possible approach 

>your fix will address this, we'd just need to sort out those couple extra scenarios. alternatively, if we wanted to try a different approach we could make that width - padding rule conditional, and not apply it when we're on a Safari instance. that might end up being easier? I haven't fully thought through that case, but based on my understanding I could see that being all that's required.

it wouldn't address the scrollbar height stuff, but I think this would solve the initial bug in a way that involves less changes . I'm pretty sure anyway, admittedly I haven't tried it out!

nad182:
> ah, we're starting to get into the fun of FormField
> 
> the "basic" or non `multiline` variant of `InputText` doesn't work so well with the `childWrapper` padding
> 
> image
> vs `master`
> 
> image
> it's because the input (not textarea) has its own padding that handles things AND it has a label, so now it's double dipping on padding 😭
> 
> you're welcome to keep going with this approach but I do want to mention another possible approach
> 
> > your fix will address this, we'd just need to sort out those couple extra scenarios. alternatively, if we wanted to try a different approach we could make that width - padding rule conditional, and not apply it when we're on a Safari instance. that might end up being easier? I haven't fully thought through that case, but based on my understanding I could see that being all that's required.
> 
> it wouldn't address the scrollbar height stuff, but I think this would solve the initial bug in a way that involves less changes . I'm pretty sure anyway, admittedly I haven't tried it out!

@ZakaryH , I've tried adding `:not(.safari)` selector and it works (in Safar), but it doesn't work in other browsers in mobile view (Samsung selected in the inspector):
image

Also, just like you said, the scrollbar is being hidden behind the label, which was the bug @MichaelParadis [fixed](https://github.com/GetJobber/atlantis/pull/1216) 14 moths ago. So I'm worried that a simple change for label to have a `width: 100%` and `padding-right: 0`, will just reintroduce that bug.

packages/components/src/FormField/FormField.css
```diff
@@ -233,6 +233,8 @@
 }
 
 .textarea.safari .input {
+  margin-top: var(--field--padding-top);
```
ZakaryH, line 236: any reason you went with margin?

it's doing that thing again where as you scroll the text feels like it disappears into the white block of margin rather than going out of view of the textarea

if we use padding, that should fix it

nad182: I guess we might have been looking at different stories/variations, since I didn't see that originally. But I changed back to using padding instead of margin. I also, changed the label width to allow some space (minimal amount) for the scrollbar, so it shouldn't be getting cut off/hidden behind the label.

ZakaryH:
ah I think the difference is that I'm trying to do this on a normal Safari browser

Kapture 2024-08-01 at 15 40 49

the extra space to not cut off the scrollbar isn't quite enough here since the space-small is being applied as margin and not counted as part of the container that hold the scrollbar. we'd need a bit more to properly leave room for the scrollbar

but also on Chrome with the dev tools simulating an iphone, we have the initial problem of letters being visible in the space
image

I'm not sure if there's gonna be a great/reliable way to leave the room for the scrollbar. I think your initial idea of using padding top might be better. I'm gonna go back to one of those commits and see if we can get it working with a minor tweak. I think you were really close at one point!

ZakaryH (approved):
LGTM

appreciate your patience with all this back and forth!

PR #1975, opened 2024-08-06
fix(components): Adjust z-index of clickable areas in FormatFile

taylorvnoj:
## Motivations

`FormatFile` has an issue when both `onClick` and `onDelete` are present, where you cannot click the delete Icon immediately after clicking the `FormatFile` component itself.  Instead, you'd have to remove focus from the `FormatFile` in order to click the delete Icon.

### Added

- a new `ExpandedwithDelete` story; good for testing this functionality but also, we should have a story with the `onDelete`
- added a `deleteButton` class so that we can raise the z-index of that element

#### Before

When an `onClick` & an `onDelete` are both present, you cannot click the delete button immediately after triggering the `onClick` (volume up):

https://github.com/user-attachments/assets/abaded2d-d0df-4f8d-9c89-8f1e940e6f9b

#### After

When `onClick` & `onDelete` are both present, you can trigger the `onClick` and then trigger the `onDelete` without removing focus/tabbing (volume up):

https://github.com/user-attachments/assets/7fbf334e-fe35-42e1-86a5-60eb8e3f1c2d

## Testing

- test this in the new `ExpandedwithDelete` story
- pre-release sent in slack msg

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormatFile/FormatFile.css
+++ b/packages/components/src/FormatFile/FormatFile.css
@@ -77,6 +77,10 @@
   box-shadow: var(--shadow-focus);
 }
 
+.deleteButton {
+  z-index: 1;
+}
+
 .expanded .thumbnail {
   border-right: none;
   border-bottom-left-radius: var(--radius-base);
--- a/docs/components/FormatFile/Web.stories.tsx
+++ b/docs/components/FormatFile/Web.stories.tsx
@@ -42,3 +42,18 @@ Collapsed.args = {
   displaySize: "large",
   onDelete: () => alert("Deleted"),
 };
+
+export const ExpandedWithDelete = BasicTemplate.bind({});
+ExpandedWithDelete.args = {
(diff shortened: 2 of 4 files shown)


packages/components/src/FormatFile/FormatFile.css
```diff
@@ -77,6 +77,10 @@
   box-shadow: var(--shadow-focus);
 }
 
+.deleteButton {
+  z-index: 1;
```
ZakaryH, line 81: lil nitpick: is there a token value we can use here instead?

taylorvnoj: @ZakaryH that's a good nit. I can put `elevation-base` on there and send it to you for review?

ZakaryH: sounds good to me!

PR #1978, opened 2024-08-07
feat(components): Update positioning of dropzone content

taylorvnoj:
## Motivations

When `InputFile` is within a tall container and is stretched to fill the space, the dropzone content (`Button`, `hintText`) should be centered vertically and horizontally.

### Added

- new styles for `dropzoneContent`

### Changed

Instead of the dropzone content sticking to the top of a tall `InputFile`, that content is now in the center. 

#### Before

Screenshot 2024-08-07 at 8 51 33 PM

#### After

Screenshot 2024-08-07 at 8 39 03 PM

If the `InputFile` is set to `size="small"` (so only a button shows up), this is what we see:

Screenshot 2024-08-07 at 8 54 33 PM

Note that any `InputFile` not stretched within a tall container retains the same styling as before:

Screenshot 2024-08-07 at 8 56 25 PM

## Testing

Pre-release showing prod example will be in separate PR tagged in slack msg

To see this in storybook, update the StatefulTemplate in `docs/components/InputFile/Web.stories.tsx` and see the screenshots above:

```
const StatefulTemplate: ComponentStory<typeof InputFile> = args => {
  const [files, setFiles] = useState<FileUpload[]>([]);

  return (
      <>
      <div style={{ height: '220px', width: '620px', display: 'flex', flexDirection: 'column' }}>
        <div style={{ flexGrow: 1, display: 'flex' }}>
        <InputFile
          {...args}
          onUploadStart={handleUpload}
          onUploadProgress={handleUpload}
          onUploadComplete={handleUpload}
          style={{ flexGrow: 1 }}
        />
        </div>
        </div>
        {files.map(file => (
          <FormatFile file={file} key={file.key} />
        ))}
      </>
  );

  function handleUpload(file: FileUpload) {
    setFiles(oldFiles => updateFiles(file, oldFiles));
  }
};
```

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -235,14 +235,16 @@ export function InputFile({
         <input {...getInputProps()} />
 
         {variation === "dropzone" && (
-          <Content spacing="small">
-            <Button label={buttonLabel} size="small" type="secondary" />
-            {size === "base" && (
-              <Typography size="small" textColor="textSecondary">
-                {hintText}
-              </Typography>
-            )}
-          </Content>
+          <div className={styles.dropzoneContent}>
+            <Content spacing="small">
+              <Button label={buttonLabel} size="small" type="secondary" />
+              {size === "base" && (
+                <Typography size="small" textColor="textSecondary">
+                  {hintText}
+                </Typography>
+              )}
+            </Content>
+          </div>
         )}
 
         {variation === "button" && (
--- a/packages/components/src/InputFile/InputFile.css
+++ b/packages/components/src/InputFile/InputFile.css
@@ -31,3 +31,10 @@
   display: flex;
   flex-direction: column;
 }
+
+.dropzoneContent {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+}
--- a/packages/components/src/InputFile/InputFile.css.d.ts
+++ b/packages/components/src/InputFile/InputFile.css.d.ts
@@ -4,6 +4,7 @@ declare const styles: {
   readonly "active": string;
   readonly "error": string;
   readonly "validationErrors": string;
+  readonly "dropzoneContent": string;
 };
 export = styles;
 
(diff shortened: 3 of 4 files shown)


packages/components/src/InputFile/InputFile.tsx
```diff
@@ -235,14 +235,16 @@ export function InputFile({
...
-          <Content spacing="small">
-            <Button label={buttonLabel} size="small" type="secondary" />
-            {size === "base" && (
-              <Typography size="small" textColor="textSecondary">
-                {hintText}
-              </Typography>
-            )}
-          </Content>
+          <div className={styles.dropzoneContent}>
+            <Content spacing="small">
```
JeffMarvin, line 239: I'm not sure that `<Content>` is doing much for us here... Consider just styling the children directly if that's what we need (i.e. adding the vertical padding to the children through the CSS relating to `.dropzoneContent`). 

(There's also something weird happening where Content isn't getting the right value for `--public-content--padding` but that's out of scope here.)

JeffMarvin (commented):
Consider one tiny tweak to the CSS -- let me know what you think!

packages/components/src/InputFile/InputFile.tsx
taylorvnoj: `Content` here is pre-existing and is keeping things spaced/aligned nicely:

Screenshot 2024-08-09 at 2 21 16 PM

If we want to accomplish the same thing through css it would look like the following (which, we're getting for free just by wrapping the children in `Content`)

```
.dropzoneContent {
  display: flex;
  align-items: center;
  justify-content: center;
  height: 100%;
  flex-direction: column; // add this
  gap: var(--space-small); // add this
}
``` 

I think I hear what you're saying - I added a css class to adjust styling in that area, so why not. But I'm leaning toward keeping the `Content` there - I feel that we should be keeping that component-based architecture and reusing our own components to get what we need.  

The way I see it, the dropzone (button and text) is a container that is wrapped by `Content` so that all children within, are consistently spaced. No matter how many children get added to that container in the future, we know that `Content` will handle it

The `dropzoneContent` is making sure that the alignment of that dropzone container, is where it needs to be when `InputFile` is stretched.  

Is there a reason to move the logic to css that I'm not thinking about? @JeffMarvin

JeffMarvin: Ok, that's fair. You convinced me with the whole composability argument! 🥇

JeffMarvin (approved):

PR #1983, opened 2024-08-14
fix(components): Always show delete button on the FormatFile when on a small screen

MichaelParadis:
## Motivations

On mobile breakpoints if a `FormatFile` was in displaying in `compact` with an onDelete the delete icon would not display all the time. This would make deleting files hard if there was an `onClick` action on the `FormatFile`

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- fix(components): Have the compact `FormatFile`s always display the delete button when on a mobile breakpoint
### Security

- 

## Testing

 - Verify compact formatFile doesn't display the delete button when the thumbnail isn't hovered on non-mobile breakpoints
 - Verify the compact fomatFile will always display the delete button when on a mobile breakpoint

See mentioned PR for to test the pre-release in JO

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormatFile/FormatFile.css
+++ b/packages/components/src/FormatFile/FormatFile.css
@@ -137,8 +137,10 @@
   right: var(--space-smaller);
 }
 
-.deleteable ~ .deleteButton {
-  visibility: hidden;
+@media (--medium-screens-and-up) {
+  .deleteable ~ .deleteButton {
+    visibility: hidden;
+  }
 }
 
 .deleteButton:focus-within,
--- a/packages/components/src/FormatFile/FormatFile.css
+++ b/packages/components/src/FormatFile/FormatFile.css
@@ -141,11 +141,11 @@
   .deleteable ~ .deleteButton {
     visibility: hidden;
   }
-}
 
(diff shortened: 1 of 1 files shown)


packages/components/src/FormatFile/FormatFile.css
```diff
@@ -137,13 +137,15 @@
...
-.deleteButton:hover,
-.deleteable:hover ~ .deleteButton,
-.deleteable:focus ~ .deleteButton {
-  visibility: visible;
+@media (--medium-screens-and-up) {
+  .deleteable ~ .deleteButton {
+    visibility: hidden;
+  }
+
+  .deleteButton:focus-within,
```
MichaelParadis, line 145: Moved this into the media query because it will default to be visible on small breakpoints, as a result we only need to do the `hover` updates when on non-mobile breakpoints

taylorvnoj (approved):
I used the build link on my phone and it works really well - the delete button persists and I can very easily click that, or the image to trigger the `onClick`. Things still work as they should in web! 
Thanks for doing this so quickly! LGTM 🚀

PR #1987, opened 2024-08-20
feat(components): use text color for menu items

chris-at-jobber:
## Motivations

Because the menu items are not "secondary" to any other content, they should just use the default text color.

### Changed

- Menu items use `text` color instead of `textSecondary`
- Also noticed that Typography doesn't have access to `destructive` so included that as well

## Testing

Test in Storybook and via prerelease

Changes can be

---

Random photo of Atlantis


nad182 (approved):
I tested the `Menu` component in preview app vs prod. The colours appear to look darker, which aligns with Chris's expectation/definition of menu items not being secondary to some other elements.
Code looks good too!

PR #1989, opened 2024-08-20
feat(components): Export component names as meta.json

jdeichert:
## Motivations

This PR adds a new `meta.json` file to both `components` and `components-native` packages. This new file is meant to export metadata that other projects can consume (meant for CI use cases). For now it just exports a list of all component names in that package. In the future it could export more info as we see fit.

We need to manually update meta.json whenever we ship new components (or remove them). The tests I added will verify meta.json is always up to date, which prevents us from forgetting to update it.

### Added

- Added meta.json to components, manually added the list of components
- Added meta.json to components-native, manually added the list of components
- Wrote tests to verify this file is up to date

### Fixed

- Fixed a tsconfig error when viewing in VS Code

## Testing

To test this, modify meta.json and delete a few component names. The unit tests should fail accordingly. 

---

### DIFF
--- a/packages/components/src/utils/meta/meta.test.tsx
+++ b/packages/components/src/utils/meta/meta.test.tsx
@@ -13,6 +13,7 @@ describe("meta", () => {
   });
 });
 
+// NOTE: Keep this in sync with components-native/src/utils/meta/meta.test.tsx
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function isComponent(name: string, value: any): boolean {
   const isFirstLetterUppercase = /^[A-Z]/.test(name);
@@ -21,6 +22,7 @@ function isComponent(name: string, value: any): boolean {
   return isFirstLetterUppercase && isFunctionComponent;
 }
 
+// NOTE: Keep this in sync with components-native/src/utils/meta/meta.test.tsx
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function isContext(value: any): boolean {
   return (
@@ -31,6 +33,7 @@ function isContext(value: any): boolean {
   );
 }
 
+// NOTE: Keep this in sync with components-native/src/utils/meta/meta.test.tsx
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function isForwardedRef(name: string, value: any): boolean {
   const isFirstLetterUppercase = /^[A-Z]/.test(name);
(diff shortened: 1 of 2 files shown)


packages/components/src/utils/meta/meta.test.tsx
ZakaryH, line 1: all of the `findComponentNamesDeep` and the functions it uses are the same at the `components-native` version of `meta.test.tsx` right?

I'm tempted to see if we can find a way of sharing it but we don't really have a great place to put shared utils, specifically internal-use-only utils that we want to use across packages...

that said, with these functions only being duplicated once each I'm ok if we can't find a nice way to DRY it up

packages/components/src/utils/meta/meta.test.tsx
```diff
@@ -0,0 +1,79 @@
+import * as fs from "fs/promises";
+import * as allExports from "../../index";
+
+describe("meta", () => {
+  // If this test fails, please update meta.json accordingly.
+  it("verifies that the meta.json file is up to date", async () => {
```
ZakaryH, line 6: nice. I was going to ask about how we might ensure the meta entires stay up to date.

I had been thinking about it the other way around, where we have a script like `findComponentNamesDeep` and use that to write the `meta.json` file as part of the build step or something which might require less manual intervention

ZakaryH: fair enough. adding an entry for your new component when you get a failing test seems like a trivial task. I'm happy with this solution!

packages/components-native/src/utils/meta/meta.json
```diff
@@ -0,0 +1,68 @@
+{
+  "exportedComponents": [
+    "ActionItem",
+    "ActionItemGroup",
+    "ActionLabel",
+    "ActivityIndicator",
+    "AtlantisContext.Provider",
+    "AtlantisContext.Consumer",
```
ZakaryH, line 8: part of me finds these Consumer entries kind of weird

on one hand, I don't expect anyone to ever use that since it's been replaced by `useContext`

on the other, it is correct that it is something we are exporting when we do `export SomeContext = React.createContext("whatever");`

ZakaryH (approved):
made a few new components in `components` and `components-native` and ran the tests

- React.memo'd
- ForwardRef'd
- standard
- a Context (provider & consumer)

all of these failed the test as expected. I have a couple comments but they're mostly thoughts. nothing blocking.

LGTM

PR #1991, opened 2024-08-21
feat(components): De-dupe `Toast` by implementing an `id` prop

timoteialbu:
## Motivations

Wanted to prevent a `Toast` from appearing multiple times if the page re-renders. Change the `id` type that already existed to a string so that we can have more freedom in terms of what it can be. Was worried that allowing it to be a number might lead to collisions with the incremental `toastKey` state we had as its likely someone would use `1` or something as the id, which would be the same as auto generated one. 

### Added

- Implement an `id` for the `showToast` method that de-dupes the Toast component and makes sure its not displayed multiple times when the page re-renders.

## Testing

Changes can be

---

Random photo of Atlantis


packages/components/src/Toast/Toast.tsx
```diff
@@ -9,7 +9,7 @@ import { Typography } from "../Typography";
 interface BaseToastProps {
   readonly variation?: "info" | "success" | "error";
   readonly message: string;
-  readonly id?: number;
+  readonly id?: string;
```
ericchernuka, line 12: I think we need to support backwards compatibility by doing `string | number`?

timoteialbu: I thought about that but I don't see it being used anywhere in JO (I opened a PR to see if any errors appear in CI, but I also search for all references of `id` and nothing). If you mean in terms of 3rd parties using this and us breaking it, that's fair, but is it worth it? Cause I doubt this will affect anyone either.

ericchernuka: That's probably right, but I'll let Atlantis folks weigh in.

JeffMarvin: Yeah, thanks for making this backward compatible. We use semantic versioning and this would be a breaking change.

PR #1992, opened 2024-08-21
feat(Disclosure): allow custom title

demoraesgui:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


packages/components/src/Disclosure/Disclosure.test.tsx
```diff
@@ -32,3 +33,23 @@ it("renders a Disclosure that is opened if `defaultOpen` is set", () => {
...
+        }
+      >
+        <span>Content</span>
+      </Disclosure>,
+    );
+
+    expect(screen.getByText("Custom Title")).toBeInTheDocument();
+    expect(screen.getByTestId("archive")).toBeInTheDocument();
+  });
+});
```
nad182, line 55: I think it might be a good idea to add a test to make sure the "standard"/"string" `title` is not getting rendered when the "custom title" is passed. This might be not so straightforward, since we don't seem to be using an "id", "role" or other identifiers for string titles, but I see 2 alternatives:
1. Using a `toHaveClass` accessor to make sure `customSummaryWrap` class is present on a `div` (although you'd need some identifier on that `div`. Perhaps, `role="presentation"`?)
2. To check that `h4` element is NOT in the document (since this is what would be rendered in case the string was passed). However, this one might be a slightly riskier option, since the implementation of that title can be changed (e.g. from `<h4>` to `<p>`). Though we have snapshot testing in place to catch that, so shouldn't be a big issue.

nad182 (changes requested):
Overall looks good. Left 2 comments/suggestions. Let me know what you think, since neither is a hard request!

packages/components/src/Disclosure/Disclosure.test.tsx
nad182: Makes sense!

demoraesgui:
Thanks for the feedback @nad182. Would you mind giving another look?

packages/components/src/Disclosure/Disclosure.test.tsx
```diff
@@ -32,3 +33,35 @@ it("renders a Disclosure that is opened if `defaultOpen` is set", () => {
...
+    render(
+      <Disclosure title={<span>Custom Title</span>}>
+        <span>Content</span>
+      </Disclosure>,
+    );
+
+    expect(screen.getByText("Custom Title").parentElement).toHaveClass(
+      "customSummaryWrap",
+    );
+  });
```
nad182, line 66: I think in this case it would be more pragmatic to test the opposite/negative scenario, i.e. in case the string title is passed, the "customSummaryWrap" should not be on the parent:

```suggestion
  it("should not render a custom title if the passed 'title' is a string", () => {
    const title = "String Title";
    const customTitleClass = "customSummaryWrap";
    render(
      <Disclosure title={title}>
        <span>Content</span>
      </Disclosure>,
    );

    expect(screen.getByText(title).parentElement).not.toHaveClass(customTitleClass);
  });
```

demoraesgui: Yeah, you're right, I think it is better, 1sec.

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -73,3 +76,24 @@ export function Disclosure({
...
+   * Title for the disclosure pane.
+   */
+  readonly title: string | ReactNode | ReactNode[];
+  /**
+   * Size when the title is string.
+   */
+  readonly size: "base" | "large";
+}
+
+const DisclosureTitle = ({ title, size }: DisclosureTitleProps) => {
```
nad182, line 91: This is my mistake, but I think using function declarations would a preferred way (vs. function expressions):

```suggestion
function DisclosureTitle({ title, size }: DisclosureTitleProps) => {
```

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -42,10 +42,12 @@
 .summaryWrap {
   align-items: baseline;
   display: flex;
+  justify-content: space-between;
 }
 
-.summaryWrap h4 {
-  flex: 1 0 0;
```
ZakaryH, line 48: curious about this removal. do we not need this anymore?

I'm not entirely sure why we wanted an inline element to fill out the full width available to it in the first place if all it is is some text, but yeah this would technically be behaving different from `master`

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -36,6 +37,7 @@ export function Disclosure({
   const [isMounted, setMount] = useState(false);
   const [titleRef, { exactWidth }] = useResizeObserver<HTMLDivElement>();
   const isBelowBreakpoint = exactWidth && exactWidth < Breakpoints.small;
+  const isTitleString = typeof title === "string";
```
ZakaryH, line 40: we're doing this same check twice. once up here and then again inside `DisclosureTitle`.

not the end of the world, but since `DisclosureTitle` is an internal component, we might as well add a prop that can leverage the check we already did

maybe something like `isTitleString: boolean`

if this was a component/prop being exposed to consumers I'd be tempted to make it more behavior focused rather than implementation details but since it's not, I don't think we need to fuss about that.

packages/components/src/Disclosure/Disclosure.css
demoraesgui: Yeah, this is where I'm a bit concern as well, If we keep it, we'll need to remove this behavior when using a customTitle, because I may want a h4 that doesn't follow this flex config.

I figured that having space between will look the same, but I agree with you, it behaves differently from `master`.

demoraesgui: Do you have any suggestion? What's the default behavior for flex when isn't set?

Maybe I just overwrite the flex behavior in the customSummaryWrap like:

```css
customSummaryWrap h4 {
  flex: default?
}
```

ZakaryH: ahh right. if someone passes in custom things with their own flex/layout and it happens to contain an h4 - then this rule would interfere with it in a non obvious, and kind of annoying way since it's an internal style.

so while it definitely behaves differently from `master` I'm trying to decide if that even matters....

image

I took a look at the original PR and couldn't find any comments or explanation as to why we wanted the text to grow

I'm struggling to think of why you'd want your element to fill the full width if it didn't need it. since it's just text, it's going to keep looking like empty space either way.... and it's not going to interfere with clicks or anything

I'll see if anyone on the team remembers

ZakaryH: ahhh ok I get it now

on master if you remove those rules and don't do anything else at all. you get this
image

which is not desirable. it should be spanning the entire width. so that's why we needed it.

I don't think it's doing the same thing on this PR though. I swear it was still filling the whole width even with only a string, and a small amount of text.

ZakaryH: @demoraesgui we're good. 

the `h4` selector rule is no longer needed. `space-between` does the same thing in a better way!

nad182 (approved):
Looks good!

packages/components/src/Disclosure/Disclosure.css
demoraesgui: > ahh right. if someone passes in custom things with their own flex/layout and it happens to contain an h4 - then this rule would interfere with it in a non obvious, and kind of annoying way since it's an internal style.

Yeah this is pretty much what I was trying to avoid.

> @demoraesgui we're good.

Awesome, thanks for the feedback :)

PR #1995, opened 2024-08-23
feat(components): Separate Chip from Chips Again

ZakaryH:
## Motivations
a fresh, new take on: https://github.com/GetJobber/atlantis/pull/1694

merging/rebasing was a headache with how much has changed since that PR was made. given it consists of 2 commits, I decided cherry picking them onto a fresh branch was the way to go.

## Changes

Please refer to the original linked PR for details.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Chips/Web.stories.tsx
+++ b/docs/components/Chips/Web.stories.tsx
@@ -2,9 +2,7 @@ import React, { useState } from "react";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
 import { Content } from "@jobber/components/Content";
 import { Chip, Chips } from "@jobber/components/Chips";
-import { Icon } from "@jobber/components/Icon";
 import { Text } from "@jobber/components/Text";
-import { Avatar } from "@jobber/components/Avatar";
 import { useFakeOptionQuery } from "./utils/storyUtils";
 
 export default {
@@ -32,22 +30,15 @@ const BasicTemplate: ComponentStory<typeof Chips> = args => {
       <Text>
         You are <u>{selected ? selected : "_______"}</u>
       </Text>
-      <Chips {...args} selected={selected} onChange={setSelected}>
-        <Chip
-          prefix={<Avatar initials="AZ" />}
-          label="Amazing"
-          value="Amazing"
-        />
-        <Chip
-          prefix={<Icon name="video" />}
-          label="Wonderful"
-          value="Wonderful"
-        />
-        <Chip
-          prefix={<Icon name="starFill" />}
-          label="Brilliant"
-          value="Brilliant"
-        />
+      <Chips
+        {...args}
+        selected={selected}
+        onChange={setSelected}
+        type="singleselect"
+      >
+        <Chip label="Amazing" value="Amazing" />
+        <Chip label="Wonderful" value="Wonderful" />
+        <Chip label="Brilliant" value="Brilliant" />
         <Chip label="Magnificent" value="Magnificent" />
       </Chips>
     </Content>
@@ -97,6 +88,7 @@ const SelectionTemplate: ComponentStory<typeof Chips> = args => {
   return (
     <Chips
       {...args}
+      type="dismissible"
       selected={selected}
       onChange={handleSelect}
       onCustomAdd={handleCustomAdd}
--- a/packages/components/src/Chip/Chip.css
+++ b/packages/components/src/Chip/Chip.css
@@ -5,11 +5,11 @@
   --chip-radius: 20px;
   --chip-base-bg: var(--color-surface--background);
   --chip-base--hover: var(--color-surface--background--hover);
-  display: flex;
+  display: inline-flex;
   position: relative;
   height: var(--chip-height);
   min-width: 0;
-  padding: var(--space-small) var(--base-unit);
+  padding: 0 var(--base-unit);
   border: var(--border-base) solid transparent;
   border-radius: var(--chip-radius);
   color: var(--color-heading);
@@ -88,8 +88,7 @@
   --chip--gradient-color-variation: var(--chip-base-bg);
 }
 
-.base:hover,
-.base:focus-visible {
+.base:hover {
   background-color: var(--chip-base--hover);
   --chip--gradient-color-variation--hover: var(--chip-base--hover);
 }
--- a/packages/components/src/Chip/Chip.css
+++ b/packages/components/src/Chip/Chip.css
@@ -9,6 +9,7 @@
   position: relative;
   height: var(--chip-height);
   min-width: 0;
+  box-sizing: border-box;
   padding: 0 var(--base-unit);
   border: var(--border-base) solid transparent;
   border-radius: var(--chip-radius);
--- a/packages/components/src/Chip/Chip.tsx
+++ b/packages/components/src/Chip/Chip.tsx
@@ -16,11 +16,12 @@ export const Chip = ({
   invalid,
   label,
   value,
+  dataTestID,
   onClick,
   onKeyDown,
   children,
   role = "button",
-  tabIndex = 0,
+  tabIndex,
   variation = "base",
 }: ChipProps): JSX.Element => {
   const classes = classnames(styles.chip, {
@@ -41,19 +42,21 @@ export const Chip = ({
     label,
     heading,
   );
+  const Tag = onClick ? "button" : "div";
 
   return (
-    <Tooltip message={tooltipMessage}>
-      <button
+    <Tooltip message={tooltipMessage} setTabIndex={false}>
+      <Tag
         className={classes}
-        onClick={(ev: React.MouseEvent<HTMLButtonElement>) =>
+        onClick={(ev: React.MouseEvent<HTMLButtonElement | HTMLDivElement>) =>
           onClick && onClick(value, ev)
         }
         tabIndex={tabIndex}
         onKeyDown={onKeyDown}
-        aria-label={ariaLabel || label}
+        aria-label={ariaLabel}
         disabled={disabled}
         role={role}
+        data-testid={dataTestID}
         type="button"
       >
         {prefix}
@@ -79,7 +82,7 @@ export const Chip = ({
           )}
         </div>
         {suffix}
-      </button>
+      </Tag>
     </Tooltip>
   );
 };
--- a/packages/components/src/Chip/Chip.tsx
+++ b/packages/components/src/Chip/Chip.tsx
@@ -49,7 +49,7 @@ export const Chip = ({
       <Tag
         className={classes}
         onClick={(ev: React.MouseEvent<HTMLButtonElement | HTMLDivElement>) =>
-          onClick && onClick(value, ev)
+          onClick?.(value, ev)
         }
         tabIndex={tabIndex}
         onKeyDown={onKeyDown}
--- a/packages/components/src/Chip/Chip.tsx
+++ b/packages/components/src/Chip/Chip.tsx
@@ -21,7 +21,7 @@ export const Chip = ({
   onKeyDown,
   children,
   role = "button",
-  tabIndex,
+  tabIndex = 0,
   variation = "base",
 }: ChipProps): JSX.Element => {
   const classes = classnames(styles.chip, {
--- a/packages/components/src/Chip/Chip.tsx
+++ b/packages/components/src/Chip/Chip.tsx
@@ -16,7 +16,7 @@ export const Chip = ({
   invalid,
   label,
   value,
-  dataTestID,
+  testID,
   onClick,
   onKeyDown,
   children,
@@ -56,7 +56,7 @@ export const Chip = ({
         aria-label={ariaLabel}
         disabled={disabled}
         role={role}
-        data-testid={dataTestID}
+        data-testid={testID}
         type="button"
       >
         {prefix}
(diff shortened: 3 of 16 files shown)


packages/components/src/Chip/Chip.css
```diff
@@ -5,11 +5,11 @@
   --chip-radius: 20px;
   --chip-base-bg: var(--color-surface--background);
   --chip-base--hover: var(--color-surface--background--hover);
-  display: flex;
+  display: inline-flex;
   position: relative;
   height: var(--chip-height);
   min-width: 0;
-  padding: var(--space-small) var(--base-unit);
```
chris-at-jobber, line 12: What's this change coming from?

packages/components/src/Chip/Chip.css
```diff
@@ -88,8 +88,7 @@
   --chip--gradient-color-variation: var(--chip-base-bg);
 }
 
-.base:hover,
-.base:focus-visible {
+.base:hover {
```
chris-at-jobber, line 91: I've recently realized that Chip actually gets a hover whether or not it's interactive, which shouldn't be the case - possible to change that here?

packages/components/src/Chips/InternalChipSingleSelect.tsx
```diff
@@ -34,7 +35,15 @@ export function InternalChipSingleSelect({
               }}
               disabled={child.props.disabled}
             />
-            <InternalChip {...child.props} active={isSelected} />
+            <InternalChip
+              {...child.props}
+              {...(isSelected
+                ? {
+                    suffix: <Icon size="small" name="checkmark" />,
```
chris-at-jobber, line 42: Need to spec `color="interactiveSubtle` here otherwise we get the default "success" green: 
image

packages/components/src/Chips/InternalChipMultiSelect.tsx
```diff
@@ -33,7 +33,9 @@ export function InternalChipMultiSelect({
             <InternalChip
               {...chip.props}
               active={isChipActive}
-              suffix={checkmarkIcon(isChipActive)}
+              {...(isChipActive
+                ? { suffix: <Icon size="small" name="checkmark" /> }
```
chris-at-jobber, line 37: Same as here:  https://github.com/GetJobber/atlantis/pull/1995/files#r1731794668

packages/components/src/Chip/Chip.css
ZakaryH: disclaimer: I can't speak to most of this since I'm really the reviewer and not the author

my best guess is that it's not needed because we're doing `align-items: center` on the parent with a flex direction of row so it gets centered vertically

image

packages/components/src/Chip/Chip.css
ZakaryH: as in if there's no `onClick` it should not have a hover nor `cursor: pointer`?

we do have this in the AC

>Out of scope is introduction of any new state to chip or any new styling changes (will be covered in different ticket)

would you be alright with us addressing that separately?

ZakaryH: your comment did make me notice something that we'd want to address now though

when you have an onClick the height is 40. when you don't, and we render a `div` it is 42px tall.

image

packages/components/src/Chips/InternalChipSingleSelect.tsx
ZakaryH: these were both quick fixes. pushed up em.

packages/components/src/Chip/Chip.css
ZakaryH: ah. it's because `button` gets `box-sizing: border-box` from the user agent stylesheet and `div` does not so it ends up being `content-box` and adds the 2pixels of height to 40 rather than making it be 40 accounting for the border

chris-at-jobber: Rule 26 of front-end development: It's always box-sizing

docs/components/Chips/Web.stories.tsx
```diff
@@ -97,6 +88,7 @@ const SelectionTemplate: ComponentStory<typeof Chips> = args => {
   return (
     <Chips
       {...args}
+      type="dismissible"
```
ZakaryH, line 91: seems like we're doing this same prop assignment twice

```
export const Selection = SelectionTemplate.bind({});
Selection.args = {
  type: "dismissible",
};
```

so `{...args}` is doing `type="dismissible"` but since it's defined as `ChipProps` it _could_ be dismissible which is not enough to satisfy the XOR-ish behavior

I'm not entirely sure why we do this whole `bind` pattern. it seems to upset a bunch of types. regardless, I guess this is fine. at worst it's a bit redundant.

ZakaryH:
@chris-at-jobber I can't approve this PR because I made it 😅 

I was going to approve all the changes that weren't by me, and then ask you for a quick look at the last few commits I've made which are all CSS updates

would you be down to approve this PR? I used the prerelease in product and checked out a handful of existing usages. they all were working as expected.

chris-at-jobber:
I'll pull it down and do a QA/design review, but I would feel _significantly_ more comfortable with another eng at least giving a cursory glance over things 😅

packages/components/src/Chip/Chip.types.d.ts
```diff
@@ -6,6 +6,11 @@ export interface ChipProps extends PropsWithChildren {
    */
   readonly ariaLabel?: string;
 
+  /**
+   * The testing id for the chip if necessary. Defaults to `chip-wrapper`.
+   */
+  dataTestID?: string;
```
MichaelParadis, line 12: I think it may be worth changing this to `testID` to match other test ID props we expose such as 
https://github.com/GetJobber/atlantis/blob/JOB-79340/JOB-80485/actually-separate-third-go/packages/components/src/Icon/Icon.tsx#L31

packages/components/src/Chip/Chip.tsx
```diff
@@ -16,11 +16,12 @@ export const Chip = ({
   invalid,
   label,
   value,
+  dataTestID,
```
MichaelParadis, line 19: The jsdoc comment mentions it defaults to `chip-wrapper` but that doesn't seem to be the case
https://github.com/GetJobber/atlantis/pull/1995/files#diff-93fff87012a73ee50578dd51a34c028ffeea01f74ada82dc08cf121ef8b80011R10

I may also be worth having it follow our other testID formats such as `ATL-CHIP-WRAPPER`

MichaelParadis: I noticed it is defaulted in `Chips` usages but since `Chip` is exported separately it should be defaulted here

packages/components/src/Chip/Chip.tsx
```diff
@@ -41,19 +42,21 @@ export const Chip = ({
...
 
   return (
-    <Tooltip message={tooltipMessage}>
-      <button
+    <Tooltip message={tooltipMessage} setTabIndex={false}>
+      <Tag
         className={classes}
-        onClick={(ev: React.MouseEvent<HTMLButtonElement>) =>
+        onClick={(ev: React.MouseEvent<HTMLButtonElement | HTMLDivElement>) =>
           onClick && onClick(value, ev)
```
MichaelParadis, line 52: Nit: could use optional chaining instead

```suggestion
          onClick?.(value, ev)
```

MichaelParadis:
Some nice to do general cleanup (can be done as a followup PR):
I think [this](https://github.com/GetJobber/atlantis/blob/blob/JOB-79340/JOB-80485/actually-separate-third-go/packages/components/src/Chips/Chip.tsx#L17) error message should be updated to tell consumers to use `Chip` from the `@jobber/components/Chip` to use the Chip separately.


Some stuff that can be moved to another PR: 
There is probably some stuff you could do with the exports [here](https://github.com/GetJobber/atlantis/blob/JOB-79340/JOB-80485/actually-separate-third-go/packages/components/src/Chips/index.ts#L3) to link the `Chip` from the main `Chip` folder instead and remove the file entirely since the "empty" chip component isn't exported / used when just importing from `@jobber/components` based on [this](https://github.com/GetJobber/atlantis/blob/JOB-79340/JOB-80485/actually-separate-third-go/packages/components/src/index.tsx#L14).

packages/components/src/Chip/Chip.tsx
```diff
@@ -41,19 +42,21 @@ export const Chip = ({
...
         }
         tabIndex={tabIndex}
         onKeyDown={onKeyDown}
-        aria-label={ariaLabel || label}
+        aria-label={ariaLabel}
```
MichaelParadis, line 56: Just a question, was a particular reason not to default this to the label if an `ariaLabel` wasn't present?

No change needed if it was intentional

packages/components/src/Chip/Chip.tsx
```diff
@@ -16,11 +16,12 @@ export const Chip = ({
   invalid,
   label,
   value,
+  dataTestID,
   onClick,
   onKeyDown,
   children,
   role = "button",
-  tabIndex = 0,
+  tabIndex,
```
MichaelParadis, line 24: https://github.com/GetJobber/atlantis/pull/1694/files#r1430475459 

> The original Chip implementation didn't do this, so we shouldn't either.

MichaelParadis: I wanted to mention but I don't know if we really want to support this / if this needs to be fixed here but I tested and the Chip is not focusable through the keyboard if no `onClick` is provided but an `onKeyDown` is.

If you want to test for yourself remove `tabIndex` from the default args and only provide an `onKeyDown` but no `onClick` in storybook.

MichaelParadis: Continuing this thread. If no `onClick` or `tabIndex` is provided then it isn't possible to view the tooltip for truncated labels.

Maybe @scotttjob has more context (although this was months ago) on this, based on [this](https://github.com/GetJobber/atlantis/pull/1694/files#r1430476308) comment there were issues with keyboard navigation setting the tab index to 0.

Again, not sure how likely this is but it is something to consider.
CC: @chris-at-jobber

packages/components/src/Chip/Chip.types.d.ts
ZakaryH: good catch, yeah that makes sense to me

packages/components/src/Chip/Chip.tsx
ZakaryH: ah yeah that default isn't true. I'm not sure if it's necessary to have a default either.

we might be fine keeping it optional, and then for our specific usage in `InternalChip` we pass the `ATL-Chip-Wrapper` or whatever we think makes sense

ZakaryH:
undoing one of the things we did in the initial PR
https://github.com/GetJobber/atlantis/pull/1694/files#r1430475459

>The original Chip implementation didn't do this, so we shouldn't either.

in reference to the `tabIndex = 0`

we didn't have to do it before because the single select and multi select versions of Chips are actually inputs, and so we want to leverage their built in behavior. the dismissible version however is a `div` and we did need to pass in `tabIndex=0`

then for singular `Chip`s I think we do want this to default to being focusable via `tabIndex=0`. if we don't do that, then when it's a div and we don't have an `onClick` we can't focus it and won't be able to see things like the truncation tooltip unless we hover it

packages/components/src/Chips/InternalChipSingleSelect.tsx
```diff
@@ -34,7 +35,22 @@ export function InternalChipSingleSelect({
...
+                      <Icon
+                        size="small"
+                        name="checkmark"
+                        color="interactiveSubtle"
+                      />
+                    ),
+                  }
+                : {})}
+              active={isSelected}
+              tabIndex={undefined}
```
MichaelParadis, line 52: Since the default for tab index is set to 0 when it is undefined this should probably be `-1` since the Checkbox / Radio in the internal chips will handle the focus of the chip.

MichaelParadis:
```suggestion
              tabIndex={-1}
```

packages/components/src/Chips/InternalChipMultiSelect.tsx
```diff
@@ -33,7 +33,18 @@ export function InternalChipMultiSelect({
...
+                    suffix: (
+                      <Icon
+                        size="small"
+                        name="checkmark"
+                        color="interactiveSubtle"
+                      />
+                    ),
+                  }
+                : {})}
+              tabIndex={undefined}
```
MichaelParadis, line 47: Same as https://github.com/GetJobber/atlantis/pull/1995/files#r1736369895
```suggestion
              tabIndex={-1}
```

packages/components/src/Chips/InternalChipSingleSelect.tsx
MichaelParadis: Not actually needed here so I will resolve this

packages/components/src/Chips/InternalChipMultiSelect.tsx
ZakaryH: ah right. addressed both of these.

MichaelParadis (approved):
LGTM!

PR #1997, opened 2024-08-27
fix(components): Fix flakey spec in useBreakpoints

MichaelParadis:
## Motivations

In some cases the `useBreakpoints` tests could flake when a PR is merged into master. This PR aims to resolve that issue

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Fixed cases where `useBreakpoints` tests could flake 

### Security

- 

## Testing

`git revert f8fd9ee`
Run the tests for `useBreakpoints` verify that the tests don't fail

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/hooks/src/useBreakpoints/useBreakpoints.test.ts
+++ b/packages/hooks/src/useBreakpoints/useBreakpoints.test.ts
@@ -112,7 +112,7 @@ describe.each(["1", "2", "3", "4", "5", "6"])("useBreakpoints", () => {
 
   describe.each(["1", "2", "3", "4", "5", "6"])("in between values", () => {
     function randomBreakpoint(min: number, max: number) {
-      return Math.floor(min + Math.random() * (max - min + 1));
+      return Math.floor(min + Math.random() * (max - min));
     }
 
     it("should have the correct breakpoint values on higher than xl size screens", () => {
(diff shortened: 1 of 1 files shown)


packages/hooks/src/useBreakpoints/useBreakpoints.test.ts
```diff
@@ -110,9 +110,9 @@ describe("useBreakpoints", () => {
     });
   });
 
-  describe("in between values", () => {
+  describe.each(["1", "2", "3", "4", "5", "6"])("in between values", () => {
     function randomBreakpoint(min: number, max: number) {
-      return Math.floor(min + Math.random() * (max - min + 1));
+      return Math.floor(min + Math.random() * (max - min));
```
MichaelParadis, line 115: Before: `Math.floor(min + Math.random() * (max - min + 1))`  could be greater than the max breakpoint value causing the flakes.

If `Math.random()` was very close to 1 floating point multiplication can get weird
  * `min + 0.999... (max - min + 1)` 
  * ≈ <strike>min</strike> + .999... (max  - <strike>min</strike> + 1)
  *  0 ≈ `min - .999... * min` 
  * 0.999... * (max + 1) can be > max

After:
 *  `min + 0.999... (max - min)` 
 *  ≈ <strike>min</strike> + .999... (max  - <strike>min</strike>)
 * (0.999... * max) < max

MichaelParadis: Based on the docs it seems to suggest it is an inclusive range where as we want an exclusive range
image

MichaelParadis: The tests also seem to suggest that as well
https://github.com/lodash/lodash/blob/4.17.15/test/test.js#L18495-L18505

MichaelParadis: I realized I should just subtract 1 from max and use lodash 🤦

PR #1998, opened 2024-08-27
fix(components):  Update description of title prop in Disclosure component

nad182:
## Motivations

This PR provides a better documentation and example of the acceptable way to pass `title` to the `Disclosure` component in case it's not just a string, but a React Component

### Added

- Better description of the the `title` prop in the Storybook Docs.
- Better description of the the `title` in the <Disclosure> component prop interface.

### Changed

- Storybook example to pass `span` instead of `h4`, since this component appears inside of `<summary`, hence the content should be of ["phrasing content"](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content) type, which was added to the docs.
- Change the typing for the `title` prop to be `ReactElement` and `ReactElement[]` instead of `ReactNode` and `ReactNode[]`, since type `string` IS a `ReactNode`.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Verify that the Storybook Docs for the `Disclosure` component provide a comprehensive explanation of what can be passed as a `title`.
- Verify that the `CustomTitle` story behaves and looks as expected.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Disclosure/Disclosure.stories.mdx
+++ b/docs/components/Disclosure/Disclosure.stories.mdx
@@ -35,11 +35,15 @@ to change but is not required to.
   - This can either be a string or a React component. If a React component is
     used (containing multiple children), it should be wrapped in a container
     element and not a `<Fragment>` to maintain correct UI styling.
-  - The elements passed as a `title` can be any heading content, plain text, or
-    HTML that can be used within a paragraph (AKA
+    - Caveat: The container element should NOT be `<div>` as it would break the
+      semantics of the `<summary>` element. Instead, use a `<span>` or `<p>`
+      tag.
+  - The elements passed as a `title` can be plain text, or HTML that can be used
+    within a paragraph (AKA
     ["phrasing content"](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content))
     to ensure that the title is accessible and can be properly read by screen
-    readers.
+    readers. A heading may also be used but will not be treated as a heading by
+    assistive technologies.
 - `children` should be actionable and clear. The contents of a Disclosure can
   be:
   - plain text
--- a/docs/components/Disclosure/Disclosure.stories.mdx
+++ b/docs/components/Disclosure/Disclosure.stories.mdx
@@ -31,7 +31,14 @@ to change but is not required to.
 ## Content guidelines
 
 - `title` should be informative and label the type of content grouped in the
-  body of the Disclosure
+  body of the Disclosure.
+  - This can either be a string or a React component. If a React component is
+    used (containing multiple children), it should be wrapped in a container
+    element and not a `<Fragment>` to maintain correct UI styling.
+  - The elements passed as a `title` should ideally be of a
+    ["phrasing content"](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content)
+    type to ensure that the title is accessible and can be properly read by
+    screen readers.
 - `children` should be actionable and clear. The contents of a Disclosure can
   be:
   - plain text
--- a/docs/components/Disclosure/Disclosure.stories.mdx
+++ b/docs/components/Disclosure/Disclosure.stories.mdx
@@ -35,10 +35,11 @@ to change but is not required to.
   - This can either be a string or a React component. If a React component is
     used (containing multiple children), it should be wrapped in a container
     element and not a `<Fragment>` to maintain correct UI styling.
-  - The elements passed as a `title` should ideally be of a
-    ["phrasing content"](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content)
-    type to ensure that the title is accessible and can be properly read by
-    screen readers.
+  - The elements passed as a `title` can be any heading content, plain text, or
+    HTML that can be used within a paragraph (AKA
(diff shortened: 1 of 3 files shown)


nad182:
@demoraesgui , I assigned you just as an FYI, since you worked on this originally.

docs/components/Disclosure/Disclosure.stories.mdx
```diff
@@ -31,7 +31,14 @@ to change but is not required to.
...
 - `title` should be informative and label the type of content grouped in the
-  body of the Disclosure
+  body of the Disclosure.
+  - This can either be a string or a React component. If a React component is
+    used (containing multiple children), it should be wrapped in a container
+    element and not a `<Fragment>` to maintain correct UI styling.
+  - The elements passed as a `title` should ideally be of a
+    ["phrasing content"](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content)
+    type to ensure that the title is accessible and can be properly read by
+    screen readers.
```
ZakaryH, line 41: oh actually maybe we should mention that Heading is acceptable too

docs/components/Disclosure/Web.stories.tsx
```diff
@@ -35,7 +35,7 @@ CustomTitle.args = {
   title: (
     <Flex template={["shrink", "grow"]} gap="small">
       <Icon name="sparkles" />
-      <Heading level={5} element="h4">
+      <Heading level={5} element="span">
```
ZakaryH, line 38: I think it's alright for us to leave it as a h4 no?

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -16,8 +16,9 @@ interface DisclosureProps {
 
   /**
    * Title for the disclosure pane.
+   * If ReactElement[] is provided, it must be wrapped in a container element (not a Fragment).
    */
-  readonly title: string | ReactNode | ReactNode[];
+  readonly title: string | ReactElement | ReactElement[];
```
ZakaryH, line 21: nice. the type is a bit clearer that string is acceptable now in the docs

though the far right interpretation of the types is horribly unreadable. that's not because of this PR. our tooling is being a bit more verbose than is helpful.

ZakaryH: image

docs/components/Disclosure/Disclosure.stories.mdx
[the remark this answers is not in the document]
nad182: So if I understand the description, technically `Heading content` CAN be used, but it won't work "properly" (as a heading) on assistive technologies. Are we good with that?

docs/components/Disclosure/Web.stories.tsx
nad182: It should be fine. However, I changed it based on that accessibility concern you mentioned above (just to gently "nudge" consumers to use "span" or any other phrasing content). I can revert this change if that would be your preference! Just LMK.

packages/components/src/Disclosure/Disclosure.tsx
nad182: Yep, I tried providing more specificity there, but it wasn't getting any shorter...

docs/components/Disclosure/Disclosure.stories.mdx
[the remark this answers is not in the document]
ZakaryH: that's my understanding too.

you won't be able to navigate to a heading because it'll lose its role, but it'll still look like a heading.

given the spec permits it, I'm good with us advising the same thing!

docs/components/Disclosure/Web.stories.tsx
ZakaryH: that's fair. if we mention the details about the content type + accessibility in our docs, I'm fine with offering an example of something more accessible via the `span`.

docs/components/Disclosure/Disclosure.stories.mdx
```diff
@@ -31,7 +31,15 @@ to change but is not required to.
 ## Content guidelines
 
 - `title` should be informative and label the type of content grouped in the
-  body of the Disclosure
+  body of the Disclosure.
+  - This can either be a string or a React component. If a React component is
+    used (containing multiple children), it should be wrapped in a container
+    element and not a `<Fragment>` to maintain correct UI styling.
+  - The elements passed as a `title` can be any heading content, plain text, or
```
ZakaryH, line 38: I'm wondering if this reads like a contradiction at all. we first say to basically use a `div`. then we say correct elements to use are phrasing content, which a `div` is not. after playing around with this a bit I think I understand it, but it's not super obvious.

what do you think?

similarly, I'd restructure this point to make it a bit clearer. right now it says

>The elements passed as a `title` can be any heading content, plain text, or HTML that can be used within a paragraph (AKA phrasing content) to ensure that the title is accessible and can be properly read by screen readers.

except that heading content will have the opposite effect

so instead, how about...

>The elements used within `title` can be plain text, or valid paragraph content HTML (AKA phrasing content) to ensure that the title is accessible and can be properly read by screen readers. A heading may also be used but will not be treated as a heading by assistive technologies.

or something along those lines.

nad182: That sounds perfect! Thanks for the suggestion!

PR #2007, opened 2024-09-03
feat(components): Add maxFiles prop to InputFile

taylorvnoj:
## Motivations

In the process of building a component using `InputFile` to upload and display expanded `FormatFile` components, we discovered a limitation with setting a `max_files` validation.

While setting `MAX_FILES = 10` validation via the `InputFile` `onUploadComplete` prop, we noticed that although the UI will only show max 10 files, any additional images that were selected are continuing to be uploaded to storage.  We want to avoid that, and actually limit # of files that can be uploaded while using `InputFile`.

### Added

An optional `maxFilesValidation` prop which is an object, consisting of `maxFiles` and `numberOfCurrentFiles`.  This prop is being mapped to `react-dropzone`'s `maxFiles` [option](https://react-dropzone.js.org/#!/Accepting%20specific%20number%20of%20files) via `useDropzone`.

### Changed

When uploading more than allowed `maxFiles`, `react-dropzone` will fail all uploads and throw an error for every attempted file upload:
Screenshot 2024-09-06 at 10 35 01 PM

To avoid a potential long list of validation errors, we thought it best to group all `maxFile` errors into one `InputValidation` message.  
Screenshot 2024-09-06 at 10 43 36 PM

This means that the `validationErrors` constant is now checking if the acc array already has a "too-many-files" error and if not, creating a new error object. Any other error code, like going beyond file size, will go into a different error object and surface as it's own validation error.  File size validations are remaining the same (multiple may appear) because they are helpful in indicating which specific files are too large.

I've updated the `InputFile` tests, added a `maxFiles` test & added some info for the Docs page.

## Testing

I've created a Max Files Limit story to test these changes.

Changes can be
[tested via Pre-release](https://github.com/GetJobber/atlantis/actions/workflows/trigger-qa-build.yml) commented below

---

Random photo of Atlantis


packages/components/src/InputFile/InputFile.test.tsx
```diff
@@ -112,9 +114,11 @@ describe("Post Requests", () => {
         onUploadComplete={handleComplete}
       />,
     );
-    const input = container.querySelector("input[type=file]");
+    const input = container.querySelector(
+      "input[type=file]",
+    ) as HTMLInputElement;
 
-    fireEvent.change(input, { target: { files: [testFile] } });
+    await userEvent.upload(input, testFile);
```
MichaelParadis, line 121: Love these changes to userEvent (:

docs/components/InputFile/Web.stories.tsx
```diff
@@ -51,6 +51,29 @@ const StatefulTemplate: ComponentStory<typeof InputFile> = args => {
   }
 };
 
+const MaxFilesTemplate: ComponentStory<typeof InputFile> = args => {
+  const [files, setFiles] = useState<FileUpload[]>([]);
```
MichaelParadis, line 55: I think it would be worth adding the "manually" managing the state for the errors described here
image

That way it is more clear as to how to manage it instead of leaving it up to developers to figure it out

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -123,6 +123,11 @@ interface InputFileProps {
    */
   readonly allowMultiple?: boolean;
 
+  /**
+   * Maximum number of files that can be uploaded via the dropzone.
+   */
+  readonly maxFiles?: number;
```
MichaelParadis, line 129: I think it would be worth adding a note about the scenario described here
image

Since people are likely going to set it and forget it

MichaelParadis: Another thought could be to do some typescript stuff with `XOR` to require a numberOfFiles prop to be passed to the component to shift this into our component so we can figure out the number of remaining files that can be uploaded

MichaelParadis (changes requested):
Blocking:
 - Updating the story to show the managing state for already uploaded files that is described in the docs.
 - Updating the JSDocs to describe the scenario mentioned in the storybook docs
Ideal but non-blocking and open to discussion:
- Changing the logic so `maxFiles` needs to be used in conjunction with a `numberOfCurrent` files like prop to allow us to shift some of the functionality instead of expecting consumers to remember that last step

taylorvnoj:
@MichaelParadis we now have `maxFilesValidation` containing `maxFiles` and `numberOfCurrentFiles`.

While I was in there I added a file size validator story separate from the max files story & documentation updated.

packages/components/src/InputFile/InputFile.test.tsx
```diff
@@ -176,14 +182,18 @@ describe("Post Requests", () => {
       const fetchParams = jest.fn(fetchUploadParams);
       const handleError = jest.fn();
 
-      (axios.request as jest.Mock).mockReturnValue(Promise.reject("error"));
+      (axios.request as jest.Mock).mockRejectedValue(
```
MichaelParadis, line 185: Nice use of `mockRejectedValue`

packages/components/src/InputFile/InputFile.test.tsx
```diff
@@ -212,15 +222,43 @@ describe("Post Requests", () => {
...
+
+      await userEvent.upload(input, [testFile, testFile, testFile]);
+
+      await waitFor(() => {
+        expect(
+          getByText("Cannot exceed a maximum of 2 files."),
+        ).toBeInTheDocument();
+      });
+
+      expect(axios.request).not.toHaveBeenCalled();
```
MichaelParadis, line 259: Appreciate verifying the axios request doesn't fire instead of just checking the error message is visible

MichaelParadis (approved):
LGMT!

PR #2009, opened 2024-09-03
fix(components): update Disclosure focus and hover styling

chris-at-jobber:
## Motivations

Improve the appearance of hover and focus states and make them more consistent with other Atlantis elements

### Changed

- Apply a `radius` of small to Disclosure (as it is typically a child component)
- Adjust how the left alignment of the Disclosure works so that on hover and focus-visible, the title is not flush to the edge of the changing surface

### Fixed

- fix transitions on background-color and focus shadow
- changed outline color to transparent to support high-contrast modes

## Testing

Test in Storybook and via prerelease

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Disclosure/Disclosure.css
+++ b/packages/components/src/Disclosure/Disclosure.css
@@ -9,13 +9,17 @@
 }
 
 .summary {
-  padding: var(--space-smaller) 0 var(--space-smallest) 0;
+  --disclosure--interactionOffset: var(--space-smaller);
+  margin-left: calc(-1 * var(--disclosure--interactionOffset));
+  padding: var(--space-smaller) 0 var(--space-smallest)
+    var(--disclosure--interactionOffset);
   border: none;
+  border-radius: var(--radius-small);
   list-style: none;
   cursor: pointer;
-  outline-color: none;
+  outline-color: transparent;
   transition:
-    background var(--timing-quick),
+    background-color var(--timing-base) ease-out,
     box-shadow var(--timing-base);
 }
 
@@ -24,15 +28,18 @@
 }
 
 .summary:hover,
-.summary:focus {
-  background: var(--color-surface--hover);
+.summary:focus-visible {
+  background-color: var(--color-surface--hover);
 }
 
 .summary:focus,
 .summary:active {
-  box-shadow: var(--shadow-focus);
   border: none;
-  outline: none;
+  outline: transparent;
+}
+
+.summary:focus-visible {
+  box-shadow: var(--shadow-focus);
 }
 
 .summary * {
--- a/packages/components/src/Disclosure/Disclosure.tsx
+++ b/packages/components/src/Disclosure/Disclosure.tsx
@@ -59,7 +59,7 @@ export function Disclosure({
             isTitleString={isTitleString}
           />
           <span className={styles.arrowIconWrapper}>
-            <Icon size="large" name="arrowDown" color="green" />
+            <Icon name="arrowDown" color="interactive" />
           </span>
         </div>
       </summary>
--- a/packages/components/src/Disclosure/__snapshots__/Disclosure.test.tsx.snap
+++ b/packages/components/src/Disclosure/__snapshots__/Disclosure.test.tsx.snap
@@ -21,13 +21,13 @@ exports[`renders a Disclosure 1`] = `
         >
           <svg
             data-testid="arrowDown"
-            style="fill: var(--color-icon); display: inline-block; vertical-align: middle; width: 32px; height: 32px;"
+            style="fill: var(--color-icon); display: inline-block; vertical-align: middle; width: 24px; height: 24px;"
             viewBox="0 0 24 24"
             xmlns="http://www.w3.org/2000/svg"
           >
             <path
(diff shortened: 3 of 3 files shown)


packages/components/src/Disclosure/Disclosure.css
```diff
@@ -9,13 +9,17 @@
 }
 
 .summary {
-  padding: var(--space-smaller) 0 var(--space-smallest) 0;
+  --disclosure--interactionOffset: var(--space-smaller);
```
chris-at-jobber, line 12: This might be a shade too clever but wanted to explicitly tie this value to the left spacing and the negative offset

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -9,13 +9,17 @@
...
+    var(--disclosure--interactionOffset);
   border: none;
+  border-radius: var(--radius-small);
   list-style: none;
   cursor: pointer;
-  outline-color: none;
+  outline-color: transparent;
   transition:
-    background var(--timing-quick),
+    background-color var(--timing-base) ease-out,
```
chris-at-jobber, line 22: `background` applies to (4?) multiple `background-` properties, `background-color` is specific to the one property we actually need here, and it works, where prior it did not.

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -24,15 +28,18 @@
 }
 
 .summary:hover,
-.summary:focus {
-  background: var(--color-surface--hover);
+.summary:focus-visible {
```
chris-at-jobber, line 31: Focus visible is how we have been treating interactions across all focusable components by default. Disclosure is a little older and was missed during the retheme.

packages/components/src/Disclosure/Disclosure.css
```diff
@@ -24,15 +28,18 @@
...
+  background-color: var(--color-surface--hover);
 }
 
 .summary:focus,
 .summary:active {
-  box-shadow: var(--shadow-focus);
   border: none;
-  outline: none;
+  outline: transparent;
```
chris-at-jobber, line 38: High-contrast mode ignores box-shadows, and forces an outline color. `none` blocks high-contrast mode from working as expected, transparent is compatible, with no visual impact to non-high-contrast mode users.

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -59,7 +59,7 @@ export function Disclosure({
             isTitleString={isTitleString}
           />
           <span className={styles.arrowIconWrapper}>
-            <Icon size="large" name="arrowDown" color="green" />
+            <Icon name="arrowDown" color="interactive" />
```
chris-at-jobber, line 62: semantic color + default icon size

packages/components/src/Disclosure/Disclosure.css
fagnanm: @chris-at-jobber could we go even further here and use `small` instead? The space is good but it's still pretty tight.

chris-at-jobber: Just poking around, I think we could - might add a bit of the same to the right side in that case

image
image

image
image

It does bug me a _bit_ when the hovered surface/focus ring is more noticeably misaligned to the content, but those states are fleeting and the component itself looks cleaner - thoughts?

chris-at-jobber: I don't love it for a few reasons
- less obvious what area is actually interactive
- inconsistent with the majority of our "hoverable" interactive elements (Button, List, DataList, Menu items, Combobox, Chip, etc)
- since teams can pass in a `{child}` as the title, would create either inconsistent or unpredictable interactive states (both for consuming teams and end users)

fagnanm: @chris-at-jobber k, lets just do more padding on each side?

fagnanm:
Some QA findings:
Feature comparison uses disclosures but changes a bunch of the styling which has it's issues with negative margins.
image
image

chris-at-jobber:
@fagnanm good catch, can make sure to go in and fix that up when we pull this in

fagnanm (commented):
Design approved

taylorvnoj:
@chris-at-jobber Just wanted to highlight a usage we will see with more reactification - the hover/focus goes out of content bounds. I know you previously mentioned these states are fleeting, but I imagine we might hear feedback on this? What do you think

Screenshot 2024-09-24 at 2 06 35 PM

chris-at-jobber:
@taylorvnoj does that example have custom styling applied to it?

taylorvnoj:
@chris-at-jobber it does - but even without the custom styling to title, the offset still goes out of bounds. Lets chat!

Screenshot 2024-09-24 at 3 33 49 PM

taylorvnoj:
👀 just going to get an updated pre-release going and will do one final pass.

taylorvnoj (approved):
Screenshot 2024-09-25 at 1 26 08 PM
Screenshot 2024-09-25 at 1 26 16 PM
Screenshot 2024-09-25 at 1 26 38 PM

Looking good! Thanks for chatting it out with me about it. I think what you've landed on is the best compromise

PR #2010, opened 2024-09-04
fix(components): Add ref prop to primary and secondary actions on Page component

nateb-jobber:
## Motivations
I needed a popover on a button, but the button was a page primary action. 

### Added

- 

### Changed
- added `ref` to div wrapping primary and secondary action buttons
- modified storybook to show example of change
- added test to make sure the `ref` prop is not passed into the button component

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing
- Open `Page` "With Actions" storybook
- Click the primary and secondary actions to see the Popover 

Screenshot 2024-09-04 at 10 15 49 AM

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Page/Web.stories.tsx
+++ b/docs/components/Page/Web.stories.tsx
@@ -1,9 +1,10 @@
-import React from "react";
+import React, { useRef, useState } from "react";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
 import { StatusLabel } from "@jobber/components";
 import { Page } from "@jobber/components/Page";
 import { Content } from "@jobber/components/Content";
 import { Text } from "@jobber/components/Text";
+import { Popover } from "@jobber/components/Popover";
 
 export default {
   title: "Components/Layouts and Structure/Page/Web",
@@ -22,6 +23,52 @@ const BasicTemplate: ComponentStory<typeof Page> = args => (
   </Page>
 );
 
+const PopoverTemplate: ComponentStory<typeof Page> = args => {
+  const primaryDivRef = useRef(null);
+  const [showPrimaryPopover, setShowPrimaryPopover] = useState(false);
+
+  const secondaryDivRef = useRef(null);
+  const [showSecondaryPopover, setShowSecondaryPopover] = useState(false);
+
+  return (
+    <>
+      <Page
+        primaryAction={{
+          label: "Trigger Food Popover",
+          onClick: () => setShowPrimaryPopover(true),
+          ref: primaryDivRef,
+        }}
+        secondaryAction={{
+          label: "Trigger Drink Popover",
+          onClick: () => setShowSecondaryPopover(true),
+          ref: secondaryDivRef,
+        }}
+        {...args}
+      >
+        <Content>
+          <Text>Page content here</Text>
+        </Content>
+      </Page>
+      <Popover
+        attachTo={primaryDivRef}
+        open={showPrimaryPopover}
+        onRequestClose={() => setShowPrimaryPopover(false)}
+        preferredPlacement="bottom"
+      >
+        <Content>Your food order: 🥨</Content>
+      </Popover>
+      <Popover
+        attachTo={secondaryDivRef}
+        open={showSecondaryPopover}
+        onRequestClose={() => setShowSecondaryPopover(false)}
+        preferredPlacement="bottom"
+      >
+        <Content>Your drink order: 🍹</Content>
+      </Popover>
+    </>
+  );
+};
+
 const titleMetaData = (
   <StatusLabel label={"In Progress"} alignment={"start"} status={"warning"} />
 );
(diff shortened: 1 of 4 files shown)


docs/components/Page/Web.stories.tsx
```diff
@@ -22,6 +23,52 @@ const BasicTemplate: ComponentStory<typeof Page> = args => (
   </Page>
 );
 
+const PopoverTemplate: ComponentStory<typeof Page> = args => {
```
nateb-jobber, line 26: Example usage of a Page with Popovers

packages/components/src/Page/Page.tsx
```diff
@@ -14,6 +14,10 @@ import { Button, ButtonProps } from "../Button";
 import { Menu, SectionProps } from "../Menu";
 import { Emphasis } from "../Emphasis";
 
+export type ButtonActionProps = ButtonProps & {
+  ref?: React.RefObject<HTMLDivElement>;
+};
```
nateb-jobber, line 19: Not sure if there's a better way to declare this type? It's a button, but with an optional ref prop. 🤷

packages/components/src/Page/Page.tsx
```diff
@@ -186,3 +193,10 @@ export function Page({
     </div>
   );
 }
+
+export const getActionProps = (actionProps: ButtonActionProps): ButtonProps => {
+  const buttonProps = { ...actionProps };
+  if (actionProps.ref) delete buttonProps.ref;
+
+  return buttonProps;
+};
```
nateb-jobber, line 202: This takes in the primary or secondary action prop, and cleans it by removing the ref prop so it is not passed into the button component.

packages/components/src/Page/Page.test.tsx
```diff
@@ -106,6 +108,23 @@ describe("When actions are provided", () => {
...
+
+  describe("getActionProps", () => {
+    it("given action props, it returns the correct props", () => {
+      const buttonActionProps = {
+        label: "Label",
+      } as ButtonProps;
+      const buttonActionWithRefProps = {
+        label: "Label",
+        ref: { current: null },
+      } as ButtonActionProps;
```
ZakaryH, line 120: for this and the one above it, do we need to cast these objects as these types or can we let Typescript figure it out for us?

nateb-jobber: It's definitely not needed. TS can figure it out no problem. It was more just readability for the person looking at the test to see that it's explicitly being cast as this type.

But I'm happy to remove it.

packages/components/src/Page/Page.tsx
```diff
@@ -154,13 +158,16 @@ export function Page({
             {showActionGroup && (
               <div className={styles.actionGroup}>
                 {primaryAction && (
-                  <div className={styles.primaryAction}>
-                    <Button {...primaryAction} />
+                  <div className={styles.primaryAction} ref={primaryAction.ref}>
```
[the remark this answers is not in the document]
ZakaryH: hmmm I don't love that either.

actually why is this getting margin bottom in the first place... it looks like the `margin-bottom`  rule should be getting stomped on for bigger screen sizes

image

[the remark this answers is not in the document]
ZakaryH: I don't think it's the margin bottom. at a different screen size, this rule also applies and it looks good

image

so something else is making the actions be taller than their content.... flex maybe?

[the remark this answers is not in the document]
ZakaryH: ah ok. it's the h1 that has some margin bottom, and those actions are in the same container making them try to be as tall as the h1. 

I don't think that's right. I can't think of any reason why you'd want your action container elements to be taller than they need to be. it's fine if the header needs to grow, but there's no need for them to grow with it.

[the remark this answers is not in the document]
ZakaryH: giving it an align self of start _should_ fix it

```
.titleBar > .actionGroup {
  flex: 1 0 auto;
  align-self: flex-start;
}
```

gonna play around with this in a few different screen sizes quick to make sure that doesn't have any unwanted side effects.

packages/components/src/Page/Page.test.tsx
ZakaryH: gotcha. personally I prefer to avoid casting unless Typescript is in a situation where it can't figure it out, so if it's all the same to you - let's remove it

ZakaryH:
@nateb-jobber looks great!

one little thing about the type casting and then the `align-self` on the `actionGroup` should fix the extra spacing between the popover and the buttons on large screens

nateb-jobber:
@ZakaryH All fixed!

Here are some screenshots
Screenshot 2024-09-04 at 3 58 29 PM
Screenshot 2024-09-04 at 3 58 38 PM

ZakaryH (approved):
LGTM

PR #2011, opened 2024-09-04
fix(components): Keep dismiss on drawers aligned to top when titles wrap

fagnanm:
## Motivations
When there are longer titles on drawers (like in Automations where the title is user-generated) the dismiss button tries to align center with the height of the title making things look real weird.

## Changes
Adjusted the structure of the headers on `drawer` and `sideDrawer` slightly so that the titles appear center aligned but are actually aligned to the start of the container. This keeps the dismiss button and subsequent actions in the right place even if the title wraps onto more than one line.

## Before
Side Drawer
image

Drawer
image

## After
Side Drawer
image

Drawer
image

## Testing

Try adding titles that wrap onto more than one line and ones that do not. Actions in the headers should still align center with the heading.

Changes can be

---

Random photo of Atlantis


packages/components/src/Drawer/Drawer.css
```diff
@@ -1,6 +1,6 @@
 .container {
   --drawer-width: 280px;
-  --drawer--base-padding: var(--space-large);
+  --drawer--base-padding: var(--space-base);
```
fagnanm, line 3: Made the padding match sideDrawer so they are the same

packages/components/src/Drawer/Drawer.css
```diff
@@ -48,10 +48,21 @@
...
+  align-items: flex-start;
+  gap: var(--space-base);
+}
+
+.heading {
+  margin-top: var(--space-small);
+}
+
+/* Make sure the dismiss button doesn't get squished when title is longer */
+.header > button {
```
fagnanm, line 64: Fixed an existing issue with the button shrinking if the title was too long

packages/components/src/Drawer/Drawer.css
```diff
@@ -48,10 +48,21 @@
...
+  padding: var(--drawer--base-padding) var(--drawer--base-padding) 0
+    var(--drawer--base-padding);
   background-color: transparent;
   justify-content: space-between;
-  align-items: center;
+  align-items: flex-start;
+  gap: var(--space-base);
+}
+
+.heading {
```
taylorvnoj, line 59: Do we need `overflow: hidden` in `heading` or something similar?

If I have a long word, the dismiss icon goes beyond the page at a smaller breakpoint/when the drawer snaps to a smaller size.
Screenshot 2024-09-20 at 5 04 39 PM

SideDrawer seems to handle it a bit better

Screenshot 2024-09-20 at 5 16 34 PM

taylorvnoj (commented):
It's looking better! I did notice that with a longer word and smaller breakpoint, I'm seeing the dismiss button go beyond the page in Drawer. Top is local and bottom is current behaviour.

Screenshot 2024-09-20 at 5 20 20 PM

packages/components/src/Drawer/Drawer.css
fagnanm: @taylorvnoj I'll see if I can fix that, sometimes limited CSS solutions to really long words when their containers are smaller.

fagnanm: @taylorvnoj played with a few things but I don't believe this will be fixable without some potential unpredictability in where words break. We could do something like `word-break: break-word` which will work in some cases but it doesn't have a predictable pattern on where it breaks the words, which is usually an accessibility concern for a screen reader.

I'd say we don't solve it given how many drawers have titles set by us. In cases where drawer titles are set by the user, I imagine it would be very uncommon to have words that exceed the width of the drawer.

taylorvnoj (approved):
LGTM

PR #2013, opened 2024-09-04
fix(components): Display Datalist checkbox based on device type instead of breakpoint

MichaelParadis:
## Motivations

On touch screen devices that were wider than our `--medium-screens-and-up` the DataList checkbox wouldn't be displayed because the user wouldn't be able to hover over the row, this PR works to fix that.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- components: Display DataList checkboxes on touch screen devices that are medium screen sized and up.

### Security

- 

## Testing

Can be tested in storybook here: https://5607bc22.atlantis.pages.dev/iframe?args=&id=components-lists-and-tables-datalist-web--basic&viewMode=story

Can also be tested in JO, see the PR mentioning this one.

To test enable responsive mode in your browser inspector (this is wider screen with touch simulation enabled):
image

Verify:
 -  When on the medium-screens-and-up size the check box appears when touch simulation is enabled
 - Verify the checkboxes do not appear on the medium-screens-and-up size when touch simulation is disabled
 - Verify smaller breakpoints work as expected when touch simulation is enabled / disabled

Wider screen with touch simulation disabled (hovering on a row):
image

Smaller screen with touch simulation disabled:
image

Smaller screen with touch simulation enabled:
image

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -175,7 +175,7 @@
   transition: opacity var(--transition-properties);
   --transition-properties: var(--timing-quick) ease-in-out;
 
-  @media (--medium-screens-and-up) {
+  @media (--medium-screens-and-up) and (any-pointer: fine) {
     padding-top: 0;
     opacity: 0;
   }
--- a/packages/components/src/DataList/DataList.css
+++ b/packages/components/src/DataList/DataList.css
@@ -175,7 +175,7 @@
   transition: opacity var(--transition-properties);
   --transition-properties: var(--timing-quick) ease-in-out;
 
-  @media (--medium-screens-and-up) and (any-pointer: fine) {
+  @media (any-pointer: fine) {
     padding-top: 0;
     opacity: 0;
   }


chris-at-jobber:
@MichaelParadis one thought - do we actually need a breakpoint query at all? If the user can hover a narrow list, why do we need to show the checkbox permanently?

zevpermack:
Thanks for taking this on!! Tested the prerelease and works to solve what we talked about on quotes and client.
I don't see the checkboxes come up on jobs. I'm guessing that is because they don't have bulk actions yet?

I'm not sure I 100% follow what Chris was saying but I was wondering if `(--medium-screens-and-up)` might be redundant now if we want all pointer cases. I guess the case of smaller breakpoints having the checkboxes available is handled elsewhere. Happy to let your team handle the code review since I don't have a lot of context there though and I'm sure I'm missing something.

Also just curious to learn why you ended up going for `fine` instead of `coarse` for `any-pointer`?

MichaelParadis:
@zevpermack and @chris-at-jobber 

## RE: The `--medium-screens-and-up` comments
I left it as before because the checkbox wouldn't display all the time on small screen sizes without hovering on desktop. I feel like this _could_ lead to some confusion on smaller breakpoints. 

Current change
image

Removing the `medium-screens-and-up` and being on a non-touchscreen device
image


## RE: `any-pointer: fine vs any-pointer: course` I went with `fine` 
The behaviour we are trying to have is have the checkbox hidden when there is a mouse and display it on hover. The media query here is for hiding the checkbox unless it is hovered so using `course`
 would reverse the behaviour (it would show all the time for mice and need to be hovered for it to display when there isn't a mouse (not possible).
[This](https://github.com/GetJobber/atlantis/blob/JOB-97858/datalist-checkbox-based-on-device-pointer/packages/components/src/DataList/DataList.css#L184-L188) is the CSS that changes the opacity when the checkbox is hovered

[any-pointer mdn page](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/any-pointer)

zevpermack:
Loaded up Atlantis repo and confirmed all of that. LGTM, thanks @MichaelParadis

chris-at-jobber:
Thanks @zevpermack ! @MichaelParadis I'm going to chat with @jlelford (the original designer who flagged this) and @diajjiao for their thoughts on the need for a media query vs cursor query.

taylorvnoj (approved):
LGTM - tested in prod also 👍

chris-at-jobber:
Bringing the outcome of the design conversation here:

If we can ensure that in cases where the `cursor` query might fail, we show the checkbox, then we can do away with the breakpoint check.

packages/components/src/DataList/DataList.css
```diff
@@ -175,7 +175,7 @@
   transition: opacity var(--transition-properties);
   --transition-properties: var(--timing-quick) ease-in-out;
 
-  @media (--medium-screens-and-up) and (any-pointer: fine) {
+  @media (any-pointer: fine) {
```
MichaelParadis, line 178: @chris-at-jobber I updated it so that there is no more breakpoint.

Based on mdn docs if the media query isn't support it will return false (aka always be visible)
image
To test this you can rewrite this so it has a typo `@media (an-pointer: fine)` to simulate an unknown media query and the checkbox will always be displayed.
CC: @taylorvnoj

taylorvnoj (approved):
- with mouse, boxes only show up on hover regardless of breakpoint ✅ 
- touch device always shows boxes ✅ 
- if @media isn't supported boxes are always visible ✅

chris-at-jobber (commented):
Design approved!

PR #2019, opened 2024-09-09
feat(components): Add description prop to InputFile

taylorvnoj:
## Motivations

The purpose behind adding `supportText` is to communicate to a user, the permitted file types and max file size when uploading files via `InputFile` dropzone.  This can help in avoiding upload errors that need to be rectified after a failed upload.

### Added

1. `supportText` prop, which is a string. Why a string?
- Although the motivation behind adding this prop is to display file types and size limit for files, I feel that keeping this as customizable as possible, allows for flexibility moving forward
- Allowing for a string also makes localization easier on the consumer side

Screenshot 2024-09-11 at 5 53 46 PM

[Figma](https://www.figma.com/design/rIIhulZvcp9M82lNOCGv16/branch/FRyCZpCX92n6dzD6alwyKj/Product%2FOnline?m=auto&node-id=13630-0&t=JI0ecH7XrhDzgHw6-1)

Why not dynamically use `allowedTypes` here?  
Thinking about instances where consumers are putting `allowedTypes` into an array, if they allow DOCX files, that looks like this: `"application/vnd.openxmlformats-officedocument.wordprocessingml.document"`.  I personally do not think it's worth having all possible MIME types (there's too many) mapped in this component, just to convert them to uppercase and then a string to be displayed with a file size that is set in the consumers component.  This would not only be adding a lot of logic to this component, but I can foresee that becoming brittle. 

2. Added an example of `supportText` to the Variations and Sizes story, as well as a stand alone Support Text story

### Changed

The `hintText` is no longer combined with the Button text.  That sentence format spanning across button and `hintText` will not scale well with various languages in the future.

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -124,9 +124,9 @@ interface InputFileProps {
   readonly allowMultiple?: boolean;
 
   /**
-   * Support text to display in the dropzone.
+   * Further description of the input.
    */
-  readonly supportText?: string;
+  readonly description?: string;
 
   /**
    * A callback that receives a file object and returns a `UploadParams` needed
@@ -186,7 +186,7 @@ export function InputFile({
   buttonLabel: providedButtonLabel,
   allowMultiple = false,
   allowedTypes = "all",
-  supportText,
+  description,
   getUploadParams,
   onUploadStart,
   onUploadProgress,
@@ -247,9 +247,9 @@ export function InputFile({
               {size === "base" && (
                 <>
                   <Typography size="small">{hintText}</Typography>
-                  {supportText && (
+                  {description && (
                     <Typography size="small" textColor="textSecondary">
-                      {supportText}
+                      {description}
                     </Typography>
                   )}
                 </>
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -124,6 +124,11 @@ interface InputFileProps {
   readonly allowMultiple?: boolean;
 
   /**
+   * Support text to display in the dropzone.
+   */
+  readonly supportText?: string;
+
+  /**
    * A callback that receives a file object and returns a `UploadParams` needed
    * to upload the file.
    *
@@ -181,6 +186,7 @@ export function InputFile({
   buttonLabel: providedButtonLabel,
   allowMultiple = false,
   allowedTypes = "all",
+  supportText,
   getUploadParams,
   onUploadStart,
   onUploadProgress,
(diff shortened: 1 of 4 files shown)


packages/components/src/InputFile/InputFile.tsx
```diff
@@ -358,14 +369,14 @@ function getLabels(
 ) {
   let buttonLabel = multiple ? "Upload Files" : "Upload File";
   let hintText = multiple
-    ? "or drag files here to upload"
-    : "or drag a file here to upload";
+    ? "Select or drag files here to upload"
```
taylorvnoj, line 372: I've experimented with a few different ways of writing this function now, since there's so much repetition, but I keep ending up with an equally convoluted, hard to read function.  Any feedback/insight on this would be appreciated!

chris-at-jobber:
@taylorvnoj did you look at similar properties on other components when naming this prop? The two patterns I see that are similar are

`description` or `assistiveText`

- applies supplementary text content to a component that otherwise has primary text to help the user 

image
image

As you can see, typically the `components-native` versions use `assistiveText` while the web versions use `description`, which is an unfortunate discrepancy, but it seems like this prop does something similar and should follow one of those approaches - although if you are referencing another pattern or approach to this, happy to discuss!

taylorvnoj:
@chris-at-jobber great call out - admittedly I was narrow focused on this component and wasn't doing the full homework. SHAME! Renamed to `description` here 7790894e536bd530bf299263c5b6ead2ac031a6d

packages/components/src/InputFile/InputFile.tsx
MichaelParadis: I was able to rewrite it as
```ts
function getLabels(
  providedButtonLabel: string | undefined,
  multiple: boolean,
  allowedTypes: string | string[],
) {
  const fileType =
    allowedTypes === "images" || allowedTypes === "basicImages"
      ? "Image"
      : "File";
  let buttonLabel = multiple ? `Upload ${fileType}s` : `Upload ${fileType}`;
  const fileTypeDeterminer = fileType === "Image" ? "an" : "a";
  const hintText = multiple
    ? `Select or drag ${fileType.toLowerCase()}s here to upload`
    : `Select or drag ${fileTypeDeterminer} ${fileType.toLowerCase()} here to upload`;

  if (providedButtonLabel) buttonLabel = providedButtonLabel;

  return { buttonLabel, hintText };
}
```
This didn't change any snapshots, I feel like it is more readable but that may not be the case

taylorvnoj: Okay that's verrry close to what I had - in some ways I feel like it's a lot but it's still readable, gets the job done and dry's it up. Ok I'll update that

taylorvnoj: Updated here 1d39876ac162b2d64ba3a01338a49e7843c8d5e3
Also just changed one InputFile in Variations and Sizes example to be plural and better fit the description I've set.

PR #2021, opened 2024-09-11
feat(components): Add underline support to Typography component for web (JOB-104497)

sabidhasan:
## Motivations
This PR adds the ability to add underlining to the `Typography` for web components.

This is accomplished by adding an underline prop to the `Typography` component, which is a space separated string consisting of a `style` and optionally, a color. The value for `style` can be set to one of the four allowed values, whereas the color can either be omitted or be set to one of the color values from our color tokens. If the color is omitted from the `underline` prop, it defaults to the text's color itself (whether provided via textColor or inherited naturally via CSS).

## Changes
**Type Safety for Prop**
Full type safety and autocomplete is provided for the new prop:
image

**Browsers**
This is how the functionality looks  in Storybook in the three major browsers:

| Firefox | image |
|---------|---------------------------------------------------------------------------------------------------------------------|
| Chrome  | image |
| Safari  | image |

**Props**
Here are some different variations of the props, changing the style of the underline, the color of the text and when setting and style and color of text:

image
image
image

**Our in app changes using this**

image

<details><summary>Details</summary>
Apply the following diff to see this:

```diff
diff --git a/app/javascript/jobber/jobberAssistant/components/ConversationListItem.module.css b/app/javascript/jobber/jobberAssistant/components/ConversationListItem.module.css
index 7e5e5911b5b..4b2d54081f7 100644
--- a/app/javascript/jobber/jobberAssistant/components/ConversationListItem.module.css
+++ b/app/javascript/jobber/jobberAssistant/components/ConversationListItem.module.css
@@ -24,6 +24,7 @@
 
 .responseContainer {
   display: flex;
+  flex-wrap: wrap;
   box-sizing: border-box;
   padding: 0;
   border: none;
@@ -41,6 +42,7 @@
   display: flex;
   min-width: 0;
   margin-bottom: 0;
+  margin-left: 32px;
   text-align: left;
   flex-direction: column;
   gap: var(--space-base);
diff --git a/app/javascript/jobber/jobberAssistant/components/ConversationListItem.tsx b/app/javascript/jobber/jobberAssistant/components/ConversationListItem.tsx
index ad062fa1c9d..1589d2c0851 100644
--- a/app/javascript/jobber/jobberAssistant/components/ConversationListItem.tsx
+++ b/app/javascript/jobber/jobberAssistant/components/ConversationListItem.tsx
@@ -4,7 +4,7 @@ import { Icon } from "@jobber/components/Icon";
 import { Text } from "@jobber/components/Text";
 import { Emphasis } from "@jobber/components/Emphasis";
 import classNames from "classnames";
-import { Markdown } from "@jobber/components";
+import { Markdown, Tooltip } from "@jobber/components";
 import type {
   AiAssistantAction,
   AiAssistantFollowup,
@@ -119,6 +119,13 @@ const Response = ({ item, showFeedbackOptions }: ConversationListItemProps) => {
           <Icon name={icon} color={iconColor} />
         </div>
       )}
+      <div style={{ flex: 1, marginBottom: 8 }}>
+        <Tooltip message="Copilot may rephrase your question to best use the data available to it">
+          <Text underline="dashed" size="large">
+            How many unique customers had recurring jobs that started in the year 2022?
+          </Text>
+        </Tooltip>
+      </div>
       <div className={styles.responseContentColumn}>
         <Markdown
           onLinkClick={({ href }) => trackExternalLinkOpen(href, "inline")}
```
</details> 

### Added
- Added a new `underline` prop to the `Text` component, allowing it to accept basic values like "solid" or "dashed" values, or compound values with a style and color. 
- Added a new `underline` prop to the `Typography` component, supporting the same values as above

## Testing
- Added a test case for rendering underlined text in the `Text` component.
- Added test cases for rendering different underline styles in the `Typography` component.
- Fixed some minor test description typos.

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Text/Text.test.tsx
+++ b/packages/components/src/Text/Text.test.tsx
@@ -47,7 +47,7 @@ it("renders a success text", () => {
   `);
 });
 
-it("renders an error text", () => {
+it("renders a error text", () => {
   const { container } = render(<Text variation="error">Name is required</Text>);
   expect(container).toMatchInlineSnapshot(`
     <div>
@@ -75,7 +75,7 @@ it("renders a warning text", () => {
   `);
 });
 
-it("renders an info text", () => {
+it("renders a info text", () => {
   const { container } = render(
     <Text variation="info">
       Drag to rearrange the order that fields show up in Jobber
@@ -123,7 +123,7 @@ it("renders a small text", () => {
   `);
 });
 
-it("renders an end-aligned text", () => {
+it("renders a end-aligned text", () => {
   const { container } = render(<Text align="end">End align me</Text>);
   expect(container).toMatchInlineSnapshot(`
     <div>
@@ -144,20 +144,6 @@ it("renders a center-aligned text", () => {
         class="base regular base text center"
       >
         Center align me
-      </p>
-    </div>
-  `);
-});
-
-it("renders underlined text", () => {
-  const { container } = render(<Text underline="dashed">Underline me</Text>);
-  expect(container).toMatchInlineSnapshot(`
-    <div>
-      <p
-        class="base regular base text basicUnderline"
-        style="text-decoration-style: dashed; text-decoration-color: text;"
-      >
-        Underline me
       </p>
     </div>
   `);
--- a/packages/components/src/Text/Text.test.tsx
+++ b/packages/components/src/Text/Text.test.tsx
@@ -47,7 +47,7 @@ it("renders a success text", () => {
   `);
 });
 
-it("renders a error text", () => {
+it("renders an error text", () => {
   const { container } = render(<Text variation="error">Name is required</Text>);
   expect(container).toMatchInlineSnapshot(`
     <div>
@@ -75,7 +75,7 @@ it("renders a warning text", () => {
   `);
 });
 
-it("renders a info text", () => {
+it("renders an info text", () => {
   const { container } = render(
     <Text variation="info">
       Drag to rearrange the order that fields show up in Jobber
@@ -123,7 +123,7 @@ it("renders a small text", () => {
   `);
 });
 
-it("renders a end-aligned text", () => {
+it("renders an end-aligned text", () => {
   const { container } = render(<Text align="end">End align me</Text>);
   expect(container).toMatchInlineSnapshot(`
     <div>
@@ -144,6 +144,19 @@ it("renders a center-aligned text", () => {
         class="base regular base text center"
       >
         Center align me
+      </p>
+    </div>
+  `);
+});
+
+it("renders underlined text", () => {
+  const { container } = render(<Text underline="dashed">Underline me</Text>);
+  expect(container).toMatchInlineSnapshot(`
+    <div>
+      <p
+        class="base regular base text dashed"
+      >
+        Underline me
       </p>
     </div>
   `);
--- a/packages/components/src/Text/Text.test.tsx
+++ b/packages/components/src/Text/Text.test.tsx
@@ -154,7 +154,8 @@ it("renders underlined text", () => {
   expect(container).toMatchInlineSnapshot(`
     <div>
       <p
-        class="base regular base text dashed"
+        class="base regular base text basicUnderline"
+        style="text-decoration-style: dashed; text-decoration-color: text;"
       >
         Underline me
       </p>
--- a/packages/components/src/Typography/Typography.tsx
+++ b/packages/components/src/Typography/Typography.tsx
@@ -10,6 +10,7 @@ import emphasis from "./css/Emphasis.css";
 import truncate from "./css/Truncate.css";
 import alignment from "./css/TextAlignment.css";
 import fontFamilies from "./css/FontFamilies.css";
+import underlineStyles from "./css/Underline.css";
 
 interface TypographyProps {
   readonly id?: string;
@@ -50,6 +51,11 @@ interface TypographyProps {
   readonly fontFamily?: keyof typeof fontFamilies;
   readonly children: ReactNode;
   readonly numberOfLines?: number;
+
+  /**
+   * Style of underline for the text
+   */
+  readonly underline?: "solid" | "double" | "dotted" | "dashed" | undefined;
 }
 export type TypographyOptions = Omit<TypographyProps, "children">;
 
@@ -65,6 +71,7 @@ export function Typography({
   emphasisType,
   numberOfLines,
   fontFamily,
+  underline,
 }: TypographyProps) {
   const shouldTruncateText = numberOfLines && numberOfLines > 0;
   const className = classnames(
@@ -76,6 +83,7 @@ export function Typography({
     emphasisType && emphasis[emphasisType],
     fontFamily && fontFamilies[fontFamily],
     shouldTruncateText && truncate.textTruncate,
+    underline && underlineStyles[underline],
     {
       ...(align && { [alignment[align]]: align !== `start` }),
     },
--- a/packages/components/src/Typography/Typography.tsx
+++ b/packages/components/src/Typography/Typography.tsx
@@ -11,7 +11,7 @@ import truncate from "./css/Truncate.css";
 import alignment from "./css/TextAlignment.css";
 import fontFamilies from "./css/FontFamilies.css";
 import underlineStyles from "./css/Underline.css";
-import { Underline, UnderlineStyle } from "./types";
+import { UnderlineStyle, UnderlineStyleWithColor } from "./types";
 
 interface TypographyProps {
   readonly id?: string;
@@ -61,7 +61,7 @@ interface TypographyProps {
    * @example "solid" for a non-dashed underline of the same color as `textColor`
    * @example "double color-invoice" for a double underline in the specified color
    */
-  readonly underline?: UnderlineStyle | Underline | undefined;
+  readonly underline?: UnderlineStyle | UnderlineStyleWithColor | undefined;
 }
 export type TypographyOptions = Omit<TypographyProps, "children">;
 
--- a/packages/components/src/Typography/Typography.tsx
+++ b/packages/components/src/Typography/Typography.tsx
@@ -125,7 +125,8 @@ function computeUnderlineColor(
   textDecorationColor: string,
   textColor?: keyof typeof textColors,
 ): string | undefined {
-  // Use underline color otherwise if a custom text color is specified, use that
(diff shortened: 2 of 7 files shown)


packages/components/src/Typography/css/Underline.css
```diff
@@ -0,0 +1,19 @@
+.solid {
+  text-decoration: underline solid var(--color-base-grey--300);
+  text-underline-offset: 4px;
```
sabidhasan, line 3: Is there a variable we should be using for this? @jlelford

packages/components/src/Typography/css/Underline.css
```diff
@@ -0,0 +1,19 @@
+.solid {
+  text-decoration: underline solid var(--color-base-grey--300);
```
sabidhasan, line 2: Not sure what the intended colour here should be @jlelford

packages/components/src/Typography/css/Underline.css
jlelford: Are both of the requests here for the underline? If so the comment below would apply, if there is a variation I'm not understanding let me know.

sabidhasan: This is for the distance between the text and the underline (wondering what the 4px should be ).

jlelford: Oh gotcha, the distance looks good in the images you posted above. Do you think that should be the Atlantis 
`--space-smaller` variable rather than specific pixel size?

NickBoers-Jobber (changes requested):
The underline thickness/width differs slightly between Chrome and Firefox. I wonder whether we should explore some additional CSS in `packages/components/src/Typography/css/Underline.css`? For example, `text-decoration-thickness`.

Chrome:

image

Firefox:

image

packages/components/src/Text/Text.test.tsx
```diff
@@ -148,3 +148,17 @@ it("renders a center-aligned text", () => {
     </div>
   `);
 });
+
+it("renders underlined text", () => {
```
NickBoers-Jobber, line 152: I wondered whether it might be worth including a test that combines a colour with the underline style? For example, 

```html
<Text variation="error" underline="dashed color-green" size="large">
  Some horrible error message.
</Text>
```

That said, you have the test for `<Typography>`. Given how it all works, the test is probably quite unnecessary.

packages/components/src/Typography/Typography.tsx
```diff
@@ -76,23 +89,48 @@ export function Typography({
...
       {children}
     </Tag>
   );
 }
+
+function computeUnderlineColor(
+  textDecorationColor: string,
+  textColor?: keyof typeof textColors,
+): string | undefined {
+  // Use underline color otherwise if a custom text color is specified, use that
```
NickBoers-Jobber, line 128: Nit, this comment feels a bit awkwardly worded to me. Totally non-blocking.

NickBoers-Jobber (approved):
LGTM!

packages/components/src/Typography/Typography.tsx
sabidhasan: @NickBoers-Jobber  Updated in 280855c889630f3a9a5a7fb74728162ca4050f70 thanks for pointing this out

packages/components/src/Text/Text.test.tsx
sabidhasan: Yeah I figured since it was just passing the props in, there isn't really a need to test it in both places. To Atlantis reviewers: if you feel this is needed, I can add it. Otherwise, feels like a lower value test to me but happy to be wrong.

ZakaryH:
something I'm still on the fence about, which I believe is the reason we didn't have underline to begin with, is the overlap between underlined text and Links.

both will look almost identical, and you won't be able to tell if something is a link by looking at it. for example, even here on Github you can tell these are links
image

however in our system, they could be underlined blue Typography

ZakaryH:
how might this work with Text/Typography that isn't just a single sentence?

imagine I only want **this part** of a sentence to be underlined

this approach won't enable that, so if people ever want that - I think they'll ignore this prop and use custom css

that doesn't necessarily mean we shouldn't do it at all, but I do think this won't cover a pretty reasonable usage of underlined text

sabidhasan:
> something I'm still on the fence about, which I believe is the reason we didn't have underline to begin with, is the overlap between underlined text and Links.
> 
> both will look almost identical, and you won't be able to tell if something is a link by looking at it. for example, even here on Github you can tell these are links image
> 
> however in our system, they could be underlined blue Typography
> 
> after taking a quick peek in product, the vast majority of places we are adding `text-decoration: underline` is on links

@ZakaryH this is a good question, and I will tag @jlelford to provide some of the design impetus/motivations behind this change because some thought and discussion has been been put into it.

> how might this work with Text/Typography that isn't just a single sentence?

Wouldn't this already be the case for **all** of the other props to Text/Typography as it is? For instance, how could I only italicize a part of the text? Or highlight a part of it? Or bold a part of it? Or change the color for a part?

The way I am thinking about it is that `Typography` is a semantic equivalent for the HTML `p` element. Since one can't underline a part of `p`, we also shouldn't support underlining a part of `Typography`.

chris-at-jobber:
Related: [<Emphasis>](https://atlantis.getjobber.com/?path=/docs/components-text-and-typography-emphasis--docs) is typically what we use to add stylistic emphasis to a portions of a piece of text
image

jlelford:
> something I'm still on the fence about, which I believe is the reason we didn't have underline to begin with, is the overlap between underlined text and Links.

@ZakaryH the concern between straight underline and links is warranted, but what I really wanted as a tool is the "dotted" version of it, which does have associations with presenting tooltip like information in other design systems. It is a method for us to display interactivity within text content that does imply actual navigation.

I'm not sure we have a need for the "other variants" of underline that aren't covered by the existing use of links. Is there something you or @chris-at-jobber would recommend us do to allow us to have the desired effect of the dotted text without opening the door on regular underlines diluting text links?

ZakaryH:
that makes sense having the dotted version to indicate interactivity, though if that's our main priority - that's the one thing we don't have available on Android. are you guys considering following up with adding that at all?

keeping the APIs between mobile and web as consistent as possible is something we're trying to do, so at this point I think we continue adding `underline` to Text/Typography in web

if that wasn't already in place I think I probably would have thought about maybe Emphasis having an interactive variant or something along those lines since I'm really not seeing a need for Text/Typography to support underlines outside of links.

though Emphasis doesn't exist in mobile at all, and we aren't currently able to combine multiple Text/Typography components so that would involved a bit more work to enable that.

sabidhasan:
> Related: [Emphasis](https://atlantis.getjobber.com/?path=/docs/components-text-and-typography-emphasis--docs) is typically what we use to add stylistic emphasis to a portions of a piece of text image

@chris-at-jobber - would it be better to have a new `variation` for emphasis for `dottedUnderline` (name pending of course) or something like that? That approach might address your concerns in constraining the scope of the underlining possible, and address Zak's concerns around links looking like underlines, since only.

Though this approach does open up an issue around the color of the underline, where @jlelford 's intention was to have a different color for the underline vs the text. It also creates a discrepancy between the way underlining is done on mobile and desktop (per the other PR we released last week)

ZakaryH:
I know it's a bit late, since we've already committed to this via the mobile implementation but I'm having doubts about adding it to `Text`.

Typography sure - that one has tons of customizable props

looking at the rest of the props for Text, it feels out of place. also if Text is more like a paragraph, is it reasonable to have an entire paragraph underlined? I worry about readability.

what would be your use case for having underlined Text?

ZakaryH:
actually, I don't think it's too late. the odds of someone else using that prop in the past few days is exceedingly low.

talked with Chris, and what we'd like to do is:

- revert the change to Text in components-native, so the prop no longer exists on Text
- have your usage in mobile swap to use Typography instead
- do the same thing in web components, where only Typography implements it

what do you think about that? I see we're currently using Text in mobile for the instance we needed last week. is there any reason Typography wouldn't work there?

sabidhasan:
> actually, I don't think it's too late. the odds of someone else using that prop in the past few days is exceedingly low.
> 
> talked with Chris, and what we'd like to do is:
> 
> * revert the change to Text in components-native, so the prop no longer exists on Text
> * have your usage in mobile swap to use Typography instead
> * do the same thing in web components, where only Typography implements it
> 
> what do you think about that? I see we're currently using Text in mobile for the instance we needed last week. is there any reason Typography wouldn't work there?

I am OK with that approach - so to confirm: no change goes into `Text`, and we only make changes in  `Typography`.

For mobile, what priority would you prefer on this work? We already have the [mobile cleanup jira ticket](https://jobber.atlassian.net/browse/JOB-104571) and we can do it as part of that, or I can make the change now to introduce revert the Text changes.

In this PR, once I make the change to remove any changes to `Text`, we should be good?

sabidhasan:
@ZakaryH - made the change in this PR to revert changes to `Typography` in 7efb402986f3191c505c86134407f2d34bf4b3ae. I think this should address the concerns for this PR. For the mobile work, we can create a separate PR to scope the changes to mobile - we can either do this as a fast follow to this PR, or if you don't think this is high priority then I can do it when we pull in the aforementioned ticket. 🙏🏾

ZakaryH:
>I am OK with that approach - so to confirm: no change goes into Text, and we only make changes in Typography.

correct. it would only be exposed by Typography, which is the component that we allow all sorts of customization on vs Text which is pretty locked down with minimal options for manipulating the appearance outside of a few presets

>For mobile, what priority would you prefer on this work? We already have the [mobile cleanup jira ticket](https://jobber.atlassian.net/browse/JOB-104571) and we can do it as part of that, or I can make the change now to introduce revert the Text changes.

I suppose it depends how soon you're thinking you'll be on the follow up work? be great to have the revert sooner than later to minimize the possibility someone starts using the prop on Text in mobile.

I'd say medium priority. the odds of someone using the prop are pretty low, but it would technically be a breaking change so keeping the window of opportunity as small as we can is preferable.

>In this PR, once I make the change to remove any changes to Text, we should be good?

yup! I'll give it one quick read over, but I can't think of anything else pressing.

ZakaryH (approved):
LGTM though I do have one request

could you please update the PR title to `Typography`?

our tooling will use the PR title in the changelog

nevermind, I forgot we can edit other people's PR titles

PR #2022, opened 2024-09-11
feat(components-native): Add underline support to Text/Typography components for mobile (JOB-104505)

sabidhasan:
## Motivations
This PR adds the ability to add underlining to the Text component for the native-components.

## Visual Changes 
**Android**
image

image

**iOS**
image

## Discussion
The underline styles here (dashed, etc) are not implemented yet on the Android side. For this reason Android defaults to the basic underline style regardless of what value is passed in. For iOS, the proper dashed style should be displayed.

### Added
- Added `underline` prop to `TextProps` with options `"solid"` or `"dashed"`
- Added `underline` prop to `TypographyProps` with options `"solid"`, `"double"`, `"dotted"`, or `"dashed"` (I covered more than what is exposed from Text, because this is an internal component; if the reviewer thinks this is unnecessary, I can change it to support only the solid/dashed cases that we need)
- Added underline styles to `typographyTokens`

## Testing
To test locally, run `npm run pack @jobber/components-native` and then install into the mobile repo via `npm i /path/to/atlantis/jobber-components-native-x.x.x.tgz`. Then change a `Text` component to pass an underline prop value.

For the screenshots above, I used the branch `JOB-103929-jobber-mobile-implement-explainability-ui` at commit `2c1c74b7c4c9f7b5ae569f4121c3505968e9bed3` and just locally installed the version of Atlantis that was packed from the command above.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/Text/Text.tsx
+++ b/packages/components-native/src/Text/Text.tsx
@@ -73,7 +73,7 @@ interface TextProps
   /**
    * Underline text
    */
-  readonly underline?: "solid" | "dashed";
+  readonly underline?: "solid" | "dotted";
 
   /**
    * This will make the text inaccessible to the screen reader.
--- a/packages/components-native/src/Text/Text.tsx
+++ b/packages/components-native/src/Text/Text.tsx
@@ -71,7 +71,9 @@ interface TextProps
   readonly italic?: boolean;
 
   /**
-   * Underline text
+   * Underline style to use for the text. The non-solid style is only supported
+   * on iOS, as per React Native's Text component's limitations.
+   * https://reactnative.dev/docs/text-style-props#textdecorationstyle-ios
    */
   readonly underline?: "solid" | "dotted";
 
--- a/packages/components-native/src/Text/Text.tsx
+++ b/packages/components-native/src/Text/Text.tsx
@@ -71,6 +71,11 @@ interface TextProps
   readonly italic?: boolean;
 
   /**
+   * Underline text
+   */
+  readonly underline?: "solid" | "dashed";
+
+  /**
    * This will make the text inaccessible to the screen reader.
    * This should be avoided unless there is a good reason.
    * For example this is used in InputText to make it so the label isn't
@@ -118,6 +123,7 @@ export function Text({
   italic = false,
   hideFromScreenReader = false,
   maxFontScaleSize,
+  underline,
   selectable,
 }: TextProps): JSX.Element {
   const accessibilityRole: TextAccessibilityRole = "text";
@@ -130,6 +136,7 @@ export function Text({
       fontWeight={getFontWeight({ level, emphasis })}
       maxFontScaleSize={maxFontScaleSize || TEXT_MAX_SCALED_FONT_SIZES[level]}
       selectable={selectable}
+      underline={underline}
       {...{
         ...levelStyles[level],
         allowFontScaling,
--- a/packages/components-native/src/Typography/Typography.style.ts
+++ b/packages/components-native/src/Typography/Typography.style.ts
@@ -88,6 +88,11 @@ export const typographyTokens: { [index: string]: TextStyle } = {
     fontStyle: "italic",
   },
 
+  underline: {
+    textDecorationColor: tokens["color-base-grey--300"],
+    textDecorationLine: "underline",
+  },
+
   startAlign: {
     textAlign: "left",
   },
--- a/packages/components-native/src/Typography/Typography.style.ts
+++ b/packages/components-native/src/Typography/Typography.style.ts
@@ -89,7 +89,7 @@ export const typographyTokens: { [index: string]: TextStyle } = {
   },
 
   underline: {
-    textDecorationColor: tokens["color-base-grey--300"],
+    textDecorationColor: tokens["color-text--secondary"],
     textDecorationLine: "underline",
   },
 
--- a/packages/components-native/src/Typography/Typography.tsx
+++ b/packages/components-native/src/Typography/Typography.tsx
@@ -85,7 +85,9 @@ export interface TypographyProps<T extends FontFamily>
   readonly fontStyle?: T extends "base" ? BaseStyle : DisplayStyle;
 
   /**
-   * Style of underline for the text
+   * Underline style to use for the text. The non-solid style is only supported
+   * on iOS, as per React Native's Text component's limitations.
+   * https://reactnative.dev/docs/text-style-props#textdecorationstyle-ios
    */
   readonly underline?: "solid" | "double" | "dotted" | "dashed" | undefined;
 
--- a/packages/components-native/src/Typography/Typography.tsx
+++ b/packages/components-native/src/Typography/Typography.tsx
@@ -7,6 +7,7 @@ import {
   // eslint-disable-next-line no-restricted-imports
   Text,
   TextProps,
+  TextStyle,
   ViewStyle,
 } from "react-native";
 import { TypographyGestureDetector } from "./TypographyGestureDetector";
@@ -84,6 +85,11 @@ export interface TypographyProps<T extends FontFamily>
   readonly fontStyle?: T extends "base" ? BaseStyle : DisplayStyle;
 
   /**
+   * Style of underline for the text
+   */
+  readonly underline?: "solid" | "double" | "dotted" | "dashed" | undefined;
+
+  /**
    * Font weight
    */
   readonly fontWeight?: T extends "base" ? BaseWeight : DisplayWeight;
@@ -124,6 +130,7 @@ const maxNumberOfLines = {
 
 export const Typography = React.memo(InternalTypography);
 
+// eslint-disable-next-line max-statements
 function InternalTypography<T extends FontFamily = "base">({
   fontFamily,
   fontStyle,
@@ -143,6 +150,7 @@ function InternalTypography<T extends FontFamily = "base">({
   hideFromScreenReader = false,
   accessibilityRole = "text",
   strikeThrough = false,
+  underline,
   selectable = true,
 }: TypographyProps<T>): JSX.Element {
   const sizeAndHeight = getSizeAndHeightStyle(size, lineHeight);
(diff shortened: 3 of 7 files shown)


jlelford (commented):
Adding the underline to the text looks good, but it appears that the we've changed the spacing between the top of the message bubble and the first line of text? Is it because the react native text is missing padding that the message previously had?

Minor nit, the "Copilot may rephrase..." text doesn't need a period.

sabidhasan:
> Adding the underline to the text looks good, but it appears that the we've changed the spacing between the top of the message bubble and the first line of text? Is it because the react native text is missing padding that the message previously had?
> 
> Minor nit, the "Copilot may rephrase..." text doesn't need a period.

I wouldn't worry about that - that is coming from the container styles, so shouldn't be related to the text itself I don't think.

jlelford:
> I wouldn't worry about that - that is coming from the container styles, so shouldn't be related to the text itself I don't think.

So it shouldn't be concerned that the spacing appears to be inconsistent in the Copilot response? (Pixel measurements are just to indicate they're the same size squares in each message)
Screenshot 2024-09-11 at 9 56 38 AM

sabidhasan:
> > I wouldn't worry about that - that is coming from the container styles, so shouldn't be related to the text itself I don't think.
> 
> So it shouldn't be concerned that the spacing appears to be inconsistent in the Copilot response? (Pixel measurements are just to indicate they're the same size squares in each message) Screenshot 2024-09-11 at 9 56 38 AM

Correct, the spacing there is set on the Markdown component's container, and for this screenshot I didn't do that; here it is with the margin set. Actually in looking at this a bit more, I discovered we are probably setting the wrong margin on the container as-is: [the library](https://github.com/iamacup/react-native-markdown-display/blob/fc9c5e1401729e30807a5781d8606e6a1054e158/src/lib/styles.js#L174) sets the paragraph margins to 10px and I assume we want 8 or 12?

Regardless, if we set them to 10 here, we get the same spacing:

image

jlelford:
> Correct, the spacing there is set on the Markdown component's container, and for this screenshot I didn't do that; here it is with the margin set. Actually in looking at this a bit more, I discovered we are probably setting the wrong margin on the container as-is: [the library](https://github.com/iamacup/react-native-markdown-display/blob/fc9c5e1401729e30807a5781d8606e6a1054e158/src/lib/styles.js#L174) sets the paragraph margins to 10px and I assume we want 8 or 12?

Ideally yes, would that then impact 2way sms? I'm a bit hesitant to add more scope here changing and reviewing this though given our timelines.

sabidhasan:
> > Correct, the spacing there is set on the Markdown component's container, and for this screenshot I didn't do that; here it is with the margin set. Actually in looking at this a bit more, I discovered we are probably setting the wrong margin on the container as-is: [the library](https://github.com/iamacup/react-native-markdown-display/blob/fc9c5e1401729e30807a5781d8606e6a1054e158/src/lib/styles.js#L174) sets the paragraph margins to 10px and I assume we want 8 or 12?
> 
> Ideally yes, would that then impact 2way sms? I'm a bit hesitant to add more scope here changing and reviewing this though given our timelines.

No nothing we do would impact 2WSMS - we are configuring gifted chat separately from them.

packages/components-native/src/Typography/Typography.style.ts
```diff
@@ -88,6 +88,11 @@ export const typographyTokens: { [index: string]: TextStyle } = {
     fontStyle: "italic",
   },
 
+  underline: {
+    textDecorationColor: tokens["color-base-grey--300"],
```
chris-at-jobber, line 92: This needs to be a semantic token - we shouldn't be specifying base values anywhere in an Atlantis component unless there's a very specific case where we don't want it to respond to theming

sabidhasan: Would `tokens["color-text--secondary"]` be acceptable @chris-at-jobber ?

chris-at-jobber: What is the expected experience if a consumer is using a variation of Typography? (ie `critical`); do we expect that this color should respond to the color of the text itself?

chris-at-jobber: Another example would be places where we are implementing a "link-looking" element (which some teams have done on their own with custom solutions), the underline uses `color-interactive` to match the text element.

chris-at-jobber: @sabidhasan I would try 

```suggestion
    textDecorationColor: tokens["color-border"],
```

as a starting point

sabidhasan: I'll let @jlelford answer here, as the designs explicitly do _not_ use the same color for the underline as they do for the text.

jlelford: @chris-at-jobber Abid was trying to find a token I was using in Figma that doesn't seem to align correctly in Atlantis. I've since changed the request to be `var(--color-text--secondary)` 

FYI I was originally using `border--interactive` which seems to be much darker in Figma than what is listed in Atlantis `var(--color-base-blue--300)`

jlelford: It took me awhile to reply so I missed this conversation, I intentionally had the dotted line a different colour from the text in the desktop version in an effort to differentiate the action as a tooltip/popover instead of a regular link.

I'd used [Polaris' tooltip](https://polaris.shopify.com/components/overlays/tooltip?example=tooltip-with-underline) underline as inspiration.

Wouldn't `color-border` be difficult to differentiate on top of the already light grey message background here?

chris-at-jobber:
@jlelford and I discussed "offline" regarding [the desired experience for consumers regarding setting underline color](https://github.com/GetJobber/atlantis/pull/2022#discussion_r1755587326)

Because there are presently no usages of the underline property as it's a net-new, I'm OK with the hard-coded color approach we're using here but I believe the use case in this PR (underline color ≠ text color) is likely the minority use case, and this PR should be followed with
- a new default that changes the underline color to match the specified `Typography` color
- adding the ability for `Typography` consumers to specify the underline color 
  - (ideally/potentially as an addition to the `underline` property, like `underline: dashed interactive--subtle` so that we're not adding another prop that Android is not even aware of)
- then updating the Copilot usages to leverage `Typography` with custom underline color

sabidhasan:
> @jlelford and I discussed "offline" regarding [the desired experience for consumers regarding setting underline color](https://github.com/GetJobber/atlantis/pull/2022#discussion_r1755587326)
> 
> Because there are presently no usages of the underline property as it's a net-new, I'm OK with the hard-coded color approach we're using here **as a means to unblock**, but I believe the use case in this PR (underline color ≠ text color) is likely the minority use case, and this PR should be followed with
> 
> * a new default that changes the underline color to match the specified `Typography` color
> * adding the ability for `Typography` consumers to specify the underline color
>   
>   * (ideally/potentially as an addition to the `underline` property, like `underline: dashed interactive--subtle` so that we're not adding another prop that Android is not even aware of)
> * then updating the Copilot usages to leverage `Typography` with custom underline color

@chris-at-jobber a ticket for this work has been made here: https://jobber.atlassian.net/browse/JOB-104571

packages/components-native/src/Text/Text.tsx
```diff
@@ -70,6 +70,11 @@ interface TextProps
    */
   readonly italic?: boolean;
 
+  /**
+   * Underline text
```
ZakaryH, line 74: if the `dotted` value isn't supported/implemented on Android, at the very least we should call that out in the JSDocs here

I think it's reasonable for a consumer to assume that it works on both platforms

packages/components-native/src/Typography/Typography.tsx
```diff
@@ -83,6 +84,11 @@ export interface TypographyProps<T extends FontFamily>
    */
   readonly fontStyle?: T extends "base" ? BaseStyle : DisplayStyle;
 
+  /**
+   * Style of underline for the text
```
ZakaryH, line 88: same thing here

if Android doesn't support any of these values it will appear to offer let's call that out at the very least

sabidhasan: Addressed in e9900e023d046db9397ce4e7f083b93e9b47e7b3

packages/components-native/src/Text/Text.tsx
sabidhasan: Addressed in e9900e023d046db9397ce4e7f083b93e9b47e7b3

garrett-octopusapp (approved):
LGTM, but I do notice that the Text component accepts the same values for `underline` as the Typography component, even though they don't appear on the type. I dunno how big of a deal that is.


https://github.com/user-attachments/assets/97abdc47-f36a-49b4-8f46-b55373b7df90

On Android, every style just looks like `solid`.

packages/components-native/src/Typography/Typography.tsx
```diff
@@ -83,6 +84,13 @@ export interface TypographyProps<T extends FontFamily>
    */
   readonly fontStyle?: T extends "base" ? BaseStyle : DisplayStyle;
 
+  /**
+   * Underline style to use for the text. The non-solid style is only supported
```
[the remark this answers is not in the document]
garrett-octopusapp: Confirmed that it just appears as "solid" on Android, as does every defined style.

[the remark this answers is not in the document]
sabidhasan: Yeah anything except solid is unsupported on Android.

sabidhasan:
> LGTM, but I do notice that the Text component accepts the same values for `underline` as the Typography component, even though they don't appear on the type. I dunno how big of a deal that is.

Might an IDE or language server problem, but the type specified for text is not the same as Typography, and at least for me I only see the types I should:

image

MichaelParadis (approved):
LGTM!

Tested on iOS and Android.

Non-blocking but it may be worth covering the `strike-through` with an underline case in the props cleanup ticket.

When the strike-through prop is used it will add a `textDecorationLine` from React Native [see](https://github.com/GetJobber/atlantis/blob/master/packages/components-native/src/Typography/Typography.style.ts#L387-L389). We may want to support both underline and strike-through by passing `underline line-through` See React Native [Docs](https://reactnative.dev/docs/0.70/text-style-props#textdecorationline)

PR #2035, opened 2024-09-18
feat(components): Re-separate Chip from Chips w/Background Color Fixes

ZakaryH:
## Motivations

Bringing back the changes we tried to make to separate Chip from Chips in another PR that we reverted due to some background color issues with an instance of `Chips` using `Chips/Chip` as the children.

very original PR: https://github.com/GetJobber/atlantis/pull/1694
PR where we actually merged the changes: https://github.com/GetJobber/atlantis/pull/1995

### Added

the ability to override the appearance of a `Chips` "active" `Chip` with

- `--public-chip-base-active-bg-color` for the resting background color
- `--public-chip-base-active-hover-bg-color` for the hover/focus color (issue with focus styles right now though)
- `--public-chip-base-active-content-color` for the text and separator line that can appear on a Chip

question: do we need one more for the active content in a hover state?

also, is the name too wordy? it was originally `chip-base-bg-color` etc. but I thought that was somewhat unintuitive in the context of `Chip` where it will not style the unselected chips, only the selected/active chips will be overridden. this is because under the hood they are the "base" variant.

it's a bit awkward because a standalone `Chip` it doesn't have anything to do with being active at all. it's because `Chips` is now using standalone `Chip` within `InternalChip`, and there it is being set as a `base`.

I could be convinced to simply go with `chip-base-bg-color` since it **does** affect the standard `base` `Chip` . I simply thought it was confusing, or expecting consumers to know too much about the internal implementation within the `Chips` context.

### Changed

it might look like a lot, however most of it is from the other PR. the only new additions in this PR are:
- fixing the cursor for `disabled` divs that are rendered when there's no `onClick`
- removing unused classnames from `InternalChip.css`
- adding 3 `--public` CSS vars for people to change the `active` (technically `base`) appearance
- making the `Chips` unselected `Chip` be `subtle` and the active ones `base`
- changing how we're using CSS vars in `Chip.css` to hopefully make it a bit cleaner

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

same testing as https://github.com/GetJobber/atlantis/pull/1995

on top of that, try to set up some `Chips` with overrides eg.

```
import { Chips, Chip } from "@jobber/components/Chips";

...
// try all types
<Chips 
  type="dismissible"
>
  <Chip label="whatever" value="whatever" />
  <Chip label="invalid" value="invalid" invalid />
  <Chip label="disabled" value="disabled" disabled />
  <Chip label="selected" value="select me" />
</Chips>

```

and wrap that instance in something that does

```
      <div
        style={
          {
            "--public-chip-base-active-bg-color":
              "var(--color-interactive--subtle)",
            "--public-chip-base-active-hover-bg-color":
              "var(--color-interactive--subtle--hover)",
            "--public-chip-base-active-content-color": "var(--color-base-white)",
          } as React.CSSProperties
        }
      >
        ...
    </div>

```

or you can do it with a proper stylesheet too. probably good to try that too. assuming you scope it adequately, you should be able to provide 2 different sets of overrides to 2 different Chips implementations on the same page 

then, when you select a Chip - it should apply the overrides on that active chip. the other should remain untouched, and disabled/invalid also doesn't change. unselected chips should have the `subtle` appearance/variant.

for example
image

the truncation should all still work as intended with the new setup mirroring the background color of the chips as they change.

`disabled` Chips that don't have an `onClick` get rendered as `divs` and so we can't rely on `cursor: auto` on them anymore. both `button` and `div` `Chip` when disabled should have just the little arrow cursor.

Changes can be

---

Random photo of Atlantis


ZakaryH:
last commit is a bit of cleanup we should have done in the Separate Chip from Chips PR when we started using standalone Chip

https://github.com/GetJobber/atlantis/pull/1995/files#diff-94e5c37e1604458d3cff18d0ffd0612251c263234bb46b8b72d0a307c95e86fbL21-L37
image


a bunch of css rules and classnames were left in the InternalChip.css file and nothing was referencing them so I removed all of the unused rules and selectors that cannot happen anymore

there are still a few other places referencing `icon` `button` `wrapper` and `input` so I've kept those

image

ZakaryH:
last commit changes how the background, border and hover variations of those are set

instead of setting it multiple times in each variant eg.

```
.subtle { background-color: var(--some-color); }
```

we're going to have a single `border`, and `background-color` in the `.chip` class that refers to a `--public` var for their colors, each having their own. we were never changing the border width or type, only the color so that's the only variable that each variant will manipulate.

then, in a single hover/focus-visible selector, we have another `border` and `background-color` - which will have priority over the base with the higher specificity

lastly, the concept of the gradient variables became redundant. there was no point in tracking that separately. the whole purpose of the gradient is to match the background color, so now it simply mirrors whatever the background is set to with the public variables used to change the hover/non-hover background color

I think this is much simpler, and hopefully easier to follow. `hover` and `focus-visible` selectors will only modify the `hover` public variable. non-hover selectors will modify the border color and non-hover background public variable.

this does raise the question of if there should be a border-hover equivalent to allow changing that in the same way you can now change the background color on hover.

one drawback to this approach is that since everything looks at the public color, you would not be able to surgically specify one color for a subtle Chip, and then a different override for an invalid Chip. if that's something we care about, then we'd need to add public vars for each and change things up a bit.

ZakaryH:
noticed that for single & multiselect `Chips` on this branch if you focus on the chip it has a different appearance from if you hover on it. this isn't exactly what we want. I was hoping the fix would be to merge the `focus-visible` and `hover` styles, however it's a bit more complicated than that.

it's a result of the separation PR, where we are giving the `InternalChip` for those two variants a `tabIndex={-1}`. this is causing it to not be focusable. you can see the `dismissible` AKA `selection` variant doesn't exhibit this behavior because it is focusable.

it seems to be a descendant that is getting focused, and we had to set the tabIndex for some other issue I recall reading a note about. this makes it hard to set the color of the parent.

docs/components/Chips/Chips.stories.mdx
```diff
@@ -144,3 +144,34 @@ colors for parity with Jobber's mobile platform, but we will not be implementing
 this is for the web components as of this first iteration of our web Chip. The
 mobile use-case is very specific to a filtering use case that we as of now do
 not anticipate will be required for the web.
```
chris-at-jobber, line 146: L142-146 are now out of date and should either be removed, or reflect the new functionality being added in this PR

docs/components/Chips/Chips.stories.mdx
```diff
@@ -144,3 +144,34 @@ colors for parity with Jobber's mobile platform, but we will not be implementing
...
+
+### Chips Composition with Standalone Chip (Web)
```
chris-at-jobber, line 148:
```suggestion
### Chips composition with standalone Chip (Web)
```

Casing should follow our [content guidelines](https://atlantis.getjobber.com/?path=/docs/content-formatting--docs#capitalization)

docs/components/Chips/Chips.stories.mdx
```diff
@@ -144,3 +144,34 @@ colors for parity with Jobber's mobile platform, but we will not be implementing
...
+
+Standalone Chip's `role`, `<Chip.Prefix>`, `<Chip.Suffix>` `heading` and
```
chris-at-jobber, line 150: The point of this section feels a bit fuzzy, it contains

- a list of features that don't work
- an example of how to work around one of those features
- an example of how not to use one of the features we've already told you won't work

Could this be a clearer "Do/Don't"? We have examples in our [frontend style guide](https://atlantis.getjobber.com/?path=/docs/guides-frontend-styleguide--docs#spacing-margins-and-padding).

ZakaryH: I can give the examples a do/don't heading but they wouldn't change beyond that tbh

there is no equivalent for role, suffix, heading nor variation so there could only be a "don't" section for those. `Chips` simply doesn't support those props that standalone Chip offers :/

overall the goal of this section is to try to provide a minimal amount of guidance of what to do when you try to import `Chip` to use within a `Chips`

image

the answer of which you should use is "it depends". if you need a `prefix` you must use `Chips/Chip`. 

if you don't, you should probably use the new standalone `Chip` however you should be aware that while it is valid to give `Chip` various props, they will do nothing in the context of `Chips`

it's a bunch of non obvious stuff that we don't have documented anywhere, so that's what this is.

docs/components/Chips/Chips.stories.mdx
ZakaryH: right. yeah I can delete these.

the notes about the CSS overrides wouldn't really belong here. they are being applied to standalone Chip, so I'll make note of them over in the stories for that component.

though that is somewhat awkward, since we know the case where someone wants to override this **is** a `Chips`. looking... maybe I'll write them there and then link to them from `Chips`

with that in mind, I feel like I should remove the `active` word from the variable. that's only applicable in the `Chips` context, but it belongs to `Chip` generically.

packages/components/src/Chip/Chip.css
```diff
@@ -129,31 +128,32 @@
 */
 
 .subtle:focus-visible {
-  --chip--gradient-color-variation--hover: var(--color-surface);
+  /* this one looks suspicious */
+  /* why does this not match the hover style? */
```
nad182, line 132: Were you able to find the answer? In the [original Figma](https://www.figma.com/design/rIIhulZvcp9M82lNOCGv16/branch/S6J5803Pjmy4VvpsQx0o8S/Product%2FOnline?node-id=18944-3858&node-type=frame&m=dev) it looks like the bg colour in hover and focus states should be the same. I didn't find any comments in previous PRs that modified this file that would explain why "subtle:focus-visible" is getting a different styling compared to ALL other chip variants.

I looked around (A LOT actually) and from what I can tell, there might be chance that `.subtle:focus-visible` background-color was set incorrectly 11 months ago in [this PR](https://github.com/GetJobber/atlantis/pull/1593/files#diff-5dd56e5b5ba5cd1980ca822ded40f143927988157353a03ea46d24e70e41d84fR111-R112) 

1. It looks like the focus state wasn't styled in the [original PR by Scott](https://github.com/GetJobber/atlantis/pull/1571/files#diff-5dd56e5b5ba5cd1980ca822ded40f143927988157353a03ea46d24e70e41d84fR89-R96).
2. In the following PR by Zak (the one I mentioned above), there was a new CSS rule added to style `.subtle:focus`, however, for some reason the background-color CSS variable was assigned a value different from the one in `.subtle:hover` block.

So I think that the incorrect colour could have been set in [this commit](https://github.com/GetJobber/atlantis/pull/1593/commits/fe7d93738b61725d6bc0c5f8a229b6a3a772faa0#diff-5dd56e5b5ba5cd1980ca822ded40f143927988157353a03ea46d24e70e41d84fR108-R111) and was potentially missed during testing.
My opinion is that the hover and focus bg colours should be the same. Perhaps, it would be great if @chris-at-jobber provided his opinion on this.

nad182 (changes requested):
I reviewed about 30% of the changes, but decided to post the comments I have so far before finishing the review, in case you'd like to address them sooner.

packages/components/src/Chip/Chip.css
ZakaryH: yeah it was 2 things. we made some changes 11 months ago that might have introduced some divergence, though I think that might have been intentional, and then with the separate-chip-from-chips PR, the focus visible broke altogether and we can't really use that selector when the Chip is rendered as a `div`, which is exactly what happens in multi/single select chips since the `onClick` isn't actually passed to the Chip, it's on some element above it.

https://github.com/GetJobber/atlantis/pull/2035#issuecomment-2369910585

packages/components/src/Chip/Chip.css
```diff
@@ -82,45 +96,30 @@
...
 
 .subtle {
-  border: var(--border-base) solid var(--color-border--interactive);
-  background-color: var(--color-surface);
-  --chip--gradient-color-variation: var(--color-surface);
+  --chip-border-color: var(--color-border--interactive);
+  --chip-base-bg-color: var(--color-surface);
 }
 
 /* CAUTION: Order is important here since
```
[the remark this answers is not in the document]
ZakaryH: ah, yeah I can fix the typo

the comment might not even be valid anymore though. previously I think we wanted 2 different states, I'm not so sure that's true anymore. also, it's not working anyway 😭

packages/components/src/Chip/Chip.css
```diff
@@ -1,28 +1,42 @@
...
-  padding: var(--space-small) var(--base-unit);
-  border: var(--border-base) solid transparent;
+  box-sizing: border-box;
+  padding: 0 var(--base-unit);
+  border: var(--border-base) solid var(--chip-border-color);
   border-radius: var(--chip-radius);
   color: var(--color-heading);
   text-align: left;
+  background-color: var();
+  background-color: var(--chip-base-bg-color);
```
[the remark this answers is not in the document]
ZakaryH: it most definitely is. good catch, I will get rid of it.

ZakaryH:
@nad182 

>I reviewed about 30% of the changes, but decided to post the comments I have so far before finishing the review, in case you'd like to address them sooner.

wanted to mention that you're definitely not expected to review all of the revert code again. it has been approved once already https://github.com/GetJobber/atlantis/pull/1995

if you're finding it hard to separate things, my commits have been relatively atomic - and the new things are everything after the first revert PR

you're still welcome to take a look at the other stuff, and I will say I have noticed a few things we missed on the first review, however it's not required

packages/components/src/Chip/Chip.tsx
```diff
@@ -41,19 +42,21 @@ export const Chip = ({
     label,
     heading,
   );
+  const Tag = onClick ? "button" : "div";
```
nad182, line 45: I realize that this might be out of scope of this PR, since it's only trying to re-revert the original `Chip` PR. However, I have some accessibility concerns with using `button` as the container for `Chip`.
[According to W3C](https://www.w3.org/TR/2011/WD-html5-20110525/the-button-element.html), the button can't have `div` elements as its descendants, only "phrasing content". However, right now this is not the case:
image

I think the safer approach from the accessibility standpoint would be to always use `div role="button"` and change internal elements (prefix, content, suffix) to be buttons (if they don't have any descendants) or other `div`s, `span`s, `p`s etc.

I'm also not quite sure whether it's currently the case, but there's a chance that the parent component that would use `Chip` instances could also be a `button` element itself. In which case, my suggestion with using `div` as the container for `Chip` will also break accessibility, meaning that it will probably need to be a `<span role="button">`.

If you feel like this is not the best time and place to refactor this, I will totally understand. But at least I think this would be worth logging it somewhere as future improvement.

packages/components/src/Chips/InternalChip.tsx
```diff
@@ -1,60 +1,34 @@
-import React, { useState } from "react";
-import classnames from "classnames";
-import styles from "./InternalChip.css";
+import React from "react";
```
nad182, line 1: This is just a question, not a change request. Why are we importing React? We're on version ^18, so technically it's not necessary (after version 17).

ZakaryH: it's an old component, so I'm sure it's from when that was necessary to include.

though my linter gets very upset if we don't import React in a TSX file
image

if we truly don't need it, and this linter is incorrectly flagging something that is fine, we would need to update our linting rules

nad182: Yep, that seems right. We don't need it as of React v17. 
As for the linter rule, if we add `"react/react-in-jsx-scope": "off",`, this would remove the errors.
image

Or instead we could add the `"plugin:react/jsx-runtime"` in the `extends` block.
image

Both will do the trick. But like I said that was just a question and requesting this change would be outside of the scope of this ticket.

nad182 (changes requested):
This might be unrelated to your changes, but since it's in the same area, I think it would be nice to fix it in this PR:
Selectable chip icons are not correct colour in dark mode:
image
image

You cannot even see the "plus" icon in Dark Mode:
image

If you don't wish to do it now, let me know and I will log the bug ticket.

packages/components/src/Chip/Chip.css
```diff
@@ -129,31 +128,32 @@
...
-  border: var(--border-base) solid var(--color-disabled--secondary);
+  --chip-border-color: var(--color-disabled--secondary);
   color: var(--color-disabled);
-  background-color: var(--color-disabled--secondary);
-  cursor: auto;
-  --chip--gradient-color-variation: var(--color-disabled--secondary);
-  --chip--gradient-color-variation--hover: var(--color-disabled--secondary);
+  --chip-base-bg-color: var(--color-disabled--secondary);
+  --chip-base-hover-bg-color: var(--color-disabled--secondary);
+  cursor: default;
```
nad182, line 156: This is more of a design question, but should this, perhaps, be `cursor: not-allowed`? We are using this typo of cursor for disabled buttons, checkboxes and radio and switches.

nad182 (commented):
I looked at the testing instructions, and I have some thoughts.
You mentioned to "on top of that, try to set up some Chips with overrides eg.", as well as testing it "with a proper stylesheet".
I was wondering, would it make sense to do the following 2 things:
1. Add a storybook story, which would provide the necessary setup for testing overrides and verifying that unselected chips have the subtle appearance/variant;
2. Add an automated test(s) to do just that. Like render the setup above, trigger user event and test that appropriate style/props are applied.

packages/components/src/Chip/Chip.css
ZakaryH: good question

right now on master with the old `Chips` setup, if you disable a chip you still get a `cursor: pointer` which is probably not the right thing to do

image

packages/components/src/Chip/Chip.tsx
ZakaryH: that's a fair callout. divs as descendants isn't correct for a button.

I think this has changed quite a bit in the past year, and even if we did change `chipContent` and other internal elements into `span`s or something since we have `Chip.Prefix` open to be whatever, we cannot guarantee that whatever is provided will be semantically valid. 

it probably does make sense to have it be a div all the time. that said, unless it's as simple as making it a div and everything still works - I'd be hesitant to make that change with this PR. I assume there's a reason we want it to be a button sometimes.

>I'm also not quite sure whether it's currently the case, but there's a chance that the parent component that would use Chip instances could also be a button element itself. In which case, my suggestion with using div as the container for Chip will also break accessibility, meaning that it will probably need to be a <span role="button">.

I'd argue that's possible with every component. you could put any component inside a button if you wanted to, and it would become semantically invalid. I don't think we should worry about that when building components unless we reasonably expect people to put them inside a button. since Chip has an `onClick` already, I would say wrapping it in a button isn't a reasonable thing to do.

ZakaryH:
>This might be unrelated to your changes, but since it's in the same area, I think it would be nice to fix it in this PR:
Selectable chip icons are not correct colour in dark mode:

we should fix that! ..... just not in this PR 😅  that's happening right now on `master` so not a result of this PR.

packages/components/src/Chip/Chip.css
ZakaryH: @chris-at-jobber how important would you say this is to have the subtle focus state match the hover state?

fixing it for Chip, isn't too bad. we just choose which appearance we want and apply it in both cases

fixing it for Chips is a bit more. I have a not that ideal solution with duplicating CSS.

if we're not overly worried about it, I can leave it be.

packages/components/src/Chip/Chip.css
ZakaryH: @chris-at-jobber any preference for the cursor style on disabled `Chip`s?

packages/components/src/Chip/Chip.css
chris-at-jobber: It would be preferred, I feel like a bit of extra CSS is worth standardizing the experience

packages/components/src/Chip/Chip.css
chris-at-jobber: `not-allowed` would be preferred!

packages/components/src/Chip/Chip.css
ZakaryH: ok, got them matching for normal Chip

then added the less-than-ideal copied styles to Chips > InternalChip.css which will keep them looking the same

the TL;DR of the problem is that we have `focus-visible` selectors and styles in `Chip.css` however in multi and single select `Chips` it's actually a hidden `input` that we are focusing, so it doesn't trigger the focus styles of `Chip`

previously, and now with the update, we apply the same focus style via a sibling selector in `InternalChip.css` but it is a second place with the same definition 😿

ZakaryH:
**Outstanding Topics to Follow Up**
- not needing to import "React" and our linting
- Dark Mode compatibility with Selectable Chip
- `div`s and other markup as children of `button` when Chip has an `onClick`



one other thing I just noticed is that you can now focus `disabled` Chip. I'm not sure that's how we want that to behave. I believe it's because previously Chip used to always be a `button` and since it's `disabled` you aren't permitted to focus it
image


however now, if you don't provide an `onClick` it'll be rendered as a `div` and it doesn't implement the same un-focusable behavior when `disabled`

image

packages/components/src/Chip/Chip.tsx
```diff
@@ -41,19 +42,21 @@ export const Chip = ({
...
-      <button
+    <Tooltip message={tooltipMessage} setTabIndex={false}>
+      <Tag
         className={classes}
-        onClick={(ev: React.MouseEvent<HTMLButtonElement>) =>
-          onClick && onClick(value, ev)
+        onClick={(ev: React.MouseEvent<HTMLButtonElement | HTMLDivElement>) =>
+          onClick?.(value, ev)
         }
         tabIndex={tabIndex}
```
nad182, line 54: If you make the following change, you should be able to fix the issue with disabled chips being currently focusable:
```suggestion
        tabIndex={disabled ? -1 : tabIndex}
```

nad182 (changes requested):
Looks great! Just one minor suggestion, otherwise, I'm ready to approve this PR!

chris-at-jobber:
Cleaned up the colors in the SelectableChip story in #2051

packages/components/src/Chip/Chip.tsx
ZakaryH: done and done!

nad182 (approved):

ZakaryH:
wanted to give it one last try in our use case with the overrides and I'm glad I did. two things were no longer working.

`--public-chip-base-content-color` was not being applied because we have a stylesheet that is targeting `p` directly, applying a `color` - this would get priority over 

```
.base {
  color: var(--chip-base-content-color);
}
```

meaning our variables did nothing. I've increased the specificity, also targeting `.base p` now. I think this is pretty safe since we always use `Typography` without a different type of `element` specified.

then, with my efforts to keep the hover/focus styles of subtle Chips consistent, I overlooked that my selectors were too general. they were applying a hover/focus style to any sibling of the input that is active, meaning that if you provided an override for `--public-chip-base-hover-bg-color` again it would lose to this other rule

to fix it, I've added an `inactive` class that we apply in single/multi select. this class is what we target to add the background hover/focus styles to subtle chips. then, if it's active AKA `base` then we don't apply any kind of background color styling, allowing the `Chip` and its overrides to apply.

@chris-at-jobber 
@nad182 

this isn't the exact place in product where we had the issue, but I've replicated the background color and applied some overrides that would be the same as what we wanna do elsewhere.

Kapture 2024-10-01 at 10 54 19

nad182:
@ZakaryH , I might be doing something wrong, but when I test changes with the pre-release branch in `/work_configuration/edit/work_settings` route, I see that
1. I'm not able to tab through the chips with the `TAB` key, only with arrows, since those elements are radio buttons. This might be intentional, just wanted to call it out.
2. The hover and focus colour are different. This results in a slightly weird behaviour (also different from prod), when hovering the selected chip changes its colour:

https://github.com/user-attachments/assets/862edee8-7f4c-4ab5-8e2e-622bb933d40c

ZakaryH:
@nad182 

>I'm not able to tab through the chips with the TAB key, only with arrows, since those elements are radio buttons. This might be intentional, just wanted to call it out.

yup. that is intentional. internally we are using those checkboxes and radios, and leveraging their built in keyboard navigation behavior.


>The hover and focus colour are different. This results in a slightly weird behaviour (also different from prod), when hovering the selected chip changes its colour:


I'm not sure I follow, or maybe we have different CSS override setups. when I hover over non selected chips, they change color which is desired. then when I hover over the selected one, mine is not changing color - but that's very much dependent on how you implement your overrides. 

do you have one for `--public-chip-base-hover-bg-color:` and is the the same as the base bg color?

nad182:
> @nad182
> I'm not sure I follow, or maybe we have different CSS override setups. when I hover over non selected chips, they change color which is desired. then when I hover over the selected one, mine is not changing color - but that's very much dependent on how you implement your overrides.
> 
> do you have one for `--public-chip-base-hover-bg-color:` and is the the same as the base bg color?

I'm not applying any overrides. I just published your code changes as a pre-release version (you can see above) and used it locally on JO. I thought that it should work without any overrides on the JO side.

ZakaryH:
>I'm not applying any overrides. I just published your code changes as a pre-release version (you can see above) and used it locally on JO. I thought that it should work without any overrides on the JO side.

oh! ok I see what you mean.

I thiiink that's intentional. where the unselected chips are "subtle" and they have a hover colour
then the selected chip is a "base" chip and it has a different hover colour

@chris-at-jobber does that sound right?

nad182:
> oh! ok I see what you mean.
> 
> I thiiink that's intentional. where the unselected chips are "subtle" and they have a hover colour then the selected chip is a "base" chip and it has a different hover colour
> 
I originally thought that it could have been intentional too, but it is different from what we currently have in production. If we hover over selected chip, its colour stays the same:

https://github.com/user-attachments/assets/1a56b4dc-db36-46c8-9c1d-2e0ebd20b2a9

Also it's a bit of a weird behaviour when "active" element (selected/focused) gets a different effect, when it gets hovered. It kind of intuitively feels like it should stay the same.

chris-at-jobber:
It's slightly _different_ from what we have right now, but the active Chip is still interactive, so having a hover state doesn't seem particularly unexpected or incorrect.

PR #2037, opened 2024-09-23
refactor(hooks): Add useOnMount and useCallbackRef [96563]

Aiden-Brine:
## Motivations

There are many copies of `useOnMount` and `useIsMounted` throughout the codebase. `useOnMount` performs a callback one time once the component is mounted whereas `useIsMounted` returns a ref that stores a boolean on whether the component is mounted or not. I think combining them would violate the Single Responsibility Principle because providing information about if the component is mounted and performing actions once the component is mounted are two different responsibilities.

Unfortunately, there are many cases where `useIsMounted` is used when `useOnMount` would be preferable. For example [here](https://github.com/GetJobber/jobber-mobile/blob/fa287540dd6e1dab133d347dc9dec7ec3a292039/src/components/SubscriptionPromptBanner/components/AppStoreIntroOfferBanner/AppStoreIntroOfferBanner.tsx#L58) and [here](https://github.com/GetJobber/jobber-mobile/blob/fa287540dd6e1dab133d347dc9dec7ec3a292039/src/features/Note/views/NoteUpsert/NoteUpsert.tsx#L124) the logic could be simplified by using `useOnMount`. That being said there are occasions like [here](https://github.com/GetJobber/jobber-mobile/blob/679d4c0751e6c4678b8a68d96fcc1d04f8b7a7d3/src/atlantis/FormatFile/hooks/useCreateThumbnail/useCreateThumbnail.ts#L18) and [here](https://github.com/GetJobber/atlantis/blob/95de228a7e422b3d2ef4d937b83ab5742b2c4362/packages/components/src/Autocomplete/Menu.tsx#L116) where `useIsMounted` is required. For this reason we need both the current functionality of `useIsMounted` and `useOnMount` but we only need one version of each available in Atlantis that can be used in the other repos. There will be a followup PRs to replace the versions of these two hooks in the other repos with the Atlantis versions. `useOnMount` also uses a hook `useCallbackRef` which will also be pulled into Atlantis and is currently only ever used by `useOnMount`.

The only difference between the two current instances of useIsMounted ([atlantis](https://github.com/GetJobber/atlantis/blob/master/packages/hooks/src/useIsMounted/useIsMounted.ts) and [jobber-mobile](https://github.com/GetJobber/jobber-mobile/blob/master/src/hooks/useIsMounted/useIsMounted.ts)) is the fact that Atlantis uses `useSafeLayoutEffect` instead of `useLayoutEffect` which was added as a part of Remix. I chatted about this with Scott and he suggested that even though we are backing away from Remix it would still be good to use `useSafeLayoutEffect`.

### Added

Adds `useCallbackRef` and `useOnMount`. You can see these are identical to those used in other projects:
- Jobber Online: [useCallbackRef](https://github.com/GetJobber/Jobber/blob/master/frontend/src/utilities/hooks/useCallbackRef/useCallbackRef.ts) & [useOnMount](https://github.com/GetJobber/Jobber/blob/master/frontend/src/utilities/hooks/useOnMount/useOnMount.ts)
- jobber-frontend (jobber-online): [useCallbackRef](https://github.com/GetJobber/jobber-frontend/blob/main/apps/jobber-online/app/lib/react/useCallbackRef.ts) & [useOnMount](https://github.com/GetJobber/jobber-frontend/blob/main/apps/jobber-online/app/lib/react/useOnMount.ts)
- jobber-frontend (spa-websites): [useCallbackRef](https://github.com/GetJobber/jobber-frontend/blob/main/apps/sp-websites/app/lib/react/useCallbackRef.ts) & [useOnMount](https://github.com/GetJobber/jobber-frontend/blob/main/apps/sp-websites/app/lib/react/useOnMount.ts)
- jobber-mobile: Does not use `useCallbackRef` or `useOnMounted`

### Changed

No existing functionality is changed

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

I added `useOnMount` to Storybook. I also tested that the Atlantis `onMount` will work via pre-release on this [this branch](https://github.com/GetJobber/Jobber/pull/51057) and for a little added testing I loaded one of the components with some console logging to verify that `onMount` is only called once despite the component being rendered multiple times:

Screenshot 2024-09-24 at 11 03 04 AM

Changes can be
[tested via Pre-release](
https://github.com/GetJobber/atlantis/actions/workflows/trigger-qa-build.yml)

---

Random photo of Atlantis


packages/hooks/README.md
```diff
@@ -18,4 +19,4 @@ Shared hooks for components in Atlantis.
 
 ## Installing
 
-`npm install hooks @jobber/hooks`
```
Aiden-Brine, line 21: This is a typo we discovered while debugging pre-releases

packages/hooks/src/useCallbackRef/useCallbackRef.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useMemo, useRef } from "react";
+
+/**
+ * A custom hook that converts a callback to a ref to avoid triggering re-renders when passed as a
+ * prop or avoid re-executing effects when passed as a dependency
+ */
+export function useCallbackRef<T extends (...args: unknown[]) => unknown>(
```
taylorvnoj, line 7: I see that this only used in `useOnMount` but I am curious if you think it's worth surfacing this as a story? I'm not sure so don't take my question as something I'm alluding to 😆  Just curious if you think there's a scenario where someone looks at Hooks in storybook and thinks, _there's no useCallbackRef_ and then creates it again for another purpose? Or do you think this would really only be used in `useOnMount` and it's not worth surfacing?

taylorvnoj (commented):
I just have one question that I'll ask while I'm testing quickly - otherwise looking great!

taylorvnoj (approved):
Hey Aiden! I smoke tested and everything behaves as expected. Let me know if you have thoughts on my earlier comment but otherwise, this is good to go!

packages/hooks/src/useCallbackRef/useCallbackRef.ts
Aiden-Brine: This is a good suggestion. I didn't originally because there wouldn't really be any functionality to show in Storybook so it didn't seem like documenting it would really help folks learn how to use the hook and I left that out. But looking at it from a discoverability perspective it would definitely be beneficial for folks to know it is there. So I added some simple documentation 😄

taylorvnoj: Awesome - thank you for adding that!

taylorvnoj (approved):
Seems like the old approval stuck, but giving one more stamp of approval now, given the documentation update. Thanks Aiden!

PR #2043, opened 2024-09-24
docs: Add an example how to clear all filters in DataList (JOB-103665)

nad182:
## Motivations

Some teams needed to create a more streamlined Filtering experience in `DataList` and weren't sure how to achieve this with existing components.

### Added

- A Story to showcase how to add a "Clear filters" button to clear all filters in all filter groups,
- as well as how to clear all filters within one filter group without opening a `Combobox` drop down, e.d. 1 step instead of 2-3 steps.
- add "@apollo/client" as an external dependency to install when running playground

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Regex that was parsing Story source code to generate import strings. It was picking up TypeScript types, since they also start with a capital letter and frequently added after "<" symbol (e.g. `useState<StateInterface>` or `type CustomType<T>`) 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


docs/components/DataList/DataList.stories.mdx
```diff
@@ -246,6 +246,112 @@ children and inserts it above the list of items.
 </DataList>
 ```
 
+#### Clearing all filters
+
+##### In all filter groups (e.g. with the "Clear Filters" button)
+
+You can add a button that will clear all the filters by using the combination of
+the `Combobox` and `Button` components (check out the `Code` tab in the "Clear
```
MichaelParadis, line 254: I think it may be worth adding a link to the Clear All Filters story with something like
`[Clear All Filters Story](../?path=/story/components-lists-and-tables-datalist-web--clear-all-filters)

docs/components/DataList/Web.stories.tsx
```diff
@@ -404,6 +408,309 @@ export const Basic: StoryObj<typeof DataList> = {
...
+
+    getCollectionByPath(items) {
+      return items?.allPeople;
+    },
+  });
+
+  const items = data?.allPeople.edges || [];
+  const totalCount = data?.allPeople.totalCount || null;
+
+  const mappedData = items.map(({ node }) => ({
```
MichaelParadis, line 522: I noticed that this doesn't actually filter the data based on the filters. Do you think it would be worth filtering the data as well?

nad182: Correct, I didn't implement the actual filtering for 2 reasons:
1. We're not doing it in any other stories, even though technically we're exposing filters - we only show alerts or console logs.
2. I figured filtering would be done on the server side anyways, so I thought there's no point in showing how to implement it.

Given my reasoning, do you still think I should add data filtering logic?

docs/components/DataList/DataList.stories.mdx
nad182: Great suggestion!

docs/components/DataList/Web.stories.tsx
```diff
@@ -404,6 +408,309 @@ export const Basic: StoryObj<typeof DataList> = {
   ),
 };
 
+export const ClearAllFilters: StoryFn<typeof DataList> = args => {
+  interface SelectedFilters {
```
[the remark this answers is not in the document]
nad182: Yep, that would require either refactoring `Playground.tsx` to work with stories written in CSV3.0, or rewriting stories to to CSV2.0 and making tweaks inside of the stories.
I'm in favour of the first approach, since there's no reason to "go backwards", using old standards/API. But you are right, it is a separate task (or even a set of tasks).

docs/components/DataList/DataList.stories.mdx
```diff
@@ -246,6 +246,113 @@ children and inserts it above the list of items.
...
+  <Button label="Clear filters" type="tertiary" onClick={removeAllFilters} />
+</DataList.Filters>;
+```
+
+##### Within one filter group (e.g. Suffix on the Chip)
+
+If you'd like to to implement a filter mechanism to clear all filters within one
+group without opening the `Combobox` "dropdown", you could achieve this by using
+`<Combobox.Activator>` and `<Chip>` components (check out the `Code` tab in the
+"Clear All Filters" story for a more elaborate example).
```
MichaelParadis, line 299: I hate that I missed this initially but could you also add the link here?

docs/components/DataList/Web.stories.tsx
MichaelParadis: I think that makes sense especially when you consider that the filtering will be done server side

chris-at-jobber:
Small request - we should document with a `subtle` button as the preferred/suggested pattern:
image

docs/components/DataList/DataList.stories.mdx
```diff
@@ -287,7 +287,12 @@ function removeAllFilters() {
     ))}
   </Combobox>
 
-  <Button label="Clear filters" type="tertiary" onClick={removeAllFilters} />
+  <Button
+    label="Clear filters"
+    type="tertiary"
+    variation="subtle"
```
chris-at-jobber, line 293: 🙏

MichaelParadis (approved):
LGTM! I think we should update the other DataList stories for the code tab but that is very much out of scope for this PR

chris-at-jobber (commented):
Design approved 👌

PR #2046, opened 2024-09-25
feat(components): Create FileTypes enum for InputFile and use in validation error

taylorvnoj:
## Motivations

If consumers are setting `allowedTypes` on `InputFile` to be `string[]`, it's possible they might end up with an array like this (where the last string is for DOCX files):

```
allowedTypes={[
    "image/jpeg",
    "image/png",
    "image/heic",
    "application/pdf",
    "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
  ]}
```

When a user then tries to upload a non-allowed file type, they'd see this behemoth validation error:

Screenshot 2024-09-24 at 11 06 11 PM

and that is **not** what we want to see.  Let's clean up the validation errors that surface when a non-allowed file type tries to be uploaded.  

_**Also,**_ then consumers can then implement `allowedTypes` like this:

```
allowedTypes={["JPEG", "PNG", "HEIC", "PDF", "DOCX"]}
```
Existing implementations that use the long strings within `allowedTypes`, will still be converted properly. See mentioned test PR in prod.

### Added

I added a new `FileTypes.ts` file which holds a `FileTypes` enum of common MIME types & their corresponding readable format.  Moving forward, if teams need a new file type, it can easily be added to this enum.

The file provides utility functions to convert MIME types to readable formats (`mimeTypeToReadable`), convert readable file type names to their corresponding MIME types (`convertToMimeTypes`), and format a list of MIME types into a human-readable string (`formatMimeTypes`).

I also added a Custom Allowed Types story, along with some documentation going over `allowedTypes`.

### Changed

Note: When `allowedTypes={images}`, that can include a long long list of possible MIME types. I'm making the assumption that we don't want _all_ the potential MIME types to appear in the error because that would be long and unnecessary. So I've made the error message simple.

| `allowedTypes`  | Before        | After           |
| ---------------- | ----------- | ----------- |
| `images`            | Screenshot 2024-09-24 at 11 01 54 PM | Screenshot 2024-09-24 at 10 59 00 PM |
| `basicImages`   | Screenshot 2024-09-24 at 11 02 51 PM | Screenshot 2024-09-24 at 10 59 35 PM |
| `string[]`           | Screenshot 2024-09-24 at 11 06 11 PM | Screenshot 2024-09-24 at 11 00 25 PM |

If a different file type needs to be added to the `FileTypes` enum, that is a simple process teams can do on their own, moving forward. 

## Testing

https://github.com/user-attachments/assets/70713ef5-b3c5-4b4b-9901-004438e88125

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputFile/FileTypes.ts
+++ b/packages/components/src/InputFile/FileTypes.ts
@@ -26,7 +26,11 @@ export const convertToMimeTypes = (types: string[]): string[] => {
 };
 
 export function formatMimeTypes(types: string[]): string {
-  if (types.length <= 1) return types.join(", ");
-
-  return `${types.slice(0, -1).join(", ")} or ${types[types.length - 1]}`;
+  if (types.length === 0) {
+    return "";
+  } else if (types.length === 1) {
+    return types[0];
+  } else {
+    return `${types.slice(0, -1).join(", ")} or ${types[types.length - 1]}`;
+  }
 }
--- a/packages/components/src/InputFile/FileTypes.ts
+++ b/packages/components/src/InputFile/FileTypes.ts
@@ -1,0 +1,32 @@
+export enum FileTypes {
+  JPEG = "image/jpeg",
+  JPG = "image/jpg",
+  PNG = "image/png",
+  HEIC = "image/heic",
+  PDF = "application/pdf",
+  DOCX = "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+}
+
+export const BASIC_IMAGE_TYPES = ["PNG", "JPEG", "JPG"];
+
+// Converts a MIME type to its readable format
+export const mimeTypeToReadable = (mimeType: string): string => {
+  const entry = Object.entries(FileTypes).find(
+    ([, value]) => value === mimeType,
+  );
+
+  return entry ? entry[0] : mimeType;
+};
+
+// Converts an array of readable file type names to their corresponding MIME types
+export const convertToMimeTypes = (types: string[]): string[] => {
+  return types.map(
+    type => FileTypes[type.toUpperCase() as keyof typeof FileTypes] || type,
+  );
+};
+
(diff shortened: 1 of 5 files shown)


packages/components/src/InputFile/FileTypes.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// Converts an array of readable file type names to their corresponding MIME types
+export const convertToMimeTypes = (types: string[]): string[] => {
+  return types.map(
+    type => FileTypes[type.toUpperCase() as keyof typeof FileTypes] || type,
+  );
+};
+
+export function formatMimeTypes(types: string[]): string {
+  if (types.length <= 1) return types.join(", ");
```
Aiden-Brine, line 29: This line works but is a bit weird because you are singling out empty arrays or arrays with one item and then calling join with `", "`  on them which intuitively doesn't make sense but is handled properly in the end by `join`. I would suggest changing it to be explicit about what you want. Something along the lines of:
```
if (types.length === 0) {
    return ""
} else if (types.length === 1) {
    return types[0]
} else {
    return `${types.slice(0, -1).join(", ")} or ${types[types.length - 1]}`;
}
```
This is definitely more verbose but the next person who comes along doesn't need take that extra step to think 🤔 "What does `.join` do with an array with 1 value again?"

Aiden-Brine:
This is a great improvement to the user experience!! Nicely done. Just the small nit-picky change to make and then it is ready to ship

Aiden-Brine (changes requested):
This is a great improvement to the user experience!! Nicely done. Just the small nit-picky change to make and then it is ready to ship

Aiden-Brine (approved):
Appreciate you making the change, looks great

chris-at-jobber (commented):
Design approved, big fan 🪭

PR #2047, opened 2024-09-25
feat(components): Make Suffix Optional on Single and Multi Select Chips

ZakaryH:
## Motivations

sometimes, consumer don't want to have the little checkmark on the selected Chip for single & multi select Chips

### Added

the ability to opt out of the checkmark on the selected Chip(s)

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

add the `hideSuffix={true}` prop to single and multi select Chips. select a chip or two, and see that the checkmark suffix does not show.

instances without the prop will continue to show the checkmark, making this behavior opt-in

dismissible/selection chips will be unaffected since it works in a different way with its suffix.

omitting `hideSuffix` or `hideSuffix = false`
image
image

`hideSuffix = true`

image

image

Changes can be

---

Random photo of Atlantis


chris-at-jobber:
Question around the API - given the pattern for [open/close](https://atlantis.getjobber.com/?path=/docs/guides-frontend-styleguide--docs#open-and-closed-states) props, which feels similar in nature, would it not be a bit more intuitive to have the prop be `showSuffix` and default to true? Then by "turning off" the property, you "turn off" the suffix?

chris-at-jobber:
Experience-wise, this is doing the thing we want 

- [x] Consumer: can specify whether or not the checkmark suffix is visible in single- and multi-select Chips
- [x] End-user: sees (or does not see) the checkmark as per the consumer's specification

ZakaryH:
>Question around the API - given the pattern for [open/close](https://atlantis.getjobber.com/?path=/docs/guides-frontend-styleguide--docs#open-and-closed-states) props, which feels similar in nature, would it not be a bit more intuitive to have the prop be showSuffix and default to true? Then by "turning off" the property, you "turn off" the suffix?

👍  using the affirmative makes sense to me. I see quite a few instances of `show*` props, and only one `hide*` prop. we can continue to have the opt-in behavior with the default being true like you said.

>(And possibly, although a bit more verbose - would showSelectionSuffix be more immediately understandable?)

I don't mind it. `showSelectedSuffix` is ever so slightly shorter. it's pretty specific to the selected suffix, but that's probably fine. I don't see us adding the ability to use other suffixes, at least not with `Chips/Chip`.

right now you cannot use standalone `Chip.Suffix` with `Chips`. if you could, `showSuffix=false` paired with a `Chip.Suffix` that you provided would be weird. if you didn't want to show it, why are you giving it? at that point the `showSelectedSuffix` makes it clearer.

as an aside, Chips + standalone Chip with a Suffix is going to be a fun problem 😬

packages/components/src/Chips/hooks/UseInternalChip.tsx
```diff
@@ -0,0 +1,32 @@
+import React from "react";
```
MichaelParadis, line 1: Nit: shouldn't the file name be `useInternalChip.tsx` instead of `UseInternalChip.tsx`?

Another thought on this is could this not be a utility function since the hook is only returning a callback that isn't ever changing since it is passed everything as a parameter to the callback?

ZakaryH: ah good catch. it should be lowercase.

I started out with it as a util, and then realized it needed to be a `.tsx` file in order for it to handle the `<Icon/>` we are returning.

for no real reason other than convention, I thought it was weird to have a `util.tsx` rather than `util.ts` though I do see we have a `Datalist.utils.tsx` which I assume is that way for the same reason

I don't feel strongly about it being a hook vs a util. there's certainly no need for it to be a hook.

MichaelParadis: I think a hook is fine but I think we should at least make it so that callback isn't recreated on every re-render. I feel like that may cause issues in the future

ZakaryH: I'm not sure I agree

using `useCallback` isn't always going to improve performance, and React is pretty good at re-rendering. this computation isn't particularly expensive.

if there's a specific problem we want to solve with a `useCallback` I'm open to it, but I don't like using it "just in case"
https://kentcdodds.com/blog/usememo-and-usecallback

MichaelParadis: I was initially thinking of pulling the function out of the hook and returning it there. I think it is minor enough we can leave it.
Ideally we rename the file to `useInternalChip.tsx` and I can approve

ZakaryH: >I was initially thinking of pulling the function out of the hook and returning it there. I think it is minor enough we can leave it.

gotcha. the only thing with that is we need it in 2 places, single and multiselect, so having it be an unexported function in the file would mean we duplicate it. it is only 2 instances, so I'm not totally opposed to doing that.

alternatively, I can make it just a util/function that we call rather than a hook like you had originally asked about.

ZakaryH: renamed ✅
moved function outside of hook to avoid being remade ✅

MichaelParadis (approved):
LGTM!
Thanks for the changes

chris-at-jobber (commented):
Design approved! Thanks for the change on the boolean logic / prop name

PR #2048, opened 2024-09-26
fix(docx): Fix regex in Playground for the code tab to work correctly

nad182:
### Added

- 

### Changed

- Fixed regex to use negative lookahead instead of negative lookbehind, since the latter doesn't seem to be supported by the Cloudflare JS engine.
- Improve existing regex to select words that start with capital letter (to match React components, and not HTML tags), which would remove at least one loop.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Check the `code` tab (e.g. in `DataList` "Clear All Filters" story.
- Make sure it's rendered, since it's currently broken on `master`:
image
- Verify that there are no unused/weird imports, such as 

```
import { T } from "@jobber/components/T";
import { SelectedFilters } from "@jobber/components/SelectedFilters";
```

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/.storybook/components/Playground/Playground.tsx
+++ b/.storybook/components/Playground/Playground.tsx
@@ -192,20 +192,22 @@ function getSingleModuleImport(
 }
 
 function parseSourceStringForImports(source: string, extraImports: string[]) {
-  // Grab the first word after <
-  // and only if there isn't a "use" or "type " in front of it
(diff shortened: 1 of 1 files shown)


MichaelParadis (approved):
LGTM!

PR #2051, opened 2024-09-30
docs(components): use semantic colors in `SelectableChip` story

chris-at-jobber:
## Motivations

Story for Selectable Chip did not work in dark mode

Before:
image

After:
image

### Changed

- used semantic color
- removed conditional logic for which color to show based on state (we don't use this pattern)

### Fixed

- SelectableChip story works in dark mode

image
image

## Testing

Review [SelectableChip story locally](http://localhost:6005/?path=/story/components-selections-chip-web--selectable) or on [Cloudflare build](https://update-states-for-non-clicka.atlantis.pages.dev/?path=/story/components-selections-chip-web--selectable)

---

Random photo of Atlantis


nad182 (changes requested):
This looks good!
Do you mind removing `{}`, since they're not really needed? Also I just want to highlight that the icon ("checkmark") will no longer be green ever. It looks fine, just wondering if this won't be a "UI breaking change" for other teams?

chris-at-jobber:
@nad182 pushed up the braces change! In terms of the color change, I have confirmed that there are currently 0 usages of `ChipSelectable` in `jobber` or `frontend`, and no imports within other Atlantis components outside of the component itself, so I am confident that

- the component itself is not used by any of our "primary" consuming applications
- the change itself would not have undesired impacts if it were used

image
image
image

nad182 (approved):
Thanks for double checking the usages!

PR #2054, opened 2024-10-02
feat(components): Decouple shadow from clickability [JOB-105101]

Aiden-Brine:
## Motivations

It was [requested](https://getjobber.slack.com/archives/C03BJHV3PMG/p17266038179375890) that the presence of a drop shadow on a card on the web not depend on if the card is clickable and instead is controllable via an elevation prop. Note that [this PR](https://github.com/GetJobber/Jobber/pull/51539) should be reviewed at the same time since it adds breaking changes.

## Changes

Removed the drop shadow CSS from clickable cards on web. Also changed the `transition` prop for clickable cards and added a shadow focus on clickable cards `focus-visible` to be more consistent with other clickable elements in the codebase. 

### Added

An elevation prop to the `Card` component which controls if the card has a drop shadow

### Changed

Cards will no longer automatically have a drop shadow if they are clickable. Instead, the presence of drop-shadow depends on the new `elevation` prop. Clickable cards are more consistent with other clickable elements in Atlantis.

### Deprecated
N/A

### Removed

N/A

### Fixed

N/A

### Security

N/A

## Testing

I added a new elevation story and this should be testable via Storybook. You can also test it with the pre-release included [here](https://github.com/GetJobber/Jobber/pull/51539)

Changes can be

---

Random photo of Atlantis


chris-at-jobber (changes requested):
Looks great with one exception!

When there is no elevation provided but the Card is clickable, Card should still have a border.

When there is elevation provided, regardless of interactivity, Card should not have a border.

Quick demo video from [the Figma](https://www.figma.com/design/rIIhulZvcp9M82lNOCGv16/branch/BvEdLezlOV8FjQ1ckxqyC9/Product%2FOnline?m=auto&node-id=22677-10824&t=ndDbuD8FJWfjsxea-1):

▶️ [Watch here](https://getjobber.zoom.us/clips/share/bSQjWDrIdRNvs1h9BH5g93wiXcllG6LggXfTxBrRoIQBcyEeppa9MmWlfeR3Yp_8nNHTx_vmufHtQI0pClDrpzXkyg.dwumWgEQPncIIvUN)

packages/components/src/Card/Card.css
```diff
@@ -42,7 +42,6 @@
 .clickable {
   --card--border: none;
```
chris-at-jobber, line 43: I think we just need to delete this line (and move it to the elevation styles?)

chris-at-jobber:
One other note not directly tied to your change, but that I believe we should make, is standarizing the interactions on clickable cards with other elements in the system:

- should get a `--shadow-focus` on `:focus-visible`, in addition to the change in background color
- the [transition](https://github.com/GetJobber/atlantis/pull/2054/files#diff-1bf632c749827ced2b3e80addf18ec8b10759e48e8973b61c30d2fa618ae950cR48) should be `all var(--timing-base) ease-out`

packages/components/src/Card/Card.css
Aiden-Brine: Ah good call, sorry I should have caught this based on the Figma. Thank you for pointing it out

packages/components/src/Card/Card.css
```diff
@@ -40,19 +40,19 @@
...
-  transition: all var(--timing-quick) linear;
+  transition: all var(--timing-base) ease-out;
 }
 
 .clickable:hover,
 .clickable:focus {
   background-color: var(--color-surface--hover);
 }
+.clickable:focus-visible {
+  box-shadow: var(--shadow-focus);
```
chris-at-jobber, line 54: I've just realized that we're already handling the focus in the "standard" (ie leveraging `outline`) on `Card` [(up on L10)](https://github.com/GetJobber/atlantis/pull/2054/files#diff-1bf632c749827ced2b3e80addf18ec8b10759e48e8973b61c30d2fa618ae950cR10) and this now conflicts with that (particularly noticeable in a "high" elevation), as the "elevation" shadow gets replaced with the "focus" shadow. 

Example:
https://github.com/user-attachments/assets/01cef92d-4920-44d9-a1d3-54e0a6a09253

We could stack the shadows by comma-separating them like

`box-shadow: var(--shadow-focus, var(--shadow-{myCurrentElevation})` 

but I think we can actually just drop this new shadow declaration here and get more or less the same end result with much less complexity.

Aiden-Brine: I see what you mean. Always game for less complexity so I have dropped this.

chris-at-jobber (commented):
Design approved! Thanks @Aiden-Brine

taylorvnoj:
@Aiden-Brine @chris-at-jobber In the slack thread mentioned, there's conversation about having the `arrowRight` Icon as an indicator that the card is clickable.  I think it might be a good idea to build this out as a story example/update the current With On Click story, so that we're giving an example of a suggested pattern?

Screenshot 2024-10-08 at 2 54 13 PM

chris-at-jobber:
Oh yeah that's a great idea

packages/components/src/Card/cardelevations.css
```diff
@@ -0,0 +1,17 @@
+.lowElevation,
```
taylorvnoj, line 1: I know that in this folder, `cardcolors.css` exists, but we should be using pascal case for our file names.  Don't worry about updating `cardcolors` but if we can update these new files to `CardElevations.css` along with the d.ts file, that would be great!

Also, I think Jacob just merged a PR which makes each css file a module file, and that's leading to some merge conflicts. Might want to pull that in and update this at the same time.

taylorvnoj (changes requested):
Look great - I do have a suggestion to update the css file name and pull in Jacob's latest module changes. Plus that story example question, if we think it's valuable.

packages/components/src/Card/cardelevations.css
Aiden-Brine: Ah great call, I was definitely basing my naming off of `cardcolors.css` but it makes sense that I should be using pascal case based on what I have seen elsewhere in the codebase. I was waiting until I saw if there were other changes required before doing the rebase but I will do both now

PR #2055, opened 2024-10-02
feat(components): Add keepOpenOnClick prop to Combobox.Action

taylorvnoj:
## Motivations

We want to provide a way for the consumer to close the Combobox after interacting with a `Combobox.Action`, other than the user clicking outside of the combobox.

If a `Combobox.Action` opened a modal for example, the Combobox would remain open and on top of the modal layer. Now, after clicking a `Combobox.Action`, the `Combobox` menu will close by default.  If the expected behaviour is that the `Combobox` remain open even after clicking an action, `keepOpenOnClick` can be added to `<Combobox.Action />`.

### Changed
`Combobox.Action` now closes on default after clicking the action.

### Added

New prop `keepOpenOnClick` on `ComboboxActionProps` which defaults to `false`.

Updated the Action section on the Combobox Docs page, and added a new story example, showcasing a use case where you might want to keep the `Combobox` open even after clicking an action.

Also updated the `ComboboxAction.test.tsx` file. Since I'm now utilizing `setOpen` from `ComboboxContext`, I wanted to tuck away the `ComboboxContextProvider` wrapper and keep the test file clean.  I added a test, _it("keeps the Combobox open when keepOpenOnClick is true")_

## Testing

Test via the [Keep Open on Click](http://localhost:6005/?path=/story/components-selections-combobox-web--keep-open-on-click) story

Pre-release test branch is tagged below

---

Random photo of Atlantis


ZakaryH:
two thoughts I have

1. might people want to close to combobox from something other than an action?

if they did, this wouldn't solve for that whereas the ref solution would. 

that said, I'm struggling to imagine when you would have a combobox open, and you're doing something else that would need to close it that isn't finishing off whatever comobobx interaction you're in the middle of doing. I can't think of anything reasonable really.

2. looking at some other combobox implementations, I'm wondering if there's even a need for a prop like this

maybe instead of making this optional via prop, we make it always close when you interact with a combobox's action?

we only have 3 instances of `Combobox.Action`, I'm playing around with them and imagining if they closed on action press. I'm kinda thinking it improves the flow.

taylorvnoj:
@ZakaryH we did have the offline chat/agreement about what to do with this component, but I'll respond to your comment here anyway.

> might people want to close to combobox from something other than an action?

I look at it this way. In Combobox, you're either searching, selecting or actioning. Can I call search a living element? It's not something that should dictate whether Combobox is open or closed. Options, in a way, dictate the open/close of Combobox when the component is single select.  Then if it's multi select, users can click outside of the Combobox or press the escape key. Clicking on the activator controls open/close. 
Actions are the only interactive part of Combobox left where users can't dictate whether the Combobox is opened/closed.  So I agree with you - I can't think of a situation outside of those, where the Combobox would need to be opened/closed in another way.


> looking at some other combobox implementations, I'm wondering if there's even a need for a prop like this

This is where we (Chris lol) noticed the IndexView Tag implementation in product. A dynamic action may be used to create a new option over and over again and in that case, if we closed the Combobox each time, that would add extra clicks and be time consuming. To emphasize that this might be a common reason _not_ to close the Combobox on action click, I made a version of this in the story I added. Otherwise, the Combobox should close on default (until another implementation comes up we aren't thinking of 😅 ). 

So in summary - what we landed on with the help of Chris, is to default the prop to false so that the Combobox closes after interacting with a Combobox Action.  If an implementation of Combobox Action should **not** close after clicking the action, they can add `keepOpenOnActionClick`.

docs/components/Combobox/Web.stories.tsx
```diff
@@ -436,6 +438,49 @@ const ComboboxInfiniteScroll: ComponentStory<typeof Combobox> = args => {
...
+          >
+            <Combobox.Option id="1" label="Search" />
+            <Combobox.Option id="2" label="for a" />
+            <Combobox.Option id="3" label="friend" />
+
+            <Combobox.Action
+              visible={({ searchValue }) => Boolean(searchValue)}
+              label={({ searchValue }) => `Add "${searchValue}" as Chip`}
+              onClick={(_, { searchValue }) => handleActionClick(searchValue)}
+              keepOpenOnActionClick
```
ZakaryH, line 467: I'm debating the name of this prop and if we can either improve clarity while also shortening it.

it's on `Combobox.Action` so I'd say we've already got the context of `OnActionClick` and we can drop the `Action`.

`keepOpen` feels verbose to me. I think it's the `keep` that's doing it for me.

I know we want to try to do the positive of things, so `open` rather than `close` but if we went with `close` that does shorten it...

`closeOnClick`? it's a bit ambiguous what we're closing. `closeComboboxOnClick`? `closePopoverOnClick`?

I'm really struggling to keep it worded as the positive. which is because it's not going to open, it's already open. it's either going to do nothing, staying open, or it's going to close. which is of course why we have to qualify it with "keep".

I think if we must keep it positive: `keepOpenOnClick`

docs/components/Combobox/Web.stories.tsx
```diff
@@ -436,6 +438,49 @@ const ComboboxInfiniteScroll: ComponentStory<typeof Combobox> = args => {
...
+            onSelect={setSelected}
+            selected={selected}
+            label="Add Chip"
+          >
+            <Combobox.Option id="1" label="Search" />
+            <Combobox.Option id="2" label="for a" />
+            <Combobox.Option id="3" label="friend" />
+
+            <Combobox.Action
+              visible={({ searchValue }) => Boolean(searchValue)}
```
ZakaryH, line 464: any reason we want to have to search before exposing the action in this example?

it's for sure my fault for literally not reading the words in front of my eyes, but I was confused why there was no action when checking out this story for an action. 

normally the content in the stories is just silly names that I default to not reading it, or at least not actually processing it. it just goes in and out immediately.

docs/components/Combobox/Web.stories.tsx
taylorvnoj: @jdeichert  Chris identified a Combobox in the Clients Index where it should stay open even when clicking the action.

@ZakaryH this demo video hopefully explains your question below. The web story replicates a common reason why we'd want to use this prop to keep the Combobox open. Let me know if I'm misunderstanding your comment



https://github.com/user-attachments/assets/b2ac3540-e274-4390-9608-27f8c046eb23

ZakaryH: oh yeah I get that it's closer to a real flow. lemme reword my question

what is the goal of the storybook example?

is it to demonstrate how the `keepOpenOnClick` works?

or is it to provide an example flow or design of when you might use `keepOpenOnClick`?

while the latter does include the former, it does it in a way that requires an extra step to get there.

I think a comparable example would be the "Clear selection" story for Combobox. the flow would be more realistic if we made people make selections first, but instead we pre-populate it so you can try the thing the story is about immediately.

also another point for shortening the name, it doesn't even fit in the side bar 😄 
image

taylorvnoj: Zak and I jumped on a call, to summarize:

We'll leave the story as is.  The story is both to demonstrate how the prop works and also, to provide an example flow for when you might use the prop.  It's fair that yes, the user will have to do some reading within the Combobox and clicking, but I feel like the story is a golden example of why we might want to use this prop.  I personally feel like a simple story where you click an action and an alert pops up, saying "the Combobox is still open" doesn't paint a picture of the intended prop behaviour.  Yes this prop can keep the Combobox open, we state that on the Docs page. But _how_ should the prop be used? I feel like it's the purpose of the story example to inform designers/engineers.  Also, we have similar behaviour in the Dynamic Action story, where you type in the search and that builds the action label.

taylorvnoj: And the prop name has been changed to `keepOpenOnClick` here 25bb2bc835ba1890ec9bd7d3b0aac169c47bc8f3

docs/components/Combobox/Web.stories.tsx
taylorvnoj: summarized our chat in the comment thread above ^

packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.tsx
```diff
@@ -5,7 +5,7 @@ import { ComboboxActionProps } from "../../Combobox.types";
 import { ComboboxContext } from "../../ComboboxProvider";
 
 export function ComboboxAction(props: ComboboxActionProps) {
-  const { searchValue } = useContext(ComboboxContext);
+  const { searchValue, setOpen } = useContext(ComboboxContext);
```
ZakaryH, line 8: wait, I totally forgot - we need to use `handleClose` not just `setOpen` when we are closing the combobox. 

```
const handleClose = useCallback(() => {
    setOpen(false);
    setSearchValue("");

    onSearch && onSearch("");
    onClose && onClose();

    if (selectedOptions.length > 0) {
      shouldScroll.current = true;
    }

```
- sets the open state to false
- clears search
- calls callbacks for a cleared search & on close
- scroll state management

if we don't do all that, when it is defaulting to close on click you'll see things like your previous search still being present when you open it up again.

not great that we didn't have a test for that. we should add one!

so it would need to be something like


```
const { searchValue, handleClose } = useContext(ComboboxContext);
...
const handleClick = (e: React.MouseEvent<HTMLButtonElement>) => {
  if (!props.keepOpenOnClick) {
    handleClose()
  }

  props.onClick(e, options);
};
```

one thing I'd wanna make sure of is if `!props.keepOpenOnClick` is the safest way to do that. since it's optional but without a literal code implemented default it could be
- `undefined` which gives us `true`
- `false` which gives us `true`
- `true` which gives us `false`

I think that's all correct?

ZakaryH (changes requested):
forgot we need to use `handleClose` rather than just `setOpen`!

ZakaryH:
now that I think about it more. I don't even know why we expose `setOpen` from the ComboboxContext. we shouldn't.

we should probably add a `handleOpen`, and then expose that rather than `setOpen`. we'd still have it used within some of the hooks where we need it, we just wouldn't expose it from the context since we know we the `setOpen(false)` usage isn't what we want to do

taylorvnoj:
@ZakaryH I'm pretty confident that this test [here](https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Combobox/Combobox.test.tsx#L405) is covering the case you mentioned.  The Combobox is closed via `handleClose` and then the search term is `""`

I'm also writing a separate ticket regarding the `setOpen` prop, as that is outside the scope of this ticket.

ZakaryH:
so yes that is testing the search gets cleared on close, but that's specifically testing when you close it by clicking outside the combobox

`await userEvent.click(screen.getByTestId(OVERLAY_TEST_ID));`

so imagine for our flow, if tomorrow someone changes the `handleClick` in `Combobox.Action` to be `setOpen(false)`. this existing test would still pass since it's not using the action to close.

ZakaryH (approved):
I totally missed that there was some tests added that verified checking `handleClose` was called. my bad!

I believe we have that behavior tested elsewhere, so knowing that it's called when it should be is enough for this component's unit tests to know about

PR #2057, opened 2024-10-03
fix(components): Remove center alignment in FormField

taylorvnoj:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormField.css
+++ b/packages/components/src/FormField/FormField.css
@@ -72,7 +72,6 @@
   display: flex;
   height: var(--field--height);
   flex: 1;
-  align-items: center;
 }
 
 .textarea.safari .horizontalWrapper {


packages/components/src/FormField/components/ClearAction.css
```diff
@@ -12,6 +12,7 @@
   cursor: pointer;
   justify-content: center;
   align-items: center;
+  transform: translateY(50%);
```
taylorvnoj, line 15: Screenshot 2024-10-03 at 11 52 13 AM

Added this back as well (it was removed in initial PR) to avoid clearable being at the top of the input

ZakaryH:
I have a question about the problem the initial work was looking to resolve - how/when was that happening?

the clearable button shows up where I'd expect it to when it has no placeholder
image


and with a placeholder
image

taylorvnoj:
> I have a question about the problem the initial work was looking to resolve - how/when was that happening?
> 
> the clearable button shows up where I'd expect it to when it has no placeholder image
> 
> and with a placeholder image

On large size inputs:
It was like this
Screenshot 2024-10-03 at 12 38 14 PM

and we noticed with the initial fix it was better aligned
Screenshot 2024-10-03 at 12 38 21 PM

But I didn't test multiline because multiline inputs can't have `clearable`. BUT I didn't realize the entire context of the text input would also align center

chris-at-jobber:
Note for future: I think if we want to solve that clearable placement on large inputs, something along the lines of

```
position: absolute;
top: 50%;
tranform: translateY(-50%);
```

OR

wrap the clearable in a container, set that container to take up the full height of the input, and use flex to manipulate the position of the clearable button

Might be safer paths we take to get there. I think we have this working as expected in mobile so maybe we can take a peek at what we did there!

ZakaryH:
@chris-at-jobber I also thought the `top: 50%` + `position: absolute` would fix it, which technically it does - but it breaks other combinations

https://github.com/GetJobber/atlantis/pull/1856#issuecomment-2062480661

if it's absolute, then it will overlap with other elements like the loading spinner or icons that can be in the same general area. which is why we made it relative.

we could try the container approach though, as long as that lets sibling elements position correctly. I do wonder what that's going to do to the available space for the actual text input though 🤔

PR #2066, opened 2024-10-10
feat(components): Handle closing Combobox with escape key

taylorvnoj:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/hooks/useEscapeKey.ts
+++ b/packages/components/src/Combobox/hooks/useEscapeKey.ts
@@ -1,0 +1,32 @@
+import { useCallback, useEffect, useRef } from "react";
+
+export function useEscapeKey(
+  isOpen: boolean,
+  handleClose: () => void,
+  activatorRef: React.RefObject<HTMLDivElement>,
+) {
+  // create ref to store handleclose and update it when it changes
+  const handleCloseRef = useRef(handleClose);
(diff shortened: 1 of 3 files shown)


packages/components/src/Combobox/hooks/useEscapeKey.ts
```diff
@@ -0,0 +1,32 @@
...
+  handleCloseRef.current = handleClose;
+
+  const listener = useCallback(
+    (event: KeyboardEvent) => {
+      // is the Combobox open and the escape key is pressed
+      if (event.key === "Escape" && isOpen) {
+        event.stopPropagation();
+        handleCloseRef.current();
+        // finds the first button in the activator and focuses it
+        activatorRef.current?.querySelector("button")?.focus();
```
taylorvnoj, line 19: This works right now for our default trigger as well as custom ComboboxActivator (Button or Chip) but I wonder if this will become brittle with more activator flexbility/customization? Although the [Combobox role docs from mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role) suggest that the 'control'/activator should be of type `button`.

PR #2069, opened 2024-10-12
fix(components): DataList toggle sorting header doesn't transpose id and key

JeffMarvin:
## Motivations
Fix a bug happening where DataList is returning an incorrect state for clicks in a sorted header. (It was swapping the key and id of the sorted column)

- needs tests

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed
DataList no longer swaps incorrectly the `key` and `id` in the sorting state returned from the sorting callback triggered on the header. 

### Security

- 

## Testing

Changes can be

A test has been added to show the broken and fixed state. (Check the CI runs - the test was added first, shows broken, and then should pass with the fix.)

NOTE: This has not been tested with any DataLists in other clients, so a smoke check would be a good idea. 

---

Random photo of Atlantis


MichaelParadis (approved):
LGTM!
Tested on the various DataList usages. We didn't notice this before because for toggle sorts the `key` and `id` where the same value in those usages

PR #2072, opened 2024-10-15
fix(docx): Fix bug in PR title generator 

Aiden-Brine:
## Motivations

As I showed [here](https://getjobber.slack.com/archives/C03BKPST4KC/p1728432648402599) there was a bug in the PR title generator where the selected scope would not show up as having been selected. Also, the list of scope values did not match the accepted list of scopes so it was possible that the title generator gave you an invalid title.

## Changes

- Change the value on the scope selector from type to scope
- Updates to the list of options for scopes so that only valid scopes can be selected

### Added

A couple of new scope values

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

Only valid scopes can be selected and they now show up in the scope selector

### Security

- 

## Testing

Play around with the [PR Title Generator](http://localhost:6005/?path=/docs/guides-pull-request-title-generator--docs) to verify things are looking good.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/.storybook/components/PullRequestGenerator/PullRequestGenerator.tsx
+++ b/.storybook/components/PullRequestGenerator/PullRequestGenerator.tsx
@@ -33,14 +33,19 @@ export function PullRequestGenerator() {
               <Option value="refactor">doesn't fix a bug or introduce a new feature</Option>
             </Select>
             <Text>in our...</Text>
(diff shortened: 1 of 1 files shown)


MichaelParadis (approved):
LGTM!

PR #2074, opened 2024-10-16
refactor(components): move Grid to leverage compositional patterns

ericchernuka:
Makes Grid allow children and leverage Grid.Cell to handle the CSS.
## Motivations
Currently, Grid does not allow application developers to abstract away its children into re-usable components. This was due to mapping over the Children directly and proxying their props to an InternalGridCell component that expected the props of the immediate children to be Grid.Cell properties. 

The original intent was to guide developers in leveraging Grid.Cell, as the first child causes issues for any developers who wish to create intermediary components for re-use. The solution was to just remove the intermediary guards in favour of rendering the children and allowing CSS to handle the logic of properly handling layout concerns.

## Changes
Refactors Grid to allow any children and expose Grid.Cell directly.

## Testing
- Leveraging Storybook, no API changes to the output will be seen because Grid.Cell is now mapped directly to what was previously InternalGridCell.

Changes can be

---

Random photo of Atlantis


MichaelParadis (approved):
LGTM! Thanks for this, it will make other people's lives easier.

PR #2075, opened 2024-10-16
feat(components): Allow Fully Custom Combobox Activator

ZakaryH:
## Motivations

Some designs call for even further customization beyond a Chip or Button.

To enable this, we are going to open up the Activator even further to allow more customization.

### Added

Combobox Activator can accept custom markup. No longer limited to only Chip and Button.

### Changed

**toggleOpen & handleOpen**
regardless of if we want to expose the API through render props or not, calling `setOpen` directly can be risky. `setOpen(true)` is fine. `setOpen(false)` is not since we want to make sure to do all the cleanup and invoke all the callbacks in addition to closing the combobox. 

to that end, I've created an equivalent wrapper for `setOpen(true)`: `handleOpen`. if we ever want to add callbacks for `onOpen` then it will be easy to add it to one place over all the places we were using `setOpen(true)`. right now it's very bare bones.

by not making the setter directly available from the Context, or even exporting it from where we make the handlers, it is much harder to accidentally use it which we almost did with a recent change. additionally, if we end up exposing methods via render props it's much safer to expose the `handleOpen`

I went one step further and made a `toggleOpen` method that makes it even simpler where we call the toggle. rather than having the logic in a couple places, you just call this method. I could potentially be convinced to go back to handleClose/handleOpen. it's not a ton of logic to repeat and does offer more granular control, though I'm not too sure what you'd do any differently with it. you can only go back and forth between the states. calling open while it's open, or closed while it's closed make no sense.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Combobox/Combobox.stories.mdx
+++ b/docs/components/Combobox/Combobox.stories.mdx
@@ -66,7 +66,7 @@ needs, you may instead provide a function as the child to `Combobox.Activator`.
 
 As arguments, the function will receive the toggle method to swap between open
 and closed states, as well as accessibility attributes required for a Combobox.
-_You must implement these yourself_. See the custom example for reference.
+**You must implement these yourself**. See the custom example for reference.
 
 The render function must return a single, valid JSX Element (it can of course
 have children, it simply must have a single wrapping tag). Ensure that tabindex,
--- a/docs/components/Combobox/Combobox.stories.mdx
+++ b/docs/components/Combobox/Combobox.stories.mdx
@@ -66,7 +66,9 @@ needs, you may instead provide a function as the child to `Combobox.Activator`.
 
 As arguments, the function will receive the toggle method to swap between open
 and closed states, as well as accessibility attributes required for a Combobox.
-**You must implement these yourself**. See the custom example for reference.
+**You must implement these yourself**. See the
+[render function example](../?path=/story/components-selections-combobox-web--custom-activator)
+for reference.
 
 The render function must return a single, valid JSX Element (it can of course
 have children, it simply must have a single wrapping tag). Ensure that tabindex,
--- a/docs/components/Combobox/Combobox.stories.mdx
+++ b/docs/components/Combobox/Combobox.stories.mdx
@@ -36,6 +36,9 @@ initiate option selection.
 
 ##### Default activator
 
+To use the default activator simply omit any kind of `Combobox.Activator`
+subcomponent in your instance.
+
 Examples of using a default activator can be found in the
 [single-select](../?path=/story/components-selections-combobox-web--single-select)
 and
@@ -50,14 +53,25 @@ activator.
 ##### Custom activator
 
 The custom activator should be used in any instance where the default activator
-is not suitable. Your custom activator should be a child of
-`<Combobox.Activator />`. An example of using a custom activator can be found in
-the
+is not suitable.
+
+**Child React Component:** To use a customized `Chip` or `Button` your custom
+activator should be a child of `<Combobox.Activator />`. An example of using a
+custom activator can be found in the
 [custom activator](../?path=/story/components-selections-combobox-web--custom-activator)
 story.
 
-Omitting a custom activator subcomponent will result in the default activator
-being used.
+**Child Render Function:** If neither a Chip nor a Button is sufficient for your
+needs, you may instead provide a function as the child to `Combobox.Activator`.
+
+As arguments, the function will receive the toggle method to swap between open
+and closed states, as well as accessibility attributes required for a Combobox.
+_You must implement these yourself_. See the custom example for reference.
+
+The render function must return a single, valid JSX Element (it can of course
+have children, it simply must have a single wrapping tag). Ensure that tabindex,
+focus, and keyboard interactivity are considered and handled appropriately if a
+non traditionally interactive element such as a div is used.
 
 #### Content
 
--- a/packages/components/src/Combobox/Combobox.tsx
+++ b/packages/components/src/Combobox/Combobox.tsx
@@ -56,6 +56,7 @@ export function Combobox(props: ComboboxProps): JSX.Element {
       handleClose={handleClose}
       shouldScroll={shouldScroll}
       searchValue={searchValue}
+      label={props.label}
     >
       <div ref={wrapperRef} className={styles.wrapper}>
         {open && (
--- a/packages/components/src/Combobox/Combobox.tsx
+++ b/packages/components/src/Combobox/Combobox.tsx
@@ -32,8 +32,8 @@ export function Combobox(props: ComboboxProps): JSX.Element {
     searchValue,
     setSearchValue,
     open,
-    setOpen,
     handleClose,
+    handleOpen,
     handleSelection,
     internalFilteredOptions,
     handleSearchChange,
@@ -52,7 +52,7 @@ export function Combobox(props: ComboboxProps): JSX.Element {
       selected={selectedOptions}
       selectionHandler={handleSelection}
       open={open}
-      setOpen={setOpen}
+      handleOpen={handleOpen}
       handleClose={handleClose}
       shouldScroll={shouldScroll}
       searchValue={searchValue}
@@ -79,7 +79,6 @@ export function Combobox(props: ComboboxProps): JSX.Element {
           setSearchValue={setSearchValue}
           wrapperRef={wrapperRef}
           open={open}
-          setOpen={setOpen}
           options={props.onSearch ? options : internalFilteredOptions}
           loading={props.loading}
           handleSearchChange={handleSearchChange}
--- a/packages/components/src/Combobox/Combobox.tsx
+++ b/packages/components/src/Combobox/Combobox.tsx
@@ -33,8 +33,8 @@ export function Combobox(props: ComboboxProps): JSX.Element {
     setSearchValue,
     open,
     handleClose,
-    handleOpen,
     handleSelection,
+    toggleOpen,
     internalFilteredOptions,
     handleSearchChange,
   } = useCombobox(
@@ -52,7 +52,7 @@ export function Combobox(props: ComboboxProps): JSX.Element {
       selected={selectedOptions}
       selectionHandler={handleSelection}
       open={open}
-      handleOpen={handleOpen}
+      toggleOpen={toggleOpen}
       handleClose={handleClose}
       shouldScroll={shouldScroll}
       searchValue={searchValue}
--- a/packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
+++ b/packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
@@ -5,11 +5,11 @@ import { Chip } from "@jobber/components/Chip";
 import { ComboboxActivator } from "./ComboboxActivator";
 import { ComboboxContextProvider } from "../../ComboboxProvider";
 
-const setOpen = jest.fn();
+const handleOpen = jest.fn();
 const handleClose = jest.fn();
 
 afterEach(() => {
-  setOpen.mockClear();
+  handleOpen.mockClear();
   handleClose.mockClear();
 });
 
@@ -29,7 +29,7 @@ describe("ComboboxActivator", () => {
   });
 
   describe("when open is false", () => {
-    it("should call setOpen when the activator is clicked", () => {
+    it("should call handleOpen when the activator is clicked", () => {
       const { getByRole } = renderComboboxActivator(
         <Button label="Teammates" />,
         false,
@@ -38,7 +38,7 @@ describe("ComboboxActivator", () => {
 
       activator.click();
 
-      expect(setOpen).toHaveBeenCalled();
+      expect(handleOpen).toHaveBeenCalled();
     });
   });
   it("can render a Button with role 'combobox' and onClick handler", () => {
@@ -74,7 +74,7 @@ describe("ComboboxActivator", () => {
 function renderComboboxActivator(child: ReactElement, open: boolean) {
   return render(
     <ComboboxContextProvider
-      setOpen={setOpen}
+      handleOpen={handleOpen}
       handleClose={handleClose}
       selectionHandler={jest.fn()}
       shouldScroll={{ current: false }}
--- a/packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
+++ b/packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
@@ -5,17 +5,17 @@ import { Chip } from "@jobber/components/Chip";
 import { ComboboxActivator } from "./ComboboxActivator";
 import { ComboboxContextProvider } from "../../ComboboxProvider";
 
-const handleOpen = jest.fn();
+const toggleOpen = jest.fn();
 const handleClose = jest.fn();
 
 afterEach(() => {
-  handleOpen.mockClear();
+  toggleOpen.mockClear();
   handleClose.mockClear();
 });
 
 describe("ComboboxActivator", () => {
   describe("when open is true", () => {
-    it("should call handleClose when the activator is clicked", () => {
+    it("should call toggleOpen when the activator is clicked", () => {
       const { getByRole } = renderComboboxActivator(
         <Button label="Teammates" />,
         true,
@@ -24,12 +24,12 @@ describe("ComboboxActivator", () => {
 
       activator.click();
 
-      expect(handleClose).toHaveBeenCalled();
+      expect(toggleOpen).toHaveBeenCalled();
     });
   });
 
   describe("when open is false", () => {
-    it("should call handleOpen when the activator is clicked", () => {
+    it("should call toggleOpen when the activator is clicked", () => {
       const { getByRole } = renderComboboxActivator(
         <Button label="Teammates" />,
         false,
@@ -38,7 +38,7 @@ describe("ComboboxActivator", () => {
 
       activator.click();
 
-      expect(handleOpen).toHaveBeenCalled();
+      expect(toggleOpen).toHaveBeenCalled();
     });
   });
   it("can render a Button with role 'combobox' and onClick handler", () => {
@@ -74,7 +74,7 @@ describe("ComboboxActivator", () => {
 function renderComboboxActivator(child: ReactElement, open: boolean) {
   return render(
     <ComboboxContextProvider
-      handleOpen={handleOpen}
+      toggleOpen={toggleOpen}
       handleClose={handleClose}
       selectionHandler={jest.fn()}
       shouldScroll={{ current: false }}
--- a/packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
+++ b/packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
@@ -7,11 +7,9 @@ import { ComboboxContextProvider } from "../../ComboboxProvider";
 import { ComboboxCustomActivatorProps } from "../../Combobox.types";
 
 const toggleOpen = jest.fn();
-const handleClose = jest.fn();
 
 afterEach(() => {
   toggleOpen.mockClear();
-  handleClose.mockClear();
 });
 
 describe("ComboboxActivator", () => {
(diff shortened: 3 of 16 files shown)


ZakaryH:
~setOpen` -> `handleOpen`

assuming we do some sort of render prop, either as a child or a named prop, we're going to need to provide the close/open methods to the Activator. actually even if we don't we almost did this exact thing with an internal change the other day!

so internally we were using `setOpen(open)` directly which is fine. however `setOpen(false)` is not fine. we always want to do all the cleanup in `handleClose` which ensures search is cleared, some scroll state is reset, setOpen is called and that callbacks are fired when we expect them to

with that, I don't want to give people the ability to make a pretty easy mistake and I've hidden `setOpen` away so all consumers get is `handleOpen` which is currently only a wrapper around `setOpen(true)`

I wouldn't be totally opposed to making this one method that takes in an open/closed arg so that it's just one method rather than 2. let me know what you think.

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -79,7 +79,6 @@ export function Combobox(props: ComboboxProps): JSX.Element {
           setSearchValue={setSearchValue}
           wrapperRef={wrapperRef}
           open={open}
-          setOpen={setOpen}
```
ZakaryH, line 82: this was not being used inside `ComboboxContent` at all. it was very likely a leftover from before we moved `setOpen` into the Context and moved things around.

packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.tsx
```diff
@@ -1,24 +1,19 @@
...
-import { Button } from "@jobber/components/Button";
-import { Chip } from "@jobber/components/Chip";
 import { ComboboxContext } from "../../ComboboxProvider";
 import { ComboboxActivatorProps } from "../../Combobox.types";
 
 export function ComboboxActivator(props: ComboboxActivatorProps) {
-  const { handleClose, open, setOpen } = React.useContext(ComboboxContext);
+  const { handleClose, open, handleOpen } = React.useContext(ComboboxContext);
 
-  if (props.children.type === Button || props.children.type === Chip) {
```
ZakaryH, line 10: a few reasons I want to try doing it this was VS a children render prop, or a named render prop - despite us wanting to use that same format as often as we can

**Reduce Responsibility**

if we take in whatever markup someone wants to provide and then clone it along with the enhancements of aria attributes and the `onClick` needed for the activator - then the consumer doesn't have to go read our docs about what's required of a combobox as far as accessibility.

also, if for some reason we want to change the behavior on click - since that's still under our control, a consumer wouldn't have to update their implementation to match.

all the consumer needs to do is provide the element, and if they wanted to - a separate clickable/interactive element that they'd have to stop event propagation inside so as to not trigger the open/close

**Avoid multiple ways of doing the same thing**

since Combobox and this Activator are some of the earlier usages of compound components and (limited) composition/flexibility - we already have the API for deciding to provide your own markup

if you want the default. don't add a `Combobox.Activator`. 

```
<Combobox {...props}>
  // Options and Actions
</Combobox>
```

if you do, provide one with either a `Chip` or `Button` as the immediate child

```
<Combobox {...props}>
  <Combobox.Activator>
    <Button label="whatever"/>
  </Combobox.Activator>
  // Options and Actions
</Combobox>
```

we currently have 24 instances of custom activators with this pattern and around 11 that use the default.

with that in mind, if we do the `customRenderItem` that's 2 ways of providing your own markup. if it's a Chip or Button, use the children. if it's something else, use the prop.

if we do the children render function, that's a bit better since it's more similar however it's still 2 ways. if it's a Chip or Button, provide markup. if it's something else, provide a function and implement the accessibility, and open/close behavior yourself.

```
 <Combobox.Activator>
   <Chip variant="subtle"/>
 </Combobox.Activator/>
 
 or
 
 <Combobox.Activator>
   {{...activatorAPI} => (<div 
   onClick={activatorAPI.onClick} 
   role={activatorAPI.role} 
   aria-expanded={activatorAPI.open}>
     <Avatar url="smile.jpg" />
   </div>)}
 </Combobox.Activator>
```

as I mentioned above - this is a lot of responsibility for a consumer. the counter point there is that it's more power, but I'm not convinced anyone really wants the power to change the behavior of the Activator nor is that necessarily a place for us to be un-opinionated.

I think it's fine for us to say, when you click on this element it toggles the combobox's visibility. that is the exact purpose and function of this component. how it looks though, is up to you.

chris-at-jobber: Seems reasonable

chris-at-jobber (commented):
In the spirit of validating concepts, would it be possible to update the "custom activator" story with something like this? (I've wrapped a `Heading level={2}` and `Icon` in a custom container that would semantically be a `button`)

image

packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.tsx
```diff
@@ -1,24 +1,20 @@
...
-        } else {
-          setOpen(true);
-        }
-      },
-    });
-  }
-
-  return props.children;
+  return React.cloneElement(props.children, {
+    ...htmlAttributes,
```
ZakaryH, line 11: this puts us in a funny state

if you give us a `div` for a custom activator it will actually have better accessibility since we're adding aria-controls and aria-expanded

image


however providing a Chip, or Button - we won't apply those 😬 which isn't great since our own internal `ComboboxTrigger` is a Chip

image

this is because Chip and Button don't accept an `ariaControls` nor `ariaExpanded`. we'd need to open it up to either those specific values, or aria stuff in general

packages/components/src/Combobox/hooks/useComboboxActivatorAccessibility.ts
```diff
@@ -0,0 +1,16 @@
+import { useContext } from "react";
+import { ComboboxContext } from "../ComboboxProvider";
+import { COMBOBOX_MENU_ID } from "../constants";
+
+export function useComboboxActivatorAccessibility() {
+  const { label, open } = useContext(ComboboxContext);
+
+  return {
+    htmlAttributes: {
```
ZakaryH, line 9: specifying the html-ness of this and making it a distinct set because I had wanted to apply them to the `Combobox.Trigger` too, however that's a `Chip` and throwing these on it does nothing. it would want React props eg. `ariaControl`

packages/components/src/Combobox/hooks/useComboboxValidation.ts
```diff
@@ -32,7 +32,7 @@ export function useComboboxValidation(children?: ComboboxProps["children"]): {
     ComboboxActivator,
     children,
   );
-
+  console.log(activatorElements[0]);
```
ZakaryH, line 35: whoops. you can go.

packages/components/src/Combobox/components/ComboboxTrigger/ComboboxTrigger.tsx
ZakaryH, line 1: I don't love the `onClick` being defined here and the Activator.

might try and turn that into a `handleClick` or something to match the other handlers. keeps it all in one place.

ZakaryH: it would also mean we don't have to pull in `handleClose` and `handleOpen` in this component. not massively impactful or anything. still a minor improvement imo.

ZakaryH:
@chris-at-jobber added a webstory with a div wrapping a h2 and an Icon for you to check out

about the "semantically a button" - we're going to stomp on the `role` provided and make it a `combobox`. not sure if that's exactly what you meant by that or something else entirely.

chris-at-jobber:
> we're going to stomp on the role provided and make it a combobox

Ah right we're handling that for the consumer now. That works!

chris-at-jobber:
The story looks good, helps validate that you can pass in something truly custom. In terms of things like keyboard operability, is there anything we can/should do to automatically ensure the element receives keyboard focus, or is that something we could rely on consumers to handle? (ie always pass in a `tabindex`, or documenting that regardless of `role` the trigger element should use an html `button` so it gets the benefits of operability?)

ZakaryH:
>Ah right we're handling that for the consumer now. That works!

yeah! at least I'd like to. I can still be convinced of doing it another way, but this makes things quite simple for a consumer which I like.

ZakaryH:
>The story looks good, helps validate that you can pass in something truly custom. In terms of things like keyboard operability, is there anything we can/should do to automatically ensure the element receives keyboard focus, or is that something we could rely on consumers to handle? (ie always pass in a tabindex, or documenting that regardless of role the trigger element should use an html button so it gets the benefits of operability?)


good question

we can for sure also automatically give it `tabindex=0` incase it's not a focusable element like a div

though, in that case we might also need to define an `onKeydown` to handle return and spacebar presses since we won't get the nice benefit of a `button` where you only have to define the `onClick`. if we don't explicitly define that for `onKeydown` you'd only be able to click on the div to interact with it

it's pretty trivial to add. it's just weird in the case where we're given a button which has no need for those extra definitions. as far as I'm aware though it doesn't hurt.

ZakaryH:
after implementing both approaches I'm torn.

on one hand, I think this is a vastly superior consumer experience

```
          <div
            tabIndex={0}
            style={{
              display: "inline-flex",
              alignItems: "center",
              gap: "10px",
            }}
          >
            <Heading level={2}>Heading Two</Heading>
            <Icon name={"arrowDown"} />
          </div>
```


<details>
  <summary>which is achieved via ComboboxActivator.tsx</summary>

  ```
  return (
    <div
      role={accessibilityAttributes.role}
      aria-expanded={accessibilityAttributes.ariaExpanded}
      aria-controls={accessibilityAttributes.ariaControls}
      aria-label={accessibilityAttributes.ariaLabel}
      onClick={toggleOpen}
      onKeyDown={event => {
        if (event.key === "Enter" || event.key === " ") {
          toggleOpen();
        }
      }}
    >
      {props.children}
    </div>
  );
```
</details>

compared to

```
          {activatorAPI => (
            <div
              role={activatorAPI.role}
              tabIndex={0}
              aria-controls={activatorAPI.ariaControls}
              aria-expanded={activatorAPI.ariaExpanded}
              onClick={activatorAPI.toggleOpen}
              onKeyDown={e => {
                if (e.key === "Enter" || e.key === " ") {
                  activatorAPI.toggleOpen();
                }
              }}
              style={{
                display: "inline-flex",
                alignItems: "center",
                gap: "10px",
              }}
            >
              <Heading level={2}>Heading Two</Heading>
              <Icon name={"arrowDown"} />
            </div>
          )}
```

**tabindex, focus & keyboard navigation**
we could even move the `tabindex=0` into the div we're wrapping everything in for the first case. though there are some cases where that can get weird. if you provide a `button`, `input` or really any element where you want that thing to be the activator and it supports focus, then you'd have to tab twice to get to it since the wrapping div is focusable. that's specifically if you want that top level item to be the activator.

if you provide a layout where there's another button like a chip suffix to clear the selection or something, that's fine to have to tab twice because it's a distinct other interactive element.

I don't know if that's something people would want to do? provide a `button` that wraps a layout? honestly that seems redundant. the wrapping div we've provided is acting as the button. we have a firm opinion that the first element needs to be button-like and the activator.

I'm trying to imagine a case where you want the activator to be a piece inside of the activator, not the wrapping element. I don't think that makes sense.

image

in this case, if someone wanted to make the OPEN the focusable element that would trigger you wouldn't be able to. however I don't think this is a good design. your clear button has no need to be in the activator. just make it a separate element, you have access to the state/ability to clear it. there's no reason to put it.

**input**

input is the exception to all this. giving the div a tabindex makes that flow bad. whereas exposing all of the activator API would enable someone to make a much better experience.

the downside is the above code. you have to put it all together yourself.


**final thought**

to me it comes down to what we want to optimize for. do we want to optimize for what is essentially custom button layouts, where it's easy and we take care of most of it for you, with an emphasis on accessibility/keyboard features.

or do we want to optimize for full customizability, where you build whatever you want but it's more complicated, you need to read the instructions and it's possible to miss pieces that are required.

File: docs/components/Combobox/Combobox.stories.mdx
Comment (ZakaryH) [inline docs/components/Combobox/Combobox.stories.mdx:1]: I'm really skeptical of this approach.

Despite keeping it as brief as I thought was reasonable while still keeping it useful and descriptive, this is a decent amount of writing and instructions to go over.

I like the freedom and flexibility the render function offers. I don't like how much you have to read and implement yourself.

File: packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.tsx [hunk 6]
```diff
@@ -3,22 +3,23 @@ import { Button } from "@jobber/components/Button";
...
+  const { toggleOpen } = React.useContext(ComboboxContext);
+  const accessibilityAttributes = useComboboxActivatorAccessibility();
 
-  if (props.children.type === Button || props.children.type === Chip) {
+  if (
+    typeof props.children !== "function" &&
+    (props.children.type === Button || props.children.type === Chip)
+  ) {
     return React.cloneElement(props.children, {
-      role: "combobox",
```
MichaelParadis, line 12: Just a comment for myself and anyone who looks at this in the future.
This exact logic has been moved to `useMakeComboboxHandlers`
https://github.com/GetJobber/atlantis/pull/2075/files#diff-09225568c5cd891a5fb7a1d356821d17dee48e6ffa609dbda4de510ad6478162R33-R44

packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.tsx
MichaelParadis: I think this seems reasonable. I feel like always using a named render prop vs `children` will require us to have exceptions. Also another reason to use `children` in this case is that `Combobox.Activator` is explicitly how we allow consumers to define a custom activator (you can't use it without passing `children`). 

If people want the default activator they wouldn't pass a `Combobox.Activator` into into the`Combobox`. In the future we _could_ remove `Combobox.Activator` and move it into a named prop on the `Combobox` but I think that is not a good idea, especially right now

packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
```diff
@@ -61,20 +62,25 @@ describe("ComboboxActivator", () => {
...
+function renderComboboxActivator(
+  child: ReactElement | ((args: ComboboxCustomActivatorProps) => JSX.Element),
+  open: boolean,
+) {
   return render(
     <ComboboxContextProvider
-      setOpen={setOpen}
+      toggleOpen={toggleOpen}
       handleClose={handleClose}
       selectionHandler={jest.fn()}
```
MichaelParadis, line 85: Just a comment, I don't think `handleClose` needs to be mocked anymore. Looking at the tests we could remove change line 84 to be `handleClose={jest.fn()}` but this isn't blocking

packages/components/src/Combobox/constants.ts
```diff
@@ -0,0 +1 @@
+export const COMBOBOX_MENU_ID = "ATL-Combobox-Content";
```
MichaelParadis, line 1: It may be worth exporting this in `index.ts`. We have done that for Glimmer and Grid but no where else so it isn't major and us exporting Test ids  can be something we discuss as a team

MichaelParadis:
RE: [This](https://github.com/GetJobber/atlantis/pull/2075#issuecomment-2422945572) comment regarding the children approach. I feel that the current approach is preferable, while it is a bit more work for consumers and they need to read documentation about how to use it I feel that because this is for advanced usages, we should expect developers to handle the various edgecases that may arise (such as keyboard navigation, deciding when to open the Combobox, etc.). An example of this was when we decided to autocreate the sort labels on DataList but that ended up causing issues for us PR [here](https://github.com/GetJobber/atlantis/pull/1883).

> the simple way enables everything we've seen so far for custom Activator designs.

and while this is true right now, I think doing "magic" to get it working with limited configuration is likely to lead to developers having issues implementing new custom activator designs in the future (as our design language evolves)

docs/components/Combobox/Combobox.stories.mdx
```diff
@@ -50,14 +53,25 @@ activator.
...
 story.
 
-Omitting a custom activator subcomponent will result in the default activator
-being used.
+**Child Render Function:** If neither a Chip nor a Button is sufficient for your
+needs, you may instead provide a function as the child to `Combobox.Activator`.
+
+As arguments, the function will receive the toggle method to swap between open
+and closed states, as well as accessibility attributes required for a Combobox.
+_You must implement these yourself_. See the custom example for reference.
```
MichaelParadis, line 69:
```suggestion
**You must implement these yourself**. See the custom example for reference.
```
I think bolding this might make it "pop" out more

MichaelParadis: Just confirming the markdown I gave was valid

packages/components/src/Combobox/constants.ts
ZakaryH: fair point. I haven't checked to see if people are referencing this directly in product or not.

also related to your other idea - this should have some sort of unique identifier appended for the multiple combobox case. I suppose the base ID wouldn't change, assuming we do just append something unique anyway.

docs/components/Combobox/Combobox.stories.mdx
MichaelParadis: The markdown is valid if we want to change it
image

packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
ZakaryH: you're right. at the Activator level we are no longer concerned with `handleClose` being called directly. it's still exposed in the provider at all for other needs, but yeah there's no need for it to be a function we're storing in a variable anymore.

packages/components/src/Combobox/constants.ts
MichaelParadis: I think so too, we could allow the passing of a "testIDSuffix` for these types of components and append it and export the "base" test ids so they can get it by using the suffix they add. However this is way out of scope and a bigger discussion because there is probably a way better way to do this

docs/components/Combobox/Combobox.stories.mdx
MichaelParadis: The markdown I gave is valid

MichaelParadis (approved):
LGTM! No real blockers but some comments. I am happy to re-review if you change anything

docs/components/Combobox/Combobox.stories.mdx
ZakaryH: yeah I think I like that better. if we're going ahead with the DIY-approach it's pretty important to call out.

I started running out of heading levels and had to use bold as the `Child Render Function` bit to make it stand out which is probably a signal this document is getting a bit big.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -24,8 +25,8 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
   const template = (
     <div
       ref={popperRef}
-      id="ATL-Combobox-Content"
-      data-testid="ATL-Combobox-Content"
+      id={COMBOBOX_MENU_ID}
```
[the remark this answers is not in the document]
ZakaryH: ah. yeah I don't think there's any functional issues, at least I sure hope not since we have multiple comboboxes on a few pages now.

I was thinking more about testing, where if you use this selector it's going to match multiple. maybe that's a good thing though? consumers should try their very best to avoid testing Atlantis. a more valuable test would be to find a button that says "select thing" and then choose an option or something like that - rather than checking the internal ID to see if the combobox menu is showing up which is something we have unit/integration tests for already.

packages/components/src/Combobox/index.ts
```diff
@@ -1,3 +1,3 @@
 export * from "./Combobox";
 export { ComboboxContextProvider } from "./ComboboxProvider";
-export { ComboboxOption } from "./Combobox.types";
+export { ComboboxOption, ComboboxCustomActivatorProps } from "./Combobox.types";
```
MichaelParadis, line 3: Great catch, I didn't catch this initially 🎉

MichaelParadis (approved):
The new changes still look good

PR #2078, opened 2024-10-18
feat(components): Improve list composability [JOB-107035]

Aiden-Brine:
## Motivations

The List component is rarely used because it can currently only be used to render list items in a very particular format. If consumers were able to specify how the items are rendered it could be used in many more places. This PR adds that flexibility.

## Changes

The `content` prop in `ListItemProps` is now optional

### Added

Before these changes `ListItem` was used to render the items passed into `List`. This PR adds a `customRenderItem` prop to `List` that will take precedence over `ListItem`. If no `customRenderItem` is provided the behaviour with using `ListItem` should be unchanged.

### Changed

For existing `List` implementations there should be no change to functionality

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Some before and after photos from testing things using the pre-release:

### Selecting item type when adding a new item on the Schedule Alpha page
Before:
Screenshot 2024-10-22 at 11 32 24 AM

After:
Screenshot 2024-10-22 at 11 08 56 AM

### Selecting a line item on the same page
Before:
Screenshot 2024-10-22 at 11 32 35 AM

After:
Screenshot 2024-10-22 at 11 09 11 AM

### Promo stats [here](http://anchor.localhost.test:3000/administration/promotions)
Before:
Screenshot 2024-10-22 at 11 33 12 AM

After:
Screenshot 2024-10-22 at 11 05 26 AM

Christain Ellinger was kind enough to offer to test out the 5 instances in disputes as well, Also be sure to play around with my new stories in Storybook.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/List/List.stories.mdx
+++ b/docs/components/List/List.stories.mdx
@@ -21,8 +21,8 @@ find an object that leads to its full representation.
 
 The majority of list implementations can be achieved without needing to
 customize rendering. If your items conform to the `ListItemProps` interface, the
-component will render them automatically with the default behavior. The
-`ListItemProps` must be used when using default rendering.
+component will render them automatically with the default behavior. A prop of
+type `ListItemProps` must be used when using default rendering.
 
 #### Basic Flat List
 
@@ -97,13 +97,6 @@ their section and display them under section headers.
   </Card>
 </Canvas>
 
-#### Context Complexity
-
-The default full bleed (width) + bottom bordered style is preferred when the
-list item contains complex content where non-complex content would be something
-like Menu's content which has a different style but identical functionality
-otherwise.
-
 ### **Custom Rendering (Advanced Usage)**
 
 If you need more control over the rendering of list items, you can provide a
@@ -112,9 +105,6 @@ the default rendering of the `ListItem` component doesn’t meet your
 requirements. If a custom render function is provided onClick, url, and hover
 functionality will still work. Sectioned list functionality will also remain the
 same.
-
-Lastly, you will be required to implement the spacing, alignment and border
-bottom CSS logic.
 
 **Example: Custom Render Function with Custom Interface**
 
@@ -137,10 +127,10 @@ const items: ProductListItemProps[] = [
 
 
 const renderProductItem = (item: ProductListItemProps) => (
-    <div className="customContainer">
+    <Flex template={["shrink", "grow"]} align="start">
         <Text>{item.price}</Text>
         <Heading level={4}>{item.name}</Heading>
-    </div>
+    </Flex>
   </div>
 );
 
@@ -148,22 +138,6 @@ export const CustomRenderExample = () => (
   <List items={items} customRenderItem={renderProductItem} />
 );
 
-// *.module.css
-
-.customContainer {
-  display: flex;
-  alignItems: center;
-  justifyContent: flex-start;
-  text-align: left;
-}
-
-.customContainer > * {
-  padding: var(--space-small);
-}
-
-li > ul > li:not(:last-child) .customContainer {
-  border-bottom: var(--border-base) solid var(--color-border);
-}
 ```
 
 **In this example**:
@@ -172,7 +146,6 @@ li > ul > li:not(:last-child) .customContainer {
   `BaseListItemProps`.
 - The `customRenderItem` function provides complete control over how each item
   is rendered.
-- We replicate the padding, alignment and border logic
 
 **When to Use Custom Rendering:**
 
@@ -181,12 +154,22 @@ li > ul > li:not(:last-child) .customContainer {
 - Create your own interface extending `BaseListItemProps` to define the fields
   you need.
 
-##### Replicating Default List Styles
+#### Using Your own Custom Item Styles
 
-In the above example we must use `li > ul > li:not(:last-child) .{customClass}`
-because it is a sectioned list having nested ul and li elements.
+Some styles (such as hover and active item styles) are always applied to list
+items, even when using a custom render function. However, other styles (like
+padding, alignment, and border-bottom) are applied by default but can be
+disabled by setting `customItemStyles={true}` alongside `customRenderItem`. This
+option allows you to remove some of the default item styles and apply your own.
 
-For a non sectioned list `li:not(:last-child) .{customClass}` would suffice.
+For instance, in the Simple List with Custom Styles example, setting
+`customItemStyles={true}` allows for the creation of a list with rounded corners
+and no border-bottom. You can also experiment by toggling this prop in the other
+custom render examples to observe its impact.
 
-For the spacing, it is achieved with `small` padding on the item container, then
-an additional `small` padding on the elements inside.
+Note: The `customItemStyles` prop is only valid when a `customRenderItem`
+function is provided.
+
+```tsx
+
+```
--- a/docs/components/List/List.stories.mdx
+++ b/docs/components/List/List.stories.mdx
@@ -20,41 +20,11 @@ like billing or activity feed. The main purpose of a list is to help the user
 find an object that leads to its full representation.
 
 The majority of list implementations can be achieved without needing to
-customize rendering. If your items conform to the ListItemProps interface, the
-component will render them automatically with the default behavior.
+customize rendering. If your items conform to the `ListItemProps` interface, the
+component will render them automatically with the default behavior. The
+`ListItemProps` must be used when using default rendering.
 
-#### ListItemProps Interface
-
-When using the default rendering, your items must conform to the
-**`ListItemProps`** interface. Here’s a breakdown of the properties:
-
-**Required:**
-
-- **`id`**: A unique identifier for the list item. This is essential to ensure
-  proper rendering and handling of batch actions.
-
-**Optional:**
-
-- **`title`**: A heading or title for the item.
-
-- **`caption`**: Subdued text shown below the main content.
-
-- **`content`**: Main content, either a string or an array of strings. Supports
-  basic markdown (e.g., **bold** and _italic_).
-
-- **`icon`**: An icon to display on the left side of the content.
-
-- **`iconColor`**: Sets the color of the icon.
-
-- **`isActive`**: Highlights the list item to indicate it needs attention.
-
-- **`section`**: Used to group the list items into sections.
-
-- **`url`**: Makes the item clickable, rendering it as a link.
-
-- **`value`**: A numeric or string value aligned to the right of the item.
-
-- **`onClick`**: A callback function triggered when the item is clicked.
+**`ListItemProps`** interface.
 
 #### Basic Flat List
 
--- a/docs/components/List/List.stories.mdx
+++ b/docs/components/List/List.stories.mdx
@@ -24,8 +24,6 @@ customize rendering. If your items conform to the `ListItemProps` interface, the
 component will render them automatically with the default behavior. The
 `ListItemProps` must be used when using default rendering.
 
-**`ListItemProps`** interface.
-
 #### Basic Flat List
 
 <Canvas>
@@ -103,7 +101,11 @@ their section and display them under section headers.
 
 If you need more control over the rendering of list items, you can provide a
 **custom render function** using the `customRenderItem` prop. This is useful if
-the default `ListItem` component doesn’t meet your requirements.
+the default rendering of the `ListItem` component doesn’t meet your
+requirements. If a custom render function is provided onClick, url, and hover
+functionality will still work. Sectioned list functionality will also remain the
+same. Lastly, the styles wrapping the list items will match those of a `List`
+with no custom render function.
 
 **Example: Custom Render Function with Custom Interface**
 
@@ -113,11 +115,15 @@ import { List } from "./List";
 interface ProductListItemProps extends BaseListItemProps {
   readonly name: string;
   readonly price: string;
+  readonly section: string;
+  readonly onClick?: () => void;
 }
 
 const items: ProductListItemProps[] = [
-  { id: 1, name: "Lawn Mower", price: "$250.00" },
-  { id: 2, name: "Hedge Trimmer", price: "$85.00" },
+  { id: 1, name: "Lawn Mower", price: "$250.00", section: "Garden Tools", onClick: () => alert("Lawn Mower clicked") },
+  { id: 2, name: "Hedge Trimmer", price: "$85.00", section: "Garden Tools", onClick: () => alert("Hedge Trimmer clicked") },
+  { id: 3, name: "Drill", price: "$50.00", section: "Power Tools", onClick: () => alert("Drill clicked") },
+  { id: 4, name: "Screwdriver", price: "$10.00", section: "Hand Tools", onClick: () => alert("Screwdriver clicked") },
 ];
 
 const renderProductItem = (item: ProductListItemProps) => (
--- a/docs/components/List/List.stories.mdx
+++ b/docs/components/List/List.stories.mdx
@@ -1,9 +1,178 @@
 import { Meta } from "@storybook/addon-docs";
+import { Card } from "@jobber/components/Card";
+import { List } from "@jobber/components/List";
 
 <Meta title="Components/Lists and Tables/List" />
 
 # List
 
-Displays a collection of objects that are either the same type or related to one
-another, like billing or activity feed. The main purpose of a list is to help
-the user find an object that leads to its full representation.
+The `List` component renders a collection of items in a structured way. It
+supports both [flat lists](#basic-flat-list) and
+[sectioned lists](#sectioned-list), and allows for
+[custom rendering](#custom-rendering-advanced-usage) of list items via a render
+function.
+
+## Design and Usage Guidelines
+
+The `List` component is a great solution if you are looking to display a
+collection of objects that are either the same type or related to one another,
+like billing or activity feed. The main purpose of a list is to help the user
+find an object that leads to its full representation.
+
+The majority of list implementations can be achieved without needing to
+customize rendering. If your items conform to the ListItemProps interface, the
+component will render them automatically with the default behavior.
+
+#### ListItemProps Interface
+
+When using the default rendering, your items must conform to the
+**`ListItemProps`** interface. Here’s a breakdown of the properties:
+
+**Required:**
+
+- **`id`**: A unique identifier for the list item. This is essential to ensure
+  proper rendering and handling of batch actions.
+
+**Optional:**
+
+- **`title`**: A heading or title for the item.
+
+- **`caption`**: Subdued text shown below the main content.
+
+- **`content`**: Main content, either a string or an array of strings. Supports
+  basic markdown (e.g., **bold** and _italic_).
+
+- **`icon`**: An icon to display on the left side of the content.
+
+- **`iconColor`**: Sets the color of the icon.
+
+- **`isActive`**: Highlights the list item to indicate it needs attention.
+
+- **`section`**: Used to group the list items into sections.
+
+- **`url`**: Makes the item clickable, rendering it as a link.
+
+- **`value`**: A numeric or string value aligned to the right of the item.
+
+- **`onClick`**: A callback function triggered when the item is clicked.
+
+#### Basic Flat List
+
+<Canvas>
+  <Card>
+    <List
+      items={[
+        {
+          id: 1,
+          content: "Content",
+          icon: "work",
+          iconColor: "interactive",
+          title: "Title",
+          value: "Value",
+          onClick: () => alert("First list item clicked"),
+        },
+        {
+          id: 2,
+          content: "Content",
+          icon: "work",
+          iconColor: "interactive",
+          title: "Title",
+          value: "Value",
+          onClick: () => alert("Second list item clicked"),
+        },
+      ]}
+    />
+  </Card>
+</Canvas>
+
+#### Sectioned List
+
+If a section field is present in any item, the list will group the items by
+their section and display them under section headers.
+
+<Canvas>
+  <Card>
+    <List
+      items={[
+        {
+          id: 1,
+          content: "Content",
+          icon: "work",
+          iconColor: "interactive",
+          title: "Title",
+          value: "Value",
+          section: "Section 1",
+          onClick: () => alert("First list item clicked"),
+        },
+        {
+          id: 2,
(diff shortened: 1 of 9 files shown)


docs/components/List/List.stories.mdx
```diff
@@ -1,9 +1,178 @@
...
+
+- **`isActive`**: Highlights the list item to indicate it needs attention.
+
+- **`section`**: Used to group the list items into sections.
+
+- **`url`**: Makes the item clickable, rendering it as a link.
+
+- **`value`**: A numeric or string value aligned to the right of the item.
+
+- **`onClick`**: A callback function triggered when the item is clicked.
```
ZakaryH, line 57: not sure how I feel about writing out all the props in our docs. we have all this documented in the JSDocs already, it's just that our tooling doesn't expose it in Storybook

image

on one hand, it is nice to have it available in storybook, on the other hand it's adding something we have to make sure to keep in sync going forward. additionally it means there are 3 places to find props of a component
- component docs
- component examples page
- JS Docs

I'm tempted to say we should avoid duplicating documentation and instead find a solution to nested props to leverage the JSDocs (in general , definitely not in this ticket!)

packages/components/src/List/ListItem.tsx
```diff
@@ -9,7 +9,15 @@ import { Typography } from "../Typography";
 import { Markdown } from "../Markdown";
 import { Emphasis } from "../Emphasis";
 
-export interface ListItemProps {
+export interface BaseListItemProps {
+  /**
+   * The ID of the list item. This will be helpful to know the selected list
+   * items when a batch action is implemented.
```
ZakaryH, line 15: I know this is a copy/paste of an existing JSDoc, but I'm not sure I get what this is trying to say. don't we need an id all the time for arrays in React and not just for batch actions? actually what even is a batch action in this context...

docs/components/List/List.stories.mdx
Aiden-Brine: You make a good point, I was being a bit overzealous but it does add another layer to the maintainability so I will remove them.

packages/components/src/List/ListItem.tsx
Aiden-Brine: Yeah, good question 🤔 I didn't pay it much attention when I copied it over. I will update it to something that makes more sense.

ZakaryH:
I've been thinking about something we talked about last week, if the consumer is the one who wants to add a border bottom to their list items, how would they avoid the double border bottom?

```
      <List
        items={myAdvancedOptions}
        customRenderItem={item => (
          <div style={{ display: "flex", borderBottom: "1px solid #ddd" }}>
            <Text>{item.text}</Text>
            <span style={{ marginLeft: "auto" }}>{item.icon}</span>
          </div>
        )}
      />
```

image

the argument we give to the render function doesn't say if it's the first, last or middle item. without us providing it, they have to figure it out themselves.

the only way you can safely do it with a list whose order isn't static would be with CSS imo. you'd likely add a custom classname to the `div` which has the border bottom, and then say `.myCustomItemClass:last-child { border: none }`

that's not the worst honestly, but I am wondering about if there's a nice way for us to expose that internal state/knowledge the List component has to each of the items. what if the modification you want to do to the last item isn't CSS, and you can't just use `last-child`?

Aiden-Brine:
@ZakaryH you raise an interesting question. I previously was just thinking along the lines of CSS following the approach you mentioned "you'd likely add a custom classname to the div which has the border bottom, and then say .myCustomItemClass:last-child { border: none }". I hadn't thought about non-CSS use cases. A couple of options:

1) We change the typing for `customRenderItem` so that it takes in an optional `index` argument. Then List just always passes the `index` into `customRenderItem` when it calls it. If the consumer wrote their custom render function to use the index they will be provided it, if not, their render function will just ignore it

2) If we thought there are likely to be other pieces of info that the render function could need in the future we could say that the custom render function takes an options object like:
```
interface CustomRenderOptions<T> {
  item: string | number;
  index: number?;
  future options...
}
```

I think option 2 is definitely overkill right now, especially because of how simple List is I'm not sure what options we could possibly want to pass along with index in the future. I am on the fence about going option 1 or saying this is over engineering 🤔 Adding the optional index prop would be easy and improve the composability but it also adds one more piece of complexity and I'm not sure it is worth it. What do you think?

ZakaryH:
@Aiden-Brine the more I think about this the more I'm talking myself out of it.

as a consumer you'll always have access to the data you're passing in. you'll have an opportunity to `map` over it and maybe add your own `index` or `isLasItem`, since we're not trying to do anything clever inside `List`. it simply maps over what was provided and renders it, meaning the order is entirely predictable/knowable.

with an extended type for `items` you can add whatever you want... so seems like you'd be able to solve a non-CSS last item problem yourself.

I think if we start seeing a repeated trend of people solving exclusively for the last item, then we should consider solving for it at a component level. so yeah, let's wait until it's an actual problem. we're not painting ourselves into a corner or making an unsolvable mess by waiting.

Aiden-Brine:
@ZakaryH I'm on board with that. We could always add an optional prop to the render function if we wanted to in the future without much work or blowing things up.

Aiden-Brine:
@chris-at-jobber there are no design code changes for you to review but I tagged you in case you wanted to look at the two new [Storybook stories](https://job-107035-add-custom-render.atlantis.pages.dev/?path=/story/components-lists-and-tables-list-web--list-with-custom-renderer) I added based on your mockup

chris-at-jobber:
@ZakaryH to the divider/avoiding a "double bottom border" question - is _that_ aspect something List _should_ have opinions about? ie 

Either you want a divider, or you don't
- defaulting to having dividers
- If the divider is on a section header, use `color-border--section`

?

ZakaryH:
>to the divider/avoiding a "double bottom border" question - is that aspect something List should have opinions about? ie

>Either you want a divider, or you don't

>defaulting to having dividers
If the divider is on a section header, use color-border--section
?

@chris-at-jobber  do you mean the default version? that already has borders and manages them.

image

if you're referring to custom markup for list items, I'm hesitant to bake that in as behavior the List provides - unless we have strong opinions about the style/appearance of the border.

as for the section headers, in the case of custom markup we won't know much of anything about what they gave us. we can't really know if it's a header, a normal item or something else entirely. we're leaving that up to them to sort out.

chris-at-jobber:
@ZakaryH sorry, I know that List already handles borders but we also have known, planned, instances that we won't want dividers on every element (Combobox, Menu) - just wondering how/if we will solve that!

ZakaryH:
building out various custom lists and seeing how it feels

**Sectioned List**

image

code to achieve this is... fine. I don't love the type narrowing needed to handle Icons names though.

```
  const myItems: MyListItemProps[] = [
    {
      id: 0,
      sectionTitle: "Country",
    },
    {
      id: 1,
      name: "Canada",
      value: "CA",
      icon: "checkmark",
    },
    {
      id: 2,
      name: "America",
      value: "USA",
      icon: "truck",
    },
    {
      id: 3,
      name: "Mexico",
      value: "MX",
      icon: "quote",
    },
  ];

  return (
    <Page title={header}>
      <Content spacing="large">
        <List
          items={myItems}
          customRenderItem={(item: MyListItemProps) => <ListItem {...item} />}
        />
      </Content>
    </Page>
  );
}

interface MyListItemProps extends BaseListItemProps {
  id: number;
  name?: string;
  value?: string;
  icon?: IconNames;
  sectionTitle?: string;
}

function ListItem(props: MyListItemProps) {
  if (props.sectionTitle) {
    return (
      <div style={{ borderBottom: "1px solid #ccc" }}>
        <strong>{props.sectionTitle}</strong>
      </div>
    );
  }

  if (props.icon) {
    return (
      <div data-value={props.value}>
        {props.name}
        <Icon name={props.icon} />
      </div>
    );
  }

  return <></>;
}
```

packages/components/src/List/List.tsx
```diff
@@ -2,20 +2,41 @@ import React from "react";
...
-  const isSectioned = items.some(item => item.section);
+export function List<T extends BaseListItemProps = ListItemProps>({
+  items,
+  customRenderItem,
+}: ListProps<T>) {
+  if (customRenderItem) {
+    return (
+      <ul className={styles.list}>
+        {items.map(item => (
+          <li key={item.id}>{customRenderItem(item)}</li>
```
ZakaryH, line 34: image

one thing I'm realizing is that we will need access to the li to pass in all sorts of attributes 😭 

let's imagine we remove the `li` and pass everything to the custom render method... is that ok? we'd have to wrap it in a fragment, and then the consumer has to implement `key` so React doesn't get lost. bleh. that's not ideal.

packages/components/src/List/ListItem.tsx
```diff
@@ -73,68 +76,77 @@ export interface ListItemProps {
...
       className={actionClasses}
-      href={url}
-      onClick={onClick}
+      href={props.url}
+      onClick={props.onClick}
       {...buttonProps}
     >
-      {icon && (
+      {props.customRenderItem ? (
+        props.customRenderItem(props as T)
```
Aiden-Brine, line 108: This typecast is currently necessary becuase `customRenderItem` is expecting something of type `T` and it is getting something of type `T & {readonly customRenderItem?: (item: T) => React.ReactNode}`. It's not the worst type cast ever because we know it will have all the properties of `T` but if someone knows of a cleaner solution I am open to it.

ZakaryH:
explanation of `move styles around to give customRender a blank slate`

with the current List, the items are complex enough that we do the full-width + border bottom on it. 

image


this is different from the simple, padded, rounded list style when the content is minimal
image

**Focus/Hover State**

currently this is all provided by the button's `.hoverable` class and `.isActive`

that's all well and good for the full width version, however if you want the rounded version with your custom markup we actually want that background to be on that element, not the one behind it otherwise you'd see the not-rounded edges peeking out behind the rounded ones.

so instead, I've made the selector apply that to the first child of `.isActive` or `.hoverable` which is either the default container I've added, OR the custom markup provided.

**button padding**

we want the button. that's part of the value of List. it handles links and clickable items along with their hover/focus styles.

however we don't necessarily want the padding. the button wrapping our custom markup means we can't use the full width. with that, we've gotta remove it in the custom case and if you want it, add it yourself.

**Border Bottom on items**

once again, this is part of the appearance for complex list items. we can't guarantee the custom content will be complex, so we should only apply it in the default case.

it does mean that someone providing custom markup will have to add their own border bottom, but I don't see any way around that. we just won't know what they're giving us. is it text and an icon, or a whole bunch of stuff?

**Why now?**

you could make the argument this is a requirement of Menu implementing List and we should do it at that point.

the reason I want to do it now is because it changes the canvas you are given for the custom markup, and I can see those limitations affecting how you implement custom markup. in the off chance someone uses this between now and then, it would become somewhat breaking change.

ZakaryH:
this caught my eye while looking at border radius stuff. it's 100% existing and not something we've introduced.

since the section header doesn't have rounded corners, but the parent `Card` does - it cuts out the little intersection between them.

image

super minor and an interesting problem that would exist with any Card content that both has a background color AND has un-rounded borders 🤔

docs/components/List/Web.stories.tsx
```diff
@@ -20,63 +32,258 @@ const BasicTemplate: ComponentStory<typeof List> = args => (
...
+    justify-content: flex-start;
+    text-align: left;
+    padding: var(--space-small);
+  }
+
+  .product-list-item > * {
+    padding: var(--space-small);
+  }
+
+  li > ul > li:not(:last-child) .product-list-item {
```
ZakaryH, line 258: this is a bit awkward. we've gotta increase the specificity of what we're targeting with `li > ul >` rather than just `li:not(:last-child)` because the structure is 

```
<ul>
  <li>
    <ul>
      <li></li>
    </ul>
  </li>
  <li>
    <ul>
      <li></li>
    </ul>
  </li>
</ul>
```

so we have to say that we want to target the last item in each set of nested `ul`s

if we don't then it will only avoid adding a border bottom to the very last instance, meaning the li right above a new section **would** get a border bottom

docs/components/List/List.stories.mdx
```diff
@@ -1,9 +1,175 @@
...
+        },
+      ]}
+    />
+  </Card>
+</Canvas>
+
+#### Sectioned List
+
+If a section field is present in any item, the list will group the items by
+their section and display them under section headers.
```
ZakaryH, line 59: one thing that I discovered when playing around with Sectioned Lists is that giving a `section` basically sets a flag saying "I am now in section mode"

since `section` is optional, it's possible to not have a `section` in each of your items.

the currently undocumented behavior that happens in this case is that you get a hard coded header of `Other`

for the existing example I've removed the `section` from something in the middle, and added one more entry at the end that has no `section`

image

it'll group everything with the first occurrence of something missing a `section`

I did some digging around in the product and I'm about 98% sure that all the cases where we use a `section` it will never have a missing `section`.

I had no idea it did this and it's kind of debatable if it even should..... but that's another topic. at the very least, let's document the behavior.

maybe something like 

> If even one `section` key exists in any of `items` then it will be expected that every object within that collection will have a `section` defined. If a `section` is missing, "Other" will be used, and any "Other" items will be grouped with that first occurrence that has no `section`.

docs/components/List/List.stories.mdx
```diff
@@ -1,9 +1,175 @@
...
+- We define a custom interface (`ProductListItemProps`) that extends
+  `BaseListItemProps`.
+- The `customRenderItem` function provides complete control over how each item
+  is rendered.
+
+**When to Use Custom Rendering:**
+
+- Use a custom render function when you need complex layouts that the `ListItem`
+  component can’t provide.
+- Create your own interface extending `BaseListItemProps` to define the fields
```
ZakaryH, line 154: nitpick: this is less about "when to use custom rendering" and more about how

maybe a heading of `About Custom Rendering` would encapsulate both of these points. the first describing "when" to use it, and the second "how"

docs/components/List/Web.stories.tsx
```diff
@@ -19,64 +31,294 @@ const BasicTemplate: ComponentStory<typeof List> = args => (
...
+  listItem,
+}: {
+  readonly listItem: SPListItemProps;
+}) {
+  return (
+    <Flex template={["shrink", "grow", "shrink"]} align="center">
+      <Avatar imageUrl={listItem.avatarUrl} />
+      <Heading level={4}>{listItem.name}</Heading>
+      <Icon name="arrowRight" size="base" color="interactive" />
+    </Flex>
```
ZakaryH, line 153: I'm very happy with how we ended up getting this to work!

being able to throw in a `Flex` with your layout and not fuss with any padding, hover, border and all that is really great.

packages/components/src/List/List.tsx
```diff
@@ -2,44 +2,91 @@ import React from "react";
...
    */
-  readonly items: ListItemProps[];
+  readonly items: T[];
+  /**
+   * A function that will be called for each item instead of the default
+   * rendering
+   */
+  readonly customRenderItem?: (item: T) => React.ReactNode;
+  /**
+   * Set to true for more control over the item styles
```
ZakaryH, line 25: let's call out that it only works in conjunction with `customRenderItem`

packages/components/src/List/List.tsx
```diff
@@ -2,44 +2,91 @@ import React from "react";
...
+function SectionedList<T extends BaseListItemProps = ListItemProps>({
+  items,
+  customRenderItem,
+  customItemStyles,
+}: ListProps<T>) {
   const sectionedItems = groupBy(items, item => get(item, "section", "Other"));
   const sectionHeaderClassNames = classnames(sectionStyles.sectionHeader);
 
+  const omitDefaultStyles = customRenderItem && customItemStyles;
+  const itemClassNames = classnames(!omitDefaultStyles && styles.item);
```
ZakaryH, line 88: I'm tempted to reduce this repetition since we're doing the exact same thing in `DisplayList`. though, it's 2 place, which isn't that bad at all.... plus what, it would save us 1 line of code in each component.

ehh. it's probably fine. I don't have a very compelling argument to do it right now.

packages/components/src/List/ListItem.module.css
```diff
@@ -1,18 +1,28 @@
...
-  display: flex;
   width: 100%;
-  box-sizing: border-box;
-  padding: var(--space-small);
+  padding: 0;
   border: none;
-  text-align: left;
-  text-decoration: none;
   background-color: transparent;
-  transition: background-color var(--timing-base);
```
ZakaryH, line 10: just noticing this is gone now

gonna check vs master to see if the background color timing is noticeable different now, or if this was strictly needed for `.action` when it was on the `button`

packages/components/src/List/ListItem.tsx
ZakaryH: following up on this, even after all the time we've spent in List I have no idea what a "batch action" means in this context

my best guess is that you want to be using this in a selection context, and onClick you want to know the id of the thing you clicked to add it to a batch of items you wanna do _something_ with

however I'm of the opinion `List` is a really bad fit for selection style interactions, at least in its current state since we don't pass `id` as an arg to `props.onClick` meaning you'd have to dig through the `event.currentTarget` to get it which is gross.

packages/components/src/List/ListItem.tsx
```diff
@@ -9,7 +9,15 @@ import { Typography } from "../Typography";
 import { Markdown } from "../Markdown";
 import { Emphasis } from "../Emphasis";
 
-export interface ListItemProps {
+export interface BaseListItemProps {
+  /**
+   * The ID of the list item. This will be helpful to know the selected list
+   * items when a batch action is implemented.
+   */
+  readonly id: number | string;
```
ZakaryH, line 17: something I've been thinking about using more often is `React.Key` which with our current version of react-types is exactly this

in newer versions it has `bigint` as another possible value, but I don't think that's a problem as long as we don't try to use it as an index.

you know what, this is fine. if we wanna start using `React.Key` more, I wanna put a tiny bit more thought into it.

packages/components/src/List/ListItem.tsx
Aiden-Brine: This all makes sense. Thank you for following up, with the other more major refactoring comments I missed addressing this one but will do so now

packages/components/src/List/List.tsx
Aiden-Brine: I thought about this as well, I landed on the fact that it would be preferable to have the interface between the components be simpler (since they are already so complicated) and live with the duplicated line

packages/components/src/List/ListItem.module.css
chris-at-jobber: Why is this gone now?

ZakaryH: confirmed that we do in fact need it. must have gotten lost in the shuffle.

we want that all the time no matter what, since it's part of the hover/focus styles so I think we need to apply it on a class that is always there.

```
.action > * {
  transition: background-color var(--timing-base);
}

```

packages/components/src/List/ListItem.module.css
```diff
@@ -1,18 +1,28 @@
...
   appearance: none;
+}
+
+.defaultContainer > * {
+  display: flex;
+  box-sizing: border-box;
+  text-align: left;
+  flex: 1;
   align-items: flex-start;
```
ZakaryH, line 14: I noticed on the custom example with the SP avatars that it's not centering properly

`Flex` is trying to apply `align-items: center` but this is more specific and overriding that.

Components___Lists_and_Tables___List___Web_-_List_With_Custom_Renderer_⋅_Storybook

the `flex-start` is an existing rule from when all of this was on the `.action` class that we apply to the `button`

I'm not 100% sure I understand why that's needed. I suppose because the default value is `normal`?

image

>[normal](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items#normal)
The effect of this keyword is dependent of the layout mode we are in:

>In absolutely-positioned layouts, the keyword behaves like start on replaced absolutely-positioned boxes, and as stretch on all other absolutely-positioned boxes.
In static position of absolutely-positioned layouts, the keyword behaves as stretch.
For flex items, the keyword behaves as stretch.
For grid items, this keyword leads to a behavior similar to the one of stretch, except for boxes with an [aspect ratio](https://developer.mozilla.org/en-US/docs/Glossary/Aspect_ratio) or an intrinsic sizes where it behaves like start.
The property doesn't apply to block-level boxes, and to table cells.

ZakaryH: I'm thinking this might be another piece similar to the previous ` > * { padding: var(--space-small) }`

where it makes sense with the exact static layout and content we had, but we can't possibly know if it makes sense for the custom stuff we've received since it's very reasonable to use flex to position your custom content

perhaps we move it inside `.paddedChild` which is probably not the best name anymore. maybe `.defaultContent` instead?

it does make me wonder if there are any other attribute in `defaultContainer > *` that don't belong...

ZakaryH:
@Aiden-Brine I thought about your question around if `text-align: left` should be in `.defaultContent`

the reason it was on `.action` is because `button` elements have a `text-align: center` from the user-agent stylesheet

image

while this is still technically a button, we are treating it more like a row of content where I'd argue having it be center aligned would be an outlier.

with that, I think we should put `text-align: left` back in `.action` so that you'll have the standard alignment by default. if you do want something centered, you can do that in your custom markup with an additional rule

----------

looking at what's left in `.defaultContainer` we have: `display: flex`, `box-sizing: border-box`, `flex:1`

and in `.defaultContent` there's a lonely `align-items: flex-start` which only makes sense with flex. we are giving it flex, but in a kind of non obvious way via `.defaultContainer > *`

looking at our example usages it's also weird that we're giving a `Flex` component `display: flex`.`Flex` doesn't actually use `flex` haha

image

so we're trying to apply it though it's not doing anything since `grid` wins begging the question of why are we trying to make the children flex. the way we have this set up, people need to provide a wrapping container for their custom markup. I'd say it's fine to leave up to them to make it flex/grid/block whatever they want really.

`border-box` is somewhat useful but I don't see us needing to set that for everything.

TL;DR - I think all these can safely be moved into `.defaultContent`, allowing us to delete `.defaultContainer`, the React classname conditional, and `omitDefaultStyles`

ZakaryH (approved):
LGTM

PR #2083, opened 2024-10-23
refactor(components): Replace ToggleOpen with HandleOpen

ZakaryH:
## Motivations

Combobox doesn't allow you to actually interact with the Activator again once the combobox is open. Maybe some day it will, but that day is not today.

With that, the `toggleOpen` method is a bit misleading. in practice it is a one way toggle that can only ever toggle from closed -> open.

the `toggleOpen` method has a dead code path that only exists in theory. functionally, you have no way to interact with the activator while the combobox is open. we will close it when you click outside the combobox, when you press ESC, when you interact with an Action that has the default close on click, or when you make a selection in single select mode.

the only way you can interact with it would be through DOM manipulation which is not realistic.

it's similar to how we handle the overlay

```
        {open && (
          <div
            className={styles.overlay}
            onClick={() => handleClose()}
            data-testid="ATL-Combobox-Overlay"
          />
        )}
```

this one is more explicit since the condition ensures we'd never even have this element rendered if it was closed, but there conceptually it's the same. we know that in practice, you can't interact with the activator once the combobox is open.

### Added

- 

### Changed

the exposed method from the API is now named `open`

### Deprecated

- 

### Removed

remvoed some tests that were checking cases that aren't real, or doing interactions no user could do

### Fixed

- 

### Security

- 

## Testing

play around with some normal comboboxes, custom Chip/Button activator comboboxes and finally the fully customized activator combobox.

all of them should still open when the activator is interacted with, and close as expected when clicking outside, pressing ESC, making selections in single select mode etc.

it will even look like it closes when you click the activator while it's open, just know that's the `handleClose` on the overlay doing it in reality.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -52,11 +52,6 @@ describe("Combobox", () => {
       expect(screen.getByText("Add Teammate")).toBeInTheDocument();
     });
 
-    it("should close the menu when clicking the activator", async () => {
-      await userEvent.click(screen.getByRole("combobox"));
-      expect(screen.getByTestId(MENU_TEST_ID)).toHaveClass("hidden");
-    });
-
     it("should close the menu when clicking outside the menu (which clicks the overlay)", async () => {
       await userEvent.click(screen.getByTestId(OVERLAY_TEST_ID));
       expect(screen.getByTestId(MENU_TEST_ID)).toHaveClass("hidden");
--- a/packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
+++ b/packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
@@ -1,43 +1,30 @@
 import React, { ReactElement } from "react";
 import { render } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 import { Button } from "@jobber/components/Button";
 import { Chip } from "@jobber/components/Chip";
 import { ComboboxActivator } from "./ComboboxActivator";
 import { ComboboxContextProvider } from "../../ComboboxProvider";
 import { ComboboxCustomActivatorProps } from "../../Combobox.types";
 
-const toggleOpen = jest.fn();
+const handleOpen = jest.fn();
 
 afterEach(() => {
-  toggleOpen.mockClear();
+  handleOpen.mockClear();
 });
 
 describe("ComboboxActivator", () => {
-  describe("when open is true", () => {
-    it("should call toggleOpen when the activator is clicked", () => {
-      const { getByRole } = renderComboboxActivator(
-        <Button label="Teammates" />,
-        true,
-      );
-      const activator = getByRole("combobox");
-
-      activator.click();
-
-      expect(toggleOpen).toHaveBeenCalled();
-    });
-  });
-
   describe("when open is false", () => {
-    it("should call toggleOpen when the activator is clicked", () => {
+    it("should call handleOpen when the activator is clicked", async () => {
       const { getByRole } = renderComboboxActivator(
         <Button label="Teammates" />,
         false,
       );
       const activator = getByRole("combobox");
 
-      activator.click();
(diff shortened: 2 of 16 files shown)


packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -52,11 +52,6 @@ describe("Combobox", () => {
       expect(screen.getByText("Add Teammate")).toBeInTheDocument();
     });
 
-    it("should close the menu when clicking the activator", async () => {
-      await userEvent.click(screen.getByRole("combobox"));
-      expect(screen.getByTestId(MENU_TEST_ID)).toHaveClass("hidden");
-    });
-
```
ZakaryH, line 59: this test was doing something users cannot do by finding the DOM element and programatically clicking it

packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.test.tsx
```diff
@@ -1,43 +1,30 @@
...
-        <Button label="Teammates" />,
-        true,
-      );
-      const activator = getByRole("combobox");
-
-      activator.click();
-
-      expect(toggleOpen).toHaveBeenCalled();
-    });
-  });
```
ZakaryH, line 28: another test doing something a user can't

packages/components/src/Combobox/components/ComboboxActivator/ComboboxActivator.tsx
```diff
@@ -15,10 +15,10 @@ export function ComboboxActivator(props: ComboboxActivatorProps) {
   ) {
     return React.cloneElement(props.children, {
       role: accessibilityAttributes.role,
-      onClick: toggleOpen,
+      onClick: handleOpen,
     });
   } else if (typeof props.children === "function") {
-    return props.children({ ...accessibilityAttributes, toggleOpen });
+    return props.children({ ...accessibilityAttributes, open: handleOpen });
```
ZakaryH, line 21: I like `open()` better than `handleOpen()` when considering this will be used by a consumer.

packages/components/src/Combobox/components/ComboboxTrigger/ComboboxTrigger.test.tsx
```diff
@@ -5,30 +5,20 @@ import { ComboboxTrigger } from "./ComboboxTrigger";
...
+    it("calls handleOpen", async () => {
       renderTrigger();
       const trigger = screen.getByRole("combobox");
 
       await userEvent.click(trigger);
-      expect(toggleOpen).toHaveBeenCalled();
-    });
-  });
-
-  describe("when open is true", () => {
```
ZakaryH, line 25: another test that is doing something a user can't

packages/components/src/Combobox/hooks/useMakeComboboxHandlers.ts
```diff
@@ -34,14 +34,6 @@ export function useMakeComboboxHandlers(
     setOpen(true);
   }, [setOpen]);
 
-  const toggleOpen = useCallback(() => {
```
ZakaryH, line 37: the only place this was used was in Trigger & Activator, both of which can't actually be interacted with while `open` is `false` making the first code path inaccessible.

if the first code path is not real, then we're left with `handleOpen` and we might as well use that and get rid of `toggleOpen`

packages/components/src/DataList/components/DataListFilters/components/DataListSort/DataListSort.test.tsx
```diff
@@ -81,7 +81,7 @@ describe("DataListSort", () => {
     });
 
     it("should close the popover when the button is clicked again", async () => {
-      await userEvent.click(screen.getByRole("combobox"));
+      await userEvent.click(screen.getByTestId("ATL-Combobox-Overlay"));
```
ZakaryH, line 84: I'm fixing this quickly but I'm not sure how I feel about this set of tests in DataListSort. it's testing Combobox behavior that is already tested elsewhere.

PR #2088, opened 2024-10-25
feat(components): Add support for controlled disclosures

MichaelParadis:
## Motivations

The disclosure isn't able to be used as a controlled component as a result there wasn't a way to open disclosures based on other user interactions. It also wasn't possible to know if a disclosure had been toggled

### Added

- Added `onToggle` prop to `Disclosure`
- Added ability to control the `Disclosure` component

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Can be tested in storybook. Verify that the controlled story allows the state of the disclosure to be controlled.
Verify that the uncontrolled stories work the same as before. Also verify default open works for uncontrolled disclosures

Changes can be

---

Random photo of Atlantis


packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -27,35 +24,35 @@ interface DisclosureProps {
...
+  opened,
 }: DisclosureProps) {
-  const [isOpen, setOpen] = useState(defaultOpen);
-  const [isMounted, setMount] = useState(false);
-  const [titleRef, { exactWidth }] = useResizeObserver<HTMLDivElement>();
-  const isBelowBreakpoint = exactWidth && exactWidth < Breakpoints.small;
+  const [internalIsOpened, setInternalIsOpen] = useState(defaultOpen || opened);
+  const isOpen = opened !== undefined ? opened : internalIsOpened;
+  const { smallAndUp } = useBreakpoints();
+  console.log({ internalIsOpened, isOpen, opened, defaultOpen });
```
MichaelParadis, line 42: Remove console log

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -27,35 +24,35 @@ interface DisclosureProps {
...
+  onToggle,
+  opened,
 }: DisclosureProps) {
-  const [isOpen, setOpen] = useState(defaultOpen);
-  const [isMounted, setMount] = useState(false);
-  const [titleRef, { exactWidth }] = useResizeObserver<HTMLDivElement>();
-  const isBelowBreakpoint = exactWidth && exactWidth < Breakpoints.small;
+  const [internalIsOpened, setInternalIsOpen] = useState(defaultOpen || opened);
+  const isOpen = opened !== undefined ? opened : internalIsOpened;
+  const { smallAndUp } = useBreakpoints();
```
MichaelParadis, line 41: changed to use `useBreakpoints`

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -27,35 +24,35 @@ interface DisclosureProps {
...
+  readonly onToggle?: (newOpened: boolean) => void;
+  readonly opened?: boolean;
 }
 
 export function Disclosure({
   children,
   title,
   defaultOpen = false,
+  onToggle,
+  opened,
```
ericchernuka, line 37: Prefer `open` rather than `opened` since opened is past tense and this is describing the current state.

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -67,15 +64,12 @@ export function Disclosure({
...
     const { open: currentToggleState } = event.target as HTMLDetailsElement;
-
-    if (!isMounted || currentToggleState === isOpen) {
-      return;
-    }
-
-    setOpen(!currentToggleState);
+    if (currentToggleState === isOpen) return;
+    setInternalIsOpen(() => !isOpen);
+    onToggle?.(!isOpen);
```
MichaelParadis, line 72: something odd is happening, it looks like even though it is controlled the disclosure still toggles even if `opened` doesn't change. Going to look into this

MichaelParadis: This looks to be because of [this](https://github.com/facebook/react/issues/15486) issue. I have an idea that I will test out

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -27,35 +24,43 @@ interface DisclosureProps {
...
 
-  useEffect(() => {
-    setMount(true);
-  }, []);
-
   return (
-    <details open={isOpen} onToggle={onToggle} className={styles.details}>
-      <summary className={styles.summary}>
+    <details open={isOpen} className={styles.details}>
+      <summary className={styles.summary} onClick={handleToggle}>
```
MichaelParadis, line 55: Use `onClick` here instead of `onToggle` to prevent `details` not working as controlled disclosure. For more information [see](https://github.com/facebook/react/issues/15486#issuecomment-488028431)

packages/components/src/Disclosure/Disclosure.tsx
MichaelParadis: I think this is because the `useResizeObserver` uses the deprecated breakpoints [see](https://atlantis.getjobber.com/?path=/docs/hooks-useresizeobserver--docs#the-breakpoints-object) while `useBreakpoints` uses the "new" ones [see](https://atlantis.getjobber.com/?path=/docs/design-breakpoints--docs). I noticed that when one the new breakpoints it was the same size as before it just needed to be on a different viewport size.

CC: @chris-at-jobber

MichaelParadis: Actually after reading the docs for `useResizeObserver` I should revert this change. `useResizeObserver` cares about the size of the actual component and not the viewport. I will undo the change

MichaelParadis: @jdeichert I undid the change in [3adf1e5](https://github.com/GetJobber/atlantis/pull/2088/commits/3adf1e5680ff948e8b23b24aae3b75d469c724a8)

PR #2089, opened 2024-10-25
feat(components): re-position search to be primary element

chris-at-jobber:
## Motivations

Roughly ~1 year of usage data on interaction patterns with the new DataList has given us some clear indicators that Search is a much more primary interaction than was anticipated.

### Changed

- Updates the order of elements in the `DataListStickyHeader` so that Search is the leading UI element, now followed by the filter Chips if present

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.module.css
+++ b/packages/components/src/DataList/DataList.module.css
@@ -26,10 +26,9 @@
 }
 
 .headerFilters {
-  display: grid;
+  display: flex;
   padding: calc(var(--space-small) + var(--space-smaller)) 0;
   gap: var(--space-small);
-  grid-template-columns: auto fit-content(calc(var(--base-unit) * 14));
 }
 
 .headerFilters:empty {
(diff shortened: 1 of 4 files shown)


packages/components/src/DataList/DataList.module.css
```diff
@@ -26,10 +26,9 @@
 }
 
 .headerFilters {
-  display: grid;
```
chris-at-jobber, line 29: I'm open to grid still being a potential way to handle this layout, but I'm not certain what we'd actually gain from it

Also, looking at this whole selector, I'm wondering if I can't just use `<Flex>` here 🤔

chris-at-jobber: The use of `:empty` just below is a nice touch but it feels like that could also be handled by not rendering the element at all rather than rendering and then checking its' contents 🤷

PR #152, opened 2019-10-17
feat(components): Create dem list component yo

darryltec:
### Added

- List component

### Changed

- Added orange on icon color

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #2090, opened 2024-10-25
fix(components): Fix SectionHeader Overlap

ZakaryH:
## Motivations

noticed this in another PR

image

it's from the Section List web story rendering a List with sections, inside a `Card` which has rounded borders

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

removed a bunch of color styles that I don't believe we use anymore. seems like they existed for Icon colors, but we don't do it like that anymore

there's a single place we reference the `SectionHeader.module.css` and in that file we only use `sectionStyles.sectionHeader` nothing else.

image

### Fixed

removed the background color of the section header, so that it avoids overlapping with the rounded borders of a `Card`

there's no need for the header to have a background color explicitly defined as far as I can tell. here I've given the background a red color and everything seems fine.

image

### Security

- 

## Testing

on the Section List webstory which has a Card wrapping it, there should not be gaps in the corners of the container

the section heading background should be the same color as the rest of the List

the color of the icons in the DisplayList and SectionList should be correctly applied via the `iconColor` specified in the data eg.

```
    {
      id: 1,
      icon: "wallet",
      iconColor: "orange",
      content: "Payment for Invoice #39",
      value: "-$300.00",
      caption: "Sep 25, 2019",
      onClick: () => alert("TODO: Implement onClick"),
    },
```

go ahead and try a few different colors please and thank u

Changes can be

---

Random photo of Atlantis


chris-at-jobber:
Should we strip out L41 and directly call `styles.sectionHeader` in L47 of List while we're here?

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/List/List.tsx#L41-L47

ZakaryH:
>Should we strip out L41 and directly call styles.sectionHeader in L47 of List while we're here?

might as well. really no need for the `classnames` if it's always applying a single class.

packages/components/src/List/SectionHeader.module.css
```diff
@@ -3,36 +3,4 @@
   top: 0;
   padding: var(--space-small) var(--space-base);
   border-bottom: 1px solid var(--color-border--section);
```
chris-at-jobber, line 5:
```suggestion
  border-bottom: var(--border-base) solid var(--color-border--section);
```

while we're here

ZakaryH:
@chris-at-jobber 

>Should we strip out L41 and directly call styles.sectionHeader in L47 of List while we're here?

it's exceedingly likely we will need to bring that back for making the section header custom. if it's alright with you I might just leave it for now.

chris-at-jobber (approved):

chris-at-jobber (commented):
Design approved, also. We confirmed (by going back to the 2019 PR that introduced List) that the "section header color" options have never actually been used in this component, and were left behind in a commit that removed the ability for the consumer to specify header colors.

ZakaryH:
commit that removed the code that would have used the section heading colors for reference

https://github.com/GetJobber/atlantis/pull/152/commits/51e413f8a56f2094581b5c04e0d2af440fdd6481

seems like we simply forgot to remove the colors in the css file

PR #90, opened 2019-08-14
Time for iBrowse

seanhealy:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- `InputTime` now works in Safari.

### Security

- 

## Testing

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #2091, opened 2024-10-26
feat(components): BREAKING CHANGE: remove Civil Date and Time libraries

JeffMarvin:
## Motivations
We should remove this long-since deprecated, never intended for production proposal library. With it, `CivilDate`, `CivilTime` and `CivilDateTime` are no longer valid data types for the components that used to use them. JavaScript Date objects work well enough as a standard for now, and things should improve in the language soon. 

The following web components have their APIs affected:
- Countdown
- FormatDate
- FormatRelativeDateTime
- FormatTime
- InputTime

`InputTime` _required_ the use of `CivilTime` (and didn't have other types available as input values) and is the least compatible with this change. 

This also cleans up an unused polyfill (time-input-polyfill) that was [originally introduced](https://github.com/GetJobber/atlantis/pull/90/files) to make InputTime compatible with Safari.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- This removes the deprecated proposal implementation of the Temporal library, which and removes the ability to use CivilDate, CivilTime and CivilDateTime in related components. JS Date objects should be used instead.

### Fixed

- 

### Security

- 

## Testing

Changes can be

---


packages/components/src/Countdown/Countdown.tsx
```diff
@@ -12,10 +11,10 @@ type GranularityOptions = "dhms" | "hms" | "ms" | "s" | "dhm" | "dh" | "d";
 interface CountdownProps {
   /**
    * The date that is being counted down to
-   * Civil Time of time is to be displayed.
+   * time is to be displayed.
```
Aiden-Brine, line 14: I think the wording on this docstring could be improved

packages/components/src/InputTime/InputTime.tsx
```diff
@@ -52,3 +48,33 @@ export function InputTime({
     }
   }
 }
+
+function dateToTimeString(date?: Date): string {
+  if (date == undefined || !(date instanceof Date)) {
```
Aiden-Brine, line 53: I don't think you need the `date == undefined` check since `undefined instanceof Date` will return `false`

packages/components/src/InputTime/InputTime.tsx
```diff
@@ -52,3 +48,33 @@ export function InputTime({
...
+    return "";
+  }
+
+  // Extract hours and minutes from the Date object
+  const hours = date.getHours().toString().padStart(2, "0");
```
Aiden-Brine, line 58: You could utilize [toTimeString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toTimeString) to do this in one line. If you think it is more readable to explicitly grab the hours and minutes then I am okay with that as well

packages/components/src/InputTime/InputTime.tsx
```diff
@@ -52,3 +48,33 @@ export function InputTime({
...
+
+  // Return the time string in HH:MM format
+  return `${hours}:${minutes}`;
+}
+
+export function timeStringToDate(timeString: string): Date | undefined {
+  try {
+    const [hours, minutes] = timeString.split(":").map(Number);
+
+    if (isNaN(hours) || isNaN(minutes)) {
```
Aiden-Brine, line 69: Should we add some validation that the hours and minutes values are valid (ie minutes less than 61) or are we confident that if it is a number string it is a valid one?

packages/components/src/Countdown/Countdown.tsx
JeffMarvin: Addressed.

packages/components/src/InputTime/InputTime.tsx
JeffMarvin: Yeah, good call. Updated.

packages/components/src/InputTime/InputTime.tsx
JeffMarvin: I think I prefer the reliability of the API over hoping the string format doesn't change and break the parsing.

packages/components/src/InputTime/InputTime.tsx
JeffMarvin: The current implementation relies heavily on the `useTimePredict` hook which already tightly constrains what values can get set.

That said, I added some bound checking here just in case.

Aiden-Brine (approved):
Looks good, feel free to pull me in if you need any help with actually releasing this breaking change

PR #2106, opened 2024-11-05
fix(design): Updated CSS selectors for dark scrollbar colour [JOB-108461]

Beccaneer:
## Motivations

We wanted to move away from using `::-webkit-scrollbar` as a CSS selector as it is considered non-standard (https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar).

Unfortunately the change does not work on Safari, so the previous code remains for that case only. Safari's current behaviour seems to be on initial page load, the scrollbar is light but turns dark on subsequent page loads and this behaviour is unaffected by this change.

### Fixed

- Updated the CSS selector for the dark mode of the scrollbar for supported browsers

## Testing
Turn on dark mode and check out the scrollbar 👀 

Changes can be

## Screenshots
Removed -- don't believe their lies

---

Random photo of Atlantis

### DIFF
--- a/packages/design/src/styles/dark.mode.suffixStyles.css
+++ b/packages/design/src/styles/dark.mode.suffixStyles.css
@@ -1,5 +1,13 @@
-/* autoprefixer: off */
-@supports selector(::-webkit-scrollbar) {
+@supports (scrollbar-color: #000 #000) {
+  /* stylelint-disable-next-line selector-max-type */
+  html {
+    scrollbar-color: var(--color-surface--background--subtle)
+      var(--color-surface--background);
+  }
+}
+
+/* Safari doesn't support scrollbar-color yet, so we're using the non-standard selector for now. */
+@supports not (scrollbar-color: #000 #000) {
   [data-theme="dark"]:root html ::-webkit-scrollbar,
   [data-theme="dark"]:root body ::-webkit-scrollbar {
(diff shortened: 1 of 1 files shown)


packages/design/src/styles/dark.mode.suffixStyles.css
```diff
@@ -1,5 +1,13 @@
-/* autoprefixer: off */
```
chris-at-jobber, line 1: Was this autoprefixer comment doing anything previously that we don't need it anymore?

Beccaneer: Hmmm I think I should add it back since we are keeping the current code for the safari use case 👍  Thanks 😅

packages/design/src/styles/dark.mode.suffixStyles.css
```diff
@@ -1,5 +1,14 @@
 /* autoprefixer: off */
-@supports selector(::-webkit-scrollbar) {
+@supports (scrollbar-color: #000 #000) {
+  html[data-theme="dark"] {
+    scrollbar-color: var(--color-surface--background--subtle)
+      var(--color-surface--background);
+  }
+}
```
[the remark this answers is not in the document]
Beccaneer: Thank you so much for QAing

PR #2108, opened 2024-11-06
POC Update useRefocusOnActivator for SPAR testing

taylorvnoj:
## Motivations

`useRefocusOnActivator` is designed to refocus the element that was active before an overlay element was opened, once the overlay is dismissed. 

Without this change, the issue discussed offline might be that the hook is trying to refocus an element that no longer exists or is not appropriate to refocus when the component unmounts.

This PR is to test if checking that the document contains the activator will fix the issue, as to not attempt to focus an element that has been removed from the DOM.

**Disclaimer**: I haven't dug too much into the issue, so this PR is really just an initial test to see if behaviour changes at all in product. I have also not tested this in Atlantis itself or any other places in product to see if there are any side effects.

---

Random photo of Atlantis

### DIFF
--- a/packages/hooks/src/useRefocusOnActivator/useRefocusOnActivator.ts
+++ b/packages/hooks/src/useRefocusOnActivator/useRefocusOnActivator.ts
@@ -15,8 +15,8 @@ export function useRefocusOnActivator(active: boolean) {
     }
 
     return () => {
-      if (active) {
-        if (activator instanceof HTMLElement) {
+      if (active && activator instanceof HTMLElement) {
+        if (document.contains(activator)) {
           activator.focus();
         }
         activator = undefined;


taylorvnoj:
cc @JeffMarvin @nad182

ericchernuka:
@taylorvnoj don't think that's gonna work since these are valid elements. I think the fix is probably only to mount this hook when the overlay is triggered to open. Right now its registering the refocus element even if you've never interacted with the headers.

taylorvnoj:
> @taylorvnoj don't think that's gonna work since the element is still present in the UI. I think the fix is probably only to mount this hook when the overlay is triggered to open. Right now its registering the refocus element even if you've never interacted with the headers.

Ahhh I see. This was just to do a quick test but what you said makes sense.  I don't have capacity to work on this further right now but we'll coordinate with Cleo on when/who can address this further

PR #2096, opened 2024-11-01
feat(components-native): Add size variation for progress bar

jjwlee94:
## Motivations
- To allow for a smaller size as we do with the progress bar for web.

### Added
- Size prop on `ProgressBar`.

| Small | Base (default) |
| ----- | ----- |
| image | image |

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be
[tested via Pre-release](https://github.com/GetJobber/atlantis/actions/workflows/trigger-qa-build.yml)

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #2110, opened 2024-11-08
feat(components-native): Updated Divider component - added 'vertical' mode and 'testID' prop [JOB-97734]

Beccaneer:
## Motivations

Since web's Divider component has a 'vertical' option, we also wanted to offer this to the mobile Divider component. Also added the 'testID' prop for testing purposes.

### Added

- 'vertical' orientation for mobile Divider
- Added in 'larger' size option for mobile Divider as in line with web
- testID prop for mobile Divider

## Testing

I added in the follow code to the Home page after installing the pre-release package:
```
      <Content>
        <Text>
          Test{"   "}
          <Divider direction={"vertical"} /> {"   "}
          The {"   "}
          <Divider direction={"vertical"} size={"large"} /> {"   "}
          Vertical{"   "}
          <Divider direction={"vertical"} size={"larger"} /> {"   "}
          Divider{"   "}
          <Divider direction={"vertical"} size={"largest"} /> {"   "}T
        </Text>
      </Content>
      <Content>
        <Divider direction={"horizontal"} size={"base"} />
        <Text>Make sure the horizontal divider is still working</Text>
        <Divider direction={"horizontal"} size={"largest"} />
      </Content>
```
I have only tested on iOS so far as I am working on getting my Android environment working 😅 
Screenshot 2024-11-12 at 9 47 58 AM
Screenshot 2024-11-12 at 9 47 51 AM

Changes can be

---

Random photo of Atlantis


packages/components-native/src/Divider/DividerHorizontal.style.ts
```diff
@@ -1,17 +1,22 @@
...
 
-export const styles = StyleSheet.create({
+export const horizontalStyles = StyleSheet.create({
   base: {
     height: tokens["space-minuscule"],
     margin: 0,
-    borderBottomWidth: tokens["border-base"],
+    width: "auto",
     borderBottomColor: tokens["color-border"],
+    borderBottomWidth: tokens["space-minuscule"],
```
Aiden-Brine, line 10: This seems to be fairly common throughout the codebase and I did the same myself but Chris [recently pointed out](https://github.com/GetJobber/atlantis/pull/2096#discussion_r1829960573) that space values should not be used to widths. There are a couple of instances of this in these two styles files

Beccaneer: yeahhh that's fair - updated!

packages/components-native/src/Divider/Divider.tsx
```diff
@@ -1,22 +1,40 @@
...
+  direction = "horizontal",
+  testID = "Divider",
+}: DividerProps): JSX.Element {
+  const directionalStyles =
+    direction === "horizontal" ? horizontalStyles : verticalStyles;
   const style = [
-    styles.base,
-    size === "large" && styles.large,
-    size === "largest" && styles.largest,
+    directionalStyles.base,
```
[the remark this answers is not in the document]
Beccaneer: This would work okay for vertical, but for the horizontal I would need to set the right borders to 0 and apply styling to the bottom border instead. I did originally have a base style and set both borders, then removed the other directional border but thought that was confusing.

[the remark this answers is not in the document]
Aiden-Brine: Ah my bad, I misread the styles. I agree that it is better left the way it is now

packages/components-native/src/Divider/DividerVertical.style.ts
```diff
@@ -0,0 +1,27 @@
...
+  large: {
+    borderRightWidth: tokens["border-thick"],
+    opacity: 0.875,
+  },
+  larger: {
+    borderRightWidth: tokens["border-thicker"],
+    opacity: 0.625,
+  },
+  largest: {
+    borderRightWidth: tokens["space-small"],
```
Aiden-Brine, line 21: I know this is just a copy over from the existing styles but I think it should also be updated to no longer user space as well as the original

chris-at-jobber: Would suggest using `border-thickest` here ([source](https://atlantis.getjobber.com/?path=/docs/design-borders--docs#border-widths))

packages/components-native/src/Divider/DividerHorizontal.style.ts
```diff
@@ -1,17 +1,22 @@
...
   large: {
     borderBottomWidth: tokens["border-thick"],
     opacity: 0.875,
   },
+  larger: {
+    borderBottomWidth: tokens["border-thicker"],
+    opacity: 0.625,
+  },
   largest: {
     borderBottomWidth: tokens["space-small"],
```
chris-at-jobber, line 21: Would also suggest using `border-thickest` ([source](https://atlantis.getjobber.com/?path=/docs/design-borders--docs#border-widths)) here!

Beccaneer: wait.. shouldn't the `largest` border be the thickest?

chris-at-jobber: Yes, that was my suggestion (L21?)

chris-at-jobber:
> https://github.com/GetJobber/atlantis/pull/2110#issuecomment-2471155109

Just noticed that as well! I think we can update those in a separate PR.

packages/components-native/src/Divider/DividerHorizontal.style.ts
Beccaneer: yes you are right sorry!! I misread and I was so confused 😂

Aiden-Brine:
Just looking at the stories on web the dividers span the full height but on mobile they do not. is that purposeful?
Screenshot 2024-11-12 at 5 16 56 PM
Screenshot 2024-11-12 at 5 16 45 PM

Beccaneer:
@Aiden-Brine  Yes, I had manually set a height in the story styling, but I've updated it to be more similar to web:
Screenshot 2024-11-13 at 7 59 22 AM

Beccaneer:
@Aiden-Brine In regards to the comment about the web vertical story styling looking a bit differently from the mobile vertical story styling - I found in storybook for the mobile part I needed to set an explicit height on the container in order for the dividers to show. However, it seems when viewing the vertical divider in the mobile simulators no height needed to be set, so this current implementation should be similar usage between mobile and web.

iOS:
Screenshot 2024-11-14 at 11 52 54 AM


Android:
Screenshot 2024-11-14 at 12 36 20 PM

PR #2111, opened 2024-11-08
fix(components): Remove unnecessary useRefocusOnActivator from DataListHeaderTile

nad182:
## Motivations

The video of the issue can be found here: https://jobber.atlassian.net/browse/JOB-108034

Basically, navigating to Home from Requests tab (where DataList -> DataListHeaderTile is instantiated) caused Requests nav link item to refocus, making it appear as if this is still an active link.

While investigating the issue, I noticed that the `useRefocusOnActivator` hook was called twice, once in `DataListSortingOptions` (to move focus back onto the column header after the sorting option was selected with an assistive technology/keyboard) and in `DataListHeaderTile` (where I believe it should NOT be called). So this PR removes calling this hook in `DataListHeaderTile`.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 
- `visible` prop from `DataListHeaderTile`
- `useRefocusOnActivator` hook call in `DataListHeaderTile`

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/DataList.utils.tsx
+++ b/packages/components/src/DataList/DataList.utils.tsx
@@ -87,7 +87,7 @@ export function generateHeaderElements<T extends DataListObject>(
   const headerElements = Object.keys(headers).reduce(
     (acc, key) => ({
       ...acc,
-      [key]: <DataListHeaderTile headers={headers} headerKey={key} visible />,
+      [key]: <DataListHeaderTile headers={headers} headerKey={key} />,
     }),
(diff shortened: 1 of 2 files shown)


taylorvnoj (approved):
LGTM - I tested in Atlantis and in product, behaviour remains the same. The change makes sense to me.

As long as this is also working for your use case, we're good to go.

PR #2118, opened 2024-11-14
fix(components): Fixing NPM Link within Atlantis

scotttjob:
## Motivations

We've heard varying reports of `npm link` not working with Atlantis.

We want `npm link` to be functional. The change in this PR seems to make it functional.

Try it out and let us know!

We have associated documents internally with more information about this PR if you pick it up for review. Ping me and I'll share them with you. :)

## Changes

Shipped jobber/formatters as ESM.

## Testing

You should be able to use `npm link` in this branch without error and have your experience update on build

Changes can be

---

Random photo of Atlantis


nad182:
Just chiming in, since I did a lot of stuff locally to make it semi functional. There are multiple steps necessary.
1. Updating node version in Atlantis to 20+ OR removing `engine-strict=true` from `.npmrc` in repos where you're trying to link Atlantis.
2. Fixing dependencies between Atlantis packages, since there were issues when I imported an Atlantis component, which used Atlantis hooks, formatters or design imports. These dependencies don't seem to be built appropriately, so their exported methods were undefined. Perhaps, an issue with "path-ing".
3. When I tried it, I had to link multiple Atlantis packages separately. Since if you linked just `components`, for example, changes are the view/page you have open uses a component which has `hooks` or any other package imports.
4. There were some issues due to different package managers (npm vs pnpm), but I don't remember exactly what they were and how I got around them.

scotttjob:
We will not be merging this PR directly anymore. We will break some of the work in this PR down into smaller chunks and merging separately.

PR #2120, opened 2024-11-15
feat(components): Add size prop to SegmentedControl

taylorvnoj:
## Motivations

Some current implementations of similar components within product have varying heights. We can match our `Button` component and offer 2 sizes, `base` & `small`

Screenshot 2024-11-14 at 11 57 25 PM

### Added

- `size` prop that is set on the parent
- added a `Sizes` story to show what `base` and `small` look like

### Changed

- the default `SegmentedControl` is now `40px` in height to match other components like `Button`, where the `base` is also `40px`
- now the `small` variation is `32px` in height

## Testing

Check out the Sizes story.

I'll also share video, to show that a long label (although not recommended) will still give the same spacing, etc between `base` & `small`

https://github.com/user-attachments/assets/e611eee1-15f8-4151-86f5-ea8fbea49d4d

---

Random photo of Atlantis

### DIFF
--- a/docs/components/SegmentedControl/Web.stories.tsx
+++ b/docs/components/SegmentedControl/Web.stories.tsx
@@ -4,6 +4,7 @@ import { SegmentedControl } from "@jobber/components/SegmentedControl";
 import { Button } from "@jobber/components/Button";
 import { Text } from "@jobber/components/Text";
 import { Typography } from "@jobber/components/Typography";
+import { Heading } from "@jobber/components/Heading";
 
 const meta: Meta = {
   title: "Components/Selections/SegmentedControl/Web",
@@ -64,5 +65,29 @@ export const Uncontrolled: Story = () => {
         </SegmentedControl.Option>
       ))}
     </SegmentedControl>
+  );
+};
+
+export const Sizes: Story = () => {
+  return (
+    <div>
+      <Heading level={4}>Small</Heading>
+      <SegmentedControl defaultValue="day" size="small">
+        {options.map(option => (
+          <SegmentedControl.Option key={option.value} value={option.value}>
+            {option.label}
+          </SegmentedControl.Option>
+        ))}
+      </SegmentedControl>
+      <br />
+      <Heading level={4}>Base</Heading>
+      <SegmentedControl defaultValue="day">
+        {options.map(option => (
+          <SegmentedControl.Option key={option.value} value={option.value}>
+            {option.label}
+          </SegmentedControl.Option>
+        ))}
+      </SegmentedControl>
+    </div>
   );
 };
--- a/docs/components/SegmentedControl/Web.stories.tsx
+++ b/docs/components/SegmentedControl/Web.stories.tsx
@@ -86,6 +86,14 @@ export const Sizes: Story = () => {
           </SegmentedControl.Option>
         ))}
       </SegmentedControl>
+      <Heading level={4}>Large</Heading>
+      <SegmentedControl defaultValue="day" size="large">
+        {options.map(option => (
+          <SegmentedControl.Option key={option.value} value={option.value}>
+            {option.label}
+          </SegmentedControl.Option>
+        ))}
+      </SegmentedControl>
     </Content>
   );
 };
--- a/docs/components/SegmentedControl/Web.stories.tsx
+++ b/docs/components/SegmentedControl/Web.stories.tsx
@@ -5,6 +5,7 @@ import { Button } from "@jobber/components/Button";
 import { Text } from "@jobber/components/Text";
 import { Typography } from "@jobber/components/Typography";
 import { Heading } from "@jobber/components/Heading";
+import { Content } from "@jobber/components/Content";
 
 const meta: Meta = {
   title: "Components/Selections/SegmentedControl/Web",
@@ -33,12 +34,11 @@ export const Controlled: Story = () => {
   const [activeOption, setActiveOption] = useState("week");
 
   return (
-    <div>
+    <Content>
       <Text>
         Control the state without interacting with the SegmentedControl
         directly. Navigate through the options and reset state via the button.
       </Text>
-      <br />
       <Typography>Current activeOption: {activeOption}</Typography>
       <SegmentedControl
         selectedValue={activeOption}
@@ -50,9 +50,8 @@ export const Controlled: Story = () => {
           </SegmentedControl.Option>
         ))}
       </SegmentedControl>
-      <br />
       <Button label="Reset to Week" onClick={() => setActiveOption("week")} />
-    </div>
+    </Content>
   );
 };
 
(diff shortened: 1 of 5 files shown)


taylorvnoj:
@chris-at-jobber I made some assumptions with this, so am looking for any ideas/thoughts/feedback:
- are the `base` sized examples ok on the Docs page? Or should I change those back to `small`?
- is aligning with `Button` in terms of heights the right thing to do?

chris-at-jobber:
> is aligning with Button in terms of heights the right thing to do? ✅ 

> are the base sized examples ok on the Docs page? ✅

packages/components/src/SegmentedControl/SegmentedControl.module.css
```diff
@@ -3,13 +3,17 @@
...
-  min-height: 32px;
+  min-height: 40px;
   box-sizing: border-box;
   border: var(--border-base) solid var(--color-border--interactive);
   border-radius: var(--radius-base);
   background-color: var(--color-surface--background);
 }
 
+.small {
+  min-height: 32px;
```
chris-at-jobber, line 14: I don't know if we have an opinion on this, but would it make sense to follow the pattern from Button and use a CSS var in these `min-height:` property declarations, so we're only manipulating a variable instead of setting new declarations?

image
image

chris-at-jobber (commented):
Design approved! Added one thought around the CSS that's not a blocker.

Also not blocking, but I think we should consider adding a **Note** to the Notes section of the docs to the effect of "No `large` size was added to match the `large` Button because we have no known use cases, but it can and should be added if needed"

(Or we could just add a 48px large variant here but nobody needs it that we know of)

docs/components/SegmentedControl/Web.stories.tsx
```diff
@@ -66,3 +67,27 @@ export const Uncontrolled: Story = () => {
     </SegmentedControl>
   );
 };
+
+export const Sizes: Story = () => {
+  return (
+    <div>
```
chris-at-jobber, line 73:
```suggestion
    <Content>
```

docs/components/SegmentedControl/Web.stories.tsx
```diff
@@ -66,3 +67,27 @@ export const Uncontrolled: Story = () => {
...
+      <br />
+      <Heading level={4}>Base</Heading>
+      <SegmentedControl defaultValue="day">
+        {options.map(option => (
+          <SegmentedControl.Option key={option.value} value={option.value}>
+            {option.label}
+          </SegmentedControl.Option>
+        ))}
+      </SegmentedControl>
+    </div>
```
chris-at-jobber, line 91:
```suggestion
    </Content>
```

docs/components/SegmentedControl/Web.stories.tsx
```diff
@@ -66,3 +67,27 @@ export const Uncontrolled: Story = () => {
...
+    <div>
+      <Heading level={4}>Small</Heading>
+      <SegmentedControl defaultValue="day" size="small">
+        {options.map(option => (
+          <SegmentedControl.Option key={option.value} value={option.value}>
+            {option.label}
+          </SegmentedControl.Option>
+        ))}
+      </SegmentedControl>
+      <br />
```
chris-at-jobber, line 82:
```suggestion

```

docs/components/SegmentedControl/Web.stories.tsx
```diff
@@ -4,6 +4,7 @@ import { SegmentedControl } from "@jobber/components/SegmentedControl";
 import { Button } from "@jobber/components/Button";
 import { Text } from "@jobber/components/Text";
 import { Typography } from "@jobber/components/Typography";
+import { Heading } from "@jobber/components/Heading";
```
chris-at-jobber, line 8:
```suggestion
import { Content } from "@jobber/components/Content";
```

chris-at-jobber: nitpick but could use Content for the layout instead of the `div` and `br`

packages/components/src/SegmentedControl/SegmentedControl.module.css
taylorvnoj: I debated using a variable but when I only had `small` and `base` I thought I'd leave it as is. I don't think we have a solid opinion on this as a team.

_But_ I also like the idea of just including `large` while we're here to align with our other components. So `large` was added here 6441414ba6dd73bf8d7d40548b0ca7ef019f0189 and then I cleaned up the css to match `Button` here e40927ed5fea2d320cdd1d51275ebc79739c6520

docs/components/SegmentedControl/Web.stories.tsx
taylorvnoj: 6a9939a5557b44ef3b70c4c9cc1cbbb1f448f867

docs/components/SegmentedControl/Web.stories.tsx
taylorvnoj: 6a9939a5557b44ef3b70c4c9cc1cbbb1f448f867

docs/components/SegmentedControl/Web.stories.tsx
taylorvnoj: 6a9939a5557b44ef3b70c4c9cc1cbbb1f448f867

chris-at-jobber (commented):
Design remains approved! I'd do the full approval because I get what's going on here and it works, but I think an eng should look at the React pieces because I don't have a sense of "the best way" to accomplish this 😄

packages/components/src/SegmentedControl/SegmentedControlBase.tsx
```diff
@@ -6,26 +7,36 @@ import React, {
 } from "react";
 import styles from "./SegmentedControl.module.css";
 
-const SegmentedControlBase = forwardRef<HTMLDivElement, PropsWithChildren>(
-  function SegmentedControlBase({ children }, ref) {
-    const optionCount = Children.count(children);
+export interface SegmentedControlBaseProps extends PropsWithChildren {
```
Aiden-Brine, line 10: I don't _think_ we need to export this interface as they are only used here?

Aiden-Brine:
Not a suggestion for you to add them to this PR but I think we should have a conversation as a team about if we add tests for this type of thing. I am seen tests in our codebase for this type of thing (essentially testing for the existence of a CSS class). Definitely pros and cons to those but would just be good to lock down either yes or no to us writing those types of tests. But again, not a suggestion here

packages/components/src/SegmentedControl/SegmentedControlBase.tsx
Aiden-Brine: This does need to be exported since it is used as part of a `forwardRef` in `SegmentedControlBase`

Aiden-Brine (approved):
Nice improvement!

PR #2128, opened 2024-11-18
feat(components): Updated Tabs label prop to accept a ReactNode [JOB-109275]

Beccaneer:
## Motivations

Users wanted to insert items other than just text into the `label` area of the Tabs component, such as badges etc. This PR expands the `label` prop type to allow text or a `ReactNode`.

### Changed

- `label` prop of Tabs component now accepts `ReactNode` as well as text. 

### Removed

- `id` attribute of `InternalTab` removed. This was set to be equal to `label`, assuming `label` was always a string. Tests ran on JO with this attribute removed passed, so it does not seem as if this attribute was used anywhere. You can see the CI results here: https://github.com/GetJobber/Jobber/pull/53183

## Testing

- Pass a `ReactNode` to the Tabs `label` prop. I would recommend something that makes sense, like `Heading` or one of the examples I gave in the story. You probably *could* put `DataList` in the tab label - but should you? 😅 

Changes can be

## Screenshots of added story and docs (Updated)

Screenshot 2024-11-19 at 12 55 15 PM

Screenshot 2024-11-19 at 12 55 26 PM
---

Random photo of Atlantis

### DIFF
--- a/docs/components/Tabs/Web.stories.tsx
+++ b/docs/components/Tabs/Web.stories.tsx
@@ -91,54 +91,52 @@ const WithTabChangeCallbackTemplate: ComponentStory<typeof Tabs> = args => {
   );
 };
 
-const WithCustomReactNodeTemplate: ComponentStory<typeof Tabs> = args => {
-  const [tab, setTab] = useState(args.defaultTab ?? 0);
-
+const WithCustomReactNodeTemplate: ComponentStory<typeof Tabs> = () => {
   return (
-    <div>
-      <p>Active tab index: {tab}</p>
-      <Tabs onTabChange={setTab}>
-        <Tab
-          label={
-            <Flex template={["shrink", "shrink"]} align="center">
-              <Typography element={"span"} fontWeight={"semiBold"}>
-                Inline Label
-              </Typography>
-              <InlineLabel color="red">{"+99"}</InlineLabel>
-            </Flex>
-          }
-        >
-          Here is an example of using an Inline Label component in the Tab
-          label!
-        </Tab>
-        <Tab
-          label={
-            <Flex template={["shrink", "shrink", "shrink"]} align="center">
-              <Icon name={"happyFace"} />
-              <Typography element={"span"} fontWeight={"semiBold"}>
-                Icons
-              </Typography>
-              <Icon name={"thumbsUp"} />
-            </Flex>
-          }
-        >
-          Here is an example of using some Icon components in the tab label!
-        </Tab>
-        <Tab
-          label={
-            <Flex template={["shrink", "shrink"]} align="center">
-              <Typography element={"span"} fontWeight={"semiBold"}>
-                Status Label
-              </Typography>
-              <StatusIndicator status={"informative"} />
-            </Flex>
-          }
-        >
-          Here is an example of using a Status Indicator component in the Tab
-          label!
-        </Tab>
-      </Tabs>
-    </div>
+    <Tabs>
+      <Tab
+        label={
+          <Flex template={["shrink", "shrink"]} gap="small" align="center">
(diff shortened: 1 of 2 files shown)


docs/components/Tabs/Web.stories.tsx
```diff
@@ -86,9 +91,61 @@ const WithTabChangeCallbackTemplate: ComponentStory<typeof Tabs> = args => {
   );
 };
 
+const WithCustomReactNodeTemplate: ComponentStory<typeof Tabs> = args => {
+  const [tab, setTab] = useState(args.defaultTab ?? 0);
+
+  return (
+    <div>
+      <p>Active tab index: {tab}</p>
```
chris-at-jobber, line 99: Do we need this active tab reference?

docs/components/Tabs/Web.stories.tsx
```diff
@@ -86,9 +91,61 @@ const WithTabChangeCallbackTemplate: ComponentStory<typeof Tabs> = args => {
...
+      <Tabs onTabChange={setTab}>
+        <Tab
+          label={
+            <Flex template={["shrink", "shrink"]} align="center">
```
chris-at-jobber, line 103:
```suggestion
            <Flex template={["shrink", "shrink"]} gap="small" align="center">
```

docs/components/Tabs/Web.stories.tsx
```diff
@@ -86,9 +91,61 @@ const WithTabChangeCallbackTemplate: ComponentStory<typeof Tabs> = args => {
...
+              <InlineLabel color="red">{"+99"}</InlineLabel>
+            </Flex>
+          }
+        >
+          Here is an example of using an Inline Label component in the Tab
+          label!
+        </Tab>
+        <Tab
+          label={
+            <Flex template={["shrink", "shrink", "shrink"]} align="center">
```
chris-at-jobber, line 116:
```suggestion
            <Flex template={["shrink", "shrink", "shrink"]} gap="small" align="center">
```

docs/components/Tabs/Web.stories.tsx
```diff
@@ -86,9 +91,61 @@ const WithTabChangeCallbackTemplate: ComponentStory<typeof Tabs> = args => {
...
+              </Typography>
+              <Icon name={"thumbsUp"} />
+            </Flex>
+          }
+        >
+          Here is an example of using some Icon components in the tab label!
+        </Tab>
+        <Tab
+          label={
+            <Flex template={["shrink", "shrink"]} align="center">
```
chris-at-jobber, line 129:
```suggestion
            <Flex template={["shrink", "shrink"]} gap="small" align="center">
```

chris-at-jobber (commented):
Just a few non-blocking thoughts, also to your question in the channel - I think we could add something like this to `Tab.stories.mdx`, after the "Design & usage guidelines" section:

```
## Content guidelines

Tabs are a metaphor for physical tab folders, so think of the content of each Tab as its' own "sheet of paper". What the content is can be flexible, as long as each Tab has distinct content.

Tab labels contain only text by default, but additional content can be added if needed. Typically this should be some metadata that helps the user understand or anticipate the contents of each Tab, such as an InlineLabel to badge a count of items in each Tab, or an Icon. Do not insert additional interactive elements such as Buttons into a Tab label as it will create a confusing interaction pattern for the user.
```

docs/components/Tabs/Web.stories.tsx
Beccaneer: Nahhh, not explicitly needed

chris-at-jobber (commented):
Design approved! Thanks @Beccaneer 🙏

MichaelParadis (approved):
LGTM!

PR #2129, opened 2024-11-18
docs(docx): Introduce documentation guidelines

chris-at-jobber:
## Motivations

We don't have any internal guidance around what and how we should write docs. There are some unwritten rules that we sometimes apply, so this is an attempt to make those written.

### Added

- Docs on docs

## Testing

Run locally or on Cloudflare

Guides -> Documentation styleguide

---

Random photo of Atlantis


docs/guides/documentation-style.stories.mdx
```diff
@@ -0,0 +1,87 @@
+import { Meta } from "@storybook/addon-docs";
+import showDo from "./assets/Atlantis_Docs_Show_Do.png";
```
chris-at-jobber, line 2: This feels like a weird way to import images but it's how it's done in `/guides/create-basic-component.stories.mdx` and it seems to work.

docs/guides/documentation-style.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+import showDont from "./assets/Atlantis_Docs_Show_Dont.png";
+
+<Meta title="Guides/Documentation styleguide" />
+
+# Documentation styleguide
+
+<br />
+
+## Purpose
```
chris-at-jobber, line 11: Honestly not even sure if we need this purpose section

docs/guides/documentation-style.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+### Document for Jobber
+
+Follow Atlantis content guidelines for written content and component examples:
+
+- [formatting](../?path=/docs/content-formatting--docs)
+- [voice & tone](../?path=/docs/content-voice-tone--docs)
+- [vocabulary](../?path=/docs/content-product-vocabulary--docs)
+
+#### Relevance
```
chris-at-jobber, line 46: This is probably less critical, could drop only if we see this actually causing problems for people.

Why people do it:
- it's fun for the person doing it
- can help the author apply their mental model to a use case ("I know the different types of ships in Stargate, so I can use those types of ships as categories in this sectioned list")

Why I believe realistic data is more useful:
- the mentioned issue with readers not getting the reference (and the implied mental modal)
  - inconsistent consumer experience of different series/movies/books/etc being used as the reference point across the system depending on which dev wrote the example
- puts focus on why we're building the components/who we're building them for (if you have to look for realistic use cases, you gain a better understanding of how this element might be used in real life)

docs/guides/documentation-style.stories.mdx
```diff
@@ -0,0 +1,87 @@
...
+| **Implementation guidance**  | **Controlled vs uncontroled** |
+| -> Props                     | **Responsiveness**            |
+| -> Callbacks                 | **Component composition**     |
+| -> Component composition     | -> Props                      |
+| -> Controlled vs uncontroled |                               |
+
+Clearly distinguish code-specific guidance from conceptual guidance so consumers
+can find what they're looking for without feeling overwhelmed.
+
+The template generated by `npm run generate` should be your starting point.
```
chris-at-jobber, line 87: @scotttjob writing this made me realize/wonder... 

a) Does `npm run generate` still work? 
b) if not, is there an updated "boilerplate" command we should be referencing

docs/guides/documentation-style.stories.mdx
scotttjob: It's how image imports work with modern bundlers. You just import the image from wherever and they handle making sure it's available in the browser!

docs/guides/documentation-style.stories.mdx
scotttjob: Yeah it's not super useful.

docs/guides/documentation-style.stories.mdx
chris-at-jobber: (Update: confirmed it still works)

docs/guides/documentation-style.stories.mdx
scotttjob: Love this. I enjoy making things whimsical, but I've run into this situation at every company as they grow. Universal pop culture references are not universal at all. Not even the one you're thinking of right now in response to my statement.

docs/guides/documentation-style.stories.mdx
scotttjob: It does work! At least it's still there. I didn't use it myself.

scotttjob (approved):
In the spirit of the document, I've read through it all and approved.

I think we can refine the document over time, and encourage members of Atlantis to speak up if something doesn't make sense or if they have questions.

PR #2137, opened 2024-11-22
feat(components): Add testID to Divider and dynamically reference size and direction props

taylorvnoj:
## Motivations

Updating `Divider` in `@jobber/components` as part of our continued effort to increase prop consistency between equivalent web and mobile components.

Now `Divider` props are exactly the same in mobile and web versions

### Added

A new `testID` prop exists on the `div`.  A test for that prop was also added in `Divider.test.tsx`

### Changed

In following a similar pattern to other components in Atlantis, I decided to dynamically reference the size and direction props to ensure the component is tightly coupled to the available size/direction options.  This means a few new files were created - `DividerSizes.module.css` and `DividerDirections.module.css`. This also DRYs up a bit of the logic in `Divider.tsx`

## Testing

Testing the pre-release in product is linked below.
UI/UX wise, everything remains the same as it was before. You can now add a `testID` .

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Divider/__snapshots__/Divider.test.tsx.snap
+++ b/packages/components/src/Divider/__snapshots__/Divider.test.tsx.snap
@@ -3,7 +3,7 @@
 exports[`Divider should render 1`] = `
 <div>
   <div
-    class="divider horizontal"
+    class="divider base horizontal"
     role="none presentation"
   />
 </div>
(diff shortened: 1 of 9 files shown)


packages/components/src/Divider/__snapshots__/Divider.test.tsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`Divider should render 1`] = `
 <div>
   <div
-    class="divider horizontal"
+    class="divider base horizontal"
```
taylorvnoj, line 6: The word "base" was added to the snapshot test because the css was refactored to include a base size.

PR #2145, opened 2024-11-26
docs(components): Adding Heading to Docs

taylorvnoj:
## Motivations

Adding `Heading` to the new docs site

### Changed

Added the `Heading` component to the new docs site.

**NOTE** HeadingLevel is a type and doesn't come through on the props table like we'd expect - I see the same thing is happening with `IconNames` in `AnimatedSwitcher` and `Button` so I assume we're ok to move ahead for now. I've noted in the Figma that it would be nice to get more functionality in this area in the future

Screenshot 2024-11-26 at 6 02 10 PM

## Testing

You should see a `Heading` option after clicking on Components.

---

Random photo of Atlantis

### DIFF
--- a/packages/site/generateDocs.mjs
+++ b/packages/site/generateDocs.mjs
@@ -115,6 +115,7 @@ const components = [
   "Chip",
   "Countdown",
   "Disclosure",
+  "Heading",
   "StatusLabel",
   "Switch",
 ];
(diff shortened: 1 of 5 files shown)


scotttjob (approved):
We're going to hide that column for now (or make it read-only and just pass through what's in the json file). And when we get that functionality in a better state we'll turn it back on

PR #2166, opened 2024-11-28
docs(components): Adding Combobox to New Docs Site [JOB-109864]

Beccaneer:
### Changed

Added the Combobox component to the new docs site.

## Testing

You should see a 'Combobox' option after clicking on Components.

Changes can be

---

Random photo of Atlantis


packages/site/src/types/combobox.ts
```diff
@@ -0,0 +1,333 @@
+import { Dispatch, ReactElement, SetStateAction } from "react";
```
scotttjob, line 1: This is a smell. Can we import the types from combobox another way than copying them in like this?

scotttjob: If we need these types from Combobox, and they're not exported naturally already, I would rather export them from Combobox properly.

scotttjob (changes requested):
Let's chat about the type imports, I know you're new to typescript so we'll figure that out a different way.

Beccaneer:
This has been done in a different pr :)

PR #2170, opened 2024-11-29
fix(components): Avoid rendering DataList title container when empty 

### LINKED ISSUE #2167: Avoid rendering title and total count elements in DataList when not provided
**Description**

Hey Atlantis! Over on Quotemeal, we've noticed that when the `DataList` component can render with an margin on an empty element, causing an unexpected space on above the rows of that component. See the screenshot below:

image

This gap is due to the `totalCount` element (`DataListTotalCount` component) is rendering with container styles that have a `margin-bottom` set. 

Either, the container can be rendered with `margin: 0`, or we can avoid rendering the component all together. To me, the latter feels like the more natural choice to not render the component at all to avoid empty markup on the page.

kingscott:
## Motivations

Fixes #2167 

Quotemeal was seeing an issue that when we were rendering the title and total count elements on a DataList that there was an element with unnecessary margin being rendered. Therefore, when `title` or `totalCount` aren't provided that container doesn't render, and there are no extra expected elements. 

### Changed

- Changed how and when the `DataList` component renders is title container

### Fixed

- Fixed the issue where empty elements were rendering with margin, causing unexpected spacing

## Testing

Render a `DataList` component without `title` and `totalCount`.

Changes can be

---

Random photo of Atlantis


packages/components/src/DataList/DataList.tsx
```diff
@@ -144,13 +144,20 @@ function InternalDataList({
   const showEmptyState = !initialLoading && data.length === 0;
 
   const shouldRenderLoadMoreTrigger = !initialLoading && !showEmptyState;
+  const shouldRenderTitleContainer =
+    initialLoading || title !== "" || totalCount !== null;
```
ZakaryH, line 148: a few thoughts:

**initialLoading**
I think I understand the rationale behind `initialLoading` being one of the conditions. if we're loading, we might not currently have our data & the corresponding `totalCount` but once the query resolves it could be a proper number. that flow makes sense to me.

the scenario that might be awkward is when we aren't providing the `totalCount` prop at all and have no intention of using it AKA it is `undefined`. if we also don't have a title, it'll render this section while loading, and then disappear once loading is done because we'll fail all the checks.

if we wanted to be as correct as possible, we'd wanna combine `initialLoading` with verifying at least one of the other elements is either there, or expected to be there. that would avoid adding an element we don't need.

**title**
since the type of title is either a `string` or `undefined`, `title !== ""` will result in the section rendering if you completely omit the `title` prop or directly set it to `undefined` so we'd definitely need to check for `undefined`.

I actually don't think we need to check or protect against an empty string. I'd rather leave that up to the consumer to handle in a more explicit way. if you don't want a title, we already provide a way to opt out so simply don't provide one. short circuiting on an empty string would effectively be a second way to achieve that same functionality, but in a less explicit way imo. this will tie into the next point around what `undefined` is meant to represent.

**totalCount**
this one has a surprising amount of hidden complexity 😅 

the type of totalCount is `number | undefined | null`, where omitting the prop is `undefined` so only checking against `null` won't quite get us what we want.

looking into it deeper, `null` is something we **do** want to let through because it represents a different flow & goes down a separate code path.

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/DataList/components/DataListTotalCount/DataListTotalCount.tsx#L20-L22

inside the total count component we are treating `undefined` as the case where you have no total count at all. we show nothing. that will be one of the conditions we need to have out here.

conversely `null` is paired with `loading` to show the Glimmer on initial load only. I'm not sure I agree with that behavior, but I found a comment stating that it was intentional so I can follow up with that separately.

all that is to say, we should specifically checking that `totalCount !== undefined` out here. I'd imagine we can also refactor & remove the `undefined` check inside `DataListTotalCount` since there wouldn't be a way for us to hit that anymore.

ZakaryH:
it's not ideal that no tests failed with this change. the current check of `totalCount !== null` is preventing us from reaching that loading state flow, but since we are only testing for that inside `DataListTotalCount` the tests are passing since it's not aware that those same conditions are preventing it from being rendered by its parent.

we should definitely have something asserting the expected behavior one level up at the `DataList.test.tsx` level. either that or we refactor this title section into its own component, and then unit test that.

ZakaryH:
previously the logic here had the blind spot of rendering the empty div, so I totally get why we need to move the conditionals up a level but something I'd like to do if it's not too much hassle is have one source of truth on this.

since we must do some checking on the parent, I'm thinking about how we might do all of the checks at that level rather than having it split half there and half in the component.

I have some rough ideas but I don't know that they'll work. I'll play around with a couple and see what I can come up with.

kingscott:
Hey @ZakaryH, did a small cleanup here. I moved the check for `totalCount` against `undefined` outside of the component, as that's the only state where we'd prefer to not render anything. (If `totalCount` is `null` somehow after loading, we also render nothing, but that's more predictable).

packages/components/src/DataList/DataList.test.tsx
```diff
@@ -92,6 +92,16 @@ describe("DataList", () => {
...
     });
+
+    it("should not render the total count if the totalCount is not declared", () => {
+      render(
+        <DataList data={mockData} headers={mockHeader} title={mockTitle}>
+          <></>
+        </DataList>,
+      );
+      expect(screen.getByText(mockTitle)).toBeInTheDocument();
+      expect(screen.queryByTestId(DATALIST_TOTALCOUNT_TEST_ID)).toBeNull();
```
taylorvnoj, line 103: nit/not blocking; using `.not.toBeInTheDocument()` is, I think, slightly better to use here than `toBeNull()`. From a readability standpoint as well as semantically.  From what I understand, `toBeNull` is testing the value returned by `queryByTestId` whereas `.not.toBeInTheDocument` is checking for the actual absence of an element.

taylorvnoj (commented):
This does look good, but I'd love to pair with you @kingscott and double check with you that this is solving your problem in your specific instance. I'm having a hard time testing because I can't seem to replicate what you referenced in prod and testing locally is difficult because the API we're gathering data from for the storybook examples seems to be broken.
I'll send you a message and we'll run through together if that's ok.

packages/components/src/DataList/DataList.test.tsx
kingscott: Happy to make that change! I'll follow up with this change after we chat later today. Thanks 🚀

taylorvnoj (approved):
- If title exists, titleContainer is present ✔️ 
- if totalCount exists, titleContainer is present ✔️ 
- if totalCount is null, titleContainer is present (which is intentional) ✔️ 
- if title & totalCount are undefined, titleContainer is not present - no div, no margin-bottom ✔️ 

I've tested in Storybook examples and in multiple places in product 👍🏻 
And we both saw that it fixes your `DataList` instance also 🎉 

Thanks for your first contribution @kingscott - and on the `DataList`! I'll merge this now and then will include the update in the next product version bump.

kingscott:
Thank you both so much for your help and guidance! @taylorvnoj, you rock! cc @ZakaryH you rock also!

PR #2179, opened 2024-11-29
feat(components): Live Doc Editor

scotttjob:
## Motivations

We needed a way to edit code live in the new docs site.

We also wanted a better way to store the default code for each component.

This PR takes care of both.

What I will admit, is I did a merge near the end so not all of the components are in the new format.

So I think we can merge this, then I'll do another pass to reconvert the components that have been updated since I started this work.

## Testing

You should now be able to edit (MOST) of the components (Note, that components merged after I started this work will flat out not load, that's okay. I just need to circle back and fix the sample code)

(used to be text about a bug here, but the bug has been fixed)

Changes can be

---

Random photo of Atlantis


chris-at-jobber:
Is there any special setup needed to run this locally?

I ran `npm ci` at the atlantis root directory, then `npm run dev` from /site

The site builds, but on every component I've tried I get this error when I hit the `Implementation` tab:
image

packages/site/src/componentList.ts
```diff
@@ -318,8 +318,8 @@ export const componentList = [
   },
   {
     title: "ReccuringSelect",
```
scotttjob, line 320: Typo I noticed from the Bones PR

packages/site/src/components/AtlantisPreviewEditorProvider.tsx
```diff
@@ -0,0 +1,330 @@
+import { transform } from "@babel/standalone";
```
scotttjob, line 1: This file is a bit of a mess. I tend to circle back and tidy it up a bit. It was in the books to do today, but I think it's more important to get this over the fence to get the new format unlocked versus tidying up some messy code.

packages/site/src/components/PropsList.tsx
```diff
@@ -60,14 +60,14 @@ export const PropsList = ({
...
                     <div style={{ display: "flex", alignItems: "flex" }}>
                       {item.key}
                     </div>
                   </Grid.Cell>
-                  <Grid.Cell size={{ xs: 3 }}>{item.required}</Grid.Cell>
-                  <Grid.Cell size={{ xs: 3 }}>{item.description}</Grid.Cell>
-                  <Grid.Cell size={{ xs: 3 }}>{item.component}</Grid.Cell>
+                  <Grid.Cell size={{ xs: 4 }}>{item.required}</Grid.Cell>
+                  <Grid.Cell size={{ xs: 4 }}>{item.description}</Grid.Cell>
+                  {/*<Grid.Cell size={{ xs: 3 }}>{item.component}</Grid.Cell> */}
```
scotttjob, line 70: Removing the editable prop field from the DataList Props Table.

packages/site/src/content/AnimatedPresence/index.tsx
```diff
@@ -1,26 +1,20 @@
...
 import { getStorybookUrl } from "../../layout/getStorybookUrl";
 
-export const AnimatedPresence = (props: PropsWithChildren) => {
-  const [switched, setSwitched] = useState(false);
-
+export default {
+  content: () => <AnimatedPresenceContent />,
+  props: Props,
+  component: {
+    element: `const [switched, setSwitched] = useState(false);
```
scotttjob, line 10: Ultimately the format is now Strings. Pass your code through as a string. It will be automatically wrapped in a function. If you don't provide a return statement, I'll insert one.

This is to keep the code examples as clean as possible. I'll document how it all works properly before we go live.

packages/site/src/content/AnimatedPresence/index.tsx
```diff
@@ -1,26 +1,20 @@
...
   return (
     <Content>
       <Button
         label={switched ? "Switched" : "Initial"}
         onClick={() => setSwitched(!switched)}
       />
-      <AnimatedPresenceRoot {...props}>
```
scotttjob, line 23: We no longer have to do this Root pattern. We can just use the components as-is.

Beccaneer (approved):

PR #2192, opened 2024-12-03
Add page from Builder

Aiden-Brine:
This draft PR includes a page generated using builder.io based on [this](https://www.figma.com/design/4E8TKajLSXxA1QNtpQaWMA/RiskOps---Document-Request-%2B-Uploads?node-id=143-3488&node-type=frame&m=dev) design. 

## Custom Prompts Used
- If a frame in Figma has the styling display: flex with a gap value, when generating code use a div to represent the frame and provide spacing between the objects in that div equivalent to the gap value.
- Only provide values to props on components from @jobber/components if they are valid values for those props
- Only provide props to components from @jobber/components if they are valid props for those components. Do not provide the prop classNames or onChange to any components from @jobber/components.The page component from @jobber/components renders a header so the addition of a header with the same value as the title passed to Page is not necessary.
- Do not add any styles, HTML divs or colors that were not specified in Figma
- Do not use position absolute in the CSS
- Avoid deeply nesting HTML divs and instead break things up into smaller pieces
- Please wrap all buttons in divs with a display of block
- Avoid setting a fixed width for items in CSS where at all possible
- Make sure the code is responsive.

## Prompts from builder.io that are editable

Follow these fundamental principles when generating code:

Code Quality:
• Write clean, maintainable, and concise code
• Follow DRY (Don't Repeat Yourself) principles
• Ensure proper error handling

Documentation:
• Add meaningful comments for complex logic
• Document any assumptions or limitations

Naming Conventions:
• Use descriptive and consistent naming
• Follow language-specific conventions
• Avoid abbreviations unless widely recognized

Implement WCAG-compliant, accessible code following these essential requirements:

Semantic Structure:
• Use semantic HTML elements (nav, main, article, section)
• Create logical heading hierarchy (h1-h6)
• Implement proper landmark regions

Interactive Elements:
• Ensure keyboard navigation support
• Maintain visible focus indicators
• Implement proper focus management
• Support touch and mouse interactions

ARIA Implementation:
• Add appropriate ARIA landmarks
• Include descriptive ARIA labels
• Use ARIA states and properties correctly
• Implement ARIA live regions for dynamic content

Media and Content:
• Provide descriptive alt text for images
• Ensure proper color contrast ratios
• Support text resizing and zoom
• Include captions and transcripts for media

Testing Requirements:
• Verify screen reader compatibility
• Test with keyboard-only navigation
• Ensure functionality at 200% zoom

## Testing

You can take a look at the page [here](http://localhost:6005/?path=/docs/components-builderio-documentupload--docs)

## Styles mappings

Currently waiting to hear back from their team about doing the styling mappings as there is a gap in their docs about it

---

Random photo of Atlantis


packages/components/src/DocumentUpload/DocumentUploadForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <InputFile
+        allowedTypes="all"
+        variation="dropzone"
+        className={styles.fileInput}
+        allowMultiple
+        onChange={setFiles}
+      />
+
+      <InputText
+        align="left"
```
Aiden-Brine, line 36: This and the value for size are invalid. I have it set up so it should be mapping to correct values, I need to talk to their team about that.

packages/components/src/DocumentUpload/DocumentUploadForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <form onSubmit={handleSubmit} className={styles.uploadForm}>
+      <div className={styles.formHeader}>
+        <h1 className={styles.formTitle}>Documents requested</h1>
+        <p className={styles.formInstructions}>{instructions}</p>
+      </div>
+
+      <InputFile
+        allowedTypes="all"
+        variation="dropzone"
+        className={styles.fileInput}
```
Aiden-Brine, line 30: It is still adding `className` and `onChange` even though I am explicitly telling it not to

Aiden-Brine:
The responsiveness certainly leaves something to be desired

Screenshot 2024-12-02 at 5 27 06 PM

packages/components/src/DocumentUpload/SecurityBlurb.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Icon } from "@jobber/components";
+import styles from "./DocumentUpload.module.css";
+import { SecurityBlurbProps } from "./types";
+
+export const SecurityBlurb: React.FC<SecurityBlurbProps> = ({
+  privacyPolicyUrl,
+}) => (
+  <div className={styles.securityBlurb}>
+    <Icon name="lock" size="small" className={styles.securityIcon} />
+    <div className={styles.securityText}>
```
Aiden-Brine, line 11: huh, styles.securityText just straight up doesn't exist 😬

packages/components/src/DocumentUpload/DocumentUploadForm.tsx
Aiden-Brine: Okay I figured out what was going on with this one and it is now correctly mapping

packages/components/src/DocumentUpload/DocumentUploadForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+        allowMultiple
+        onChange={setFiles}
```
Aiden-Brine, line 32: Input file requires a `getUploadParams` prop. There is no equivalent in Figma but I added it to the mapping file so I would expect it to be added here. I am waiting to hear back from their team on that

packages/components/src/DocumentUpload/DocumentUpload.module.css
ZakaryH, line 1: there are a handful of styles that I'm pretty skeptical of. they either do nothing, or are a questionable way of achieving the desired result.

one example is the `.container` class having equal padding AND justify-content/align-items `center`. those are both doing the same thing, trying to center the content except only the padding is doing what it is supposed to do since the container has no provided dimensions to center its content within.

at first glance this seems reasonable making it easy to miss and leave behind some code that is not needed. also, we can't ask the "author" what they were going for.

ZakaryH: it's also oddly inconsistent with what it solves a problem with

in the `.container` it used `display: flex` to fill the width, which is valid. 

then in a handful of other classes it used `width:100%` to do the same thing, which is once again superfluous since it's applying that to a `div` which has `display: block`. the rule isn't providing the wrong behavior it's just that you don't need to even apply that rule for it to do the right thing.

packages/components/src/DocumentUpload/SecurityBlurb.tsx
```diff
@@ -0,0 +1,18 @@
...
+      Documents and files you upload are stored securely as per our{" "}
```
ZakaryH, line 12: a couple things here

- first it's putting text directly in a div rather than some kind of text element which I don't love
- second, the `{" "}` is an interesting way to add a space. I get that a formatter might remove a trailing whitespace but still. I don't love it.

packages/components/src/DocumentUpload/DocumentUpload.module.css
Aiden-Brine: Great callout! I also have some concerns about unnecessary CSS. It currently has no awareness of our components so I have seen it wrapping components with a `size` prop in a `div` specifying what the size should be. Potentially fixable by telling the AI "if you encounter a component with a size prop don't use CSS to control its size. This is brittle for two reasons:
- If a component comes along where a prop other than "size" controls its size
- More likely a component has a default size value so there is no size prop provided. `InputText` for example takes a size prop of `"small" | "large"` and if none are provided it uses the value of "base" (arguably bad design not allowing the default value as a prop but that is the current situation). If no size prop is provided it will assume it needs to wrap it in a div and add its own CSS to size it.


This isn't an issue right now because it isn't noticeable but I see this becoming a problem if we want to change the size in the future. The developer will change the size prop and be confused about why it isn't working. Turns out there is also CSS now dictating the wrong size built in

PR #2204, opened 2024-12-04
feat(components): Add controlled option to Tabs

taylorvnoj:
## Motivations

Adding a controlled option to the `Tabs` component

In this PR, the `Tabs` component checks if the `activeTab` prop is provided. If it is, the component uses this prop to determine which tab is active. This means the parent component can control the active tab by updating the `activeTab` prop. If the `activeTab` prop is not provided, the `Tabs` component falls back to managing its own state internally using the `internalActiveTab` state variable.

### Added

- new optional prop `activeTab` to allow `Tabs` to be controlled
- new test for controlled version
- storybook example for controlled version

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Tabs/Web.stories.tsx
+++ b/docs/components/Tabs/Web.stories.tsx
@@ -140,10 +140,38 @@ const WithCustomReactNodeTemplate: ComponentStory<typeof Tabs> = () => {
   );
 };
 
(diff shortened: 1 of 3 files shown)


scotttjob (approved):
LGTM!

PR #2210, opened 2024-12-05
docs: tweaks to dark mode experience

chris-at-jobber:
## Motivations

Felt like I could make the collection thumbnails less glaring until we have a better solution (ie mode-responsive images)

### Changed

- gently darken thumbnail backgrounds and used a darker moon icon

## Testing

Throw the new docs site in dark mode

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/components/ContentCard.tsx
+++ b/packages/site/src/components/ContentCard.tsx
@@ -39,7 +39,7 @@ export const ContentCard = ({
             style={
               imageURL && theme === "dark"
                 ? {
-                    background: "var(--color-surface)",
+                    background: "var(--color-base-blue--200)",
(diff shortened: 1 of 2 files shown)


packages/site/src/components/ContentCard.tsx
```diff
@@ -39,7 +39,7 @@ export const ContentCard = ({
             style={
               imageURL && theme === "dark"
                 ? {
-                    background: "var(--color-surface)",
+                    background: "var(--color-base-blue--200)",
```
chris-at-jobber, line 42: Arbitrary but this feels _just dark enough that it's not so glaring_ but without washing out the thumbnail contents

image

PR #2211, opened 2024-12-05
feat(components): Add arrow key navigation to Tabs

taylorvnoj:
## Motivations

In the `Tabs` component, keyboard navigation can be improved when navigating between tabs and to children within individual tabs.  Currently the only way to navigate between tabs is with the tab key and hitting enter to view the focused tab's children. 

Here is a demo of what we should now expect for navigation between tabs:

https://github.com/user-attachments/assets/cc5d24b4-47a9-48ac-936c-7d62c98b6325

### Fixed

In the Tabs component users can:
- tab key will enter or exit the Tabs component
- use arrow keys to navigate between tabs
- while focused on a Tab, if the children of that tab have focusable items, hitting tab will move focus to the child element
- while focused on the content of children in a Tab, shift-tabbing “upwards” out of the content will return focus to the active Tab selector

If 2 Tabs sets are in one page: 
- arrow key navigating will keep you within the current Tabs set
- you can move from the first tab set to the second but using the tab key

I also added a hook within the `Tabs` component called `useArrowKeyNavigation`.  There is an inline comment below, explaining why I chose to keep the hook internal to `Tab` for now.

## Testing

I did not push the example, but if you'd like to test functionality with 2 `Tabs` on one page, you can use this example:

```
const TwoTabsExample: ComponentStory<typeof Tabs> = () => {
  return (
    <div>
      <h3>First Tab Set</h3>
      <Tabs>
        <Tab label="Tab 1">
          <div>
            Content for Tab 1 in the first set.
            <Button label="Cancel" type="primary" variation="subtle" />
          </div>
        </Tab>
        <Tab label="Tab 2">Content for Tab 2 in the first set.</Tab>
      </Tabs>
      <h3>Second Tab Set</h3>
      <Tabs>
        <Tab label="Tab A">Content for Tab A in the second set.</Tab>
        <Tab label="Tab B">
          <div>
            Content for Tab B in the second set.
            <a
              href="https://www.example.com"
              target="_blank"
              rel="noopener noreferrer"
            >
              Visit Example
            </a>
          </div>
        </Tab>
      </Tabs>
    </div>
  );
};
```

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Tabs/Tabs.test.tsx
+++ b/packages/components/src/Tabs/Tabs.test.tsx
@@ -1,5 +1,6 @@
 import React from "react";
-import { fireEvent, render } from "@testing-library/react";
+import { render } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 import { InlineLabel } from "@jobber/components/InlineLabel";
 import { Tab, Tabs } from ".";
 
@@ -33,6 +34,7 @@ const originalScrollWidth = Object.getOwnPropertyDescriptor(
   "scrollWidth",
 );
 
+// eslint-disable-next-line max-statements
 describe("Tabs", () => {
   it("renders Tabs", () => {
     const { container } = render(omelet);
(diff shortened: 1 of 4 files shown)


packages/components/src/Tabs/hooks/useArrowKeyNavigation.ts
```diff
@@ -0,0 +1,47 @@
+import { useCallback } from "react";
+
+interface UseArrowKeyNavigationProps {
+  elementsRef: React.RefObject<(HTMLElement | null)[]>;
+  onActivate: (index: number) => void;
+}
+
+export const useArrowKeyNavigation = ({
```
taylorvnoj, line 8: I'm choosing to create and keep this hook within `Tabs` for a few reasons:

1. We talked about sharing arrow key navigation between `SegmentedControl` and `Tabs` since they are very similar however, `SegmentedControl` does not have any specific arrow key logic.  That is because arrow key navigation is implicitly set through the `SegmentedControl` radio button structure. The options are a type of radio and part of a group (name attribute) which inherently supports keyboard navigation. So since the component uses a `role="radiogroup"` and `name` attribute, the browser handles this behaviour automatically. There's nothing to share between these two components.

2. There is one other area (`packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts`) where this type of logic is being used but to be honest, creating a more universal version of this hook for one other component (that I'd prefer not to touch right now 😅 ) is not something I want to increase scope on.  I'm happy to create a ticket to create a more flexible hook in the future if we think that is something our library should have. I have ideas on what that hoook can look like so can document it in the ticket

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -72,13 +86,20 @@ export function Tabs({
...
+          onKeyDown={handleKeyDown}
+        >
           {React.Children.map(children, (tab, index) => (
             <InternalTab
               label={tab.props.label}
               selected={activeTab === index}
               activateTab={activateTab(index)}
               onClick={tab.props.onClick}
+              ref={el => (tabRefs.current[index] = el)}
+              tabIndex={activeTab === index ? 0 : -1}
```
taylorvnoj, line 102: This ensures that only the active tab is focusable using the tab key.  So if you have two tabs, and the first tab has a focusable child, when focus is on the tab and you hit tab key, the focus will go to the child and not the next tab.
Instead, arrow keys will help you continue to navigate to the next tab.

packages/components/src/Tabs/hooks/useArrowKeyNavigation.ts
chris-at-jobber: Love this breakdown

chris-at-jobber (commented):
Design approved! This is LONG overdue, thanks for making it happen @taylorvnoj  🙏

[Recorded my review](https://getjobber.zoom.us/clips/share/BGNhMDEWRHJJS2J5X09RMkNkLURhZkpCWVlQdwE) if it's helpful.

taylorvnoj:
Thanks for the review recording @chris-at-jobber!  I chose not to ship the 2 `Tabs` example (I don't think it's needed other than for testing). The code for that example is in the PR description if you wanted to double check but to your point, that is what you're seeing in my demo video.

Thanks again Chris!

scotttjob (approved):

packages/components/src/Tabs/hooks/useArrowKeyNavigation.ts
taylorvnoj: Yeah that sounds good - I can play around with that today and see what you and Chris think!

packages/components/src/Tabs/Tabs.test.tsx
```diff
@@ -33,133 +34,220 @@ const originalScrollWidth = Object.getOwnPropertyDescriptor(
...
-    expect(queryByText("🧀")).not.toBeInTheDocument();
+      tab1.focus();
+      await userEvent.keyboard("{Tab}");
+      expect(focusableChild).toHaveFocus();
+    });
 
-    fireEvent.click(getByText("Set Tab 1"));
-    expect(queryByText("🍳")).not.toBeInTheDocument();
-    expect(queryByText("🧀")).toBeInTheDocument();
+    it("while focused on a child element, hitting Shift+Tab moves focus back to the tab", async () => {
```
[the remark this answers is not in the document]
taylorvnoj: same! love when things are for free

packages/components/src/Tabs/Tabs.tsx
```diff
@@ -110,37 +131,42 @@ interface InternalTabProps {
...
+      <li role="presentation">
+        <button
+          type="button"
+          role="tab"
+          className={className}
+          onClick={event => {
+            activateTab();
+
+            if (onClick) {
+              onClick(event);
```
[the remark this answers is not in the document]
taylorvnoj: I knew I was doing something weird and forgot to return to it! fixed here a046635983ae2da13d96d7ce4845fc28a0ed6e85

packages/components/src/Tabs/hooks/useArrowKeyNavigation.ts
```diff
@@ -0,0 +1,47 @@
...
+  elementsRef,
+  onActivate,
+}: UseArrowKeyNavigationProps) => {
+  const handleKeyDown = useCallback(
```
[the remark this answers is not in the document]
taylorvnoj: Good call out to see what other DS are doing! I do have a working generic hook but I could see it growing and yes, it's a valid concern that it might just be more complicated.  I'll dive into this one as side desk work, do some research and we can regroup on it

chris-at-jobber:
@scotttjob we have a few components where we've added a "Notes" section (see Datepicker as an example) - sometimes that gets used for design-y things, sometimes eng-related decisions - we could apply that pattern to the Implementation/"Code" tab?

PR #2213, opened 2024-12-05
feat(components): add input text rebuild EARLY Iterations

MichaelParadis:
From: https://github.com/GetJobber/atlantis/pull/2105

## Motivations

We have an upcoming need to decouple from `react-hook-form`. We still want to leverage the library, but we no longer want our form components to be directly coupled, instead we want them to support being used in that environment but not use the library directly.

This PR is the next step by adding an rebuild `InputText` that removes the react-hook-form functionality

You can verify this new component is working through the third part of this refactor, the potential pattern for an Atlantis Shim. Currently if you use `InputText` directly, you'll get the OG version of the component. If you pass a prop called 'version' set to `2` you'll get the new component. You can see how we handle this swap in `index.tsx` under InputText.

## Testing

To test, open a storybook story (InputText/Web.stories) and use the Version comparison story to see the behaviour of version 1 components side by side with the version 2 component

Efforts were taken to gets tests as good as possible for the new version of the component. If you forcibly swap to the new version (in the shim, force flow to the new component versus the old), you should only have failing tests in Form (which makes sense, react-hook-form is not part of the flow), AutoComplete (some tests rely on react-hook-form functionality) and InputText (some tests rely on react-hook-form functionality). The rest of the tests in the system (which only use InputText downstream with basic examples) pass fine in this scenario.

## Changes

- Added InputTextSPAR (naming TBD) as a potential 'new' implementation of InputText that bypasses react-hook-form.
- Added a potential pattern for an Atlantis Shim to allow existing users of a component to get the legacy version of the component, and consumers wishing to use the new pattern can apply an additional prop `version:2` to the prop list.

## Decisions

@JeffMarvin and I decided to not dry our hooks for form inputs in `useInputTextActions.ts` and `useInputTextFormField` initially so that we could determine bare minimum functionality needed for our new input components and refactor the hooks as needed in the future.

Due to a limitation in typescript some of the props related to aria are visible on the intellisense V1 version  that are added to the V2 component and don't cause errors when the version of the input is V1. The reason for this is that typescript doesn't check dash separated attributes like `aria-...` [see](https://github.com/microsoft/TypeScript/issues/32447)
### Added

- Added InputText rebuilt that is shimmed and the version can be toggled with the version prop

## Testing

To test, open a storybook story (InputText/Web.stories) and use the Version comparison story to see the behaviour of version 1 components side by side with the version 2 component

Efforts were taken to gets tests as good as possible for the new version of the component. 

Changes can be

---

Random photo of Atlantis


packages/components/src/InputText/index.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { ForwardedRef } from "react";
+import {
+  InputText as InputTextLegacy,
+  InputTextPropOptions,
+  InputTextRef,
+} from "./InputText";
+import { InputTextSPAR } from "./InputText.rebuilt";
+import { InputTextRebuiltProps } from "./InputText.types";
+
+type CombinedProps = InputTextPropOptions | InputTextRebuiltProps;
```
MichaelParadis, line 10: Look into conditional types because typescript suggestions are really weird at the moment
https://www.typescriptlang.org/docs/handbook/2/conditional-types.html

JeffMarvin (approved):
Great stuff here! 
As discussed, since v2 isn't in use, we will merge this and keep iterating.

PR #2214, opened 2024-12-05
docs: Deploy new docs beside storybook

jdeichert:
## Motivations

This PR deploys the new docs site alongside storybook. Storybook now lives at `/storybook` and the new docs site handles some automatic redirects for us, to ensure existing storybook links/bookmarks will continue to work.

You can test out this functionality using my deployed cloudflare preview.

How it works:

* If you load the root domain directly https://jacob-new-docs-deployment2.atlantis.pages.dev
    * This by default redirects to `/storybook`. In the future, we'll load the new docs site by default!
* If you load a storybook path (`/?path=___`), it will always redirect you to storybook
    * e.g. https://jacob-new-docs-deployment2.atlantis.pages.dev/?path=/docs/components-images-and-icons-avatar--docs
    * redirects to:  https://jacob-new-docs-deployment2.atlantis.pages.dev/storybook/?path=/docs/components-images-and-icons-avatar--docs 
* If you load the site index with `?new` it will force the new docs site to render
    * e.g. https://jacob-new-docs-deployment2.atlantis.pages.dev/?new
* If you load any other url, it's assumed to be owned by the new docs site
    * e.g. https://jacob-new-docs-deployment2.atlantis.pages.dev/components will resolve
    * e.g. https://jacob-new-docs-deployment2.atlantis.pages.dev/doesnt-exist will render our SPA shell... but we do need to make a proper 404/not found route to handle that more nicely

### Changed

- Made the new docs site deploy alongside storybook

## Testing

See instructions above to verify the routes load correctly.

Locally, you will not see any differences - vite should load every docs page normally.


packages/site/src/utils/storybook.ts
```diff
@@ -0,0 +1,28 @@
...
+ * This checks if the page is attempting to load a storybook path.
+ *
+ * If the url contains `?path`, we redirect to `/storybook/?path=___`.
+ *
+ * If the url is the root page, we redirect to `/storybook` unless the `?new`
+ * param is set, which forces the new docs site to render.
+ *
+ * Any other path is assumed to be owned by the new docs site, and will render.
+ */
+export function handleStorybookRedirect() {
```
scotttjob, line 11: I adore this approach. I really like that you found a way to make this work within the confines of the docs site. And it really does work, I tried out the scenarios you provided and everything works as advertised.

I'm curious though, and we can chat about this: Could we potentially achieve the same effects with DNS/Rewrite rules at the domain/host level? If you and I were to sit down today and go through the backend UI and see what we can configure?

The ONLY reason, and I'll admit it's actually a relatively small reason and one I could be talked out of. You have to do the full round trip to the SPA, before you get redirected to Storybook. So technically the storybook experience gets a little worse while we roll out the new site.

Which ultimately, is not a huge deal.

So let's take a look today, and see if we can implement the logic you have here at that level. We could potentially leave a lot of this in place as a backup. And we'll time box ourselves so we don't burn too much trying to come up with a 'perfect' solution while the one you have here is proven and functional and it works and is easy to reason about.

Also either way, we will hold this PR back until we're okay with the new docs site leaking out if something doesn't work as intended in prod. Which I think is going to be early next week at the latest.

scotttjob (commented):
Soft Approved. This works, and the code looks great! 

I would like to explore our options at the hosting/DNS level (see longer comment), to see if we can implement some of the redirect logic there (which doesn't change 90% of this PR), only to improve the experience while we cut over to the new site.

We'll time-box ourselves though, because what you have here works and works well and ultimately the redirection will be temporary so there's diminishing returns on our time looking at the hosting level.

Excellent work!

scotttjob (approved):
Approved! Thanks for letting me use the branch to figure out some prod issues ahead of the soft launch.

I've now removed my code and will get it up in another branch.

Really appreciate that!

PR #2215, opened 2024-12-05
docs: More search disambiguation

chris-at-jobber:
## Motivations

Not all of our newly-added components had search disambiguation baked in

### Added

- new matching terms for components that can help new consumers ramp up faster 
- For example, if you've come from a place that has a "Stack" component, Content and Box will show up when you search for that item

### Fixed

- design results were not checking for additionalMatches in a case-insensitive fashion

## Testing

Search for things using terms like Notification, Stack, Layout, Dropdown
---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/componentList.ts
+++ b/packages/site/src/componentList.ts
@@ -15,21 +15,25 @@ export const componentList = [
     title: "Autocomplete",
     to: "/components/Autocomplete",
     imageURL: "/Autocomplete.png",
+    additionalMatches: ["Search", "Typeahead"],
   },
   {
     title: "Avatar",
     to: "/components/Avatar",
     imageURL: "/Avatar.png",
+    additionalMatches: ["User", "Facepile", "Profile"],
   },
   {
     title: "Banner",
     to: "/components/Banner",
     imageURL: "/Banner.png",
+    additionalMatches: ["Flash", "Alert", "Notification"],
   },
   {
     title: "Box",
     to: "/components/Box",
     imageURL: "/Box.png",
+    additionalMatches: ["Layout", "Container", "Stack"],
   },
   {
     title: "Button",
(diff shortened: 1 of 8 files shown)


packages/site/src/layout/SearchBox.tsx
```diff
@@ -33,8 +33,12 @@ export const SearchBox = ({
...
 
   const filteredDesignList = useMemo(() => {
-    return designList.filter(d =>
-      d.title.toLowerCase().includes(search.toLowerCase()),
+    return designList.filter(
+      d =>
+        d.title.toLowerCase().includes(search.toLowerCase()) ||
+        d.additionalMatches?.find(e =>
+          e.toLowerCase().includes(search.toLowerCase()),
+        ),
```
chris-at-jobber, line 41: @scotttjob this is literally a copy+paste of the `filteredComponentsList` function below, and we'll presumably want this for Hooks and whatnot as well? So _I'm_ fine with this for now but might be worth having one of yourself/Taylor/Becca/Aiden/Jacob take a peek at a "whatever you pass into search is case insensitive and accounts for additionalMatches" sort of standardized approach, because QAing the design changes it took me way too long to realize that `designList` wasn't hooked up the same way!

scotttjob: Yup we can definitely make this better. I think it was components to start with and then it was cheap at the time to just duplicate a second for design. 

Now we're entering the territory where DRY-ing out the solution has more upside than a simple copy and paste for the third time on hooks.

I'll make sure to add a ticket for this.

scotttjob (approved):

packages/site/src/componentList.ts
```diff
@@ -239,11 +254,13 @@ export const componentList = [
     title: "InputEmail",
     to: "/components/InputEmail",
     imageURL: "/InputEmail.png",
+    additionalMatches: ["Email Input"],
```
taylorvnoj, line 257: I don't think it hurts to have these listed, but if someone searches "email", InputEmail will come up anyway

chris-at-jobber: Sounds good! I'll merge this and then I can take over + manage any conflict resolution on your branch 🫡

PR #2226, opened 2024-12-09
docs: Create Implement Tab on New Docs Site

taylorvnoj:
## Motivations

The Design Tab can get pretty full, and sometimes there are developer specific topics that can live elsewhere. This PR is a PART 1 of splitting concerns.  

We're creating a (currently named) Implement Tab.  So far, all content can be sorted under a `Configuration` or `Developer Notes` heading.  

Note: There will be duplication between Design and Implement Tabs and this is because, we are pulling from Storybook for the Design Tab content and I don't want to remove information from there yet.  Once we are hard launched and the New Site is the main documentation site, we can update those `stories` files.

Screenshot 2024-12-11 at 9 43 58 AM

### Added

-  `site/src/content` component folders got a new `[componentName]Notes.mdx` file.  I decided to name is this way because we aren't set on a Tab name yet and notes encapsulates it in a general way
- `notes?` was added to our `ContentExport`. The Implement Tab will only be shown if a component has a corresponding `[componentName]Notes.mdx` file.
- `ComponentView` has a new Tab object for the Implement Tab

### Changed

- some `stories` files got updates; some grammar changes, some heading changes so all heading are the same per component.
- `FormatTime` `stories` got a bit of an update to provide more clarity

## Testing

Check that the Implement Tab, when present, has content.  If a component (ex Glimmer) doesn't have a `[componentName]Notes.mdx` file, make sure an Implement Tab isn't present.

Changes can be

---

Random photo of Atlantis


packages/site/src/content/AnimatedPresence/AnimatedPresenceNotes.mdx
```diff
@@ -0,0 +1,40 @@
+## Configuration
```
scotttjob, line 1: Do we want to include a comment in the original file(s) for these extractions letting others know that this has technically been duplicated in the short term?

I would hate for us to make a correction in the original docs somewhere and then lose it when we cut out that work later.

packages/site/src/content/AnimatedPresence/index.tsx
```diff
@@ -44,4 +45,5 @@ export default {
       ),
     },
   ],
+  notes: () => <Notes />,
```
scotttjob, line 48: 👍

packages/site/src/content/AnimatedSwitcher/index.tsx
```diff
@@ -38,4 +39,5 @@ export default {
       ),
     },
   ],
+  notes: () => <Notes />,
```
scotttjob, line 42: Yup like this pattern of linking MDX files from the content page. Great work. :)

packages/site/src/layout/ComponentView.tsx
```diff
@@ -106,6 +106,14 @@ export const ComponentView = () => {
         </>
       ),
     },
+    {
+      label: "Implement",
```
scotttjob, line 110: This looks okay to me and shouldn't conflict too heavily with either of the link PRs I have in flight.

packages/site/src/layout/ComponentView.tsx
```diff
@@ -118,6 +126,10 @@ export const ComponentView = () => {
         return false;
       }
 
+      if (!PageMeta?.notes && index === 3) {
```
scotttjob, line 129: This also looks okay. We'll have to adjust the auto-links after we merge, but not a huge lift to do that.

scotttjob (approved):
LGTM!. Let's work together on the merge, I have a feeling we're going to conflict because we were both working on the index files.

packages/site/src/content/AnimatedPresence/AnimatedPresenceNotes.mdx
taylorvnoj: Yes! Great idea! I will do that in my next set of components

PR #2229, opened 2024-12-10
feat(components): Autocomplete Custom Menu Rendering

MichaelParadis:
## Motivations

As a part of our composability and customizabilty work we wanted to add support for rendering a custom menu in the Autocomplete. To achieve this I added a `customMenuRender` prop that will be used to replace the content when the Autocomplete menu is opened.

To help achieve this I did the following:
 - Exposed two hooks to aid in handling keyboard navigation with the Autocomplete
   - `useKeyboardNavigation` for handling the existing keyboard navigation behaviour. This works when the customization of the menu only renders MenuOptions and doesn't include custom elements
   - `useCustomKeyboardNavigation` this is used when the Autocomplete menu has elements that aren't just options. The reason for this is that the Autocomplete can not be certain of how people want keyboard navigation to work. This works to prevent issues with custom selectable items such as a footer to create another option.
 - Expose various atoms that we use for rendering the content in the Autocomplete menu such as:
   - `MenuOption` which is the default behaviour of menu options in Autocomplete
   - `MenuOptionContent` the default rendering of the content of a `MenuOption`
   - `BaseMenuOption` the base wrapper of the `MenuOption` this component handles the highlighted and separator logic
   - Also added `UNSAFE_` props for overriding the styles of these atoms 

## Decisions

I decided to give full access to the Autocomplete Menu including when it is rendered when using the `customMenuRender`. The reason for this is that the existing `Autocomplete` wouldn't display a menu if no options were available. However, when custom rendering a Menu the consumer may want to show a "Create new X` option instead. 

I made the options type support generics, the reason for this is that if generics weren't provided then the `customMenuRender` wouldn't be aware of the properties of the options it is rendering which would require consumers to use typescript casting to be able to get the properties of an object for rendering.

I removed the XOR of `Option` and `Group` option for `OptionCollection` the reason for this is that I found it was causing headaches when introducing support for generic option types. I also found that `OptionCollection` wasn't actually enforcing that `initialOptions` and `getOptions` was only an array of `Option`s or `GroupOption`s

### Added

- feat(components): Added ability to render a custom menu in the Autocomplete
- Expose Atoms used for rendering the Menu in Autocomplete

## Testing

 - Verify existing `Autocomplete` stories work as the did previously (verify keyboard navigation and using a mouse, etc.)
 - Verify that the `Custom Rendering` story can be used with both a keyboard and a mouse
 - Verify existing Autocomplete usages work (see testing branch referencing this PR)

Changes can be

---

Random photo of Atlantis


packages/components/src/Autocomplete/Autocomplete.types.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export type OptionCollection = AnyOption[];
+export type AnyOption<
+  GenericOption extends Option | GroupOption = Option | GroupOption,
+> = GenericOption;
+
+export interface AutocompleteProps<
+  GenericOption extends AnyOption = AnyOption,
+  GenericOptionValue extends Option = Option,
+  GenericGetOptionsValue extends AnyOption = AnyOption,
```
MichaelParadis, line 28: The generic on GetOptionsValue is because some people were passing an initialOptions value that would be a different type to what was returned by `getOptions` e.g. Initial options would be a list of GroupOptions but getOptions would give Options based on another value in the form

packages/components/src/Autocomplete/Autocomplete.module.css
```diff
@@ -28,11 +27,12 @@
 }
 
 .options.visible {
-  display: block;
+  position: absolute;
```
MichaelParadis, line 30: Changed the popper visibility to use visibility to match the recommendations in the docs
https://popper.js.org/docs/v2/modifiers/hide/
It also helped with an issue when clicking on customRendered options that didn't use the `onMouseDown` callback

packages/components/src/Autocomplete/Menu/MenuWrapper.tsx
```diff
@@ -0,0 +1,81 @@
...
+      {...attributes.popper}
+    >
+      {children}
+    </div>
+  );
+}
+
+export const MenuWrapper = React.memo(MenuWrapperInternal);
+
+export function useMenuWrapper({
```
MichaelParadis, line 39: Rename to `useAutocompleteMenu`

packages/components/src/Autocomplete/Menu/MenuWrapper.tsx
```diff
@@ -0,0 +1,81 @@
...
+      style={{ ...popperStyles.popper, width: targetWidth }}
+      data-elevation={"elevated"}
+      {...attributes.popper}
+    >
+      {children}
+    </div>
+  );
+}
+
+export const MenuWrapper = React.memo(MenuWrapperInternal);
```
MichaelParadis, line 37: This doesn't need to be memoized

packages/components/src/Autocomplete/Menu/MenuWrapper.tsx
MichaelParadis: Add doc strings explaining this component (and others in the file)

packages/site/src/content/Autocomplete/AutocompleteExamples.tsx
```diff
@@ -0,0 +1,693 @@
...
+
+  return (
+    <MenuWrapper visible={menuVisible}>
+      {optionsToRender}
+      {footer}
+    </MenuWrapper>
+  );
+}
+
+export const AdvancedTemplateCode = `
```
MichaelParadis, line 493: I tried various ways to get the Canvas to render the code as well, since the Canvas can't use any state and expects a single component to render I had to break it out into `AdvancedCustomTemplate` but the show code button stopped working.

I tried using things like the [source](https://storybook.js.org/docs/api/doc-blocks/doc-block-canvas#source) option but it would also not work.

When I tried using the `Source` block and doing `AdvancedCustomTemplate.toString()` it was giving minified code. I am fine removing this is we wish

chris-at-jobber:
@MichaelParadis something worth checking in-product - in Storybook, I can have multiple custom Autocompletes open at once:
image

MichaelParadis:
@chris-at-jobber let me look was certain I fixed that

docs/components/Autocomplete/Autocomplete.stories.mdx
```diff
@@ -38,6 +39,623 @@ the user types in the input. This async request should return a list of possible
...
+    ({ label }) => label === "Planets"
+  )?.options;
+
+  return (
+    <Autocomplete
+      placeholder="Search for an option with section headings"
+      initialOptions={sectionHeadingOptions}
+      value={value}
+      onChange={newValue => setValue(newValue)}
+      customRenderMenu={({
```
ZakaryH, line 173: I see in the description you mention that one of the reasons for exposing the full menu is because the menu wouldn't be rendered if there are no options which is fair

looking at how much code this can result in even for a simple custom menu I do wonder if there isn't another way to unlock that behavior? maybe an `emptyOption` prop or something, and then only allow custom rendering of options

I still need to play around with this more, and maybe I'll change my mind on it but my initial thought is if people will find themselves needing access to the full menu where they need to rebuild a couple other layers - or if it's really just the options they need access to

ZakaryH:
using keyboard navigation I'm able to select a group heading as an option whereas clicking on it correctly won't select it 
Kapture 2025-01-29 at 10 32 19

docs/components/Autocomplete/Autocomplete.stories.mdx
MichaelParadis: I will double check my notes (and update the PR) but I believe another reason has if the Input lost focused but you wanted to continue rendering the menu (such as with a footer action that is keyboard accessible). If I only allowed custom rendering of options then actions such as a footer button or a header action would eventually need to become a prop or they would need to add a fake option to determine if it is the header /footer

ZakaryH:
is it expected/desired for there to be no way to close the menu with keyboard interactions? where pressing ESC doesn't do anything?

I know that's what we already do today, and so maybe adding that extra scope here isn't the right place to address it - but I do find the interaction feels a bit off

MichaelParadis:
@ZakaryH I kept the same behaviour as our existing Autocomplete, I am going to look at the selection issue though

Aiden-Brine:
Screenshot 2025-01-29 at 11 23 16 AM
I'm getting the following on the new docs site

packages/site/src/content/Autocomplete/AutocompleteExamples.tsx
Aiden-Brine: I went through basically the exact same process with my stuff, wrestled with it for a while, and wrote [this ticket](https://jobber.atlassian.net/browse/JOB-114601?focusedCommentId=281693) to come back and improve it later. So I think the way you did it is totally fine. I would maybe just suggest using a `Disclosure` or `<details>` directly to allow people to show/hide the long blocks of code as they need. Right now it is a bit of a long scroll if you aren't interested in one of the examples

MichaelParadis:
@Aiden-Brine have you done `npm run bootstrap`? I am pretty sure the new site doesn't pick up on new components without it

packages/components/src/Autocomplete/Menu/Menu.tsx
```diff
@@ -0,0 +1,94 @@
...
+  inputFocused,
+  inputRef,
+}: InternalCustomMenuProps<GenericOption, GenericOptionValue>) {
+  const { MenuWrapper, menuRef } = useAutocompleteMenu({ attachTo });
+
+  const menuContent = customRenderMenu({
+    options,
+    menuRef,
+    onOptionSelect,
+    selectedOption,
```
ZakaryH, line 87: on the fence about how I feel on this one

on one hand, I suppose it's kind of convenient to get the selectedOption as an arg if I want to do something special or different for that option's UI

on the other, don't I already know this? if I must provide `value` and update it as selections are made - I'd already have access to selected Option to compare against

Aiden-Brine:
> @Aiden-Brine have you done `npm run bootstrap`? I am pretty sure the new site doesn't pick up on new components without it

My bad, that fixed it

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -142,29 +93,30 @@ function AutocompleteInternal(
   }
 
   async function updateSearch() {
-    const updatedOptions: AnyOption[] = await getOptions(inputText);
-    const filteredOptions = updatedOptions.filter((option: AnyOption) =>
+    const updatedOptions = await getOptions(inputText);
```
Aiden-Brine, line 96: Might be worth adding a check for empty input text here to save a call to `getOptions` and filtering if the input is cleared. Instead, set options to the initialOptions if the inputText is empty

ZakaryH:
I've read the comments about keyboard navigation

>Exposed two hooks to aid in handling keyboard navigation with the Autocomplete
useKeyboardNavigation for handling the existing keyboard navigation behaviour. This works when the customization of the menu only renders MenuOptions and doesn't include custom elements

>useCustomKeyboardNavigation this is used when the Autocomplete menu has elements that aren't just options. The reason for this is that the Autocomplete can not be certain of how people want keyboard navigation to work. This works to prevent issues with custom selectable items such as a footer to create another option.

the first one, is there a reason we can't just provide the keyboard navigation behavior by default? I can't really think of a good argument for not implementing that, and so it makes me wonder why we'd have the consumer build it when using a `MenuOption`

the second one, yeah I guess that makes sense. it might make more sense once I try some very custom content, but I'm still tempted to provide a baseline keyboard experience

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -89,15 +32,23 @@ function AutocompleteInternal(
...
+    GenericOptionValue,
+    GenericGetOptionsValue
+  >,
   ref: Ref<InputTextRef>,
 ) {
-  const [options, setOptions] = useState(initialOptions);
-  const [menuVisible, setMenuVisible] = useState(false);
+  const [options, setOptions] = useState<
+    Array<GenericOption | GenericGetOptionsValue>
+  >(mapToOptions(initialOptions));
```
Aiden-Brine, line 46: I think this will cause `mapToOptions(initialOptions)` to be called on every render of the component. Instead I would suggest putting that value in a memoized variable first and then it will only be called on the first render

docs/components/Autocomplete/Web.stories.tsx
```diff
@@ -212,6 +237,422 @@ SectionHeading.args = {
...
+                {options.map((option, index) => {
+                  const optionStyle = getOptionStyling({
+                    option,
+                    highlightedIndex,
+                    options,
+                  });
+
+                  return (
+                    <MenuOption
+                      key={index}
```
ZakaryH, line 361: I ran into a non unique key problem in my own example and wanted to see how we avoided it in the webstory

I'm not sure this is a safe place to use the index since the options are not static. they could definitely change based on what we've filtered for with the search term.

ZakaryH:
is it possible to customize group headers when going with a full-custom menu?

here my data has "EMOJIS" and "NOT EMOJIS" as groups, and I was hoping to have some custom UI for them but they're being treated as normal selectable options

image

I've tried to identify which ones are group header or not, but at this point my `options` are the top level objects and not the nested options from this data structure

```
    const possibleOptions = [
      {
        label: "EMOJIS",
        options: [
          { label: "cry", value: "cry" },
          {label: "smile", value: "smile"}
        ]
      },
      {
        label: "NOT EMOJIS",
        options: [
          {label: "john", value: "johnny"},
          {label: "jane", value: "jane"},
          {label: "jacob", value: "jacob"},
        ]
      }
    ];
```

packages/components/src/Autocomplete/Menu/Menu.tsx
MichaelParadis: I decided to do this so consumers wouldn't need to have their custom menu be a Component declared inside of their wrapper component. If I didn't expose this they would either need to use some sort of provider or create a higher order component that adds it to the scope for example.
```tsx

function MyUsage() {
    const [value, setValue] = useState()
    // In order for value to be in the scope
    function CustomMenu(props) {
        const selectedOption = props.options.find((option) => option === value)
        return //My content
    }
    return (
        <Autocomplete value={value} onChange={setValue} 
        renderCustomMenu={args => <CustomMenu {...args}>}
        />
    )
}
```
as I wrote this I realized they could do 


```tsx

function CustomMenu(props: CustomOptionsMenuProp & {value?: <SomeType>}) {
    const selectedOption = props.options.find((option) => option === props.value)
    return //My content
}

function MyUsage() {
    const [value, setValue] = useState()

    return (
        <Autocomplete value={value} onChange={setValue} 
        renderCustomMenu={args => <CustomMenu value={value} {...args}>}
        />
    )
}
```

I personally prefer the existing way since the consumer doens't really need remember the selected option because it will be there from the Autocomplete

MichaelParadis: My main thought was that people may make shared Autocomplete menus that are reused and they may want to be able to isolate them instead of having a larger component. For example by providing it they could have it be imported from a completely different file.

packages/components/src/Autocomplete/Autocomplete.tsx
MichaelParadis: Good catch, I had done `useState` that use a variable value and it was only set once but I guess functions in useStates are different. I will update it, just fixing something else right now

packages/components/src/Autocomplete/Menu/MenuWrapper.tsx
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Provides a wrapper for the Autocomplete menu that handles positioning and visibility.
+ * @param attachTo - The element that the menu should be attached to.
+ */
+export function useAutocompleteMenu({
+  attachTo,
+}: {
+  attachTo: React.RefObject<Element | null>;
+}) {
+  const [menuRef, setMenuRef] = React.useState<HTMLElement | null>();
```
Aiden-Brine, line 46: I am confused about the purpose of this state and lines 56-58. We are updating this `menuRef` value whever `prop.menuRef` changes. Could we just remove this line and lines 56-58 and then return props.menuRef on line 75? Or am I missing some added benefit here?

packages/components/src/Autocomplete/Menu/MenuWrapper.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const Wrapper = useCallback(
+    ({
+      children,
+      visible,
+    }: {
+      children?: React.ReactNode;
+      visible: boolean;
+    }): React.ReactElement => {
+      const props = useRepositionMenu(attachTo, visible);
```
Aiden-Brine, line 55: Should visible be added to the dependencies for the useCallback or should this line be moved out of the useCallback?

packages/components/src/Autocomplete/Option.tsx
```diff
@@ -0,0 +1,179 @@
...
+    styles.option,
+    {
+      [styles.active]: isHighlighted,
+      [styles.separator]: addSeparators,
+    },
+    UNSAFE_className,
+  );
+
+  return (
+    <button
```
Aiden-Brine, line 171: My editor flagged an eslint warning that this should have a `type`. Should it also have an Aria role?

packages/components/src/Autocomplete/useKeyboardNavigation.ts
```diff
@@ -0,0 +1,123 @@
...
+  event: KeyboardEvent;
+  direction: KeyboardAction;
+  options: T[];
+  highlightedIndex: number;
+}) {
+  event.preventDefault();
+  const requestedIndex = options[highlightedIndex + direction];
+
+  return requestedIndex && isGroup(requestedIndex)
+    ? direction + direction
```
Aiden-Brine, line 121: Is it possible that direction + direction is also a group?

packages/components/src/Autocomplete/useRepositionMenu.ts
```diff
@@ -0,0 +1,31 @@
...
+  readonly menuRef: HTMLElement | null;
+  readonly setMenuRef: (ref: HTMLElement | null) => void;
+  readonly targetWidth: number | undefined;
+}
+
+export function useRepositionMenu(
+  attachTo: MenuProps["attachTo"],
+  visible = false,
+) {
+  const [menuRef, setMenuRef] = useState<HTMLElement | null>();
```
Aiden-Brine, line 16: I think this should default to null based on the typing

packages/components/src/Autocomplete/useRepositionMenu.ts
```diff
@@ -0,0 +1,31 @@
...
+      { name: "offset", options: { offset: [0, 8] } },
+      { name: "flip", options: { fallbackPlacements: ["top"] } },
+    ],
+  });
+
+  useSafeLayoutEffect(() => {
+    popper?.update?.();
+  }, [visible]);
+
+  const targetWidth = attachTo.current?.clientWidth;
```
Aiden-Brine, line 28: Should this also be wrapped in `useSafeLayoutEffect` in case `attachTo` is resized after this has already been called?

MichaelParadis: Popper adjusts a lot of the styling automatically. This `targetWidth` is helpful for initial positioning but once visible has changed `react-popper` starts controlling the rendering

packages/components/src/Autocomplete/Autocomplete.tsx
MichaelParadis: I would argue no and that the empty check in `updateInput` should actually be removed and the `updateSearch` should be called in the `useEffect` on line 55. My assumption as to why line 55 was done is to create a better UX when clearing the input.

 I found that some consumers were setting an `initialOptions` but based on other form state they would want to return different options in their `getOptions` if this were to be added this behaviour would break and create a worse experience for consumers

Aiden-Brine: Ah interesting, I didn't realize that folks were doing that and that it would be a breaking change. Fair enough

packages/components/src/Autocomplete/useKeyboardNavigation.ts
Aiden-Brine: Okay, thank you for clarifying that for me

packages/components/src/Autocomplete/Menu/MenuWrapper.tsx
Aiden-Brine: Chatted about this on a call and it cleared up my questions

packages/components/src/Autocomplete/Menu/MenuWrapper.tsx
Aiden-Brine: Same with this, chatted about it and it cleared up my questions

packages/components/src/Autocomplete/useRepositionMenu.ts
MichaelParadis: Updated in 59f69741e1e5952ec2c0c070663a3d9e536d18a2

packages/components/src/Autocomplete/Option.tsx
MichaelParadis: Updated in https://github.com/GetJobber/atlantis/commit/3b4a04a2d4a92b905f3cffb2e68cb9c3b89341c0

packages/components/src/Autocomplete/Autocomplete.tsx
MichaelParadis: Fixed in https://github.com/GetJobber/atlantis/pull/2229/commits/aef46223ce44a381079215b05b3c119e49af2649

MichaelParadis:
@ZakaryH I updated storybook and the docs to include information on rendering Autocompletes with custom sections. I am just updating my pre-release branch right now

MichaelParadis:
@ZakaryH

> I've read the comments about keyboard navigation
> 
> > Exposed two hooks to aid in handling keyboard navigation with the Autocomplete
> > useKeyboardNavigation for handling the existing keyboard navigation behaviour. This works when the customization of the menu only renders MenuOptions and doesn't include custom elements
> 
> > useCustomKeyboardNavigation this is used when the Autocomplete menu has elements that aren't just options. The reason for this is that the Autocomplete can not be certain of how people want keyboard navigation to work. This works to prevent issues with custom selectable items such as a footer to create another option.
> 
> the first one, is there a reason we can't just provide the keyboard navigation behavior by default? I can't really think of a good argument for not implementing that, and so it makes me wonder why we'd have the consumer build it when using a `MenuOption`
> 
> the second one, yeah I guess that makes sense. it might make more sense once I try some very custom content, but I'm still tempted to provide a baseline keyboard experience

The main reason for this is we could need to call default keyboard behaviour in the parent above and provide the "default" highlighted index as a prop to the `renderCustomMenu`. I could see 2 possible issues:
1. Since that prop would update on keyboard events the menu would have extra unneeded rerenders
2. People might see highlighted index prop and not consider that there may be issues with keyboard navigation. It makes the keyboard navigation a more explicit decision to consider if custom navigation is needed

ZakaryH:
is there anything different or special about these Add buttons? I noticed they're persisting for just a bit longer than the rest of the menu when I choose an option which is weird since I'd have thought their visibility would be exactly the same as everything else in the menu

Kapture 2025-02-03 at 14 33 50

ZakaryH:
@MichaelParadis nice, yeah that looks much better!

I haven't looked at the code too closely so I'll let @Aiden-Brine give the final approval, but for the things I was playing around with and all the comments I had made this LGTM

Aiden-Brine (approved):
Thank you for making those changes and discussing everything! LGTM

PR #2231, opened 2024-12-10
refactor(components-native): surface--element instead of background as background color for non surface/container components

edison-cy-yang:
## Motivations
- UI components that use `surface--background` have poor contrast on some surfaces

## Changes
- Components in components-native that use `surface--background` now have better visibility on all surfaces

### Added

- 

### Changed
- Changed all UI components in the components-native package (excluding surfaces & containers) that used `color-surface--background` to use `color-surface--element` instead

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/Chip/Chip.style.ts
+++ b/packages/components-native/src/Chip/Chip.style.ts
@@ -29,6 +29,6 @@ export const styles = StyleSheet.create({
     padding: tokens["space-smaller"],
   },
   activeDismissIcon: {
-    backgroundColor: tokens["color-surface--background"],
+    backgroundColor: tokens["color-surface--element"],
   },
 });
--- a/packages/components-native/src/InputText/context/InputAccessoriesProvider.style.tsx
+++ b/packages/components-native/src/InputText/context/InputAccessoriesProvider.style.tsx
@@ -14,7 +14,7 @@ export const styles = StyleSheet.create({
     height: BAR_HEIGHT,
   },
   lightTheme: {
-    backgroundColor: tokens["color-surface--background"],
+    backgroundColor: tokens["color-surface--element"],
   },
   darkTheme: {
     // PlatformColor has to be conditional for Storybook to run without error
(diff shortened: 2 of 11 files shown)


packages/components-native/src/Chip/Chip.style.ts
```diff
@@ -29,6 +29,6 @@ export const styles = StyleSheet.create({
     padding: tokens["space-smaller"],
   },
   activeDismissIcon: {
-    backgroundColor: tokens["color-surface--background"],
+    backgroundColor: tokens["color-interactive--background"],
```
chris-at-jobber, line 32: I don't actually recall _why_ we need a different background color here, but that's probably a question for another day

packages/components-native/src/InputText/context/InputAccessoriesProvider.style.tsx
```diff
@@ -14,7 +14,7 @@ export const styles = StyleSheet.create({
     height: BAR_HEIGHT,
   },
   lightTheme: {
-    backgroundColor: tokens["color-surface--background"],
+    backgroundColor: tokens["color-interactive--background"],
```
chris-at-jobber, line 17:
```suggestion
    backgroundColor: tokens["color-surface--background"],
```

This is the toolbar above the native keyboard for things like the "Done" button, this isn't a "UI element" in the sense of the other components here

packages/components-native/src/InputText/context/InputAccessory.style.ts
```diff
@@ -7,7 +7,7 @@ export const styles = StyleSheet.create({
     justifyContent: "space-between",
     alignItems: "center",
     paddingHorizontal: tokens["space-small"],
-    backgroundColor: tokens["color-surface--background"],
+    backgroundColor: tokens["color-interactive--background"],
```
chris-at-jobber, line 10:
```suggestion
    backgroundColor: tokens["color-surface--background"],
```

Fairly sure this is related to the same element as the previous file

chris-at-jobber (changes requested):
Same as web, mostly good with a few small changes!

chris-at-jobber (approved):
And that's the lot of them! Good to go @edison-cy-yang

PR #2235, opened 2024-12-10
feat(components): Make List Section Header Flexible

ZakaryH:
## Motivations

we want the heading of sections to be less rigid, allowing the ability to provide an alternative layout and heading if desired.

**"Other" default section and new prop**
additionally, during the first set of List changes we noticed there is some not obvious behavior when it comes to sectioned List that are missing one or more entries for `section`. the way it works is that if a single item in `items` has a `section` key/value pair, you have now opted into a sectioned list.

it is assumed, but not enforced in any way, that every item in items will belong to a section meaning you must provide `section` for every item to group it with all the other items of that section. if you are missing a `section` then we will create one for you with the heading value of "Other". this behavior was not previously document, nor seemingly used as far I can tell.

while removing it is very tempting, it would be a breaking change. as such we're going to keep the default of "Other", and offer a new prop to provide your own value since "Other" isn't guaranteed to make sense in every content. we considered adding an option to opt out of the behavior entirely, but that starts getting to be quite complex. List has some strong opinions on how it should be used that are paired with the structure of the data it accepts. where would un-sectioned items go? would they all go into the last known section?

**customRenderSection**
compared to the item's custom render function, we are offering much more limited customization. for the section we are only passing the section's title as the arg to the render function. this is because `items` are all data about the item, with `section` being a very brief (and repetitive) description of what section it belongs to. 

to pass in arbitrary data for the section's custom render function to use, we'd either need to pollute the item's data with section specific values, change the type of `section` or add another prop for section details specifically. all of these have their various downsides. while this approach is limiting, we don't have a specific need outside of text just yet.

ultimately, the API for this component isn't set up well to interface with the sectionedList in a clean way. extending it from the current form creates cumbersome APIs. if people are finding that `customRenderSection` isn't providing enough flexibility, I would recommend either creating a new component entirely or changing the syntax of `List` to invoke a different version of it such as 
```
<List>
  <List.Section>
   ...
  </List.Section>
   {items.map(item => <List.Item item={item}/>}
</List>
```

### Added

new prop to decide what the section-less items should be called

ability to provide a render function to customize the section header

ability to opt out of the styles of the section header, providing more of a blank slate

aria-labelledby prop

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


docs/components/List/Web.stories.tsx
ZakaryH, line 1: a specific example for the section header might not be a bad idea

packages/components/src/List/List.test.tsx
```diff
@@ -274,3 +274,89 @@ describe("When a list is provided a custom render function", () => {
...
+            id: 1,
+            content: "Build the deck",
+            onClick: () => {
+              /* do stuff */
+            },
+          },
+          {
+            id: 2,
+            content: "Hey I'm walkin here",
+            section: "Gabbagool",
```
Aiden-Brine, line 323: 😆

Aiden-Brine:
@ZakaryH The code looks good to me. I would like to see a Storybook example with a semi-interesting custom section header (and maybe one with custom header styles? but I could certainly be convinced that is overkill). Also, the docs page talks about `customRenderItem`, `customItemStyles` and the behaviour when a section header is missing so I think it would be good to update those with the new functionality available.

ZakaryH:
@Aiden-Brine added a somewhat interesting example, and updated the docs!

docs/components/List/List.stories.mdx
```diff
@@ -172,7 +178,7 @@ and no border-bottom. You can also experiment by toggling this prop in the other
 custom render examples to observe its impact.
 
 Note: The `customItemStyles` prop is only valid when a `customRenderItem`
-function is provided.
```
Aiden-Brine, line 175: This makes it sound to me like `customSectionStyles` is only valid if `customRenderItem` is provided but I believe it is only valid if `customRenderSection` is provided?

docs/components/List/List.stories.mdx
```diff
@@ -104,11 +107,14 @@ occurrence that has no section.
...
-requirements. If a custom render function is provided onClick, url, and hover
-functionality will still work. Sectioned list functionality will also remain the
-same.
+**custom render function** using one or more of the `customRenderItem` or
+`customRenderSection` props.
+
+This is useful if the default rendering of the `ListItem` or section header
+component doesn’t meet your requirements. If a custom render function is
+provided onClick, url, and hover functionality will still work. Sectioned list
+functionality will also remain the same, only the appearance will change if a
```
Aiden-Brine, line 116: I would suggest being more explicit here and say "only the appearance of the header will change" since `customRenderSection` only changes the appearance of the header

Aiden-Brine (changes requested):
I think just the last couple of docs touch ups

taylorvnoj:
@ZakaryH 👋🏻  In `List.stories.mdx` you'll see a new message about new information being duplicated into a corresponding `packages/site` docs page.  It would be great to get that information duplicated over into `ListNotes.mdx`. 
If you want a run down/explanation let me know! Happy to chat!

docs/components/List/List.stories.mdx
ZakaryH: fair. ok pushed something, lemme know if that cleared it up!

ZakaryH:
@taylorvnoj new docs also updated

packages/site/src/content/List/ListNotes.mdx
```diff
@@ -3,11 +3,16 @@
 ### Custom Rendering (Advanced Usage)
 
 If you need more control over the rendering of list items, you can provide a
-**custom render function** using the `customRenderItem` prop. This is useful if
-the default rendering of the `ListItem` component doesn’t meet your
-requirements. If a custom render function is provided onClick, url, and hover
-functionality will still work. Sectioned list functionality will also remain the
-same.
+**custom render function** using one or more of the `customRenderItem` or
```
taylorvnoj, line 6: Thank you!

Aiden-Brine (approved):
Thank you for making those changes

PR #2236, opened 2024-12-11
docs: Link Checking + E2E + Correct Storybook Links for All Components

scotttjob:
## Motivations

Most of the "Storybook" links in the right-hand panel were broken prior to this PR and I wanted to fix those.

But since those are just links based on strings in files which are relatively brittle to change, I wanted to introduce a regression test to make sure that future changes don't disrupt all of the links being functional.

So I've introduced a Playwright based test suite that is entirely optional to run. It's not currently part of any automation, but you can run the test suite whenever you want to find any broken links.

We could certainly add it to our automation in the future, but it's a little expensive to run. It takes about 4 minutes currently for a successful run. It runs each test 3 times, once for Chrome, Firefox, and Safari. In Desktop mode. We can optionally run more devices in the future if we find this useful.

## Changes

1. Fixed all the Storybook links in the side panel.
2. Added a Playwright based regression test to ensure all of our pages are rendering properly (with the exception of the design pages, will update in a future ticket)
3. Fixed rendering of the SideDrawer component (found by the tests!)
4. Broke out the base list of components in generateDocs into a shared .mjs file so I didn't have to duplicate the list of components.
5. Introduced a VITE_STORYBOOK_URL env var that swaps the link depending on which environment you're running in.

## Testing

1. Pull this branch down.
1. Run `npm ci`. 
1. Run `npx playwright install` This is going to take a minute to pull down the headless browsers.
1. Start the docs site
1. Start storybook
1. `npm run e2e` when you're in `packages/site`
1. The test suite should take about 5 minutes to run. It'll print out the component names as it runs through to the console. There is an HTML file thats generated at the end that you can look at if you're interested.

If the test suite does not run successfully, let me know! I might have to amend these instructions with more details.

Changes can be

---

Random photo of Atlantis


packages/site/index.html
```diff
@@ -1,12 +1,20 @@
...
-  </head>
-  <body>
-    <div id="root"></div>
-    <script type="module" src="/src/main.tsx"></script>
-  </body>
-</html>
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
scotttjob, line 6: We'll need a proper favicon soon and title in here I would imagine.

packages/site/src/components/ComponentLinks.tsx
```diff
@@ -21,11 +21,13 @@ export const ComponentLinks = ({
         <Heading level={2}>Links</Heading>
         <Box>
           {links?.map((link, index) => (
-            <Box key={index}>
-              <Link key={index} url={link.url}>
-                {link.label}
-              </Link>
-            </Box>
+            <div key={index} data-storybook-link>
```
scotttjob, line 24: I think there might be a better way to do this, I was having issues initially selecting the link text but didn't circle back to this, I think we could drop a ticket to get this out in the future (maybe when we get the design pages in).

packages/site/src/globalTypes.d.ts
```diff
@@ -0,0 +1,7 @@
+interface AtlantisWindow extends Window {
```
scotttjob, line 1: Is there a better way to do this in 2024? This is how I did it forever ago and GPT also gave me this answer but it always felt wrong, like there should be a generic or something we can use instead.

packages/site/src/layout/ComponentView.tsx
```diff
@@ -58,7 +58,11 @@ export const ComponentView = () => {
   }, [code, iframe?.current, iframeMobile?.current, type]);
 
   useEffect(() => {
-    if (type === "web" && !PageMeta?.component?.element) {
+    if (
```
scotttjob, line 61: This got into a nasty update loop if there was no component (a 404). This prevents the nasty loop.

packages/site/tests/site.components.e2e.ts
```diff
@@ -0,0 +1,100 @@
...
+    ...ListOfGeneratedWebComponents,
+  ].sort();
+
+  const uniqueSet = new Set(combinedList);
+
+  return Array.from(uniqueSet);
+};
+
+const uniqueList = buildUniqueComponentList();
+
```
scotttjob, line 20: I'm not the most elegant tester so there's probably better way to do a lot of this but as a first pass this works! It cycles through all the component pages making sure they're rendering properly, then it loads the links in right hand nav and follows them. If they're valid storybook pages, the tests pass!

Beccaneer (approved):
Looks good to me! The only thing I needed to do was run `npx playwright install` before the e2e tests

Beccaneer:
Also,  it's showing `packages/site/playwright-report/index.html` has changes after running the e2e tests - it might need to be gitignored? 🤔

packages/site/src/content/ActionItem/index.tsx
```diff
@@ -19,7 +20,9 @@ export default {
   links: [
     {
       label: "Storybook",
-      url: "http://localhost:6006/?path=/docs/components-utilities-ActionItem-web--docs",
+      url: getStorybookUrl(
+        `?path=/docs/components-layouts-and-structure-actionitem--docs`,
+      ),
```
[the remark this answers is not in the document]
scotttjob: Just to follow-up here publicly, yup! they're meant to be different. The should line up with the actual component links now, beforehand they were generated strings as part of the initial creation process (we always knew we'd have to go back and do this PR that corrected all the links).

Thankfully we added an E2E suite at the same time, so we can determine if any of the links break again in the future!

You and I should chat about potentially putting that E2E suite as part of our deploy process.

PR #2237, opened 2024-12-11
docs: Adjusted SideNav styling for new docs site [JOB-110489]

Beccaneer:
### Changed

Updated styling for the side nav menu. 
Screenshot 2024-12-11 at 10 09 08 AM

Screenshot 2024-12-11 at 10 09 18 AM

## Testing

You can compare how it looks with the Figma file https://www.figma.com/design/HhNrPoreO9D8R8O9Ta2AM7/Unified-Atlantis-Docs?node-id=1021-12320&node-type=frame&m=dev and make sure the side nav scrolls as expected and the links work as expected etc :)

Changes can be
---

Random photo of Atlantis


packages/site/src/assets/JobberLogo.svg.tsx
```diff
@@ -1,12 +1,18 @@
...
+        fill="#012939"
```
chris-at-jobber, line 14: `fill` should be using a semantic token - hardcoding to #012939 breaks in dark mode:
image

**Before** (using `--surface--reverse`)
image

**After**
image

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
```diff
@@ -14,27 +14,40 @@ function AnimatedPresenceDisclosure({
...
+  };
+
   return (
-    <>
+    <div>
       <Link
         to={to ?? "/"}
         href="https://www.google.com"
         tabIndex={0}
         onFocus={e => (e.currentTarget.style.boxShadow = "0 0 3px #0073e6")}
```
chris-at-jobber, line 35: Can we use the `--shadow-focus` token here?

Also noticing that the collapsable headers have an unstyled focus state:
image

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
```diff
@@ -14,27 +14,40 @@ function AnimatedPresenceDisclosure({
...
         onBlur={e => (e.currentTarget.style.boxShadow = "none")}
+        onMouseEnter={handleHover}
+        onMouseLeave={handleHover}
         style={{
           display: "flex",
           alignItems: "center",
           justifyContent: "space-between",
-          backgroundColor: "var(--color-surface--background)",
+          backgroundColor: isHovered
+            ? "var(--color-surface)"
```
chris-at-jobber, line 44: - Should have a transparent (or `--color-surface--background`) in rest state, and `--color-surface--background--hover` when hovered
image

- the `surface` (white in light mode) is for the current page state

packages/site/src/layout/NavMenu.module.css
```diff
@@ -0,0 +1,60 @@
+
+.navMenuSticky {
+  background-color: var(--color-surface--background);
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: column;
+  gap: var(--space-large);
+  padding: 36px var(--space-base) var(--space-large) var(--space-base);
+  position: fixed;
```
chris-at-jobber, line 9: This seems a bit overcomplicated, couldn't we have a single container using flex to achieve this? Then we can avoid needing to set the 188px, using `fixed`, using z-index, doing the height calc on the menu container?

packages/site/src/assets/JobberLogo.svg.tsx
Beccaneer: oops! I literally copied it from Figma 🙈  fixed

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
Beccaneer: Focus states are addressed in https://jobber.atlassian.net/browse/JOB-111070

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
Beccaneer: Oh okay! I am a bit confused because the Figma shows the backgrounds as white when hovered on? Is this what it is supposed to be? 
Screenshot 2024-12-11 at 12 02 42 PM

packages/site/src/layout/SearchBox.tsx
```diff
@@ -101,7 +101,7 @@ export const SearchBox = ({
             </Content>
           )}
           {filteredDesignList.length > 0 && (
-            <Box padding={{ top: "largest" }}>
+            <Box>
```
chris-at-jobber, line 104: Do we actually still need this to be a Box if we're not using it to apply any styling? (I guess it could stick around and if we need it, there's no real negative impact of keeping it here.)

packages/site/src/layout/NavMenu.module.css
Beccaneer: I updated with your suggestion, let me know if you like it

packages/site/src/layout/SearchBox.tsx
Beccaneer: Uhmmm tbh dunno! It doesn't seem to break anything when removed... so it's probably okay? 😅

packages/site/src/layout/NavMenu.module.css
```diff
@@ -0,0 +1,60 @@
+
+.navMenuContainer {
+    display: flex;
+    flex-direction: column;
+    padding: 36px var(--space-base) 0 var(--space-base);
+    height: calc(100dvh - 36px);
```
chris-at-jobber, line 6: If we set `box-sizing: border-box`, the `calc` here can go away and `100dvh` will work

Beccaneer: Screenshot 2024-12-11 at 1 10 28 PM
Screenshot 2024-12-11 at 1 10 35 PM

This is making it skinnier. Would it be easier to make a ticket and assign it to you to go through the css?

packages/site/src/layout/NavMenu.module.css
```diff
@@ -0,0 +1,60 @@
+
+.navMenuContainer {
+    display: flex;
+    flex-direction: column;
+    padding: 36px var(--space-base) 0 var(--space-base);
```
chris-at-jobber, line 5: I think we need to move the left and right padding off of the parent and onto the children - this approach "traps" the scrollbar inside of the nav items in a way that overlaps, and clips the edges of the focus states:
image

Moving the left and right padding to the children gives the scrollbar some room to breathe and leaves the focus states unclipped:
image

Beccaneer: You seem to want things in a very specific way, and it is just adding time going back and forth through these comments, so maybe it would be better if you wanted to make the changes on my branch directly? I don't mind!

chris-at-jobber: Sure! I can push a few things up 👍

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
chris-at-jobber, line 1: I know we have a separate ticket for this, but I was finding it really hard to suss out how it was all going to come together, particularly with element alignment, until I made adjustments here. Now that I have these adjustments, I could in theory cherry pick them out and land that first?

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
```diff
@@ -14,56 +15,61 @@ function AnimatedPresenceDisclosure({
...
+        <Link
+          to={to ?? "/"}
+          href="https://www.google.com"
+          tabIndex={0}
+          onFocus={e =>
+            (e.currentTarget.style.boxShadow = "var(--shadow-focus)")
+          }
+          onBlur={e => (e.currentTarget.style.boxShadow = "none")}
+          onMouseEnter={handleHover}
+          onMouseLeave={handleHover}
```
chris-at-jobber, line 42: I'm actually not sure if I still need these now that some of the styling lives in the CSS module

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
```diff
@@ -14,56 +15,61 @@ function AnimatedPresenceDisclosure({
...
-          aria-label="Toggle Changelog"
         >
-          <Icon name="arrowDown" />
-        </button>
-      </Link>
+          <Typography fontWeight="semiBold" size="large" textColor="heading">
+            {title}
+          </Typography>
+        </Link>
+        <Button
```
chris-at-jobber, line 53: Notice that the button is no longer a child of the Link, which was an issue semantically and I think might be why we have the `preventDefault` in place up on L25-26?

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
Beccaneer: No reason to pick it out if it's done imo! Just update/close the relevant tickets 👍

packages/site/src/layout/NavMenu.module.css
```diff
@@ -0,0 +1,97 @@
+.navMenuContainer {
+    --hoverOffset: var(--space-smallest);
+    --negativeHoverOffset: calc(-1 * var(--hoverOffset));
+    --subItemInset: calc(var(--space-small) + var(--space-smaller));
```
chris-at-jobber, line 4: I can see a case for just doing these inline, but setting them as vars (I hope) makes it more explicit what they're doing

Side note for Future Chris: the calc to get 12px is one of the most common "custom CSS" patterns I've seen, even in our own components, so look into supporting that value as a default token

packages/site/src/layout/NavMenu.module.css
```diff
@@ -0,0 +1,98 @@
...
+.navMenu li {
+  margin-bottom: var(--space-smaller);
+}
+
+.navMenuItem {
+  align-items: center;
+  display: flex;
+  min-height: var(--navItemHeight);
+  overflow-wrap: anywhere;
+  transition: all var(--timing-base) ease-out;
```
chris-at-jobber, line 46: other note for future: this is a really standard declaration we set on interactive elements. could we reduce overhead / mental effort by creating a "compound" token or util class that gives you standard interaction stuff Out Of The Box?

packages/site/src/layout/NavMenu.tsx
```diff
@@ -18,116 +19,102 @@ export const NavMenu = () => {
...
-        <Button
-          onClick={() => setOpen(true)}
-          label="Search"
-          icon="search"
-          variation="subtle"
-        />
-      </Box>
-      <SearchBox open={open} setOpen={setOpen} />
-      <MenuList>
+    <nav className={styles.navMenuContainer}>
```
chris-at-jobber, line 22: semantic element

packages/site/src/layout/NavMenu.module.css
```diff
@@ -0,0 +1,98 @@
...
+    --hoverOffset: var(--space-smallest);
+    --negativeHoverOffset: calc(-1 * var(--hoverOffset));
+    --subItemInset: calc(var(--space-small) + var(--space-smaller));
+    --navItemHeight: 40px;
+    display: flex;
+    flex-direction: column;
+    padding: 36px 0 0 0;
+    height: 100dvh;
+    background-color: var(--color-surface--background);
+    width: 200px;
```
chris-at-jobber, line 11: Wider than I'd originally mocked up - the 188 was referencing Jobber's sidenav, but we don't necessarily need to be married to that value and this solves a great deal of the wrapping issues

packages/site/src/layout/NavMenu.module.css
```diff
@@ -0,0 +1,98 @@
...
+    box-sizing: border-box;
+}
+
+.navMenuContainer > * {
```
chris-at-jobber, line 15: Could change this declaration to 

```suggestion
.navMenuHeader,
.navMenu {
```

if preferred

packages/site/src/layout/NavMenu.module.css
```diff
@@ -0,0 +1,96 @@
...
+
+.disclosureNavItem {
+  display: flex;
+  gap: var(--space-smaller);
+  border-radius: var(--radius-small);
+  transition: all var(--timing-base) ease-out;
+  margin-left: var(--negativeHoverOffset);
+  margin-right: var(--negativeHoverOffset);
+  padding-left: var(--hoverOffset);
+  padding-right: calc(var(--space-smallest) + var(--hoverOffset));
```
chris-at-jobber, line 83: Bit of a tricky means to have the rest state align with the overall structure, while avoiding the hover state touching the label / button boundaries

packages/site/src/layout/NavMenu.module.css
```diff
@@ -0,0 +1,96 @@
...
+  min-height: var(--navItemHeight);
+  align-items: center;
+  justify-content: space-between;
+
+  &:hover {
+    background-color: var(--color-surface--background--hover);
+  }
+
+  button {
+    border-radius: var(--radius-small);
```
chris-at-jobber, line 93: `base` radius inside of a `small` is gross

packages/site/src/componentList.ts
chris-at-jobber, line 1: Renames all of the " and " section titles to " & "

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
```diff
@@ -14,56 +15,61 @@ function AnimatedPresenceDisclosure({
...
-          {title}
-        </Typography>
-        <button
-          type="button"
-          onClick={handleButtonClick}
+    <div>
+      <span className={styles.disclosureNavItem}>
+        <Link
+          to={to ?? "/"}
+          href="https://www.google.com"
```
Aiden-Brine, line 35: It appears that `href` is not a [valid prop for Link](https://api.reactrouter.com/v7/interfaces/react_router.LinkProps.html)

chris-at-jobber: oh nice, since I'm assuming we actually don't want to send people to Google dot com can we just remove that line?

Aiden-Brine:
Screenshot 2024-12-11 at 3 20 13 PM

Apologies if this has been discussed but if I click on something that has sub items (like Components) a border remains, but if I click on something like Design that does not I never see a border.

Also, for a nav menu item like Design the whole hoverable part is actionable. For something like Components there is quite a bit of area that is hoverable but not actionable (The area around the outside, the area between the title and the button to expand).

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
```diff
@@ -14,56 +15,61 @@ function AnimatedPresenceDisclosure({
...
+          <Typography fontWeight="semiBold" size="large" textColor="heading">
+            {title}
+          </Typography>
+        </Link>
+        <Button
+          variation="subtle"
+          size="small"
+          type="tertiary"
+          onClick={handleButtonClick}
+          ariaLabel="Toggle Changelog"
```
Aiden-Brine, line 58: Should this value be dynamic?

chris-at-jobber: Oh yes it should be

packages/site/src/layout/NavMenu.tsx
```diff
@@ -18,116 +19,102 @@ export const NavMenu = () => {
...
+          <Content spacing="smaller">
+            {routes?.map((route, routeIndex) => {
+              if (route.inNav === false) return null;
 
-            interface MenuItem {
-              handle: string;
-              children?: MenuItem[];
-              path?: string;
-            }
+              interface MenuItem {
```
Aiden-Brine, line 45: I could be missing something but is there a benefit to defining this interface and the below functions here as opposed to elsewhere in the file? I think that would really improve the readability over all the nesting we have at the moment

packages/site/src/layout/SearchBox.tsx
Aiden-Brine: If there are uses of `Box` that aren't currently providing any benefits styling wise I would advocate for removing them

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
chris-at-jobber: Done [here](https://github.com/GetJobber/atlantis/pull/2237/commits/01372bc2588106b4485c032477d5d215ede418bf)

image

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
chris-at-jobber: [looks like it ](https://github.com/GetJobber/atlantis/pull/2237/commits/4c63a98006e68afd764df6d0a16fe8d8691c7d4a)

chris-at-jobber:
@Aiden-Brine great point about the interactions, I moved the styling and interactive state management out of inline styles and into CSS and I think we're in a much happier place now:

- [moved to CSS](https://github.com/GetJobber/atlantis/pull/2237/commits/5b861043a1b3241c177626bf33c3917f264308e5)
- [took out JS interaction handlers](https://github.com/GetJobber/atlantis/pull/2237/commits/2f7aabf1276b760e09eca639a7136d86acb95adc) that are now being handled in the CSS

packages/site/src/layout/NavMenu.tsx
chris-at-jobber: (no thoughts or opinions about this on my end)

Beccaneer: I just updated this, but I was under the understanding we don't mind code that isn't completely perfect or ideal in favour of moving faster

Aiden-Brine (approved):
Thank you for making those changes. Behaviour looks great now

packages/site/src/layout/SearchBox.tsx
Aiden-Brine: @Beccaneer if this is the case I think it is something we should clean up at some point, not worth blocking this PR though

PR #2241, opened 2024-12-11
fix(components): Fixing some issues with the docs site before launch

scotttjob:
## Motivations

Using Jacob's prod branch, I was able to identify two issues we had after going to production:

1. Mobile components were not rendering in production at all.
2. Storybook links were pointing to localhost.

I fixed those issues in that branch, and then cherry picked them out to this PR. We'll merge that one first, merge back into Jacob's PR and then soft launch tomorrow!

## Changes

1. Updated import map references to proper local files instead of `.vite` dependencies that don't exist in production.
1. Updated the name of the local `.env` file from `.local` to `.development`. I thought local meant local environment, but it didn't! Was still loading it production. `.development` means only locally in development mode.

## Testing

Tough to test this because they're all for prod fixes. You can test it doesn't break anything locally though:

1. Loading Mobile + Web components in the live editor should work the same (even with the changes to where jobber hooks is coming from)
2. Storybook links in the right hand nav should still be functional (because even though we renamed `.env.local` to `.env.development` it should still pick up `localhost:6005` locally.

Changes can be

---

Random photo of Atlantis


taylorvnoj (approved):
Looking really good -Storybook links are perfect.

One thing I noticed. I suppose it might be chalked up to buggy behaviour, because the preview doesn't always load at first, but I feel the need to point it out incase you also noticed.

Mobile Form component previews aren't loading.
As I wrote over the mobile screenshot, refreshing on the mobile tab bring me back to the Design Tab (I don't think that's being affected here) but interestingly, `?mobile` remains in the url 🤔 

Screenshot 2024-12-12 at 8 51 24 AM

Screenshot 2024-12-12 at 8 51 42 AM

EDIT: ok I went back on master and saw the same behaviour so I'll approve the PR. Maybe it's a me thing, but if not, good to know so we can dive into it. Btw I didn't go through all the mobile components but others that I checked like `ButtonGroup` previewed as expected

PR #2244, opened 2024-12-12
docs: Fix Border size calculation to avoid appending elements to the DOM

jdeichert:
## Motivations

The existing border calculation was injecting elements into the DOM, and on the new docs site, they hang around until you refresh due to it being a SPA.

This PR replaces that implementation with a simpler one: asking the DOM to compute the value of a CSS variable.

### Changed

- Simplified border width calculation for Border docs page

### Fixed

- Prevent elements from getting injected to the bottom of docs pages and persisting

## Before / After (no changes)

Before |  After
:---:|:---:
Screenshot 2024-12-12 at 7 49 36 AM | Screenshot 2024-12-12 at 7 49 20 AM

## Testing

* Navigate to the Design -> Borders page
* Observe there are no divs injected to the bottom of the DOM
* Observe the table results match what we currently have in [production](https://atlantis.getjobber.com/?path=/docs/design-borders--docs)

### DIFF
--- a/docs/design/Borders.stories.mdx
+++ b/docs/design/Borders.stories.mdx
@@ -34,10 +34,9 @@ export function Example({ of: size }) {
 }
 
 export function BorderSize({ of: size }) {
-  const element = document.createElement("div");
-  Object.assign(element.style, BaseStyle({ of: size }));
-  document.body.appendChild(element);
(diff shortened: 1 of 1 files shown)


scotttjob (approved):
LGTM

PR #2251, opened 2024-12-13
docs: use default image for ContentCard

chris-at-jobber:
## Motivations

Rather than have a broken image symbol if a thumbnail doesn't exist, use the placeholder image when no imageURL is passed in.

I did consider wrapping the image section in a conditional and not rendering if there's no URL passed in, but early feedback was this is a more lightweight approach for now as we don't have to worry about layout shifts etc so addressed that before pushing this up!

### Changed

- ContentCards with no URL provided will return the placeholder image
- Removed explicit declarations of the image URL in `contentList` that were just declaring the placeholder

## Testing

View ContentCard in the Content guides, or try removing an imageURL from any `___List` item that has an actual thumbnail image defined.

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/components/ContentCard.tsx
+++ b/packages/site/src/components/ContentCard.tsx
@@ -19,7 +19,7 @@ export const ContentCard = ({
   title,
   to,
   component,
-  imageURL,
+  imageURL = "../Placeholder.png",
   onClick,
 }: ContentCardProps) => {
   const history = useHistory();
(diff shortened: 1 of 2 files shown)


chris-at-jobber:
Closing as the imageURL is actually required across several places, so this becomes less of a quick win and more of a refactor that doesn't actually help us long run 😅

PR #2255, opened 2024-12-14
docs: Refactor Design item in side nav to use AnimatedPresenceDisclosure

taylorvnoj:
## Motivations

The Design section in the side nav was not utilizing `AnimatedPresenceDisclosure`, so this PR makes that change.

### Added

- Design in the side nav is now a disclosure!

### Changed

- updated the URL from `responsive-breakpoints` to just `breakpoints`. This matches the current url from Storybook (which doesn't have "responsive" in it). I also removed the word responsive from the side nav item. We only have "Breakpoints" in the content cards also. Now everything matches! Less words! Horray!

## Testing

All Design content loads as it should from Home/Search/Disclosure

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/designList.ts
+++ b/packages/site/src/designList.ts
@@ -13,7 +13,7 @@ export const designList = [
   },
   {
     title: "Breakpoints",
-    to: "/design/responsive-breakpoint",
+    to: "/design/breakpoints",
     imageURL: "/Breakpoints.png",
     additionalMatches: ["Responsive"],
   },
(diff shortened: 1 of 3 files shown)


packages/site/src/designList.ts
```diff
@@ -13,7 +13,7 @@ export const designList = [
   },
   {
     title: "Breakpoints",
-    to: "/design/responsive-breakpoint",
```
taylorvnoj, line 16: Doing this to match what's currently in Storybook and the figma (as far as sidenav goes)

Beccaneer (approved):

PR #2258, opened 2024-12-16
docs: Update New Docs Site to have '/' trigger Search

taylorvnoj:
## Motivations

In the new Docs site, we want a quick way to trigger search from any page. We also want it to be clear how to perform a quick search action.

Screenshot 2024-12-16 at 1 08 25 PM

### Added

- you can hit `/` anywhere within the new docs site and trigger the search modal (using `useOnKeyDown` hook to accomplish this)
- apart from the new prefix and suffix, the new HTML button is styled exactly as it was before

### Changed

- No longer using `@jobber/components/Button`; we're using HTML because we need prefix (`search` icon) and suffix (key indicator `/`) within the button 

## Testing

Go to various pages within the new docs site and hit the `/` key. 
Make sure the search button is still working

Changes can be

---

Random photo of Atlantis


packages/site/src/layout/NavMenu.module.css
```diff
@@ -1,34 +1,34 @@
...
+  display: flex;
+  flex-direction: column;
+  padding: 36px 0 0 0;
+  height: 100dvh;
+  width: 200px;
+  box-sizing: border-box;
 }
 
-.navMenuContainer > * {
+.navMenuContainer>* {
```
Aiden-Brine, line 14: The other formatting changes in this file look good but I think this one needs to be reverted

Aiden-Brine: Huh, I wonder why your editor is changing that, it is definitely less readable. You don't need any spaces in CSS so  `.navMenuContainer>*{padding-left: var(--space-base);padding-right: var(--space-base);}` would still work but I think the consensus out there is that `.navMenuContainer > *` is preferable. Anyways, I will approve it anyways as this is the only thing I came across but worth looking into your editor at some point

Aiden-Brine (approved):

packages/site/src/layout/NavMenu.module.css
taylorvnoj: @Aiden-Brine I figured it out! Just pushed a new commit & will merge after. This has been bugging me but thanks for pushing me to find a solution 😅 

Screenshot 2024-12-16 at 5 52 50 PM

Aiden-Brine: Nice! Thanks for doing that

PR #2264, opened 2024-12-18
docs: Add siderail for content pages and improve layouts [Job 111086]

Aiden-Brine:
## Changes

- Adds a `BaseView` component that contains the styles that are shared between `ComponentView` and `ContentView`
- Related to this the page layout on content pages should now be consistent with component pages
- Content pages now have a siderail with anchor tags to jump to parts of the page
- If the siderail has many links (like on changelog pages) the siderail has a scrollbar independent of the main content scrollbar
- Some other cleanup and refactoring

## Testing

- Click around content and component pages and compare the two. The layouts should be matching
- The content pages should now have anchor links in the siderail. That being said as [Taylor recently pointed out](https://getjobber.slack.com/archives/C081WMPQLUV/p1734460480376659) there are issues with many of the anchor links in the site. So testing on this is more about confirming that the links are appearing (on most pages) and when the links are correct the styles and behaviour are as expected. Fixing all of the links will come in a later PR
- Go to a changelog page and ensure the behaviour when there are a huge number of links in the siderail is a good user experience

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/components/AnchorLinks.tsx
+++ b/packages/site/src/components/AnchorLinks.tsx
@@ -1,0 +1,62 @@
+import { Box, Content, Heading } from "@jobber/components";
+import { MouseEvent, useEffect, useState } from "react";
+
+interface AnchorLinksProps {
+  /**
+   * An additional action to perform along with scrolling to the selected anchor
+   */
+  readonly header: string;
+
+  /**
+   * An additional action to perform along with scrolling to the selected anchor
+   */
+  readonly additionalOnClickAction?: () => void;
+}
+
+export function AnchorLinks({
+  header,
+  additionalOnClickAction,
+}: AnchorLinksProps) {
+  const [hlinks, setHlinks] = useState<Element[] | null>(null);
+
+  useEffect(() => {
+    const hdd = document.querySelectorAll("[data-heading-link]");
+
+    if (hdd.length > 0) {
+      setHlinks(Array.from(hdd));
+    }
+  }, []);
+
+  const click = (e: MouseEvent) => {
+    e.preventDefault();
+    const id = e.currentTarget?.getAttribute("href")?.replace("#", "");
+
+    if (id) {
+      additionalOnClickAction?.();
+      setTimeout(() => {
+        const element = document.getElementById(id);
+
+        if (element) {
+          element.scrollIntoView({ behavior: "smooth" });
+        }
+      }, 100);
+    }
+  };
+
+  return (
+    <Content>
+      <Heading level={6} element="h3">
+        {header}
+      </Heading>
+      <Content spacing="small">
+        {hlinks?.map((link, index) => (
+          <Box key={index}>
+            <a onClick={click} href={`#${link.id}`}>
+              {link.textContent}
+            </a>
+          </Box>
+        ))}
+      </Content>
+    </Content>
+  );
+}
--- a/packages/site/src/components/AnchorLinks.tsx
+++ b/packages/site/src/components/AnchorLinks.tsx
@@ -8,6 +8,11 @@ interface AnchorLinksProps {
   readonly header: string;
 
   /**
+   * A key unique to the page that controls re-rendering of the component
+   */
+  readonly key: string;
+
+  /**
    * An additional action to perform along with scrolling to the selected anchor
    */
   readonly additionalOnClickAction?: () => void;
@@ -15,6 +20,7 @@ interface AnchorLinksProps {
 
 export function AnchorLinks({
   header,
+  key,
   additionalOnClickAction,
 }: AnchorLinksProps) {
   const [hlinks, setHlinks] = useState<Element[] | null>(null);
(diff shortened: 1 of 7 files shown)


packages/site/src/components/AnchorLinks.tsx
```diff
@@ -0,0 +1,72 @@
...
+   */
+  readonly key: string;
+
+  /**
+   * An additional action to perform along with scrolling to the selected anchor
+   */
+  readonly additionalOnClickAction?: () => void;
+}
+
+export function AnchorLinks({
```
Aiden-Brine, line 21: This is all functionality that was pulled from `ComponentView`

packages/site/src/components/ComponentLinks.tsx
```diff
@@ -9,130 +9,87 @@ import { useAtlantisSite } from "../providers/AtlantisSiteProvider";
...
-        width: "200px",
-        padding: "36px var(--space-base)",
-        display: "flex",
-        flexDirection: "column",
-        flexShrink: "0",
-        boxSizing: "border-box",
-      }}
-    >
-      <Content spacing={"larger"}>
+    <Content spacing={"larger"}>
```
Aiden-Brine, line 32: The diff looks much more involved than what was actually changed. This was essentially just pulling the anchor link functionality out into its own component

packages/site/src/components/ContentLoader.tsx
```diff
@@ -28,11 +28,5 @@ export const ContentLoader = () => {
 
   const content = contentMap[type][name];
 
-  return (
-    <ContentView
-      intro={content.intro}
```
Aiden-Brine, line 33: intro and title were not being used. We should consider removing these from the mappings unless we plan to use them later?

packages/site/src/layout/BaseView.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+BaseView.Main = function Main({
+  children,
+}: {
+  readonly children: React.ReactNode;
+}) {
+  return (
+    <main
+      style={{
```
Aiden-Brine, line 15: Styles were pulled from `ComponentView` for the most part

packages/site/src/layout/ComponentView.tsx
```diff
@@ -186,55 +186,41 @@ export const ComponentView = () => {
...
-          overflowY: "scroll",
-          backgroundColor: "var(--color-surface)",
-          boxShadow: "var(--shadow-low)",
-          flexGrow: 1,
-        }}
-      >
-        <Box alignItems="center">
-          <div style={{ width: "100%", maxWidth: "768px" }}>
-            <Page width="narrow" title={PageMeta.title}>
-              <PageWrapper>
```
Aiden-Brine, line 201: As far as I can tell `PageWrapper` was providing no benefit here

packages/site/src/layout/ContentView.tsx
```diff
@@ -0,0 +1,25 @@
+import { Content } from "@jobber/components";
```
Aiden-Brine, line 1: I moved `ContentView` into layouts since I think it belongs beside `ComponentView` and `BaseView`

taylorvnoj: Great decision!

packages/site/src/layout/BaseView.tsx
```diff
@@ -0,0 +1,50 @@
+import { Box } from "@jobber/components";
+import React from "react";
+
+export function BaseView({ children }: { readonly children: React.ReactNode }) {
```
taylorvnoj, line 4: Love that you created this component!

This is a nit, not blocking, but we can make use of `PropsWithChildren` to simplify the type definitions. So we can have this instead:

```
export function BaseView({ children }: PropsWithChildren<{}>) {
  return <div style={{ display: "flex", height: "100dvh" }}>{children}</div>;
}
```

It's a utility type that automatically includes `children` as type React.ReactNode. Again, not a blocker, just something I learned a while ago.

packages/site/src/components/ContentLoader.tsx
taylorvnoj: Interesting - I didn't realize we don't need them in the maps. But good call on using title for the tabs

taylorvnoj (approved):
I have one nit but it's not blocking - love the refactor and benefits that come with it (like some hooks pages now have 100% vh)

packages/site/src/components/ContentLoader.tsx
Aiden-Brine: It seems like something we thought we would need but at the moment it doesn't seem like it we are. Something we could get rid of at the end when we are doing other cleanup

packages/site/src/layout/BaseView.tsx
Aiden-Brine: Ah good call, I have seen that in the code but forgot to reach for it when building this. Definitely an improvement and now added

PR #2265, opened 2024-12-18
docs: Make current page apparent in side nav

taylorvnoj:
## Motivations

When navigating through the New Docs site, we want to indicate where the user is, by highlighting the current page in the side nav.

Quick Demo:

https://github.com/user-attachments/assets/3bc7bff6-e06b-4efc-8c63-1b27ed09c3a4

### Added/Changed

1. New `selected` prop in `AnimatedPresenceDisclosure`
The `selected` prop is passed to each child element to determine if it matches the current URL path. This allows us to apply styles indicating the item is selected in the side nav.

2. `selectedRef` in `NavMenu`.
This allows us to perform actions on selected link like scrolling the item into view in the side nav.
We use `isSelected` to determine if the current URL matches `to` and if it does, the link is considered selected. Then, `selectedRef` is assigned to `ref` of the `Link` component.

## Testing

See how the side nav indicates what page you are on by:
- updating the url
- using search to navigate to a page
- clicking content cards in Home/wherever there's a content card
The side nav should scroll the child/page you are on, into view

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/layout/AnimatedPresenceDisclosure.tsx
+++ b/packages/site/src/layout/AnimatedPresenceDisclosure.tsx
@@ -1,5 +1,5 @@
 import { AnimatedPresence, Button, Typography } from "@jobber/components";
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { Link, useLocation } from "react-router-dom";
 import styles from "./NavMenu.module.css";
 
@@ -18,8 +18,13 @@ function AnimatedPresenceDisclosure({
 }: AnimatedPresenceDisclosureProps) {
   const { pathname } = useLocation();
 
+  const childrenArray = useMemo(
+    () => React.Children.toArray(children),
+    [children],
+  );
+
   // Determine if any child is selected based on the current URL
-  const hasSelectedChild = React.Children.toArray(children).some(
+  const hasSelectedChild = childrenArray.some(
     child => React.isValidElement(child) && pathname === child.props.to,
   );
 
@@ -76,7 +81,7 @@ function AnimatedPresenceDisclosure({
       <AnimatedPresence>
         {isOpen && (
           <ul style={{ padding: "0" }}>
-            {React.Children.toArray(children)
+            {childrenArray
               .filter((child): child is React.ReactElement =>
                 React.isValidElement(child),
               )
--- a/packages/site/src/layout/AnimatedPresenceDisclosure.tsx
+++ b/packages/site/src/layout/AnimatedPresenceDisclosure.tsx
@@ -7,12 +7,16 @@ interface AnimatedPresenceDisclosureProps {
   readonly children: React.ReactNode;
   readonly title: React.ReactNode;
   readonly to: string;
+  readonly className?: string;
+  readonly selected?: boolean;
 }
 
 function AnimatedPresenceDisclosure({
   children,
   title,
   to,
+  className,
+  selected,
 }: AnimatedPresenceDisclosureProps) {
   const [isOpen, setIsOpen] = useState(false);
 
@@ -22,8 +26,13 @@ function AnimatedPresenceDisclosure({
   };
 
   return (
-    <div>
-      <span className={styles.disclosureNavItem}>
+    <div className={className}>
+      <span
+        className={`${styles.disclosureNavItem} ${
+          selected ? styles.selected : ""
+        }`}
+      >
+        {" "}
         <Link to={to ?? "/"} tabIndex={0}>
           <Typography fontWeight="semiBold" size="large" textColor="heading">
             {title}
--- a/packages/site/src/layout/AnimatedPresenceDisclosure.tsx
+++ b/packages/site/src/layout/AnimatedPresenceDisclosure.tsx
@@ -51,6 +51,7 @@ function AnimatedPresenceDisclosure({
     setIsOpen(!isOpen);
   };
 
+  // Keeps from having the Disclosure title and the child both highlighted
   const isTitleSelected = location.pathname === to;
 
   return (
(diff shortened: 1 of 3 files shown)


packages/site/src/layout/NavMenu.tsx
```diff
@@ -137,11 +146,21 @@ export const NavMenu = ({ mainContentRef }: NavMenuProps) => {
...
   to,
   children,
-}: PropsWithChildren<{ readonly to: string }>) => {
+  selectedRef,
+}: PropsWithChildren<{
+  readonly to: string;
+  readonly selectedRef: React.RefObject<HTMLAnchorElement>;
+}>) => {
+  const isSelected =
+    location.pathname === to || (location.pathname === "/" && to === "/?new");
```
taylorvnoj, line 155: Needed to add the `to === "/?new"` because Home isn't in a disclosure, but we still need it to be highlighted when selected/the URL matches

Aiden-Brine:
Not related to your changes but why is the area for the scrollbar showing when there aren't enough items for a scroll. Here is a screenshot from the start of the video:


Screenshot 2024-12-19 at 12 56 13 PM

I haven't seen that before on my end

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
```diff
@@ -1,29 +1,66 @@
...
 }
 
 function AnimatedPresenceDisclosure({
   children,
   title,
   to,
+  selected,
 }: AnimatedPresenceDisclosureProps) {
-  const [isOpen, setIsOpen] = useState(false);
+  const location = useLocation();
```
Aiden-Brine, line 19: I should have done this in my changes to `ContentLoader` as well but if we are just using pathname it would better to just explicitly grab the pathname like `const { pathname } = useLocation();`. Not a big deal but the benefit here is that if parts of the location other than pathname change it won't cause a re-render.

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
```diff
@@ -1,29 +1,66 @@
...
+
+  // Determine if any child is selected based on the current URL
+  const hasSelectedChild = React.Children.toArray(children).some(
```
Aiden-Brine, line 22: Another minor improvement. Since we now have `React.Children.toArray(children)` twice in this component and I don't think that value will change we could pull the value into a memoized variable.

packages/site/src/layout/NavMenu.tsx
```diff
@@ -20,6 +20,8 @@ interface NavMenuProps {
 export const NavMenu = ({ mainContentRef }: NavMenuProps) => {
   const [open, setOpen] = useState(false);
   const { isMinimal } = useAtlantisSite();
+  const location = useLocation();
```
Aiden-Brine, line 23: Same as my comment above about grabbing just the pathname

packages/site/src/layout/NavMenu.tsx
Aiden-Brine: I think this is a great time to use a constant for this this value and maybe even a comment about it. Can we grab this value from another file so that when we hard launch we don't have to come back and update here?

packages/site/src/layout/NavMenu.tsx
```diff
@@ -151,11 +170,20 @@ export const StyledLink = ({
...
+  readonly to: string;
+  readonly selectedRef: React.RefObject<HTMLAnchorElement>;
+}>) => {
+  const isSelected = location.pathname === to;
+
   return (
     <Link
       to={to ?? "/"}
-      className={`${styles.navMenuItem} ${styles.navMenuSubItem} ${styles.navMenuLink}`}
+      className={`${styles.navMenuItem} ${styles.navMenuSubItem} ${
```
Aiden-Brine, line 183: I think this one is long enough that it should be build up into a variable to make it a bit easier to read

Aiden-Brine (changes requested):
Functionality looks great! Just a couple of small improvements

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
taylorvnoj: Great call out - adjusted here 7f6e6b9d2648a0bde923bd53e5e89529b48f1d1b

packages/site/src/layout/AnimatedPresenceDisclosure.tsx
taylorvnoj: Another great call out - did that here 867d9da1953700271dc57169bdb780a115717c7c

packages/site/src/layout/NavMenu.tsx
taylorvnoj: Done here 7d5e50a50e005c4b562e60e4202a58b610ab9ed5

packages/site/src/layout/NavMenu.tsx
taylorvnoj: Created a helper function here 7265874edd19e3361206e4ad666fa2a5a81d443f

packages/site/src/layout/NavMenu.tsx
taylorvnoj: To sum up a few conversations surrounding this:
Although I understand how this is a good idea in general, we're actually just going to be getting rid of that code entirely, probably in January. I've added an explanatory/TODO comment above this logic here 0cc72051928439d3de0a6cf7dbcb5152fb1ad2bb

taylorvnoj:
@Aiden-Brine currently looking into the scrollbar separately - `overflow-y: scroll;`is the issue but changing that to `auto` is giving some jumpiness. Either way, I'm ready for another review whenever you have a moment 😄

Aiden-Brine (approved):
Looking great! Thank you for making those changes 🎉

PR #2267, opened 2024-12-18
docs: Add favicon and page title [Job 111944]

Aiden-Brine:
### Added
Adds the favicon and the page title to the tab that updates as you navigate around the docs site. I choose for the home page to say "Atlantis' instead of "Home" but please let me know if you would prefer something else.

## Testing
Navigate from page to page and ensure that the title looks good and is updating as expected

Changes can be

---

Random photo of Atlantis


packages/site/index.html
```diff
@@ -4,7 +4,8 @@
 <head>
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-  <title>Atlantis@next</title>
+  <title>Atlantis</title>
+  <link rel="icon" type="image/png" href="./public/atlantis_favicon.svg" />
```
taylorvnoj, line 8: nit: I'm not sure how much it matters, since the favicon seems to be loading across browsers, but at the bottom of this [wiki](https://en.wikipedia.org/wiki/Favicon) it suggests to use `type="image/svg+xml"` since we're using an `svg` and not a `png`.

packages/site/src/hooks/usePageTitle.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect } from "react";
+
+function usePageTitle({ title }: { title: string }) {
+  useEffect(() => {
+    document.title = title;
```
taylorvnoj, line 5: What do you think about doing something like this here instead:

`document.title = title === "Atlantis" ? title : `${title} - Atlantis`;`

Looking at other sites, they tend to always have their brand name appended (ex [Shopify](https://polaris.shopify.com/components/actions/button), [React](https://react.dev/reference/react-dom/components/form), [Vue](https://vuejs.org/guide/quick-start.html))

taylorvnoj (changes requested):
Looking great - two suggestions - one a nit and the other I think an improvement we can do to match similar documentation sites.  Let me know what you think!

packages/site/index.html
Aiden-Brine: Nice catch! updated

packages/site/src/hooks/usePageTitle.ts
Aiden-Brine: I really like the suggestion. I updated the title on the home page to Home so I don't think think we need the conditional

taylorvnoj (approved):
Thanks for making the changes - this is SO GOOD! 🚢

PR #2270, opened 2024-12-19
feat(components): Add button Navigation Provider

MichaelParadis:
## Motivations

As Atlantis evolves we want to remove the react-router-dom dependency directly used in Atlantis to support other client side routing libraries such as tanstack router. Since the `to` prop has been deprecated Button the logic was added to `ButtonAnchorProps` and the client side navigation logic is tied to a new prop called `useClientSideRouting`

The reason for adding this prop is because a lot of client side routing libraries don't handle urls that navigate to a route not found. For example using `react-router-dom` to navigate to `https://google.com` would not navigate to google.com and will instead try and find the path in the router and navigate to the 404 route for that specific router.

## Decisions made:
 - the client side navigation logic is tied to a new prop called `useClientSideRouting`
    - The reason for adding this prop is because a lot of client side routing libraries don't handle urls that navigate to a route not found. For example using `react-router-dom` to navigate to `<baseURL>/https://google.com` would not navigate to google.com and will instead try and find the path in the router and navigate to the 404 route for that specific router.
  - For typescript support developers need to declare the `NavigationConfig` interface. The reason for this is to provide typescript support when consuming using the provider because regular links in this libraries would have the typehint support. 
  - However due to the age of react-router-dom version 5 the pathnames are not fully typed. 
  - There is also some limitations with tanstack router where the parameter type hints aren't fully enforced

### Added

- components: `ButtonNavigationProvider` to provide support for other client side routing libraries.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

These changes can be tested by reverting `75acaf51c124371a7c237f4fd4076dcb5ac91dcc`
 - Verify that the links navigate correctly. The internal link will navigate to `/components` with url query parameters as well.
See draft PRs that mention this PR

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Button/Button.stories.mdx
+++ b/docs/components/Button/Button.stories.mdx
@@ -331,6 +331,8 @@ to declare the `NavigationConfig` interface in your project.
 
 ```tsx
 import { useHistory } from "react-router-dom";
+import type { RouterNavigationPath } from "@jobber/components"
+
 
 // Both Push and replace have the same parameters
 type RouterActionParameters = Parameters<
@@ -353,7 +355,7 @@ function AppProvider({ children }) {
   const history = useHistory();
   return (
     <ButtonNavigationProvider
-      openLink={(href?: NavigationPath, routerOptions?: NavigateOptions, event) => {
+      openLink={(href?: RouterNavigationPath, routerOptions?: NavigateOptions, event) => {
         // Default to the push action
         const action = routerOptions?.action || "push";
         const historyAction =
@@ -395,6 +397,7 @@ function InnerComponent() {
 ```tsx
 import type { NavigateOptions, ToOptions } from "@tanstack/react-router";
 import { useRouter } from "@tanstack/react-router";
+import type { RouterNavigationPath } from "@jobber/components";
 
 declare module "@jobber/components" {
   interface NavigationConfig {
@@ -408,8 +411,11 @@ function AppProvider({ children }) {
 
   return (
     <ButtonNavigationProvider
-      buildLocationHref={(href: NavigationPath) => href?.toString() || ""}
-      openLink={(href?: NavigationPath, routerOptions?: NavigateOptions) => {
+      buildLocationHref={(href: RouterNavigationPath) => href?.toString() || ""}
+      openLink={(
+        href?: RouterNavigationPath,
+        routerOptions?: NavigateOptions
+      ) => {
         router.navigate({ to: href, ...routerOptions });
       }}
     >
--- a/docs/components/Button/Button.stories.mdx
+++ b/docs/components/Button/Button.stories.mdx
@@ -411,7 +411,10 @@ function AppProvider({ children }) {
 
   return (
     <ButtonNavigationProvider
-      buildLocationHref={(href: RouterNavigationPath) => href?.toString() || ""}
+      buildLocationHref={(href: RouterNavigationPath, routerOptions) => {
+        if (!href) return "";
+        return router.buildLocation({ to: href, ...routerOptions }).href;
+      }}
       openLink={(
         href?: RouterNavigationPath,
         routerOptions?: NavigateOptions
--- a/docs/components/Button/Button.stories.mdx
+++ b/docs/components/Button/Button.stories.mdx
@@ -301,12 +301,140 @@ A way to communicate to the user that a background action is being performed.
   <Button label="Canceling..." variation="subtle" loading={true} />
 </Canvas>
 
-## Client-side routing (Web only)
+### DEPRECATED: Client-side routing (Web only)
 
 A Button can be used for client-side routing as well. When using a Button to
 handle the client-side routing, use the `to` prop. Notice when you click below
 that the URL will change to the appropriate route. See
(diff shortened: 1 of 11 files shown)


docs/components/Button/Button.stories.mdx
```diff
@@ -301,13 +301,150 @@ A way to communicate to the user that a background action is being performed.
...
+  );
+}
+
+function InnerComponent() {
+  return (
+    <Button
+      to="/some-path/$someParam"
+      routerOptions={{
+        resetScroll: true,
+        params: { someParam: "someValue" },
```
scotttjob, line 436: How are params/search/state different? They're all just URL Query Params in the end right? Or am I missing something?

packages/components/src/Button/ButtonNavigationProvider.tsx
```diff
@@ -0,0 +1,73 @@
...
+  createContext,
+  useContext,
+  useMemo,
+} from "react";
+
+export interface NavigationConfig {}
+export type RouterOptions = NavigationConfig extends { routerOptions: infer T }
+  ? T
+  : // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    any;
```
scotttjob, line 14: 👀 I have a feeling this 'any' is actually an 'any' but if that's the case can we leave that comment at the end of the eslint disable line explaining that? Just so nobody comes back in the future and spends an afternoon trying to type it 'properly'

packages/components/src/Button/useLinkButtonProps.ts
```diff
@@ -0,0 +1,32 @@
...
+) {
+  const { routerOptions, onClick, url, useClientSideRouting } = props;
+  const { openLink, buildLocationHref } = useButtonNavigationProvider();
+  const routerHref = buildLocationHref(url, routerOptions);
+
+  const handleClick = useCallback(
+    (event: React.MouseEvent<HTMLAnchorElement | HTMLButtonElement>) => {
+      useClientSideRouting && event.preventDefault();
+
+      onClick?.(event);
```
scotttjob, line 16: I like that we call 'onClick' even when handling the url. I think we had a restriction in place before that said you can't provide an onClick if you also provide a URL? Was I mistaken, or did we remove that restriction (happy if we removed it, I never liked it).

packages/site/src/content/Button/ButtonNotes.mdx
```diff
@@ -1,12 +1,148 @@
...
+          search: "?my-query-param=some-query-param",
+        }}
+        routerOptions={{ action: "replace", { myHistoryStateValue: true}}}>
+      Go to some path
+    </Button>
+  );
+}
+```
+
+#### Tanstack router example:
```
scotttjob, line 93: Love that you supported both routers in your examples. Great work!

scotttjob (commented):
I've only read through the code so far, but I wanted to drop my comments on my code read as soon as possible (which is why this only a comment).

I'd like to get a chance to run the code and try it out first hand before approving, but it's okay if someone else gets here first.

What's in here looks great, I really appreciate your approach to backwards compatibility and providing paths forward for both of our router types.

Hopefully I'll get a chance to run the code today.

docs/components/Button/Button.stories.mdx
MichaelParadis: Search is URL query params tanstack [docs](https://tanstack.com/router/latest/docs/framework/react/guide/search-params)
Params is [path params](https://tanstack.com/router/latest/docs/framework/react/guide/path-params) in the url. Similar to how on github has the URL for this PR: `https://github.com/GetJobber/atlantis/pull/2270/files` which could be formatted as such:
`https://github.com/<repoOwner>/<repoName>/pull/<pullRequestID>/files`

State is the history state handed by the router. It is from the `ToOptions` from [tanstack](https://tanstack.com/router/latest/docs/framework/react/api/router/ToOptionsType) router. It is just an arbitrary key value [store](https://tanstack.com/router/latest/docs/framework/react/api/router/historyStateInterface#historystate-interface)

packages/components/src/Button/ButtonNavigationProvider.tsx
```diff
@@ -0,0 +1,73 @@
...
+   * Function that will be called when a user clicks on a Button with a URL and useClientSideRouting is true.
+   */
+  readonly openLink: (
+    buttonURLProp: RouterNavigationPath,
+    routerOptions?: RouterOptions,
+    event?: React.MouseEvent<HTMLAnchorElement>,
+  ) => void;
+
+  /**
+   * Function that will build the href used by the anchor tag in the Button component based on the url prop of the Button
```
scotttjob, line 31: This description + the function name make the functionality of this a bit vague. Maybe just an example of the expected input -> output? Definitely a follow-up at this point.

scotttjob (approved):
I dropped you a question in a DM, but otherwise this looks great!

As a follow-up, I wonder if it's worth documenting our approach to the Interface/Provider pattern here, so it can inform the work when we approach soon in the future (when we do the same work with Apollo and any other libraries we would rather depend on, versus consume directly)

PR #2272, opened 2024-12-20
docs: Hide NavMenu scrollbar when it's not needed

taylorvnoj:
## Motivations

When the NavMenu was fully collapsed and there was no need for scrolling, a empty scrollbar was visible (especially on light mode).
We need to hide the scrollbar until it's necessary because a large disclosure is open, etc.

### Before

Screenshot 2024-12-20 at 2 54 48 PM

### After

Screenshot 2024-12-20 at 2 54 59 PM

## Testing

When there are not enough items in the NavMenu to warrant a scrollbar, make sure you don't see one.
Open Components and you should see a scrollbar appear and it works.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/layout/NavMenu.module.css
+++ b/packages/site/src/layout/NavMenu.module.css
@@ -28,7 +28,8 @@
   display: flex;
   flex-direction: column;
   gap: var(--space-large);
-  overflow-y: scroll;
+  overflow-y: auto;
+  scrollbar-gutter: stable;
 }
 
 .navMenu li {


packages/site/src/layout/NavMenu.module.css
```diff
@@ -28,7 +28,8 @@
   display: flex;
   flex-direction: column;
   gap: var(--space-large);
-  overflow-y: scroll;
+  overflow-y: auto;
+  scrollbar-gutter: stable;
```
taylorvnoj, line 32: `overflow-y: auto;` alone makes the NavMenu really jumpy when opening and closing long disclosures. The Search input was jumping in size also.
`scrollbar-gutter: stable;` ensures space is always reserved for the scrollbar, even when it’s not visible.

Beccaneer (approved):

PR #2273, opened 2024-12-20
docs: Adjust margin on Disclosure Nav Item

taylorvnoj:
## Motivations

When title of the disclosure was in focus and you hovered on the first child, there was visual overlap.

### Before

Screenshot 2024-12-20 at 5 23 21 PM

### After

Screenshot 2024-12-20 at 5 24 07 PM

## Notes

1. I did want to adjust the animation to be a bit quicker but that started to seem a little convoluted so I'm deciding not to attempt that right now
2. This does add a tiny bit more space between title items in the nav, as seen in the gif below. I do think it's worth the tradeoff for not having overlapping focus/hover states

Kapture 2024-12-20 at 17 30 42

Changes can be

---

Random photo of Atlantis


packages/site/src/layout/NavMenu.module.css
```diff
@@ -117,6 +117,7 @@
   transition: all var(--timing-base) ease-out;
   margin-left: var(--negativeHoverOffset);
   margin-right: var(--negativeHoverOffset);
+  margin-bottom: var(--space-smaller);
```
taylorvnoj, line 120: Disclaimer: I'm not a designer 😆  But using this token is my personal preference.
Chris did say `--space-smallest` could probably do the trick but I still see overlap when the button is hovered so that's why I went with `smaller`. If anyone other than Chris feels icky reviewing this I totally understand and we can leave it till he's back.

Screenshot 2024-12-20 at 5 33 34 PM

scotttjob (approved):

packages/site/src/layout/NavMenu.module.css
chris-at-jobber: Not sure if it's worth blocking this PR on, but this change does exacerbate an issue with the sidenav.

Because we are making the items responsible for their own spacing and only updating the Disclosure version here, what was once a 2px discrepancy between "normal" vs "disclosure" headings is now a 6px discrepancy:
image

chris-at-jobber: Looking at the structure, I think it's because while the top-level nav items are all wrapped in a Content, the _sections_ are just divs with no styling, which means the "responsibility" for the spacing is being handed off halfway down the tree, which allows the spacing to only be increased for the disclosure items. 

This is a Content:
image

These are the individual sections:
image

taylorvnoj: The individual sections are also wrapped in a `Content` (just one more div up in your screenshot):
Screenshot 2025-01-08 at 9 24 06 PM

So if I remove that `Content` which is in `NavMenu.tsx` and put the `margin-bottom` back on the disclosure, we fix the issue this PR addresses while keeping the distance between sections the same.

With this commit 9808aad9888a73c518c38f999b2faffd3700c27a, we now have:

`--space-larger` height:
Screenshot 2025-01-08 at 11 00 57 PM

and overlap issue fixed:
Screenshot 2025-01-08 at 11 10 38 PM

But it's possible I'm missing something so definitely want your eye on that again @chris-at-jobber whenever possible 😄

chris-at-jobber (approved):
Looks great, thank you @taylorvnoj 🫡

PR #2274, opened 2024-12-20
docs: Add Not Found Page [JOB-111818]

Aiden-Brine:
I am opening this PR because @Beccaneer  is gone for the holidays but full credit for this lovely 404 page goes to her. I just did some routing touch ups at the end.

### Added

A new 404 page when a page isn't found on the docs site.

### Additional context
Here is an explanation of the routing changes I made.

https://github.com/user-attachments/assets/653d6f81-2455-493d-8d56-3737a413b445

## Testing

Make sure all the routing is working as expected. Make sure the 404 page is rendering when you hit an endpoint that doesn't exist.

Changes can be

---

Random photo of Atlantis


packages/site/src/routes.tsx
```diff
@@ -11,6 +11,7 @@ import { HooksPage } from "./pages/HooksPage";
 import { hooksList } from "./hooksList";
 import { GuidesPage } from "./pages/GuidesPage";
 import { PackagesPage } from "./pages/PackagesPage";
+import { ComponentNotFound } from "./components/ComponentNotFound";
```
scotttjob, line 14: This content was created for us to help us understand why a component wasn't showing up after we transferred it into the system.

So while I appreciate us keeping the content as a specific route, I don't think everyone will ever visit it directly (since before this it was acting as a 404 page).

Would it make sense to move this content into a new "Working with the Docs site" section of our docs?

packages/site/src/pages/NotFoundPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+      </div>
+    );
+  };
+
+  const [fishElement, setFishElement] = useState<JSX.Element | null>(
+    getRandomFish(),
+  );
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
+
+  useEffect(() => {
```
scotttjob, line 29: This is a smell to me, with the useEffect and two timeouts + tracking one of them for clearing it. I know this is just a 404 page, so really low-risk overall, but what is this useEffect accomplishing? I see the comment below with the setTimeout and 0 as the delay, in that case however a `requestAnimationFrame` would be the better solution, as the code will be called on the next tick like you're wanting.

Is the idea to select a random fish to display?

packages/site/src/pages/NotFoundPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <div className={styles.fish}>
+        {sealifeArray[Math.floor(Math.random() * sealifeArray.length)]}
+      </div>
+    );
+  };
+
+  const [fishElement, setFishElement] = useState<JSX.Element | null>(
+    getRandomFish(),
+  );
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
```
scotttjob, line 27: This is the wrong type. It's not a NodeJS Timeout, but just a Number. The NodeJS Timeout would be true if this was running on the server.

scotttjob (changes requested):
I'd love for us to revisit the useEffect + timeout solution, that's a bit of a code smell to me. If there's a specific reason for it, that's okay, but I'd like us to document that specific reason + why another solution didn't work (just as a JSDoc comment above the bit of code).

Also I'd like to get the NodeJS.Timeout type out, it doesn't belong.

Otherwise my comments are just comments, and I'll be able to approve after we sort out those two small issues.

packages/site/src/pages/NotFoundPage.tsx
Beccaneer: This is my newness to React showing through a bit :) I will give that a try!

packages/site/src/routes.tsx
Beccaneer: @taylorvnoj  mentioned concerns about not having this content included in some way

packages/site/src/pages/NotFoundPage.module.css
```diff
@@ -0,0 +1,77 @@
...
+
+  @media screen and (min-width: 1024px) {
+    .container {
+        align-items: center;
+        background: var(--color-informative--onSurface);
+        background: linear-gradient(to bottom, var(--color-informative--onSurface) 0%, var(--color-surface) 40% );
+    
+        /* The Wavy part */
+        --size: 50px;
+        --R: calc(var(--size)*1.28);
```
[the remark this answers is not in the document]
Beccaneer: Not generated! I frankensteined this together a bit from various open sources code pieces, I will give that a look!

chris-at-jobber:
A few suggestions:

- It would be nice to get the text content center-aligned
- should use our `Heading level={"1"}` for the h1
- I think it could be helpful to give an explicit "go home" action to give the user an explicit remediation
- Also tried some ideas for simplifying the copy a bit while still keeping the fun nautical vibes 

image
image
image

packages/site/src/routes.tsx
taylorvnoj: @scotttjob I think that content might still be valuable to keep as a link on the 404 pages for any component not found path, because it will give good hints to future devs as to why a new component might not be showing up on the site.
I'm gonna play devils advocate and say, no one will seek out a different page in our docs to figure out why they're running into issues. It's likely in both cases they'll come ask us. 
I'm of the thought that a link is the most intuitive - unless we put that info on the 404 page itself when a component isn't found, if we want to avoid creating another route.

scotttjob: Great use of devil's advocate Taylor, I completely agree. I'll sheepishly admit I didn't notice the bit of copy referencing the page and directing us there, so I thought it had become disconnected.

Really actually like this solution of keeping it a link with information, so great work all around.

scotttjob (approved):
Changes look good to me on the code front! Appreciated.

Chris left some design feedback, but we might be able to corral that into a new ticket and address it soon after the soft launch?

PR #2275, opened 2024-12-20
docs: Refactor SearchBox

taylorvnoj:
## Motivations

`SearchBox` can use some refactoring and DRYing up for better maintainability and readability.

**Note:** I snuck in two other small changes (fixed a typo and updated the favicon path), pointed out below in comments

### Changed

- Consolidated multiple useMemo instances filtering different lists into a single filteredLists array using a reusable filterItems helper function.
- Replaced the repeated list filtering logic for individual lists with a unified lists array of objects containing title and items.
- Moved the "empty results" UI into a dedicated EmptyResults component for more clarity 

## Testing

All behaviour in Search from before, should be the same after this refactor.

---

Random photo of Atlantis


packages/site/src/layout/SearchBox.tsx
```diff
@@ -24,13 +21,20 @@ import { hooksList } from "../hooksList";
 import { packagesList } from "../packagesList";
 import { ContentListItem } from "../types/components";
 
+const lists = [
```
taylorvnoj, line 24: `lists` is static; keeping it outside the function helps avoids unnecessary re-evaluation on every render

packages/site/src/layout/SearchBox.tsx
```diff
@@ -45,7 +49,7 @@ export const SearchBox = ({
 
   const filterFunction = (item: ContentListItem) =>
     item.title.toLowerCase().includes(search.toLowerCase()) ||
-    item.additionalMatches?.find((e: string) =>
+    item.additionalMatches?.some((e: string) =>
```
taylorvnoj, line 52: I feel like `.some` is all we need here. `.some` is just looking for if any `additionalMatches` match the search term and that's all we care about.
`.find` is going to return the matching element which we don't need in this case.

packages/site/src/layout/SearchBox.tsx
```diff
@@ -43,53 +47,31 @@ export const SearchBox = ({
...
-    item.title.toLowerCase().includes(search.toLowerCase()) ||
-    item.additionalMatches?.find((e: string) =>
-      e.toLowerCase().includes(search.toLowerCase()),
+  const filterItems = (items: ContentListItem[], searchTerm: string) => {
+    const lowerCaseSearchTerm = searchTerm.toLowerCase();
+
+    return items.filter(
+      item =>
+        item.title.toLowerCase().includes(lowerCaseSearchTerm) ||
+        item.additionalMatches?.some((e: string) =>
```
taylorvnoj, line 56: I feel like `.some` is all we need here. `.some` is just looking for if any `additionalMatches` match the search term and that's all we care about.
`.find` is going to return the matching element which we don't need in this case.

packages/site/index.html
```diff
@@ -5,7 +5,7 @@
   <meta charset="UTF-8" />
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   <title>Atlantis</title>
-  <link rel="icon" type="image/svg+xml" href="./public/atlantis_favicon.svg" />
+  <link rel="icon" type="image/svg+xml" href="/atlantis_favicon.svg" />
```
taylorvnoj, line 8: Removed the `./public` to remove this warning in the terminal

Screenshot 2024-12-24 at 8 38 54 PM

docs/hooks/UseBool.tsx
```diff
@@ -13,7 +13,7 @@ export function UseBool() {
       <br />
       <Button onClick={on} label={"On"} variation="subtle" />
       <br />
-      <Button onClick={off} label={"Off"} variabtion="destructive" />
+      <Button onClick={off} label={"Off"} variation="destructive" />
```
taylorvnoj, line 16: Quick typo fix!

scotttjob (approved):
Huge Thumbs Up!

I saw this file myself during my refactors and thought "I'll clean that up later"

Not realizing you had already done that.

Great work!

Also, this is a great example of building things "quickly" to start with, until the patterns start to emerge that need to be DRY'd out.

Only other comment I have, is that EmptyResults feels out of place, but I don't have a better solution. Its own file is too much, it's not really a re-usable component. We have a variety of other small components like this, tucked into other files, but maybe that's okay!

PR #2276, opened 2024-12-22
docs: Reset Tab and Props State on Component Change

taylorvnoj:
## Motivations

Fixing a few issues with state when switching between components (plus a couple small things I noticed along the way)

https://github.com/user-attachments/assets/0fccecf8-454e-440c-89bd-3edc361edcc5

### Fixed

- Props list now updates depending on what component page you're on
- When switching from component to component, tabs will revert back to Design 

## Testing

Make sure props lists update to reflect the component you are viewing

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #2277, opened 2024-12-23
docs: Add id to AnchorLinks and Add keys to Switch in Layout

taylorvnoj:
## Motivations

When viewing the site locally, the following errors were in the console:

Screenshot 2024-12-23 at 5 07 44 PM

### Changed

`ComponentLinks` which is using `AnchorLinks` had a `key` prop - however we should be avoiding calling a prop `key` to avoid confusion with React's `key` attribute.  That is why I renamed the prop to `id`.

`Layout` needed keys so I've added those to the previously empty fragments. The `React.Fragment` is grouping the dynamic routes under the Switch and adding keys makes React happy (we're aligning more with React's preference for list rendering).

## Testing

All side rail links should work as expected.
All component pages and sections render as expected.
---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/components/AnchorLinks.tsx
+++ b/packages/site/src/components/AnchorLinks.tsx
@@ -8,9 +8,9 @@ interface AnchorLinksProps {
   readonly header: string;
 
   /**
-   * A key unique to the page that controls re-rendering of the component
+   * A unique identifier for the component
    */
-  readonly key: string;
+  readonly id: string;
 
   /**
    * An additional action to perform along with scrolling to the selected anchor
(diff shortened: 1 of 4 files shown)


Aiden-Brine (approved):
Thank you very much for clearing up those errors! The changes to the Switch will likely be refactored out by [the changes I made to get Rebeca's 404 page working](https://github.com/GetJobber/atlantis/pull/2274/files) but I am happy to see this go in for now 🎉

taylorvnoj:
This PR actually fixed issues laid out in https://github.com/GetJobber/atlantis/pull/2276

What I unintentionally fixed here is:
- When switching from component to component, tabs will revert back to Design. This is important because if you are on a component and on an Implement Tab, then navigated to a component without an Implement tab, you saw a blank component
AND
- navigating from component to component, you'd often see the props tables **not** update

PR #2284, opened 2025-01-07
docs: give preview window contents more room to breathe [JOB-112177]

chris-at-jobber:
## Motivations

- Users reported that the live preview area was hard to distinguish from surrounding content
- Users reported content feeling squished / not enough room, especially for hooks or components with UI that spans a wide area
- Hard to tell which components have a defined background color when the preview window is the same color as most components that have a `surface`

### Changed

- Preview window background to `surface--background--subtle`
- Component preview background to transparent
  - Note that mobile components will look worse in dark mode because they don't have a dark mode yet, but I think that's fine, because it just highlights a gap in functionality
- reduced padding of preview window

image
image
image

## Testing

Run docs site locally

---

Random photo of Atlantis


packages/site/src/providers/AtlantisPreviewEditorProvider.tsx
```diff
@@ -360,23 +360,24 @@ const skeletonHTML = (theme: Theme, type: "web" | "mobile") => {
...
-  border-radius:10px;
+  border-radius: var(--radius-base);
 }
 </style>
 <link rel="stylesheet" href="/styles.css">
 <link rel="stylesheet" href="/foundation.css">
 <link rel="stylesheet" href="/dark.mode.css">
 </head>
-      <body>
+      <body style="background-color: transparent;">
```
chris-at-jobber, line 380: Need to define this here to overwrite the declaration from `foundation.css`

packages/site/src/providers/AtlantisPreviewEditorProvider.tsx
```diff
@@ -360,23 +360,24 @@ const skeletonHTML = (theme: Theme, type: "web" | "mobile") => {
...
-  height: 100%;
-  width: 99%;
+  height: 100vh;
+  width: 100vw;
   display: flex;
   align-items: center;
   justify-content: center;
   flex-direction: column;
-  min-height: 200px;
+  min-height: 280px;
```
chris-at-jobber, line 369: Now that the padding has gone from 48 down to 8, we have 80px more of vertical space to work with for free

packages/site/src/providers/AtlantisPreviewEditorProvider.tsx
```diff
@@ -360,23 +360,24 @@ const skeletonHTML = (theme: Theme, type: "web" | "mobile") => {
...
+  margin: 0;
```
chris-at-jobber, line 370: This, combined with the vh/vw above, let the contents use the full defined space as needed

packages/site/src/providers/AtlantisPreviewEditorProvider.tsx
```diff
@@ -360,23 +360,25 @@ const skeletonHTML = (theme: Theme, type: "web" | "mobile") => {
...
+  padding: var(--space-smallest);
```
chris-at-jobber, line 371: Minus this small space needed to prevent the very edge of things like cards from being clipped

packages/site/src/providers/AtlantisPreviewEditorProvider.tsx
```diff
@@ -360,23 +360,28 @@ const skeletonHTML = (theme: Theme, type: "web" | "mobile") => {
...
+  min-height: 280px;
+  margin: 0;
 }
   html[data-type="mobile"] {
-  border-radius:10px;
+  border-radius: var(--radius-base);
+}
+#root {
+  box-sizing: border-box;
+  padding: var(--space-smallest);
```
chris-at-jobber, line 377: This insets the wrapper of the component by _just_ a pinch to prevent the last/first pixels from being clipped by the parent iframe.

Doing this here instead of on the `body, html, #root` selector above because it had slightly unpredictable results, dependent on the components' own box-sizing/layout. For example, Card was fine, but DataTable was being clipped.

This approach is more specific to where we actually require this styling.

image
image
image

scotttjob (approved):
LGTM! Thanks Chris, great work and this really does look better.

PR #2288, opened 2025-01-08
fix(docx): Fixing issue with Minimal Mode on small Viewports

scotttjob:
## Motivations

1. As part of the vscode plugin work, we noticed that minimal mode had an issue with some sidenav changes.
2. This PR fixes that so minimal mode looks good again on small viewports.

## Changes

1. Added  a `visible` prop to BaseView.SideRail, which is determined by `isMinimal` from the provider.

### Added

- A Visible prop to BaseView.SideRail

## Testing

1. Adding ?minimal to the URL should hide the navigation + sidenav on the component pages.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/layout/BaseView.tsx
+++ b/packages/site/src/layout/BaseView.tsx
@@ -28,9 +28,15 @@ BaseView.Main = function Main({
 
 BaseView.Siderail = function Siderail({
   children,
+  visible = true,
 }: {
   readonly children: React.ReactNode;
+  readonly visible: boolean;
 }) {
+  if (!visible) {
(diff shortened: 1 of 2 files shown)


chris-at-jobber:
@scotttjob as a followup to this, or maybe just for this work, can we just do the [responsive implementation where the siderail goes away on smaller breakpoints?](https://www.figma.com/design/HhNrPoreO9D8R8O9Ta2AM7/Unified-Atlantis-Docs?node-id=1250-20308&t=n7VMofGTevmzZGwY-1)

PR #2289, opened 2025-01-08
docs: update layout component examples

chris-at-jobber:
## Motivations

Felt like some of our examples for layout components could more effectively convey what the component is/does. The examples in the site have a lot of "single column" examples for things like Grid, DataList (which isn't layout but has a similar issue because it uses Grid), and Content has a really nice "practical" example but it's not immediately clear what the Content component actually is because it's showing off a nicely spaced form.

### Added

- 

### Changed

- Component examples:

**Grid**: visually highlight the 12 column structure
image

**DataList**: show the more prevalent "non-stacked" layout
image

**Box**: highlight color and radius as additional features to better suggest the customization capabilities
image

**Card**: use the built-in header and action features
image

**Content:** highlight spacing examples
image

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


packages/site/src/content/Content/index.tsx
```diff
@@ -9,30 +9,37 @@ export default {
...
-          </Text>
-          <InputText placeholder="Name" />
-          <InputText placeholder="Phone" />
-          <InputText placeholder="Email" />
-          <InputText
-            multiline={true}
-            placeholder="Describe yourself"
-            name="describeAge"
-          />
+    element: `return (
```
taylorvnoj, line 12: Considering this fits better in the iframe and still conveys what the component does, I'm good with this change.  I do think having the old example remain in storybook/become a future example in New Docs is a good idea so we can continue sharing patterns - but that's not something we should address here.

taylorvnoj (approved):
LGTM - get's me excited to eventually get more examples in new docs but the changes you made add more polish to the iframes. Great update!

PR #2291, opened 2025-01-08
docs: fix h1s across the docs site [JOB-113010]

chris-at-jobber:
## Motivations

After a recent change to our index file, Jobber Pro was no longer being pulled in from the CDN.

Also, the h1s on our collection pages were set up using Typography to achieve a "white" version, but using the wrong weight of Jobber Pro. This is a great use case for the AtlantisThemeContextProvider.

### Fixed

- Jobber Pro is back
- Collection and detail pages h1s have consistent font-weights

image
image

## Testing

Run docs site locally/Cloudflare
- check collection page (ie Home, Components)
- check detail page (ie Button)

If you can't visually parse the difference between Jobber Pro and Poppins, remove Poppins from the fallback stack in your inspector and observe no change to the rendered font.

---

Random photo of Atlantis

### DIFF
--- a/packages/site/index.html
+++ b/packages/site/index.html
@@ -10,7 +10,7 @@
   <title>Atlantis</title>
   <link rel="icon" type="image/svg+xml" href="/atlantis_favicon.svg" />
   <link rel="preconnect" href="https://cdn.jobber.com">
-  <link rel="prefetch" href="https://cdn.jobber.com/fonts/fonts.css" as="style">
+  <link rel="stylesheet" href="https://cdn.jobber.com/fonts/fonts.css">
   <link rel="prefetch" href="/styles.css" as="style">
   <link rel="prefetch" href="/foundation.css" as="style">
   <link rel="prefetch" href="/dark.mode.css" as="style">
--- a/packages/site/src/components/HeaderBlock.tsx
+++ b/packages/site/src/components/HeaderBlock.tsx
@@ -1,4 +1,10 @@
-import { Box, Button, Content, Typography } from "@jobber/components";
+import {
+  AtlantisThemeContextProvider,
+  Box,
+  Button,
+  Content,
+  Heading,
+} from "@jobber/components";
 import { useHistory } from "react-router";
 
 interface HeaderBlockProps {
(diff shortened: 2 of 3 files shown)


packages/site/src/components/HeaderBlock.tsx
```diff
@@ -33,17 +39,16 @@ export const HeaderBlock = ({
...
-            size="extravagant"
-            fontWeight="bold"
-            fontFamily="display"
-          >
-            {title}
-          </Typography>
-          <Typography size="large" fontWeight={"semiBold"}>
-            {body}
-          </Typography>
+          <div>
```
chris-at-jobber, line 42: Slightly annoying detail for anyone interested in improving `AtlantisThemeContextProvider` - requires an additional wrapper to be placed outside of itself for `Content` to work, because the `margin` applied by Content to its direct children is subsumed by the `AtlantisThemeContextProvider`s `display: contents`

chris-at-jobber: (Still fewer LOC than the Typography solution, though 😄)

taylorvnoj: We should definitely document this somewhere. Maybe bring it in the channel for thoughts first

packages/site/index.html
```diff
@@ -10,7 +10,7 @@
   <title>Atlantis</title>
   <link rel="icon" type="image/svg+xml" href="/atlantis_favicon.svg" />
   <link rel="preconnect" href="https://cdn.jobber.com">
-  <link rel="prefetch" href="https://cdn.jobber.com/fonts/fonts.css" as="style">
+  <link rel="stylesheet" href="https://cdn.jobber.com/fonts/fonts.css">
```
taylorvnoj, line 13: I can understand why the prefetch change was made. I did some light reading and seems like `preload` could also be used for fonts but that didn't work when I tried it.
The direction in our internal docs is to have this `<link>` as you currently have it with the `stylesheet`. So if it's good enough for everyone, it's good enough for me too.

taylorvnoj (approved):
Jobber Pro is **so** back.
Everything LGTM 🚀

scotttjob (approved):

PR #2298, opened 2025-01-09
docs: Add section about UNSAFE_ props to Popover documentation

taylorvnoj:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Popover/Popover.stories.mdx
+++ b/docs/components/Popover/Popover.stories.mdx
@@ -29,6 +29,10 @@ See
 
 ### Functional
 
+<Banner type="warning" dismissible={false}>
+  A "functional" Popover hasn't been implemented in a componentized fashion yet.
+</Banner>
+
 Reveal a list of available actions to the user. For example, if the user clicks
 on an element and there are four potential actions they might take, Popover is a
 great way to present those actions. If you're looking to provide a menu of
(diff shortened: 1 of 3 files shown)


docs/components/Popover/Popover.stories.mdx
```diff
@@ -29,18 +29,76 @@ See
 
 ### Functional
 
+<Banner type="warning" dismissible={false}>
```
taylorvnoj, line 32: Just moved this to a spot that made more sense

docs/components/Popover/Popover.stories.mdx
```diff
@@ -29,18 +29,76 @@ See
...
 on an element and there are four potential actions they might take, Popover is a
 great way to present those actions. If you're looking to provide a menu of
 actions that comes complete with a trigger button,
 [Menu](../?path=/docs/components-navigation-menu--docs) has that bundle ready to
 go.
 
-<Banner type="warning" dismissible={false}>
-  A "functional" Popover hasn't been implemented in a componentized fashion yet.
-</Banner>
+### UNSAFE\_ props (Advanced Usage)
```
taylorvnoj, line 43: Used (Advanced Usage) to copy what is happening in List Docs where they use `[Custom Rendering (Advanced Usage)](https://atlantis.getjobber.com/?path=/docs/components-lists-and-tables-list--docs#custom-rendering-advanced-usage)`

docs/components/Popover/Popover.stories.mdx
```diff
@@ -29,18 +29,76 @@ See
...
+```tsx
+  UNSAFE_style={{
+    container: { backgroundColor: "lightblue" },
+    dismissButtonContainer: { paddingRight: "var(--space-larger)" },
+    arrow: { paddingRight: "var(--space-extravagant)" },
+  }}
+```
 
-### Related components
+## Related components
```
chris-at-jobber, line 101: ❤️

docs/components/Popover/Popover.stories.mdx
chris-at-jobber: Feels like this should be an h2 under something like "Component customization", until such time as it can live under its' own tab in the new docs site, rather than nested under design guidance?

Also nit, but we should be [using sentence case](https://atlantis.getjobber.com/content/formatting) (ie `(advanced usage)` instead of `(Advanced Usage)`)

taylorvnoj: Good idea - because then we see "Component customization" in the side rail. Ok I updated that here 99c7f556f63b765aff9934ecb9d8bae4e8b070c9
Added that to `List` docs as well. There will be a time where we have "Custom rendering" and "UNSAFE_ props", both of which should go under "Component customization"

Screenshot 2025-01-09 at 9 02 33 PM

chris-at-jobber: :noice:

docs/components/Popover/Popover.stories.mdx
```diff
@@ -29,18 +29,78 @@ See
...
+  background-color: var(--color-indigo);
+}
+
+.customArrow {
+  padding-left: var(--space-extravagant);
+}
+```
+
+You can also use `UNSAFE_className` to apply custom classes to any of the
+Popover's elements.
```
scotttjob, line 85: Incredibly pedantic, and I might be wrong, but I feel like this '`s' on the end of Popover makes the sentence read poorly? Or is it the 'the'?

packages/site/src/content/Popover/PopoverNotes.mdx
```diff
@@ -0,0 +1,59 @@
+## Component customization
```
scotttjob, line 1: Is this information duplicated from above? If so (for the future, not for this ticket) I wonder if there's a way we can make a 'component' out of certain pieces of content then just consume them in these MDX files like other react components? A way to put content in two places without duplicating.

scotttjob (approved):
Couple of comments, but not enough to slow this train down!

docs/components/Popover/Popover.stories.mdx
taylorvnoj: You're right and I can't let it go. How about
"You can also use `UNSAFE_className` to apply custom classes to any of the
Popover's elements."
Fixed that here 194e1d8f65fc118066d5d715141586c86567dddb

packages/site/src/content/Popover/PopoverNotes.mdx
taylorvnoj: Yes, to have it on the docs page for soft launch but I've started a clean up PR where all the duplication will be removed here https://github.com/GetJobber/atlantis/pull/2308

Beyond that I don't think we'll see such a component, since we won't have to put information in 2 places that I can foresee?

PR #2301, opened 2025-01-10
fix(docx): Final Feedback before soft launch

scotttjob:
## Motivations

1. Had some feedback to bring better organization to the hook pages, which is done in this PR.
2. Wanted to double check rendering of all components, made adjustments for three mobile components which are now rendering properly for the first time (yay). This now means all the mobile components are at least rendering in the browser (Form, FormField, Glimmer)
3. Added disclaimer to all mobile component views, with the same message as Storybook: 'Due to distinctions between web and native platform, this may not render accurately in a web browser.'
image

## Testing

1. View the hooks page. Does it have categories?
1. View the Form/FormField/Glimmer mobile components. Do they render?
1. Do you see the message listed above on every "Mobile" tab under Components?

Changes can be

---

Random photo of Atlantis


packages/site/src/layout/ComponentView.tsx
```diff
@@ -109,6 +109,12 @@ export const ComponentView = () => {
           <Box margin={{ bottom: "base" }}>
             <AtlantisPreviewEditor />
           </Box>
+          <Box margin={{ top: "base", bottom: "base" }}>
```
chris-at-jobber, line 112: Could this be `Banner type="warning"`?

scotttjob: image

Like so?

packages/site/src/pages/HooksPage.tsx
```diff
@@ -13,6 +13,7 @@ export const HooksPage = () => {
         body: {
           content: hooksList,
         },
+        useCategories: true,
```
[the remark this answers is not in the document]
chris-at-jobber: (the list of content is much smaller so you should always see all of them in one view anyways, it's more just about establishing that pattern, so not critical)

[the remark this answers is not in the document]
scotttjob: Done!

packages/site/src/layout/ComponentView.tsx
chris-at-jobber: Yeah! (could have `dismissible={false}` but I don't think it hurts anyone if they _can_ dismiss?

chris-at-jobber (commented):
Design approved! Would want to get @taylorvnoj or someone similarly more qualified than myself to look at the mobile config and ovverides

packages/site/src/hooksList.ts
```diff
@@ -2,65 +2,81 @@ export const hooksList = [
   {
     title: "useAssert",
     to: "/hooks/useAssert",
+    sections: ["React Utilities"],
```
taylorvnoj, line 5: Love this addition! Much easier to parse!

packages/site/src/content/Form/index.tsx
```diff
@@ -54,12 +54,10 @@ export default {
...
 )`,
     mobileElement: `<Form
+      localCacheKey="form"
       initialValues={{ firstName: "Greatest", lastName: "Ever", nickName: "" }}
       onSubmit={(value) => {
-        return new Promise(function (resolve) {
-          setTimeout(() => {
-            resolve(alert(JSON.stringify(value, void 0)));
-          }, 100);
+                 console.log(JSON.stringify(value, void 0))
```
taylorvnoj, line 60: Looks like there's an extra bracket here - I'm getting this error under the mobile Form example:

Screenshot 2025-01-10 at 4 38 56 PM

 The problem seems to be the `}};` below this so
 
 ```
       onSubmit={(value) => {
        console.log(JSON.stringify(value, void 0))
      }}
    >
   ```
   
   should fix this.

taylorvnoj (changes requested):
Extra brackets on the Form/index.tsx file - just need to remove that to make the error disappear and we're golden

packages/site/src/content/Form/index.tsx
scotttjob: Fixed!

taylorvnoj (approved):
Thanks for making that fix! LGTM! 🎉

PR #2302, opened 2025-01-10
feat(components): Add InputDateRebuilt to better support react-hook-form

MichaelParadis:
## Motivations

We have an upcoming need to decouple from `react-hook-form`. We still want to leverage the library, but we no longer want our form components to be directly coupled, instead we want them to support being used in that environment but not use the library directly.

This PR is the next step by adding a rebuild `InputDate` that removes the react-hook-form functionality

## Testing

To test, open a storybook story (InputDate/Web.stories) and use the Version comparison story to see the behaviour of version 1 components side by side with the version 2 component

Efforts were taken to gets tests as good as possible for the new version of the component. If you forcibly swap to the new version (in the shim, force flow to the new component versus the old), you should only have failing tests in Form (which makes sense, react-hook-form is not part of the flow).

## Changes

- Added InputDateRebuilt (naming TBD) as a potential 'new' implementation of InputDate that bypasses react-hook-form.

## Decisions

Similar to https://github.com/GetJobber/atlantis/pull/2213 not dry our hooks for form inputs in `useInputDateActivatorActions.ts` initially so that we could determine bare minimum functionality needed for our new input components and refactor the hooks as needed in the future.

Due to a limitation in typescript some of the props related to aria are visible on the intellisense V1 version  that are added to the V2 component and don't cause errors when the version of the input is V1. The reason for this is that typescript doesn't check dash separated attributes like `aria-...` [see](https://github.com/microsoft/TypeScript/issues/32447)
### Added

- Added InputDate rebuilt that is shimmed and the version can be toggled with the version prop

### Fixed:
 - Fixed `InputText` version 2 not actually supporting the `onKeyDown` prop
 - Fixed `useInputTextFormField` spreading the `rest` in the incorrect order which would override the logic we implemented on certain props [see](https://github.com/GetJobber/atlantis/pull/2302/files#diff-c78e6ba1621243b99f308d57d2ad17fde3a89fc828acd5ce99051f1339f33b29L98)

## Testing

To test, open a storybook story (InputDate/Web.stories) and use the Version comparison story to see the behaviour of version 1 components side by side with the version 2 component

You can also use the PR mentioning this PR for testing on the frontend

Efforts were taken to gets tests as good as possible for the new version of the component. 

Changes can be

---

Random photo of Atlantis


Aiden-Brine:
Screenshot 2025-01-29 at 11 23 16 AM
I'm getting the following on the new docs site

packages/components/src/InputDate/InputDate.types.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Text to display instead of a date value
+   */
+  readonly emptyValueLabel?: string;
+
+  /**
+   * Version 2 is highly experimental. Avoid using it unless you have talked with Atlantis first.
+   */
+  readonly version: 2;
```
taylorvnoj, line 45: there's something going on with `CommonFormFieldProps` and getting duplicated `version` descriptions. The description you wrote here is being layered on top of the `version` prop from `FormField`. I suppose the easiest thing would be to remove the prop description here and let `FormField` handle it? Otherwise we're extending props unnecessarily?

Screenshot 2025-01-31 at 12 41 31 PM

docs/components/InputText/Web.stories.tsx
```diff
@@ -16,7 +16,9 @@ export default {
 } as ComponentMeta<typeof InputText>;
 
 const BasicTemplate: ComponentStory<typeof InputText> = args => {
-  return <InputText {...args} />;
+  const [value, setValue] = React.useState("");
+
+  return <InputText {...args} value={value} onChange={setValue} version={2} />;
```
taylorvnoj, line 21: Are the changes in this file meant to be cleaned up?
The Readonly `InputText` story is broken again

packages/components/src/InputDate/useInputDateActivatorActions.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export function useInputDateActivatorActions({
+  onChange,
+  onBlur,
+  onFocus,
+}: useInputDateActivatorActionsProps) {
+  const [isFocused, setIsFocused] = useState(false);
+
+  function handleChange(
+    _: unknown,
```
taylorvnoj, line 22: Can you give more context for the `unknown`?
Can we get away with removing that or on the flip side, being more explicit about the value type?

taylorvnoj (commented):
I'm just going to throw this in JO to triple check nothing goes sideways. Otherwise just a couple questions and the one above ^

packages/components/src/InputDate/useInputDateActivatorActions.ts
MichaelParadis: Since it was an unused parameter I just decided to use unknown because it is used to provide the change event to the InputDateActivator

docs/components/InputText/Web.stories.tsx
MichaelParadis: Thanks for catching this, I think something went wrong resolving conflicts

packages/components/src/InputDate/InputDate.types.ts
MichaelParadis: I do think the version field being duplicated is fine since both version 1 and version 2 are supported on InputDate. I am hesitant about adding it directly to FormField only because some components don't have version 2 support and mentions of version 2 on other components may confuse people.
The merging here is because both Version 1 and Version 2 versions are usable in storybook. I did update some of the punctuation to make it more readable since there are random capitalizations.

`
Experimental: Determine which version of the FormField to use. Right now this isn't used but it will be used in the future to allow us to release new versions of our form inputs without breaking existing functionality. Version 2 is highly experimental, avoid using it unless you have talked with Atlantis first.
`

taylorvnoj: Okay that's fair, just something to think about moving forward when adding prop descriptions to more v2 components

PR #2304, opened 2025-01-10
fix(docx): style sidenav correctly in Safari

chris-at-jobber:
## Motivations

Safari sidenav items were misaligned and not interacting as expected in [Safari 17.1 and older.](https://caniuse.com/css-nesting)

image

### Changed

- Moved some nested selectors out of the "nested" CSS pattern and into standalone selectors for backwards compatibility. It might be worth doing this globally, or ensuring that our CSS compiles into backwards-compatible CSS automatically, but this PR just fixes the most egregious examples I found.

### Fixed

- Search, sidenav link and disclosure items look as expected
image

## Testing

Run locally or in CF in Safari

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/layout/NavMenu.tsx
+++ b/packages/site/src/layout/NavMenu.tsx
@@ -92,7 +92,9 @@ export const NavMenu = ({ mainContentRef }: NavMenuProps) => {
           >
             <Icon name="search" color="greyBlue" />
             <span className={styles.searchButtonText}>
-              <Typography textColor={"textSecondary"}>Search</Typography>
+              <Typography size={"base"} textColor={"textSecondary"}>
+                Search
+              </Typography>
             </span>
             <div className={styles.searchKeyIndicator}>/</div>
           </button>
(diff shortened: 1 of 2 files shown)


packages/site/src/layout/NavMenu.tsx
```diff
@@ -92,7 +92,9 @@ export const NavMenu = ({ mainContentRef }: NavMenuProps) => {
           >
             <Icon name="search" color="greyBlue" />
             <span className={styles.searchButtonText}>
-              <Typography textColor={"textSecondary"}>Search</Typography>
+              <Typography size={"base"} textColor={"textSecondary"}>
```
chris-at-jobber, line 95: Safari has userAgent styling for button text that this handles.

chris-at-jobber: image

scotttjob (approved):
LGTM.

I think testing + regression + compatibility will be the name of the game for a little bit.

We're currently not running any of our CSS through any post processors, so it's probably a build step we can run with 'maximum compatibility' enabled.

PR #2305, opened 2025-01-10
fix(docx): get component iframes rendering correctly

chris-at-jobber:
## Motivations

Initially, I set the height of the iframes to have a min-height simply to avoid the internal scrolling area that was occuring, not intending to fix the components themselves not being visible, however this also seems to have fixed the components not being visible as well. 🎉 ?

### Changed

- add min-height to the iframes themselves
  - without the min-height, they seemed to just be collapsing in on themselves

### Fixed

- Components now render in Safari
image

## Testing

Run locally in Safari/Chrome/FF

---

Random photo of Atlantis


scotttjob (approved):
I wonder if Safari has some internal mechanism for height within iframes, which wasn't being updated properly by our code to shift the size of the iframe after it renders in.

Either way, approved!

PR #2309, opened 2025-01-14
docs: Add a page for figma resources

chris-at-jobber:
## Motivations

This PR is primarily an exercise in me discovering what it's like to add a net-new documentation page that _isn't_ pulling directly from an existing `stories.mdx` file. That said, I would like to add some documentation about our Figma libraries to the Atlantis docs site.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


packages/site/src/components/ContentCard.tsx
```diff
@@ -29,7 +30,11 @@ export const ContentCard = ({
     <Card
       elevation="low"
       onClick={() => {
-        history.push(to);
+        if (url) {
+          window.location.href = url;
+        } else if (to) {
+          history.push(to);
+        }
```
chris-at-jobber, line 37: I suppose there's another version of this where I do something like

`if to pass to the onClick prop, if url, pass to url` 

I just want to be able to send people to a URL outside of the SPA sometimes

packages/site/src/pages/FigmaPage.tsx
```diff
@@ -0,0 +1,40 @@
...
+      structure={{
+        header: {
+          title: "Figma for Atlantis",
+          body: "Jobber's toolkit for building consumer-grade experiences",
+          ctaLabel: "Go to Libraries",
+          to: "https://www.figma.com/files/1154197009171393179/project/68434130/Atlantis-%2B-SG-1?fuid=596716868535775450",
+          imageURL: "Figma.png",
+        },
+        body: {
+          content: [
```
chris-at-jobber, line 15: TODO: need a way to provide mixed content. Like if I want to add a few paragraphs before or after the `ContentCardSection`, can I just drop that in somewhere?

packages/site/src/routes.tsx
```diff
@@ -338,4 +339,11 @@ export const routes: Array<AtlantisRoute> = [
     inNav: false,
     exact: true,
   },
+  {
+    path: "/figma-guide",
+    component: FigmaPage,
+    handle: "FigmaGuide",
+    inNav: true,
```
chris-at-jobber, line 346: This `true` is just so I can access the page easily locally

packages/site/src/types/components.ts
```diff
@@ -2,7 +2,8 @@ import { ReactNode } from "react";
 
 export interface ContentCardProps {
   readonly title: string;
-  readonly to: string;
+  readonly to?: string;
+  readonly url?: string;
```
chris-at-jobber, line 6: I guess I could name this something like `toExternalURL` to make it clearer?

packages/site/src/components/ContentCard.tsx
scotttjob: Let's chat about this for sure, `to` is a standard prop in router world as a stand-in for `url` or `href`. I think we could have an 'external' prop, or some other flag for 'this thing goes out'.

Another option is just to inspect the content of the 'to' prop. Does it have `http` at the start? Probably can be treated as an external URL. If it starts with a `/` it's a relative link.

packages/site/src/pages/FigmaPage.tsx
scotttjob: Can we just have that content be in the MDX file? Or can we just have a mechanism to load content from multiple MDX files at once in an array format? I'm trying to keep MDX content in MDX files wherever we can.

The other option is we create another Page type. I intended "PageBlock" to be used to show an MDX file.

If there's a need for a more dynamic type of content type, we could probably look at creating that separately versus overloading PageBlock.

packages/site/src/routes.tsx
scotttjob: Do you intend for this to be under the `/design` section? In which case, we might need to add it under that section above.

packages/site/src/types/components.ts
scotttjob: I would rather a `startsWith('http')` on the to prop. Ultimately a consumer just wants to be taken somewhere on click.

scotttjob (commented):
Added some comments, but they're just comments. Let's chat on our coffee chat today!

packages/site/src/pages/FigmaPage.tsx
chris-at-jobber: Can we have a `ContentCardSection` in an mdx file? If so, that works for me

chris-at-jobber:
Closing, did the learning I needed to here and have since built out other sections/added other pages!

PR #2317, opened 2025-01-16
chore: Update ESLint config to closer resemble .eslintrc in eslint-config package

nad182:
## Motivations

We aim to create have the same ESLint rules between the repos we maintain. Jobber-frontend is the newest (most up to date) repo, so some of the rules were ported from there (they differer slightly from `@atlantis/eslint-config` ruleset).

### Added

- ESLint rules that resemble the ones in JF repo.
- missing `"react-ga4"` package to `package.json`.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 
- Formatting (linting) of `docs/README`.

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


packages/site/.eslintrc.cjs
```diff
@@ -1,15 +1,132 @@
...
     "monorepo-cop/no-relative-import-outside-package": "off",
+    "import/order": [
+      "error",
+      {
+        groups: ["builtin", "external", "internal", "index", "sibling"],
+        "newlines-between": "never",
+      },
+    ],
     "import/no-default-export": "off",
-    "import/no-cycle": "off",
```
nad182, line 7: Same as the above: the rule is still present.

packages/site/.eslintrc.cjs
```diff
@@ -1,15 +1,132 @@
 module.exports = {
+  env: { browser: true, es6: true, node: true },
```
scotttjob, line 2: This is my fault for not being clearer in the original ticket description, but I was moreso looking for us to consume the existing configuration here, not necessarily duplicate it.

What this is highlighting for me though is we have a bit of an issue with our linting/prettier setup that I'd like us to solve holistically.

You did good work here!

scotttjob (commented):
You did great work here, but I made a mistake when defining the ticket and it resulted in a solution that I wasn't intending.

Your work here though has highlighted a glaring issue with our linting setup that is going to require a bit of a lift on its own to solve. You at all interested in helping determine what the complete solution will look like for the whole repo?

I think we can keep this PR for historical purposes, but I don't think we should merge it in this shape (with the config file duplicated), and I don't think it's worth the additional effort (in this sprint) to "correct"

Correct is in quotes because I think the original ask has proved to be not the right direction. We should be updating all of the codebase to point towards the future, instead of asking the new code to point at the past.

nad182:
Closing as WON'T DO **right now**. We will have to rethink how we want to approach creating a single ESLint config to be shared across all packages in Atlantis and (ideally) in other Jobber repos (i.e. JO, JF etc.)

PR #2318, opened 2025-01-16
fix(components): Add transparent outline for DataList sort options 

Aiden-Brine:
## Motivations

We are already adding a box shadow data list sorting options when focused so we don't need an additional outline. As you can see the current behaviour is quite busy:
image

## Changes

Makes the outline on focused sorting options transparent

## Testing

Here is a video of the new behaviour when tested via the pre-release

https://github.com/user-attachments/assets/3433ca33-39f1-4029-bad4-b0920f82f887

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.module.css
+++ b/packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.module.css
@@ -37,4 +37,5 @@
 
 .option:focus-visible {
   box-shadow: var(--shadow-focus);
+  outline: transparent;
 }


packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.module.css
```diff
@@ -37,4 +37,5 @@
 
 .option:focus-visible {
   box-shadow: var(--shadow-focus);
+  outline: transparent;
```
chris-at-jobber, line 40: ❤️

taylorvnoj (approved):
Thank you @Aiden-Brine ❤️ Tested in JO also!

packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.module.css
chris-at-jobber: I think `Card` has this issue too, if you want to take a peek at similar instances 😄

PR #2319, opened 2025-01-16
fix(components): Fix label issue with SegmentedControl on older versions of Safari [113267]

Aiden-Brine:
## Motivations

On old versions of Safari (before 18) the labels on SegmentedControl were displaying far off of where they belonged. 

Screenshot 2025-01-16 at 3 19 04 PM

This was only observed in our new docs site but could also present itself in other implementations. The reason that this was only impacting old versions of Safari is that in 18 they fixed the bug: 
> Fixed overlapping elements with flex box when height: 100% is applied on nested content.

As far as I can tell the height of 100% on the label container was providing no benefits so removing it should allow the labels to display as expected in all versions of Safari as well as other browsers.

### Removed

The height of 100% from the label container for SegmentedControl labels

### Fixed

- 

### Security

- 

## Testing

Ensure that SegmentedControl continues to behave the same and as expected on browsers other than Safari. If you have Safari 18 installed you can use Browser Stack's Local Testing feature to test out the changes with an older version of Safari.

Screenshot 2025-01-16 at 3 26 33 PM

I tested with Safari 17.3 which was broken before the changes (leading to the screenshot in the Motivations section):

Screenshot 2025-01-16 at 3 18 15 PM

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.module.css
+++ b/packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.module.css
@@ -37,4 +37,5 @@
 
 .option:focus-visible {
   box-shadow: var(--shadow-focus);
+  outline: transparent;
 }
--- a/packages/components/src/SegmentedControl/SegmentedControl.module.css
+++ b/packages/components/src/SegmentedControl/SegmentedControl.module.css
@@ -29,7 +29,6 @@
 .container label {
   display: flex;
   z-index: 1;
-  height: 100%;
   padding: 0 calc(var(--space-base) - var(--space-smaller));
   overflow: hidden;
   font-size: var(--typography--fontSize-base);


packages/components/src/DataList/components/DataListHeaderTile/components/DataListSortingOptions.module.css
```diff
@@ -37,4 +37,5 @@
 
 .option:focus-visible {
   box-shadow: var(--shadow-focus);
+  outline: transparent;
```
chris-at-jobber, line 40: He does it again!

chris-at-jobber:
Confirmed that 
- issue is fixed in Safari 17.1
image

- does not introduce any new issues in other browsers

### Chrome
image

### Edge
image

### Firefox
image

chris-at-jobber (approved):

PR #2333, opened 2025-01-22
feat(components): Allow for external label [JOB-113378]

Aiden-Brine:
## Motivations

Users would benefit from being able to provide a label for an input external to that label. Along with that, they would also benefit from having the ability to hide the mini label that appears when a user starts typing into the input. I made the decision to modify and export the existing `FormLabelField` with a couple of small API changes. Since `FormFieldLabel` was not exported from `index.ts` before this these API changes should be non-breaking. I also fixed some examples in the `InputText` section of the new docs site that were not rendering before.

## Changes

- FormFieldLabel is now exported from `index.ts` so that it can be used directly
- FormFieldLabel now accepts what it will display as a child as opposed to a `placeholder` prop. I think this API is closer to what users expect and will be more flexible in the future
- The `identifier` prop on FormFieldLabel has been renamed to `htmlFor` which should be a more familiar and logical name for users

### Added
- FormFieldLabel now takes an `external` prop to specify if the label is to be internal or external to the input
- FormField and in turn InputText now accept a `hideMiniLabel` prop to no longer show the mini label once a user adds a value
- The ability to provide an `id` to various inputs so that a user can provide the same value for the `id` of an input as the `htmlFor` of an external label

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Half of the examples in the docs for [ InputText on the new docs site](https://atlantis.getjobber.com/components/InputText) which were not rendering before. Also the show code for these were not working in Storybook.

### Security

- 

## Testing

I added an example to the [InputText docs](https://atlantis.getjobber.com/components/InputText) if you want to see it in action. Also please ensure that none of the other inputs are broken

Changes can be
[tested via Pre-release](http://localhost:6005/?path=/docs/components-forms-and-inputs-inputtext--docs)

---

Random photo of Atlantis


packages/components/src/InputLabel/InputLabel.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import styles from "./InputLabel.module.css";
+
+export function InputLabel({
+  label,
+  inputId,
+  style,
```
chris-at-jobber, line 7: Was thinking this should perhaps use the [UNSAFE pattern ](https://atlantis.getjobber.com/guides/customizing-components)if we go this route?

Aiden-Brine: Interesting thought. I thought the ticket mentioned that it needed the ability to accept custom styles but I think that might have just come from my conversation with Michael. Do either you or @MichaelParadis know if we already have plans to use this external label with custom styles? Just trying to gauge how actively we want to discourage users from providing their own styles here

MichaelParadis: I am not aware of any custom style needs but unsafe may be something to consider using

docs/components/InputText/InputText.stories.mdx
```diff
@@ -15,22 +17,38 @@ Input text is used in forms that accept short or long answers from users.
...
-          name="age"
-          placeholder="Age in words"
-        />
-      );
-    }}
-  </Story>
+  <ControlledExample />
 </Canvas>
 
+<Disclosure title="Show code">
```
Aiden-Brine, line 36: This was the best I could come up with here but I wrote up [a ticket](https://jobber.atlassian.net/browse/JOB-114601?atlOrigin=eyJpIjoiYzY1MWVmM2IxYTY0NDI2OGI2MTJiODdmZGEwYjZjNjQiLCJwIjoiaiJ9) to improve it in the future

packages/components/src/FormField/FormField.tsx
```diff
@@ -11,7 +11,8 @@ export function FormField(props: FormFieldProps) {
   // Warning: do not move useId into FormFieldInternal. This must be here to avoid
   // a problem where useId isn't stable across multiple StrictMode renders.
   // https://github.com/facebook/react/issues/27103
-  const id = useId();
+  const generatedId = useId();
```
Aiden-Brine, line 14: This will lead to an ID being generated and not used at times but that isn't expensive and `useId` can't be called conditionally because it is a hook

packages/components/src/FormField/FormFieldWrapper.tsx
```diff
@@ -157,19 +163,25 @@ export function FormFieldWrapperMain({
...
 }) {
-  if (!placeholder) return null;
+  if (!children) return null;
 
   return (
-    <label className={styles.label} htmlFor={identifier} style={style}>
-      {placeholder}
+    <label
+      className={external ? styles.externalLabel : styles.label}
+      htmlFor={htmlFor}
```
taylorvnoj, line 181: Love how when you click on the label, it focuses the input. Cool! I made a storybook example with 3 inputs and each label click focuses the corresponding input

docs/components/InputText/InputText.stories.mdx
taylorvnoj: Fair - it's good to have the code there. I don't want us to start adding disclosures in other places and bulk up the examples.  There's something to be said about having the different variations on the design page and code on an Implement tab but we don't need to think about that here and now.

docs/components/InputText/InputText.stories.mdx
```diff
@@ -162,6 +164,33 @@ See:
   <InputText placeholder="Email" value="atlantis" invalid={true} />
 </Canvas>
 
+### External label
```
taylorvnoj, line 167: Do you think it's a good idea to have an example within `InputText/Web.stories.tsx` as well?
We are still using Storybook examples outside of the new docs site _and_ hopefully we'll get to a point where we're using the web stories as examples to display on the new docs site. I think it's worth adding an example as it's another variation of this component we can test against in the future.

taylorvnoj (changes requested):
Just a couple documentation related suggestions. I still have some testing to do, although I'm sure nothing will come up.

packages/components/src/InputLabel/InputLabel.tsx
taylorvnoj: This thread made me think about 2 things:

I realize we no longer have `InputLabel` however now that `FormFieldLabel` is being exported, should `FormFieldLabel` also be moved outside of `FormFieldWrapper` - at least into the components folder, maybe? Even just for the sake of code clean up and composability?

Re: Adding `UNSAFE_` - to me, it's important to get ahead of whatever teams may implement in the future by adding escape hatches now. Otherwise they will just revert to a wrapping div and cleanup would be unlikely. That being said, `FormFieldLabel` accepts `children` so that's flexible and we'd be implementing `UNSAFE_` on the container. I wouldn't be surprised if  someone wanted to customize that in the future however I'm ok with not doing that right now.

packages/components/src/FormField/FormFieldTypes.ts
```diff
@@ -68,6 +73,13 @@ export interface CommonFormFieldProps {
    */
   readonly disabled?: boolean;
 
+  /**
+   * If this is false the placeholder is shifted to a mini label inside the input
+   * once a value has been added. Setting this to true hides the mini label.
+   */
+
+  readonly hideMiniLabel?: boolean;
+
```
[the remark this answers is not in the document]
chris-at-jobber: This feels really unintuitive

Set to true, to remove something from the view?

I know this isn't an "open/close" prop, but this feels like it should follow the mental model of our [open/close prop naming guidance.](https://arc.net/l/quote/adjdxejz)

```
Controls the visibility of the floating mini label that appears inside the input
when a value is entered. By default, the placeholder text moves up to
become a mini label. Set to false to disable this behavior.

@default true

showMiniLabel?: boolean
```

"enabled means you see the thing, false means you don't"

packages/components/src/InputLabel/InputLabel.tsx
scotttjob: > This thread made me think about 2 things:
> 
> I realize we no longer have `InputLabel` however now that `FormFieldLabel` is being exported, should `FormFieldLabel` also be moved outside of `FormFieldWrapper` - at least into the components folder, maybe? Even just for the sake of code clean up and composability?
> 
> Re: Adding `UNSAFE_` - to me, it's important to get ahead of whatever teams may implement in the future by adding escape hatches now. Otherwise they will just revert to a wrapping div and cleanup would be unlikely. That being said, `FormFieldLabel` accepts `children` so that's flexible and we'd be implementing `UNSAFE_` on the container. I wouldn't be surprised if someone wanted to customize that in the future however I'm ok with not doing that right now.

It was my suggestion to keep this within FormFieldWrapper (for now), if only to prevent an explosion of small components that are really part of a larger whole. 

BUT I also wanted to leave some time for this discussion in an upcoming chat (early next week) to review that strategy and ask ourselves do we want to continue the compound pattern, or say 'nay' and just ship smaller components with no tie-in to a larger organism (and make it easier to spin up new docs pages for them?). I'm okay with that path forward, but I'd like to see a plan for the exact small components that we feel are missing so we have some guideposts to keep things on track. Creating new standalone components is easy, but taking them away or changing them can be hard, so I want our moves in this space to feel deliberate instead of reactive.

To match what we've been building so far I would prefer if this was `FormField.Label` or `Input.Label` versus a brand-new component (but I do think leaving it as-is in FormFieldWrapper as it is in this PR is a good first step). I think in the future we'll have a variety of these small form related pieces in which you can compose the full form field component, and the discovery of typing `FormField.` in your IDE and have all the options just appear will be a powerful building tool. Strong JSDocs on each of the compound parts should also really help devs.

All that said, I do want to have this discussion with the team in case I'm not looking at the problem properly. Maybe there's a place for compound and a place for small disconnected components, in which case we can just clearly define that line and we can make them Tenants!

taylorvnoj: That all sounds great @scotttjob! I'm onboard with leaving `FormFieldLabel` inside `FormFieldWrapper` in this PR and then having that larger conversation with the team.

packages/components/src/FormField/FormFieldTypes.ts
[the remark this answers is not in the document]
Aiden-Brine: I can see your point of view and if that is more consistent with what we are doing elsewhere I am happy to change it. I did give it some thought at the time and my reasoning for going this way was that users will only be providing this prop at times when they want to hide the mini label. So keeping that in mind you either have:

`<FormFieldLabel hideMiniLabel={true}>` or `<FormFieldLabel showMiniLabel={false}>`

To me the first is more immediately clear what is happening (the mini label is being hidden). I don't think we would ever have the code:

`<FormFieldLabel hideMiniLabel={false}>` or `<FormFieldLabel showMiniLabel={true}>`

I don't feel strongly though and I do see your perspective so I'm happy to update it if you still feel as though it would be better flipped?

packages/components/src/FormField/FormFieldTypes.ts
```diff
@@ -110,7 +122,8 @@ export interface CommonFormFieldProps {
   onValidation?(message: string): void;
 
   /**
-   * Hint text that goes above the value once the form is filled out.
+   * Hint text. If hideMiniLabel is false this text provides a mini label inside the
+   * input once a value has been added.
```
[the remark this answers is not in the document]
Aiden-Brine: Yeah that's fair, I had the sense this wasn't great when I was proofreading it and I should have stopped to take the time to re-write. Good suggestions and I will re-write it now

docs/components/InputText/InputText.stories.mdx
Aiden-Brine: That's fair, it is a bit of a unique case where we have so many examples inline in this file and some of those examples are reliant on state. I don't see having this be the norm by any means

docs/components/InputText/InputText.stories.mdx
Aiden-Brine: Good call, I can add one

packages/components/src/FormField/FormFieldTypes.ts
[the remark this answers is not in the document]
chris-at-jobber: That does make sense, and I think in the single context of this usage is probably a little better ("I want to _hide_ the miniLabel, how do I do that?"). I did do a (very non-scientific) check of the system and it looks like we are more consistent in using `show / true` as a pattern, even in cases where the `show` is indeed the default so the prop will only be used to hide the element, so in terms of overall learnability of the system I would lean to following that pattern over optimizing for discoverability of this individual prop, if that makes sense?

image
image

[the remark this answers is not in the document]
chris-at-jobber: (I don't want my comment to be a blocker/required change, but just want to make sure we've considered the convention in the broader context of our components/system)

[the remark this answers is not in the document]
taylorvnoj: That's a good call out Chris, it's more intuitive to have something removed from view with false.
This brings me back to `keepOpenOnClick` chats 😅

packages/components/src/FormField/FormFieldWrapper.tsx
```diff
@@ -95,7 +95,7 @@ export function FormFieldWrapper({
         <FormFieldInputHorizontalWrapper>
           <AffixIcon {...prefix} size={size} />
           <FormFieldInputWrapperStyles>
-            {!(value && hideMiniLabel) && (
+            {!(value && !showMiniLabel) && (
```
taylorvnoj, line 98: Sorry @Aiden-Brine, tell me if this makes sense. 
Do you feel there's a difference in readability between:

`{!(value && !showMiniLabel) && (`

vs

` {(showMiniLabel || !value) && (`
to say, show the label when `showMiniLabel` is true or when there is no value.

Aiden-Brine: @taylorvnoj Nice catch! The latter is definitely clearer. I used Cursor to zip through the code and swap out `hideMiniLabel` for `showMiniLabel` but I didn't stop to improve the readability after that change which I should have. Updated now 😄

taylorvnoj (approved):
LGTM! I tested:
- `InputText` stories
- `InputText` in product
- `InputText` with external label in product
- `FormField` stories out of an abundance of caution.

Aiden-Brine:
> LGTM! I tested:
> 
> * `InputText` stories
> * `InputText` in product
> * `InputText` with external label in product
> * `FormField` stories out of an abundance of caution.

@taylorvnoj thank you very much for the extensive testing!

PR #2326, opened 2025-01-21
fix(docx): use a correct link for star wars graphQL API (useCollectionQuery hook)

nad182:
## Motivations

There were a bunch of console errors when trying to check the docs for `useCollectionQuery` hook, since the star wars API link specified no longer worked.

### Added

- 

### Changed

- Changed the graphQL API link to the one 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

1. Navigate to https://atlantis.getjobber.com/hooks/useCollectionQuery
2. Observe CORS error in the console (and network tab) and no returned results in the UI (no list of Star Wars planets)
3. Pull the branch locally or open a preview website
4. Notice the graph request is now successful (200 OK) as well as now you can see the list of planets in the UI.
  4.1. CORS error is no longer present in Chrome console.

Changes can be
[tested via Pre-release](https://github.com/GetJobber/atlantis/actions/workflows/trigger-qa-build.yml)

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

Random photo of Atlantis

PR #2335, opened 2025-01-23
fix(hooks): Force bump packages to avoid lerna problem

jdeichert:
## Motivations

We had a publish on Nov 4th that failed to hit npm, but the git tags were still pushed to github. We reverted the publish commit on github to undo the package.json version changes but subsequent publishes apparently didn't try to publish hooks nor generators packages again. 2 days ago a PR modified the hooks package https://github.com/GetJobber/atlantis/pull/2326 which lerna failed to publish because the git tags were never deleted from Nov 4th.

### Fixed

- Lerna version

## Testing

No testing necessary.

Changes can be

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jobber/components",
-  "version": "6.11.5",
+  "version": "6.11.6",
   "license": "MIT",
   "type": "module",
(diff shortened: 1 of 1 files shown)


MichaelParadis (approved):
LGTM, those versions are not on npm and the tags do not exist for these versions

PR #2337, opened 2025-01-23
feat(components): Enhance image lightbox layout and background

anagelberg:
## Motivations

This PR is the first of two enhancements to the Lightbox component. A separate PR to add thumbnails will follow. 

## Changes

- The background of the Lightbox now has a "stained glass" styling effect, as opposed to plain grey
- The title and caption of the image were changed in the layout to be more prominent
- Image count is included
- Image sizing is contained more precisely
- Navigation buttons hide when there isn't mouse movement (as per comment, animations added)

### Changed

https://github.com/user-attachments/assets/31c86ecd-523c-4530-9314-d705f2916186

Changes can be

---

Random photo of Atlantis


packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,56 +15,79 @@
...
+  z-index: var(--elevation-base);
+  margin-bottom: var(--space-larger);
+  text-align: center;
+  flex-direction: column;
+  align-items: center;
+  gap: var(--space-small);
+}
+
+.caption {
+  min-height: calc(17.5px * 3);
```
anagelberg, line 87: The purpose of this is to minimalize bouncing between images the majority of the time. Design may want to truncate this caption to 2 lines, but this is still being decided as it doesn't seem common. For now some bouncing past 2  lines is OK. 

However, I couldn't find a variable that matches the computed line height of the `Text` element anywhere?

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -133,44 +136,63 @@ export function LightBox({
...
+                  drag="x"
+                  dragConstraints={{ left: 0, right: 0 }}
+                  dragElastic={1}
+                  onDragEnd={handleOnDragEnd}
+                />
+              </AnimatePresence>
+            </div>
+
+            {images.length > 1 && <NextButton onClick={debouncedHandleNext} />}
+          </div>
```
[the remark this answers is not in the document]
viccyc: Also, they jump a wee bit when reducing screen size. Not sure if this is ok?

https://github.com/user-attachments/assets/3998b5d3-df78-4cd0-a100-7e361bdf8758

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -133,44 +136,63 @@ export function LightBox({
...
 
-          <div className={styles.toolbar}>
-            {images[currentImageIndex].caption}
+          <div className={styles.captionWrapper}>
+            <AtlantisThemeContextProvider dangerouslyOverrideTheme="dark">
```
viccyc, line 183: For some reason, the margin under the photo image number looks bigger on yours than the design, but I can't click into the design enough to see what it should be. I'm not sure if that's just my photo?

Screenshot 2025-01-27 at 12 19 09 PM

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -133,44 +136,63 @@ export function LightBox({
...
+              {images.some(image => image.title) && (
+                <Heading level={4}>
+                  {images[currentImageIndex].title || "\u00A0"}
+                </Heading>
+              )}
+              {images.some(image => image.caption) && (
+                <div className={styles.caption}>
+                  <Text>{images[currentImageIndex].caption}</Text>
```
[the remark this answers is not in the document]
viccyc: The text doesn't seem to line up under the heading when there's more. Also, is there a max allowed chars?

Screenshot 2025-01-27 at 12 23 02 PM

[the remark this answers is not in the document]
viccyc: I'm also not 100% sure the padding is correct. Maybe @kerry-mac can confirm as some of it I can't click into and maybe it's my photo is within the max area?

Screenshot 2025-01-27 at 12 27 06 PM

viccyc (changes requested):
Looks great. I have a few minor styling comments. The CSS I think Atlantis can confirm more because that's not my area 😆

packages/components/src/LightBox/LightBox.tsx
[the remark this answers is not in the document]
viccyc: The next arrows also move up if there's a lot of text underneath

https://github.com/user-attachments/assets/4f6d2168-9e8d-422f-8849-bae32953b94c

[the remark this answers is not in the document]
anagelberg: > The text doesn't seem to line up under the heading when there's more. Also, is there a max allowed chars?

> The next arrows also move up if there's a lot of text underneath

These are related. Originally design was thinking we might truncate at 2 lines but have decided to leave this until they see a use case for it from my understanding. My understanding from chatting with @kerry-mac is that since this is very infrequent and if it occurs often we should revisit truncating, the jumping was okay at this time. Maybe @chris-at-jobber might be able to confirm while she's away though, seeing this in action?

packages/components/src/LightBox/LightBox.tsx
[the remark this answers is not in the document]
anagelberg: Totally missed those button sizes on mobile 😶‍🌫️ but I've added that now. As far as this jumping on screen size, there were two reasons for it: 

1 was a mistake that they weren't entirely centered on the image on desktop, which I've fixed. However, there will still be a little bit of jumping from the change of size / handling of the image, as well as the relative available height for the image shifting due to increased space left for captions on mobile (goes up to 3 lines). I don't foresee that being a big deal or avoidable with the current treatment :/

packages/components/src/LightBox/LightBox.tsx
anagelberg: Confirmed that this is indeed a little big. Looks like the design has a 40px size, which doesn’t correspond to any spacing in the design system 🤔  (it’s 32px or 48px). @Chris-murray can we confirm?

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,126 @@
...
 
-.imagesWrapper {
-  display: flex;
-  justify-content: center;
-  position: relative;
-  flex: 1 1 0;
-  min-height: 0;
+.caption {
+  min-height: calc(17.5px * 3);
+  text-align: left;
```
anagelberg, line 123: > @viccyc The text doesn't seem to line up under the heading when there's more. 

Hmm, that is dramatic, 

It is centered on the text box: 

Screenshot 2025-01-27 at 2 55 59 PM

It's off because the way left aligned text works it won’t go all the way to the edge, which is exaggerated in your screenshot because using one word will cut off the same place every time.  I wonder if we should justify it? :thinking: 

Justified example: 

Screenshot 2025-01-27 at 3 23 32 PM

packages/components/src/LightBox/LightBox.tsx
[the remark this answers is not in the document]
anagelberg: > The design has the caption heading the same font size as the text (16px)

Good eye, fixed 

> I also think the text should be centred on smaller screen

Latest design is not text centered, even on mobile 

> I'm also not 100% sure the padding is correct. 

It wasn't, great eye, I fixed a few paddings that I mixed up. Is there anything else specific that you notice that still looks off? 

Note, since the thumbnails aren't here yet, I just left the padding we're expecting under the caption text. In addition, we do leave a height equivalent of 2 lines for captions so if there's only 1 that's going to make that space seem bigger.

packages/components/src/LightBox/LightBox.tsx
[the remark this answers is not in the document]
chris-at-jobber: Video isn't rendering for me, but if I understand correctly from the description - don't think we need to optimize for the experience while resizing the viewport (if that was the concern)

packages/components/src/LightBox/LightBox.tsx
chris-at-jobber: Let's start with 48 (largest), easy to change if needed

packages/components/src/LightBox/LightBox.module.css
chris-at-jobber: Alignment's always going to be tricky when you have centered + left paired, the SSSSNNNAKKKE example really highlights the risks but with a more natural mix of word and line-lengths I'd expect it will look fine.

packages/components/src/LightBox/LightBox.tsx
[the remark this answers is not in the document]
chris-at-jobber: At a broad level, what Kerry and I had discussed was starting off with flexibility/fewer restrictions here, experiment with content length suggestions or restrictions at a feature/instance level, and then we can layer in truncation or progressive disclosure patterns here later as needed. The arrows jumping when the content changes is a pain for sure - we could look at other ways of positioning that don't rely on matching the vertical centre of the image, or increasing the hitbox area so the user can still click through even if the button itself is moving around.

chris-at-jobber:
@anagelberg one other piece of feedback as I've been testing this is the interaction on the "hide while resting" functionality of the prev/next arrows. Quick demo video [here](https://getjobber.zoom.us/clips/share/GxQVo3dLQOOGn72y4pOUJw)

Wondering if we could extend the timeout before the arrows hide to slightly longer, and add a transition like `transition: opacity var(--timing-base) ease-in` to that show/hide change?

packages/components/src/LightBox/LightBox.tsx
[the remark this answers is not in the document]
anagelberg: thanks so much for all the feedback @chris-at-jobber!  Just wondering do mean we could add those in here to deal with the button issue with the current PR, or as a future enhancement once we experiment more? I'd be fine with either of the mentioned calls, is there a preferred pattern at this time?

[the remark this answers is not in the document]
chris-at-jobber: Discussed "offline" by which I mean "online but over a video call": we'll try fixing the position of the arrow buttons at a reasonable height as a first solve for this.

packages/components/src/LightBox/LightBox.test.tsx
```diff
@@ -1,135 +1,147 @@
...
-  expect(queryByText(title)).toBeNull();
-  expect(queryByText(caption)).toBeNull();
-});
+const { setViewportWidth } = mockViewportWidth();
 
-test("lightbox closes when user clicks close", () => {
-  const title = "Dis be a title";
-  const caption = "Dis be a caption 🎉";
+describe("LightBox", () => {
+  beforeEach(() => {
```
anagelberg, line 12: Just added the describe block here

packages/components/src/LightBox/LightBox.tsx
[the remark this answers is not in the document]
anagelberg: The navigation buttons are now set to be 43% from top of screen, which roughly translates to the middle of image. We'll have to readjust this center with addition of thumbnails which will push it up a bit.

viccyc (commented):
Looks great! Thanks for the changes.
Only thing I'd say is... when the text overlaps where the arrows are, the arrows can't be clicked. Maybe another case for limiting the number of chars?

Uploading Screen Recording 2025-01-29 at 10.32.50 AM.mov…

chris-at-jobber:
Another option there could be to constrain the text width to avoid the overlap.
image

We could also start with something like `maxLines="larger"` on the Text and ensure that we're always setting the buttons higher than 16 lines of text and see if that creates any issues. I'd prefer to avoid truncation as a default, but 16 lines feels reasonably generous.

anagelberg:
@chris-at-jobber @viccyc  I think we're going to need to go with adjusting the max width of the text if we want to avoid the overlap. The thumbnails will leave less space for the text so I think we'd need to truncate at estimated ~6 lines if we were to go that way, which might be a bit too much?

@chris-at-jobber  how do you feel about the overlap generally, if I take a look at ensuring they work? These rarely show because the user has to tap the screen first to see the buttons

chris-at-jobber:
I think this looks alright with the updates @anagelberg !

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -132,45 +143,72 @@ export function LightBox({
           aria-label="Lightbox"
           key="Lightbox"
           ref={lightboxRef}
+          onMouseMove={handleMouseMovement}
```
ZakaryH, line 146: hmmm I have some concerns about how often we're going to be triggering this since it is setting state on each call we'll be triggering a ton of re-renders

we'll either need to debounce it, or see if we can't find another way to achieve the same result. I'll have to dig into what we're using the `mouseIsStationary` for first!

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
+  max-width: 600px;
+  margin: 0 auto var(--space-larger);
+  text-align: center;
+  flex-direction: column;
+  align-items: center;
+  gap: var(--space-small);
+}
+
+.caption {
+  min-height: calc(21.438px * 3);
```
ZakaryH, line 140: this feels a bit magical/arbitrary

is there a constant we can replace it with, or rather how did we arrive at this very specific value I guess is my real question

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
+}
+
 .next,
 .prev {
-  display: flex;
   position: absolute;
-  top: 50%;
-  z-index: var(--elevation-base);
-  transform: translateY(-50%);
+  top: 43%;
```
ZakaryH, line 34: similar thought here, what's the rationale or reason for such a specific value?

packages/components/src/LightBox/LightBox.module.css
anagelberg: Oh I had called this out earlier as a question too, got lost in the mix! `21.438px` is the height of a single line of the `<Text>` component container set to large size, and I couldn't find a variable for it using line heights anywhere. I could set a constant so it's more clear what it is within the file, but I'm sure you're more familiar with how to calculate that using what's available?

packages/components/src/LightBox/LightBox.module.css
anagelberg: I agree, this is fairly arbitrary. It's approximately the center of the max image container (optimized for default 1 line title, 2 line caption), but the buttons use the full viewport for fixing, instead of that image area ([see this thread in this PR](https://github.com/GetJobber/atlantis/pull/2337#discussion_r1931085592)). Originally we had these centered based on image height itself, but it causes a UX issue in the event of larger captions making it shift so we want them fixed in a place and accepting they will look less ideal for edge cases, but won't bounce. This is also subject to change when we add thumbnails in. I'm very open to other ideas for how to address this, definitely it isn't ideal

packages/components/src/LightBox/LightBox.tsx
anagelberg: mouseIsStationary is used to track that movement so that the buttons can disappear and allow better showcasing of the images. I think that's a very valid concern though, I'll take another look at this and see if the debounce angle can work 🙏

anagelberg: I debounced this, I had definitely overcomplicated both the logic and performance to get this to work. Thanks for the guidance on it!

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -101,9 +106,12 @@ export function LightBox({
   );
   const mounted = useIsMounted();
   const prevOpen = useRef(open);
-
   useRefocusOnActivator(open);
 
+  const handleMouseMovement = debounce(() => {
+    setMouseIsStationary(true);
+  }, 1000);
```
ZakaryH, line 113: adding some logging, I'm seeing more re-renders that I think we can avoid - also something to watch out for when debouncing functions in React is to make sure we have a stable reference to the function we're debouncing. we have a handy utility hook that can take care of that for you `useDebounce`

here's something I came up with that will trigger pretty close to a minimum amount of rerenders


```
import { useDebounce } from "../utils/useDebounce";
...
const MOVEMENT_DEBOUNCE_DELAY = 1000;
...
export function LightBox({...}) {
...

  const debouncedHandleMouseMovement = useDebounce(() => {
    setMouseIsStationary(true);
  }, MOVEMENT_DEBOUNCE_DELAY);

  const handleMouseMovement = () => {
    if (mouseIsStationary) {
      setMouseIsStationary(false);
    }
    debouncedHandleMouseMovement();
  };
...
          onMouseMove={handleMouseMovement}
        >

```

the `MOVEMENT_DEBOUNCE_DELAY` can be adjusted of course to make it feel more responsive to when you stop moving the mouse.

I was tempted to use a `throttle` instead but thinking on what we actually want, we only really care about the mouse movement stopping. there's nothing relevant about the events as they keep happening that we'd need to sample from which is when I'd reach for `throttle`

lemme know if that makes sense or if you have any questions!

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
+  z-index: calc(var(--elevation-base) + 1);
```
ZakaryH, line 35: I'll have to see what else is at this same elevation that we're fighting with since it seems like this used to work with only the base elevation.

if this must go higher, we should use a higher token rather than incrementing it ourselves

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
-  align-items: center;
+.toolbar button {
+  grid-column: 3;
+  justify-self: end;
 }
 
-.imagesWrapper {
-  display: flex;
-  justify-content: center;
+.toolbar p {
```
ZakaryH, line 100: not the biggest fan of targeting the specific tag. gonna see if we can't make some generic rules about the containers that will apply regardless of the content.

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
+  height: 100%;
+  max-width: 100%;
+  margin-bottom: var(--space-large);
+  @media (--medium-screens-and-up) {
+    padding: 0 calc(var(--space-largest) + var(--space-base)); /* nav button width + padding */
+  }
+}
+
+.image {
+  position: absolute;
```
ZakaryH, line 116: doesn't seem like this absolute position is doing anything anymore if we're not telling it to be anywhere in particular

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
+  width: 100%;
+  height: 100%;
+  object-fit: contain;
+}
+
+.imageArea {
```
ZakaryH, line 122: gonna see if we can't avoid nesting a div immediately inside another div with this and `slideWrapper`

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
-  flex: 1 1 0;
-  min-height: 0;
+  width: 100%;
+  height: 100%;
+}
+
+.captionWrapper {
+  display: flex;
+  z-index: var(--elevation-base);
+  max-width: 600px;
```
ZakaryH, line 131: is there a specific reason for 600?

I'm hesitant to apply a hardcoded width/height when the sizes of the screen can vary from mobile to very large.

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
+  margin: 0 auto var(--space-larger);
+  text-align: center;
```
ZakaryH, line 133: what are we trying to target with this rule?

ZakaryH: I ask because the `align-items` is going to center the heading element, and it won't do anything to the `Text` we're using for the actual caption

if we want the caption to be centered, we'll need to give it the `align="center"` prop

packages/components/src/LightBox/LightBox.module.css
anagelberg: This is the maximum value set by design to ensure readability on the larger screens

packages/components/src/LightBox/LightBox.module.css
ZakaryH: ah gotcha. yeah we do have some variables/tokens for the line height of text but they're ratios, not direct pixel values so we'd need to adjust this a bit

taking a step back, what are we looking to accomplish with these min heights? I wonder if there isn't another way to achieve that where we don't need to make assumptions about the text size multiplied by a number of lines that may or may not be present since we really can't know how much of a caption there is.

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
-}
-
-.image {
-  position: absolute;
-  top: 50%;
-  max-width: 100%;
-  max-height: 100%;
-  margin: auto;
+  width: 100%;
+  height: 100%;
```
ZakaryH, line 154: I think these height/width declarations are redundant no? since we're already telling it to be fixed + top,right,bottom, and left of `0` - effectively telling it to completely fill the space

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,143 @@
...
+  backdrop-filter: blur(30px);
+  background-color: rgba(3, 43, 58, 0.25);
```
ZakaryH, line 156: is there a color token we can use here instead? I'm really not sure what rgb(3,43,58) is 😅

packages/components/src/LightBox/LightBox.module.css
anagelberg: Good call on backing up on it. The design request was to hold space for all the images in a set to minimum of 2 lines if any of them have a caption. This was to avoid the layout shifting between the images with various size captions, optimizing for the predicted most common use cases (1-2 lines).

 It was most important when the arrows were centered on the image itself though, now that it isn't, we might be able to simplify this to not depend on these heights. My understanding was that we should still try to minimize how the layout shifts between images with different sizes where possible, but I think we could treat it more like the title with a blank fallback now that the arrows aren't dependent (would just be 1 line, instead of 2-3). Alternatively we could use a less specific value that should still hold the space in this case, if that seems preferred?

packages/components/src/LightBox/LightBox.tsx
anagelberg: This makes a ton of sense, thank you so much for the detailed explanation here. I changed to this and it seems to work really well 🙏

packages/components/src/LightBox/LightBox.test.tsx
```diff
@@ -1,135 +1,192 @@
...
-          url: "https://i.imgur.com/6Jcfgnp.jpg",
-        },
-      ]}
-      onRequestClose={handleClose}
-    />,
-  );
-
-  fireEvent.click(getByLabelText("Close"));
-});
+  test("lightbox doesn't show up", () => {
```
ZakaryH, line 38: I know this was an existing test assertion but I don't really get _why_ the lightbox wouldn't show up in this scenario.

you have to examine the props to understand what's different about this setup and leads to the result which I think could be improved with a `context` wrapper

since it was already like this, it's probably fine to leave it as is, but if you're up for a small refactor of the tests I think it would really help make this test suite easier to understand!

packages/components/src/LightBox/LightBox.module.css
anagelberg: The design specifies surface—reverse with 25% opacity (that is a distinct color, isn't it? I just double checked), but in order to work right with the blur it needs to be defined via an rgba value. I’m double checking this if there isn’t another way

packages/components/src/LightBox/LightBox.test.tsx
```diff
@@ -1,135 +1,192 @@
...
+const { setViewportWidth } = mockViewportWidth();
 
-test("lightbox closes when user clicks close", () => {
-  const title = "Dis be a title";
-  const caption = "Dis be a caption 🎉";
+describe("LightBox", () => {
+  beforeEach(() => {
+    setViewportWidth(BREAKPOINT_SIZES.lg);
+  });
+  test("Lightbox opens and shows the image", () => {
```
ZakaryH, line 15: nitpick: since we're nested in a `Lightbox` `describe` all these `test`s don't really need to be prefixed with `Lightbox` anymore since when they run, it'll read as: "Lightbox Lightbox opens and shows the image"

personally I appreciate it when tests read as a somewhat-normal sentence that can more or less be understood from reading the message

packages/components/src/LightBox/LightBox.test.tsx
```diff
@@ -1,135 +1,192 @@
...
+            caption: captionTwo,
+            url: "https://i.imgur.com/6Jcfgnp.jpg",
+          },
+        ]}
+        imageIndex={1}
+        onRequestClose={handleClose}
+      />,
     );
+
+    expect(queryByText(titleTwo)).toBeInstanceOf(HTMLElement);
```
ZakaryH, line 117: couple small things here, let's use `toBeInTheDocument()` rather than `toBeInstanceOf(HTMLElement)`

also, the `test` name here doesn't really match up with what we're actually checking. we're not checking/displaying the `imageIndex` we're seeing if the correct title was found. this one is a bit of nitpick, but I always appreciate when tests can be easily understood by their name!

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -209,16 +249,23 @@ export function LightBox({
...
 }
 
-function PreviousButton({ onClick }: NavButtonProps) {
+function PreviousButton({ onClick, mouseIsStationary }: NavButtonProps) {
+  const { mediumAndUp } = useBreakpoints();
+
   return (
-    <div className={styles.prev}>
+    <div
+      data-testid="prev-button-wrapper"
```
ZakaryH, line 266: I don't see any tests assertion against this test id. what's the reason for adding it? especially if we're only adding it to the previous button and not the next. 

if we must have one, we should follow the `ATL-{thing}` format for test ids please and thank you!

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -209,16 +249,23 @@ export function LightBox({
...
   }
 }
-
 interface NavButtonProps {
   readonly onClick: () => void;
+  readonly mouseIsStationary: boolean;
 }
 
-function PreviousButton({ onClick }: NavButtonProps) {
+function PreviousButton({ onClick, mouseIsStationary }: NavButtonProps) {
```
ZakaryH, line 261: not strictly required since it doesn't matter a ton because we're not exposing this `PreviousButton` component to anything else, but if we were I'd probably want to rename `mouseIsStationary` to `hideButton` or something like that. 

really just trying to separate the component and its concerns from its usage so that it could be more generic and reused easily, or even making it more resilient to changes if say we decide next week that there's another condition where the button gets hidden and it's no longer just about a stationary mouse.

packages/components/src/LightBox/LightBox.module.css
anagelberg: The new layering of the backgrounds and overlay is interfering with the clickability of the buttons. I could possibly try reworking the stacking context of that further but it was giving me some hard times. I'm struggling to understand what is an appropriate elevation token in this case in any case, can you point me in the right direction?

packages/components/src/LightBox/LightBox.module.css
ZakaryH: fair enough! 

ok if that's been design approved then I can live with that. it is a max-width after all, so my mobile comment doesn't really make sense.

packages/components/src/LightBox/LightBox.tsx
anagelberg: Good call, renamed the props here

packages/components/src/LightBox/LightBox.tsx
anagelberg: Removed this, I had forgot I added it 😅

packages/components/src/LightBox/LightBox.test.tsx
anagelberg: Addressed this and the other above readability comments from prior work. Thanks for suggestions for keeping it clean!

packages/components/src/LightBox/LightBox.module.css
anagelberg: Reworked to remove this extra wrapper 🙏

ZakaryH: ok so I'm pretty sure these look the same...

Padding + 2 divs             |  Margin + 1 div
:-------------------------:|:-------------------------:
image
  | 
image


the latter is achieved with a little bit less markup and some styles shifted around

in `LightBox.tsx` I've removed `imageArea` completely so that we only have one div wrapping `AnimatePresence` and our image

then in the styles I have

```
.slideWrapper {
  z-index: var(--elevation-base);
  height: 100%;
  max-width: 100%;
  position: relative;

  @media (--medium-screens-and-up) {
    /* nav button width + padding */
    margin: 0 calc(var(--space-largest) + var(--space-base));
  }
}

...

.captionWrapper {
  display: flex;
  z-index: var(--elevation-base);
  max-width: 600px;
  margin: var(--space-large) auto var(--space-larger);
  text-align: center;
  flex-direction: column;
  align-items: center;
  gap: var(--space-small);
}
```

where the main changes are that `slideWrapper` now has the relative position, and uses margin instead of padding. oh also, the margin bottom has been removed and instead is margin top of the same amount of the `captionWrapper`

the main benefits to this are that we don't have extra div nesting when we can do it with less. keeps things simpler for future layout changes, and now all our margin for the caption is on the caption element itself, though this did make me realize something that I'll address in a different comment since it's worth focusing on separate from this!

packages/components/src/LightBox/LightBox.module.css
anagelberg: I've been thinking on this, but I really think typically we'd want to target the children to achieve this type of layout? The only other way I can think of is adding in a third dummy child container with defined width to the mix, but that feels rather clunky. 

Is the concern that the elements inside could change? Would it be better to wrap them to apply needed styles to these components?

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -162,15 +188,31 @@ export function LightBox({
...
+              <NextButton
+                onClick={debouncedHandleNext}
+                hideButton={mouseIsStationary}
+              />
             </>
           )}
 
-          <div className={styles.toolbar}>
-            {images[currentImageIndex].caption}
+          <div className={styles.captionWrapper}>
```
ZakaryH, line 202: a few things in here:

**optional attributes**
since both `caption` and `title` are optional, it's entirely possible that we have neither. the only purpose of this section is to display those.

is there a reason we want to render this section at all when neither are present? 

here you can see how much room we're allocating to... nothing

image

vs with that gone, the image has a lot more room
image

**empty elements**
similarly if we don't have content for text elements we should simply not render them

I know we were rendering an empty span before 😅  but that was also not great

instead of rendering `<h4>""</h4>` when there's no `title` we should not render it at all. same goes for the empty `p` tag we get from `caption`

this circles back to my initial question of, do we really need to render this section at all when there is neither a caption nor a title? do we actually want this bottom space to be made available to the image instead?

ZakaryH: if the answer is yes, we'll want to find another of achieving this because having empty text elements is just not a great practice and will lead to a worse screen reader experience

packages/components/src/LightBox/LightBox.module.css
ZakaryH: ah yeah I see the problem

our `var(--color-surface--reverse)` resolves to an `hsl` value, which can accept an alpha value for opactiy, but we can't get it in there

I'll have to ask Chris if he can think of a way to use our token in that way

packages/components/src/LightBox/LightBox.tsx
anagelberg: Yes, this is specifically a design requirement that was outlined to me to hold this space as I had explained in that other height comment. Specifically if any image in the set has a title and/or caption, which is why the current implementation uses the `.some` check. 

I'll touch base with design about whether we can handle this a little differently though, re: accessibility concerns

packages/components/src/LightBox/LightBox.module.css
ZakaryH: fair enough! hmm I'm trying to decide which is the better one to keep

`text-align` is more what we're trying to do, with the Heading at least, it's just a bit odd that the Text will ignore it

`align-items` happens to work here, and doesn't conflict with Text

let's keep the `text-align` and get rid of `align-items` if that's the actual behavior we're after

packages/components/src/LightBox/LightBox.module.css
ZakaryH: ah sorry I missed this and I think it answers my question about the caption area being present even if it's not applicable to that particular image

hmmm ok then yeah if that's what we want, then I think a cleaner way to keep the space would be to give the caption section a min height that we think is good, rather than inserting empty characters combined with line heights

packages/components/src/LightBox/LightBox.module.css
anagelberg: Ah gotcha that makes sense! Removed it

packages/components/src/LightBox/LightBox.module.css
ZakaryH: yeah exactly if we decide that maybe instead we want this to be a h3, nothing has really changed about what we want out of the layout. this particular column should still be `grid-column: 2; justify-self:  center;`

however because we were being so specific, it won't work anymore and we have to update it.

so what would be optimal is separating the layout concern from the elements we are currently using to display.

normally I like to reduce the amount of divs/markup, but in this case I actually think that yeah wrapping it with a div that we can target to apply the layout improves it in a meaningful way

packages/components/src/LightBox/LightBox.module.css
ZakaryH: so I would recommend one of the higher elevation tokens, even though they are named something like `elevation--menu` and this isn't a menu

buuuut I tried this out with just `--elevation-base` and it seems to be working fine?

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -15,71 +15,125 @@
...
+  @media (--medium-screens-and-up) {
+    left: var(--space-larger);
+  }
+}
+
+.buttonHidden {
+  opacity: 0;
+  transition: opacity var(--timing-base) ease-in;
+
+  &:hover {
```
ZakaryH, line 53:
```suggestion
  &:hover,
  &:focus-within {
```

so that if you are using your keyboard on these buttons directly, it stays showing too

I know you can use your arrow keys to do the same thing, but since they're focusable, we might as well make them work

packages/components/src/LightBox/LightBox.tsx
ZakaryH: I've got my replies split between this comment and another but I'll repeat it here just in case

we can still have the spacing if it's a requirement. it's just how we're achieving it that would need to change. no empty text elements, and instead we use a min-height or flex-basis to provide that spacing.

on a related note, we should have a conditional above `captionWrapper` that prevents it from rendering at all if none of our images have titles nor captions

for a small optimization we could iterate over all of the images a single time and store the answers to teh questions of "do any have a title" "do any have a caption"

then use those on their own, and combined to determine what gets rendered down in this section

I can put together a quick snippet tomorrow to demonstrate what I'm thinking, just signing off for the day now but wanted to comment before I forgot!

anagelberg: I gotcha, this makes sense! I'll rework this a bit further, thanks so much for all the feedback Zakary! 🙏

packages/components/src/LightBox/LightBox.module.css
anagelberg: Added these to resolve it 👍

packages/components/src/LightBox/LightBox.module.css
anagelberg: Ha sorry, I forgot why I added these with all the changes and was mistaken in my last reply. it's actually because of [this issue](https://github.com/GetJobber/atlantis/pull/2337#pullrequestreview-2581772489), discussed above. Since the buttons can hide (and usually are in mobile) and it's an edge edge case, we aren't going to worry too much about the overlap right now. However, the buttons should at least be clickable, which was mainly an issue of the stacking, so that's why this was added. 

Looking at this again though, I think this may hide the image itself in even more extreme cases, which is probably not ideal. When we were having convos with Chris the main websites designer was OOO, I’m going to chat with her about this tomorrow to see if we want to handle this edge case differently, so the z-index issue may be mute. 

Otherwise will move to the `--elevation-menu` as you suggest :smile:

packages/components/src/LightBox/LightBox.tsx
ZakaryH: no problemo! also my note about optimizing isn't correct. `some` is fine, it would just be storing those results to use with their individual cases, and then a combination to see if we render the entire footer/caption section at all.

one other thought I had was around the whole "2-3 lines of caption" bit. we've gotta keep in mind mobile screen sizes will have a lot less room, and so something that has 2 lines of text on a big screen could be 4-5 lines on mobile. planning around lines of text can be tricky if there's no maximum/truncation.

anagelberg: I've chatted with @kerry-mac about this. We're going to remove this space since the buttons are now fixed in place 👍

packages/components/src/LightBox/LightBox.module.css
anagelberg: We decided to set a max height on the text container. If the text goes beyond the size for longer captions, it will scroll. This resolves the issue here regarding the overlap with the buttons, so I've set these back to `--elevation-base`

packages/components/src/LightBox/LightBox.module.css
anagelberg: We've decided to leave these arrows centered on the viewport for now, so this will be less arbitrary.

packages/components/src/LightBox/LightBox.module.css
anagelberg: Backing up on design, we've decided we don't need this with different handling of buttons, so I've removed these values

ZakaryH:
@anagelberg a one line change to move a function call outside the conditional, and then a merge conflict in Gallery.test

aside from that, looks good to me!

ZakaryH (approved):
LGTM

PR #2344, opened 2025-01-29
docs: remove the feedback banner and its implementation

chris-at-jobber:
## Motivations

FeedbackBanner did not generate much feedback - we garnered more via adhoc consumer reachout.

In addition, the Banner itself was the source of some negative feedback.

So we're taking it out!

### Removed

- The FeedbackBanner
image

## Testing

Run the docs site locally or on CF, observe no FeedbackBanner

---

Random photo of Atlantis


packages/site/src/components/FeedbackBanner.tsx
```diff
@@ -1,83 +0,0 @@
...
-          />
-          <Button
-            type="tertiary"
-            variation="subtle"
-            onClick={() => {
-              localStorage.setItem("nolikeynewsite", "true");
-              window.location.href =
-                "/storybook/?path=/docs/introduction--docs";
-            }}
-            label="Back to old site"
```
taylorvnoj, line 77: I agree with removing the Banner, especially because it was centered and got in the way. 
My concern is that this is now in a sense, hard launching new docs site, no? I know we link to Storybook elsewhere but should we have a smaller "View Storybook" in the same place as what's currently on Storybook?

Screenshot 2025-01-29 at 12 58 17 PM

taylorvnoj (commented):
Agree with removing the large banner however now we're 'hard launching'? Is that the intention?

packages/site/src/components/FeedbackBanner.tsx
chris-at-jobber: No concerns with Cleo and Scott when I ran it by them, but does it help to add a small subtle link in the sidenav for now? I know we have some other tickets where we'd look to add more direct links from component to storybook in the component preview panel but until then this give us a bit of both?

image

taylorvnoj: We could - I think my main FUD is, what now constitutes the Hard Launch? Because there are ticket's we've indicated are meant to be pushed with the Hard Launch like removing duplicated material from Docs pages that are in the Implement Tab. Are we ok to now remove that duplication?

scotttjob: I think the little link at the bottom is cool! We can also revisit this decision in a day or two if we're getting feedback of "where is storybook"

In terms of the cross-over/line for the launch state, we'll just consider everything else v2 moving forward?

chris-at-jobber: @taylorvnoj really like that improvement, I pushed an inline version just to validate for myself but will extract to the stylesheet

taylorvnoj (approved):
Thank you Chris! Feel free to merge ahead, if you push the cleanup after my approval!

chris-at-jobber:
Thanks @taylorvnoj ! Scott and I will also deal with any backlog adjustments we need as a result of this 🫡

packages/site/src/layout/NavMenu.module.css
```diff
@@ -163,4 +163,11 @@
 
 :root[data-theme="dark"] .disclosureNavItem button {
   mix-blend-mode: screen;
+}
+
+a.navFooterLink {
```
chris-at-jobber, line 168: extra specificity to ensure does not inherit default link interactive color

PR #2349, opened 2025-01-30
fix(components-native): Add missing onPress prop type to InputPressable suffix (it's already accepted)

nad182:
## Motivations

`InputPressable` uses `InputFieldWrapper` internally, which accepts a `suffix` prop. While `InputFieldWrapper`'s `suffix` accepts `icon`, `label` and `onPress`, `InputPressable` was missing `onPress` from type definitions. Passing that property still worked, but TS was throwing an error due to missing prop.

### Added

- `onPress` prop as a part of `suffix` type in `InputPressable`.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/InputPressable/Mobile.stories.tsx
+++ b/docs/components/InputPressable/Mobile.stories.tsx
@@ -1,8 +1,9 @@
 import React, { useState } from "react";
-import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { ComponentStory } from "@storybook/react";
+import type { Meta, StoryObj } from "@storybook/react";
 import { InputPressable } from "@jobber/components-native";
 
-export default {
+const meta: Meta<typeof InputPressable> = {
   title: "Components/Forms and Inputs/InputPressable/Mobile",
   component: InputPressable,
   parameters: {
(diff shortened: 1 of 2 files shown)


TylerOlthuizen:
Just tested this on mobile, working as expected!

TylerOlthuizen:
> Just tested this on mobile, working as expected!

https://github.com/GetJobber/jobber-mobile/pull/10062

TylerOlthuizen:
@MichaelParadis wouldn't mind getting this in to unblock my PR. We would like to make todays code freeze if possible.

https://github.com/GetJobber/jobber-mobile/pull/10062

ZakaryH:
this is a little weird

it's true that when you do `suffix={{icon: "timer", onPress: myOnPress }}`

that we'll go down a path that will pass the `suffix.onPress` to `SuffixIcon`

https://github.com/GetJobber/atlantis/blob/master/packages/components-native/src/InputFieldWrapper/InputFieldWrapper.tsx#L227-L256

the thing is though that both `label` and `icon` are both optional, so you could provide `{label: "blah", onPress: myOnPress }` and it won't do anything. I'm wondering if we should use a TS XOR or just do it with TS itself so that it will tell you only that combination is valid.

fighting with my IDE at the moment, but wanna see what the types are telling me about this on `master` and this `branch`

ZakaryH (approved):
with the latest type changes I am able to provide the suffix itself an `onPress`

additionally Typescript properly informs me that `{label: "something", onPress: someFunction }` is not valid, nor is `onPress` alone. it requires an `icon` to be present which matches the actual behavior

nice improvement!

PR #2350, opened 2025-01-31
feat(components-native): Adding Unsafe style to mobile Text and Button [JOB-112128]

CoderDake:
## Motivations

Jobber Copilot has some unique styling needs. With @taylorvnoj 's help I was pointed to Unsafe styling as a solution.

### Added

- Add unsafe styling to Text
- Add unsafe stylling to Button

## Testing

- [ ] Apply an unsafe style change to Text
- [ ] Apply an unsafe style change to Button
- [ ] Ensure that Buttons without unsafe styles are unchanged
- [ ] Ensure that Text without unsafe styles are unchanged.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components-native/src/Button/Button.tsx
+++ b/packages/components-native/src/Button/Button.tsx
@@ -85,7 +85,7 @@ interface CommonButtonProps {
 }
 
 export interface CommonButtonStyleOverride {
-  readonly borderColor?: string; // TODO: get proper type
+  readonly borderColor?: typeof tokens;
 }
 
 interface LabelButton extends CommonButtonProps {
@@ -129,17 +129,10 @@ export function Button({
     fullHeight && styles.fullHeight,
   ];
 
-  console.log("tokens", tokens);
-
   if (styleOverrides?.borderColor) {
-    if (styleOverrides.borderColor in tokens) {
-      const borderColor =
-        tokens[styleOverrides.borderColor as keyof typeof tokens];
-      buttonStyle.push({ borderColor: borderColor });
-    } else {
-      throw new Error(`Invalid borderColor key: ${styleOverrides.borderColor}`);
-    }
+    buttonStyle.push({ borderColor: styleOverrides.borderColor });
   }
+
   // attempts to use Pressable caused problems.  When a ScrollView contained
   // an InputText that was focused, it required two presses to activate the
   // Pressable.  Using a TouchableHighlight made things register correctly
--- a/packages/components-native/src/Button/Button.tsx
+++ b/packages/components-native/src/Button/Button.tsx
@@ -246,7 +246,7 @@ function getContentStyles(
   icon: IconNames | undefined,
 ) {
   if (label && !icon) {
-    return [];
+    return undefined;
   }
 
   return [
--- a/packages/components-native/src/Button/Button.tsx
+++ b/packages/components-native/src/Button/Button.tsx
@@ -132,6 +132,10 @@ export function Button({
     fullHeight && styles.fullHeight,
   ];
 
+  if (UNSAFE_style?.container) {
+    buttonStyle.push(UNSAFE_style.container);
+  }
+
   // attempts to use Pressable caused problems.  When a ScrollView contained
   // an InputText that was focused, it required two presses to activate the
   // Pressable.  Using a TouchableHighlight made things register correctly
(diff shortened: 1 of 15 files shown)


packages/components-native/src/Button/Button.tsx
```diff
@@ -119,6 +139,10 @@ export function Button({
     fullHeight && styles.fullHeight,
   ];
 
+  if (UNSAFE_style?.container) {
+    buttonStyle.push(UNSAFE_style.container);
+  }
```
taylorvnoj, line 144: This, as well as the ternary in the other Views seems like more than we should need to do. Was the choice to do that because the tests were failing? I'm going to have to look into this a bit more

taylorvnoj: Ok I played around with it. We can remove this `if` statement.

If I didn't have this `if` statement the tests would fail horribly. The issue is about style precedence. 
Removing this `if` statement and updating the wrapping View to be:
```
<View style={[...buttonStyle, UNSAFE_style?.container]}>
```
ensures UNSAFE is applied last and properly overrides defaults.

The rest of the Views don't need the ternary and I'll add suggestions there.

taylorvnoj: We only have to spread `buttonStyle` because it's an array that's built up from multiple sources - the remaining views just combine a single base style with a single override

packages/components-native/src/Button/Button.tsx
```diff
@@ -143,17 +167,36 @@ export function Button({
...
-            <View style={styles.iconStyle}>
+            <View
+              style={[
+                styles.iconStyle,
+                ...(UNSAFE_style?.iconContainer
+                  ? [UNSAFE_style.iconContainer]
+                  : []),
+              ]}
+              testID="iconContainer"
+            >
```
taylorvnoj, line 183:
```suggestion
             <View
              style={[styles.iconStyle, UNSAFE_style?.iconContainer]}
              testID="iconContainer"
            >
```

packages/components-native/src/Button/Button.tsx
```diff
@@ -143,17 +167,36 @@ export function Button({
...
-            <View style={styles.labelStyle}>
+            <View
+              style={[
+                styles.labelStyle,
+                ...(UNSAFE_style?.actionLabelContainer
+                  ? [UNSAFE_style.actionLabelContainer]
+                  : []),
+              ]}
+              testID="actionLabelContainer"
+            >
```
taylorvnoj, line 199:
```suggestion
 <View
              style={[styles.labelStyle, UNSAFE_style?.actionLabelContainer]}
              testID="actionLabelContainer"
            >
```

packages/components-native/src/Button/Button.tsx
```diff
@@ -143,17 +167,36 @@ export function Button({
     >
       <View style={buttonStyle}>
         {loading && <InternalButtonLoading variation={variation} type={type} />}
-        <View style={getContentStyles(label, icon)}>
+        <View
+          style={getContentStyles(label, icon, UNSAFE_style?.contentContainer)}
+          testID="contentContainer"
+        >
```
taylorvnoj, line 173: We can also remove the logic added below in `getContentStyles` and revert this back to

```suggestion
<View
          style={[
            getContentStyles(label, icon),
            UNSAFE_style?.contentContainer,
          ]}
          testID="contentContainer"
        >
```

taylorvnoj (changes requested):
Hey @CoderDake - apologies it took longer than expected to get a review back.

The logic in `Button` felt a bit clunky to me but I can see that tests were probably being quite difficult. Instead of  double pushing with the `if` statement, we should be able to spread `buttonStyle`.

There's quite a bit here, I'm sorry! If you want to pair at any point let me know

taylorvnoj (approved):
This LGTM!

- Pre-release is in prod here https://github.com/GetJobber/jobber-mobile/pull/10110; CI is happy ✔️ 
- @CoderDake & I ran through the pre-release branch on iOS locally; Buttons and Text components still looked as they should and UNSAFE worked when tested ✔️ 
- I tested the android build; Buttons and Text look as they should ✔️ 

Really great job on this @CoderDake - this is a huge contribution! I'm going to pass this by the team and if I don't hear anything, I'll merge this for you tomorrow 👍🏻

PR #2360, opened 2025-02-05
fix(components): Put Popover inside a portal 

Aiden-Brine:
## Motivations

A user of Atlantis [pointed out](https://getjobber.slack.com/archives/C03BJHV3PMG/p1738701498519799) that when they use Popover inside a container that has overflow: hidden their popover is cut off. This is because the popover is not inside a portal.

### Added
A portal wrapping Popover

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

Screenshot 2025-02-04 at 4 57 59 PM

### Security

- 

## Testing

Verify that a couple of other portals are behaving as expected and have not been modified. You can grab the pre-release and test out [the endpoint](http://localhost.test:3000/business_health) where the problem was reported. Kyle said I would need a feature flag turned on to view this but I didn't end up needing to. Mentioning in case the reviewer runs into an error on that endpoint.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Popover/Popover.tsx
+++ b/packages/components/src/Popover/Popover.tsx
@@ -107,11 +107,7 @@ export function Popover({
     </div>
   );
 
-  return <>{open && <PopoverPortal>{popoverContent}</PopoverPortal>}</>;
-}
-
-function PopoverPortal({ children }: { children: React.ReactNode }) {
-  return ReactDOM.createPortal(children, document.body);
+  return <>{open && ReactDOM.createPortal(popoverContent, document.body)}</>;
 }
 
 function buildModifiers(arrowElement: HTMLElement | undefined | null) {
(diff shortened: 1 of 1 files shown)


packages/components/src/Popover/Popover.tsx
```diff
@@ -11,28 +12,23 @@ export interface PopoverProps {
    * and passed as an attachTo prop in order for the Popover to function properly
    */
   readonly attachTo: Element | React.RefObject<Element | null>;
-
   /**
    * Popover content.
    */
   readonly children: React.ReactNode;
-
```
edison-cy-yang, line 19: were these line break removals from eslint?

Aiden-Brine: I thought it was Prettier but I realize that doesn't really make sense. I wonder if it was Cursor? Regardless, added back now

edison-cy-yang (approved):
Tested and verified that the popover is on a portal outside of the main DOM. Looks good!

PR #2362, opened 2025-02-05
feat(components): Add thumbnails and change animations in lightbox component

anagelberg:
## Motivations

This is the second of 2 planned updates to the lightbox component to enhance its UX. 

## Changes

https://github.com/user-attachments/assets/e1826fba-4f77-4ff6-b594-228fcd6bb8c0

https://github.com/user-attachments/assets/bf95e83c-7356-4c30-a927-f6ac8e0e84e0

### Added

- Clickable preview thumbnails are now present within the lightbox
- Adds optional alt text field on images in the the accepted images array prop to improve accessibility 

### Changed

- Changes the main animation of switching the images 

Changes can be

---

Random photo of Atlantis


ZakaryH:
something we'll need to account for is the fact that since we've added a scroll-x to the thumbnails, if your OS settings display the scrollbar this can happen and it's not immediately clear the whitespace is actually a horizontal scroll bar

image

it also shifts the layout up a bit where the top of the border is cut off

ZakaryH:
since `Gallery` uses `Lightbox` internally, the UX/UI gets a little interesting where this is `Gallery`
image

then I click on an image and get
image

where I now have basically that same piece of UI available to me. I'm honestly not sure how I feel about that. it might be a little odd at worst or it might be totally fine.

I think I'll defer to @chris-at-jobber on this one

anagelberg:
> something we'll need to account for is the fact that since we've added a scroll-x to the thumbnails, if your OS settings display the scrollbar this can happen and it's not immediately clear the whitespace is actually a horizontal scroll bar
> 
> image
> 
> it also shifts the layout up a bit where the top of the border is cut off

@ZakaryH I don't follow how to reproduce this issue, can you explain?

ZakaryH:
@anagelberg for sure, so I use a non-mac mouse and keyboard which apparently triggers the scrollbars to show up pretty much all the time

if you're just using the trackpad or mac products I don't think it shows up. if that's the case you can go into your OS settings and change this to "always" and it should appear like the screenshot

image

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -28,9 +28,8 @@
 .next,
 .prev {
   position: absolute;
-  top: 50%;
+  top: 40%;
```
ZakaryH, line 31: what's the rationale for this change?

it's one million percent me being a little eccentric but I definitely noticed it's off centre and it bothers me way more than it reasonably should 😅

anagelberg:
> @anagelberg for sure, so I use a non-mac mouse and keyboard which apparently triggers the scrollbars to show up pretty much all the time
> 
> if you're just using the trackpad or mac products I don't think it shows up. if that's the case you can go into your OS settings and change this to "always" and it should appear like the screenshot

@ZakaryH  Ah gotcha, good callout. In this case do you think it's acceptable to just override system default scrollbar behavior (make the background transparent, set the height, etc?) or would you want to address that another way?

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -109,9 +104,11 @@
   position: relative;
   height: 100%;
   max-width: 100%;
-  margin-bottom: var(--space-large);
+  margin: 0 var(--space-base) var(--space-large);
   @media (--medium-screens-and-up) {
-    margin: 0 calc(var(--space-largest) + var(--space-base)) var(--space-large);
+    margin: 0 /* nav button width + inner padding + outer margin */
+      calc(var(--space-largest) + var(--space-base) + var(--space-large))
+      var(--space-large);
```
ZakaryH, line 111: this calculation feels a bit fragile, or overly complicate to me. wondering if we can't find a simpler solution with either flexbox or percentages.

what's the end result we're trying to achieve with this horizontal margin?

packages/components/src/LightBox/LightBox.module.css
anagelberg: It's because we wanted the arrows to be close to centered on the image itself, but this is an imperfect approximation since we are fixing the arrows to the outer container to prevent bouncing between images with different sized captions. If it's at 50%, they'll sit lower: 

Screenshot 2025-02-12 at 12 34 03 PM


So this is an adjustment to give approximate space for thumbnails

ZakaryH:
> Ah gotcha, good callout. In this case do you think it's acceptable to just override system default scrollbar behavior (make the background transparent, set the height, etc?) or would you want to address that another way?

kinda depends on what we want the experience to be.

when there's nothing to scroll, then yes I think it would make sense to try and hide the scrollbar. it adds nothing and only makes the layout look worse.

when there is something to scroll, it still doesn't look great but it at least serves the purpose of indicating that you can scroll. whether or not that's necessary is debatable. personally I'd like to hide it even then, and hope that people see the edge of the final thumbnail which would suggest it can be scrolled.

anagelberg:
> > Ah gotcha, good callout. In this case do you think it's acceptable to just override system default scrollbar behavior (make the background transparent, set the height, etc?) or would you want to address that another way?
> 
> kinda depends on what we want the experience to be.
> 
> when there's nothing to scroll, then yes I think it would make sense to try and hide the scrollbar. it adds nothing and only makes the layout look worse.
> 
> when there is something to scroll, it still doesn't look great but it at least serves the purpose of indicating that you can scroll. whether or not that's necessary is debatable. personally I'd like to hide it even then, and hope that people see the edge of the final thumbnail which would suggest it can be scrolled.
> 
> edit: if I really think about it, why do we want it to be scrollable? I would argue that's a pretty uncommon UX and I wouldn't expect to do horizontal scrolling in a section like that. how would we expect people to interact with that scrollable section when they have more images than can fit in the screen's width?

I'm gonna loop @kerry-mac in on this but from my understanding we want it to be scrollable so they have the ability to quickly browse through images and select the one they want, and I think what you're saying that they should get the idea it's scrollable from seeing a portion of the image. I think we're okay to make it transparent and match the intended behavior (where we only really see the scrollbar while currently scrolling), which gives them an indication of where they are in the list of images.

packages/components/src/LightBox/LightBox.module.css
ZakaryH: gotcha. personally I would be tempted to play around with the caption experience. maybe we always leave space for them, or we find a different way to display them that doesn't alter the available space.

I'm certainly not going to block on this, especially since I assume it was already approved but yeah I'd say that's the crux of the issue.

if you look at other gallery/lightbox type components, even if they don't have captions the space is preserved for them preventing a shift in the layout, and allowing the buttons to be centered

image
image

I'll simply leave my feedback as the fact that it definitely feels imbalanced being off center

anagelberg: We went back and forth all over trying to get a good design decision to the captions thing on the last PR, where we left space, but then it was an issue if the captions were larger. However, we did end up landing on having an overflow: scroll effect on the captions themselves. Now that that's there, I think leaving that same steady space and keeping the arrows centered is likely possible. I'll take one more look at this🙏

ZakaryH:
>I think we're okay to make it transparent and match the intended behavior (where we only really see the scrollbar while currently scrolling), which gives them an indication of where they are in the list of images.

so for mobile, yeah I think that makes sense since the scroll will support dragging to move. same for web with a track pad and swiping.

things get a little less intuitive on web with a mouse & scrollwheel. it still works, you just have to know that it's scrollable and that to do horizontal scroll you must hold shift and then use your wheel.

anagelberg:
> so for mobile, yeah I think that makes sense since the scroll will support dragging to move. same for web with a track pad and swiping.
> 
> things get a little less intuitive on web with a mouse & scrollwheel. it still works, you just have to know that it's scrollable and that to do horizontal scroll you must hold shift and then use your wheel.

I was thinking we'd still have the thumb as an indicator when scrolling was available, as is without the settings set to "always on" so that provides clicking ability as well, I'm unsure if you're thinking differently? I think we should go with something subtler with that thumb (either default, or maybe custom) but for example it would look something like this, just so the background itself is transparent : 



https://github.com/user-attachments/assets/fa08c6d1-81b0-4348-92cc-673b1de927bc



Are you saying that's unclear it's a scrollbar though ?

packages/components/src/LightBox/LightBox.module.css
ZakaryH: I disagree.

we can basically never guarantee that `inner-padding` value outside of very specific image dimensions and screen sizes.

image


image

here are two examples of that gap being something completely different. with that in mind, I'd suggest we don't try to focus on that at all.

instead, we need to figure out the consistent pieces that we do want to guarantee.

with a layout like this
image

we'd define how much the purple space can take up, and the minimum width we would let them be. then, we'd absolutely position the buttons relative to those spaces, but specifying from the left hand side.

if I'm understanding the desired layout, that would get us the same thing with no need to do any `calc` on values that are subject to change. the buttons will always be a set distance from the edges of the screen, and an arbitrary distance from image edge with a guarantee of a minimum distance so as to not overlap

ZakaryH:
>Are you saying that's unclear it's a scrollbar though ?

that's pretty much it yeah. I'm imagining the scrollbar being hidden until I scroll. with a mouse, I can't scroll it unless I hold shift and scroll wheel it. would we expect people to know to even try that?

again, I do think it works nicely for mobile and trackpads. it's just mice that would have a worse time, and maybe that's OK if we don't think that's a problem worth solving!

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -121,16 +118,81 @@
...
+  align-items: center;
+  gap: var(--space-small);
+  flex-shrink: 0;
+}
+
+.thumbnailImage {
+  width: 100%;
+  height: 100%;
+  border-radius: var(--radius-small);
+  opacity: 0.8;
```
ZakaryH, line 158: opacity overlay token please!

image


unless there's a particular reason we don't want to have that matching

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -121,16 +118,81 @@
...
+  object-fit: cover;
+}
+
+.thumbnail {
+  width: var(--space-extravagant);
+  height: var(--space-extravagant);
+  border-radius: var(--radius-small);
+  background-color: var(--color-text--reverse);
```
ZakaryH, line 166: hmmm using a text color as the background is a bit unconventional. 

what's the color we're after here?

packages/components/src/LightBox/LightBox.module.css
anagelberg: I'm going to take a look at getting this back on flex anyway back on your other comment, but the way I understand the design, we do want to define the _maximum_ image area specifically, bound by the buttons as I've defined (maybe double check the design to see what I mean). 

The way I've set this up, `.imageArea` corresponds to that space, shown in the purple box here: 

Screenshot 2025-02-12 at 1 48 37 PM
Screenshot 2025-02-12 at 1 49 05 PM


In your demonstration, this wouldn't be bound correctly by the arrows as it could take up the whole width of the screen? 

While the image is contained to preserve aspect ratio, I think we still need that maximum defined as I have to match design

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -121,16 +118,81 @@
...
+
+  &:hover {
+    cursor: pointer;
+  }
+}
+
+.thumbnail.selected {
+  margin: 0 var(--space-smaller);
+  padding: var(--space-smallest);
+  border: 2px solid var(--color-text--reverse);
```
ZakaryH, line 180: similar thought here. these are the first two times we'd be using the text color for non text. I'm hesitant to do it without a compelling reason.

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -121,16 +118,81 @@
...
+  width: 100%;
+  height: 100%;
+  border-radius: var(--radius-small);
+  opacity: 0.8;
+  object-fit: cover;
+}
+
+.thumbnail {
+  width: var(--space-extravagant);
+  height: var(--space-extravagant);
```
ZakaryH, line 164: this might be a place where we actually don't want to use tokens

is there a specific dimension we want the thumbnails to be?

or phrased differently, if tomorrow we decide to reduce `--space-extravagant` to be something smaller, is it OK for the thumbnails to shrink too, or would we want them to remain untouched?

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -61,27 +63,21 @@ const swipePower = (offset: number, velocity: number) => {
...
-      opacity: 0,
+      x: direction < 0 ? "150%" : "-150%",
     };
   },
 };
 
 const imageTransition = {
-  x: { type: "spring", stiffness: 300, damping: 30 },
-  opacity: { duration: 0.2 },
+  x: { duration: 0.65, ease: [0.42, 0, 0, 1.03] },
```
ZakaryH, line 80: question about these values. the duration isn't using any previously established timing token, and the easing numbers are rather specific.

any reason for these exact values or might we be able to use a timing token + some cleaner numbers?

image

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -135,6 +133,19 @@ export function LightBox({
     togglePrintStyles(open);
   }
 
+  useEffect(() => {
+    if (thumbnailBarRef.current) {
+      const selectedThumbnail = thumbnailBarRef.current.children[
```
ZakaryH, line 138: hmmm I'm wondering if we really need the thumbnail bar, or if we can simply set the ref on the element we wanna scroll to directly instead. I'll give this a try.

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -18,6 +19,7 @@ import { AtlantisThemeContextProvider } from "../AtlantisThemeContext";
 interface PresentedImage {
   title?: string;
   caption?: string;
+  alt?: string;
```
ZakaryH, line 22: arguably this should actually be mandatory to comply with accessibility requirements

lemme double check if we're being good about that on Gallery or not

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -208,14 +224,43 @@ export function LightBox({
...
                 )}
               </AtlantisThemeContextProvider>
             </div>
           )}
+
+          {images.length > 1 && (
+            <div
+              className={styles.thumbnailBar}
+              ref={thumbnailBarRef}
+              data-testid="thumbnail-bar"
```
ZakaryH, line 244:
```suggestion
              data-testid="ATL-Thumbnail-Bar"
```

packages/components/src/LightBox/LightBox.tsx
anagelberg: We worried about breaking any other parts of Jobber that use this?

packages/hooks/src/index.ts
```diff
@@ -1,4 +1,3 @@
-// Export all hooks
```
ZakaryH, line 1: curious about this deletion

packages/components/src/LightBox/LightBox.tsx
anagelberg: @kerry-mac defined these, thoughts on changing the animation here? (the sliding box one)

packages/components/src/LightBox/LightBox.tsx
ZakaryH:
```
  const selectedThumbnailRef = useRef<HTMLDivElement>(null);

...
  useEffect(() => {
    selectedThumbnailRef?.current?.scrollIntoView({
      behavior: "smooth",
      block: "nearest",
      inline: "center",
    });
  }, [currentImageIndex]);

....
            <div className={styles.thumbnailBar} data-testid="thumbnail-bar">
              {images.map((image, index) => (
                <div
                  key={index}
                  className={classNames(styles.thumbnail, {
                    [styles.selected]: index === currentImageIndex,
                  })}
                  onClick={() => handleThumbnailClick(index)}
                  ref={
                    index === currentImageIndex ? selectedThumbnailRef : null
                  }
                >
```

I like this a bit more because it's a bit less code, and doesn't require any type casting

packages/components/src/LightBox/LightBox.tsx
ZakaryH: that's a fair point, we'd need to do some change management first and that's a bit much for right now.

looking at Gallery it's required in a very roundabout way since we use `alt={name}` and a file must have a name

this is fine to leave as optional and I'll make a note that we should be better about this with our image-related components

packages/components/src/LightBox/LightBox.module.css
ZakaryH: >In your demonstration, this wouldn't be bound correctly by the arrows as it could take up the whole width of the screen?

I don't believe it would be but that reminds me, you'll wanna pull in `master` into this branch because I found a bug last week that I fixed with a quick change

we had to change from `object-fit: contain` so as to not distort smaller images from being stretched.

with that change, an image will never be larger than what it actually is.

I think this would be easier for me to code and be able to definitively show with a demo, but I'm quite confident there's a way to do this that would free us up from doing margin addition in the css module.

I'd have to fiddle with the exact implementation details, but the gist of it is that those buttons would be contained in invisible containers that would be responsible for their dimensions. then the main image area could be responsible for its own dimensions. combining the two, they just occupy the space as instructed, and one doesn't need to be accounting for the spacing of the other.

packages/components/src/LightBox/LightBox.module.css
chris-at-jobber: Might be closer in the semantic ballpark to use `color--focus` here, since (if I recall correctly) we're wrapping all of this UI in `dark` theme.

image

image

packages/components/src/LightBox/LightBox.module.css
anagelberg: I’m just really struggling to understand how you would be able to bind them to a container considering the buttons are absolutely positioned in the parent lightbox container. Especially in a way that doesn't make it yet more confusing... like adding empty containers to bound it seems even more complicated to me, but maybe it's more readable anyway. 

I get you though, this is not ideal, I really didn’t want to use `position:absolute` for these buttons to begin with for that reason and have already switched it back and forth several times. I am gonna try again, just wondering if we'll ever get this right 😅

ZakaryH: fair enough. I don't have a solution ready to offer you right now, so I can definitely live with this for now and circle back on it when I have some more time to play around with it.

feel free to move on from this topic, I don't consider it a blocker for this PR.

packages/hooks/src/index.ts
anagelberg: Oops, this was accidental, I've reverted this

packages/components/src/LightBox/LightBox.tsx
anagelberg: Added this! What is the "ATL" prefix mean?

packages/components/src/LightBox/LightBox.tsx
anagelberg: Very nice and works well, thanks!!

packages/components/src/LightBox/LightBox.module.css
anagelberg: Good point, I've just set these to 64px

packages/components/src/LightBox/LightBox.module.css
anagelberg: This is outside the dark mode context at this point and we just wanted white. Not sure why I chose the text value to begin with but I've just switched these to use --color-surface now for both cases 😅

packages/components/src/LightBox/LightBox.module.css
```diff
@@ -121,16 +118,97 @@
...
+  margin-bottom: var(--space-small);
+  padding-bottom: var(--space-small);
+  overflow-x: scroll;
+  justify-content: safe center;
+  align-items: center;
+  gap: var(--space-small);
+  flex-shrink: 0;
+
+  scrollbar-width: thin;
+  scrollbar-color: var(--color-interactive--subtle) transparent;
```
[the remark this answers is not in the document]
ZakaryH: looks good on everything except Safari 😭 

image

that selector is supposed to be supported though. I know we had to do something like this not too long ago, lemme see what we did for that to get it working nicely.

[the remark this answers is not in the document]
ZakaryH: tried it on Edge/windows with Browserstack. looks good there too. interesting that the blur kinda stops near the edges though.

image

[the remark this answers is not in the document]
ZakaryH: I'm very confused. applying the same sort of styles to `body ::webkit-scrollbar` and the `thumb` it shows up as expected and I can style it. this obviously has the same scrollbar show settings since that's at the OS level.
image

why on earth is this not working on the lightbox....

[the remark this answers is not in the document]
ZakaryH: alright I haven't been able to figure this out

unless we think it's a blocker I'm fine with merging and trying to figure it out later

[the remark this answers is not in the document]
anagelberg: Thanks for taking a look Zak! I haven't been able to figure this out either and I think this is edge case enough that if youre okay with it, so am I! Also @kerry-mac gave unofficial approval over slack for the color and style, so we're good here I think!

packages/components/src/LightBox/LightBox.module.css
anagelberg: I appreciate this, thank you! Notably how we handle the text is possibly subject to change as well, so I think for now I'd like to leave this as we may be able to handle the arrows better at that time more easily/clearly as well 🙏

ZakaryH (approved):
I can't explain the horizontal scrollbar behavior

3 different people are seeing 3 different appearances of the same scrollbar styling on Safari. that said, 1 of them is what we want. 1 of them is OK and 1 of them is not great (mine)

image

image

image

despite that, I don't think this is a show stopper. so we'll merge this, and keep an eye on it/see if we can figure it out because I'm very interested in what is going on

PR #2363, opened 2025-02-06
fix(components): Prevent Stretching Small Images in Lightbox

ZakaryH:
## Motivations

when you view a small image with the new Lightbox changes, it gets stretched out and doesn't look excellent

rather than stretching the image, lets let it be as big as it is and then it doesn't look so bad for small images

### Added

- 

### Changed

we were using `object-fit: contain`

>[contain](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit#contain)
>The replaced content is scaled to maintain its [aspect ratio](https://developer.mozilla.org/en-US/docs/Glossary/Aspect_ratio) while fitting within the element's content box. The entire object is made to fill the box, while preserving its aspect ratio, so the object will be ["letterboxed"](https://en.wikipedia.org/wiki/Letterboxing_(filming)) or ["pillarboxed"](https://en.wikipedia.org/wiki/Pillarbox) if its aspect ratio does not match the aspect ratio of the box.

and instead we'll now use

>[scale-down](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit#scale-down)
>The content is sized as if none or contain were specified, whichever would result in a smaller concrete object size.

which will never stretch the image to be bigger than it is

image

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/LightBox/LightBox.module.css
+++ b/packages/components/src/LightBox/LightBox.module.css
@@ -102,7 +102,7 @@
   position: absolute;
   width: 100%;
   height: 100%;
-  object-fit: contain;
+  object-fit: none;
 }
 
 .imageArea {
--- a/packages/components/src/LightBox/LightBox.module.css
+++ b/packages/components/src/LightBox/LightBox.module.css
@@ -102,7 +102,7 @@
   position: absolute;
   width: 100%;
   height: 100%;
-  object-fit: none;
+  object-fit: scale-down;
 }
 
 .imageArea {


ZakaryH:
on current master

**small image**
image

**larger image**
image

**portrait image**
image

mobile

image
image
image

chris-at-jobber (approved):
Thank you for flagging and fixing @ZakaryH !

PR #2367, opened 2025-02-10
chore: Update generate command when creating a new component

taylorvnoj:
## Motivations

Running `npm run generate` from the root creates the folders/files needed to get started when adding a new component to Atlantis. For example it creates the `packages/components` folder as well as the `stories` files.  This PR now makes it so `npm run generate` also builds the `packages/site/src/content` needed for the new docs site.

### Added

- `buildSiteContentIndex` which Scott actually made previously, but now we're bringing it back to execute the `site` templates on `npm run generate`
- an `index.tsx` template to help components be displayed on new docs site
- `{{name}}Notes.mdx` template, similar to the `stories` template but for the Implement tab

### Changed

- moved 2 sections out of the `stories` template, since _Platform considerations_ and _Notes_ are now in Implement Tab territory

## Testing

Run `npm run generate` from the atlantis root and follow the prompts. 
Name your component and generate web/mobile/both separately to make sure the appropriate folders are being created (as if someone was adding a new component to the DS)

---

Random photo of Atlantis


packages/generators/templates/docs/{{name}}.stories.{{mdx}}
```diff
@@ -47,13 +47,6 @@ The {{name}} will... _Add a brief description of the design and usage of the com
   Is there any related resource (internal or external) to this component?
 -->
 
-## Platform considerations
```
taylorvnoj, line 50: This and the Notes section are in the Implement tab and therefore, taken out of the stories template.

edison-cy-yang (commented):
Tested generating a component and everything worked well!

One thing to note, it would be great if we could add `{{name}}.props.json` or `{{name}}.props-mobile.json` to the site template as well

taylorvnoj:
@edison-cy-yang I see where you're coming from, but those props.json files should be generated after the component is built and props already exist. The generate command I've updated here is before a component exists (we have another `npm run generate` command that runs within `packages/site` to create those props files automatically). 

My fear is that if we add props files to the generate command that happens from atlantis root, it will create duplicate files later on when generating those json files in /packages/site.

In this commit (1431ba6de4e35b4af403d9321dc2b0f5d70299fd) I updated the `/component-not-found` route to include running `npm run generate` from `/packages/site` to generate those prop files - hopefully that's a bit more clear?

I hope this makes sense 😅  let me know if it doesn't and we can chat

edison-cy-yang:
> @edison-cy-yang I see where you're coming from, but those props.json files should be generated after the component is built and props already exist. The generate command I've updated here is before a component exists (we have another `npm run generate` command that runs within `packages/site` to create those props files automatically).
> 
> My fear is that if we add props files to the generate command that happens from atlantis root, it will create duplicate files later on when generating those json files in /packages/site.
> 
> In this commit ([1431ba6](https://github.com/GetJobber/atlantis/commit/1431ba6de4e35b4af403d9321dc2b0f5d70299fd)) I updated the `/component-not-found` route to include running `npm run generate` from `/packages/site` to generate those prop files - hopefully that's a bit more clear?
> 
> I hope this makes sense 😅 let me know if it doesn't and we can chat

Ah I see how this is really being used. Sounds good!

edison-cy-yang (approved):

PR #2369, opened 2025-02-11
feat(components): Checkbox v2

ZakaryH:
## Motivations

We want to have a version of Checkbox that isn't coupled with RHF internally. The hidden connection that generates its own names isn't desirable when we want to use a Checkbox with RHF on a form. Additionally, we would like to be less tied to a single form library.

We can't just remove the coupling in our existing component because too much relies on that, and so to offer a de-coupled version we're going to offer a v2 that doesn't include the same setup.

unlike some of the other v2 inputs we've made recently I've opted to not add `ref` as a prop. I can't think of a compelling reason you'd need access to the ref. I haven't heard of anyone requesting it in all the time Checkbox has existed. not strictly opposed to it though, just don't see a reason currently.

### Added

A "rebuilt" Checkbox that has no RHF included

### Changed

no longer using XOR within types to determine if label is a string or child. instead, `label` accepts a string to use the default Text element and positions, or if more customization is needed markup can be passed in.

similarly, `description` also accepts a string or markup.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Checkbox/Checkbox.types.ts
+++ b/packages/components/src/Checkbox/Checkbox.types.ts
@@ -1,0 +1,79 @@
+import { ReactElement } from "react";
+import { XOR } from "ts-xor";
+
+export interface BaseCheckboxProps {
+  /**
+   * Determines if the checkbox is checked or not.
+   */
+  readonly checked?: boolean;
+
+  /**
+   * Initial checked value of the checkbox. Only use this when you need to
+   * pre-populate the checked attribute that is not controlled by the component's
+   * state. If a state is controlling it, use the `checked` prop instead.
+   */
+  readonly defaultChecked?: boolean;
+
+  /**
+   * Disables the checkbox.
+   */
+  readonly disabled?: boolean;
+
+  /**
+   * When `true` the checkbox to appears in indeterminate.
+   *
+   * @default false
+   */
+  readonly indeterminate?: boolean;
+
+  /**
+   * Checkbox input name
+   */
+  readonly name?: string;
+
+  /**
+   * Value of the checkbox.
+   */
+  readonly value?: string;
+
+  /**
+   * Further description of the label
+   */
+  readonly description?: string;
+
+  /**
+   * ID for the checkbox input
+   */
+  readonly id?: string;
+
+  onChange?(newValue: boolean): void;
+
+  onFocus?(): void;
+
+  onBlur?(): void;
+}
+
+interface CheckboxLabelProps extends BaseCheckboxProps {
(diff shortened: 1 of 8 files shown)


packages/components/src/Checkbox/Checkbox.types.ts
```diff
@@ -0,0 +1,105 @@
...
+   */
+  description?: string | ReactElement;
+
+  /**
+   * Version 2 is highly experimental, avoid using it unless you have talked with Atlantis first.
+   */
+  version: 2;
+};
+
+export type CheckboxProps = XOR<CheckboxLabelProps, CheckboxChildrenProps> & {
```
MichaelParadis, line 103: Minor comment, I think we should export the V1 prop type as CheckboxLegacyProps similar to what we did with InputText since the V1 prop type doesn't look to be exported currently


image

packages/components/src/Checkbox/Checkbox.types.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * ID for the checkbox input
+   */
+  readonly id?: string;
+
+  /**
+   * Called when the checkbox value changes
+   */
+  onChange?(newValue: boolean): void;
```
MichaelParadis, line 52: Do you think it is worth exposing the raw event here. I am thinking it may be useful for the `onFocus` / `onBlur` events

ZakaryH: yeah I could see the value in that

are you thinking we'd do it for both the existing checkbox and the new one, or just the new one?

MichaelParadis: I think just the new one if that is easier but I could see value in adding to both (if it doesn't blow up scope)

ZakaryH: gotcha. actually wait, are you saying to add it to just `onChange` or `onChange` `onFocus` and `onBlur`?

MichaelParadis (approved):

ericchernuka:
> unlike some of the other v2 inputs we've made recently I've opted to not add ref as a prop. I can't think of a compelling reason you'd need access to the ref. I haven't heard of anyone requesting it in all the time Checkbox has existed. not strictly opposed to it though, just don't see a reason currently.

@ZakaryH Can we get the ref now? Its how form libraries will focus the input upon an error or programatically in libraries like React Hook Form.

PR #2371, opened 2025-02-11
fix(components): InputChipDismissibleInput clearing options in chip on initial render

MichaelParadis:
## Motivations

On the initial render of dismissive chips with a custom activator the options would be empty when they were passed in as props

## Changes

 - Updated the Popper for the `Chips` typeahead to update when the options change to make sure it renders correctly
 - Made the debounced search function take parameters instead of using stale values in the closure

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- fixed initial render of dismissible chips having an empty options list when data is present

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


Aiden-Brine (approved):
Tested via pre-release on the branch with the reported bug to confirm it is fixed. Also tested some other Chips usages

PR #2375, opened 2025-02-14
fix(components): Allow Chip to be used as an activator in Menu 

Aiden-Brine:
## Motivations

Gui pointed out [here](https://getjobber.slack.com/archives/C03BJHV3PMG/p1739480113250269) that a `Chip` could not be used as an activator for a `Menu`. The reason Chip doesn't work is:
- Chip is essentially a Button wrapped in a Toolbar which leads to code like this:
```
<span class="Tooltip-module__shadowActivator--G4n8v"></span>
<button class="Chip-module__chip--Ob7Ns Chip-module__base--Gz7fH" tabindex="0" role="button" type="button" aria-description="description">
```
- Before these changes we were adding a shadow ref as a sibling to the activator and then using `shadowRef.current?.nextElementSibling` to access the activator
- When a button was used as an activator this would work since the button would be the sibling of the shadow ref. Unfortunately, this was broken for a Chip since the first sibling would be the tooltip span

## Changes

- Instead of using a span with a shadow ref and grabbing the next sibling, the activator is wrapped in a div which is used to reference the activator. 

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Test out the menu functionality with an activator of a Button or a Chip (by modifying `docs/components/Menu/Web.stories.tsx`). You can also test out various menus in JO/JF by using the pre-release.

I also tested that parent click actions are handled as expected and as a more out there case I tested when a Menu is used as a toolbar to ensure none of the current functionality breaks. 

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Menu/Menu.module.css
+++ b/packages/components/src/Menu/Menu.module.css
@@ -10,10 +10,6 @@
   z-index: var(--elevation-modal);
 }
 
-.shadowRef {
-  display: none;
-}
-
 .menu {
   --menu-space: var(--space-small);
   --menu-offset: var(--space-smallest);
--- a/packages/components/src/Menu/Menu.module.css.d.ts
+++ b/packages/components/src/Menu/Menu.module.css.d.ts
@@ -1,7 +1,6 @@
 declare const styles: {
   readonly "wrapper": string;
   readonly "popperContainer": string;
-  readonly "shadowRef": string;
   readonly "menu": string;
   readonly "section": string;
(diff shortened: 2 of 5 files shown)


ZakaryH (approved):
tried it out with a few different custom activators, all of them working as expected

code changes make sense. I really don't remember why I used a shadowref in the first place 🤔

PR #2378, opened 2025-02-14
docs: Fix Safari component preview renderer bug

jdeichert:
## Motivations

This PR fixes a bug in safari with the docs site component renderer.

When loading a component's doc page, it would fail to render in Safari on both macOS/iOS. After debugging this, it only happens on https://atlantis.getjobber.com and wasn't reproducible with local _production_ builds nor in cloudflare preview builds on this PR! 

The bug was a race condition where Safari was firing the iframe's `load` event before we initialized the actual event listener, causing it to never run.

I was able to test this solution using the production site and Safari's dev tools to override the js. See the before/after for the only proof I can provide 😅 

## Before

https://github.com/user-attachments/assets/79957250-beb0-4568-ae11-db14981f930f

## After

https://github.com/user-attachments/assets/d92c5140-37cf-4493-8603-7aeacc4b6a31

### Fixed

- Fixed componet preview rendering in Safari

## Testing

NOTE: as I said above, this actually wasn't broken for me on preview branches nor local (only in production). But we should still test the preview branch to make sure other browsers work fine.

* Load the cloudflare preview branch below
* Navigate between component pages, ensure the preview renders fine
* Navigate between web/mobile tabs, ensure the preview renders fine

### DIFF
--- a/docs/components/LightBox/Web.stories.tsx
+++ b/docs/components/LightBox/Web.stories.tsx
@@ -18,11 +18,13 @@ const images = [
     url: "https://images.unsplash.com/photo-1597201278257-3687be27d954?ixlib=rb-4.0.3&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=1170&q=80",
     caption:
       "This was the view of Bushart Gardens in Victoria, BC, Canada in July from a hill.",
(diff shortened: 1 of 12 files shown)


edison-cy-yang (approved):
Tested on Safari and Firefox and everything looks fine, although this is the classic race condition only in production build

scotttjob (approved):
I know Edi already approved, but I was able to verify using the staging + prod links that this fixes the race condition.

Change:
image


Old Code:
image


Great work Jacob!

PR #2380, opened 2025-02-14
docs: Add Triton to SideDrawer

taylorvnoj:
## Motivations

We are integrating Triton into the New Docs site!

### Added

New SideDrawer opens and asks for Triton API Key
- `InputPassword` gives validation error if the key is wrong
- if API Key is correct, you go to Triton Chat

Triton UI left panel is in a `SideDrawer` in the New Docs site:
- we establish a connection to Triton
- bring in Triton UI (welcome message, InputText and chat stream)

I've also installed `react-markdown` into the `packages/site/package.json` to match using `ReactMarkdown` in `TritonConversation` (as it's used in `atlantis-ai`)

## Testing

To test, you need to first make sure you have the `atlantis-ai` `.env.development` filled out as described in that repo's README.
Next you should be running the `atlantis-ai` server using `npm run dev`.

in `atlantis/packages/site` you then start the server, `npm run dev`

### Validating via API Key:
- [ ] when you open the Triton `SideDrawer` you are prompted for a Triton API key. 
- [ ] If you enter the wrong API key you get a validation error.
- [ ] If you enter the correct Triton API key (in 1Password) you will be redirected to the Triton UI

Demo:

https://github.com/user-attachments/assets/307bdbe8-de5f-4561-b8a2-07bf837b5476

### Question & Response:
- [ ] You can ask questions via the `InputText`, and the `Ask` button will be loading until the response is fully returned
- [ ] It is clear in the UI what the question & response was
- [ ] Code blocks appear if prompted and links to component or design pages works
- [ ] Responses stream in, as if being typed
- [ ] The bottom of the response is always visible, ensuring the container scrolls as responses come in

Demo:

https://github.com/user-attachments/assets/85c787be-5e83-486a-8447-59cea9d7ddba

Changes can be

---

Random photo of Atlantis


packages/site/src/components/TritonConversation.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+const Message = ({ question, response }: MessageProps) => (
+  <div>
+    <div style={{ width: "100%", display: "flex" }}>
+      <Box
+        background="surface--background"
+        padding="small"
+        radius="base"
+        margin={{ bottom: "small" }}
```
edison-cy-yang, line 32: I feel like this could use a slightly bigger margin

packages/site/src/components/TritonConversation.tsx
```diff
@@ -0,0 +1,85 @@
...
+                  <Link url={getComponentPath(key || "")} external={false}>
+                    {children}
+                  </Link>
+                );
+              },
+              p: ({ children }) => <Text>{children}</Text>,
+              pre: ({ children }) => (
+                <pre className={styles.codeWrapper} tabIndex={0}>
+                  {children}
+                </pre>
```
edison-cy-yang, line 59: We cannot use the edit icon button here (since that was copying code to editor in Triton web) but we could make use of the copy button that simply copies to clipboard.

packages/site/src/components/TritonConversation.tsx
taylorvnoj: Good eye! Updated it to `base` to match Triton 458e6948788e7475dbdf01eeab5ff5115ca461d2

Screenshot 2025-02-19 at 2 39 12 PM

packages/site/src/utils/useTritonApi.ts
```diff
@@ -0,0 +1,26 @@
+interface TritonApiOptions {
+  endpoint: string;
+  body?: object;
+  key?: string;
+}
+
+export function useTritonApi() {
+  const fetchWithApiKey = async ({ endpoint, body, key }: TritonApiOptions) => {
```
edison-cy-yang, line 8: since these are POST requests, the function name should reflect that, maybe something like `callTritonApi` or `invokeTritonApi`

packages/site/src/components/TritonLinks.ts
```diff
@@ -0,0 +1,20 @@
+import { componentList } from "../componentList";
```
edison-cy-yang, line 1: Nice implementation to avoid hard coding! Although this file would be more suitable in the `utils` folder

packages/site/src/providers/TritonProvider.tsx
```diff
@@ -16,29 +23,99 @@ const TritonContext = createContext<TritonContextType>({
...
 });
 
+const handleStreamResponse = async (
+  fullText: string,
+  setResponses: React.Dispatch<React.SetStateAction<string[]>>,
+) => {
+  let accumulated = "";
+  const chunkSize = 5;
+
+  const scrollToBottom = () => {
```
edison-cy-yang, line 41: `scrollToBottom` is probably better suited to live outside of this function

packages/site/src/providers/TritonProvider.tsx
```diff
@@ -16,29 +23,99 @@ const TritonContext = createContext<TritonContextType>({
...
+      if (!response.body) return;
+      const reader = response.body.getReader();
+      const { value } = await reader.read();
+      const fullText = new TextDecoder().decode(value);
+
+      setQuestions(prev => [...prev, question]);
+      setResponses(prev => [...prev, ""]);
+      setQuestion("");
+
+      await handleStreamResponse(fullText, setResponses);
```
edison-cy-yang, line 113: I dont think the responses are actually being streamed with this implementation. If you look at the original implementation, it has the do while loop that is reading the response from the server as it comes in. Line 106-107 will wait til the full response is returned from the server, then hand it off to `handleStreamResponse`. `handleStreamResponse` breaks up the text into chunks to make it look streamed.

I should add that response streaming isn't working in local development, so that might be where the confusion comes in.

edison-cy-yang (commented):
Nice initial implementation! Two main things are the api key validation logic that's tied to UI, and the streaming piece. I do like the approach to break up those streaming code though!

packages/site/src/components/TritonConversation.tsx
taylorvnoj: Good call out - done here 3b31ab899ecbafa953b4ea31281a6f83ff253484

packages/site/src/utils/useTritonApi.ts
```diff
@@ -0,0 +1,26 @@
...
+    const response = await fetch(`http://localhost:8788${endpoint}`, {
```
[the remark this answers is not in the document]
taylorvnoj: Done here c74cda6b03ac0536f82f1213503ac40c981be770
I also added more logic in there to emulate what's going on in the vsCode extension code

packages/site/src/utils/useTritonApi.ts
taylorvnoj: Makes sense! I renamed it to `invokeTritonApi` here e04bdb702e22a359bf2c427c3075dd48fb4c53e6

packages/site/src/components/TritonLinks.ts
taylorvnoj: Yess good call, I really want to clean up the file structure in this area in general, but good idea to move this now. Done here 3750e01ff8e0e136cfe459e84c95d6d8a00dfb33

packages/site/src/providers/TritonProvider.tsx
taylorvnoj: done here e6273592ff46a6cbc726afd4f652f397b6b5cd6f

packages/site/src/utils/useTritonApi.ts
```diff
@@ -0,0 +1,42 @@
...
+  endpoint: string;
+  body?: object;
+  key?: string;
+}
+
+export function useTritonApi() {
+  const invokeTritonApi = async ({ endpoint, body, key }: TritonApiOptions) => {
+    try {
+      const response = await fetch(
+        `https://atlantis-ai.jobber.dev${endpoint}`,
```
nad182, line 11: This is not a request to change it, but I'm curious whether we could call local or production version of Triton based on the NODE_ENV variable?
E.g. `NODE_ENV=development`, when running the app locally and `NODE_ENV=production` when running the app in production. and use it like this:
```ts
// in some config file
export const TRITON_PROD_URL = "https://atlantis-ai.jobber.dev";
export const TRITON_DEV_URL = "http://localhost:8788";
export const TRITON_URL = process.env.NODE_ENV === 'production` ? TRITON_PROD_URL : TRITON_DEV_URL;

// in this file
const response = await fetch(
  `${TRITON_URL}${endpoint}`,
  // ...
```

edison-cy-yang: @nad182 We could, although I don't think there's enough usage for us to do that. We wouldn't really have to target different Triton environments after this work is done.

It would probably be a environment variable with the same name like `TRITON_API_URL` and is retrieved and replaced into code during build. Would be a fun side quest!

nad182: I think that could be useful to do that in case we'll be working on Triton changes locally and we'd want to see them reflected in the local Atlantis docs site, without the need to pre-release `atlantis-ai` or changing the URL manually 🤷‍♂️.
But as I mentioned earlier, that's just a suggestion.

packages/site/src/components/TritonSideDrawer.tsx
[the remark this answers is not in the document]
taylorvnoj: Thanks for pairing with me on this one! Updated the logic and moved it to `useTritonApi` here 1ab88b5838a8ed62eb13faf9054518c94ed3721f
Then have subsequently moved over `scrollToBottom` 912733fbc79a38f6b5e55890defbeefe636a5eaf and `sendSearch` & `handleStreamResponse` to `useTritonApi` also to clean up the provider 7426cd89d676a4bf23ebebcfb6623eeb857572ce

So we can no longer set a key via localstorage and gain access to the chat window.

packages/site/src/utils/useTritonApi.ts
taylorvnoj: I think it's probably a good idea, but not as part of this work. Like Edi mentioned I think would be a good side desk task or we can turn it into a ticket!

packages/site/src/providers/TritonProvider.tsx
taylorvnoj: Thanks again for pairing on this too! Copied code from Triton here 8e0d2e26f424f215e0cdff98cffd8945081cec9f and removed the commented code here 1ce0591c303b70102fd218794d34533b9527f090

Like you mentioned, whenever this merges I'll be on standby testing everything on prod!

packages/site/src/components/TritonConversation.tsx
edison-cy-yang: I think there's an issue with the typing so codeContent is null/empty and nothing is copied to the clipboard. `children` is `ReactNode[]` so the conversion to `ReactElement` did not work.

I haven't looked too deep yet but `children` in the original implementation is `ReactNode` instead of a list 🧐

taylorvnoj: How are you testing that @edison-cy-yang ? I see `codeContent` and the copy button does copy to clipboard

https://github.com/user-attachments/assets/5e898149-216e-4b5f-8ca8-65f57ae52b98

packages/site/src/utils/useTritonApi.ts
```diff
@@ -0,0 +1,169 @@
...
+
+      if (key) {
+        localStorage.setItem("tritonApiKey", key);
+      }
+
+      setIsValidKey(true);
+
+      return true;
+    } catch (error) {
+      console.error("API key validation failed:", error);
```
edison-cy-yang, line 94: Could get rid of this line since the error is already logged on line 64

taylorvnoj: This and the other clean up we talked about here c8d6c9b7870207ed4dd2b1cb62ab6dfaffd2dbcd

packages/site/src/components/TritonConversation.tsx
edison-cy-yang: Confirm that it's working properly. `react-markdown` somehow wasnt installed. Makes me wonder how the markdown is loading lol

edison-cy-yang (approved):
Nice work!

PR #2386, opened 2025-02-19
docs: Hide nav and add hamburger to collapse side nav

edison-cy-yang:
## Motivations

We want to make the Atlantis doc site mobile responsive

## Changes

- `TopNav` layout changes during narrow viewport
- Hamburger button to open nav menu during narrow viewport
- `NavMenu` layout - move Jobber logo next to the close button in narrow viewport
- Attach `toggleMobileMenu` onClick handler to every react-router-dom Link in NavMenu to close nav when clicked. The state update is only triggered if viewport is below medium.

### Added

- `LeftDrawer` is a simplified SideDrawer for the purpose of this responsive nav
- `isMobileMenuOpen` state in AtlantisSiteContext

### Changed

- `NavMenu` contains both the desktop nav and mobile nav

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

The breakpoint is 768px. Test the responsiveness in TopNav, as well as NavMenu (which can be opened from the hamburger menu button)
---

Random photo of Atlantis


packages/site/src/layout/AnimatedPresenceDisclosure.tsx
```diff
@@ -63,7 +66,7 @@ function AnimatedPresenceDisclosure({
           isTitleSelected ? styles.selected : ""
         } stickySectionHeader`}
       >
-        <Link to={to ?? "/"} tabIndex={0}>
+        <Link to={to ?? "/"} tabIndex={0} onClick={toggleMobileMenu}>
```
edison-cy-yang, line 69: these onClick handlers on Link are responsible for closing mobile menu when a route is clicked on

packages/site/src/layout/LeftDrawer.tsx
```diff
@@ -0,0 +1,37 @@
+import { Box, Button } from "@jobber/components";
```
edison-cy-yang, line 1: LeftDrawer is a simpler version of our SideDrawer, until we can get SideDrawer to slide out from different directions

packages/site/src/layout/NavMenu.module.css
```diff
@@ -5,10 +5,14 @@
   --navItemHeight: 40px;
   display: flex;
   flex-direction: column;
-  padding: 35px 0 0 0;
-  height: 100dvh;
+  height: 100%;
   min-width: var(--sideBarWidth);
   box-sizing: border-box;
+
+  @media screen and (min-width: 768px) {
```
edison-cy-yang, line 12: `--medium-screens-and-up` css variable isn't working properly due to missing post-media plugin. There's also a couple other issues with css import that I spotted along the way. Will get another ticket started to deal with it.

packages/site/src/layout/NavMenu.module.css
```diff
@@ -5,10 +5,14 @@
   --navItemHeight: 40px;
   display: flex;
   flex-direction: column;
-  padding: 35px 0 0 0;
-  height: 100dvh;
+  height: 100%;
```
edison-cy-yang, line 8: height of NavMenu in mobile viewport should not be 100dvh since it lives inside the LeftDrawer with header on top

packages/site/src/layout/SearchButton.module.css
```diff
@@ -1,40 +1,47 @@
 .searchButton {
-    display: flex;
```
edison-cy-yang, line 2: this file is mostly just spacing diff

packages/site/src/providers/AtlantisSiteProvider.tsx
```diff
@@ -50,13 +58,21 @@ export const AtlantisSiteProvider = ({
     }));
   }, []);
 
+  const toggleMobileMenu = () => {
+    if (!mediumAndUp) {
```
edison-cy-yang, line 62: added an additional check for viewport here to avoid unnecessary `isMobileMenuOpen` state updates when routes are clicked in desktop viewport (although it makes no difference UX wise)

packages/site/src/layout/LeftDrawer.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children,
+  open,
+  onClose,
+  header,
+}: LeftDrawerProps) {
+  if (!open) return null;
+
+  return (
+    <>
+      <div className={styles.overlay} onClick={onClose} />
```
ZakaryH, line 21: not sure I understand the purpose of the overlay. from my testing we are never able to see it or interact with it since the drawer covers the entirety of the screen when it's open.

edison-cy-yang: yeah good call, that was from the first run when I brought over changes from SideDrawer

packages/site/src/layout/LeftDrawer.module.css
```diff
@@ -0,0 +1,47 @@
...
+.header {
+  padding: var(--space-base);
+}
+
+.content {
+  flex: 1;
+  overflow-y: hidden;
+}
+
+@keyframes slideIn {
```
ZakaryH, line 40: I'm thinking we should also do a slide out animation so it feels a bit less abrupt on the way out

packages/site/src/layout/NavMenu.tsx
```diff
@@ -118,6 +120,27 @@ export const NavMenu = ({ mainContentRef }: NavMenuProps) => {
       </a>
     </nav>
   );
+
+  return (
+    <>
+      <div className={styles.desktopNavContainer}>{menuContent}</div>
+      <div className={styles.mobileNavContainer}>
+        <LeftDrawer
+          open={isMobileMenuOpen}
```
ZakaryH, line 129: I know the `LeftDrawer` is not a component we're looking to export and re-use much, even still I'm tempted to not have `open` be a prop on it since the only way it uses `open` is to return `null`, and instead let that decision making happen here at the level where we have it in our template

so we'd just do

```
{isMobileMenuOpen && <LeftDrawer... />

```

and the `LeftDrawer` doesn't have to worry about a case where it shouldn't show.

I wouldn't call this a blocker though.

packages/site/src/layout/LeftDrawer.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <div className={styles.drawer}>
+        <Box padding="small" direction="row" alignItems="center" gap="small">
```
ZakaryH, line 23: if we make this `padding="base"` the position of the Jobber logo more or less stays in the same place

there's still a 2px difference between the two for some reason.... overall though, it looks nicer.

Kapture 2025-02-21 at 14 44 57

vs

Kapture 2025-02-21 at 14 45 46

packages/site/src/layout/TopNav.tsx
```diff
@@ -1,33 +1,75 @@
...
   return (
-    <Box padding="base" direction="row" alignItems="center" gap="small">
+    <Box
+      padding="base"
+      direction="row"
+      alignItems="center"
+      gap="small"
+      justifyContent={mediumAndUp ? "" : "space-between"}
+    >
+      <div className={styles.mobileNavContainer}>
```
ZakaryH, line 24: hmm we already have the `mediumAndUp` out at this level, how about we use`!mediumAndUp` as the condition to render the mobile nav stuff instead eg. `{!mediumAndUp && <Box ...`

that way we can get rid of the `mobileNavContainer` div, and since that's the only style in `TopNav.module.css` that entire file can go too

then we also have the benefit of our responsive stuff being managed in 1 way rather than split between CSS and JS, plus it's always nice to just not render things we don't want rather than hiding them with CSS when possible

packages/site/src/layout/TopNav.tsx
```diff
@@ -1,33 +1,75 @@
...
-        <Box width={200}>
-          <SearchButton />
+        <Box
+          direction="row"
+          gap="small"
+          justifyContent="center"
+          width={mediumAndUp ? "grow" : "auto"}
+          padding={mediumAndUp ? { left: "extravagant" } : {}}
+        >
+          <Box width={mediumAndUp ? 200 : "auto"}>
```
ZakaryH, line 59: we have a decent amount of ternaries with `mediumAndUp`

I'm tempted to suggest breaking it into 2 components, and then instead having it be a simpler conditional of which layout to use... but I'm not sure it's really worth it. I wouldn't say the current setup is hard to understand.

what do you think?

packages/site/src/layout/NavMenu.tsx
```diff
@@ -67,19 +68,20 @@ export const NavMenu = ({ mainContentRef }: NavMenuProps) => {
...
+  const menuContent = (
     <nav className={styles.navMenuContainer}>
       <div className={styles.navMenuHeader}>
         <VisibleWhenFocused>
           <Button label="Skip to Content" onClick={skipToContent} />
         </VisibleWhenFocused>
-        <Box>
-          <Link to="/">
+        <div className={styles.navMenuHeaderLogo}>
+          <Link to="/" onClick={toggleMobileMenu}>
```
ZakaryH, line 81: when does this show up?

I've tried replacing the content with HELLO WORLD and I only ever see it when we're in not-mobile mode, at which point I'd ask if we need it to have an `onClick` to toggle a menu that doesn't apply to it

packages/site/src/layout/NavMenu.module.css
```diff
@@ -130,4 +134,28 @@ a.navFooterLink {
   margin-top: auto;
   font-size: var(--typography--fontSize-small);
   color: var(--color-text--secondary);
+}
+
+.desktopNavContainer {
+  display: none;
+
+  @media screen and (min-width: 768px) {
+    display: block;
```
ZakaryH, line 143: do we need to explicitly be setting `display: block` on divs? they should just do that naturally

then we could reduce our style rules to only be applying the thing we want to change, when we want to apply them

packages/site/src/layout/TopNav.tsx
edison-cy-yang: overall I wanted to use media query and steer away from this sort of pattern, but also wanted to keep using Box instead of divs. Agree breaking into two components will be nicer.

packages/site/src/layout/NavMenu.tsx
edison-cy-yang: ah yeah this used to be the same code for desktop and mobile but not anymore. Good catch

packages/site/src/layout/NavMenu.tsx
```diff
@@ -118,6 +120,27 @@ export const NavMenu = ({ mainContentRef }: NavMenuProps) => {
       </a>
     </nav>
   );
+
+  return (
+    <>
+      <div className={styles.desktopNavContainer}>{menuContent}</div>
+      <div className={styles.mobileNavContainer}>
```
ZakaryH, line 127: I'm getting a bit confused with these two classes called `mobileNavContainer`

what's this one for? if I delete it, as far as I can tell everything still works nicely

in general I'm having not the easiest time following what is showing/when with the combination of CSS `display: none` and JS `mediumAndUp` conditionals

packages/site/src/providers/AtlantisSiteProvider.tsx
ZakaryH: hmm if we're calling a method that doesn't make sense to call in that context, is it possible for us to just not add it in that case?

packages/site/src/layout/NavMenu.module.css
edison-cy-yang: I did this to follow the design guideline to style mobile first, also for when the `--medium-screens-and-up` css variable is added back.

packages/site/src/layout/TopNav.tsx
ZakaryH: mkay, if you wanna refactor it that definitely works for me!

again I don't think this is in currently in a state where I find it hard to read or anything, so if sharing 90% of the code with this handful of conditionals is beneficial enough, I don't have any hard objections to leaving it as is.

packages/site/src/layout/NavMenu.tsx
edison-cy-yang: This mobileNavContainer is not needed since the visibility is controlled in LeftDrawer, I removed it.

I dont like mixing plain media query with useBreakpoints hook in js either. If we had to pick, is it fine to use div instead of Box in order to get all the responsive logic inside css?

packages/site/src/providers/AtlantisSiteProvider.tsx
edison-cy-yang: For some reason I didn't think of doing this check in where toggleMobileMenu is called 🤦. Updated!

packages/site/src/layout/NavMenu.tsx
edison-cy-yang: Updated!

packages/site/src/layout/TopNav.tsx
edison-cy-yang: Didn't see this comment before my [reply](https://github.com/GetJobber/atlantis/pull/2386#discussion_r1966326645). In this case, it makes sense to only use `mediumAndUp` and get rid of the css file entirely.

packages/site/src/layout/LeftDrawer.module.css
```diff
@@ -0,0 +1,47 @@
+.overlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background: rgba(var(--color-black--rgb), 0.4);
```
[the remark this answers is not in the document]
edison-cy-yang: overlay removed

packages/site/src/layout/NavMenu.tsx
edison-cy-yang: Let me think about the combination of css and JS for a bit.
Overall the pattern I see is
- TopNav uses `mediumAndUp` only
- NavMenu uses CSS media query only, except the logic where I dont attach `toggleMobileMenu` to `Link` (but this has nothing to do with showing/hiding)

It's not really that convoluted anymore. Thoughts?

packages/site/src/layout/LeftDrawer.module.css
edison-cy-yang: Added!

ZakaryH: that looks & feels much better!

ZakaryH (approved):
LGTM

PR #2387, opened 2025-02-19
feat(components-native): AtlantisThemeContext for mobile

jdeichert:
## Motivations

This PR adds a new mobile component: `AtlantisThemeContextProvider`. This component behaves similar to our [web version](https://atlantis.getjobber.com/components/AtlantisThemeContext) with the exact same API.

We also expose a `useAtlantisTheme()` hook. This is slightly different from the web version because it also returns a `setTheme()` method which allows you to change the theme globally. The web version is a little more complex because it needs to dispatch a global event (via `updateTheme()`) that all theme providers (in various distinct react roots) listen for. Mobile doesn't need that complexity as we only have one react tree and not a bunch of disparate islands.

### Added

- New AtlantisThemeContextProvider for mobile

## Testing

There's not much testing to do. This is a new component that will not affect JM at all until we start using it, which will be in followup PRs.

* Check out the new [mobile stories](https://job-116234-theme-context-for.atlantis.pages.dev/storybook/?path=/story/components-themes-atlantisthemecontext-mobile--basic)
    * Note that the contents inside the theme provider don't look correct in dark mode! We'll be slowly adding dark theme support to all native components, starting with a few prioritized ones.
* Check out the [docs page update](https://job-116234-theme-context-for.atlantis.pages.dev/components/AtlantisThemeContext?)
    * Documents mobile usage  
 

### DIFF
--- a/docs/components/AtlantisThemeContext/AtlantisThemeContext.stories.mdx
+++ b/docs/components/AtlantisThemeContext/AtlantisThemeContext.stories.mdx
@@ -26,7 +26,8 @@ theme for the entire app.
 
 On web, you'll need to import the `updateTheme` function and call it with the
 new theme. This is a separate function because it synchronizes the theme update
-across all providers in various react trees (due to our island architecture).
+across all providers under various React trees. Synchronizing across providers
+is necessary for cases where an island-based architecture is used.
 
 ### Usage for web
 


packages/components-native/src/AtlantisThemeContext/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * The children to render.
+   */
+  readonly children: React.ReactNode;
+
+  /**
+   * Force the theme for this provider to always be the same as the provided theme. Useful for sections that should remain the same theme regardless of the rest of the application's theme.
+   * This is dangerous because the children in this provider will not be able to change the theme.
+   */
+  readonly dangerouslyOverrideTheme?: Theme;
```
Aiden-Brine, line 30: Maybe too late for this because I know we use this language on the web theme provider but ideally should this be "unsafe" instead of "dangerous" or is there a difference between an unsafe action and a dangerous action?

docs/components/AtlantisThemeContext/AtlantisThemeContext.stories.mdx
```diff
@@ -61,6 +61,50 @@ function ThemedComponent() {
 }
 ```
 
+### Usage for mobile
```
Aiden-Brine, line 64: The usage sections give enough information to figure out how to use both web and mobile but how about a small section to highlight the differences so that if someone who is familiar with one is looking at the other they can quickly see the difference?

Aiden-Brine (approved):
I left two non-blocking questions but otherwise this looks really nice!

packages/components-native/src/AtlantisThemeContext/types.ts
Aiden-Brine: Checking the code I think this is the only place we use "dangerous"

docs/components/AtlantisThemeContext/AtlantisThemeContext.stories.mdx
```diff
@@ -13,7 +13,22 @@ tokens.
...
+`theme` and `tokens` in your components.
+
+On mobile, this hook also returns a `setTheme` function which you'll use to
+update the theme for the nearest `AtlantisThemeContextProvider` ancestor.
+Typically there will only be a single provider at the root, controlling the
+theme for the entire app.
+
+On web, you'll need to import the `updateTheme` function and call it with the
+new theme. This is a separate function because it synchronizes the theme update
+across all providers in various react trees (due to our island architecture).
```
Aiden-Brine, line 29: Should this say (to handle a cases where a React Island) architecture is used since there is potential that someone outside of Jobber is reading these docs?

PR #2388, opened 2025-02-19
docs: Fix git status check and many outdated props files

jdeichert:
## Motivations

The git status check was failing to notice our _generated_ props files were very out of date. This is because jobs don't persist their working directories by default, and it's expensive to persist everything in each job.

The solution I landed on was to make the git status check a step in the existing build jobs rather than being its own separate job. This is cheaper/faster as well with one less job to run.

### Fixed

- check_git_status CI job
- Regenerated props files

## Testing

No testing is necessary. The git status check caught the missing props files (see screenshot below) and after I added them, it passes.

Before |  After
:---:|:---:
Screenshot 2025-02-20 at 7 50 52 AM | Screenshot 2025-02-20 at 8 13 41 AM

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 packages/site/src/content/**/*.props.json linguist-generated=true
-ackages/site/src/content/**/*.props-mobile.json linguist-generated=true
+packages/site/src/content/**/*.props-mobile.json linguist-generated=true


Aiden-Brine (approved):
Thank you for fixing this!

PR #2389, opened 2025-02-20
feat(components-native): Allow for stylesheets to be dynamically updated [JOB-116234]

Aiden-Brine:
## Motivations

Now that we have an [Atlantis theme context provider](https://github.com/GetJobber/atlantis/pull/2393) all of our mobile components and their styles needed to be updated to use the theme.

## Changes

There are a ton of changed files in this PR but they all fall into one of these categories:
- Adding buildThemedStyles, a function that accepts a styles factory—another function that takes tokens as an argument and returns an object suitable for StyleSheet.create. buildThemedStyles returns a hook that when called memoizes the styles factory's output and returns the resulting StyleSheet. This simplifies theming for consumers, as they only need to update their styles to use buildThemedStyles.
- Updates to styles sheets so that instead of returning styles constants they return hooks generated using buildThemedStyles
- Updates to common styles files (Typography.styles.ts and CommonInputStyles.ts) to return styles hooks, as well as deprecated styles constants (to prevent this from being a breaking change as these constants are used directly) and styles getters that return styles based on a tokens argument.
- Update all of the components to use styles hooks instead of styles constants
- Update any direct use of tokens that might change with a theme to use themed tokens as opposed to static one

## Testing

- Without being wrapped in a theme provider these changes should have no impact. Look in Storybook/using the pre-release to ensure this is the case
- Test wrapping components in AtlantisThemeContext and toggle to dark mode to see how the theme change now impacts the components

Changes can be

---

Random photo of Atlantis


packages/components-native/src/Flex/Flex.gap.styles.tsx
```diff
@@ -0,0 +1,14 @@
+import { ViewStyle } from "react-native";
+import { Spacing, spacing } from "./types";
+import { buildThemedStyles } from "../AtlantisThemeContext";
+
+export const useGapStyles = buildThemedStyles(tokens =>
+  spacing.reduce((gapObj, space) => {
+    let paddingLeft = 0;
+    if (space !== "none") paddingLeft = tokens[`space-${space}`];
```
Aiden-Brine, line 8: I know this doesn't use color tokens so could use static tokens but I was concerned that more gap styles would be added that did rely on themeing and it wouldn't be updated

packages/components-native/src/InputPressable/InputPressable.style.ts
```diff
@@ -1,29 +1,32 @@
...
+        tokens["space-smaller"],
+      lineHeight: typographyStyles.defaultSize.lineHeight,
+    },
 
-  inputDisabled: {
-    color: typographyStyles.disabled.color,
-  },
+    inputEmpty: {
+      paddingTop: 0,
+    },
```
Aiden-Brine, line 22: Not sure why the diff is being a bit weird here

packages/components-native/src/InputText/context/InputAccessory.style.ts
```diff
@@ -1,17 +0,0 @@
-import { StyleSheet } from "react-native";
```
Aiden-Brine, line 1: This wasn't being used anywhere

packages/components-native/src/ProgressBar/ProgressBar.test.tsx
```diff
@@ -45,36 +44,23 @@ describe("with stepped variation", () => {
...
-        total={3}
+        total={4}
         current={2}
         inProgress={1}
         variation={"stepped"}
       />,
     );
   });
-  it("renders the correct number of steps", () => {
-    const stepElements = screen.getAllByTestId("progress-step");
```
Aiden-Brine, line 56: These old tests were quite brittle

packages/components-native/src/InputFieldWrapper/components/Prefix/Prefix.test.tsx
```diff
@@ -156,6 +167,8 @@ describe("Prefix", () => {
         disabled: true,
       });
       const prefixLabel = tree.getByText(mockLabel);
+      // console.log(tree.debug());
+      console.log(prefixLabel.props.style);
```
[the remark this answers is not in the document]
Aiden-Brine: Thank you, I removed these and forgot to push up the commit 🤦

packages/components-native/src/InputFieldWrapper/InputFieldWrapper.test.tsx
```diff
@@ -159,7 +178,7 @@ describe("InputFieldWrapper", () => {
       const styleOverride = {
         container: {
           backgroundColor: "purple",
-          width: "50%",
+          width: "50%" as const,
```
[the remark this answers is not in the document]
Aiden-Brine: Yes, it raises the error `Type 'string' is not assignable to type 'DimensionValue | undefined'.`. If there is a better way for me to fix this please let me know. I'll admit I just got Cursor to fix the error when it popped up.

packages/components-native/src/AtlantisThemeContext/buildThemedStyles.test.tsx
```diff
@@ -47,32 +47,6 @@ describe("buildThemedStyles", () => {
     expect(result.current).toBe(initialResult);
   });
 
-  it("creates styles with dependent hooks", () => {
```
Aiden-Brine: I chose to remove this test since this isn't really a pattern we want to encourage for performance reasons. If you think I should add it back or replace it with something else please let me know.

packages/components-native/src/InputFieldWrapper/CommonInputStyles.style.ts
```diff
@@ -37,3 +48,23 @@ export const commonInputStyles = StyleSheet.create({
...
+ */
+export const commonInputStyles = StyleSheet.create(
+  // eslint-disable-next-line import/no-deprecated
+  createCommonInputTokens(staticTokens, staticTypographyStyles),
+);
+
+export const getCommonInputStyles = (
+  tokens: AtlantisThemeContextValue["tokens"],
+) => {
+  const typographyStyles = getTypographyStyles(tokens);
```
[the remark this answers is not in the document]
Aiden-Brine: Oh yeah, that is a nice cleanup! Good catch

packages/components-native/src/Typography/Typography.tsx
```diff
@@ -291,7 +301,10 @@ function getSizeAndHeightStyle(
   return fontSize;
 }
 
-function getLetterSpacingStyle(letterSpacing: LetterSpacing = "base") {
+function getLetterSpacingStyle(
+  letterSpacing: LetterSpacing = "base",
```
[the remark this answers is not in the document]
Aiden-Brine: I am confused as to why this can no longer be an optional param? It is still optional on `TypographyProps`. I moved the default up to the call to `InteralTypography` since that is where I think it should live as opposed to this function call.

PR #2393, opened 2025-02-21
docs: Exclude normal functions from prop generation

jdeichert:
## Motivations

react-docgen-typescript automatically parses every export from a file, including non-component functions. This caused us to generate props for unnecessary things.

This PR fixes that by filtering out any non-component function props from our generated files.

### Fixed

- Fixed generated props files to exclude non-component functions

## Testing

No testing is really necessary here. Check out the generated files (see `displayName`) and see that only non-components were removed.


MichaelParadis (approved):
Lgtm

PR #2397, opened 2025-02-25
fix(components): Keep only one DatePicker open at a time

taylorvnoj:
## Motivations

If an `InputDate` is open and focus is shifted to another `InputDate`, they will both stay open. This can cause UX confusion & unintended side effects, especially when inputs are very close together.  This PR addresses that and makes sure only 1 `DatePicker` can be open at a time.

### Added

- Added a Date Range story example (for testing purposes but also, because we should have this example as it's a common pattern)

### Changed

Modified the class name logic for DatePicker to ensure proper open/close behaviour when there are multiple DatePickers: 

```typescript
"react-datepicker-ignore-onclickoutside": !inline && open,
```

## Testing

- [ ] test localhost Date Range story [here](http://localhost:6005/?path=/story/components-forms-and-inputs-inputgroup-web--date-range)
- [ ] make sure only 1 picker remains open when clicking back and forth; also make sure clicking away from the picker and the ESC key still close the picker
- [ ] in the Date Range story example, you can swap out one of the `<InputDate />` for an `<InputText />` and when focus goes to the `InputText`, the picker closes

You can also test that when an `InputDate` is inside a `Modal`, hitting esc key doesn't close the whole modal, it will close the `InputDate` first. This can go into `docs/components/InputDate/Web.stories.tsx`:

```
export const InsideModal: ComponentStory<typeof InputDate> = () => {
  const [date, setDate] = useState(new Date());
  const [isOpen, setIsOpen] = useState(false);

  return (
    <Content>
      <Button
        label="Open Modal with Date Picker"
        onClick={() => setIsOpen(true)}
      />

      <Modal
        title="Select a Date"
        open={isOpen}
        onRequestClose={() => setIsOpen(false)}
        primaryAction={{
          label: "Save",
          onClick: () => setIsOpen(false),
        }}
      >
        <InputDate
          value={date}
          onChange={setDate}
          placeholder="Select a date"
        />
      </Modal>
    </Content>
  );
};
```

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -115,7 +115,7 @@ export function DatePicker({
     // It uses react-onclickoutside package and declaring some elements to be
     // ignored via said class name
     // https://www.npmjs.com/package/react-onclickoutside#marking-elements-as-skip-over-this-one-during-the-event-loop
-    "react-datepicker-ignore-onclickoutside": !inline,
+    "react-datepicker-ignore-onclickoutside": !inline && open,
     [styles.fullWidth]: fullWidth,
   });
   const datePickerClassNames = classnames(styles.datePicker, {
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable max-statements */
 import React, { ReactElement, useEffect, useRef, useState } from "react";
 import classnames from "classnames";
 import ReactDatePicker from "react-datepicker";
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -89,7 +89,7 @@ interface DatePickerCloseDetails {
 
 const DATEPICKER_SWITCH_EVENT = "atlantis.datepicker-switch";
 
-// eslint-disable-next-line max-statements
+/*eslint max-statements: ["error", 16]*/
 export function DatePicker({
   onChange,
   onMonthChange,
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -87,7 +87,7 @@ interface DatePickerCloseDetails {
   id: string;
 }
 
-const DATEPICKER_SWITCH_EVENT = "atlantis.datepicker-switch";
+export const DATEPICKER_SWITCH_EVENT = "atlantis.datepicker-switch";
 
 /*eslint max-statements: ["error", 16]*/
 export function DatePicker({
--- a/packages/components/src/DatePicker/DatePicker.tsx
+++ b/packages/components/src/DatePicker/DatePicker.tsx
@@ -172,7 +172,6 @@ export function DatePicker({
   }
 
   function handleCalendarOpen() {
-    // If there's already an open picker that's different from this one, close it
     if (
       openDatePickerRef.current &&
       openDatePickerRef.current !== pickerRef.current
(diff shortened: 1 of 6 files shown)


packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -115,7 +117,7 @@ export function DatePicker({
   const datePickerClassNames = classnames(styles.datePicker, {
     [styles.inline]: inline,
   });
-  const { pickerRef } = useEscapeKeyToCloseDatePicker(open, ref);
+  const { pickerRef } = useEscapeKeyToCloseDatePicker(open);
```
taylorvnoj, line 120: the `ref` removed from here wasn't needed for ESC key functionality; ESC key handler only needs `pickerRef` to call `setOpen(false)` and doesn't need the wrapper's `ref`.

Aiden-Brine (approved):
Looks great, nice fix 👍

MichaelParadis (dismissed):
LGTM as well!

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -115,7 +121,74 @@ export function DatePicker({
...
+        event.stopPropagation();
+      }
+    };
+
+    document.addEventListener("keydown", escFunction, true);
+
+    return () => {
+      document.removeEventListener("keydown", escFunction, true);
+    };
+  }, [open]);
```
taylorvnoj, line 191: Some of the above is not new, like Github is making it look, just moved before the return

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -83,7 +83,13 @@ interface DatePickerInlineProps extends BaseDatePickerProps {
 
 type DatePickerProps = XOR<DatePickerModalProps, DatePickerInlineProps>;
 
-/*eslint max-statements: ["error", 13]*/
+interface DatePickerCloseDetails {
+  id: string;
+}
+
+export const DATEPICKER_SWITCH_EVENT = "atlantis.datepicker-switch";
```
MichaelParadis, line 90: The[index.ts ](https://github.com/GetJobber/atlantis/blob/JOB-116104/JOB-116904/one-datepicker-at-a-time/packages/components/src/DatePicker/index.ts)file will also need to be updated for consumers to use this without importing from the `/dist/` folder

taylorvnoj: thank youuuu, done here 013701bc28fb35f50a808e4c57b50b885733d481

MichaelParadis: I could see the need when people are writing testing and it would be something they could use if needed

MichaelParadis: You are right, I was worried for people writing tests in a weird way but we should actually expect them to use "standard" ways of testing things

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -83,7 +83,13 @@ interface DatePickerInlineProps extends BaseDatePickerProps {
...
+
+/*eslint max-statements: ["error", 16]*/
```
Aiden-Brine, line 92: I think it would be good to pull some things out of the function rather than silencing the error

packages/components/src/DatePicker/DatePicker.tsx
taylorvnoj: reverted exporting the event name here d016f5e74226fb2f429c59968a33d8fdc26a58b6

packages/components/src/DatePicker/DatePicker.tsx
taylorvnoj: we've gone with a totally different approach and it's really now just a 1 line change in this file, so will respectfully not do this until we have to refactor this component in the future 😆

PR #2399, opened 2025-02-26
fix(components): Fix Chip Truncation Gradient

ZakaryH:
## Motivations

the hover/focus states of the truncation gradient did not look great. this is because you cannot simply apply a `transition` to a `linear-gradient` `background`.

the goal of this PR is to make that look better than the current state.

## Changes

rather than trying to transition between the same element having 2 different gradient colors, what we're doing is instead having 2 distinct elements with a different gradient color for each of them. one will have `opacity: 1` by default, this will be the same color as the resting Chip variant background color.

the other will be `opacity: 0` by default and will have the hover/focus background color as its gradient.

as we hover, we will swap the opacity of these two gradients that are in the exact same position. since opacity IS a value that responds to `transition` we can give the appearance of the gradient changing at the same rate as the background of the Chip itself which is just a solid color.

solution inspired by
https://keithjgrant.com/posts/2017/07/transitioning-gradients/

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

go to the Chip examples, `Truncation` example

when you hover you should NOT clearly see the gradient block as a distinct element as the background of the Chip changes on hover/focus. it should blend nicely back and forth as you hover/unhover/

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chip/Chip.module.css
+++ b/packages/components/src/Chip/Chip.module.css
@@ -212,11 +212,13 @@
 }
 
 .chip:hover .truncateGradient::before,
+.chip:focus .truncateGradient::before,
 .chip:focus-visible .truncateGradient::before {
   opacity: 0;
 }
 
 .chip:hover .truncateGradient::after,
+.chip:focus .truncateGradient::after,
 .chip:focus-visible .truncateGradient::after {
   opacity: 1;
 }
--- a/packages/components/src/Chip/Chip.module.css
+++ b/packages/components/src/Chip/Chip.module.css
@@ -185,6 +185,15 @@
   height: 100%;
 }
 
+.truncateGradient > span {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  width: 10%;
+  background-color: var(--chip-base-bg-color);
+}
+
 .truncateGradient::before,
 .truncateGradient::after {
   content: "";
@@ -196,7 +205,7 @@
 .truncateGradient::before {
   background: linear-gradient(
     to left,
-    var(--chip-base-bg-color) 5%,
+    var(--chip-base-bg-color) 10%,
     rgba(255, 255, 255, 0)
   );
   opacity: 1;
@@ -205,7 +214,7 @@
 .truncateGradient::after {
   background: linear-gradient(
     to left,
-    var(--chip-base-hover-bg-color) 5%,
+    var(--chip-base-hover-bg-color) 10%,
     rgba(255, 255, 255, 0)
   );
   opacity: 0;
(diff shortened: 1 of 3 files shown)


packages/components/src/Chip/Chip.module.css
```diff
@@ -107,6 +107,7 @@
 .subtle.invalid {
   --chip-border-color: var(--color-critical);
   --chip-base-bg-color: var(--color-critical--surface);
+  --chip-base-hover-bg-color: var(--color-interactive--background--hover);
```
ZakaryH, line 110: since we're not trying to change the value, rather now we have 2 separate gradients that we are swapping between - it no longer makes sense to have this defined only in the hover state rules. we know it should always be this color.

ZakaryH:
**before**
Kapture 2025-02-26 at 09 58 27

**after**
Kapture 2025-02-26 at 09 57 26

nad182 (changes requested):
There are 2 small issues:
1. When in focus state (not focus-visible), I still see the old issue
2. In certain cases, after hovering over the chip, an extra character (or a part of it) becomes visible for a split second and then disappears. The same happens with un-hovering (aka bluring).

https://github.com/user-attachments/assets/f81a66b2-7513-4a37-ba4c-e34f650390be


https://github.com/user-attachments/assets/52f16c10-2250-403c-aafa-2a81acf4fc0d

nad182:
This is also not a big deal, but it might be useful to add the "truncated" chip to the Comparisons story: https://atlantis.getjobber.com/storybook/?path=/story/components-selections-chip-web-comparisons--all

ZakaryH:
>There are 2 small issues:

>When in focus state (not focus-visible), I still see the old issue
>In certain cases, after hovering over the chip, an extra character (or a part of it) becomes visible for a split second and then disappears. The same happens with un-hovering (aka bluring).

oh interesting, nice catch. that focus one is happening on master right now too. that's not great. I'll see what that's about.

yeah there seems to be a moment where both gradients have an opacity of `0` and an extra character or two becomes visible. I can see if adjusting the easing curves on each of the gradients can help to hide that.

ZakaryH:
the layout isn't as symmetrical as it was before

image


the thing is that with the way the columns are laid out now, you could conceivably have multiple permutations for each type, combination and then some for when the label is too long vs the heading being too long at which point it's going to 2-4x in size.

nad182 (changes requested):
I still see a flicker of a small "piece" of the next character on hover on/off, but it's such a tiny issue, that I don't don't want to block this PR because of that. However, in the Comparison story for some reason the Chips are not getting truncated on my end.



https://github.com/user-attachments/assets/56611f06-8170-4596-82db-a2d31b33fab2

ZakaryH:
@nad182 fixed the truncation examples

yeah unfortunately depending on the word and where the characters end - there's the potential for a bit of it to be visible. I've tried a number of different approaches and can't get anything that's perfect.

I'll keep thinking about this problem and see if I can't get it to work, but overall I'd say this is a big improvement over what it's doing today

nad182 (approved):
Thanks for the great improvement!

PR #2403, opened 2025-03-03
feat(components): Add anchored positioning support for SideDrawer

nad182:
## Motivations

The `SideDrawer` component previously could only be anchored to the viewport's right edge. This limited its flexibility in more complex layouts where the drawer needs to be positioned relative to specific container elements. This refactor enables the `SideDrawer` to be place inline similar to how the existing `Drawer` component works, making it more versatile for various layout requirements.

### Added

- Add inline positioning support
- Add support for custom width and positioning using `UNSAFE_style` and `UNSAFE_className`
- Tests for new functionality, such as inline rendering and UNSAFE props
- Storybook examples of using new props

### Changed

- Modified the drawer's positioning logic to support both viewport-fixed and inline modes
- Refactored the component's internal state management into a custom hook for better organization
- Updated CSS to handle both fixed and inline positioning scenarios
- Update new site docs for `SideDrawer` to include new props and notes about UNSAFE props.

### Deprecated

- 

### Removed

- 

### Fixed

- Improved the drawer's positioning to ensure it aligns properly with the right edge of its anchor element
- Ensured the drawer's width is properly constrained based on available space

### Security

- 

## Testing

The changes can be tested by:
1. Opening the Storybook and navigating to the new "Inline" story
2. Verifying that the side drawer opens and mimics the Drawer behaviour
3. Testing the drawer's behaviour on window resize
4. Confirming that the original viewport-anchored functionality still works as expected

Changes can be

---

Random photo of Atlantis


MichaelParadis:
Still reviewing but I noticed that the sidedrawer doesn't position correctly on the anchored story
image

Compared to:
> Verifying that the drawer aligns correctly with the right edge of the middle panel

I also have some concerns about there not being a way to have the `SideDrawer` shift content like our existing `Drawer` component which it seems like we want to mark legacy in the future

docs/components/SideDrawer/Web.stories.tsx
```diff
@@ -268,6 +268,133 @@ const BackButtonTemplate: ComponentStory<typeof SideDrawer> = args => {
...
+};
+
+const CustomWidthAnchoredTemplate: ComponentStory<typeof SideDrawer> = args => {
+  const [sideDrawerOpen, setSideDrawerOpen] = useState(args.open);
+  const anchorRef = useRef<HTMLDivElement>(null);
+
+  // Calculate position manually to ensure proper alignment
+  const [position, setPosition] = useState({ left: 0 });
+
+  useEffect(() => {
```
MichaelParadis, line 323: Could this logic be exposed in a hook, we seem to have something very similar in `useAnchorPosition`. 

Another thought could be allowing the width to be changed through a `width` prop

packages/components/src/SideDrawer/SideDrawer.tsx
```diff
@@ -38,20 +43,77 @@ interface SideDrawerProps extends PropsWithChildren {
...
+    width: 0,
+  });
+
+  useEffect(() => {
+    if (anchorElement?.current && open) {
+      const updatePosition = () => {
+        const rect = anchorElement.current?.getBoundingClientRect();
+
+        if (rect) {
+          const width = Math.min(420, rect.right);
```
MichaelParadis, line 97: I think the `420` should be defined as a constant since right now it is just a "magic number"

packages/components/src/SideDrawer/SideDrawer.tsx
```diff
@@ -171,16 +243,21 @@ export function SideDrawer({
...
-    </SideDrawerContext.Provider>,
-    container,
+    </SideDrawerContext.Provider>
   );
 
+  if (inline) {
+    return SideDrawerRender;
+  }
+
+  return createPortal(SideDrawerRender, container);
```
ericchernuka, line 259: This would mount the sidebar even if its not open.

Aiden-Brine:
Screenshot 2025-03-19 at 4 40 29 PM

I think some padding on the left would improve the look

Aiden-Brine (approved):
Thanks for the change, looks good to go to me

PR #2417, opened 2025-03-11
feat(components): Make InputFile composable [JOB-110281]

Aiden-Brine:
## Motivations

Consumers would like the ability to customize the content in inside the dropzone for `InputFile`

## Changes

The behaviour of the current implementations should not be impacted by these changes

### Added

- A `hintText` prop so that users can easily modify the hint text without having to do much customization

### Changed

- Made the component a compound component by pulling the wrapping styles div, button, hint text, description, and validation into separate components to allow users to pick and choose which components they want
- Allowed for components to accept children so they can build their content with whatever components they want
- Added a context for subcomponents (provided or otherwise) to access information about the input

### Deprecated

- The `"button"` value for the `variation` prop has been deprecated. InputFile is essentially a dropzone component so having a button variation that looks like a button but behaves like a dropzone is unexpected. The only uses of this variation in our codebase is for the InputAvatar so once that is updated we can remove the variation prop

## Testing

- Hopefully the different options are fairly well tested via stories and tests and I tested some existing implementations via pre-release

Changes can be

---

Random photo of Atlantis


packages/components/src/Button/Button.tsx
```diff
@@ -8,6 +8,7 @@ import { Icon } from "../Icon";
 import { Typography } from "../Typography";
 
 type ButtonType = "button" | "submit";
+export type ButtonSize = "small" | "base" | "large";
```
Aiden-Brine, line 11: I recognize this will need to be updated slightly once Michael's [button work](https://github.com/GetJobber/atlantis/pull/2412) is merged in

packages/components/src/InputFile/InputFile.tsx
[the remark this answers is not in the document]
scotttjob: We've got library upgrades on the docket for the upcoming quarter (potentially even removing react-dropzone for react-aria's dnd solution). This composability work will be about changing as little as possible to get to a composable state, then we can start to iterate afterwards.

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -95,6 +99,7 @@ export interface UploadParams {
 interface InputFileProps {
   /**
    * Display variation.
+   * @deprecated The only supported variation is "dropzone"
```
MichaelParadis, line 102: I don't think we should deprecate this. Especially since the [base input file](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file) is a button and there are a few usages of the button variant.

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -531,3 +527,9 @@ export function updateFiles(updatedFile: FileUpload, files: FileUpload[]) {
 function generateId() {
   return Math.floor(Math.random() * Date.now()).toString(16);
 }
+
+InputFile.Button = InputFileButton;
+InputFile.Description = InputFileDescription;
+InputFile.DropzoneWrapper = InputFileDropzoneWrapper;
+InputFile.HintText = InputFileHintText;
+InputFile.ValidationErrors = InputFileValidationErrors;
```
MichaelParadis, line 535: It appears like consumers wouldn't be able to use  `InputFileValidationErrors` independently since `validationErrors` is determined internally.

If someone where to use this component in their `InputFile` content there would be two copies rendered because of this line https://github.com/GetJobber/atlantis/pull/2417/files#diff-7fcad67b1d9d5aa5f9968a9476c8fa6bbb08ef65afde3c73ec8543304baa49e5R376

packages/components/src/InputFile/InputFileButton.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import { useInputFileContentContext } from "./InputFileContentContext";
+import { Button, ButtonProps } from "../Button";
+
+export function InputFileButton({
+  label,
+  size,
+  fullWidth = false,
+  ...buttonProps
+}: Omit<ButtonProps, "label" | "submit" | "name" | "value" | "to" | "url"> & {
```
MichaelParadis, line 10: I am assuming the omit on `label` is due to some issue with `label` being marked as required when label for `InputFileButton` can be undefined because it is determined below.

MichaelParadis: I just tested and it looks like the omit of label isn't needed and you don't need the 
``` 
& {
 readonly label?: string 
 }
 ```

File: packages/components/src/InputFile/InputFile.tsx (hunk 3 above)
Comment (Aiden-Brine) [reply packages/components/src/InputFile/InputFile.tsx:535]: My thinking of pulling this out into its own component was more along the lines:
- Simplifying this component
- Having this available in case someone was building their own input file implementation and using this component would be beneficial for that.

I recognize that if you are using `InputFile` you wouldn't be using it independently. So I think my options are:
- Get ride of `InputFileValidationErrors` and move the functionality back in here
- Get rid of this line and just import and use `InputFileValidationErrors` directly.

I think I could argue it either way so please let me know your preference

File: packages/components/src/InputFile/InputFileButton.tsx (hunk 4 above)
Comment (Aiden-Brine) [reply packages/components/src/InputFile/InputFileButton.tsx:10]: Thank you, updated. You were right about my line of thinking but I'm not totally sure why I thought this was necessary either 🤔

File: packages/components/src/InputFile/InputFile.tsx (hunk 2 above)
Comment (Aiden-Brine) [reply packages/components/src/InputFile/InputFile.tsx]: Updated to remove the deprecation

File: packages/components/src/InputFile/InputFile.tsx (hunk 3 above)
Comment (MichaelParadis) [reply packages/components/src/InputFile/InputFile.tsx:534]: I think maybe not exporting it makes sense since it will likely cause confusion.
I think if we want to allow people to create their own full implementation of InputFile some of the logic inside of `InputFile` should be moved into a hook to allow people to get access to the basic logic of the `InputFile` (could be named `useInputFile`)

File: docs/components/InputFile/Web.stories.tsx [hunk 5]
```diff
@@ -1,13 +1,18 @@
 import React, { useCallback, useState } from "react";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { Card } from "@jobber/components/Card";
```
MichaelParadis, line 3: Since the button variation is no longer deprecated I think it would be worth having an example of a `Button` variant of the InputFile with customization like icons on the button.

You could probably update the CustomButton story and have a Custom dropzone button story

packages/site/src/content/InputFile/InputFileNotes.mdx
```diff
@@ -118,7 +126,194 @@ allowed to be uploaded.
...
+```
+
+### Using children
+
+For complete control over the layout, you can provide custom children:
+
+```tsx
+<InputFile getUploadParams={getUploadParams}>
+  <InputFile.DropzoneWrapper>
+    <h2>Upload Your Files</h2>
```
MichaelParadis, line 203: Two things,
1. I think this h2 might make more sense to be a `Heading` level 2 to make it clear that this uses Atlantis instead of using base HTML elements
2. I think you should explicitly show the usage in such a way that the Above component is rendered

I also pasted this example into a storybook story and this is what I saw. I think you should wrap these in a content to show it "matching" what a prop driven usage of InputFile 
image

packages/site/src/content/InputFile/InputFileNotes.mdx
```diff
@@ -118,7 +126,194 @@ allowed to be uploaded.
...
+function CustomUploadContent() {
+  const { fileType, allowMultiple, size } = useInputFileContentContext();
+
+  return (
+    <div>
+      <h3>Upload {allowMultiple ? `${fileType}s` : fileType}</h3>
+      <InputFile.Button size={size} />
+    </div>
+  );
+}
```
MichaelParadis, line 265: I think showing an explicit usage of this inside an `InputFile` would make sense. I know we have the example in storybook but people may not check both places

packages/site/src/content/InputFile/InputFileNotes.mdx
```diff
@@ -118,7 +126,194 @@ allowed to be uploaded.
...
+    </Content>
+  </InputFile.DropzoneWrapper>
+</InputFile>
+
+// size="small"
+<InputFile size="small" getUploadParams={getUploadParams}>
+  <InputFile.DropzoneWrapper>
+    <Content spacing="small">
+      <InputFile.Button size="small" fullWidth={false} />
+      {/* HintText and Description are omitted in small size */}
```
MichaelParadis, line 164:
```suggestion
      {/* In the default InputFile HintText and Description are omitted in small size */}
```

Makes it more clear they can be used in small variations but in the default implementations they aren't present. Also if you want to reword the suggestion I am fine with that

packages/components/src/InputFile/InputFile.test.tsx
```diff
@@ -333,3 +333,173 @@ describe("PUT requests", () => {
...
+    const button = screen.getByRole("button", { name: "Upload file" });
+    expect(
+      screen.getByRole("button", { name: "Upload file" }),
+    ).toBeInTheDocument();
+    expect(button).not.toHaveClass("fullWidth");
+    expect(screen.queryByText("Description text")).toBeInTheDocument();
+    expect(screen.queryByText("Hint text")).toBeInTheDocument();
+  });
+
+  it("renders default content without hint text and description in small size", () => {
```
MichaelParadis, line 364: In general I love that we are now testing the actual rendering of the component. I am surpised those tests didn't exist before

packages/components/src/InputFile/InputFileDescription.tsx
```diff
@@ -0,0 +1,22 @@
...
+  children,
+  ...textProps
+}: PropsWithChildren<TextProps>) {
+  const { description } = useInputFileContentContext();
+
+  if (!children && !description) {
+    return null;
+  }
+
+  return (
```
MichaelParadis, line 17: I noticed that the original Hint and description used `Typography`. Any specific reason we aren't going that here?

MichaelParadis (commented):
Just some questions.

I think 1 of two things can be done about `InputFile.ValidationErrors`
1. Remove the export of `InputFile.ValidationErrors` since the validation logic isn't exposed.
2. Add a hook to expose the internal logic of `InputFile` so `InputFile.ValidationErrors` can be used in custom content / externally.

I personally prefer option 2 but that will likely add scope and I think it can be iterated on in the future.

Other comments related to minor docs changes to make things more explicit

packages/components/src/InputFile/InputFileDescription.tsx
Aiden-Brine: This was a suggestion from Chris, I think the goal is that `Typography` is only used by `Text`

Aiden-Brine:
I thin pulling out the other functionality into hooks is a follow-up so I wrote up [a ticket](https://jobber.atlassian.net/browse/JOB-118704?atlOrigin=eyJpIjoiMWVlZWQ4OTAxYzBiNDFkNmFmOGI4ZjUyMGRkMTQzMzAiLCJwIjoiaiJ9) for that

packages/components/src/Button/Button.types.ts
```diff
@@ -45,24 +45,6 @@ interface ButtonLabelProps extends ButtonNonComposableProps {
   readonly label: string;
 }
 
-type BaseButtonProps = XOR<
```
MichaelParadis, line 48: Thanks for cleaning up this. For reference the types were accidentally duplicated and this just removes the extras that aren't needed

taylorvnoj:
I tested in a few key areas in product - no changes in behaviour 👍🏻 
Outside of what Michael already reviewed, the only thing I was thinking of was having `FormatFile` uploaded underneath each specific `InputFile` in the stories, but that is definitely **not** something you should do in this PR and is a trivial change we can do later.

packages/components/src/Button/Button.types.ts
```diff
@@ -31,6 +30,7 @@ export interface ButtonFoundationProps {
 }
 
 interface ButtonNonComposableProps extends ButtonFoundationProps {
+  readonly ariaLabel?: string;
```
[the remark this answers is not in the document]
MichaelParadis: That is correct, @Aiden-Brine can you move this to foundation props? The snippet I sent you wasn't fully correct

packages/components/src/InputFile/InputFileButton.tsx
```diff
@@ -21,7 +21,6 @@ export function InputFileButton({
   return (
     <Button
       {...buttonProps}
-      ariaLabel="Upload file"
       label={label || contextLabel}
```
MichaelParadis, line 24: Remove aria-label here because of https://accessibleweb.com/question-answer/do-i-need-aria-attributes-or-roles-on-native-elements/

This change was breaking tests and buttons by default use the text content as the label so it wasn't needed

packages/components/src/InputFile/InputFile.test.tsx
```diff
@@ -388,7 +396,11 @@ describe("Content", () => {
       />,
     );
 
-    const button = screen.getByRole("button", { name: "Upload file" });
+    const button = screen.getByRole("button", {
```
MichaelParadis, line 399: This is because the dropzone wrapper also gets the button role so this is to find the actual button and make sure it is styled correctly

packages/components/src/InputFile/InputFileButton.tsx
MichaelParadis: @jdeichert this is the other commit I made

MichaelParadis (approved):
LGTM!

PR #2428, opened 2025-03-18
fix(components): fix controlled SegmentedControl render state update

edison-cy-yang:
## Motivations
The controlled variation of `SegmentedControl` calls `onSelectValue` during rendering, which is usually a `setState` call. There should not be side effects during render phase.

This was first discovered as part of this [PR](https://github.com/GetJobber/atlantis/pull/2423) where `onSelectValue` (which is a `setPersonality` setState call in its parent) is called during render phase. This is the error:
`Cannot update a component (TritonProvider) while rendering a different component (SegmentedControlProvider).`

### Added

- 

### Changed

- Wrapped onSelectValue call in SegmentedControlProvider in `useEffect` to make sure it only gets called after initial render

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/SegmentedControl/SegmentedControlProvider.tsx
+++ b/packages/components/src/SegmentedControl/SegmentedControlProvider.tsx
@@ -4,6 +4,7 @@ import React, {
   createContext,
   useCallback,
   useContext,
+  useEffect,
   useRef,
   useState,
 } from "react";
@@ -66,10 +67,12 @@ export function SegmentedControlProvider<T>({
 
   const hasInitialized = useRef(false);
 
-  if (!hasInitialized.current && currentSelectedOption !== undefined) {
-    onSelectValue?.(currentSelectedOption);
-    hasInitialized.current = true;
-  }
+  useEffect(() => {
+    if (!hasInitialized.current && currentSelectedOption !== undefined) {
+      onSelectValue?.(currentSelectedOption);
+      hasInitialized.current = true;
+    }
+  }, [currentSelectedOption, onSelectValue]);
 
   return (
     <SegmentedControlContext.Provider
--- a/packages/components/src/SegmentedControl/SegmentedControlProvider.tsx
+++ b/packages/components/src/SegmentedControl/SegmentedControlProvider.tsx
@@ -5,7 +5,6 @@ import React, {
   useCallback,
   useContext,
   useEffect,
-  useRef,
   useState,
 } from "react";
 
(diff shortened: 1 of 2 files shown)


packages/components/src/SegmentedControl/SegmentedControlProvider.tsx
```diff
@@ -66,10 +67,12 @@ export function SegmentedControlProvider<T>({
 
   const hasInitialized = useRef(false);
 
-  if (!hasInitialized.current && currentSelectedOption !== undefined) {
-    onSelectValue?.(currentSelectedOption);
-    hasInitialized.current = true;
-  }
+  useEffect(() => {
```
edison-cy-yang, line 70: @taylorvnoj (or even @scotttjob ) I'm not familiar with the context when this got implemented. But I'm thinking we could get rid of the `hasInitialized` ref and just have `useEffect` with empty dependency array?
```
  useEffect(() => {
    if (currentSelectedOption !== undefined) {
      onSelectValue?.(currentSelectedOption);
    }
  }, []);
```

we're still making sure `onSelectValue` only gets called after initial render. Subsequent controlled value update is done via handleChange. Was there any particular use case for `hasInitialized` ? Seeing that it creates a mutable ref across renders to indicate initial value is propagated back, this useEffect also makes sure the initial value is sent back to the parent after initial render.

edison-cy-yang: it's similar situation with the original implementation. Assuming `selectedValue` and `onSelectValue` come from a `useState` hook in the parent, the parent already knows the value (which is why I was a bit confused about this block of code that calls `onSelectValue`)

As for `isControlled` vs `!isControlled` in the if statement, `onSelectValue` should be undefined when it's  uncontrolled, so this block of code is not applicable to uncontrolled variant.

taylorvnoj: right, so we should be good with 

```
useEffect(() => {
  if (currentSelectedOption !== undefined) {
    onSelectValue?.(currentSelectedOption);
  }
}, []);
```

edison-cy-yang: yeah I just thought by evaluating `isControlled` first we save two statements from being executed:
- `currentSelectedOption !== undefined`
- checking if `onSelectedValue` is undefined (that line gets compiled to a ternary operator in js)

taylorvnoj (approved):
LGTM thanks for helping test in product! TritonSideDrawer also no longer gives a console warning. 🚢

PR #2430, opened 2025-03-19
feat(components): Add UNSAFE_ props to Typography

taylorvnoj:
## Motivations

As we add `UNSAFE_` props to move and more components, it's important and helpful to get `UNSAFE_` into atom components first. Updating `Typography` will help unlock maximum customization in components like `Text`, `Disclosure`, etc.

### Added

- `UNSAFE_className` and `UNSAFE_style` on `Typography`. I decided to make these props a keyed object, even though we just have one element here.  This is for a few reasons:
1. Consistency - the mobile `Typography` component has a `textStyle` key; now `UNSAFE_` is 1:1 in mobile and web
2. Future-proofing - if we ever need to add more targetable elements in the future, we don't need to change the API structure

- Updated the Implement tab to include `UNSAFE_` info
- tests

## Testing

Add this to `docs/components/Typography/Web.stories.tsx` to see that you can override via `UNSAFE_`. See how if you then try and change `textColor` via props table in Storybook, the red remains.

```
export const UNSAFE_style = BasicTemplate.bind({});
UNSAFE_style.args = {
  UNSAFE_style: {
    textStyle: {
      color: "red",
      fontSize: "20px",
      textDecoration: "underline",
      textDecorationColor: "blue",
      textDecorationThickness: "13px",
    },
  },
};
```

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/content/Typography/TypographyNotes.mdx
+++ b/packages/site/src/content/Typography/TypographyNotes.mdx
@@ -1,3 +1,51 @@
+## Component customization
+
+### UNSAFE\_ props (advanced usage)
+
+General information for using `UNSAFE_` props can be found
+[here](/guides/customizing-components).
+
+**Note**: Use of `UNSAFE_` props is **at your own risk** and should be
+considered a **last resort**. Future Typography updates may lead to unintended
+breakages.
+
+#### UNSAFE_className (web)
+
+Use `UNSAFE_className` to apply custom classes to the Typography component. This
+can be useful for applying styles via CSS Modules.
+
+```tsx
+// Typography.tsx
+<Typography
+  UNSAFE_className={{
+    textStyle: styles.customText,
+  }}
+>
+  Text with UNSAFE className
+</Typography>
+
+// Typography.module.css
+.customText {
+  color: var(--color-purple);
+}
+```
+
+#### UNSAFE_style (web & mobile)
+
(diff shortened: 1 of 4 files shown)


taylorvnoj:
If you want to see how a component would consume this https://github.com/GetJobber/atlantis/pull/2431/files

packages/site/src/content/Typography/TypographyNotes.mdx
```diff
@@ -1,3 +1,51 @@
...
+
+#### UNSAFE_style (web & mobile)
+
+Use `UNSAFE_style` to apply inline custom styles to the Typography component.
+
+```tsx
+<Typography
+  UNSAFE_style={{
+    textStyle: {
+      color: "var(--color-purple)",
```
MichaelParadis, line 41: Super duper minor nitpik, since mobile doesn't have CSS variables this wouldn't actually work on mobile.
It may be worth showing grabbing the tokens from the `useTheme` hook since those are both available on web and mobile

taylorvnoj: Ahhh thank you!  Ok I made updates here 2464adf00e88ad05296b8b9d1071e197472f2433 - this is also the first component where we have `UNSAFE_` on mobile and web so I've formatting things differently and used Tabs. Let me know what you think

MichaelParadis (approved):
LGTM!

PR #2431, opened 2025-03-19
feat(components): Add UNSAFE_ props to Disclosure

taylorvnoj:
## Motivations

`Disclosure` is a high priority component for composability so we've decided to add `UNSAFE_` props.

### Added

- `UNSAFE_className` & `UNSAFE_style` props for each element of the `Disclosure` (including `Typography` and `Icon` being surfaced)
- Tests for both `UNSAFE_` prop type
- Implement Tab for `Disclosure`, outlining how to use `UNSAFE_` props 

## Testing

You can add this to `docs/components/Disclosure/Web.stories.tsx` and see how we can successfully target `Disclosure` elements including the `Icon` and `Text`

```
export const UNSAFECustomization = BasicTemplate.bind({});
UNSAFECustomization.args = {
  title: "Customized Disclosure",
  UNSAFE_style: {
    container: {
      border: "3px solid var(--color-purple)",
      borderRadius: "var(--radius-base)",
    },
    summary: {
      background: "var(--color-yellow--lighter)",
      padding: "var(--space-base)",
    },
    summaryWrap: {
      border: "2px dotted var(--color-purple--light)",
    },
    title: {
      textStyle: {
        color: "var(--color-pink)",
      },
    },
    icon: {
      path: {
        fill: "var(--color-pink)",
      },
    },
    arrowIconWrapper: {
      border: "2px solid var(--color-pink)",
      borderRadius: "var(--radius-base)",
    },
    content: {
      backgroundColor: "var(--color-purple--lightest)",
      padding: "var(--space-base)",
    },
  },
};
```

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Disclosure/Disclosure.tsx
+++ b/packages/components/src/Disclosure/Disclosure.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, ReactNode, useState } from "react";
+import React, { CSSProperties, ReactElement, ReactNode, useState } from "react";
 import classnames from "classnames";
 import {
   Breakpoints,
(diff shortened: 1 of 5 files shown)


packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -54,26 +92,77 @@ export function Disclosure({
   const isBelowBreakpoint = exactWidth && exactWidth < Breakpoints.small;
   const isTitleString = typeof title === "string";
 
+  const containerClassNames = classnames(
```
taylorvnoj, line 95: Although we have a `details` element and reference `details` in styles, I do feel like it's more intuitive for users for this to be called `container`. `details` is fine internally but I feel like users would confused that with `content`

packages/components/src/Disclosure/Disclosure.tsx
```diff
@@ -54,26 +92,77 @@ export function Disclosure({
...
+    styles.details,
+    UNSAFE_className.container,
+  );
+
+  const summaryClassNames = classnames(
```
taylorvnoj, line 100: I left this as `summary` only because we also have `summaryWrap` and I thought it would be better to leave these names matching rather than rename 2 more elements in `UNSAFE`

packages/components/src/Disclosure/Disclosure.tsx
MichaelParadis: I think container makes sense as well.

MichaelParadis (approved):
LGTM!

PR #2432, opened 2025-03-19
feat(components): add optional intent prop to Tooltip

bradpurchase:
## Motivations

- We wish to update the `Tooltip` component to allow for specific treatment (such as custom cursor styles) depending on the intent of the tooltip
- Specifically, we want to enable users to specify that a tooltip's intent is to display help information, so we can use the `help` cursor style in this case ([docs](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor))

### Added

- Added an optional `intent` prop to the `Tooltip`
- The default value of the `intent` prop is `info` which will use the default cursor style and apply no special treatments
- `help` is also supported as a value for the `intent` prop, and if specified, it will apply the `help` cursor style on hover

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.module.css
+++ b/packages/components/src/Tooltip/Tooltip.module.css
@@ -85,7 +85,7 @@
   left: var(--tooltip--offset);
 }
 
-.help {
+.help > * {
   cursor: help;
 }
 
--- a/packages/components/src/Tooltip/Tooltip.module.css
+++ b/packages/components/src/Tooltip/Tooltip.module.css
@@ -85,6 +85,10 @@
   left: var(--tooltip--offset);
 }
 
+.help {
+  cursor: help;
+}
+
 .tooltipMessage {
   margin: 0;
   color: var(--color-text--reverse);
--- a/packages/components/src/Tooltip/Tooltip.module.css.d.ts
+++ b/packages/components/src/Tooltip/Tooltip.module.css.d.ts
@@ -7,8 +7,8 @@ declare const styles: {
   readonly "bottom": string;
   readonly "left": string;
   readonly "right": string;
-  readonly "tooltipMessage": string;
   readonly "help": string;
+  readonly "tooltipMessage": string;
 };
 export = styles;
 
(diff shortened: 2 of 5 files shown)


ZakaryH:
hey @bradpurchase would you be able to include some real examples of how we imagine using this?

since the tooltips only show when you are either hovering on the element the Tooltip is wrapping or focused on it, I don't get how you'd ever see the `help` cursor - since that's on the tooltip itself and in the hover case that would mean as soon as you move your cursor onto the tooltip to see the `help` the tooltip itself would disappear

that leaves us with the focus case for it to show and I'm finding that a bit awkward. I'd imagine if I'm using the keyboard to navigate & focus it, I have to stop using the keyboard and then mouse over the tooltip to see the help cursor.

bradpurchase:
> another consideration is that even in 2014 the `help` cursor was considered to be an uncommon pattern
> 
> > That said, it's not a commonly-used interface anymore, and most UIs tend to embed the contextual help information into the interface from the beginning, or based on keyboard focus more than mouse focus, since modern touch interfaces don't work well with the idea of pointing at elements of the UI.
> 
> https://ux.stackexchange.com/a/63413
> 
> so I'd be curious to hear our reasons for implementing it. after a quick search of some popular component libraries/design systems I'm not seeing this on other tooltips. that's not to say it's inherently wrong, but it does make me wonder why no one else has adopted it.

Yeah, something I considered too - if it's too much effort, the value is not really worth it. Good callout.

ZakaryH:
thanks for the example that's helpful!

image

I'm assuming the little question mark icon is actually an icon and part of the UI, and to me that's already a good indication there's some sort of contextual help available. 

so I guess my question is just what does the help cursor state add, or communicate that the icon doesn't already?

I definitely agree the help icon + hover for tooltip pattern is something that I could see people needing to reach for often, but I'd be tempted to make that a simple composition/shared component within product rather than making something like `HelpIcon` within Atlantis.

----------

then as far as implementation, yeah the span might be an issue. coincidentally I'm looking into a bug where that existing shadowactivator span is causing issues where we have some code moving focus to the next sibling element and it's targeting that span rather than the `children`, so I'd definitely want to try to find a way to include the style without tinkering with the markup.

bradpurchase:
Not moving forward with this one - after some thought, I agree that overall the benefit is not worth the effort here. Closing!

PR #2433, opened 2025-03-19
fix(components): Fix Dismissible Chip Focus

ZakaryH:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useRef } from "react";
 import styles from "./InternalChipDismissible.module.css";
 import { useInternalChipDismissible } from "./hooks";
 import { InternalChipDismissibleInput } from "./InternalChipDismissibleInput";
@@ -17,7 +17,32 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
     handleChipRemove,
     handleCustomAdd,
     handleWrapperKeyDown,
+    setChipRef,
   } = useInternalChipDismissible(props);
+
+  const chipElementsRef = useRef<Map<string, HTMLElement | null>>(new Map());
+
+  // Effect to update refs when chips are rendered
+  useEffect(() => {
+    // Wait for the DOM to be updated
+    setTimeout(() => {
+      if (wrapperRef.current) {
+        const chipElements =
+          wrapperRef.current.querySelectorAll('[role="option"]');
+
+        // Update ref map
+        chipElementsRef.current.clear();
+        sortedVisibleChipOptions.forEach((chip, index) => {
+          const element = chipElements[index] as HTMLElement | undefined;
+
+          if (element) {
+            chipElementsRef.current.set(chip.value, element);
+            setChipRef(index, element);
+          }
+        });
+      }
+    }, 0);
+  }, [sortedVisibleChipOptions, setChipRef, wrapperRef]);
 
   return (
     <div
@@ -27,11 +52,11 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
       onKeyDown={handleWrapperKeyDown}
       role="listbox"
     >
-      {sortedVisibleChipOptions.map(chip => (
+      {sortedVisibleChipOptions.map((chip, index) => (
         <InternalChip
           key={chip.value}
           {...chip}
-          onKeyDown={handleChipKeyDown(chip.value)}
+          onKeyDown={handleChipKeyDown(chip.value, index)}
           onClick={handleChipClick(chip.value)}
           ariaLabel={`${chip.label}. Press delete or backspace to remove ${chip.label}`}
           tabIndex={0}
@@ -41,7 +66,7 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
               invalid={chip.invalid}
               disabled={chip.disabled}
               label={chip.label}
-              onClick={handleChipRemove(chip.value)}
+              onClick={handleChipRemove(chip.value, index)}
             />
           }
         />
--- a/packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissible.test.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissible.test.tsx
@@ -199,12 +199,7 @@ describe("Deleting a chip", () => {
     );
   }
 
-  // There is a quick focus on the body before focusing on the next chip,
-  // this makes the the component difficult to test comprehensively.
-  // Strategies tried: userEvent library (typing backspace),
-  // and waitFor on the first chip's removal.
-  // eslint-disable-next-line jest/no-disabled-tests
-  it.skip("should focus on the next chip if the very first chip has focus", async () => {
+  it("should focus on the next chip if the very first chip has focus", async () => {
     render(<MockChips />);
 
     const first = getByChipLabelText("chip");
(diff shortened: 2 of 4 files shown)


packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
```diff
@@ -17,8 +17,33 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
...
     handleWrapperKeyDown,
+    setChipRef,
   } = useInternalChipDismissible(props);
 
+  const chipElementsRef = useRef<Map<string, HTMLElement | null>>(new Map());
+
+  // Effect to update refs when chips are rendered
+  useEffect(() => {
+    // Wait for the DOM to be updated
+    setTimeout(() => {
```
ZakaryH, line 28: we're doing this 3 times and I'm not sure I love it. might we find a better way to ensure the DOM has been updated? can we not wait for and trust React's render cycles + state updating?

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
```diff
@@ -17,8 +17,33 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
...
+      if (wrapperRef.current) {
+        const chipElements =
+          wrapperRef.current.querySelectorAll('[role="option"]');
```
ZakaryH, line 31: fragile but maybe acceptable....

an alternative method could be using React.Children iteration methods and checking the `type` of the `children` rather than mixing in DOM querying

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
```diff
@@ -17,8 +17,33 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
...
+
+        // Update ref map
+        chipElementsRef.current.clear();
+        sortedVisibleChipOptions.forEach((chip, index) => {
+          const element = chipElements[index] as HTMLElement | undefined;
```
ZakaryH, line 36: this casting feels weird. why would we want to cast it as `undefined`? how or why are we iterating over chips that aren't valid?

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
```diff
@@ -17,8 +17,33 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
...
+
+          if (element) {
+            chipElementsRef.current.set(chip.value, element);
+            setChipRef(index, element);
```
ZakaryH, line 40: does this really need to be a function?

it's a one-liner to begin with

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissible.tsx
```diff
@@ -17,8 +17,33 @@ export function InternalChipDismissible(props: InternalChipDismissibleProps) {
...
+          }
+        });
+      }
+    }, 0);
+  }, [sortedVisibleChipOptions, setChipRef, wrapperRef]);
```
ZakaryH, line 45: pretty sure this `wrapperRef` dependency does nothing since it's a ref

packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissible.test.tsx
```diff
@@ -199,36 +199,72 @@ describe("Deleting a chip", () => {
...
-  it.skip("should focus on the next chip if the very first chip has focus", async () => {
+  it("should focus on the next chip if the very first chip has focus", async () => {
     render(<MockChips />);
 
     const first = getByChipLabelText("chip");
     await userEvent.tab();
     expect(first).toHaveFocus();
 
     await userEvent.keyboard("{Backspace}");
-    expect(first).not.toBeInTheDocument();
```
ZakaryH, line 215: this assertion seems relevant to the test but it was removed. bring it back.

packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissible.test.tsx
```diff
@@ -199,36 +199,72 @@ describe("Deleting a chip", () => {
...
 
-    const second = getByChipLabelText("chip2");
+    // Focus the second chip
     await userEvent.tab();
+    await userEvent.keyboard("{ArrowRight}");
+    const second = getByChipLabelText("chip2");
     expect(second).toHaveFocus();
+
     await userEvent.keyboard("{Backspace}");
-    expect(second).not.toBeInTheDocument();
```
ZakaryH, line 229: same here. this is a worthwhile assertion. either bring it back or test it separately but don't fully ditch it.

packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissible.test.tsx
```diff
@@ -199,36 +199,72 @@ describe("Deleting a chip", () => {
...
-    expect(first).toHaveFocus();
+    // After deletion, the first chip should have focus
+    await waitFor(() => {
+      const first = getByChipLabelText("chip");
+      expect(first).toHaveFocus();
+    });
+  });
+
+  it("should focus on the Add button when the last chip is removed", async () => {
+    function SingleChipMock() {
```
ZakaryH, line 237: not really necessary

we could simply delete all the chips that exist and achieve the same functionality without writing test functions that then themselves may fail or need tests of their own

packages/components/src/Chips/InternalChipDismissible/__tests__/KeyboardNavigation.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+// Helper function to get a chip by label
+function getByChipLabelText(chipName: string) {
+  return screen.getByLabelText(
+    chipName + ". Press delete or backspace to remove " + chipName,
+  );
+}
+
+// Helper to get remove button within a chip
+function getRemoveButton(chip: HTMLElement) {
+  const button = chip.querySelector('[data-testid="remove-chip-button"]');
```
ZakaryH, line 16: while this is just a test and I'm more open to querying in a test than I am in the actual component, we should see if we can't find a better way to find it.

actually, if it has a test-id why are we not using testing-library selectors?

packages/components/src/Chips/InternalChipDismissible/__tests__/KeyboardNavigation.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+  if (!button) {
+    throw new Error("Remove button not found");
```
ZakaryH, line 19: using a proper testing-library selector saves us from having to write anything like this. use that instead if possible.

packages/components/src/Chips/InternalChipDismissible/__tests__/KeyboardNavigation.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+      const addButton = screen.getByRole("button", { name: "Add" });
+      expect(addButton).toHaveFocus();
+
+      // Try to go past Add button (should stay on Add)
+      await userEvent.keyboard("{ArrowRight}");
+      expect(addButton).toHaveFocus();
+    });
+  });
+
+  describe("Focus Management When Deleting Chips", () => {
```
ZakaryH, line 152: this deletion describe already exists in the other file. merging them seems better than having the coverage duplicate/split.

packages/components/src/Chips/InternalChipDismissible/__tests__/KeyboardNavigation.test.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+        // After deletion, the Add button should have focus
+        await waitFor(() => {
+          const addButton = screen.getByRole("button", { name: "Add" });
+          expect(addButton).toHaveFocus();
+        });
+      });
+    });
+
+    describe("When using keyboard Delete/Backspace", () => {
```
ZakaryH, line 200: I do like having coverage for deletion VS removal

not sure why we were making a distinction between the two previously. I wonder if that was intentional because it seems very strange.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -22,13 +19,110 @@ export function useInternalChipDismissible({
     chip => !selected.includes(chip.value),
   );
 
+  return { chipOptions, sortedVisibleChipOptions, availableChipOptions };
```
ZakaryH, line 22: nice, I like having this as a function

I think we're mixing arrow functions and `function` functions though. we should try sticking to one syntax.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -22,13 +19,110 @@ export function useInternalChipDismissible({
...
+};
+
+// Helper function for keyboard navigation
+const getNavigationData = (wrapper: HTMLElement, target: HTMLElement) => {
+  // Find all focusable elements
+  const chipElements = Array.from(
+    wrapper.querySelectorAll('[role="option"]'),
+  ) as HTMLElement[];
```
ZakaryH, line 30: one of the downsides to querying the DOM, we lose inferred types and have to cast everything. don't love it.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -22,13 +19,110 @@ export function useInternalChipDismissible({
...
+};
+
+export function useInternalChipDismissible({
+  children,
+  selected,
+  onChange,
+  onClick,
+  onCustomAdd,
+}: InternalChipDismissibleProps) {
+  const ref = useRef<HTMLDivElement>(null);
```
ZakaryH, line 54: we're dealing with a lot of refs now. being more specific about what this is a ref for would improve readability.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -22,13 +19,110 @@ export function useInternalChipDismissible({
...
   const { liveAnnounce } = useLiveAnnounce();
 
   const actions = {
-    handleChipRemove: (value: string) => {
+    handleChipRemove: (value: string, index?: number) => {
       return () => {
         liveAnnounce(`${value} Removed`);
-        onChange(selected.filter(val => val !== value));
+
+        // If index is provided, handle focus management
```
ZakaryH, line 75: when are we not provided index? why wouldn't we be?

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -52,49 +146,95 @@ export function useInternalChipDismissible({
...
+            // If it's the first chip and there are remaining chips, focus the new first chip
+            nextFocusIndex = 0;
           }
-          actions.handleChipRemove(value)();
+
+          // First update the array (important to do this before focus)
+          onChange(newSelected);
+
+          // After the state update and component re-render, attempt to focus
+          setTimeout(() => {
```
ZakaryH, line 196: feel weird about all these settimeouts.

maybe they're necessary, maybe not.

ZakaryH:
this is OK. it works, but the way it works is not my favorite. going to see if it can be adjusted and improved. if not, might go with something else.

will try to keep the tests at least. happy with those as a goal and how much coverage they provide now.

ZakaryH:
not happy with this approach. closing.

PR #2437, opened 2025-03-21
fix(components): remove unnecessary useEffect in SegmentedControlProvider

taylorvnoj:
## Motivations

The `useEffect` was unnecessary. In a controlled component pattern, the parent is the
source of truth  and the child should only notify of actual user interactions,
not try to sync or manage state with the parent.

This change:
- Removes the `useEffect` that was calling `onSelectValue` on mount
- Maintains proper parent-child relationship where parent is all-knowing
- Prevents unwanted navigation when the component loads

## Testing

Changes can be

---

Random photo of Atlantis


edison-cy-yang (approved):
Looks good to me!

PR #2447, opened 2025-03-27
fix(components): Dismissible Chip Focus Management

ZakaryH:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

A couple keyboard focus state related issues were happening on Chips, specifically the Dismissible/Selection variant where you can press delete to remove, or use arrow keys to navigate between them.

the other versions are working as expected  where single select uses arrows to change the selection, and multi select uses tabbing. the reason they aren't affected is because the markup for the other arrow based interaction is quite different where the individual Chip isn't focusable, it's actually a wrapping `input` that is being focused

the specific issues with Dismissible Chips are
- deleting a Chip with backspace/delete was not moving focus to any other visible element, instead it was going to the invisible "shadow activator" `span` of a `Tooltip` that wraps `Chip` for truncation behavior
- using arrows to navigate between Chips would also not work because again the span was getting the focus

this was due to our logic walking up and down the DOM with methods like `nextElementSibling` and `previousElementSibling` that assumed a specific structure that was no longer true due to the Tooltip + span

### Security

- 

## Testing

Single select chips should be totally unaffected and work the same as `master`

same for Multi select chips

selection/dismissible chips should allow you to use arrow keys to navigate between them, and deleting a chip using keyboard should move focus to a sensible item. it should not disappear.

try some other simple interactions like typing in the input and then trying to move focus back to the selected chips.

Changes can be

---

Random photo of Atlantis


packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -3,6 +3,34 @@ import sortBy from "lodash/sortBy";
...
+    direction === "previous"
+      ? element.previousElementSibling
+      : element.nextElementSibling;
+
+  if (!nextElement) return null;
+
+  // Check if element is a div or button
+  if (
+    nextElement instanceof HTMLElement &&
+    (nextElement.tagName === "DIV" || nextElement.tagName === "BUTTON")
```
ZakaryH, line 25: we're trying to avoid giving the `span classname="shadow-activator"` focus

the valid sibling elements within Chips are
- button
- div
- input

Chip can be either a `div` or `button` depending on if the top level `Chips` was passed an `onClick` or not which is kind of weird. it's always interactive, I don't really agree with making it a `div` when no `onClick` is passed in to the `Chips`.

this criteria for being a focusableElement is pretty dependent on the current markup. though I am ok with having it be so tightly coupled with `Chip` because we don't support any other children for Chips.

Aiden-Brine (approved):
Works great!

PR #2451, opened 2025-03-31
feat(components): Popover composability

jdeichert:
## Motivations

As part of our composability initiative, we're working on making more components flexible by offering their internal building blocks as subcomponents. This allows consumers to build their own version with more control over design and behaviour.

This PR updates `Popover` with a few new subcomponents: `Popover.Provider`, `Popover.Arrow` and `Popover.DismissButton`.

Check out the new "Composed" [story here](https://job-101345-composable-popove.atlantis.pages.dev/storybook/?path=/story/components-overlays-popover-web--composed) to see what the new API is capable of.

Big thanks to @scotttjob for getting the majority of this in place a few months ago! ([branch](https://github.com/GetJobber/atlantis/compare/master...scott_t/menu-popper-provider-1))

### Added

- A helper component `AtlantisPortalContent` to ensure theming is correctly applied to portalled content. This is internal-only for now while we dogfood it.

### Changed

- Exposed internal building blocks as `Popover` subcomponents
- Updated docs, stories, tests

### Fixed

- Popover's now respect the current theme if it was overridden via `dangerouslyOverrideTheme={theme}`

## Testing

* Check out the new "Composed" [story here](https://job-101345-composable-popove.atlantis.pages.dev/storybook/?path=/story/components-overlays-popover-web--composed)
* Review the updated "Implement" tab docs for Popover [here](https://job-101345-composable-popove.atlantis.pages.dev/components/Popover)

### DIFF
--- a/docs/guides/customizing-components.stories.mdx
+++ b/docs/guides/customizing-components.stories.mdx
@@ -28,25 +28,20 @@ concerns and any issues that may arise in future versions.
 
 ## Compound components
 
-Some components expose their foundational building blocks. This usually
-includes subcomponents and hooks for providing styles, state, and
-additional logic.
+Some components expose their foundational building blocks. This usually includes
+subcomponents and hooks for providing styles, state, and additional logic.
 
-These components provide greater flexibility, giving you much more control
-over their appearance and behaviour.
+These components provide greater flexibility, giving you much more control over
+their appearance and behaviour.
 
 ### Example: [Button](../components/Button)
 
-The `Button` component offers a very simple API allowing you to specify a
-label, icon, and icon position. However, because it doesn't accept children,
-it's very inflexible.
+The `Button` component offers a very simple API allowing you to specify a label,
+icon, and icon position. However, because it doesn't accept children, it's very
+inflexible.
 
 ```tsx
-<Button
-  label="Home"
-  icon="home"
-  iconOnRight
-/>
+<Button label="Home" icon="home" iconOnRight />
 ```
 
 For greater control, `Button` is also available as a compound component and has
@@ -74,17 +69,16 @@ function Example() {
 
 `Button` also exposes a few hooks that can be used within your own components:
 
-* `useButtonStyles`: Returns a map of css classes that `Button` uses internally.
-Consumers might use this to make something else look like a `Button`.
-* `useButtonContext`: Returns internal context that the `Button` shares with its
-subcomponents. Consumers might use Button's context within a custom child
-component, allowing that child to adapt to the `size` prop for example.
-
+- `useButtonStyles`: Returns a map of css classes that `Button` uses internally.
+  Consumers might use this to make something else look like a `Button`.
+- `useButtonContext`: Returns internal context that the `Button` shares with its
+  subcomponents. Consumers might use Button's context within a custom child
+  component, allowing that child to adapt to the `size` prop for example.
 
 ## Named `customRender____` props
 
-Some of our components use render props following the `customRender____`
-naming convention.
+Some of our components use render props following the `customRender____` naming
+convention.
 
 This allows you to both:
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -8622,20 +8622,6 @@
         "hermes-estree": "0.23.1"
       }
     },
-    "node_modules/@react-native/community-cli-plugin/node_modules/image-size": {
-      "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/image-size/-/image-size-1.1.1.tgz",
-      "integrity": "sha512-541xKlUw6jr/6gGuk92F+mYM5zaFAc5ahphvkqvNe2bQ6gVBkd6bfrmVJ2t4KDAfikAYZyIqTnktX3i6/aQDrQ==",
-      "dependencies": {
-        "queue": "6.0.2"
-      },
-      "bin": {
-        "image-size": "bin/image-size.js"
-      },
-      "engines": {
-        "node": ">=16.x"
-      }
-    },
     "node_modules/@react-native/community-cli-plugin/node_modules/import-fresh": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/import-fresh/-/import-fresh-2.0.0.tgz",
@@ -23648,6 +23634,21 @@
       },
       "engines": {
         "node": ">=10"
+      }
+    },
+    "node_modules/image-size": {
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/image-size/-/image-size-1.2.1.tgz",
+      "integrity": "sha512-rH+46sQJ2dlwfjfhCyNx5thzrv+dtmBIhPHk0zgRUukHzZ/kRueTJXoYYsclBaKcSMBWuGbOFXtioLpzTb5euw==",
+      "license": "MIT",
+      "dependencies": {
+        "queue": "6.0.2"
+      },
+      "bin": {
+        "image-size": "bin/image-size.js"
+      },
+      "engines": {
+        "node": ">=16.x"
       }
     },
     "node_modules/import-fresh": {
@@ -44693,7 +44694,7 @@
     },
     "packages/components": {
       "name": "@jobber/components",
-      "version": "6.28.0",
+      "version": "6.29.0",
       "license": "MIT",
       "dependencies": {
         "@floating-ui/react": "^0.27.5",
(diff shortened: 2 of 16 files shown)


packages/components/src/Popover/Popover.tsx
```diff
@@ -1,146 +1,86 @@
...
-    <div
-      role="dialog"
-      data-elevation={"elevated"}
-      ref={setPopperElement}
-      style={{ ...popperStyles.popper, ...(UNSAFE_style.container ?? {}) }}
-      className={popoverClassNames}
-      {...attributes.popper}
-      data-testid="popover-container"
+  return (
+    <Popover.Provider
```
scotttjob, line 22: Big fan of the final API here. Just the base pieces, nothing more. Great example for others moving forward.

packages/components/src/Popover/Popover.tsx
```diff
@@ -1,146 +1,86 @@
...
+    <div
+      ref={setArrowElement}
+      className={classes}
+      style={{ ...popperStyles.arrow, ...UNSAFE_style?.arrow }}
+      data-testid="popover-arrow"
+    />
+  );
+};
+
+Popover.DismissButton = function PopoverDismissButton(
```
scotttjob, line 63: Big fan of how basic these are. There really isn't much to them, which is how they should be! :)

packages/components/src/Popover/types.ts
```diff
@@ -0,0 +1,124 @@
+import { CSSProperties, PropsWithChildren } from "react";
+import { XOR } from "ts-xor";
+import { ButtonDismissProps } from "../ButtonDismiss";
+
+export interface PopoverProps {
```
scotttjob, line 5: I really like locating all component types in this 'types.ts' file as a pattern moving forward. It forces us to export our types, which we should be doing when possible anyway.

packages/components/src/Popover/usePopover.tsx
```diff
@@ -0,0 +1,53 @@
...
+      modifiers: buildModifiers(arrowElement),
+      placement: preferredPlacement,
+    },
+  );
+  useRefocusOnActivator(open);
+
+  return { setPopperElement, setArrowElement, popperStyles, attributes };
+};
+
+function buildModifiers(arrowElement: HTMLElement | undefined | null) {
```
scotttjob, line 27: Totally a follow-up, but I wonder if this could live as a useMemo instead of a function call. Uses 'arrowElement' to trigger the render.

packages/components/src/Popover/Popover.tsx
```diff
@@ -1,146 +1,86 @@
...
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-function isHTMLElement(el: any): el is Element {
-  return globalThis?.document && el instanceof Element;
-}
+  return (
+    <div
+      ref={setArrowElement}
+      className={classes}
+      style={{ ...popperStyles.arrow, ...UNSAFE_style?.arrow }}
+      data-testid="popover-arrow"
```
ZakaryH, line 58: I know it's not from this PR but we should try to use `ATL-` for our testids to avoid collision with fairly common names, and also hint at the fact that a test might be dipping its toes into testing Atlantis behavior if it's targeting these containers rather than the expected content

https://jobber.atlassian.net/wiki/spaces/teamuxf/pages/2476933462/How+should+we+declare+test+ID+s+in+Atlantis+components

I do see a couple instances in product where we've already targeted this value, so it might be more effort than we'd like to do right now on that topic. don't need to address it now.

packages/components/src/AtlantisPortalContent/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface AtlantisPortalContentProps {
+  /** The content that will be rendered inside a portal. */
```
MichaelParadis, line 2: Minor comment on the types files.

It seems like we have been doing `<componentName>.types.ts` for our types files in components. I think this is preferable because it makes searching & finding the types easier. It also keeps it consistent with other types files

packages/components/src/Popover/types.ts
```diff
@@ -0,0 +1,124 @@
+import { CSSProperties, PropsWithChildren } from "react";
```
MichaelParadis, line 1: See https://github.com/GetJobber/atlantis/pull/2451/files#r2025090611
But I do love the use of types files

packages/components/src/AtlantisPortalContent/AtlantisPortalContent.tsx
```diff
@@ -0,0 +1,18 @@
...
+  children,
+}: AtlantisPortalContentProps) {
+  const { theme } = useAtlantisTheme();
+
+  return (
+    <AtlantisThemeContextProvider dangerouslyOverrideTheme={theme}>
+      {children}
+    </AtlantisThemeContextProvider>
+  );
+}
```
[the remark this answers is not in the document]
MichaelParadis: Not for this PR, however I think we should have documentation somewhere for our internal components so that future Atlantis maintainers can know how to use them without having to dig into examples.

This one is simple enough that I would say it is fine without it but it is something to consider in the future if we do make more internal components like this (unlikely from what I see at the moment)

packages/components/src/AtlantisPortalContent/types.ts
MichaelParadis: My bad, I must have missed that when reviewing it in the Card PR

packages/components/src/Popover/Popover.types.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * **Use at your own risk:** Custom style for specific elements. This should only be used as a
+   * **last resort**. Using this may result in unexpected side effects.
+   * More information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).
+   */
+  readonly UNSAFE_style?: {
+    dismissButtonContainer?: CSSProperties;
+  };
+} & XOR<PopoverDismissButtonWithChildren, PopoverDismissButtonWithoutChildren>;
```
MichaelParadis, line 104: I am really starting to like `XOR` in combination with being more explicit in the `XOR`'d types. Especially with a commit like this https://github.com/GetJobber/atlantis/pull/2451/commits/cc1e8168e506d84af8bb9f77830f2be5d7bc6458. Instead of children just not being there it is quite clear

MichaelParadis: @jdeichert actually, I have a minor comment about this. I personally am not a fan of an interface / type intersected / extended with an `XOR` I find it can be confusing at times (for an extreme example see Button)

I found a possible way to use extends in the base types being `XOR`'d you may find interesting 

<details><summary>Possible rewrite of Popover Dismiss Button types</summary>

interface PopoverDismissButtonPropsStyles {
  /**
   * **Use at your own risk:** Custom class names for specific elements. This should only be used as a
   * **last resort**. Using this may result in unexpected side effects.
   * More information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).
   */
  readonly UNSAFE_className?: {
    dismissButtonContainer?: string;
  };

  /**
   * **Use at your own risk:** Custom style for specific elements. This should only be used as a
   * **last resort**. Using this may result in unexpected side effects.
   * More information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).
   */
  readonly UNSAFE_style?: {
    dismissButtonContainer?: CSSProperties;
  };
}

interface PopoverDismissButtonWithChildren
  extends PopoverDismissButtonPropsStyles {
}

type PopoverDismissButtonWithoutChildren = Partial<ButtonDismissProps> & {
  readonly children?: never;
} & PopoverDismissButtonPropsStyles;

export type PopoverDismissButtonProps = XOR<
  PopoverDismissButtonWithChildren,
  PopoverDismissButtonWithoutChildren
>;
</details> 


<details><summary>Type testing</summary>
// Invalid, both types of props
// eslint-disable-next-line @typescript-eslint/no-unused-vars
const IncompatibleChildren: PopoverDismissButtonProps = {
  ariaLabel: "fdafda",
  children: "Has Children",
};

// Valid no children
// eslint-disable-next-line @typescript-eslint/no-unused-vars
const _test2: PopoverDismissButtonProps = {
  onClick: () => {},
  ariaLabel: "fdafda",
  UNSAFE_className: {
    dismissButtonContainer: "hihi",
  },
};

// Valid
const _test3: PopoverDismissButtonProps = {
  children: "Fdafda",
  UNSAFE_className: {
    dismissButtonContainer: "hihi",
  },
};
</details>

docs/components/Popover/Web.stories.tsx
MichaelParadis, line 1: What are your thoughts on adding an example of a "reverse themed" Popover?

MichaelParadis: Yeah, like on light mode a popover that is dark mode themed

MichaelParadis (approved):
LGTM!

PR #2452, opened 2025-04-01
refactor(hooks): Remove usages of watch() and stop unnecessary renders [JOB-120031]

KyleKreutzer:
## Motivations
Noticed while testing our client create/edit form that performance felt sluggish. Using the react dev tools, I discovered that usages of `watch()` within our hooks library were causing full form rerenders every time a form value changed. This PR addresses that issue by removing our dependency on `watch()`.

image

### Changed

- Prevent full form rerenders when a form value changes.

## Testing
Easiest to test in `jobber-frontend`. Install the package locally using `pnpm install {path to package}` on the `jobber-online` app.

Ensure you have the following flags enabled in your jobber override file. This will ensure you have access to the new requests and client create/edit form built in SPA:
```
    "spa_requests_wip" => "on",
    "spa_requests_ga" => "on",
     "react_client_create_online" => "on"
```

Try creating clients + requests. When viewing them afterwords, you should see all the fields you entered correctly persisted to the DB.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Checkbox/Checkbox.tsx
+++ b/packages/components/src/Checkbox/Checkbox.tsx
@@ -1,6 +1,6 @@
 import React, { ChangeEvent, useEffect, useId, useState } from "react";
 import classnames from "classnames";
-import { Controller, useForm, useFormContext } from "react-hook-form";
+import { Controller, useForm, useFormContext, useWatch } from "react-hook-form";
 import styles from "./Checkbox.module.css";
 import { CheckboxLegacyProps } from "./Checkbox.types";
 import { Icon } from "../Icon";
@@ -20,7 +20,7 @@ export function CheckboxLegacy({
   onChange,
   onFocus,
 }: CheckboxLegacyProps) {
-  const { control, setValue, watch } =
+  const { control, setValue } =
     useFormContext() != undefined
       ? useFormContext()
       : // If there isn't a Form Context being provided, get a form for this field.
@@ -37,11 +37,16 @@ export function CheckboxLegacy({
     name ? name : `generatedName--${identifier}`,
   );
 
+  const watchedValue = useWatch({
+    control,
+    name: controlledName,
+  });
+
   useEffect(() => {
     if (value != undefined) {
       setValue(controlledName, value);
     }
-  }, [value, watch(controlledName)]);
+  }, [value, watchedValue]);
 
   const wrapperClassName = classnames(
     styles.wrapper,


packages/components/src/FormField/hooks/useAtlantisReactHookForm.ts
```diff
@@ -46,7 +45,7 @@ export function useAtlantisReactHookForm({
     if (value != undefined) {
       setValue(name, value);
     }
-  }, [value, watch(name)]);
+  }, [value, field.value]);
```
KyleKreutzer, line 48: We already have access the the field here through `useController`, so we use `field.value` in the dependency array instead and save us from having to use watch.

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -37,11 +37,16 @@ export function CheckboxLegacy({
     name ? name : `generatedName--${identifier}`,
   );
 
+  const watchedValue = useWatch({
```
KyleKreutzer, line 40: This will only rerender the checkbox component when the watched value changes.

ericchernuka (commented):
Great find. I'll let others approve, but this seems pretty straightforward fix for a huge peformance gain.

packages/components/src/FormField/hooks/useAtlantisReactHookForm.ts
ZakaryH: I've got some FUD about this change.

hundreds of `<Input....` in product will be affected by this since it's within `FormField` that they all use under the hood and the entire point of this change is that it has very different behavior. my worry is that some of those instances are, for better or worse, relying on the current behavior .

we'll need to do more than just test in JF, since that's a different setup when it comes to forms.

packages/components/src/Checkbox/Checkbox.tsx
ZakaryH: same FUD here. we've got 100+ instances of Checkbox in product. at the minimum we've gotta have a passing CI run to verify none of those were relying on this behavior.

ZakaryH:
sorry but we need certainty before going ahead with this.

checking the new project alone isn't sufficient given how broad the scope would be for this change. as a starting place, let's create a pre-release of this branch, use that version in product and see what the CI run thinks.

KyleKreutzer:
> as a starting place, let's create a pre-release of this branch, use that version in product and see what the CI run thinks.

How can I go about doing this?

ZakaryH:
@KyleKreutzer looks like someone has already done the first part, but the steps would be: 
- github actions
- trigger pre release build
- wait for the build to be commented on the PR
- go to the `Jobber` repo
- make a new branch
- install the prerelease build with the command listed on the comment
- commit and push that branch to remote to trigger a CI run (you could try running the full test suite locally but I think there are some known flakes when doing that)
- then just see what the CI runs shows!

packages/components/src/FormField/hooks/useAtlantisReactHookForm.ts
ericchernuka: I guess the rationale here is `watch(name) === field.value`, but targetted. Now, if everyone's leaning on the full-form reload behaviour that's going to be a pretty thing to unwind.

ZakaryH: that makes sense, and it does seem like this is what we should have been using all along. 

I have no idea if anyone/someone/everyone has been relying on this behavior intentionally or otherwise. ideally, it's just an improvement across the board. not-so-ideally, it has the potential to impact a tremendous amount of stuff - which is why I'd like to take some steps to increase our certainty.

ericchernuka: Ya, fully agree. Places we can look for possible issues would be uses of `getValues()` anywhere since many think leverage it thinking it works cause of the re-render, but in this change it means that only if the component is being watched would the getValues re-run so its possible to have stale values reflected. :(

ZakaryH:
I ran this prerelease build thru a CI run https://github.com/GetJobber/Jobber/pull/58173

5 failures with at least 2 different types of reasons at a glance
- failing to call a function
- failing to find an element

while 5 is a pretty tame, manageable, number it doesn't inspire a ton of confidence for me. we know our test coverage isn't 100%, so I'm fearful of how many similar issues exist that tests aren't catching.

haven't had a chance to dig into the reasons for the failures. perhaps there's a quick fix, but yeah I'm feeling nervous about this change.

PR #2454, opened 2025-04-01
feat(components): Add UNSAFE props to Text

taylorvnoj:
## Motivations

We are currently making our way through atom components, adding `UNSAFE_` props in these small building blocks so it is easier to surface in larger components later on.

### Added

- `UNSAFE_` props in web `Text` (just surfacing via `Typography`)
- tests for `UNSAFE_` props
- new section in the `Text` `Implement` tab explaining how to use the props
- update in `Text` props file
- update in `InputFile` props file (comment below talking about it)

### Changed

- some markup was changed in `TypographyNotes` while I'm in the area

## Testing

Update existing Storybook example and look at the Basic story:

```
Basic.args = {
  children:
    '"640K is more memory than anyone will ever need on a computer" - Not Bill Gates',
  UNSAFE_style: {
    textStyle: {
      color: "red",
      fontSize: "20px",
    },
  },
};
```

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/site/src/content/InputFile/InputFile.props.json
+++ b/packages/site/src/content/InputFile/InputFile.props.json
@@ -520,6 +520,32 @@
         "type": {
           "name": "\"small\" | \"base\" | \"large\""
         }
+      },
+      "UNSAFE_className": {
+        "defaultValue": null,
+        "description": "**Use at your own risk:** Custom classNames for specific elements. This should only be used as a\n**last resort**. Using this may result in unexpected side effects.\nMore information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).",
+        "name": "UNSAFE_className",
+        "parent": {
+          "fileName": "packages/components/src/Text/Text.tsx",
+          "name": "TextProps"
+        },
+        "required": false,
+        "type": {
+          "name": "{ textStyle?: string; }"
+        }
+      },
+      "UNSAFE_style": {
+        "defaultValue": null,
+        "description": "**Use at your own risk:** Custom style for specific elements. This should only be used as a\n**last resort**. Using this may result in unexpected side effects.\nMore information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).",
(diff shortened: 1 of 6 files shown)


packages/site/src/content/InputFile/InputFile.props.json
```diff
@@ -520,6 +520,32 @@
         "type": {
           "name": "\"small\" | \"base\" | \"large\""
         }
+      },
+      "UNSAFE_className": {
```
taylorvnoj, line 524: At first, this felt wrong to me and I considered omitting `UNSAFE_` in `InputFile.HintText` & `InputFile.Description`.  I thought, maybe it would be strange to have `UNSAFE_` in these subcomponents and not in `InputFile` itself.
But I'm not really surfacing `UNSAFE_`, we're getting it for free because we're referencing `TextProps` in these subcomponents.  There could be legitimate use cases for having `UNSAFE_` in these subcomponents, so I can't really think of a good reason to omit it.

Using this in `InputFile` would look something like this:

```
<InputFile.HintText
    UNSAFE_style={{
      textStyle: {
      fontStyle: "italic",
      color: "purple",
      textDecoration: "underline",
    },
  }}
>
```

Obviously not a good real world example but you could see how someone may need to target that text for different styling:

Screenshot 2025-04-01 at 8 33 26 PM

The question is, do we leave this undocumented for now, until we handle `UNSAFE_` in `InputFile`?

Aiden-Brine (approved):
LGTM!

packages/site/src/content/InputFile/InputFile.props.json
Aiden-Brine: We chatted about this briefly and I agreed with leaving it undocumented. Taylor is going to add a note to the ticket though to make sure we don't forget when we handle `InputFile`.

taylorvnoj: Note added! Thanks Aiden!

PR #2460, opened 2025-04-03
feat(components): Make DataList.Search optionally controllable

jdeichert:
## Motivations

This PR allows the `DataList.Search` input to be controlled on the consumer side, if desired.

This means consumers now have complete control over the search input which is beneficial in cases where the input state is stored somewhere else (like url query params) and where the consumer wants to make a clear all filters button, which includes clearing the search input.

### Changed

- Made `DataList.Search` support a controlled `value` prop
- Updated tests and documentation

## Testing

* Check out the updated docs Implement tab: https://job-110225-controllable-data.atlantis.pages.dev/components/DataList
  * The search section is updated with info about controlling the value
* Check out the [Clear All Filters](https://job-110225-controllable-data.atlantis.pages.dev/storybook/?path=/story/components-lists-and-tables-datalist-web--clear-all-filters) story. 
  * Type in the search input, then click "Clear filters", verify the input clears
* Check out the [Basic](https://job-110225-controllable-data.atlantis.pages.dev/storybook/?path=/story/components-lists-and-tables-datalist-web--basic) story
  * Verify it still works as expected in the uncontrolled case: type in the input and view the console logs, verify the debounce still works as expected

### DIFF
--- a/docs/components/DataList/Web.stories.tsx
+++ b/docs/components/DataList/Web.stories.tsx
@@ -18,6 +18,7 @@ import { Chip } from "@jobber/components/Chip";
 import { Icon } from "@jobber/components/Icon";
 import { Combobox, ComboboxOption } from "@jobber/components/Combobox";
 import { Flex } from "@jobber/components/Flex";
+// eslint-disable-next-line import/no-internal-modules
 import { useDebounce } from "@jobber/components/utils/useDebounce";
 
 const meta: Meta = {


packages/site/src/content/DataList/DataListNotes.mdx
```diff
@@ -328,6 +328,26 @@ DataList.Search adds a search input to the list. This is debounced by 500ms.
 </DataList>
 ```
 
+#### Controlling the search input
+
+By default, DataList.Search is uncontrolled and internally manages its value.
```
edison-cy-yang, line 333: can we provide a quick example of how consumers should setup debouncing in their component when using controlled search? Or maybe too trivial ?

edison-cy-yang (commented):
LGTM, tested it and it works nicely. Just a small question on implementation note.

edison-cy-yang (approved):

PR #2461, opened 2025-04-04
feat(components): Layout System 2.0 

scotttjob:
## Motivations

We've had a need for a while to better support layouts using Atlantis. Our existing components are either heavily opinionated, or have absolutely no opinions at all and are doing their best to mimic CSS.

These new components sit somewhere in the middle. They're not opinionated about how you use them, but they are opinionated about how they respond to their content.

These components are heavily based on the [Every Layout](https://every-layout.dev/layouts/) system, which feature components designed to self-govern based on some basic setup properties, and the width of the child content.

Which is a slightly fancy way of saying these components can be magically responsive without media queries. Container queries though are fully supported with this system, and some layout components will be fast-followed with media query based overrides.

Each of these components also tackle a single use-case, and are meant to be mixed and matched together in a composed way to form really powerful layouts together. No CSS knowledge is required to use these new components, and the props they accept are centered around how they make the component react, versus how the underlying CSS is structured.

## Stack - Vertically Stacking Elements
image
image

## Content Block - Setting Block Width and horizontally justifying the content within
image

## Cluster - Laying out Elements Horizontally based on their width
image

## Tiles - Laying out Elements Horizontally with a fixed width
image

## Cover - Vertically Centering Elements
image

## Frame - Forcing content into a specific aspect ratio. Great for images.
image

## ResponsiveSwitcher - Responsively reflow content equally based on its width. No media queries required.
image
image

## SideKick - Responsively reflow content based on your defined widths + content width. No media queries required.
image
image

## Changes

Added new Layout components designed to help Atlantis consumers layout out their experiences, widgets, pages, containers, and views without requiring CSS knowledge or concerning themselves too much with responsiveness.

### Added

New Components:
- Cluster
- ContentBlock
- Cover
- Frame
- ResponsiveSwitcher
- SideKick
- Stack
- Tiles

Storybook Stories For all the above Components
Documentation Pages for all the above Components

New Guides:
- Page Layouts - Small Examples of how to layout individual Page elements with the new components
- Scaffolding - Full Page Examples of how layout entire Pages with the new components (will be filled in more soon)

### Changed

- Added the ability to go 'full width' on Doc Site pages. This is used for Scaffolding page, because they're full-page layouts.

### Removed

- Nothing

### Fixed

- Lack of Quality Layout Options

### Security

- No Issues

## Testing

1. Boot up the new docs site to see them in action.
2. Play around the props on the docs pages to see changes.
3. Bring in the pre-release in this PR to use the new components within applications, attempt to replace custom css with these components to see if they work in place of custom options.

Changes can be

---

Random photo of Atlantis


packages/site/src/preview/useAtlantisPreviewSkeleton.tsx
```diff
@@ -58,6 +58,50 @@ html,body,#root {
         var root;
         var rootElement;
       </script>
+      <style>
```
scotttjob, line 61: This is for the Container component example. Didn't want to introduce css modules into this preview component (at least, not for this), so for now just an easy override that we can remove later if required.

packages/site/src/types/maps.ts
```diff
@@ -4,6 +4,7 @@ export interface ContentMapItem {
   intro: string;
   title: string;
   content: ContentExport["content"];
+  noMaxWidth?: boolean;
```
scotttjob, line 7: This is for the scaffolding page, so we can override the max width so the page layouts can be a full page.

packages/site/tests/visual/site.visual.ts
```diff
@@ -0,0 +1,23 @@
+import { expect, test } from "@playwright/test";
```
scotttjob, line 1: The visual tests!

Ultra, ultra basic for now. Load the route, check the screenshot. We can quickly iteratively build on these first tests.

packages/site/src/layout/BaseView.tsx
```diff
@@ -35,7 +37,9 @@ BaseView.Main = function Main({
             <div
               style={{
                 width: "100%",
-                maxWidth: "calc(768px + var(--space-large)",
+                maxWidth: noMaxWidth
```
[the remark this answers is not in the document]
chris-at-jobber: Does this not cause issues with the docs pages? (I haven't pulled this branch down but that value keeps doc contents from running full-width and having wayyyyy too long character-count lines)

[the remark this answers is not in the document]
scotttjob: It's just an override for a single page (the new scaffolding page, or any other page that wants to run full width for any reason), and we do want that particular one to run full width, because you're building full-page layouts (presumably with content width protection built in, like Page does)

packages/components/src/Cluster/Cluster.module.css
```diff
@@ -0,0 +1,18 @@
+.cluster {
+  display: flex;
+  flex-wrap: wrap;
+  gap: var(--public-cluster-space, var(--space-base, 16px));
+  justify-content: var(--public-cluster-justify, flex-start);
+  align-items: var(--public-cluster-align, center);
+  width: 100%;
```
taylorvnoj, line 7: So I've been playing around with this type of layout, and what I see are nested Clusters (one parent Cluster, with Clusters around Icon/StatusLabel and Cluster around Buttons):

Screenshot 2025-04-11 at 1 37 16 PM

For example I took this into Storybook and tried this
```
 <Cluster justify="between">
    <Cluster>
      <Button label="Save" />
      <Button label="Cancel" type="secondary" />
    </Cluster>
    <Button label="Delete" variation="destructive" type="secondary" />
</Cluster>
```
But then we get something like this:

Screenshot 2025-04-11 at 1 40 24 PM

And from what I can gather that is happening because of the 100% width in the css.  

What fixes it, is something like this:
```
.cluster .cluster {
  width: auto;
}
```
so when a Cluster is nested inside another Cluster, we don't force full width.

Then I do get the layout I want easily without further wrapping divs around groups of components

Screenshot 2025-04-11 at 1 44 45 PM

What do you think?

taylorvnoj: I initially lean toward the CSS solution, just because the consumer doesn't have to think about anything and things _just work_ - but together also works! Do you see that happening in this PR or a follow up?

taylorvnoj: Oh maybe that prop would help in the case where we try to put any Input in a Cluster, because those just end up stacking vertically I think because of the 100% width on FormField

packages/site/src/pages/VisualTestComponentPage.tsx
```diff
@@ -0,0 +1,314 @@
...
+                  <Avatar size="large" initials="ST" />
+                </Cluster>
+              </section>
+
+              {/* Avatar with Images */}
+              <section>
+                <Text size="large">Avatar with Images</Text>
+                <Cluster>
+                  <Avatar
+                    imageUrl="https://i.pravatar.cc/300?img=1"
```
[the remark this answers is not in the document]
scotttjob: Updated to a less weird domain for placeholder images!

packages/components/src/Stack/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  | "larger"
+  | "largest"
+  | "extravagant";
+
+export interface StackProps {
+  readonly children: React.ReactNode;
+  /** The amount of space between the children. Semantic tokens are available. */
+  readonly space?: string | Spaces;
+  /** The number of children to split the stack after. Requires parent to have height greater than the sum of the children. */
+  readonly splitAfter?: number;
```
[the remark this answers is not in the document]
scotttjob: I went down the road of making some elegant like we had for Box, but it ended up being a little convoluted, so opted for 1 -> 15, because it's just easier to understand.

packages/components/src/Container/types.ts
```diff
@@ -0,0 +1,12 @@
+import { ReactNode } from "react";
+
+export interface ContainerProps {
+  readonly children: ReactNode;
+  readonly name: string;
+  readonly className?: string;
+}
```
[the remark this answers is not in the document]
scotttjob: Done!

packages/components/src/ResponsiveSwitcher/ResponsiveSwitcher.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useMemo } from "react";
+import classNames from "classnames";
+import styles from "./ResponsiveSwitcher.module.css";
+import { ResponsiveSwitcherProps, Spaces } from "./types";
+
+const spaceTokens: Record<Spaces, string> = {
```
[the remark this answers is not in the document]
scotttjob: Great opportunity to dedupe, which I did. I created a new directory called 'sharedHooks' within components, but open to other names/locations on where to put the 'useSpaces' hook

packages/components/src/Cluster/Cluster.module.css
scotttjob: Hey so I pushed an update. I have the .cluster .cluster { piece in place, but I also added 'autoWidth' as a boolean prop to all the components that had the new width: 100%, so you could say "don't do that" and it defaults back to width auto

packages/components/src/Cluster/Cluster.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { useMemo } from "react";
+import { Breakpoints } from "@jobber/hooks/useResizeObserver";
+import { useMediaQuery } from "@jobber/hooks";
+import classNames from "classnames";
+import styles from "./Cluster.module.css";
+import { ClusterProps, Spaces } from "./types";
+
+const spaceTokens: Record<Spaces, string> = {
```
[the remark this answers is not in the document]
scotttjob: I've now done this because of the space-slim token being added

packages/components/src/Stack/types.ts
```diff
@@ -0,0 +1,31 @@
+import { Spaces } from "../sharedHooks/useSpaces";
+
+export interface StackProps {
+  readonly children: React.ReactNode;
+  /** The amount of space between the children. Semantic tokens are available. */
+  readonly space?: string | Spaces;
```
taylorvnoj, line 6: As I was writing this in my doc

Screenshot 2025-04-12 at 6 01 32 PM

I realized that our _og_ layout components (Box, Flex, Grid) use `gap`, while our new layout components use `space`. Is it intentional to use `space` moving forward? We normally strive to have the same prop names for the same behaviour, although I think I like `space` more.

Not necessarily suggesting a change here but curious on what your thoughts are.

scotttjob: I very much want to keep these new components in line with our existing components. I'll update 'space' to 'gap' in all the new components this morning.

Any more prop differences you've noted that I can update. Now is the time! :)

packages/site/src/content/Cover/Cover.mdx
```diff
@@ -0,0 +1,52 @@
...
+
+## Basic example
+
+<Canvas>
+  <Cover minHeight="50vh">
+    <Text>Content above</Text>
+    <Cover.Center>
+      <Heading>Welcome back!</Heading>
+      <Text>Sign in to continue to your account.</Text>
+    </Cover.Center>
```
nad182, line 27: Not a huge issue, but in the new doc site this is displayed as `CenterContent`, instead of `Cover.Center`:
image

packages/site/src/content/Cluster/Cluster.mdx
```diff
@@ -0,0 +1,166 @@
...
+  </Stack>
+</Canvas>
+
+## Related components
+
+Use [Stack](/components/Stack) when you need to arrange elements vertically
+instead of in a horizontal cluster.
+
+Use [Tiles](/components/Tiles) when you need elements arranged in a grid with
+equal spacing and consistent sizing.
```
nad182, line 166: I assume `Flex` and `Grid` are not mentioned because they're going to be deprecated?

nad182:
The **Code** tab is broken in Storybook for every new component:
image

From what I understand, it's because of this code in `Playground.tsx`:
```tsx
<SandpackProvider
      template="react-ts"
      customSetup={{
        dependencies: {
          "@jobber/components": "latest",
          "@jobber/design": "latest",
          "@jobber/hooks": "latest",
...
...
```

Probably is not an issue, but wanted to call it out just in case.

packages/components/src/ContentBlock/ContentBlock.tsx
```diff
@@ -0,0 +1,51 @@
...
+      "--content-block-gutters": guttersMapped,
+    } as React.CSSProperties;
+  }, [maxWidthMapped, guttersMapped]);
+
+  return (
+    <div
+      style={style}
+      className={classNames(
+        styles.contentBlock,
+        andText && styles.andText,
```
nad182, line 41: This might not be the case in the "real world", but I found out that if nest `ContentBlock` components and the parent has `andText`, while child(ren) has `andText={false}`, the nested one is ignored. This is what it looks like (my understanding is that the text should NOT be centered, but I might be wrong):
image

packages/components/src/Cover/Cover.module.css
```diff
@@ -0,0 +1,22 @@
...
+  margin-block-start: 0;
+}
+
+.cover > :last-child:not(.centerContent) {
+  margin-block-end: 0;
+}
+
+.cover > .centerContent {
+  margin-block: auto;
+}
```
nad182, line 22: This might be a bad/weird question, but shouldn't be the center content centered vertically in the parent container, rather than between top and bottom content (i.e. header and footer)?
If you look at the screenshot, the center content will only be truly vertically cenetered if there's no content above and below or if the content above AND below is the same height.

image

packages/components/src/SideKick/types.ts
```diff
@@ -0,0 +1,17 @@
...
+export interface SideKickProps {
+  readonly children: React.ReactNode;
+  /** The width of the sidekick. */
+  readonly sideWidth?: string;
+  /** The minimum width of the content. */
+  readonly contentMinWidth?: string;
+  /** The amount of space between the sidekick and the content. Semantic tokens are available. */
+  readonly space?: string | Spaces;
+  /** Whether to place the sidekick on the right. */
+  readonly onRight?: boolean;
```
nad182, line 12: It might be just me, but I find the name of this prop confusing. Regardless of if I set the value to `true` or `false`, the layout/markup is preserved, meaning the side element remains on it's original side. Whereas I thought that it would act more like `flex-direction: row-reverse`.
Otherwise, I'm not sure I understand the value/role of this prop. All it does is it changes the size of the second element (on the right).

packages/components/src/SideKick/SideKick.tsx
```diff
@@ -0,0 +1,44 @@
...
+        onRight ? styles.right : styles.left,
+        sideWidth
+          ? onRight
+            ? styles.withWidthRight
+            : styles.withWidthLeft
+          : null,
+        collapsed && collapseBelow ? styles.collapsed : null,
+      )}
+    >
+      {children}
```
nad182, line 41: Shouldn't we limit the amount of children to 2? It feels like the purpose of this component is to have "main content" (1 child) on one side and the "supplementary content" (1 child) on another side.
Currently it's possible to have unlimited children, thus creating a confusion which content should be main and/or supplementary:
image

packages/components/src/Cluster/Cluster.module.css
taylorvnoj: Amazing, that is going to be so helpful!

packages/site/src/content/Cover/Cover.mdx
taylorvnoj: Stack will also show up as Stack5 lol - this can definitely be a follow up ticket imo

packages/site/src/content/Stack/Stack.mdx
```diff
@@ -0,0 +1,84 @@
...
+
+### Basic Stack
+
+<Canvas>
+  <Stack>
+    <Text>First item</Text>
+    <Text>Second item</Text>
+    <Text>Third item</Text>
+  </Stack>
+</Canvas>
```
nad182, line 35: For some reason, on the new doc site the component is called `Stack5` in all examples:
image

packages/components/src/Stack/types.ts
taylorvnoj: thanks Scott! I didn't see any more prop differences 👍🏻

packages/components/src/SideKick/SideKick.tsx
taylorvnoj: Don't want to speak for Scott but I would suggest we don't restrict. The example in that screenshot is a layout that will be in every object in product - in fact there will be situations where there are multiple children on the right and SideKick will give us the responsiveness in one go, rather than having to account for an unknown amount of children

packages/components/src/Stack/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  /** The number of children to split the stack after (1-15). Requires parent to have height greater than the sum of the children. */
+  readonly splitAfter?:
```
nad182, line 8: I think it would be great to give a bit more explanation for this prop. It's not clear straight away that by splitting we mean pushing the rest of the content to the bottom.

packages/components/src/SideKick/SideKick.tsx
nad182: I understand that we might need to have more elements, but it seems like the proper layout would be to wrap ALL main elements in one container and all supplementary in another:

image

packages/site/src/content/Cover/Cover.mdx
scotttjob: Yes I very much want to fix this! I'll make a note to create that ticket, and maybe I'll even dig in myself today, I want these examples to be usable without modification.

packages/site/src/content/Cluster/Cluster.mdx
scotttjob: Or possibly just change how they work + add more props. I'm still not 100% sold on the direction for those components, so that's why they're left out of these new ones related components.

packages/components/src/SideKick/types.ts
```diff
@@ -0,0 +1,17 @@
...
+  /** The breakpoint to collapse the sidekick at. */
+  readonly collapseBelow?: string;
```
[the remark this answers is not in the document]
scotttjob: Based on feedback from Eric around the potential performance implications of `useMediaQuery` on potentially a lot of layout components on the same page, this has been modified to no longer support arbitrary values (semantic tokens only tied to specific breakpoints).

That said, had we kept that feature I probably would not have strongly typed that value beyond a string, or possibly a number. I might have provided more guidance in the JSDoc about what types of values are supported, or I would have tried to find a DOM based type that just supported all the values that a 'width' value would support. My intention was for this to be extremely flexible.

Even though I removed that flexibility from within the component, I did preserve the original functionality of 'break when you want to' by providing an override 'collapsed' prop to the component, which is up to the consumer to tell us when to break, so they can set up their own media query if need be (or break on other arbitrary reasons).

packages/site/src/content/Stack/Stack.mdx
scotttjob: Commented above on the other item, we've got an issue with how this component (RenderComponentShowCode) maps react components back to their originals. We may just have to setup an override map.

packages/components/src/Stack/types.ts
scotttjob: Done! (in the follow-up branch, targeting this one, which will be pushed up and created shortly)

taylorvnoj (approved):
THANK YOU for separating out the feedback updates and thank you for all of this incredible work! 
This is in a great state to release and we can continue to iterate  👌🏻

PR #2470, opened 2025-04-08
feat(components): Add ref to Button and Chip

ZakaryH:
## Motivations

With React Hook Forms, there's a feature where if there's an error on a field it will scroll to that input, and also focus it for the user.

If a Combobox were to be used in such a way, we don't have a way to offer the ref of it's activator, thus blocking that behavior in the ideal way anyway. 

To enable this behavior, we're going to expose a ref for Chip. It's a bit odd because Chip can be either a div or a button, and I'm not too sure what the use case would even be for needing a ref to the div.

Button is a reasonable other usage for a Combobox activator, so I'm preemptively adding a ref to that too. Button I've omitted the ref prop when the `to` prop is provided because that is deprecated and going away. I've also omitted it when `url` is provided, making it into an `a` tag. Perhaps that's not right though, and it would be valid to expose the ref to the anchor tag.

## Changes

updated the meta.test.tsx helper function because now that Chip is both a forwarded ref component AND it has child components (Chip.Prefix & Chip.Suffix) we need to apply some recursive searching like we do in other components.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -25,6 +25,7 @@ function ButtonWrapper(props: ButtonProps) {
     ariaLabel,
     disabled = false,
     external,
+    htmlButtonRef,
     id,
     name,
     onClick,
@@ -62,6 +63,7 @@ function ButtonWrapper(props: ButtonProps) {
     "aria-label": ariaLabel,
     role: role,
   };
+  const refProp = url ? {} : { ref: htmlButtonRef };
 
   const buttonInternals = children || <ButtonContent {...props} />;
 
@@ -73,9 +75,15 @@ function ButtonWrapper(props: ButtonProps) {
     );
   }
 
-  const Tag = url ? "a" : "button";
+  if (url) {
+    return <a {...tagProps}>{buttonInternals}</a>;
+  }
 
-  return <Tag {...tagProps}>{buttonInternals}</Tag>;
+  return (
+    <button {...tagProps} {...refProp}>
+      {buttonInternals}
+    </button>
+  );
 }
 
 Button.Label = ButtonLabel;
--- a/packages/components/src/Button/Button.types.ts
+++ b/packages/components/src/Button/Button.types.ts
@@ -121,6 +121,8 @@ interface BasicButtonProps extends ButtonFoundationProps {
    * Used to override the default button role.
    */
   readonly role?: string;
+
+  readonly htmlButtonRef?: React.Ref<HTMLButtonElement>;
 }
 
 type BaseButtonProps = XOR<
--- a/packages/components/src/Chip/Chip.tsx
+++ b/packages/components/src/Chip/Chip.tsx
@@ -9,56 +9,58 @@ import { useChildComponent } from "./hooks/useChildComponent";
 import { Typography } from "../Typography";
 import { Tooltip } from "../Tooltip";
 
-export const Chip = ({
-  ariaLabel,
-  disabled,
-  heading,
-  invalid,
-  label,
-  value,
-  testID,
-  onClick,
-  onKeyDown,
-  children,
-  role = "button",
-  tabIndex = 0,
-  variation = "base",
-}: ChipProps): JSX.Element => {
-  const classes = classnames(styles.chip, {
-    [styles.invalid]: invalid,
-    [styles.base]: variation === "base",
-    [styles.subtle]: variation === "subtle",
-    [styles.disabled]: disabled,
-  });
+// Define a type for the Chip component with the properties we add to it
+type ChipComponent = React.ForwardRefExoticComponent<
+  ChipProps & React.RefAttributes<HTMLButtonElement | HTMLDivElement>
+> & {
+  Prefix: typeof ChipPrefix;
+  Suffix: typeof ChipSuffix;
+  displayName: string;
+};
 
-  const prefix = useChildComponent(children, d => d.type === Chip.Prefix);
-  const suffix = useChildComponent(children, d => d.type === Chip.Suffix);
+export const Chip = React.forwardRef<
+  HTMLButtonElement | HTMLDivElement,
+  ChipProps
+>(
+  (
+    {
+      ariaLabel,
+      disabled,
+      heading,
+      invalid,
+      label,
+      value,
+      testID,
+      onClick,
+      onKeyDown,
+      children,
+      role = "button",
+      tabIndex = 0,
+      variation = "base",
+    },
+    ref,
+  ) => {
+    const classes = classnames(styles.chip, {
+      [styles.invalid]: invalid,
+      [styles.base]: variation === "base",
+      [styles.subtle]: variation === "subtle",
+      [styles.disabled]: disabled,
+    });
 
-  const [labelRef, labelFullyVisible] = useInView<HTMLSpanElement>();
-  const [headingRef, headingFullyVisible] = useInView<HTMLSpanElement>();
-  const tooltipMessage = getTooltipMessage(
-    labelFullyVisible,
-    headingFullyVisible,
-    label,
-    heading,
-  );
-  const Tag = onClick ? "button" : "div";
+    const prefix = useChildComponent(children, d => d.type === Chip.Prefix);
+    const suffix = useChildComponent(children, d => d.type === Chip.Suffix);
 
-  return (
-    <Tooltip message={tooltipMessage} setTabIndex={false}>
-      <Tag
-        className={classes}
-        onClick={(ev: React.MouseEvent<HTMLButtonElement | HTMLDivElement>) =>
-          onClick?.(value, ev)
-        }
-        tabIndex={disabled ? -1 : tabIndex}
-        onKeyDown={onKeyDown}
-        aria-label={ariaLabel}
-        disabled={disabled}
-        role={role}
-        data-testid={testID}
-        type="button"
-      >
+    const [labelRef, labelFullyVisible] = useInView<HTMLSpanElement>();
+    const [headingRef, headingFullyVisible] = useInView<HTMLSpanElement>();
+    const tooltipMessage = getTooltipMessage(
+      labelFullyVisible,
+      headingFullyVisible,
+      label,
+      heading,
+    );
+
+    const chipContent = (
+      <>
         {prefix}
         <div className={styles.chipContent}>
           {heading && (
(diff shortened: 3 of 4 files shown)


packages/components/src/Button/Button.types.ts
```diff
@@ -121,6 +121,8 @@ interface BasicButtonProps extends ButtonFoundationProps {
    * Used to override the default button role.
    */
   readonly role?: string;
+
+  readonly htmlButtonRef?: React.Ref<HTMLButtonElement>;
```
ZakaryH, line 125: not in love with this name

the reason I've gone with this, and not `ref` via a forwarded ref is because in my mind there _could_ be a distinction to be made between a ref that is associated with the html `button` directly, which is this case

and then another where perhaps we start using an imperativeHandler to add specific Atlantis Button API methods. maybe something like `shake()` who knows.

that said we currently don't offer anything like that

packages/components/src/Chip/Chip.tsx
```diff
@@ -84,10 +86,47 @@ export const Chip = ({
...
+                className: classes,
+                tabIndex: disabled ? -1 : tabIndex,
+                onKeyDown,
+                "aria-label": ariaLabel,
+                role,
+                "data-testid": testID,
+                ref,
+              },
+              chipContent,
+            )}
```
ZakaryH, line 125: this isn't a 1:1

we're now missing the `onClick`. even if that's questionable to have for a div, I am not trying to refactor it or remove functionality to that degree right now.

omitting `type: "button"` makes sense but is also not a 1:1

`disabled` is another weird one to apply to a div, but it was doing that before so perhaps I should continue

somewhat surprised the `onClick` isn't upset about the type of the event

ZakaryH:
I'm not going to address it in this PR but there's another possible problem with the Combobox setup

right now if you want the default Chip trigger, you completely omit `Combobox.Activator` however that's the element we'd want a `ref` to. so where does the `ref` go? on the `Combobox` top level component but as something like `defaultActivatorRef`?

if you're doing a non default activator, you can provide the component and you'd have no need for us to expose the `ref` on `Combobox.Activator` - you can just do it yourself

that _might_ be the way to proceed, if we like it that is. where if you want to provide the ref, you must use a custom activator. it keeps things cleaner, and it is a more advanced, and arguably custom setup

packages/site/src/content/Chip/Chip.props.json
ZakaryH, line 1: this doesn't seem right. why is the entire props.json being deleted?

packages/components/src/Button/Button.types.ts
ZakaryH: I suppose we could still enhance `ref` with more capabilities if we wanted, and then implement a `ref.focus()` which would do the same thing.

that's probably better now that I think about it more. the only thing there is that we've called it `inputRef` on inputs. I wonder why not use `ref` directly for those instances.

packages/site/src/content/Chip/Chip.props.json
ZakaryH, line 1: after fighting with this for quite some time, this is the best state I've been able to manage thus far

it does result in us losing the props for Prefix and Suffix, well really just Suffix since Prefix has nothing

image

packages/components/src/Chip/Chip.types.ts
ZakaryH, line 1: I changed this from `Chip.types.d.ts` to `Chip.types.ts` as part of my troubleshooting but I think I'm going to leave it like this

my understanding is that the `d.ts` version is the exception, not the norm which lines up with how we have all our other type files named. if I'm missing something though I can change it back.

packages/components/src/Chip/Chip.tsx
ZakaryH: wait, I'm being silly. of course there's no `onClick` for `div`, the logic of `Tag` specifically only renders a `div` if there is no `onClick` so that part is fine. `type: "button"` also wouldn't make sense for a `div`.

`disabled` is debatable. I suppose it would depend on if we're targeting that attribute for styles, that might be a valid reason to have it on a div.

packages/components/src/Button/Button.tsx
```diff
@@ -1,83 +1,115 @@
...
+        <ButtonWrapper {...props} ref={ref} />
+      </ButtonProvider>
     );
-  }
+  },
+);
+ButtonComponent.displayName = "Button";
+
+const ButtonWrapper = forwardRef<ButtonRefElement, ButtonProps>(
+  // eslint-disable-next-line max-statements
```
ZakaryH, line 27: it was specifically upset about an arrow function

could try changing this to a named function and perhaps it will be happy, sparing us from disabling eslint

packages/components/src/Button/Button.tsx
```diff
@@ -1,83 +1,115 @@
...
+      id,
+      style: UNSAFE_style.container,
+      ...(!disabled && { onClick: onClick }),
+      ...(!disabled && { onMouseDown: onMouseDown }),
+      "aria-controls": ariaControls,
+      "aria-haspopup": ariaHaspopup,
+      "aria-expanded": ariaExpanded,
+      "aria-label": ariaLabel,
+      role: role,
+    };
```
ZakaryH, line 69: if you look closely you'll notice that `url`, `external`, `submit`, and `href` conditional attribute assignment is no longer happening in this object that is now called `commonProps`

instead, those specific props/attributes are being dealt with in their respective code blocks

personally I think this is easier to read, follow and reason about - but it's not a necessary change. it could be left as it was.

packages/components/src/Button/Button.tsx
```diff
@@ -1,83 +1,115 @@
...
+
+    const buttonInternals = children || <ButtonContent {...props} />;
+
+    if (to) {
+      // Props specific to Link
+      const linkProps = {
+        ...commonProps,
+        to: to as LinkProps["to"],
+      };
```
ZakaryH, line 78: I've omitted the ref here since `to` is deprecated but maybe this is just making it more confusing than it's worth since the type won't tell you that in this scenario the ref won't be attached to anything. it'll appear valid but do nothing.

ZakaryH:
here's a demo of using Zod + RHF + custom activator with a Chip ref on Combobox


https://github.com/user-attachments/assets/43a29e39-5e2a-4c00-85e8-464f278034de

try to ignore some of the jank, it's an imperfect example that is getting upset with the types on select.

and a snippet of the code to implement it minus all the zod/RHF setup

```
<Controller
              control={control}
              name="country"
              render={({ field, fieldState }) => (
                <>
                <Combobox
                  selected={country}
                  onSelect={(option) => {
                    setCountry(option);
                    field.onChange(option);
                  }}
                >
                  <Combobox.Activator>
                    <Chip label="Press Me" ref={field.ref} tabIndex={0}/>
                  </Combobox.Activator>
              {countryOptions.map(option => (
                <Combobox.Option key={option.value} label={option.label} id={option.value}/>
              ))}
              </Combobox>
                {fieldState.error && <strong style={{ color: "red" }}>{fieldState.error.message}</strong>}
                </>
              )}/>
```

packages/components/src/Chip/Chip.tsx
```diff
@@ -9,56 +9,46 @@ import { useChildComponent } from "./hooks/useChildComponent";
...
+      [styles.invalid]: invalid,
+      [styles.base]: variation === "base",
+      [styles.subtle]: variation === "subtle",
+      [styles.disabled]: disabled,
+    });
 
-  const prefix = useChildComponent(children, d => d.type === Chip.Prefix);
-  const suffix = useChildComponent(children, d => d.type === Chip.Suffix);
+    const prefix = useChildComponent(children, d => d.type === ChipPrefix);
+    const suffix = useChildComponent(children, d => d.type === ChipSuffix);
```
ZakaryH, line 39: this should be equivalent since `Chip.Prefix` was literally `ChipPrefix` before, however now with the change in how we're setting up the namespace we can't yet reference Chip.Prefix.

packages/components/src/Chip/Chip.tsx
```diff
@@ -9,56 +9,46 @@ import { useChildComponent } from "./hooks/useChildComponent";
...
 
-  const [labelRef, labelFullyVisible] = useInView<HTMLSpanElement>();
-  const [headingRef, headingFullyVisible] = useInView<HTMLSpanElement>();
-  const tooltipMessage = getTooltipMessage(
-    labelFullyVisible,
-    headingFullyVisible,
-    label,
-    heading,
-  );
-  const Tag = onClick ? "button" : "div";
```
ZakaryH, line 45: this might have been nice at one point but I feel like it's more confusing than the benefits it provides since it looks like we add all the same props to `<Tag>` in both cases but we're not exactly doing that in practice.

packages/components/src/Button/Button.tsx
edison-cy-yang: Nice! This is so much cleaner than before with the different props that get set before its respective element code block is reached.

packages/site/src/content/Button/Button.props.json
```diff
@@ -330,6 +330,32 @@
...
+        "parent": {
+          "fileName": "atlantis/node_modules/@types/react/index.d.ts",
+          "name": "RefAttributes"
+        },
+        "required": false,
+        "type": {
+          "name": "Ref<ButtonRefElement>"
+        }
+      },
+      "key": {
```
edison-cy-yang, line 347: Not a blocker, but I wonder if we can exclude generating this `key` prop. I'm assuming our doc generation script picks this up when component's `forwardRef` exposes React's internal structure. Having the key prop on our docs doesn't really add values

packages/components/src/Chip/Chip.tsx
edison-cy-yang: Nice one! Same with the Button's Tag usage, this is cleaner to read and more explicitly applying the corresponding props to the relevant element.

edison-cy-yang (commented):
I really like the breakdown of anchor/button and div/button into its own code block & tsx - easier to understand.

I've personally only touched `Button` before and not `Chip`. However, everything looks good to me from a 1:1 standpoint with the addition of `ref`, other than the ones in `Chip`, but they make sense. 

I just have one question on the generated `key` prop for our docs, otherwise I'm ready to approve.

packages/site/src/content/Button/Button.props.json
ZakaryH: ah sorry I missed this question, I'm not sure if there's a way to do so. this is all coming from the react docgen script.

I'm not seeing a way to easily omit these props that are coming from `forwardRef`. would you be alright with merging as-is?

PR #2471, opened 2025-04-08
feat(components): Add ability to only open Chips menu on search [JOB-119726]

Aiden-Brine:
## Motivations

The motivation was to achieve the A/C of this ticket, essentially allowing Chips to be used as an email selection input that has similar functionality to GMail. I'll admit, I don't feel fantastic about these changes as they add more complexity to already quite complex component. I had a conversation with though and we agreed that it made sense to add these additions to Chips for now and then look into building a new component, potentially using autocomplete, to achieve this functionality in the future without relying on Chips. In the meantime though I tried really hard to keep the new logic as simple as possible and add new testing coverage/improve the tests that were existing before.

### Added
Adds two new props:
- `onlyShowMenuOnSearch`: If this is `false` (the default) when you click on the activator the menu will appear immediately to provide the user with options to select. When set to `true` the menu of options will not appear until you start typing in a search value. When the input is cleared, the menu disappears again.
- `submitInputOnFocusShift`: When `false` (the default) when a user has added a search term to the input and then shifts focus away by clicking outside of the input the search term will be cleared. When `true` instead of clearing the search term it will submit the best available value based on the search term. That value is:
  - An exact match if present
  - A new custom value if custom values are allowed
  - The closest match if custom values are not allowed

I didn't add more docs that the prop docstrings. If anyone thinks I should add anything else into the docs please let me know. 

### Changed

- There should be no changes to existing functionality

## Testing

The story for this is kind for this functionality is pretty borked but I have a pre-release of these changes on [this branch](https://github.com/GetJobber/jobber-frontend/pull/2571) and you can test them out there. Also worth double checking that none of the other uses of Chips are broken

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Chips/ChipsTypes.tsx
+++ b/packages/components/src/Chips/ChipsTypes.tsx
@@ -112,6 +112,8 @@ export interface ChipDismissibleProps extends ChipFoundationProps {
 
   /**
    * If true, automatically selects an option based on the current search value when the input loses focus.
+   * The automatic selection order is: an exact match of the search value if available, a custom option if
+   * onCustomOptionSelect is provided, or the closest match.
    * @default false
    */
   readonly submitInputOnFocusShift?: boolean;
--- a/packages/components/src/Chips/ChipsTypes.tsx
+++ b/packages/components/src/Chips/ChipsTypes.tsx
@@ -103,6 +103,18 @@ export interface ChipDismissibleProps extends ChipFoundationProps {
    * @param searchValue - The input value
    */
   onLoadMore?(searchValue: string): void;
+
+  /**
+   * Control whether the menu only appears once the user types.
+   * @default false
+   */
+  readonly onlyShowMenuOnSearch?: boolean;
+
+  /**
+   * Select the last available option when focus shifts away from the input.
+   * @default false
+   */
+  readonly submitInputOnFocusShift?: boolean;
 }
 
 export type ChipsProps =
--- a/packages/components/src/Chips/ChipsTypes.tsx
+++ b/packages/components/src/Chips/ChipsTypes.tsx
@@ -116,7 +116,7 @@ export interface ChipDismissibleProps extends ChipFoundationProps {
    * onCustomOptionSelect is provided, or the closest match.
    * @default false
    */
-  readonly submitInputOnFocusShift?: boolean;
+  readonly autoSelectOnClickOutside?: boolean;
 }
 
 export type ChipsProps =
--- a/packages/components/src/Chips/ChipsTypes.tsx
+++ b/packages/components/src/Chips/ChipsTypes.tsx
@@ -111,7 +111,7 @@ export interface ChipDismissibleProps extends ChipFoundationProps {
   readonly onlyShowMenuOnSearch?: boolean;
 
   /**
-   * Select the last available option when focus shifts away from the input.
+   * If true, automatically selects an option based on the current search value when the input loses focus.
    * @default false
    */
   readonly submitInputOnFocusShift?: boolean;
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
@@ -1,6 +1,5 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect } from "react";
 import classNames from "classnames";
-import debounce from "lodash/debounce";
 import { useSafeLayoutEffect } from "@jobber/hooks/useSafeLayoutEffect";
 import styles from "./InternalChipDismissible.module.css";
 import { ChipDismissibleInputProps } from "./InternalChipDismissibleTypes";
@@ -13,8 +12,6 @@ import {
 import { Text } from "../../Text";
 import { Button } from "../../Button";
 import { Spinner } from "../../Spinner";
-
-const DEBOUNCE_TIME = 200;
 
 // eslint-disable-next-line max-statements
 export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
@@ -35,6 +32,7 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     menuId,
     menuOpen,
     searchValue,
+    showInput,
     generateDescendantId,
     handleBlur,
     handleOpenMenu,
@@ -44,12 +42,10 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     handleSetActiveOnMouseOver,
     handleKeyDown,
     handleSelectOption,
+    handleShowInput,
     handleDebouncedSearch,
   } = useInternalChipDismissibleInput(props);
 
-  // Controls whether the input field or the activator button is rendered.
-  // This state is only used when `onlyShowMenuOnSearch` is true.
-  const [showInput, setShowInput] = useState(false);
   const menuRef = useScrollToActive(activeIndex);
   const { ref: visibleChildRef, isInView } = useInView<HTMLDivElement>();
 
@@ -73,41 +69,11 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     isInView && onLoadMore && onLoadMore(searchValue);
   }, [isInView]);
 
-  if (onlyShowMenuOnSearch) {
-    if (!showInput) {
-      const handleActivate = () => {
-        setShowInput(true);
-        setTimeout(() => inputRef.current?.focus(), 0);
-      };
-
-      return React.cloneElement(activator, { onClick: handleActivate });
-    }
-  } else {
-    if (!menuOpen) {
-      return React.cloneElement(activator, { onClick: handleOpenMenu });
-    }
+  if (!showInput) {
+    return React.cloneElement(activator, { onClick: handleShowInput });
   }
 
-  const handleInputBlur = () => {
-    if (onlyShowMenuOnSearch) {
-      const valueBeforeBlur = inputRef.current?.value;
-
-      handleBlur();
-
-      setTimeout(() => {
-        if (valueBeforeBlur === "") {
-          setShowInput(false);
-        }
-      }, DEBOUNCE_TIME);
-    } else {
-      debounce(handleBlur, DEBOUNCE_TIME)();
-    }
-  };
-
-  const shouldShowMenu =
-    menuOpen &&
-    (hasAvailableOptions || isLoadingMore) &&
-    (!onlyShowMenuOnSearch || !!searchValue);
+  const shouldShowMenu = menuOpen && (hasAvailableOptions || isLoadingMore);
 
   return (
     <>
@@ -124,7 +90,7 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
         value={searchValue}
         onChange={handleSearchChange}
         onKeyDown={handleKeyDown}
-        onBlur={handleInputBlur}
+        onBlur={() => setTimeout(handleBlur, 200)}
         onFocus={onlyShowMenuOnSearch ? undefined : handleOpenMenu}
         autoFocus={true}
       />
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
@@ -13,7 +13,6 @@ import { Text } from "../../Text";
 import { Button } from "../../Button";
 import { Spinner } from "../../Spinner";
 
-// eslint-disable-next-line max-statements
 export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
   const {
     activator = <Button icon="add" type="secondary" ariaLabel="Add" />,
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
@@ -1,6 +1,7 @@
-import React, { useEffect } from "react";
-import debounce from "lodash/debounce";
+/* eslint-disable max-statements */
+import React, { useEffect, useState } from "react";
 import classNames from "classnames";
+import debounce from "lodash/debounce";
 import { useSafeLayoutEffect } from "@jobber/hooks/useSafeLayoutEffect";
 import styles from "./InternalChipDismissible.module.css";
 import { ChipDismissibleInputProps } from "./InternalChipDismissibleTypes";
@@ -14,6 +15,8 @@ import { Text } from "../../Text";
 import { Button } from "../../Button";
 import { Spinner } from "../../Spinner";
 
+const DEBOUNCE_TIME = 200;
+
 export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
   const {
     activator = <Button icon="add" type="secondary" ariaLabel="Add" />,
@@ -21,6 +24,7 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     isLoadingMore = false,
     onLoadMore,
     options,
+    onlyShowMenuOnSearch = false,
   } = props;
 
   const {
@@ -43,6 +47,7 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     handleDebouncedSearch,
   } = useInternalChipDismissibleInput(props);
 
+  const [showInput, setShowInput] = useState(false);
   const menuRef = useScrollToActive(activeIndex);
   const { ref: visibleChildRef, isInView } = useInView<HTMLDivElement>();
 
@@ -66,9 +71,31 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     isInView && onLoadMore && onLoadMore(searchValue);
   }, [isInView]);
 
-  if (!menuOpen) {
+  if (onlyShowMenuOnSearch && !showInput) {
+    const handleActivate = () => {
+      setShowInput(true);
+      setTimeout(() => inputRef.current?.focus(), 0);
+    };
+
+    return React.cloneElement(activator, { onClick: handleActivate });
+  } else if (!onlyShowMenuOnSearch && !menuOpen) {
     return React.cloneElement(activator, { onClick: handleOpenMenu });
   }
+
+  const handleInputBlur = () => {
+    handleBlur();
+
+    setTimeout(() => {
+      if (inputRef.current?.value === "") {
+        setShowInput(false);
+      }
+    }, DEBOUNCE_TIME);
+  };
+
+  const shouldShowMenu =
+    menuOpen &&
+    (hasAvailableOptions || isLoadingMore) &&
+    (!onlyShowMenuOnSearch || !!searchValue);
 
   return (
     <>
@@ -80,17 +107,21 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
         aria-label="Press up and down arrow to cycle through the options or type to narrow down the results"
         aria-autocomplete="list"
         aria-owns={menuId}
-        aria-expanded={hasAvailableOptions}
+        aria-expanded={shouldShowMenu}
         aria-activedescendant={generateDescendantId(activeIndex)}
         value={searchValue}
         onChange={handleSearchChange}
         onKeyDown={handleKeyDown}
-        onBlur={debounce(handleBlur, 200)}
-        onFocus={handleOpenMenu}
+        onBlur={
+          onlyShowMenuOnSearch
+            ? handleInputBlur
+            : debounce(handleBlur, DEBOUNCE_TIME)
+        }
+        onFocus={!onlyShowMenuOnSearch ? handleOpenMenu : undefined}
         autoFocus={true}
       />
 
-      {(hasAvailableOptions || isLoadingMore) && (
+      {shouldShowMenu && (
         <div
           ref={setPositionedElementRef}
           className={styles.menu}
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
@@ -73,9 +73,7 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     isInView && onLoadMore && onLoadMore(searchValue);
   }, [isInView]);
 
-  // Conditionally render the activator button or the input section.
   if (onlyShowMenuOnSearch) {
-    // Mode: Only show menu after typing.
     if (!showInput) {
       const handleActivate = () => {
         setShowInput(true);
@@ -84,34 +82,24 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
 
       return React.cloneElement(activator, { onClick: handleActivate });
     }
-    // If showInput is true, fall through to render the input section below.
   } else {
-    // Mode: Show menu on focus/click (default).
     if (!menuOpen) {
       return React.cloneElement(activator, { onClick: handleOpenMenu });
     }
-    // If menuOpen is true, fall through to render the input section below.
   }
 
-  // If we didn't return an activator above, render the input and potentially the menu.
   const handleInputBlur = () => {
     if (onlyShowMenuOnSearch) {
-      // Mode: Only show menu after typing.
-
-      // Capture input value BEFORE calling the hook's blur
       const valueBeforeBlur = inputRef.current?.value;
 
-      // Allow hook to reset menu state immediately
       handleBlur();
 
       setTimeout(() => {
-        // Check the value *as it was when blur occurred*
         if (valueBeforeBlur === "") {
-          setShowInput(false); // This will trigger rerender with activator
+          setShowInput(false);
         }
       }, DEBOUNCE_TIME);
     } else {
-      // Mode: Show menu on focus/click (default).
       debounce(handleBlur, DEBOUNCE_TIME)();
     }
   };
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable max-statements */
 import React, { useEffect, useState } from "react";
 import classNames from "classnames";
 import debounce from "lodash/debounce";
@@ -17,6 +16,7 @@ import { Spinner } from "../../Spinner";
 
 const DEBOUNCE_TIME = 200;
 
+// eslint-disable-next-line max-statements
 export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
   const {
     activator = <Button icon="add" type="secondary" ariaLabel="Add" />,
@@ -47,6 +47,8 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     handleDebouncedSearch,
   } = useInternalChipDismissibleInput(props);
 
+  // Controls whether the input field or the activator button is rendered.
+  // This state is only used when `onlyShowMenuOnSearch` is true.
   const [showInput, setShowInput] = useState(false);
   const menuRef = useScrollToActive(activeIndex);
   const { ref: visibleChildRef, isInView } = useInView<HTMLDivElement>();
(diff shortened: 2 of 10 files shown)


packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissibleInput.test.tsx
```diff
@@ -67,16 +104,19 @@ describe("Menu open", () => {
   });
 
   it("should not show the add button", () => {
-    expect(addButton).not.toBeInTheDocument();
+    // Button role might still exist if input is also a button, query specifically
+    expect(
+      screen.queryByRole("button", { name: "Add" }),
```
Aiden-Brine, line 109: By re-querying the DOM to ensure no add buttons are present this ensures the add button referenced by `addButton` wasn't removed and replaced by a new add button.

packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissibleInput.test.tsx
```diff
@@ -67,16 +104,19 @@ describe("Menu open", () => {
...
+    ).not.toBeInTheDocument();
   });
 
   it("should show an input and menu", () => {
-    expect(screen.queryByRole("combobox")).toBeInTheDocument();
-    expect(screen.queryByRole("listbox")).toBeInTheDocument();
+    expect(screen.getByRole("combobox")).toBeInTheDocument();
```
Aiden-Brine, line 114: On production the following error is present on all the calls of `screen.getByRole`:
```
Argument of type 'HTMLElement | null' is not assignable to parameter of type 'Element | Node | Document | Window'.
  Type 'null' is not assignable to type 'Element | Node | Document | Window'.ts(2345)
```
`getByRole` fails if the query fails so always returns an element

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -14,13 +14,17 @@ import { Text } from "../../Text";
 import { Button } from "../../Button";
 import { Spinner } from "../../Spinner";
 
+const DEBOUNCE_TIME = 200;
+
+// eslint-disable-next-line max-statements
```
Aiden-Brine, line 19: I don't think there is a way to avoid this without a decently significant refactor and I was _trying_ to be as unobtrusive as possible. I know adding to Chips is not ideal and pushing it over its max statements is definitely not ideal. I think it is the lesser of two evils though. If anyone would prefer I do the refactor though please let me know and I can dive in on that.

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -43,6 +47,9 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     handleDebouncedSearch,
   } = useInternalChipDismissibleInput(props);
 
+  // Controls whether the input field or the activator button is rendered.
```
Aiden-Brine, line 50: I was trying to strike the balance both here and in the tests of having the right amount of comments but if you think any of them are not helpful I am happy to remove them

packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissibleInput.test.tsx
```diff
@@ -171,22 +208,202 @@ describe("Activator", () => {
...
 function isOptionhighlighted(highlighted: string) {
   const activeClass = "activeOption";
-  expect(screen.getByRole("option", { name: highlighted })).toHaveClass(
-    activeClass,
-  );
+  const option = screen.getByRole("option", { name: highlighted });
+  expect(option).toHaveClass(activeClass);
+
+  const input = screen.getByRole("combobox");
+  expect(input).toHaveAttribute("aria-activedescendant", option.id);
```
Aiden-Brine, line 398: Added a test for the aria attribute along with just the CSS class

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -66,10 +73,42 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     isInView && onLoadMore && onLoadMore(searchValue);
   }, [isInView]);
 
-  if (!menuOpen) {
-    return React.cloneElement(activator, { onClick: handleOpenMenu });
+  if (onlyShowMenuOnSearch) {
+    if (!showInput) {
+      const handleActivate = () => {
+        setShowInput(true);
+        setTimeout(() => inputRef.current?.focus(), 0);
```
ZakaryH, line 80: I'm assuming the `setTimeout(..., 0)` is to try to ensure the `inputRef.current` exists?

I'll have to think about this a bit more, but maybe there's a sensible `useEffect` we could put this in instead - even though avoiding `useEffect`s is _usually_ something to aim for.

packages/components/src/Chips/ChipsTypes.tsx
```diff
@@ -103,6 +103,20 @@ export interface ChipDismissibleProps extends ChipFoundationProps {
    * @param searchValue - The input value
    */
   onLoadMore?(searchValue: string): void;
+
+  /**
+   * Control whether the menu only appears once the user types.
+   * @default false
+   */
+  readonly onlyShowMenuOnSearch?: boolean;
```
ZakaryH, line 111: it's mostly a personal preference where I really like short, succinct props that makes me wonder if we couldn't change this to work off an enum style prop like....

```
enum MenuVisibility {
  FOCUS = "FOCUS",
  SEARCH = "SEARCH",
}

MenuVisibility.FOCUS;

/**
* @default "FOCUS"
*/
readonly showMenu?: MenuVisibility;
...

menuVisibility={"SEARCH"}

```

maybe an enum isn't the right tool since we'd have to export it. a regular ol' string like we do with the `Clearable` type might work fine.

maybe "search present" would be more descriptive too.

packages/components/src/Chips/ChipsTypes.tsx
```diff
@@ -103,6 +103,20 @@ export interface ChipDismissibleProps extends ChipFoundationProps {
...
+
+  /**
+   * If true, automatically selects an option based on the current search value when the input loses focus.
+   * The automatic selection order is: an exact match of the search value if available, a custom option if
+   * onCustomOptionSelect is provided, or the closest match.
+   * @default false
+   */
+  readonly submitInputOnFocusShift?: boolean;
```
ZakaryH, line 119: hmmm, losing focus could traditionally be described as `onBlur` do you think that would make sense here? 

that aside, this is a pretty complicated prop. I don't have any immediate suggestions for how to improve. I'll continue checking things out and come back to this later.

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
Aiden-Brine: Yes indeed, it ensures `inputRef.current?.focus` is called in the next event loop cycle after all of the rendering has completed. I think [this](https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful) is a decently common practice and not concerning and that this is an acceptable time to use it. But if you come up with a better solution that uses `useEffect` instead I am definitely open to changing it.

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -66,10 +73,42 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
...
+
+      handleBlur();
+
+      setTimeout(() => {
+        if (valueBeforeBlur === "") {
+          setShowInput(false);
+        }
+      }, DEBOUNCE_TIME);
+    } else {
+      debounce(handleBlur, DEBOUNCE_TIME)();
```
ZakaryH, line 103: does this work? typically using `debounce` directly in React components has a bad time due to re-rendering and functions being remade in such a way that `debounce` doesn't think the same function is being called.

we have a handy `useDebounce` hook that handles all of the React stuff

packages/components/src/Chips/ChipsTypes.tsx
Aiden-Brine: I don't think these new props are immediately intuitive. I think that speaks to the smell that we have already talked about that comes from overloading Chips further. My approach to this was making the prop names very verbose to try to communicate their meaning as much as possible. In an ideal world that would be a sign that this component needs to be refactored. We both know it does and that it won't be happening right now. 

I don't feel terribly strongly about this. If as an external person coming in who hasn't be working with these props for a few days you thinking simpler names would be more intuitive for users I am happy to change them. Naming is hard

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
Aiden-Brine: Screenshot 2025-04-11 at 2 57 11 PM

This is actually the current behaviour on prod. I didn't pay it much attention other than making sure if `onlyShowMenuOnSearch` is false the behaviour remains the same. I can do some testing though to see if there is a bug in prod that could be improved by this PR 😄

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -66,10 +73,42 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
...
+        setShowInput(true);
+        setTimeout(() => inputRef.current?.focus(), 0);
+      };
+
+      return React.cloneElement(activator, { onClick: handleActivate });
+    }
+  } else {
+    if (!menuOpen) {
+      return React.cloneElement(activator, { onClick: handleOpenMenu });
+    }
```
ZakaryH, line 88: these two paths end up doing extremely similar things, where it's really just the `onClick` that differs

I wonder if we could keep this relatively simple, and introduce a new `handleClick` function that we stuff all this logic into

```
function handleClick() {
// all the same logic that would determine if we call handleActivate or handleOpenMenu
}
```

that should also be safer because we'd be using the ref inside a callback. wait, actually it's already inside a callback. do we even need the `setTimeout`?

ZakaryH: ok yes we do need `setTimeout` with that setup, and moving into a handler doesn't work. we'd need to do some other adjustments.

also I'm only now noticing how this works. I'm not a huge fan of the fact that this component will either return a cloned activator element OR it will return an entirely different piece of UI (the input)

I feel like that should not be the responsibility of this component. one level up should be deciding to show either the activator or the input.

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
ZakaryH: actually hold on, why do we need to all `inputRef.current?.focus()` in the first place?

the input attached to `inputRef` has `autofocus={true}` so isn't that just going to focus it as soon as it's rendered anyway?

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
Aiden-Brine: That is totally fair, I agree with that. That was the behaviour before I started making changes but that doesn't make it correct. It certainly took me a while to wrap my head around how it worked before I started making changes. I approached my changes from the approach of "what are the smallest amount of changes I can make to Chips to get this functionality to work" and maybe I should have said "Are there fundamental changes I can make to chips first to make my changes easier?" I can admit I had some tunnel vision of "I just need to change this and this to make it work.. oh and this...oh and this..."

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
ZakaryH: ok yeah that all tracks that this would be doing this on prod, and it's replicating that behavior not net new.

also now that I think about it more, why are we even debouncing a function that is called `onBlur`? 🤔 

it's not as though you can successively trigger a blur event in any reasonable usage of the input. makes me wonder if we were mistakenly relying on the incorrect implementation to cause a sort of delay rather than a true debounce.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
```diff
@@ -67,6 +69,7 @@ export function useInternalChipDismissibleInput({
     handleReset: () => {
       setActiveIndex(activeIndex === 0 ? activeIndex : activeIndex - 1);
       setSearchValue("");
+      actions.handleCloseMenu();
```
ZakaryH, line 72: I don't think this is safe to move into here.

we call `handleReset` in `handleSelectOption`, so this will cause the menu to close when we make a selection.

ZakaryH: though in the scenario where we only want to have the menu open when there's a search term, `setSearchValue("")` will also close it haha - well that's if we use `!!searchValue` as a criteria anyway. we can make it more explicit if we'd like.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
```diff
@@ -81,14 +84,35 @@ export function useInternalChipDismissibleInput({
...
+
+      if (
+        submitInputOnFocusShift &&
+        searchValue.length > 0 &&
+        allOptions.length > 0
+      ) {
+        // If there's a custom option, select it. Otherwise select the best match
+        const optionToSelect = canAddCustomOption
+          ? allOptions[allOptions.length - 1]
+          : allOptions[0];
```
ZakaryH, line 96: what's the relevance of `allOptions.length - 1`?

is the last option when you can custom add the "add _____ as whatever"?

packages/components/src/Chips/ChipsTypes.tsx
ZakaryH: now that I understand the behavior attached to this prop pretty well, I'm fine with it as is

yes it's a bit long, but it's descriptive enough that I think it's justified.

packages/components/src/Chips/ChipsTypes.tsx
ZakaryH: this one I feel less great about
- submit feels too form-ey, whereas we're using it for a selection
- input is a very technical detail - yes the input is technically powering it but it's more about the search term being used as a selection or Chip
- onFocusShift to me feels a lot like `onBlur` assuming that's the event it attaches to


with that in mind what do you think about any of these:
- `createFromSearch`
- `autoSelectOnBlur`
- `useSearchAsChip`
- `convertSearchToSelection`


with some of the ones that don't include the "when" of "onBlur" maybe having that tacked onto the end. though actually that reminds me, will this also control the behavior of using the search term as a selection when you press `Enter`? if yes, then it would be when you click outside, tab, and type `,` right?

ZakaryH:
seeing some strange behavior in a few scenarios. it's possible that my webstory is set up incorrectly, I've got this:

<details>
  <summary>code</summary>

 ```
const SelectionTemplate: ComponentStory<typeof Chips> = args => {
  const [selected, setSelected] = useState<string[]>(["Mando"]);
  const [options, setOptions] = useState<string[]>([
    "Mando",
    "Luke",
    "Leia",
    "Han",
    "Chewie",
    "R2D2",
    "C3PO",
  ]);
  const [loading, setLoading] = useState(false);
  const [search, setSearch] = useState("");

  const handleCustomAdd = (value: string) => {
    setOptions([...options, value]);
  };

  const handleSearch = (value: string) => {
    options.filter(option =>
      option.toLowerCase().includes(value.toLowerCase()),
    );
  };

  return (
    <Chips
      {...args}
      type="dismissible"
      selected={selected}
      onChange={setSelected}
      onCustomAdd={handleCustomAdd}
      isLoadingMore={loading}
      onSearch={handleSearch}
      onLoadMore={() => console.log("load more")}
      onlyShowMenuOnSearch={true}
      submitInputOnFocusShift={true}
    >
      {options.map(name => (
        <Chip key={name} label={name} value={name} />
      ))}
    </Chips>
  );
};

```
</details>


https://github.com/user-attachments/assets/5318c42f-c617-43a3-8bf7-55d83e016b29

basically it's taking 2 `tab`s or 2 clicks outside for the search term to be used, though it's correctly adding items to the list it would seem?

then the behavior of typing a comma doesn't seem to work at all. I see in the `master` version of this event handler we're pretty much doing the same thing as the new behavior

```
        ",": () => {
          if (searchValue.length === 0) return;
          actions.handleSelectOption(generateCustomOptionObject(searchValue));
        },
```

so unless we specifically don't want that to happen in the new flow I feel like it should continue to work in the same way as the new stuff that uses the search value as a selection

File: packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx [hunk 13]
```diff
@@ -66,10 +68,12 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     isInView && onLoadMore && onLoadMore(searchValue);
   }, [isInView]);
 
-  if (!menuOpen) {
-    return React.cloneElement(activator, { onClick: handleOpenMenu });
+  if (!showInput) {
+    return React.cloneElement(activator, { onClick: handleShowInput });
   }
```
ZakaryH, line 73: nice, even though I don't love what this component is doing to begin with - keeping it to a single return is great.

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -80,17 +84,17 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
...
-        aria-expanded={hasAvailableOptions}
+        aria-expanded={shouldShowMenu}
         aria-activedescendant={generateDescendantId(activeIndex)}
         value={searchValue}
         onChange={handleSearchChange}
         onKeyDown={handleKeyDown}
-        onBlur={debounce(handleBlur, 200)}
-        onFocus={handleOpenMenu}
+        onBlur={() => setTimeout(handleBlur, 200)}
+        onFocus={onlyShowMenuOnSearch ? undefined : handleOpenMenu}
```
ZakaryH, line 93: I don't feel too strongly about this since it's a component we don't want long term, but if it were - I'd be tempted to make an `handleFocus` method rather than calling `handleOpenMenu` directly

it probably made sense at one point since that was the only thing we wanted to do on focus, but now that there's some branching logic I think it'd be nice to have that encapsulated in a function rather than inline. maybe there is _something_ we'd want to do on focus even for the only-show-on-search case.

as it stands this is fine though since we have no such need, just thinking out loud.

packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissibleInput.test.tsx
```diff
@@ -96,57 +145,77 @@ describe("Arrow keys", () => {
...
+
+    act(() => {
+      jest.runOnlyPendingTimers();
+    });
+
+    fireEvent.keyDown(input, { key: "Enter" });
+
+    expect(handleCustomOptionSelect).toHaveBeenCalledWith(newValue);
+    expect(handleOptionSelect).not.toHaveBeenCalled();
+  });
```
ZakaryH, line 218: I notice we're missing a test asserting the comma typing behavior, and it seems to be having an issue right now.

once we sort that out, lets add a test for it for regression purposes and also to provide some level of documentation aroudn expected behavior

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
```diff
@@ -107,23 +138,35 @@ export function useInternalChipDismissibleInput({
...
           actions.handleSelectOption(computed.activeOption);
-        },
-        Tab: () => actions.handleSelectOption(computed.activeOption),
-        ",": () => {
+        };
+        callbacks.Tab = () => actions.handleSelectOption(computed.activeOption);
+
+        callbacks[","] = () => {
           if (searchValue.length === 0) return;
           actions.handleSelectOption(generateCustomOptionObject(searchValue));
```
ZakaryH, line 161: as I mentioned in the video, this case is being handled differently from the rest

I wonder if it's worth having a single wrapping function that handles all cases where we want to create based on the search input, and in there we have the logic that looks at our new props + `generateCustomOptionObject` related conditions

Aiden-Brine:
I appreciate you flagging the behaviour with the video. I think that is a WebStory issue though as I am not seeing any issues with the tab or comma functionality on my local Atlantis WebStory that has my Cursor fix for the querying functionality. More importantly I generated a new pre-release and added it to [this branch](https://github.com/GetJobber/jobber-frontend/pull/2571) an things seem to be working as expected (attaching a video). Happy to hop on and pair if that is helpful


https://github.com/user-attachments/assets/2703b854-c4ef-4ed1-9990-0217f681ae29

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx
Aiden-Brine: So right off the bat I think I should replace `allOptions[allOptions.length - 1]` in the above code with `generateCustomOptionObject(searchValue)` which is the same result but is consistent with here and much more readable. 

I am having a hard time seeing how a single wrapping function would be beneficial so I might need a code example to do that. Is it this code:
```
      if (
        submitInputOnFocusShift &&
        searchValue.length > 0 &&
        allOptions.length > 0
      ) {
        // If there's a custom option, select it. Otherwise select the best match
        const optionToSelect = canAddCustomOption
          ? generateCustomOptionObject(searchValue)
          : allOptions[0];
        actions.handleSelectOption(optionToSelect);
      }
    ``` 
and the code here that you think could be handled by a single wrapping function? Tab, comma, and selecting on blur when `submitInputOnFocusShift` is true all have different desired behaviours so I think keeping them all separate makes sense. Again, maybe I am missing your suggestion and there is some nice cleanup here but I think I need a code snippet or a quick call to be able to see it

File: packages/components/src/Chips/InternalChipDismissible/hooks/__tests__/useInternalChipDismissibleInput.test.tsx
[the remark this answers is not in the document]
Comment (Aiden-Brine) [reply packages/components/src/Chips/InternalChipDismissible/hooks/__tests__/useInternalChipDismissibleInput.test.tsx:1]: Interesting. I have always tried to take the approach of fully testing out hooks directly and not relying on a component to test out the hooks functionality indirectly. Certainly a different perspective but I think sometimes testing out hooks through the components it is almost like testing Atlantis component functionality in JO and JF. You should trust that the hook is doing what it says it does and that that functionality is tested elsewhere. I think the component tests do test a lot of the hook functionality but I definitely took the approach of not trying to get too under the hood of the hook in the component tests. 

I asked ChatGPT when you should test the hook functionality directly and when you should test the hook through the component. The start of the response was: 
> Test the hook directly when:
> - The hook contains complex logic (e.g., state management, side effects, keyboard interactions).
So to me this is a prime example of a time when testing the hook directly makes sense. That being said I _think_ I did a fairly good job with more integration style tests that test the hook functionality through the component. I don't think I should remove any of these hook tests but if you think there is an area that is lacking in the component tests I would be more than happy to add those.

Also just worth noting I worked on a repo for ~7 years that had 100% test coverage. Your CI would fail if you didn't fully cover every new line. I know there are arguments against this and it isn't the philosophy Jobber takes but my default is to test the heck out of things, hooks included.

File: packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx (hunk 14 above)
Comment (Aiden-Brine) [reply packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx:93]: I see where you are coming from and happy to do this. I think this was a bit of a carry over from me messing around, refactoring and trying to keep the number of statements in this component low. I will add a `handleFocus` and put it in the hook 👍

File: packages/components/src/Chips/ChipsTypes.tsx (hunk 8 above)
Comment (Aiden-Brine) [reply packages/components/src/Chips/ChipsTypes.tsx:119]: This will only be changing the behaviour when you have a search term and click outside. Enter, tab, and `,` should all continue to work as they did before. FWIW I am drawn to `autoSelectOnBlur` from this list but I guess it isn't fully accurate either. Technically it is `autoSelectOnClickOutside`. Too wordy?

File: packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissibleInput.test.tsx (hunk 15 above)
Comment (Aiden-Brine) [reply packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissibleInput.test.tsx:218]: This behaviour is tested in the hook tests but as I said here I am happy to add any additional test cases to the component tests. I will add a comma behaviour test here

**Comment by Aiden-Brine** [comment]
@ZakaryH to summarize things:
#### I have added a new commit that:
- Adds a test for the comma behaviour to the component tests
- Adds a `handleFocus` method to the hook
- Improves the logic when selecting custom options on blur when `submitInputOnFocusShift` is true

#### Still outstanding:
- I wasn't able to recreate what you are seeing in the video. I am hoping that it is an issue with your WebStory fix and you can verify the behaviour on [this branch](https://github.com/GetJobber/jobber-frontend/pull/2571)
-  Locking down a new name for `submitInputOnFocusShift`
- It seems we differ on testing philosophies for hooks. I think the easiest way to reconcile this would be to have me add in tests for any gaps you see in the component testing
- I'm not sure where to take things on:
> a single wrapping function that handles all cases where we want to create based on the search input

File: packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx (hunk 16 above)
Comment (ZakaryH) [reply packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissibleInput.tsx:167]: what you're saying about using the `generateCustomOptionObject` is pretty much the idea, assuming that the `allOptions[allOptions.length - 1]` can be replaced

I was thinking if those were 2 different ways of adding an item, that we'd have one 

```
addItemFromSearchContent() {
 if (newFlow) {
  allOptions[allOptions.length - 1]
 } else {
generateCustomOptions(...)
}
```

but if we can use that one method for both then there's no need for a wrapping function

packages/components/src/Chips/InternalChipDismissible/__tests__/InternalChipDismissibleInput.test.tsx
ZakaryH: yeah if you wouldn't mind. we have these other tests for the various keyboard interactions tested, might as well have this case too.

packages/components/src/Chips/ChipsTypes.tsx
ZakaryH: right it doesn't modify the behavior of those other ones. those will happen regardless.

`autoSelectOnClickOutside` does describe it the best, it's just such a specific prop that I don't know if there's a short way to describe it in any meaningful kind of way.

I'm OK with `autoSelectOnClickOutside` if you are

ZakaryH:
@Aiden-Brine 

- I'll verify the comma behavior on that branch. could definitely just be a webstory thing.
- the name, yeah I don't think there's going to be anything we're really happy with - I'm fine to settle for something good enough
- testing: I'm fine to leave the hook tests as is. I really don't have a strong enough argument to remove them.
- the wrapping function comment can be disregarded if we're going to share the same function that the `Tab` press uses. I really just wanted to make sure we had a single shared entrypoint/way of doing what seems like the same operation

ZakaryH:
alright so I've played around with this and verified that the strange tab behavior isn't a real problem

there is one real problem that I've encountered though where after making a selection the arrow keys stop working. you have to have multiple options in the list to be able to encounter it. in this gif after I make a selection I'm pressing my arrow keys and nothing is happening.
Kapture 2025-04-17 at 12 29 56

also, it might be due to the delay we added for making the menu show up? but when it goes away I'm seeing it stick around in an odd way - you can see it at the end of the gif.

ZakaryH:
to solve the menu staying open after a selection with the new props, we could either bring back that check in `shouldShowMenu` to verify the prop is present and a search value is present.

or we could include a conditional call to `actions.handleCloseMenu` in `handleSelectOption` 

my preference would be to call `handleCloseMenu` so that that method continues to do what it says.

PR #2473, opened 2025-04-08
feat(components): Add unsafe props to Avatar web

edison-cy-yang:
## Motivations

Adding UNSAFE props to atom components to allow more customizations

### Added

- Added UNSAFE_className and UNSAFE_style props to Avatar, which target the container, `Initials`, and `Icon`
- Tests for UNSAFE props
- Notes in the Implementations tab with UNSAFE props use explanation
- Avatar props file update

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Add this story to make sure everything renders as expected
```
// Target the fallback icon
export const FallbackIconCustomStyles = BasicTemplate.bind({});
FallbackIconCustomStyles.args = {
  color: "var(--color-indigo)",
  initials: "",
  name: "The Jobbler",
  size: "large",
  UNSAFE_style: {
    fallbackIcon: {
      svg: {
        width: "59px",
        height: "59px",
      },
      path: {
        fill: "var(--color-blue)",
      },
    },
  },
};

// Target the container
export const ContainerCustomStyles = BasicTemplate.bind({});
ContainerCustomStyles.args = {
  color: "var(--color-indigo)",
  imageUrl:
    "https://images.unsplash.com/photo-1533858539156-90ea20bafd17?ixid=MXwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHw%3D&ixlib=rb-1.2.1&auto=format&fit=crop&w=1650&q=80",
  name: "The Jobbler",
  size: "large",
  UNSAFE_style: {
    container: {
      borderColor: "var(--color-green)",
    },
  },
};

// Target Initials
export const InitialsCustomStyles = BasicTemplate.bind({});
InitialsCustomStyles.args = {
  color: "var(--color-indigo)",
  initials: "JBLR",
  name: "The Jobbler",
  size: "large",
  UNSAFE_style: {
    initials: {
      color: "var(--color-blue)",
    },
  },
};
```

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Avatar/Avatar.tsx
+++ b/packages/components/src/Avatar/Avatar.tsx
@@ -3,7 +3,7 @@ import classnames from "classnames";
 import { XOR } from "ts-xor";
 import styles from "./Avatar.module.css";
 import { isDark } from "./utilities";
-import { Icon } from "../Icon";
+import { Icon, IconProps } from "../Icon";
 
 type AvatarSize = "base" | "large" | "small";
 interface AvatarFoundationProps {
@@ -42,10 +42,7 @@ interface AvatarFoundationProps {
   readonly UNSAFE_className?: {
     container?: string;
     initials?: string;
-    fallbackIcon?: {
-      svg?: string;
-      path?: string;
-    };
+    fallbackIcon?: IconProps["UNSAFE_className"];
   };
 
   /**
@@ -56,10 +53,7 @@ interface AvatarFoundationProps {
   readonly UNSAFE_style?: {
     container?: CSSProperties;
     initials?: CSSProperties;
-    fallbackIcon?: {
-      svg?: CSSProperties;
-      path?: CSSProperties;
-    };
+    fallbackIcon?: IconProps["UNSAFE_style"];
   };
 }
 
@@ -111,7 +105,19 @@ export function Avatar({
       aria-label={name}
     >
       {!imageUrl && (
-        <Initials initials={initials} dark={shouldBeDark} iconSize={size} />
+        <Initials
+          initials={initials}
+          dark={shouldBeDark}
+          iconSize={size}
+          UNSAFE_className={{
+            initials: UNSAFE_className.initials,
+            fallbackIcon: UNSAFE_className.fallbackIcon,
+          }}
+          UNSAFE_style={{
+            initials: UNSAFE_style.initials,
+            fallbackIcon: UNSAFE_style.fallbackIcon,
+          }}
+        />
       )}
     </div>
   );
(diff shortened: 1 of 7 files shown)


packages/components/src/Avatar/Avatar.tsx
```diff
@@ -31,6 +31,30 @@ interface AvatarFoundationProps {
...
   readonly size?: AvatarSize;
+
+  /**
+   * **Use at your own risk:** Custom class names for specific elements. This should only be used as a
+   * **last resort**. Using this may result in unexpected side effects.
+   * **Note:** If you are applying fill override to buttonIcon.path, you will need to add !important due
+   * to Button's children element css inheritance.
+   * More information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).
+   */
+  readonly UNSAFE_className?: {
```
edison-cy-yang, line 42: thought about strong typing where `AvatarWithImageProps` can only target `container` and `AvatarWithInitialProps` can target all named properties, but things start to get messy quick

packages/components/src/Avatar/Avatar.tsx
```diff
@@ -83,22 +127,44 @@ interface InitialsProps {
...
+
+  readonly UNSAFE_className?: {
+    initials?: string;
+    fallbackIcon?: IconProps["UNSAFE_className"];
+  };
+
+  readonly UNSAFE_style?: {
+    initials?: CSSProperties;
+    fallbackIcon?: IconProps["UNSAFE_style"];
+  };
```
nad182, line 139: Would it make sense to create shared ("Base") interfaces for `UNSAFE_className` and `UNSAFE_style` to be used here and on lines [42-57](https://github.com/GetJobber/atlantis/pull/2473/files#diff-839c127bd9443ba9b8e788cc9e89af8904d4b9f992f9906cd9ff9e866c770db8R42-R57)?

The reasoning is that if we ever need to change an interface, we could do it just in the Base instance, and the other interfaces would inherit it. This is not a hard requirement though, just a suggestion.

nad182 (approved):
Looks good! Just left one small suggestion regarding creating base/shared interfaces for unsafe styles and classnames to share between Avatar and Initials, but it's definitely non-blocking, so I'm approving the PR as is.

packages/components/src/Avatar/Avatar.tsx
edison-cy-yang: @nad182 Good callout! It's definitely more robust to change. Created shared className and style interfaces.

nad182: Looks like you might need to regenerate the docs.

PR #2477, opened 2025-04-09
feat(components): InputTime without RHF

ZakaryH:
## Motivations

Currently InputTime uses FomField for the input logic and styling. FormField uses RHF internally, and this can cause issues with external instances of RHF trying to use the component.

As such, we're going to make an FormField (RHF) decoupled version of the component under the `version=2` group that can be opted into.

### Added

- a separate InputTime.rebuilt component that does not use FormField
- internal test-ids to the Clearable button and the InputTime's input because `getByRole("textbox")` is failing to find the `input type="time"` for some reason 

### Changed

- time utils extracted to their own file to be shared across the two implementations
- index.ts -> index.tsx with logic to determine which component to use based off `version`
- types file renamed to match how we name these types of files
- InputTime.rebuilt doesn't even attempt to call `props.onKeyUp` because it's not a real prop we'd ever receive

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

InputTime v1 should be totally unchanged

inputTime v2 should work as expected for base functionality 
- onBlur/onFocus
- time prediction
- clearing
- other props like `inline`
- etc.

additionally now you can pass in "name" without a FormProvider and it'll work nicely with no generated name

I don't know how to test the `useEffect` block, I still don't quite understand what it should be doing.

Changes can be

---

Random photo of Atlantis


packages/components/src/InputTime/InputTime.tsx
```diff
@@ -38,50 +45,16 @@ export function InputTime({
...
   function handleBlur() {
     params.onBlur?.();
+    console.log("handleBlur");
 
     // Time inputs doesn't clear the typed value when it's invalid. This should
     // force it to reset the input value when the typed value is invalid.
     if (ref.current) {
+      // when would this ever be invalid? we don't allow any
+      // of the attributes that would make it invalid
       if (!ref.current.checkValidity()) {
```
ZakaryH, line 55: I'm unclear how we'd ever hit this block. I must be missing something because it seems to me that we'd always be comparing a timestamp to some string value.

image
image
https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation

doesn't really matter in the context of what this PR is meant to achieve though.

packages/components/src/InputTime/InputTime.tsx
```diff
@@ -38,50 +45,16 @@ export function InputTime({
 
   function handleBlur() {
     params.onBlur?.();
+    console.log("handleBlur");
 
     // Time inputs doesn't clear the typed value when it's invalid. This should
     // force it to reset the input value when the typed value is invalid.
     if (ref.current) {
```
ZakaryH, line 52: another **existing issue** with this component is that it accepts an `inputRef` prop that gets applied to `FormField` in a non obvious way via `{...fieldProps}` where that spread is done after `inputRef={ref}` with a ref of our own that we're attempting to use

so, if you provide `inputRef` it will override the other `ref` leaving it `null` forever, and preventing this code block from ever executing

as I mention below, I'm not too clear on the implications of that since I don't know when it's supposed to trigger even when we have the ref.

packages/components/src/InputTime/InputTime.tsx
```diff
@@ -38,50 +45,16 @@ export function InputTime({
...
-    }
-
-    const date = new Date();
-    date.setHours(hours, minutes, 0, 0);
-
-    return date;
-  } catch {
-    return undefined;
-  }
-}
```
ZakaryH, line 87: I'm going to try my best to stay on target but I'm very very tempted to reach for `date-fns` rather than doing our own date/string parsing. I've found that library to be really nice for date/time operations, and there's plenty of edge cases to consider when it comes to dates and time.

packages/components/src/InputTime/hooks/useTimePredict.ts
```diff
@@ -1,6 +1,6 @@
 import debounce from "lodash/debounce";
 import { useCallback, useEffect, useState } from "react";
-import { InputTimeProps } from "../InputTimeProps";
```
ZakaryH, line 3: minor rename to better match how we name these types of files in the project

packages/components/src/InputTime/InputTime.rebuilt.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const { inputStyle } = useFormFieldWrapperStyles(params);
+
+  const id = getId(params);
+  const fieldProps = {
+    onChange: handleChange,
+    ...(defaultValue && {
+      defaultValue: dateToTimeString(defaultValue),
+    }),
+    ...(!defaultValue && { value: dateToTimeString(value) }),
+  };
```
ZakaryH, line 32: this object doesn't make much sense anymore. we need to manage the default value/value logic in a different way since previously this was something we'd make use of inside FormField's RHF logic

ZakaryH:
pretty much all the props of InputTime make sense. there's only a couple that are questionable:

- `maxLength` is weird in the context of time, and if you want the shorter input behavior that we've paired with this prop that should be a CSS layout responsibility
- `autocomplete` also makes little sense

packages/components/src/InputTime/InputTime.tsx
```diff
@@ -9,15 +11,20 @@ export function InputTime({
   onChange,
   ...params
 }: InputTimeProps) {
+  // this will only work if inputRef is not provided
+  // otherwise due to spreading fieldProps after inputRef, the inputRef will be overwritten
   const ref = useRef<HTMLInputElement>(null);
   const { setTypedTime } = useTimePredict({ value, handleChange });
 
-  const fieldProps: FormFieldProps = {
```
ZakaryH, line 15: the FormField complexity is really such a nuisance.

because we've said that `fieldProps` are of type `FormFieldProps` it believes that `onKeyUp` is a possible prop, which it is - for FormField. however if you look at the type for `InputTimeProps` you'll see that we do not accept an `onKeyUp` prop, meaning that the `params` we are spreading here should never contain `onKeyUp`

but because of the `FormFieldProps` type, line 36 thinks that's valid, that we have a `fieldProps.onKeyUp?.(e)` to call. it's safe because of the safe navigator, so it won't case any harm but this will never do anything until InputTime accepts an `onKeyUp`.

I'm not going to fix this for now here, and in the new implementation I'm just not going to even attempt to call it because well it doesn't exist, and that's how I went down this rabbit hole on my version which is detaching from FormField

docs/components/InputTime/Web.stories.tsx
ZakaryH, line 1: reminder to remove all this when it's ready

packages/components/src/InputTime/InputTime.rebuilt.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export function InputTimeRebuilt({
+  defaultValue,
+  value,
+  onChange,
+  ...params
+}: InputTimeRebuiltProps) {
+  const ref =
+    (params.inputRef as React.RefObject<HTMLInputElement>) ??
+    useRef<HTMLInputElement>(null);
```
ZakaryH, line 15: either use the ref we were provided via `inputRef` or use our own instance to avoid ever having 2 separate instances where one would never be set

packages/components/src/InputTime/InputTime.rebuilt.tsx
```diff
@@ -0,0 +1,94 @@
...
+    </FormFieldWrapper>
+  );
+
+  function handleChangeEvent(event: ChangeEvent<HTMLInputElement>) {
+    handleChange(event.target.value);
+  }
+
+  function handleChange(newValue: string) {
+    onChange?.(timeStringToDate(newValue));
+  }
```
ZakaryH, line 64: this is a bit silly, but the reason is that `useTimePredict` wants a handler that takes in a string as the arg, so I'd like to have a function like that.

then, for the `input` the `onChange` will provide a ChangeEvent, so we need a function for that too.

it's also an option to modify `useTimePredict` to accept the event, though I'm trying to minimize changes during this work and the other version is all about strings.

packages/components/src/InputTime/InputTime.rebuilt.tsx
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  function handleClear() {
+    handleBlur();
+    onChange?.(undefined);
+    ref.current?.focus();
+  }
+
+  function handleFocus() {
```
ZakaryH, line 85: need to provide the events to these focus/blur callbacks

packages/components/src/InputTime/InputTime.rebuilt.tsx
```diff
@@ -0,0 +1,100 @@
...
+    params.onBlur?.(event);
+
+    if (ref.current) {
+      if (!ref.current.checkValidity()) {
+        // don't we need to update the value too?
+        // is it valid for them to be out of sync?
+        // onChange?.(undefined);
+        ref.current.value = "";
+      }
+    }
```
ZakaryH, line 82: need to delete this comment. for me the questions still stand though.

A) how would we ever trigger `checkValidity()`
B) is it correct to only adjust the `ref.current.value` and not inform `onChange` of the same new value

packages/components/src/InputTime/InputTime.rebuilt.tsx
```diff
@@ -0,0 +1,100 @@
...
+  };
+
+  return (
+    <FormFieldWrapper
+      disabled={params.disabled}
+      size={params.size}
+      align={params.align}
+      inline={params.inline}
+      name={params.name}
+      error={""}
```
ZakaryH, line 37: this is also a bit silly. we don't have an `error` prop on `InputTime` so there's nothing reasonable to pass it.

I suppose we could add an `error` prop, so that you can leverage `FormFieldWrapper`'s validation message styling and markup

packages/components/src/InputTime/InputTime.rebuilt.tsx
```diff
@@ -0,0 +1,100 @@
...
+      descriptionIdentifier={`descriptionUUID--${id}`}
+      invalid={Boolean(params.invalid)}
+      description={params.description}
+      clearable={params.clearable ?? "never"}
+      onClear={handleClear}
+      placeholder={params.placeholder}
+      value={value}
+    >
+      <input
+        ref={ref}
```
MichaelParadis, line 48: The rebuilt version can no longer be disabled

packages/components/src/InputTime/InputTime.rebuilt.tsx
MichaelParadis, line 37: What are you thoughts on being able to pass an error prop to the InputDateRebuilt so that people don't need to use `InputValidation` on their own for displaying an error?

packages/components/src/InputTime/InputTime.rebuilt.tsx
```diff
@@ -0,0 +1,100 @@
...
+      size={params.size}
+      align={params.align}
+      inline={params.inline}
+      name={params.name}
+      error={""}
+      identifier={id}
+      descriptionIdentifier={`descriptionUUID--${id}`}
+      invalid={Boolean(params.invalid)}
+      description={params.description}
+      clearable={params.clearable ?? "never"}
```
MichaelParadis, line 42: I noticed that clearable doesn't work when only a `defaultValue` is provided and the component is uncontrolled. This is likely due to [this](https://github.com/GetJobber/atlantis/blob/JOB-120387/inputtime-but-without-rhf/packages/components/src/FormField/FormFieldWrapper.tsx#L71-L78) line. I found that if `readonly value: Date | undefined;`  (note that this isn't `readonly value?: Date` is added to  `InputTimeRebuiltProps` the value needs to be passed but it can be undefined. This way the `InputTimeRebuilt` would only be usable as a controlled input.

packages/components/src/InputTime/InputTime.rebuilt.tsx
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  function handleBlur(event?: React.FocusEvent<HTMLInputElement>) {
+    params.onBlur?.(event);
+
+    if (ref.current) {
+      if (!ref.current.checkValidity()) {
+        // don't we need to update the value too?
+        // is it valid for them to be out of sync?
+        // onChange?.(undefined);
```
MichaelParadis, line 79: I would argue the value in the html input and the value the consumers are aware of should be the same especially since this behaviour was likely handled because RHF was integrated before

packages/components/src/InputTime/hooks/useTimePredict.ts
MichaelParadis: Minor comment, most of our types files are ___.**t**ypes.ts not ___.**T**ypes.ts

packages/components/src/InputTime/InputTime.rebuilt.tsx
MichaelParadis: Thought I replied to this comment or maybe my browser wasn't refreshed with the comments you made but I found this: https://github.com/GetJobber/atlantis/pull/2477/files#r2044745324

MichaelParadis (changes requested):
Blocking:
- `disabled` prop doesn't work on the V2
- `readonly` also doesn't work in V2
- Both are caused by these values not being passed to the underlying input

Nice to do:
- Rename types file to use lowercase **t** to match other types files

Unsure of:
- I think fixing the clearable not displaying when uncontrolled ([see](https://github.com/GetJobber/atlantis/pull/2477/files#r2044600675))by making the component need to be controlled makes sense but I am open to discuss it

packages/components/src/InputTime/InputTime.rebuilt.tsx
ZakaryH: ah yes, the disabled is only being passed to the `FormFieldWrapper` but we need it on the input itself. good catch.

packages/components/src/InputTime/InputTime.rebuilt.tsx
ZakaryH: interesting. I am going to try to remove `defaultValue` as a possible prop, so if that's the only scenario where clearable is misbehaving it _should_ go away soon.

packages/components/src/InputTime/InputTime.rebuilt.tsx
ZakaryH: >After testing it looks like the value of the original implementation wasn't provided to the consumer when it is cleared


hold on, how did you test this???

I still don't see how you'd ever hit this block of code. as far as I can tell it is inaccessible.

ZakaryH: right but that's the thing, how is this comparison ever going to work?

it seems to me we're always going to be comparing apples and oranges
https://github.com/GetJobber/atlantis/pull/2477#discussion_r2036272865

we don't necessarily **have to** solve this, since it's existing code - but also I'd rather not port over what is basically not functional code if we know it's not going to do anything.

MichaelParadis: ohh, yeah. I meant to say "we can leave this as is"

ZakaryH: sounds good. I'll make a ticket with a couple of the bugs and inaccessible code paths I've found in the existing implementation but not fuss with that for now.

ZakaryH:
@MichaelParadis for `readonly` it's worth noting that behavior doesn't work properly even on master

with a controlled component + `readonly` we don't have anything preventing `keyUp` behavior, so you can still get all the time predict functionality

I'm down to put in an early return in the rebuilt to prevent that. was there anything else you were seeing with it that doesn't do what you'd expect?

ZakaryH:
`disabled` is now being passed to the input
`readonly` is now being passed to the input

added a check in the `onKeyUp` to short circuit if `disabled` or `readonly` is true, preventing the time prediction functionality which effectively writes to the input. interestingly Firefox treats this correctly, Safari and Chrome do not - oh and it only happens on a controlled version of InputTime.

renamed the types file to match our casing

_trying_ to remove the `defaultValue` but it's not been successful so far. that in theory would resolve the clearable issue, though now that you mention it, I wonder if it should also factor in disabled + readonly behavior. clearing it is modifying it.

packages/components/src/InputTime/InputTime.tsx
```diff
@@ -12,12 +14,15 @@ export function InputTime({
...
-    onChange: handleChange,
-    ...(defaultValue && { defaultValue: dateToTimeString(defaultValue) }),
-    ...(!defaultValue && { value: dateToTimeString(value) }),
-    ...params,
-  };
+  const fieldProps: FormFieldProps = omit(
+    {
+      onChange: handleChange,
```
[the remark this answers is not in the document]
ZakaryH: hmm we do have that as an optional argument on `FormField`'s `onChange`

```
  onChange?(
    newValue: string | number | boolean | Date,
    event?: ChangeEvent<
      HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement
    >,
  ): void;

```

mine is explicitly overriding it, so I'm kinda leaning towards leaving it alone

```
  /**
   * Function called when user changes input value.
   */
  onChange?(newValue?: Date): void;
```

for Select I think it's inheriting/extending the FormField version which does have the optional `event` arg, so in that case I think it's fine to provide since it's optional anyway.

ZakaryH:
summary of the last few commits
- added an `error` prop to the rebuilt version (maybe I should have do it for both?)
- passed down some more props to FormFieldWrapper that were missing and causing the miniLabel behavior to misbehave
- adjusted the `useShowClear` to take into account `readonly` which IMO should also prevent it from showing up because that's definitely not readonly if you can clear it - **this will apply to much more than just InputTime**


the Safari behavior of setting the time to 12:30 is something the browser is doing itself and we'd have to find a way to override. it's doing it even when the input is disabled or readonly though important to note it's already doing that today on master.

the input still predicting time when it's set to `readonly` and a user types in the input is broken on Chrome and Safari but works correctly on Firefox. again this is an existing issue that I'd like to solve separately.

packages/components/src/InputTime/InputTime.tsx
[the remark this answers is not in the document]
MichaelParadis: I think leaving it as is make sense especially since the V1 version also didn't expose the event. We can always update it if people need it

MichaelParadis (approved):
LGTM!

Great job on this and thanks for fixing some of the weird browser behaviour

PR #2484, opened 2025-04-14
feat(components): Select without RHF

edison-cy-yang:
## Motivations

Current `Select` uses FormField for the input styling and logic, and FormField uses RHF. We want to decouple the use of RHF so RHF instances outside of the component do not cause issues.

The decoupled version of Select has `version=2`

### Added

- `Select.rebuilt` component without dependency on RHF

### Changed

- `index.ts` -> `index.tsx` to conditionally render different Select versions with `version` prop
- props for v1 Select use `Pick` instead of Pick and Exclude (Omit)

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Select v1 should be unchanged

Select v2 should continue to work how you'd expect it:
- onBlur/onFocus/onChange
- props like `inline`, `disabled`, `size` etc

This following change is needed in JO with the version bump. Please test against [this branch](https://github.com/GetJobber/Jobber/tree/JOB-120388-do-it-select)

This change is needed in JFE with the version bump: https://github.com/GetJobber/jobber-frontend/tree/JOB-120388-do-it-select

Changes can be

---

Random photo of Atlantis


packages/components/src/Select/Select.types.ts
```diff
@@ -0,0 +1,43 @@
...
+  readonly value?: string;
+  onChange?(newValue?: string): void;
+  version?: 1 | 2;
+}
+
+/**
+ * Rebuilt version of the Select component without React Hook Form dependency.
+ */
+export interface SelectRebuiltProps extends Omit<SelectProps, "defaultValue"> {
+  version: 2;
```
edison-cy-yang, line 42: I think Select V2 should be controlled so it doesn't need `defaultValue`. In that case, should `value` and `onChange` be required here?

packages/components/src/Select/Select.types.ts
```diff
@@ -0,0 +1,43 @@
...
+      | "readonly"
+      | "autofocus"
+      | "onEnter"
+      | "onBlur"
+      | "onFocus"
+      | "inputRef"
+      | "wrapperRef"
+      | "validations"
+      | "children"
+      | "prefix"
```
edison-cy-yang, line 29: Not actually sure what prefix and suffix would be useful for. They were already included in the original Select types

packages/components/src/Select/Select.rebuilt.tsx
```diff
@@ -0,0 +1,92 @@
...
+import {
+  FormFieldWrapper,
+  useAtlantisFormFieldName,
+  useFormFieldWrapperStyles,
+} from "../FormField";
+import { FormFieldPostFix } from "../FormField/FormFieldPostFix";
+
+export function SelectRebuilt(props: SelectRebuiltProps) {
+  const selectRef =
+    (props.inputRef as React.RefObject<HTMLSelectElement>) ??
```
edison-cy-yang, line 15: FormField already has `inputRef` prop so we should use that instead of forwardRef here

packages/components/src/Select/Select.types.ts
```diff
@@ -0,0 +1,43 @@
...
+      | "suffix"
+    > {
+  readonly defaultValue?: string;
+  readonly value?: string;
```
edison-cy-yang, line 33: It seems like there are usages of value being a number (Pagination.tsx). However value in `Option` is just a string. Should it be both?

edison-cy-yang: That's what CI is failing for. I can get a quick fix in once confirmed if this is desired behaviour

packages/components/src/Select/Select.types.ts
MichaelParadis: I think keeping them optional isn't the worst especially since they may want to have a "undefined" value set. Just a heads up you may need to do something like [this](https://github.com/GetJobber/atlantis/pull/2477/files#diff-a25d7efa4fc4f45e4aa11bd967a57cd799030ea361908c02d08765a15c109501R52-R56) to truly remove default value from the version 2

packages/components/src/Select/Select.types.ts
MichaelParadis: The prefix would be useful for something like this in the V1
image

packages/components/src/Select/hooks/useSelectActions.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  function handleBlur() {
+    onBlur?.();
+  }
+
+  function handleFocus() {
+    onFocus?.();
+  }
+
+  function handleClear() {
```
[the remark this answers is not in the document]
MichaelParadis: I think `FormFieldWrapperProps` requiring `onClear` was an oversight. I think keeping Select not clearable makes sense since it was what the V1 did

packages/components/src/Select/Select.rebuilt.tsx
MichaelParadis: I think them being a forward ref makes sense, however I think for now this is sufficient since we did it in `InputTime` and we have been having discussions about exposing refs in general

packages/components/src/Select/Select.rebuilt.test.tsx
edison-cy-yang, line 1: perhaps I shouldnt test against snapshot here?

packages/components/src/Select/hooks/useSelectActions.ts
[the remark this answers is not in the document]
edison-cy-yang: Cool, I changed the handleClear function to an empty function to pass to FormFieldWrapper.

packages/components/src/Select/Select.rebuilt.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+        </SelectRebuilt>,
+      );
+
+      const select = container.querySelector("select") as HTMLSelectElement;
+      fireEvent.blur(select);
+
+      expect(blurHandler).toHaveBeenCalledTimes(1);
+    });
+
+    it("should handle all events correctly when they are combined", () => {
```
MichaelParadis, line 181: Same event and `getByRole` as above

packages/components/src/Select/Select.rebuilt.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+      // Call native methods through ref
+      if (ref.current) {
+        ref.current.focus();
+        ref.current.blur();
+      }
+
+      // Verify
+      expect(selectElement?.focus).toHaveBeenCalled();
+      expect(selectElement?.blur).toHaveBeenCalled();
+    });
```
MichaelParadis, line 256: You can use typescript expects to not have to do the conditional checks and other changes as well.

```suggestion
    it("can access native select methods through inputRef", () => {
      const ref = React.createRef<HTMLSelectElement>();
      const mockFocus = jest.fn();
      const mockBlur = jest.fn();
      render(
        <SelectRebuilt version={2} inputRef={ref}>
          <Option>Foo</Option>
        </SelectRebuilt>,
      );

      const selectElement = screen.getByRole("combobox");
      expect(selectElement).toBeInstanceOf(HTMLSelectElement);
      // Setup mocks
      selectElement.focus = mockFocus;
      selectElement.blur = mockBlur;

      // Call native methods through ref
      ref.current?.focus();
      ref.current?.blur();

      // Verify
      expect(mockFocus).toHaveBeenCalled();
      expect(mockBlur).toHaveBeenCalled();
    });
```

packages/components/src/Select/hooks/useSelectActions.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  function handleFocus() {
+    onFocus?.();
+  }
+
+  // Noop for SelectRebuilt since onClear is a required prop for FormFieldWrapper
+  // but Select is not clearable.
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  function handleClear() {}
```
MichaelParadis, line 36: What are your thoughts on updating FormFieldWrapper to no longer require `onClear` since it was likely an oversight making it required?

MichaelParadis (changes requested):
Some changes to tests, I may not have provided suggestions for all of them but the gist is to use `screen.getBy...` instead of `container.querySelector`, use the jest matches such has `toHaveValue` and to use `userEvent` instead of `fireEvent`

Minor comment on the `handleClear` stuff

packages/components/src/Select/Select.rebuilt.test.tsx
```diff
@@ -0,0 +1,258 @@
+import React from "react";
```
MichaelParadis, line 1: It might be worth having cursor write some stories for the version comparisons on props to make seeing the differences easier. An example of this is done in [InputText](https://github.com/GetJobber/atlantis/blob/master/docs/components/InputText/Web.stories.tsx#L104)

packages/components/src/Select/hooks/useSelectActions.ts
edison-cy-yang: I think that makes sense if you're on board. It should be a straight forward change. I can get that started

edison-cy-yang: actually, let's do it as a separate PR and I'll come back to take out the empty function.

packages/components/src/Select/Select.types.ts
```diff
@@ -0,0 +1,43 @@
+import { CommonFormFieldProps, FormFieldProps } from "../FormField";
+
+export interface SelectProps
+  extends Pick<
+      CommonFormFieldProps,
+      | "id"
+      | "align"
+      | "description"
+      | "disabled"
+      | "invalid"
```
edison-cy-yang, line 10: I took out `loading` since it was not being used in v1 as well

edison-cy-yang:
> Some changes to tests, I may not have provided suggestions for all of them but the gist is to use `screen.getBy...` instead of `container.querySelector`, use the jest matches such has `toHaveValue` and to use `userEvent` instead of `fireEvent`
> 
> Minor comment on the `handleClear` stuff

- `screen.getBy...` instead of `container.querySelector` ✅
- `useEvent` instead of `fireEvent` ✅
- Would like to make `onClear` optional on `FormFieldWrapper` in a separate PR. This one is a big change already.

packages/components/src/Select/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { Select as SelectLegacy } from "./Select";
+import { SelectRebuilt } from "./Select.rebuilt";
+import { SelectProps, SelectRebuiltProps } from "./Select.types";
+
+export { Option } from "./Option";
+export type SelectShimProps = SelectProps | SelectRebuiltProps;
```
MichaelParadis, line 7: I think we should expose `SelectProps` as `SelectLegacyProps` and `SelectRebuiltProps` to help with people who want to write components that use the `SelectProps` in their own components. We did the same with `InputText` because there were some issues with typescript when people used `React.ComponentProps<ComponentThatWasShimmed>`

packages/components/src/Select/Select.rebuilt.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+    );
+
+    const selectElement = screen.getByRole("combobox");
+
+    // Verify the value is correctly set
+    expect(selectElement).toHaveValue(expectedValue);
+
+    // Verify the correct option is selected
+    const select = selectElement as HTMLSelectElement;
+    expect(select.options[select.selectedIndex].text).toBe("Baz");
```
MichaelParadis, line 100: You could do the following but it isn't blocking

```suggestion
     expect(select.options[select.selectedIndex]).toHaveTextContent("Baz");
```

packages/components/src/Select/index.tsx
MichaelParadis: We did this here
https://github.com/GetJobber/atlantis/blob/JOB-120388-do-it-select/packages/components/src/InputText/index.tsx#L36

packages/components/src/Select/Select.types.ts
```diff
@@ -0,0 +1,43 @@
...
+      | "onBlur"
+      | "onFocus"
+      | "inputRef"
+      | "wrapperRef"
+      | "validations"
+      | "children"
+      | "prefix"
+      | "suffix"
+    > {
+  readonly defaultValue?: string | number;
```
MichaelParadis, line 30: I noticed that this breaks V1 usages because the initial implementation to `Select` also used `FormFieldProps` which could be a `Date` as well

MichaelParadis: This also applies to value and `onChange` (I would override these in the V2 version) and keep the original ones that were from `FormFieldProps`

MichaelParadis (changes requested):
The main thing is some of the exports and the typescript stuff breaks existing V1 usages in various ways. I can chat and help as needed but you can see the breaking stuff by installing the pre-release

edison-cy-yang:
just noting that I think there are also some changes needed for JFE

MichaelParadis (approved):
LGTM!

MichaelParadis (changes requested):
I just noticed that `validations` and `onValidation` are exposed on the V2 variant. I think they also need to be omitted

MichaelParadis (approved):
LGTM!

PR #2485, opened 2025-04-15
feat(components): Feedback on Layout 2.0

scotttjob:
## Motivations

1. Got some feedback on the Layout components.
2. Wanted the review of the feedback to be easy (this is a big PR), so I broke the feedback out into its own PR so it can be reviewed independently.
3. If we approve this PR, it will be an implied approval on the original Layout PR as well (we can rubber stamp that through, just to make things less complicated).

## Changes

1. Updated 'space' token for 'gap' everywhere in the new layout components. There was a few references, and they've all been updated.
1. Removed `useMediaQuery` usages for 'collapseBelow'. Instead relying on common Atlantis breakpoints. Also included a 'collapsed' prop so the consumer can have completely control beyond the standard breakpoints.
1. Added the ability to set new layout components to specified list of semantic HTML elements (additional elements can be added in the future if needed).
1. Added the ability for data and aria pass through, open to alternative patterns for this approach.
1. Added role/id pass through for all components, along with UNSAFE styles and classnames.
1. Updated naming of some functions away from 'use' based react hooks to 'get' based standard functions.
1. Moved all Inline styles configuration out of disconnected (and prematurely optimized) useMemo instances.
1. Updated some descriptions to be clearer

Changes can be

---

Random photo of Atlantis


packages/components/src/Cluster/types.ts
```diff
@@ -17,15 +26,40 @@ export interface ClusterProps {
...
+   * The HTML tag to render the cluster as.
+   */
+  as?: CommonAllowedElements;
+
+  /** **Use at your own risk:** Custom class names for specific elements. This should only be used as a
+   * **last resort**. Using this may result in unexpected side effects.
+   * More information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).
+   */
+  readonly UNSAFE_className?: {
+    container?: string;
```
ericchernuka, line 56: I had a question on this? I realize its consistent, but requires an object to be created for elements with only 1 component like this? Same comment about styles.

packages/components/src/Cover/Cover.tsx
```diff
@@ -1,27 +1,67 @@
...
+  readonly UNSAFE_className?: {
+    centerContent?: string;
+  };
+  readonly UNSAFE_style?: {
+    centerContent?: React.CSSProperties;
+  };
 }) {
-  return <div className={styles.centerContent}>{children}</div>;
+  return (
+    <div
```
ericchernuka, line 57: Should this also have a Tag?

packages/components/src/sharedHelpers/getCommonProps.ts
```diff
@@ -0,0 +1,11 @@
+export function dataPropsMapped(data?: Record<string, string>) {
+  return Object.entries(data || {}).map(([key, value]) => ({
+    [key]: value,
+  }));
+}
```
ericchernuka, line 5: Is `data` supposed to be used like
```
data={{ 'data-foo': true }}
```
or
```
data={{ foo: true }}
```

packages/components/src/sharedHelpers/getCommonProps.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function ariaPropsMapped(aria?: React.AriaAttributes) {
+  return Object.entries(aria || {}).map(([key, value]) => ({
+    [key]: value,
+  }));
+}
```
ericchernuka, line 11: Same comment about aria attributes.

ericchernuka: Wonder if we can just extend `AriaAttributes` from React?

packages/components/src/sharedHelpers/getMappedBreakpointWidth.ts
```diff
@@ -0,0 +1,23 @@
+export const AtlantisBreakpoints = {
+  xs: 0,
+  sm: 490,
+  md: 768,
+  lg: 1080,
+  xl: 1440,
+};
```
ericchernuka, line 7: I noticed that we seem to redefine these across many components? Do we need a central version that they all extend from?

ericchernuka (commented):
Overall liking the direction and flexibility. Few questions.

packages/components/src/Cover/Cover.tsx
scotttjob: Sure should! Nice catch.

packages/components/src/Cluster/types.ts
ericchernuka: I think it could become a clear rule that if it has more than one override, then it's an object; otherwise, it's a string.

packages/components/src/Cover/Cover.tsx
scotttjob: Addressed!

packages/components/src/sharedHelpers/getCommonProps.ts
scotttjob: I updated this while you were reviewing (my spidey sense was tingling)!  it's now a proper data- attribute and it lets you know how to use it.

packages/components/src/sharedHelpers/getCommonProps.ts
scotttjob: Donezo! This was one was in flight too while you were reviewing. You can now pass fully qualified `aria-` and it autocompletes for you.

packages/components/src/sharedHelpers/getMappedBreakpointWidth.ts
scotttjob: Yeah it's going to be this one. We seemed to have a pattern in the past where we didn't re-use anything across components unless they got promoted to hooks, but that's too much effort for some of these component-only usages.

packages/components/src/Cluster/types.ts
taylorvnoj: Myself, @jdeichert & @ZakaryH have discussed this before when adding `UNSAFE_` to atom components that are also just a container. Jacob's suggestion was that we used a keyed map to make sure we're handling future scenarios/future proofing (even if it seems unlikely any other sub element will be added to a component). 
I did initially have the same thoughts, and we _could_ have just a string for className and an object for `_styles` if people feel strongly. I tend to lean into consistency at this point.

packages/components/src/sharedHelpers/types.ts
```diff
@@ -0,0 +1,20 @@
...
+  /** Standard HTML data attributes. Accepts anything in a {{"data-key":"value"}} format. */
+  data?: { [key: `data-${string}`]: string };
+  /** Standard HTML aria attributes. Accepts all standard HTML aria attributes. */
+  aria?: React.AriaAttributes;
+  /** Standard HTML role attribute. */
+  role?: React.AriaRole;
+  /** Standard HTML id attribute. */
+  id?: string;
+}
+export type CommonAllowedElements =
```
ericchernuka, line 11: Let's also add `p` which would be useful in some scenerios

packages/components/src/Cluster/types.ts
ericchernuka: That's fair. I think it probably speaks to components needing more composition so that the overrides could be targetted, but another way to approach that with the current structure would be to actually give `data-slot="<componentName>"` which could then be targetted by the CSS through data attribute selectors rather than needing to be the ones responsible for plumbing classnames through.

Forgive the tailwind nuances here that they're hiding behind Components, but he presents an interesting approach to controlling styles from above using data attributes. https://www.youtube.com/watch?v=MrzrSFbxW7M

taylorvnoj:
Looks really good to me. We have been adding `UNSAFE_` documentation on Implement tabs but that and other Implement tab type info can definitely be in a follow up. I might even be able to do some of that work during my support hero rotation that starts today.
Not sure if you have any other changes for this PR - ping me for when ready/when circleci is green 👍🏻

packages/components/src/Cluster/types.ts
ericchernuka: Wonder if we just leverage what Braid does? https://github.com/seek-oss/braid-design-system/blob/0de668c186fd972a197160b799b980243ef58dc7/packages/braid-design-system/src/lib/components/private/buildDataAttributes.ts

Then you don't need to re-type data?

packages/components/src/sharedHelpers/getCommonProps.ts
ericchernuka: Yup. I dug into the types and was like why am I getting aria attributes!?

packages/components/src/Stack/types.ts
```diff
@@ -1,10 +1,16 @@
...
+  CommonAtlantisProps,
+} from "../sharedHelpers/types";
 
-export interface StackProps {
+export interface StackProps extends CommonAtlantisProps {
   readonly children: React.ReactNode;
   /** The amount of space between the children. Semantic tokens are available. */
-  readonly space?: string | Spaces;
-  /** The number of children to split the stack after (1-15). Requires parent to have height greater than the sum of the children. */
+  readonly gap?: string | Spaces;
```
nad182, line 10: Same comment as above regarding the type.

packages/components/src/Tiles/types.ts
```diff
@@ -1,11 +1,32 @@
...
+} from "../sharedHelpers/types";
 
-export interface TilesProps {
+export interface TilesProps extends CommonAtlantisProps {
   readonly children: React.ReactNode;
   /** The minimum size of the tiles. */
   readonly minSize: string;
   /** The amount of space between the tiles. Semantic tokens are available. */
-  readonly space: string | Spaces;
+  readonly gap: string | Spaces;
```
nad182, line 12: Same as above. Also is `gap` required for `Tiles`?

packages/components/src/SideKick/types.ts
```diff
@@ -1,17 +1,40 @@
...
-export interface SideKickProps {
+export interface SideKickProps extends CommonAtlantisProps {
   readonly children: React.ReactNode;
   /** The width of the sidekick. */
   readonly sideWidth?: string;
   /** The minimum width of the content. */
   readonly contentMinWidth?: string;
   /** The amount of space between the sidekick and the content. Semantic tokens are available. */
-  readonly space?: string | Spaces;
+  readonly gap?: string | Spaces;
```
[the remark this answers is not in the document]
scotttjob: Not intentionally no! I've replaced these various Spaces | options with a new 'GapSpacing' type that should be used everywhere this loose type appeared before.

packages/components/src/Stack/types.ts
scotttjob: Replaced with the above!

packages/components/src/Tiles/types.ts
scotttjob: Gap is not required! Good catch. I've updated this one as well with the new type

taylorvnoj (approved):

PR #2489, opened 2025-04-16
chore: New manual GitHub Action to diff bundled outputs

jdeichert:
## Motivations

This PR adds a new manually-triggered GitHub Action that _diffs the outputs of our packages_ on a given branch.

If you're testing a PR... let's say a dependabot version bump of some dependency... and you're not sure whether it's safe to ship because it could affect the final built assets that we ship, THIS ACTION IS FOR YOU!

Simply run this GH Action against your branch and you'll see one of two things:

✅ No differences found in built assets!

OR

⚠️ Differences found in built assets! This indicates that the compiled outputs in this branch don't match what master produces.
👉 Please investigate the differences above to see if they are expected.

If you see the latter warning, look above it and you'll see a git diff showing what exactly changed. Usually it will be something within a `dist` directory in one of the packages. Just review the diff to see if it makes sense and whether it's safe to ship OR if you need to do more testing, then do that!

### Added

- New GitHub Action to diff bundle/package outputs

## Testing

No need to test, I ran this on this PR a few times.

UPDATE 2026: The screenshots below are a bit outdated due to slight modifications to the echo logs, but the general concept is the exact same as below.

### When there's no changes

https://github.com/GetJobber/atlantis/actions/runs/14503388190

Screenshot 2025-04-16 at 3 12 30 PM

### When there are changes

https://github.com/GetJobber/atlantis/actions/runs/14503583543/job/40688621884

Screenshot 2025-04-16 at 3 15 59 PM

### DIFF
--- a/.github/workflows/diff-outputs.yml
+++ b/.github/workflows/diff-outputs.yml
@@ -1,6 +1,6 @@
 name: Diff Outputs
 
-on: pull_request
+on: workflow_dispatch
 
 jobs:
   diff-outputs:
--- a/packages/components/src/Checkbox/Checkbox.rebuilt.tsx
+++ b/packages/components/src/Checkbox/Checkbox.rebuilt.tsx
@@ -109,4 +109,4 @@ export const CheckboxRebuilt = forwardRef(function CheckboxRebuiltInternal(
   );
 });
 
-CheckboxRebuilt.displayName = "CheckboxRebuiltTestDiff";
+CheckboxRebuilt.displayName = "CheckboxRebuilt";


.github/workflows/diff-outputs.yml
```diff
@@ -0,0 +1,69 @@
...
+            echo "Removing existing node_modules..."
+            find . -type d -name "node_modules" -prune -exec rm -rf {} \; -exec echo "Deleted: {}" \;
+
+            echo "Installing deps and building..."
+            npm install
+          }
+
+          # Unignore dist directories and stage them to commit later
+          sed -i '/\<dist\>/d' .gitignore
+          git add -A
```
Aiden-Brine, line 41: I think we should force add them with `-f` instead of modifying the `.gitignore` file

.github/workflows/diff-outputs.yml
```diff
@@ -0,0 +1,69 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Configure git author
+        run: |
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "actions@github.com"
+
+      - name: Check for differences in built assets
+        run: |
```
Aiden-Brine, line 17: How about having `set -e` at the top of the script so that if something like `npm ci` fails it will exit out quickly?

.github/workflows/diff-outputs.yml
```diff
@@ -0,0 +1,69 @@
...
+          clean_dist() {
```
Aiden-Brine, line 18: I'm also wondering if it would be more readable if this was broken up into more name/run commands? You have lots of echo statements so it is still fairly easy to follow and I feel less strongly about this one.

.github/workflows/diff-outputs.yml
jdeichert: Good call, done!

Aiden-Brine (approved):
Thank you for making those changes

PR #2492, opened 2025-04-17
fix(docx): Fixing Local Dev Flow with the docs site.

scotttjob:
## Motivations

1. When working on the Layout components, I realized that the dev flow wasn't great when making changes to Atlantis.
2. I tried to fix it as part of the Layout components, but it required updating how we import types in a bunch of cases. I believe there still more of these mis imported types, but this is enough to get going with the dev site.

## Changes

1. Fixed the docs site so it pulls from `src` instead of `dist` for components, which means instant updates.
2. Had to adjust how we import types in a lot of cases.

## Testing

The new visual testing pages are a great working ground now. They don't have any wrapping CSS or structure.

For instance, open up the new layout testing pages /visual-testing/layout

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/site/.eslintrc.cjs
+++ b/packages/site/.eslintrc.cjs
@@ -8,7 +8,11 @@ module.exports = {
     "import/no-unresolved": [
       "error",
       {
-        ignore: ["^@atlantis/docs", "^@atlantis/packages"],
+        ignore: [
+          "^@atlantis/docs",
+          "^@atlantis/packages",
+          "^@jobber-component-styles",
+        ],
       },
     ],
   },
--- a/packages/site/vite.config.ts
+++ b/packages/site/vite.config.ts
@@ -63,7 +63,7 @@ export default defineConfig({
       "@jobber/formatters": path.resolve(__dirname, "../formatters/src"),
       "@jobber/hooks": path.resolve(__dirname, "../hooks/src"),
       "@jobber/components": path.resolve(__dirname, "../components/src"),
-      "@jobber/design-system": path.resolve(__dirname, "../design/src"),
+      "@jobber-component-styles": path.resolve(__dirname, "../components/dist"),
       "@storybook/addon-docs": path.resolve(
         __dirname,
         "./src/components/StorybookOverrides",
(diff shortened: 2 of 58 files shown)


packages/site/.eslintrc.cjs
```diff
@@ -8,7 +8,11 @@ module.exports = {
     "import/no-unresolved": [
       "error",
       {
-        ignore: ["^@atlantis/docs", "^@atlantis/packages"],
+        ignore: [
+          "^@atlantis/docs",
+          "^@atlantis/packages",
+          "^@jobber/component/styles",
```
scotttjob, line 14: By moving to load components out of /src/ meant that we no longer had access to the built css output from the dist folder.

This alias gives us access to the dist folder again, specifically for loading styles, which is why the key is prefixed with /component/styles.

packages/site/vite.config.ts
```diff
@@ -62,6 +62,8 @@ export default defineConfig({
     alias: {
       "@jobber/formatters": path.resolve(__dirname, "../formatters/src"),
       "@jobber/hooks": path.resolve(__dirname, "../hooks/src"),
+      "@jobber/components": path.resolve(__dirname, "../components/src"),
```
scotttjob, line 65: This line here is the "big" change. Every other change in this repo is just telling the system that we're importing types, instead of values. We had to make that change because vite expects you to do that, and now vite is responsible for building the components within the docs site (still the existing flow for the rest). It's rollup all the way down though.

We'll need to do this work when we move to vite entirely in the jobber components side either way, so good to get them out of the way now.

scotttjob (commented):
Left a few comments detailing what's happening in here.

packages/site/vite.config.ts
scotttjob: There's something up with running rollup in watch mode, it doesn't react to changes. Instead of spending more time down that rabbit hole, I spent 20 min on the other route to update to use the src + types.

If you can get rollup + watch going, that's definitely better (and also my first course of action)!

I think in general though, this update to properly define when we're importing/exporting a 'type' specifically is a good move, even if we don't update how we build the docs site (but could come in a different PR).

Maybe as a stepped approach (follow-up) to get these two pieces in sync, we can peek under hood to see what rules that vite applies, and apply them one-by-one to the components. Getting our component build process updated + modernized was always planned as a follow-on to the docs site work, so maybe we can chip away it incrementally over the course of the year, and then quietly roll over to vite when they're in sync.

scotttjob: Yup, I was aiming for the visual testing pages specifically. They're fantastic for fast iteration because they have no other wrapping page parts, just components.

PR #2493, opened 2025-04-17
fix(components): Provide Exact Button Event Type

ZakaryH:
## Motivations

With a recent change to Button, the types being returned from each different Button markup changed. realistically you are never going to get an anchorElement or a buttonElement event once you've decided which type of button you want (Link, url or standard)

there's no need for our types to suggest otherwise so cleaning that up

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis


ZakaryH:
related to some other work we decided not to move ahead with

PR #2496, opened 2025-04-17
fix(components): Fix Chip Type Issue

ZakaryH:
## Motivations

`Chip` types and props are not working. required props are not being added because TS doesn't say to add them, and non-existent props are being added with TS saying nothing about it. not great 😬 

we don't have this problem in Atlantis because here we have direct access to the type files regardless of what they are named. there's a step in the pipeline of preparing the packages for distribution that is not working, causing package consumers to not be able to use the types.
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

properly export `Chip` types so that when using `Chip` you get TypeScript help as far as what exists and what doesn't.

### Security

- 

## Testing

this should have no functional changes to the components, only to the dev experience where TS should correctly tell you what props exist and don't exist on Chip

use the prerelease, import it into a project, make sure you install and clear any caches - and then go to somewhere with a `Chip`. it should correctly tell you props, and if you give it props that don't exist it should tell you as much.

this will break CI since we have some invalid and missing props in product, so the PR that introduces this will have some work to do.

Changes can be

---

Random photo of Atlantis


packages/components/src/Chip/Chip.types.ts
ZakaryH, line 1: I'm 99% sure this is the culprit

in my Chip & Button PR I noticed we had the file named `Chip.types.d.ts` and I thought that was strange. typically `d.ts` files are generated, and that's not how any of our other type files are named so I changed it in my PR. I might have had a minor comment about it.

what I think was happening is that something either in vite or our own pipeline is trying to generate a `d.ts` file for this, but since it already exists it either doesn't generate it - or it's looking for something else and doesn't generate it.

either way, in a project using Chip if you ctrl click into the type definitions you'll hit a deadend when trying to see what `ChipProps` is
image

whereas other component like Combobox will link you to the `Combobox.types.d.ts`
image

basically Chip was missing all its types 😱

packages/components/src/Chip/Chip.tsx
```diff
@@ -9,7 +9,7 @@ import { useChildComponent } from "./hooks/useChildComponent";
 import { Typography } from "../Typography";
 import { Tooltip } from "../Tooltip";
 
-export const Chip = ({
+export function Chip({
```
ZakaryH, line 12: I thought this was a possible cause, it's not.

I'm fine to revert it, but also this is more our style for function definitions in general.

ZakaryH:
I've narrowed it down to 2 possible causes

- `Chip.types.d.ts` file name that we "manually" made
- **not** exporting any Chip types from `index.ts`

my PR that initially surfaced this problem did 2 things with Chip: renamed `Chip.types.d.ts` -> `Chip.types.ts` and used a different pattern for namespacing the `Chip` component along with `Chip.Prefix` and `Chip.Suffix`. bringing this into product, it (correctly) caused CI to fail on those violations of `Chip` that are missing required props or using non-existent props.

with that, I figured it was the `Chip.types.d.ts` change. however; doing that in isolation did not fix it. I had to also start exporting the types from `index.ts` for it to work, so I'm missing something there.

worth noting that every other component with subcomponents either exports _something_ from their `Component.types.ts` in `index.ts` OR they simply don't have a `Component.types.ts` and instead the types are defined in the component and we do `export * from './Component` - so I'm pretty confident that's part of the cause.

anyway I don't have the capacity to dig into this any further at the moment, but I want to unblock the Chip/Button PR without making it deal with a bunch of other problems - so I'm gonna do this in isolation, fix the Chip(s) that are wrong and then circle back on this some day when I have more time

taylorvnoj (approved):
LGTM, thanks for digging into it this far. Might be worth a ticket for follow up if you haven't written one already. Just so we have context/your thoughts down somewhere

PR #2498, opened 2025-04-21
docs: add guide to Figma libraries

chris-at-jobber:
## Motivations

There's some good Figma onboarding pieces scattered around internal wikis that would be helpful for new Jobber designers to have consolidated access to. Would like to link this on the Home page once it lands.

image

### Added

- Guide to our Figma library setup and general Jobber-specific approaches to file setup and structure

### Security

- Note: Figmas are not publicly viewable and are also have no real "trade secret" content - IMO this is secure enough. If the team feels strongly otherwise, I can leave the content in internal wikis, or we could look at some of the ideas around conditional authentication like we have in place on Triton. FWIW we already have embedded Figmas in our icon guidance docs and some patterns docs.

## Testing

- Run locally/in CF
- When not logged in to Figma, you should not see the Figma files previewed in the iframes

---

Random photo of Atlantis

### DIFF
--- a/docs/guides/figma-101.mdx
+++ b/docs/guides/figma-101.mdx
@@ -5,7 +5,23 @@ import { Banner, Tiles } from "@jobber/components";
 
 # Figma 101
 
-The basics to building with Atlantis in Figma.
+## How we use Figma at Jobber
+
+This guide will help you get oriented on the ways we organize our projects,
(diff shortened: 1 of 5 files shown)


taylorvnoj (approved):
I think it's a great move to put figma's in our docs like this - centralize as much as we can! LGTM

PR #2502, opened 2025-04-22
fix(components): Properly Exclude DefaultValue from InputTimev2 Props

ZakaryH:
## Motivations

we removed the functionality of `defaultValue` on `InputTime` v2 - since it's meant to be a controlled component. however the types are not correctly reflecting that, I believe it's because we had defined `version?: 1 | 2` and so it could meet that criteria of the "old" set of props and not exclude it like we'd like

## Changes

re-jigged the types a bit to properly tell TS that if you do `version={2}` there is no `defaultValue` prop

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

`<InputTime version={2} defaultValue={new Date()} .../>` should NOT do anything with the provided date, and TS should tell you that's invalid.

`<InputTime version={1} defaultValue={new Date()} ... />` should continue to work

`<InputTime defaultValue={new Date()} ... />` should continue to work

anything referencing `InputTimeProps` should continue to work. I made a new type that extends from the old one, making sure to not remove it. though nothing really should be referencing it, nor can I find anything trying to.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InputTime/InputTime.types.ts
+++ b/packages/components/src/InputTime/InputTime.types.ts
@@ -50,7 +50,10 @@ export interface InputTimeLegacyProps extends InputTimeProps {
 }
 
 export interface InputTimeRebuiltProps
-  extends Omit<InputTimeProps, "defaultValue" | "version"> {
+  extends Omit<
+    InputTimeProps,
+    "defaultValue" | "version" | "validations" | "onValidation"
+  > {
   /**
    * Version 2 is highly experimental, avoid using it unless you have talked with Atlantis first.
    */
--- a/packages/components/src/InputTime/InputTime.types.ts
+++ b/packages/components/src/InputTime/InputTime.types.ts
@@ -52,6 +52,9 @@ export interface InputTimeLegacyProps extends InputTimeProps {
 export interface InputTimeRebuiltProps
   extends Omit<InputTimeProps, "defaultValue" | "version"> {
   defaultValue?: never;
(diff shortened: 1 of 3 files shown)


ZakaryH:
image

image

these messages are decent I think. I'd perhaps like it to be a bit clearer that `defaultValue` doesn't exist on the component.

ZakaryH:
oh wait, that's because I was doing `defaultValue?: never`

if I go back to how we had it at one point, I get that superior messaging that it doesn't exist

image
image

packages/components/src/InputTime/InputTime.types.ts
```diff
@@ -43,15 +43,17 @@ export interface InputTimeProps
...
-   */
-  version?: 1 | 2;
+}
+
+export interface InputTimeLegacyProps extends InputTimeProps {
+  version?: 1;
 }
 
 export interface InputTimeRebuiltProps
   extends Omit<InputTimeProps, "defaultValue" | "version"> {
```
MichaelParadis, line 53: So I noticed that the rebuilt props provides hints for validations and `onValidation` which comes from RHF. I think it should be omitted from this as well

ZakaryH: ah right. good catch.

MichaelParadis (approved):
LGTM!

PR #2503, opened 2025-04-22
fix(components): Fix Type Definition Linking for Clearable

ZakaryH:
## Motivations

the `Clearable` type was not coming through in non-Atlantis projects using Atlantis. this is because there was a bad path to it that makes sense in Atlantis, but wouldn't be valid when the separate packages (hooks, components etc) are bundled and used.

we need to reference it in the correct way

credit goes to Jacob for figuring this out!

## Changes

updated `Clearable` path/source to work when we bundle Atlantis

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

use this pre-release build in another project. see that when using the `clearable` prop on components like `InputTime` `InputText` `Autocomplete` and more, that your IDE and the types correctly inform you what is available and what is invalid eg. `clearable={false}` is not OK

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/FormFieldTypes.ts
+++ b/packages/components/src/FormField/FormFieldTypes.ts
@@ -1,7 +1,7 @@
 import React, { ChangeEvent, ReactNode, RefObject } from "react";
 import { RegisterOptions } from "react-hook-form";
 import { XOR } from "ts-xor";
-import { Clearable } from "@jobber/hooks/src/useShowClear";
+import { Clearable } from "@jobber/hooks";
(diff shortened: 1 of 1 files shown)


ZakaryH:
a nice QoL improvement in Atlantis might be a linting rule that catches this sort of thing. it's so subtle and will work just fine when you're building and even testing it out within this repo.

edison-cy-yang (approved):
great catch @jdeichert @ZakaryH !

PR #2506, opened 2025-04-23
feat(components): Add Ref to Chip

ZakaryH:
## Motivations

Same as https://github.com/GetJobber/atlantis/pull/2470

except this is reduced scope not including Button. there's some extra complexity causing that one to give trouble to existing usages and their types. we have a couple options to play around with to get it to work, but in the meantime I'd rather get this Chip part merged and available.

both the Chip and Button changes are trying to get ahead of a requirement, but the Button is even more so since Combobox typically uses a Chip as an activator.

### Added

`ref` prop can now be passed to `Chip`

this enables some nice RHF behavior with say a required Combobox where on error, we'd focus the Chip and scroll to it for you to change what is needed

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Chip should be functionally identical to how it was before, including how we interact with its types.

the only new thing is that you can do `<Chip {...args} ref={myRef} />` and if you wanted to do a full integration with RHF and Zod, it would work using a `<Combobox.Activator><Chip {...args} ref={myRef}/></Combobox.Activator>` where the ref is being registered with RHF and has validation connected to this field. I have a video of this behavior in the linked original PR.

Changes can be

---

Random photo of Atlantis


MichaelParadis (approved):
LGTM!

PR #2507, opened 2025-04-23
feat(components): make onClear optional on FormFieldWrapper

edison-cy-yang:
## Motivations

For some inputs, they are not clearable (e.g. `Select`). However even if `clearable` is set to "never", `onClear` handler is still a required prop on `FormFieldWrapper`, making us create `handleClear` function that will never be used.

### Added

- 

### Changed

- `onClear` prop of `FormFieldWrapper` is now optional
- `onClick` prop of `ClearAction` now has a `?` conditional check.
- Removed empty `handleClear` function in Select rebuilt 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Test in legacy InputText's clearable story, where `onClear` handler comes from `useAtlantisFormFieldActions`. Confirm value is still being cleared.
- In Select rebuilt, `onClear` is not being passed into `FormFieldWrapper`. Confirm no typescript error.

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/FormField/components/ClearAction.tsx
+++ b/packages/components/src/FormField/components/ClearAction.tsx
@@ -6,7 +6,7 @@ interface ClearActionProps {
   /**
    * Click handler
    */
-  readonly onClick: () => void;
+  readonly onClick?: () => void;
   readonly visible?: boolean;
 }
 
@@ -19,7 +19,7 @@ export function ClearAction({
   return (
     <button
       className={styles.clearInput}
-      onClick={onClick}
+      onClick={() => onClick?.()}
       type="button"
       aria-label="Clear input"
       data-testid="ATL-FormField-clearButton"
--- a/packages/components/src/FormField/components/ClearAction.tsx
+++ b/packages/components/src/FormField/components/ClearAction.tsx
@@ -19,7 +19,7 @@ export function ClearAction({
   return (
     <button
       className={styles.clearInput}
-      onClick={() => onClick?.()}
+      onClick={onClick}
       type="button"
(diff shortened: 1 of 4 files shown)


packages/components/src/FormField/components/ClearAction.tsx
```diff
@@ -19,7 +19,7 @@ export function ClearAction({
   return (
     <button
       className={styles.clearInput}
-      onClick={onClick}
+      onClick={() => onClick?.()}
```
nad182, line 22: I'm curious why you needed to change this. Seems like the original code should be already doing what we want. `onClick` can be `undefined` according to the type definition above; as well as `button`'s _onClick_ can also be `undefined`.

image

nad182 (commented):
It is probably a nitpick, so feel free to ignore. Just something I noticed in Storybook.
According to your PR description:

> For some inputs, they are not clearable (e.g. Select). 

However, in Storybook, you can set the `clearable` prop to any of the given 3 values.

image

I looked at the `Select` component implementation and it seams like `clearable` always has a set/static value `"never"`, so I don't think this is going to be an issue - it seems to be just a Storybook thing. Not sure if it's worth doing anything about it, but wanted to call it out just in case.

packages/components/src/FormField/components/ClearAction.tsx
edison-cy-yang: you're right, for some reason I thought button's onClick cannot be undefined

nad182 (approved):
Looks good!

PR #2510, opened 2025-04-23
feat(components): Add maxLength prop to both Select versions

edison-cy-yang:
## Motivations

It turns out `maxLength` is not only used on `input` and `textarea`, but also fed into `useFormFieldWrapperStyles` to determine the width of the input. This PR adds the prop back to `SelectLegacy` that originally had it.

### Added

- 

### Changed

- Added `maxLength` to `SelectLegacyProps` and `SelectRebuiltProps`

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Select/Select.types.ts
+++ b/packages/components/src/Select/Select.types.ts
@@ -30,8 +30,12 @@ export interface SelectLegacyProps
       | "suffix"
       | "defaultValue"
       | "version"
-      | "maxLength"
-    > {}
+    > {
+  /**
+   * Changes the width to roughly the same size as the maximum character length
+   */
+  readonly maxLength?: number;
+}
 
 /**
  * Rebuilt version of the Select component without React Hook Form dependency.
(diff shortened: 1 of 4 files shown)


packages/components/src/Select/Select.types.ts
```diff
@@ -30,6 +30,7 @@ export interface SelectLegacyProps
       | "suffix"
       | "defaultValue"
       | "version"
+      | "maxLength"
```
ZakaryH, line 33: rather than inheriting the `maxLength` JS Docs that are defined on `FormFieldProps` that are going to say stuff about limiting the characters, I think it'd be nice to instead provide more accurate JS Docs for `Select` so something like

```
   | "version"
    > {
  /**
   * Changes the width of the Select to roughly the size of the maximum character length.
   */
  maxLength?: FormFieldProps["maxLength"];
}
```

that way, if someone chooses to read/use this - it's at least giving a good description of what it will actually do

edison-cy-yang: That makes sense

ZakaryH: I was thinking we'd still reference `FormFieldProps["maxLength"` but I suppose it doesn't matter much. I doubt we're going to be making any changes to that prop.

ZakaryH (approved):
LGTM

PR #2511, opened 2025-04-24
feat(components): Add support for composable modals - Take 2

MichaelParadis:
Reverts GetJobber/atlantis#2508
Original PR https://github.com/GetJobber/atlantis/pull/2475


packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -81,7 +85,17 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
     </div>
   );
 
+  if (parentNodeId) {
```
MichaelParadis, line 88: This is the big change from the initial PR. Floating UI breaks in JO so this only uses the floating UI tree logic if it is inside another Floating UI element (like modal). If not it will use the default portal behaviour

packages/components/src/Modal/Modal.tsx
```diff
@@ -39,7 +22,7 @@ export function Modal({
   secondaryAction,
   tertiaryAction,
   onRequestClose,
-}: ModalProps) {
+}: ModalLegacyProps) {
```
MichaelParadis, line 25: The other change from the original PR was I removed the `AtlantisPortalContent` stuff for allowing the reverse theme in the Modal since other issues cropped up

packages/components/src/Modal/Modal.rebuilt.tsx
```diff
@@ -0,0 +1,155 @@
...
+        <FloatingNode id={floatingNodeId}>
+          <FloatingPortal>
+            <AtlantisPortalContent>
+              <FloatingOverlay className={overlay}>
+                <FloatingFocusManager
+                  context={floatingContext}
+                  returnFocus={activatorRef?.current ? activatorRef : true}
+                  initialFocus={floatingRefs?.floating}
+                >
+                  <div
```
MichaelParadis, line 126: Moved this because I found that the overlay was slightly changed unintentionally, this better matches the dom structure of the original modal

Aiden-Brine:
Screenshot 2025-04-24 at 8 58 24 AM

The changes are looking good 👍 I was just wondering if we are okay with this level of opacity where you can see the lines of the input through the modal. It isn't the best look IMO

PR #2513, opened 2025-04-24
feat(components): Add Default Activator Ref to Combobox

ZakaryH:
## Motivations

Recently we added `ref` to `Chip` so that if you are using a Combobox to make a selection that is required on a form, specifically Zod + RHF though a similar effect could be achieved other ways, we'd have a ref to register with the form to scroll to and focus.

The only way to provide a ref would be thru a custom activator, which is not ideal if all you want is the default activator with a ref. Having to rebuild the activator just to access that isn't great.

So, to allow for the default to have a ref we're adding a `defaultActivatorRef` prop. The intent of the name is to communicate that it is ONLY going to work if you are using the default activator since the way COmbobox works is that if you want to override the default you provide `Combobox.Activator`, if you want the default you do not provide one at all.

### Added

`defaultActivatorRef` to top level `Combobox` to pass a ref to the `Chip` we use as the default activator

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Combobox/Combobox.types.ts
+++ b/packages/components/src/Combobox/Combobox.types.ts
@@ -62,7 +62,7 @@ export interface ComboboxProps {
   /**
    * A ref to the default activator button element.
    */
-  readonly defaultActivatorRef: React.RefObject<HTMLButtonElement>;
+  readonly defaultActivatorRef?: React.RefObject<HTMLButtonElement>;
 }
 
 export interface ComboboxCustomActivatorProps {
@@ -107,7 +107,7 @@ export interface ComboboxActivatorProps {
 export interface ComboboxTriggerProps
   extends Pick<ComboboxContentProps, "selected"> {
   readonly label?: string;
-  readonly activatorRef: React.RefObject<HTMLButtonElement>;
+  readonly activatorRef?: React.RefObject<HTMLButtonElement>;
 }
 
 export interface ComboboxOptionProps {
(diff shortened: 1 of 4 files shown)


packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -58,6 +58,11 @@ export interface ComboboxProps {
    * Should the Combobox display the loading state.
    */
   readonly loading?: boolean;
+
+  /**
+   * A ref to the default activator button element.
+   */
+  readonly defaultActivatorRef?: React.RefObject<HTMLButtonElement>;
```
MichaelParadis, line 65: Minor comment, I think this should should `React.Ref<HTMLButtonElement>` because the ref might be a callback ref (this is what RHF uses). That way consumers don't need to cast it

packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -102,6 +107,7 @@ export interface ComboboxActivatorProps {
 export interface ComboboxTriggerProps
   extends Pick<ComboboxContentProps, "selected"> {
   readonly label?: string;
+  readonly activatorRef?: React.RefObject<HTMLButtonElement>;
```
MichaelParadis, line 110: Minor comment, I think this should should React.Ref<HTMLButtonElement> because the ref might be a callback ref (this is what RHF uses). That way consumers don't need to cast it

MichaelParadis (changes requested):
Minor changes on the Ref type for ease of use with other libraries like RHF

packages/components/src/Combobox/Combobox.types.ts
ZakaryH: that's fair. I'm fine to change that here if it's the more flexible implementation

though worth noting that's not typically what we use, so we might need to make some updates to other components someday

image

PR #2515, opened 2025-04-25
feat(components): Add element prop to Text

taylorvnoj:
## Motivations

Allow users of the `Text` component to specify which HTML element they want to render, providing the same flexibility as the `Typography` component. This enables more semantic markup while maintaining consistent text styling.

### Added

- Added `element` prop to `Text` component that accepts all HTML elements supported by `Typography` (h1-h6, p, b, em, dd, dt, strong, span)
- Default value is "p" to maintain existing behavior
- Type is derived from `Typography` component to ensure consistency

## Testing

Use `Text` component with different element values:
2. Verify that each renders with the correct HTML tag while maintaining Text component styling
3. Verify that existing Text component usage without the element prop continues to render as a paragraph

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Text/Text.tsx
+++ b/packages/components/src/Text/Text.tsx
@@ -2,6 +2,12 @@ import React, { PropsWithChildren } from "react";
 import { Typography, TypographyOptions, TypographyProps } from "../Typography";
 
 export interface TextProps {
+  /**
+   * The HTML element to render the text as.
+   * @default "p"
+   */
+  readonly element?: TypographyProps["element"];
+
   readonly maxLines?:
     | "single"
     | "small"
@@ -44,6 +50,7 @@ export function Text({
   variation = "default",
   size = "base",
   align = "start",
+  element = "p",
   children,
   maxLines = "unlimited",
   UNSAFE_className,
@@ -70,6 +77,7 @@ export function Text({
 
   return (
     <Typography
+      element={element}
       textColor={textColors[variation] as TextColor}
       size={size}
       numberOfLines={maxLineToNumber[maxLines]}
--- a/packages/components/src/Text/Text.tsx
+++ b/packages/components/src/Text/Text.tsx
@@ -1,12 +1,17 @@
 import React, { PropsWithChildren } from "react";
 import { Typography, TypographyOptions, TypographyProps } from "../Typography";
 
+type TextElement = Extract<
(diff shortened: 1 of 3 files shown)


packages/components/src/Text/Text.tsx
```diff
@@ -2,6 +2,12 @@ import React, { PropsWithChildren } from "react";
 import { Typography, TypographyOptions, TypographyProps } from "../Typography";
 
 export interface TextProps {
+  /**
+   * The HTML element to render the text as.
+   * @default "p"
+   */
+  readonly element?: TypographyProps["element"];
```
MichaelParadis, line 9: One minor comment, is the I think you should [extract](https://www.typescriptlang.org/docs/handbook/utility-types.html#extracttype-union) all elements but `h<x>` or [exclude](https://www.typescriptlang.org/docs/handbook/utility-types.html#excludeuniontype-excludedmembers) the `h<x>` elements to keep text as text and headings as heading

MichaelParadis (commented):
Just one comment on the types of elements Text should support.

packages/components/src/Text/Text.tsx
taylorvnoj: Great call out - done here 43bcf2e65591b99369586edd0890ed5a0ff559bb
I decided to extract everything except for headings to maintain that clear separation 🫡

packages/components/src/Text/Text.tsx
```diff
@@ -1,12 +1,17 @@
 import React, { PropsWithChildren } from "react";
 import { Typography, TypographyOptions, TypographyProps } from "../Typography";
 
+type TextElement = Extract<
+  TypographyProps["element"],
+  "p" | "b" | "em" | "dd" | "dt" | "strong" | "span"
```
chris-at-jobber, line 6: Semantically I bet you could probably end up with a Text that's somehow an h4/5/6 🤔

MichaelParadis: This was my thoughts behind it https://github.com/GetJobber/atlantis/pull/2515#discussion_r2060498140
We can always add it if need be but I personally think we can do this and expand as needed. Basically headings looking like Text seems odd

MichaelParadis (approved):
LGTM!

PR #2523, opened 2025-04-28
fix(components-native): Adjusting Type mismatch in ThumbnailList

scotttjob:
## Motivations

1. We had a report that there was a type mismatch from an older version of ThumbnailList, which I confirmed to be true. We used to not require `createThumbnail` which is confirmed by following the codepath down. It's allowed to be undefined elsewhere, and the underlying component expects it can be undefined and it creates a dummy function in that instance.

1. Will generate a pre-release so this can be tested.

## Changes

1. `createThumbnail` is no longer required on ThumbnailList.

## Testing

1. You should be able to use `ThumbnailList` in a react-native application without requiring the 'createThumbnail' prop.

Changes can be

---

Random photo of Atlantis


classtronaut:
Tested this within Jobber mobile and it looks to resolve the issue, comparing before and after seems all good! Thanks

PR #2527, opened 2025-04-28
feat(components): InputEmail V2 without RHF

edison-cy-yang:
## Motivations

Current `InputEmail` uses `FormField` for the input styling and logic, and `FormField` uses RHF. We want to decouple the use of RHF so RHF instances outside of the component do not cause issues.

The decoupled version of `InputEmail` has version=2

### Added

- `InputEmail.rebuilt` component without dependency on RHF

### Changed

- `index.ts` -> `index.tsx` to conditionally render different `InputEmail` versions with `version` prop

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

`InputEmail` v1 should be unchanged

`InputEmail` v2 should continue to work how you'd expect it:

onBlur/onFocus/onChange
props like inline, disabled, size etc

"Version comparison" story added to showcase v1 and v2 with the same props

Changes can be

---

Random photo of Atlantis


packages/components/src/InputEmail/InputEmail.rebuilt.tsx
```diff
@@ -0,0 +1,92 @@
...
+      clearable={props.clearable ?? "never"}
+      onClear={handleClear}
+      wrapperRef={wrapperRef}
+      maxLength={props.maxLength}
+      disabled={props.disabled}
+      type="email"
+      value={props.value}
+    >
+      <input
+        {...fieldProps}
```
edison-cy-yang, line 81: As I was going through this part, I almost wondered if we should be explicitly setting these props instead of just spreading most of them, considering that we want to avoid abstraction and duplicate now, then figure out what can be abstracted

docs/components/InputEmail/Web.stories.tsx
```diff
@@ -19,3 +22,233 @@ export const Basic = BasicTemplate.bind({});
...
+          "Size small",
+          {
+            placeholder: "Small input",
+            size: "small",
+            ...extraProps,
+          },
+          "sizeSmall",
+        )}
+
+        {renderBothVersions(
```
MichaelParadis, line 173: I think a [readonly](https://github.com/GetJobber/atlantis/blob/JOB-121642-do-it/docs/components/InputText/Web.stories.tsx#L110) variant would be useful like `InputText` has

packages/components/src/InputEmail/InputEmail.rebuilt.tsx
```diff
@@ -0,0 +1,92 @@
...
+        ref={inputRef}
+        type="email"
```
MichaelParadis, line 83: We should probably add [inputMode](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#inputmode) as well

packages/components/src/InputEmail/InputEmail.types.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Experimental version 2 of the InputEmail component.
+ * Do not use unless you have talked with Atlantis first.
+ */
+export interface InputEmailRebuiltProps
+  extends Omit<
+    React.InputHTMLAttributes<HTMLInputElement>,
+    | "onChange"
+    | "onBlur"
+    | "maxLength"
```
MichaelParadis, line 23: This omit isn't needed since `input` takes a `maxLength` that is the same as the one you have defined below

packages/components/src/InputEmail/InputEmail.rebuilt.tsx
edison-cy-yang: According to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inputmode#email), inputs that require email should just use type="email" instead

MichaelParadis: That makes sense to me. Feel free to ignore this. I think I confused textInput on RN which if you don't do that it will give you a regular keyboard without the @ key easily accessible

edison-cy-yang: let me test if it makes a difference in mobile web

MichaelParadis: It looks to work for me on Android

edison-cy-yang: looks like type="email" already prompts for the keyboard with @

MichaelParadis (approved):
LGTM!

PR #2528, opened 2025-04-29
feat(components-native): BREAKING CHANGE: Upgrade React Native Version

ZakaryH:
## Motivations

We are behind on our React Native version for components-native, and would like to catch up a bit.

please ignore the branch name. I optimistically had hoped to get to 0.78 with this branch. it proved to be more effort that we're going to have to tackle with a different approach.

### Added

- 

### Changed

while this was incrementally approached, the end result of this PR is that we will be "on" RN 0.77.

in reality we're not using anything new from RN 0.76, so it's more that we are _compatible_ with RN 0.77.

**bumped our peer dependencies up quite a bit from >=0.69 to >=0.76 which is the reason for the breaking change tag**

the reason for this is because it's a bit safer for us to have our minimum close to what we build with to prevent say using a 0.76 feature, but our minimum version says 0.73 is fine - and then it won't work

as far as what we had to update in our own library, luckily nothing about what we're currently using in the actual components needed to change. it was all config, test, package version and type updating.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

I've noted the aspects that warrant some actual QA to verify they work as intended (SwitchBase, Glimmer)

aside from that it's basically confirming that everything works: CI, tests, docs site web components, docs site native components (where applicable), package building/pipeline, and finally trying out the packages in product where it should work as intended and there's no CI issues there due to behavior changes or missing types.
Changes can be

---

Random photo of Atlantis


packages/components-native/src/Form/Form.test.tsx
```diff
@@ -545,7 +545,7 @@ describe("Form", () => {
   });
 
   describe("Safe Area", () => {
-    it("does render a safe area when there's a saveButtonOffset is provided", () => {
+    it("does render a safe area when there's NO saveButtonOffset provided", () => {
```
ZakaryH, line 548: test assertion didn't make sense

pretty minor thing I came across running this test that was failing for totally different reasons. not necessary but it wasn't accurate before.

packages/components-native/src/__mocks__/__mocks.ts
```diff
@@ -22,6 +22,7 @@ jest.mock("react-native-reanimated", () => {
     FadeIn: {
       duration: () => undefined,
     },
+    useEvent: () => ({ listeners: [] }),
```
ZakaryH, line 25: this the the one we actually use in our test suite.

I don't get why we have 2 separate implementations, where we tell consumers to use `jestSetup.js` but we don't use it ourselves? seems strange since the file is right there.

also, I don't think this is the right way to handle this. I was hitting this exact issue of `useEvent` not being found, and there's a clearly recommended way to setup your jest tests.
https://github.com/software-mansion/react-native-reanimated/issues/6335#issuecomment-2308806958

checking in `jobber-mobile` we're doing it via `require('react-native-reanimated').setUpTests();`

and then also using the `jestSetup` from `components-native` but I would bet money that if we removed the reanimated mocks, tests over there would still pass.

packages/components-native/src/Switch/components/BaseSwitch/__snapshots__/BaseSwitch.test.tsx.snap
ZakaryH, line 1: I couldn't find any great documentation on the `handlerTag` which is coming from RNGH's `Switch` component.

digging around in the source code you can find references to it, so I'm thinking it's a pretty internal implementation detail that we happened to be capturing in a snapshot.

I'll wanna verify the behavior is good. assuming it is, I think this is not something we need to care about since it's possible these values are (somewhat) arbitrary numbers. again, assuming `-1` doesn't signify something is not working.

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -146,7 +146,6 @@ export interface SimpleConfirmationModalProps
 
 export interface ComplexConfirmationModalProps
   extends BaseConfirmationModalProps {
-  readonly ref: Ref<ConfirmationModalRef>;
```
ZakaryH, line 149: I'll have to check the notes, but it seems like something around types and refs got stricter.

```

type ConfirmationModalProps =
  | (SimpleConfirmationModalProps & (StringMessage | ChildrenMessage))
  | (ComplexConfirmationModalProps & (StringMessage | ChildrenMessage));

export const ConfirmationModal = forwardRef(function ConfirmationModalInternal(
  {...},
  ref: Ref<ConfirmationModalRef>,
```

this is being used in a forwardRef, so it makes sense to me that the `ref` definition shouldn't be here, it should be  on the `forwardRef` function call

ZakaryH: this was causing `npm i` to fail on 0.74.0+

packages/site/package.json
```diff
@@ -42,7 +42,7 @@
     "@types/babel__standalone": "^7.1.9",
     "@types/node": "^22.10.1",
     "@types/prismjs": "^1.26.4",
-    "@types/react": "18.2.0",
+    "@types/react": "18.2.45",
```
ZakaryH, line 45: couple notes on this, first I don't know why but the tsconfig that we have which says to use `./node_modules/@types/react` seems like a fine thing to do but for some reason, even after installing a different version of `@react/types` within `packages/site` it was not updating and we were still on `18.2.0`

I'm skeptical that's the entire problem since this is passing on `master` with that same version. something about bumping RN to `0.74.5` must have triggered a peer dependency change or something along those lines because that's when this started blowing up

anyway, the problem was that everything in the visual tests folder was very upset with the types just like this 
https://github.com/react-bootstrap/react-bootstrap/issues/6717
https://github.com/facebook/react/issues/24304#issuecomment-1094565891

the fix was removing node modules, and re-installing the version I wanted which doesn't have the issue

ZakaryH:
all of the `*props.json` updates are coming from the updated `@types/react` presumably where there are really only a couple changes in a bunch of files

- ref -> legacyRef and now it has a description
- `needsOffscreenAlphaCompositing` and some some RN `ViewProps` are now showing up

the latter is because we are directly extending `ViewProps`
`export interface ActivityIndicatorProps extends ViewProps {...`

debatable if we should or not without `Pick`, but that's a separate topic.

only thing to consider testing with this is that the `ref`s still work on the components implementing what looks to be forward refs, and then using the package in product to make sure there's no conflicts with the new type compared to what people might have been typing their refs as

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
ZakaryH: having dealt with more type problems in `site` something definitely bumped the version of `@types/react` that was being used causing all sorts of TS errors

ZakaryH:
**RN 0.74.5 (and accompanying changes) testing**

running the app on this pre-release looks and feels fine. nothing blowing up in the console either.

image

packages/components-native/src/Disclosure/__snapshots__/Disclosure.test.tsx.snap
```diff
@@ -370,7 +370,7 @@ exports[`should not render the caret when the Disclosure is empty 1`] = `
     }
     accessible={true}
     collapsable={false}
-    focusable={true}
+    focusable={false}
```
ZakaryH, line 373: this change is expected

image

where in TouchableOpacity prior to `0.75.0` the logic for `focusable` was 

```
        focusable={
          this.props.focusable !== false && this.props.onPress !== undefined
        }
```

and in `0.75.0`

```
        focusable={
          this.props.focusable !== false &&
          this.props.onPress !== undefined &&
          !this.props.disabled
        }
```

and our `TouchableOpacity` in this snapshot is disabled

```
    <TouchableOpacity
      activeOpacity={tokens["opacity-pressed"]}
      onPress={onToggle}
      disabled={isEmpty}
    >
```

packages/site/src/content/ActivityIndicator/ActivityIndicator.props-mobile.json
ZakaryH, line 1: a side effect of extending ViewProps

`export interface ActivityIndicatorProps extends ViewProps {`

packages/components-native/babel.config.js
ZakaryH, line 1: https://github.com/facebook/react-native/issues/48228#issuecomment-2547160218

ZakaryH:
ActivityIndicator's props.json keeps needing to be updated because we're extending ViewProps. not sure it's of benefit to us to do that, might consider `Pick` ing what we need rather than grabbing everything.

ZakaryH:
Kapture 2025-05-05 at 16 04 12

I haven't been paying close enough attention during my smoke testing of previous builds to definitively say if this is a new behavior after my upgrades or not. I'll have to check against current master. it did catch my eye though.

packages/components-native/src/Form/Form.tsx
```diff
@@ -138,6 +138,7 @@ function InternalForm<T extends FieldValues, S>({
 
   const styles = useStyles();
 
+  // Comment to trigger build once more
```
ZakaryH, line 141: need to remove these

packages/components/src/Chip/Chip.tsx
```diff
@@ -9,6 +9,7 @@ import { useChildComponent } from "./hooks/useChildComponent";
 import { Typography } from "../Typography";
 import { Tooltip } from "../Tooltip";
 
+// Comment to trigger build
```
ZakaryH, line 12: this too

packages/site/package.json
scotttjob: It's the .vite cache in the node_modules folder holding on to previous versions. I have a PR floating around the updates `site` to pull from `src` dirs instead of `dist` dirs which fixes the cache problem, but potentially opens up new issues we haven't fully investigated.

I suppose if we can run the e2e tests properly with the `src` change in place, we should feel confident enough to roll that out

package.json
```diff
@@ -135,7 +135,7 @@
     "react-dom": "18.2.0",
     "react-ga4": "^2.1.0",
     "react-intl": "^6.4.2",
-    "react-native": "0.73.9",
+    "react-native": "^0.77.0",
```
scotttjob, line 138: I know we chatted about this last week, but I can't remember if we talked about why these need to be different?

In this root package.json you've got '^0.77.0' and in the components-native package.json we've got `>=0.76.0'

is it just different syntax for regular deps vs peerdeps, or was there something else I'm just blanking on now.

scotttjob (commented):
Just dropping a comment to say this looks good. Happy to hear testing is going well, I still haven't run things directly (aside from loading storybook and clicking around)

Have we reached out to our mobile team yet, for them to give the branch a whirl? Or is that at the end, when we're relatively confident the branch is in a good state?

Is this issue in Storybook where the props are showing up behind the experience, is this known?

If Storybook ends up being a blocker here btw, let me know. We've been looking for an excuse to start moving away from our custom-built Storybook 7 experience towards a vanilla Storybook-latest, so if this is the impetus we need then I'm all for it!

image

PR #2538, opened 2025-05-05
feat(components): Format File Composed

scotttjob:
## Motivations

1. We had a request to make some style changes to FormatFile, so as per our recent efforts, we made it composable instead so consumers can change what they need.

## Changes

1. FormatFile is now composable.

### Changed

- Format File refactored to be composable.

### Fixed

- FormatFile is now in a composable state for extension.

## Testing

- Everything should still work as-is.

Changes can be

---

Random photo of Atlantis


timoteialbu (approved):
LGTM - First of all, thanks so much for pushing this through with no notice today! I am using it in this [PR](https://github.com/GetJobber/jobber-frontend/pull/2806) and did a bunch of testing and made some changes based on my needs. Hopefully those are okay.

edison-cy-yang (approved):
Non-composed version continues to work in storybook, and Timotei tested the composed version in JF. lgtm!

A quick followup we should do though is expose the props types of components like FormatFile.Body. Components like FormatFile.ThumbnailContainer can maybe live without props exported since they're very much simple containers.

packages/components/src/FormatFile/FormatFile.tsx
```diff
@@ -46,86 +17,213 @@ export function FormatFile({
...
+  children,
+  className,
+}: {
+  readonly children: React.ReactNode;
+  readonly className?: string;
+}) {
+  return <div className={className}>{children}</div>;
+};
+
+FormatFile.Body = function FormatFileBody({
```
edison-cy-yang, line 140: I will be moving the props for some of these sub components into `types.ts` and exporting them before we actually ship this

PR #2540, opened 2025-05-06
feat(components-native): BREAKING CHANGE: Upgrade to React Native 0.77

ZakaryH:
## Motivations

Same concept as https://github.com/GetJobber/atlantis/pull/2528

this was split off because the other PR was getting somewhat large. in practice I wouldn't want us to merge the other PR without this one so that we do one big jump from 0.73->0.77

if it's not helpful having as a separate PR we can merge in into the other one and review it all, since that's effectively what would be happening if you review this one built off the others. there's not really a point to doing QA twice.

## Changes

again, nothing in our components needed to change luckily. just the way we test components and then dependencies of RN

### Added

- 

### Changed

**INCREASED PEER DEPENDENCIES OF `components-native` TO MATCH THE VERSION WE'RE BUILDING WITH**

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

same process as the other PR, I'd recommend testing on this one to only have to do it once.

all CI, tests, building should work as expected. site, components, web components should all work.

Glimmer in theory changed but I've checked it out using this version and it's still doing what it is supposed to do, but worth double checking.

ideally trying things on both a physical Android and iOS device is preferable, but we will have some time for QA when this merges into the release candidate as long as there's nothing clearly amiss.

Changes can be

---

Random photo of Atlantis


ZakaryH:
the other PR was getting big

0.75 and 0.76 didn't require a ton for changes whereas 0.77 is looking like it'll require a bit more work, so I'm making this an isolated PR for it to be easier to review

package-lock.json
```diff
@@ -36614,9 +36719,9 @@
       }
     },
     "node_modules/react-devtools-core": {
-      "version": "5.3.2",
-      "resolved": "https://registry.npmjs.org/react-devtools-core/-/react-devtools-core-5.3.2.tgz",
-      "integrity": "sha512-crr9HkVrDiJ0A4zot89oS0Cgv0Oa4OG1Em4jit3P3ZxZSKPMYyMjfwMqgcJna9o625g8oN87rBm8SWWrSTBZxg==",
+      "version": "6.1.1",
```
ZakaryH, line 36722: major version bump, though it's for dev-tools so I'm not particularly worried

ZakaryH:
Disclosure tests are failing on snapshots
image
if you check the element causing it, it's the `ScrollView`
image

there's a mention of some changes around ScrollView which I assume are related
https://reactnative.dev/blog/2025/01/21/version-0.77#other-breaking-changes

packages/components-native/src/Disclosure/Disclosure.test.tsx
ZakaryH, line 1: admittedly this is a bit more than I needed to do to fix the tests

I just don't love snapshots and would prefer to test in a more reliable way that is less susceptible to failing due to things like the underlying `View` or `ScrollView` changing

ZakaryH:
there's an actual bug with Disclosure where we're not accounting for the padding-top causing it to get cut off

image

there are a few ways to solve this (possibly with the new box sizing that we'll now have available?), but I'm going to make a ticket for it since it's an existing issue

I'm also at a loss for why the Disclosure is built the way it is

```
    <ReanimatedScrollView
      scrollEnabled={false}
      showsHorizontalScrollIndicator={false}
      showsVerticalScrollIndicator={false}
      style={[styles.contentContainer, animatedStyle]}
    >
```

why are we using a Reanimated Scroll View hard coded to not be scrollable with all the scroll indications off?

I'll assume there was an issue with Reanimated at the time, but there's really no reason to not use a view now - or if you do want to set a max height, then allow scrolling if the content exceeds that amount, don't hard code it to false

packages/components-native/src/Glimmer/Glimmer.test.tsx
ZakaryH, line 1: this test was failing due to a change in how React Native handles looping animations

>Animation
>Native looping animation do not send a React state update every time the loop ends.
https://reactnative.dev/blog/2025/01/21/version-0.77#other-breaking-changes

which is what Glimmer does

I tried a few different approaches, re-rendering and trying to verify the animated style value but nothing was working. doing this for now, which is a bit different of a way of trying to verify the same behavior.

maybe there's a better way but I'm going to stick with this for now for the purposes of moving forward.

ZakaryH: I've verified the component works fine in reality, so this is specifically an issue around how to test it

ZakaryH:
all the props.json updates are coming from RN props

3 of them are some new possible values for `textContentType`
https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Components/TextInput/TextInput.d.ts#L270-L317

ZakaryH:
using the prerelease in mobile on both iOS and Android simulator seemed fine :)

ZakaryH:
looking over the changes in the other PR and this one, there's extremely little that had to change in our actual implementations in fact almost all the changes minus some babel config were all tests, props and other package updates which is to say nothing about our compatibility changed in any meaningful way between 0.73 and 0.77

with that in mind, I think it's OK to keep `components-native` as `>=0.69.2` for the react-native peer dep. nothing we've used is only available in say 0.77 that would need us to increase that requirement

as such, if we're not increasing our peer dep requirement, it's technically not even a breaking change that would require a major version bump of components-native. our stuff works with 0.77, but if you're still running 0.70 it would still work.

packages/components-native/src/Form/Form.tsx
```diff
@@ -138,7 +138,7 @@ function InternalForm<T extends FieldValues, S>({
 
   const styles = useStyles();
 
-  // Comment to trigger build once more
+  // Comment to trigger build once more and again
```
ZakaryH, line 141: remove

packages/components/src/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -163,7 +163,7 @@ type ConfirmationModalProps =
   | (SimpleConfirmationModalProps & (StringMessage | ChildrenMessage))
   | (ComplexConfirmationModalProps & (StringMessage | ChildrenMessage));
 
-// Comment to trigger build once more
+// Comment to trigger build once more and again
```
ZakaryH, line 166: remove

packages/components-native/package.json
```diff
@@ -70,7 +70,7 @@
     "date-fns-tz": "^2.0.0",
     "react": "^18.2.0",
     "react-intl": "^6.4.2",
-    "react-native": ">=0.69.2",
+    "react-native": ">=0.77.0",
```
ZakaryH, line 73: doing the opposite of this comment https://github.com/GetJobber/atlantis/pull/2540#issuecomment-2860211146

after talking with Scott, we've decided it's safer and better for Jobber for this to closer match the actual version we are building with

keeping it as >= to avoid blocking the app, but worth noting we can't actually guarantee compatibility with say 0.78 yet

we might work fine with it, we might not. it's unclear right now. 

bumping the peer deps is a breaking change though so I'll update the title and tagging on both of these PRs

ZakaryH: with this change, it'll be important to try installing the package in product

I'll add another comment and trigger a new prerelease build once we get a passing CI run

ZakaryH: ah, yeah this is no good. it's too high, our own app is slightly behind this version. I'll reduce it slightly.

PR #2541, opened 2025-05-06
feat(components): Add InputDate prop to replace invalid values with last value on blur

jdeichert:
## Motivations

This is a hacky workaround for a problem we're currently facing. I'm looking at better solutions and will file followup tickets as necessary.

I already found an improvement and will ship this separately after it's reviewed: https://github.com/GetJobber/atlantis/pull/2542

### Added

- Added a new `InputDate` (v1) prop `restoreLastValueOnBlur` which replaces empty/invalid values with the last value when the form field is blurred

## Testing

Test out this story locally in`docs/components/InputDate/Web.stories.tsx`

1. Click the textarea
2. Tab to focus the input date field
3. Backspace the value
4. Shift tab to go back to the textarea
    * Observe the form field input replaces the empty value with the previous value
5. Tab back into the input date field
6. Enter an invalid date string
7. Shift tab to go back to the textarea
    * Observe the form field input replaces the bad value with the previous value
8. Confirm the normal date picker works as expected (mouse click to focus which opens the date picker)

```jsx
const BasicTemplate: ComponentStory<typeof InputDate> = args => {
  const [date, setDate] = useState(new Date("11/11/2011"));

  return (
    <Content>
      <textarea />
      <InputDate
        value={date}
        restoreLastValueOnBlur
        onChange={d => {
          if (!d) return;
          setDate(d);
        }}
      />
    </Content>
  );
};
```

### DIFF
--- a/packages/site/src/content/InputDate/InputDate.props.json
+++ b/packages/site/src/content/InputDate/InputDate.props.json
@@ -90,7 +90,7 @@
         "defaultValue": {
           "value": "false"
         },
-        "description": "Whether to replace empty/invalid values with the original value on blur",
+        "description": "Experimental:\nWhether to replace empty/invalid values with the original value on blur.\n\nThis solves an immediate UX problem and is likely to change in the future.\nIt prevents the input from retaining empty/invalid values when the user tabs\ninto it, clears the value, and tabs away. In this scenario, the original\nvalue will be restored.",
         "name": "restoreLastValueOnBlur",
         "parent": {
           "fileName": "packages/components/src/InputDate/InputDate.types.ts",


JeffMarvin:
@jdeichert Probably not required (because I think you'd need to recreate the PR) but watch out for upper case letters in branch names. Can cause issues on Mac sometimes.

packages/components/src/InputDate/InputDate.tsx
```diff
@@ -54,6 +58,17 @@ export function InputDate(inputProps: InputDateProps) {
                 inputProps.onBlur && inputProps.onBlur();
                 activatorProps.onBlur && activatorProps.onBlur();
                 setIsFocused(false);
+
+                if (inputProps.restoreLastValueOnBlur) {
```
JeffMarvin, line 62: A comment here describing why we're allowing this and what issue it solves would be good. (Could even say it is experimental and may change.)

Same thing with the prop - a comment and status of experimental may help future maintainers.

PR #2547, opened 2025-05-09
feat(components): Chips Experiment

ZakaryH:
## Motivations

there's a request to support a custom template for the menu items in dismissible chips, and we've got a reasonably good idea of how to solve that with a custom render function. the only tricky part about that is if you want the render function to provide you with the data that corresponds to each item, you'll have to pass that in - meaning we have to update the types on `Chip` to accept a new prop that can have pretty much anything as a key. we've done this before on Menu and some other custom render function components

the next part is trickier though. we also want to be able to search and filter based on these arbitrary values. integrating that into Chip's search logic is not really an option. we have no idea what the data will be or if it's relevant to a search, we'd have no way of knowing. Chip's search works in a very complex way that I'd want to change if it was staying around, and there's a bug with it currently being called when it shouldn't and it's not actually debounced.

furthermore, there's a bunch of internal logic not only filtering, but using the search value and results of the filtering to decide if we should show the "+" button to add, and the content is based on the search term. note that this is NOT working with this example yet. we should be able to solve that with a `customRenderMenuItems` where you are responsible for rendering all the menu items yourself, and then you can include an extra one at the end. though that approach does also mean that you have to implement all the handlers yourself which is a lot.

with that in mind, it's best to delegate the search to the consumer. the only ways I see us doing that is either by providing some sort of search logic prop where we will have the search term as an argument, then you provide the function that it will use.

Or, what I've tried here. add a separate code path that just ignores almost all of the logic around filtering, searching, selected and defer the responsibilities to the consumer.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/docs/components/Chips/Web.stories.tsx
+++ b/docs/components/Chips/Web.stories.tsx
@@ -3,6 +3,8 @@ import { ComponentMeta, ComponentStory } from "@storybook/react";
 import { Content } from "@jobber/components/Content";
 import { Chip, Chips } from "@jobber/components/Chips";
 import { Text } from "@jobber/components/Text";
+import { MenuItemsAPI } from "@jobber/components/Chips/ChipsTypes";
+import { ChipProps } from "@jobber/components/Chip/Chip.types";
 
 export default {
   title: "Components/Selections/Chips/Web",
@@ -212,6 +214,21 @@ const ControlledSelectionTemplate: ComponentStory<typeof Chips> = () => {
     setSelected(prev => [...prev, value]);
   };
 
+  const customRenderMenu = (
+    options: ChipProps[],
+    menuItemsAPI: MenuItemsAPI,
+  ) => {
+    return (
+      <div>
+        {options.map(option => (
+          <div onBlur={menuItemsAPI.handleCancelBlur} key={option.value}>
+            {option.label}
+          </div>
+        ))}
+      </div>
+    );
+  };
+
   return (
     <>
       <Text>
@@ -231,6 +248,7 @@ const ControlledSelectionTemplate: ComponentStory<typeof Chips> = () => {
         onChange={handleSelect}
         onCustomAdd={handleCustomAdd}
         onSearch={handleSearch}
+        customRenderMenu={customRenderMenu}
       >
         {displayedOptions.map(name => (
           <Chip key={name} label={name} value={name} />
--- a/docs/components/Chips/Web.stories.tsx
+++ b/docs/components/Chips/Web.stories.tsx
@@ -75,31 +75,65 @@ export const MultiSelect = MultiSelectTemplate.bind({});
 MultiSelect.args = {};
 
 const SelectionTemplate: ComponentStory<typeof Chips> = args => {
-  const {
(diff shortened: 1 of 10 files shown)


ZakaryH:
this sort of works? I wonder how it's going to interact with onLoadMore though. maybe you'd just have to keep adding to your list and it'll be ok? maybe not the most performant but at least functional.

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -57,12 +56,6 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     update?.();
   }, [allOptions, menuOpen, update, options]);
 
-  useEffect(() => {
-    handleDebouncedSearch(searchValue, options);
-
-    return handleDebouncedSearch.cancel;
-  }, [searchValue, options]);
```
ZakaryH, line 64: while we should avoid doing this because it would be calling the search at times where we would not want to call the search, I'm not confident this isn't going to re-introduce the problem it was added to address

docs/components/Chips/Web.stories.tsx
```diff
@@ -121,3 +145,130 @@ Selection.parameters = {
...
+
+    if (value === "") {
+      setDisplayedOptions(allOptions.slice(0, 7)); // Reset to initial set
+    } else {
+      // Filter options based on search term
+      // Note: We don't need to include selected options in the filtered list
+      // The component will ensure selected chips remain visible
+      const filteredOptions = allOptions.filter(option =>
+        option.toLowerCase().includes(value.toLowerCase()),
+      );
```
ZakaryH, line 201: here is where one would plop in all their logic

you've received the search term, now it's up to you to see if that term exists in 1+ fields - ideally in a performant way such that you're not filtering N x M amount of times where N is the number of fields you have to search for a given entry and M is the amount of entries

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -107,24 +100,36 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
...
+                  generateDescendantId,
+                  handleSelectOption,
+                  handleSetActiveOnMouseOver,
+                  handleCancelBlur,
+                  handleEnableBlur,
+                  handleShowInput,
+                  activeIndex,
+                  menuListOptionStyles: styles.menuListOption,
+                  activeOptionStyles: styles.activeOption,
+                })
```
ZakaryH, line 114: not the most elegant that each and every one of these has to be implemented, but what this does enable is that someone can drop their own button at the end of these items

however if you want the default one there's not really a great way to get that with this setup because we opted to skip out on all the search behavior and that's tied to the internal search. we'd need another way to have both.

PR #2548, opened 2025-05-09
feat(components): Add semantic token support for Grid gap spacing

taylorvnoj:
## Motivations

This change improves the flexibility of the Grid component by allowing developers to use semantic tokens for spacing, maintaining consistency with our design system's spacing scale. The implementation aligns better with how we handle spacing across other components.

### Added

- Support for semantic tokens (`small`, `base`, `large`, etc) in Grid gap spacing
- New Storybook examples demonstrating different gap spacing options
- Use of type `GapSpacing` for semantic token support

### Changed

- Grid gap implementation from CSS class-based to style-based using tokens
- Default gap behavior to use `spaceTokens.base` when `gap={true}`
- Grid component tests to verify token-based spacing

## Testing

Check out the new story http://localhost:6005/?path=/story/components-layouts-and-structure-grid-web--spacing to see that we can either allow the boolean/default base spacing or assign a different spacing token

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/Grid/Grid.tsx
+++ b/packages/components/src/Grid/Grid.tsx
@@ -3,12 +3,19 @@ import classNames from "classnames";
 import styles from "./Grid.module.css";
 import alignments from "./GridAlign.module.css";
 import { GridCell } from "./GridCell";
+import { GapSpacing } from "../sharedHelpers/types";
+import {
+  getMappedAtlantisSpaceToken,
+  spaceTokens,
+} from "../sharedHelpers/getMappedAtlantisSpaceToken";
 
 interface GridProps {
   /**
-   * Add spacing between elements.
+   * Add spacing between elements. Can be a boolean for default spacing,
+   * or a semantic token for custom spacing.
+   * @default true
    */
-  readonly gap?: boolean;
+  readonly gap?: boolean | GapSpacing;
 
   /**
    * Adjust the alignment of columns. We only support a few select properties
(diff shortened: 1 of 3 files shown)


packages/components/src/Grid/Grid.tsx
```diff
@@ -3,12 +3,19 @@ import classNames from "classnames";
...
 
 interface GridProps {
   /**
-   * Add spacing between elements.
+   * Add spacing between elements. Can be a boolean for default spacing,
+   * or a semantic token for custom spacing.
+   * @default true
    */
-  readonly gap?: boolean;
+  readonly gap?: boolean | GapSpacing;
```
edison-cy-yang, line 18: I'm not sure if I am onboard with the approach of `gap` being boolean or GapSpacing. It seems to overcomplicate things that should've been straight forward. Can we perhaps have a `none` or `zero` space token with 0px?

packages/components/src/Grid/Grid.tsx
```diff
@@ -30,12 +37,24 @@ export function Grid({
   gap = true,
   children,
 }: GridProps) {
-  const classnames = classNames(styles.grid, alignments[alignItems], {
-    [styles.gap]: gap,
-  });
+  const gapValue = (() => {
```
edison-cy-yang, line 40: this logic block can be avoided if we have the above suggestion

packages/components/src/Grid/Grid.tsx
taylorvnoj: My thinking was this is backwards compatible. I definitely agree that this is complicating things but `gap` being boolean is pre-existing, unfortunately the flexibility wasn't implemented initially. There's about 30 implementations in JO. 
I did do some looking around and some design systems have 0px spacing tokens, others do not. I feel like our spacing tokens represent intentional, systematic & semantic spacing values, but 0 represents the absence of space and something about that feels...odd.

Regardless, it's more about being backwards compatible 😅

taylorvnoj:
Have a path forward, just need the time. Since this isn't top priority it will have to wait a little while, but change is coming!

PR #2549, opened 2025-05-09
refactor(components): replace calcs to 12 with slim in web

chris-at-jobber:
## Motivations

Since we've introduced a token for `slim` to meet needs that were being worked around repeatedly, let's clean up the old workarounds.

The most common was subtracting `smaller` (4) from `base` (16), but also found some multiplications. I checked for division like `large` (24) / `2`, but did not find anything 

### Changed

- No visual changes
- Replacing calcs with the `--space-slim` token
  - Banner
  - Button
  - FormField
  - InlineLabel
  - MultiSelect
  - SegmentedControl

## Testing

Testing in Storybook should be sufficient - expect no visual diff

Changes can be

---

Random photo of Atlantis

### DIFF
--- a/packages/components/src/InlineLabel/InlineLabel.module.css
+++ b/packages/components/src/InlineLabel/InlineLabel.module.css
@@ -12,7 +12,7 @@
 }
 
 .large {
-  padding: calc(var(--space-small) * 1.25) calc(var(--space-small) * 1.5);
+  padding: calc(var(--space-small) * 1.25) var(--space-slim);
   border-radius: var(--radius-larger);
 }
 
@@ -21,7 +21,7 @@
 }
 
 .larger {
-  padding: calc(var(--space-small) * 1.5) calc(var(--space-base));
+  padding: var(--space-slim) calc(var(--space-base));
   border-radius: var(--radius-larger);
 }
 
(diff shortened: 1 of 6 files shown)


packages/components/src/InlineLabel/InlineLabel.module.css
```diff
@@ -12,7 +12,7 @@
 }
 
 .large {
-  padding: calc(var(--space-small) * 1.25) calc(var(--space-small) * 1.5);
+  padding: calc(var(--space-small) * 1.25) var(--space-slim);
```
chris-at-jobber, line 15: It is at least slightly funny that we _still_ have a calc left in here, but the 10px value is much less widespread

packages/components/src/MultiSelect/MultiSelect.module.css
```diff
@@ -48,8 +48,8 @@
 }
 
 .small {
-  --field--padding-left: calc(var(--space-base) - var(--space-smaller));
-  --field--padding-right: calc(var(--space-base) - var(--space-smaller));
+  --field--padding-left: var(--space-slim);
```
chris-at-jobber, line 51: These changes are identical to the FormField one and as a larger question, if MultiSelect is sticking around is there a way it could inherit/import styling from FormField rather than be re-declared?

PR #2551, opened 2025-05-12
feat(components): Add InputNumber v2

JeffMarvin:
## Changes
This pull request introduces a revamped `InputNumber` component (v2), featuring:
- no RHF dependency
- value is always a strict `number` type, making it easier to work with in Zod contexts
  - will either set value to a number or `Nan` when invalid
- powered by React Aria Components (Adobe)
- provides additional formatting options and more sophisitcate

The new implementation uses `react-aria-components` for accessibility and modularized styles for better maintainability
* **Shim for Backward Compatibility**: Introduced a shim in `index.tsx` to switch between the legacy and rebuilt `InputNumber` implementations based on the `version` prop (`packages/components/src/InputNumber/index.tsx`).

**Storybook Adjustment**: Enhanced the `VersionComparison` story for `InputText` to include additional props like `description` and `error` (`docs/components/InputText/Web.stories.tsx`).

### Future followups needed:
1. Simplify CSS selectors by moving some of the state logic into the component (e.g. small and disabled)
2. Implementations of min/max/step will be handled in a separate task looking at field level validations as a whole.
3. Implementations of pre and post fix content will be handled in a future upgrade.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing
The storybook examples have some additional stories:
 - there's a [version comparison page](https://job-121643-input-number-v2.atlantis.pages.dev/storybook/?path=/story/components-forms-and-inputs-inputnumber-web--version-comparison)
 - there's a [v2 formatting story](https://job-121643-input-number-v2.atlantis.pages.dev/storybook/?path=/story/components-forms-and-inputs-inputnumber-web--formatting-examples-v-2-only) to show what the new component can do

Changes can be

---

Random photo of Atlantis


packages/components/src/InputNumber/index.ts
```diff
@@ -1 +0,0 @@
-export { InputNumber, InputNumberRef } from "./InputNumber";
```
JeffMarvin, line 1: This file becomes `index.tsx` and is more sophisticated to include the version shim.

packages/components/src/InputNumber/InputNumber.rebuilt.module.css
```diff
@@ -0,0 +1,251 @@
...
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  pointer-events: none;
+  transform: translateY(var(--field--placeholder-transform));
+}
+/* When not hiding the label, animate the transition to the miniLabel */
+.inputWrapper:not(.hideLabel) .label {
+  transition: all var(--timing-quick);
+}
+/**
```
JeffMarvin, line 208: Some of the rule logic gets a bit complex here. I promised @ZakaryH I would address that in a followup. For now, I wanted to get this into review but I'll work on the followup alongside this. (It's a bit finicky to make sure all of the logic gets correctly preserved.)

packages/components/src/InputNumber/InputNumber.rebuilt.types.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import { CommonFormFieldProps, FormFieldProps } from "../FormField";
+
+export type InputNumberVersion = 1 | 2 | undefined;
+
+export interface InputNumberRebuiltProps
+  extends Omit<CommonFormFieldProps, "version">,
+    Pick<
```
edison-cy-yang, line 8: I may have missed some context but do we not want prefix and suffix in V2?

packages/components/src/InputNumber/InputNumber.rebuilt.tsx
```diff
@@ -0,0 +1,135 @@
...
+};
+
+export interface InputNumberRebuiltRef {
+  blur(): void;
+  focus(): void;
+}
+
+export const InputNumberRebuilt = forwardRef(
+  (props: InputNumberRebuiltProps, ref: React.Ref<InputNumberRebuiltRef>) => {
+    const inputRef = createRef<HTMLInputElement>();
```
edison-cy-yang, line 29: should we combine the forwarded `ref` with this?

packages/components/src/InputNumber/InputNumber.rebuilt.module.css
```diff
@@ -0,0 +1,256 @@
...
+  --field--padding-top: var(--space-small);
+  --field--padding-bottom: var(--space-small);
+  --field--height: calc(var(--space-larger) + var(--space-smaller));
+  --postfix-right: calc(var(--space-small) + var(--space-smaller));
+}
+
+.large {
+  --field--padding-left: var(--space-large);
+  --field--padding-right: var(--space-large);
+  --field--height: calc(var(--space-extravagant));
```
edison-cy-yang, line 123: I noticed size large isn't matching up to v1 and not respecting this field height. I wonder if it's due to explicitly setting aria Label vs using label from html input

packages/components/src/InputNumber/InputNumber.rebuilt.types.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import { CommonFormFieldProps, FormFieldProps } from "../FormField";
+
+export type InputNumberVersion = 1 | 2 | undefined;
+
+export interface InputNumberRebuiltProps
+  extends Omit<CommonFormFieldProps, "version">,
```
edison-cy-yang, line 7: is `clearable` going to be implemented in a followup? It'll be brought over from here but not implemented

edison-cy-yang (commented):
Mostly looks good! Just a few questions. And bringing in React Aria is 👌 🚀

I also noticed `handleChange` in V2 is only triggered when the input is blurred, I'm assuming that's just how react aria handles it internally

packages/components/src/InputNumber/InputNumber.rebuilt.module.css
JeffMarvin: Apparently this issue has existed from over a year ago. I'll file a separate bug and fix it in a followup.

packages/components/src/InputNumber/InputNumber.rebuilt.types.tsx
JeffMarvin: Good call -- I updated this section to exclude some of the props that won't be implemented yet, and added some comments.

packages/components/src/InputNumber/InputNumber.rebuilt.tsx
JeffMarvin: I'm not familiar with what we've done on some of the other rebuilt inputs. What would we want to combine the forwarded `ref` with the internal `inputRef` for here?

JeffMarvin:
> I also noticed `handleChange` in V2 is only triggered when the input is blurred, I'm assuming that's just how react aria handles it internally

@edison-cy-yang Yup! That's by design, it allows React Aria components to support input values that are invalid while they are being added (e.g. pasting in, or half-typed values) and then only parses them on blur.

PR review comments have been addressed, please re-review! Thanks.

edison-cy-yang (approved):
LGTM.

PR #2552, opened 2025-05-12
chore(components): Use version range for hooks peerDep

jdeichert:
## Motivations

This is safer for when we do major version bumps. If we were to publish hooks v3.0.0, lerna's publish commit would fail in this case because components currently requires v2.x.x.

Using `>=2` allows any version above 2, including major version 3+.

### Changed

- Updated hooks peerDep version range 

## Before

This dry run shows that lerna would fail in the publish step because components requires a peerDep of v2.x.x

Screenshot 2025-05-12 at 11 15 18 AM

## After

Now it's successful because v2 or above is allowed.

Screenshot 2025-05-12 at 11 16 37 AM

## Testing

No need to test anything.

Here's the dry run command if you really want to test it:

```bash
npx lerna version --no-push --no-git-tag-version --yes --no-private
```

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -518,7 +518,7 @@
   "peerDependencies": {
     "@apollo/client": "^3",
     "@jobber/design": "*",
-    "@jobber/hooks": "^2",
+    "@jobber/hooks": ">=2",
     "@tanstack/react-table": "^8",
     "axios": "^1",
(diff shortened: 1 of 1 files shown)


scotttjob (approved):

MichaelParadis (approved):
Thing I missed this but LGTM!

PR #2560, opened 2025-05-14
fix(components): Prevent Autocompletes within Modals from dismissing the Modal on select

jdeichert:
## Motivations

This PR fixes Autocompletes that are rendered inside composable Modals (Modal.Provider).

It prevents menu clicks from dismissing the parent Modal.

See comments for more details!

## Before

https://github.com/user-attachments/assets/19e6c7a7-5b23-45da-9f05-17a21d4e5c4a

## After

https://github.com/user-attachments/assets/aa32c876-5fb5-495b-a822-ecc2e36b6b59

### Fixed

- Provided a workaround to prevent Autocompletes from dismissing parent Modals 

## Testing

To test this...

1. Go to this story which showcases the problem _before_ the fix: https://89e3cd3f.atlantis.pages.dev/storybook/?path=/story/components-forms-and-inputs-autocomplete-web--within-modal
2. Click the button to open the modal
3. Focus the first Autocomplete, the menu will appear
    * Select an option, note that it DOES NOT dismiss the modal (this is good ✅)
    * This default menu works fine, the problem exists with the following Autocompletes that use `customRenderMenu`
4. Focus the second Autocomplete, the menu will appear
    * Select an option, observe the entire Modal gets dismissed (bad ❌)
    * Repeat with the other Autocompletes below
    * Observe how clicking "+ Add new client" also leads to the Modal being dismissed (bad ❌)
5. Now go to this story which showcases the solution working as intended: https://job-124341-fix-autocomplete.atlantis.pages.dev/storybook/?path=/story/components-forms-and-inputs-autocomplete-web--within-modal
    * Repeat steps 2-4
    * Observe how selecting any Autocomplete option (or custom buttons) now works as expected and DOES NOT dismiss the parent Modal

---

### DIFF
--- a/packages/components/src/Autocomplete/Menu/Menu.tsx
+++ b/packages/components/src/Autocomplete/Menu/Menu.tsx
@@ -39,7 +39,7 @@ export function Menu<
    * outside of the Modal.
    */
   const specialModalWorkaround = isWithinOpenModal && customRenderMenu;
-  if (specialModalWorkaround && (!inputFocused || !options.length)) return null;
+  if (specialModalWorkaround && !inputFocused) return null;
 
   if (customRenderMenu) {
     return (


docs/components/Autocomplete/Web.stories.tsx
```diff
@@ -1098,3 +1099,448 @@ function useCustomSectionOptions(
...
+                  <Icon name="checkmark" size="small" />
+                )}
+                <Text>{option.label}</Text>
+                <Text variation="subdued">{option.address}</Text>
+                <Text variation="subdued">{option.contact}</Text>
+              </Content>
+            </Flex>
+            <StatusLabel status={status} label={label} />
+          </Flex>
+        </BaseMenuOption>
```
ZakaryH, line 1460: not a huge deal but something in the example is off with the layout causing it to exceed the width

image

packages/components/src/Autocomplete/Menu/Menu.tsx
```diff
@@ -21,6 +22,25 @@ export function Menu<
   inputRef,
   customRenderMenu,
 }: MenuProps<GenericOption, GenericOptionValue>) {
+  const { open } = useModalContext();
```
ZakaryH, line 25: somewhat of a nitpick, `open` does make sense as a concept within `Autocomplete` and `Autocomplete/Menu`

but this `open` isn't referring to either of those, so maybe we alias it as `parentModalOpen` or `modalOpen`?

pretty minor thing though, I don't feel that strongly about it

packages/components/src/Autocomplete/Menu/Menu.tsx
```diff
@@ -21,6 +22,25 @@ export function Menu<
...
+   * replace any `onClick` handlers with `onMouseDown` handlers on those Buttons.
+   *
+   * The check below prevents Autocomplete from rendering the `customRenderMenu` when it's not visible
+   * (when the input isn't focused). This prevents Modals (FloatingUI) from marking the Autocomplete's
+   * menu as data-floating-ui-inert and aria-hidden="true". As a result, this prevents the bug where clicking
+   * within its menu would cause FloatingUI to close the parent Modal because it determined the click was
+   * outside of the Modal.
+   */
+  const specialModalWorkaround = open && customRenderMenu;
+  if (specialModalWorkaround && (!inputFocused || !options.length)) return null;
```
ZakaryH, line 42: we're going to have an issue here with `options.length`

you could have a search term of `something that doesn't match an option` so you'll have `options.length` of 0

but lets say one of your menu footer items is `Add your search term as a new option`- we wouldn't be rendering it 

image

image

ZakaryH (approved):
definitely want us to figure out how to make this work better in the future, but since it's an internal workaround not exposed to consumers in any real way it's perfectly fine as a short term fix

code changes make sense, and playing around with it in the web stories it's behaving as expected now

I haven't tried it in product yet, gonna take a look at that PR in a moment but I wouldn't imagine that would affect my thoughts on this PR - it would most likely be something we need to adjust over there

LGTM pending passing CI run

PR #2562, opened 2025-05-14
fix(components): Remove z-index from affix icons with actions [JOB-119173 ]

Aiden-Brine:
## Motivations

[This bug](https://jobber.atlassian.net/browse/JOB-119173) was reported where the calendar from the date picker is overlapping with the save bar. 

image-20250319-185030

Looking into it I can't determine the need for this z-index. I asked [here](https://getjobber.slack.com/archives/C03BKPST4KC/p1747244072043699) and Chris who was a reviewer on the original PR thinks it was to make sure that the icons are always above the input content (value/placeholder). I've done some testing though and it doesn't seem necessary for that. It's worth noting that only suffixes can have actions, so we only need to test this behaviour for suffixes and not prefixes.

### Removed

The z-index for affix icons with actions has been removed.

## Testing

To do some initial testing I added the following to one of the stories for form field:
```
  suffix: {
    icon: "googlePlay",
    onClick: () => console.log("clicked suffix"),
    ariaLabel: "Suffix button",
  },
```
To ensure that at various screen sizes the placeholder text/input value don't overlap with the suffix button. I have also created a pre-release for testing the behaviour on JO and JFE on various inputs. I recognize that my solution is based on not seeing issues with the style removed so it is important that we test this thoroughly.

Screenshot 2025-05-14 at 1 11 11 PM

Screenshot 2025-05-14 at 1 11 25 PM

Changes can be

---

### DIFF
--- a/packages/components/src/FormField/FormField.module.css
+++ b/packages/components/src/FormField/FormField.module.css
@@ -352,11 +352,6 @@
   padding-left: var(--space-small);
 }
 
-.affixIcon.hasAction {
-  position: relative;
-  z-index: var(--field--postfix-action-elevation);
-}
-
 .affixLabel {
   display: flex;
   margin: 0 calc((var(--field--padding-left) - var(--space-smallest)) * -1) 0 0;
--- a/packages/components/src/FormField/FormField.module.css.d.ts
+++ b/packages/components/src/FormField/FormField.module.css.d.ts
@@ -25,7 +25,6 @@ declare const styles: {
   readonly "postfix": string;
   readonly "affixIcon": string;
   readonly "suffix": string;
-  readonly "hasAction": string;
   readonly "affixLabel": string;
   readonly "description": string;
   readonly "toolbar": string;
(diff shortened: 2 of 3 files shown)


MichaelParadis (approved):
LGTM! This line was never doing anything after checking the initial PR that added it

JeffMarvin:
Ok, Michael did more sleuthing so maybe this isn't used... I was wondering if we should be safer and just use an isolation context. 
The only other thing I'm wondering is if the css variable that is getting its reference removed here still needs to exist or if it can also be removed.

JeffMarvin:
@MichaelParadis @Aiden-Brine

Aiden-Brine:
@JeffMarvin good catch on the variable. Since this wasn't used in the first place I don't think it is worth complicating things by adding the isolation context. I'll get this into JFE and then I'll put up a one liner cleanup PR to remove that variable and tag you and @MichaelParadis

PR #2563, opened 2025-05-14
fix(components): Allowing error passthrough on <Select version={2} />

scotttjob:
## Motivations

We want to allow the error to be passed through to the `SelectRebuilt` component.

## Changes

1. The error is now passed through to the SelectRebuilt component.

## Testing

You should be able to add an 'error' prop to the `SelectRebuilt` component and see the error displayed

"Version comparison" story's Toggle Error verifies `error` prop is being passed through

Changes can be

---


edison-cy-yang (approved):
LGTM! Thanks for jumping on this

PR #2567, opened 2025-05-20
feat(hooks): Adding useStepper hook [JOB-124608]

kshelley-jobber:
## Motivations
There are a number of places where we have multi-step workflows in JO and JFE, with differing implementations. The aim is to provide a standard implementation for multi-step workflows which can be shared among repositories.

This hook has been vetted in JFE, having been used to build the multi-step setup guide, and will also be used in certain multi-step workflows in JO, such as the self-serve cancellation/hibernation workflow, and upcoming self-serve downgrade process

### Added

- Adds `useStepper` hook based on the [existing implementation](https://github.com/GetJobber/jobber-frontend/blob/d8277dbb9971ab2c320ea352e59791aac89b1cc8/apps/jobber-online/app/shared/stepper/useStepper.tsx) in JFE

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing
A basic example of the stepper hook has been added to storybook and can be used to test the usage of navigating forward/back in a stepper workflow. The checks for first/last step are also being used in the example to enable/disable stepper buttons

Changes can be

---

### DIFF
--- a/docs/hooks/useStepper.stories.mdx
+++ b/docs/hooks/useStepper.stories.mdx
@@ -14,6 +14,8 @@ type safety for step definitions.
 </Canvas>
 
 ```tsx
+import { useStepper } from "@jobber/hooks/useStepper";
+
 const steps = ["step1", "step2", "step3"] as const;
 
 const {


packages/hooks/src/useStepper/useStepper.ts
```diff
@@ -0,0 +1,59 @@
+import { useMemo, useState } from "react";
+import invariant from "tiny-invariant";
```
ethanmillard, line 2: I'm not sure if atlantis has a direct dependancy of `tiny-invariant`, I think it's just a dependancy from a bunch of other packages.

We should probably add the dependancy ourselves into the package.json

kshelley-jobber: You're right, its not included explicitly - any opinions from @GetJobber/ux-foundations about adding this dependency? seems like the impact on bundle size should be pretty inconsequential

JeffMarvin: Thanks for adding it explicitly.

docs/hooks/useStepper.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+  defaultStep?: StepName;
+}
+
+function useStepper<StepName extends string>(
+  steps: readonly StepName[],
+  options: UseStepOptions<StepName> = {}
+);
+```
+
+## Return Values
```
JeffMarvin, line 30:
```suggestion
## Return Value
```
Consider using the singular here, since it is just an object being returned.

docs/hooks/useStepper.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+import { UseStepper } from "./UseStepper";
+
+<Meta title="Hooks/useStepper" />
+
+# UseStepper
+
+The `useStepper` hook provides a type-safe way to manage multi-step flows in
+React applications. It handles step progression and navigation while ensuring
+type safety for step definitions.
+
```
JeffMarvin, line 11: Suggest adding a code snippet for how to import. (We're not super consistent on this but it is helpful.) Github messes with the formatting here, but I think you get the idea:
```suggestion
```typescript
import { useStepper } from "@jobber/hooks/useStepper";
````
```

docs/hooks/useStepper.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+<Canvas>
+  <UseStepper />
+</Canvas>
```
JeffMarvin, line 14: The example is great. Consider adding another block here to make the usage example easy to access on this page. Something like: 


```suggestion
## Example
<Canvas>
  <UseStepper />
</Canvas>

```tsx
const steps = ["step1", "step2", "step3"] as const;

const { currentStep, nextStep, previousStep, isFirst, isLast } = useStepper(
  steps,
  {
    defaultStep: "step1", // optional
  },
);

return (
  <div>
    {currentStep === "step1" && <div>Step 1</div>}
    {currentStep === "step2" && <div>Step 2</div>}
    {currentStep === "step3" && <div>Step 3</div>}
    <Flex gap="base" direction="row" align="end" template={["grow", "grow"]}>
      <Button
        label="Previous"
        onClick={previousStep}
        disabled={isFirst}
        type="primary"
      />
      <Button
        label="Next"
        onClick={nextStep}
        disabled={isLast}
        type="primary"
      />
    </Flex>
  </div>
);
```

```

File: packages/hooks/src/useStepper/useStepper.test.tsx [hunk 5]
```diff
@@ -0,0 +1,171 @@
...
+
+  describe("error handling", () => {
+    beforeEach(() => {
+      jest.spyOn(console, "error").mockImplementation(() => undefined);
+    });
+
+    afterEach(() => {
+      jest.restoreAllMocks();
+    });
+
```
JeffMarvin, line 114: Is this necessary? It can be problematic to mock like this, and I'm not sure if we need to here.

packages/hooks/src/useStepper/useStepper.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+    it("throws error when steps array is empty", () => {
+      expect(() => {
+        renderHook(() => useStepper([]));
+      }).toThrow("Steps array cannot be empty");
+    });
+  });
```
JeffMarvin, line 120: Consider being a bit more precise with the type here too:
```
expect(() => {
  // code that should throw
}).toThrow(new Error("expected error message"));
```

packages/hooks/src/useStepper/useStepper.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+});
+
+describe("useStepper implementation", () => {
```
JeffMarvin, line 123: These tests are good, but do they add anything that the above approach isn't covering? 
Consider eliminating one of the two approaches here unless they are both required.

docs/hooks/useStepper.stories.mdx
JeffMarvin: Also consider including `goToStep` in your example here, as it is a very useful feature.

packages/hooks/src/useStepper/useStepper.ts
```diff
@@ -0,0 +1,59 @@
...
+    [currentStep, steps],
+  );
+
+  const handlers = useMemo(
+    () => ({
+      goToStep: (step: StepName) => {
+        setCurrentStep(step);
+      },
+
+      nextStep: () => {
```
JeffMarvin, line 33: I find `nextStep` and `previousStep` ambiguous - they seem to be nouns, like `currentStep` since they omit their verb. 

Consider making these clearer with a noun. One example, consistent with `goToStep` could be `goToNextStep` and `goToPreviousStep`

I recognize that that will mean existing usages of this hook will need to be migrated to the new functions but I think it is worth it for clarity in the long run.

JeffMarvin (changes requested):
Thanks for the excellent contribution! I had some questions and suggestions, let me know if you want to chat through anything.

packages/hooks/src/useStepper/useStepper.test.tsx
kshelley-jobber: Good point, its probably a bit redundant to have both 🤔

docs/hooks/useStepper.stories.mdx
kshelley-jobber: Good call, I've added another button here that is rendered when we're at the last step, and returns to the beginning of the stepper to demonstrate that functionality

packages/hooks/src/useStepper/useStepper.test.tsx
kshelley-jobber: It doesn't look like it is necessary!

packages/hooks/src/useStepper/useStepper.ts
kshelley-jobber: I've renamed these properties as you suggested, there are only a couple of existing usages and it should be a fairly easy refactor when we start using the Atlantis component

PR #2568, opened 2025-05-21
fix(components): Re-render Autocomplete when initialOptions change

timoteialbu:
## Motivations

We initialize some state internally when initialOptions are passed in. I would like those options to be update-able for this use case:
- I am using Autocomplete to show a list of values I get from a graphql query. I am using the CustomMenuRender prop to render a button that allows me to create another item in the list. At the end of that creation I want to add that item back to the initialOptions, but that is currently not possible because of the way the internal options are saved.

### Added

- 

### Changed

- the `options` useState will now be updated internally based on changes to `initialOptions`.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

### DIFF
--- a/packages/components/src/Autocomplete/Autocomplete.tsx
+++ b/packages/components/src/Autocomplete/Autocomplete.tsx
@@ -56,6 +56,10 @@ function AutocompleteInternal<
   const inputRef = useRef<InputTextRef | null>(null);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


timoteialbu:
Closing because we have a workaround + autocomplete is scheduled to be rewritten so that we can control its state easier.

PR #2578, opened 2025-05-30
fix(components): Fix modal spacing issue due to css specificity recently changing

jdeichert:
## Motivations

This PR fixes 2 classes that were trying to override other component's styles.

My recent [PR](https://github.com/GetJobber/atlantis/pull/2573) fixed our `styles.css` bundle to have a consistent classname order. This change led to this problem because class order changed compared to what we had previously, which was non-deterministic/random and we've just been really lucky this never appeared before.

My change is still important, we absolutely need css consistency. This spacing issue could have appeared at any point before my change due to the random bundling order of our css before.

## Before

Screenshot 2025-05-30 at 8 26 07 AM

## After

This matches a preview build from _before_ my change landed: https://3ce4c73a.atlantis.pages.dev/components/ConfirmationModal

Screenshot 2025-05-30 at 8 26 01 AM

### Fixed

- Modal button and content gaps

## Testing

* Pull this branch down
* npm install
* cd packages/site
* npm run dev
* go here: http://localhost:5173/components/ConfirmationModal
    * cmd-shift-r to force a reload (sometimes editorBundle.js is cached)
* open the modal
    * observe the primary/secondary actions have a gap
    * observe the text content under the title DOES NOT have top padding

### DIFF
--- a/packages/components/src/Modal/Modal.rebuilt.module.css
+++ b/packages/components/src/Modal/Modal.rebuilt.module.css
@@ -73,7 +73,7 @@
  */
 
 .header + * {
-  padding-top: 0;
+  padding-top: 0 !important;
 }
 
 .closeButton {
@@ -107,7 +107,7 @@
 /* This is in a correct position and order */
 /* stylelint-disable-next-line no-descending-specificity */
 .rightAction > * {
-  margin-left: var(--space-small);
+  margin-left: var(--space-small) !important;
 }
 
 .rightAction > *:first-child {


nad182 (changes requested):
I believe we need to make the similar changes in `Modal.rebuilt.module.css`.

nad182 (approved):
So cool to see visual snapshots!

packages/components/src/Modal/Modal.module.css
```diff
@@ -109,7 +109,7 @@
 /* This is in a correct position and order */
 /* stylelint-disable-next-line no-descending-specificity */
 .rightAction > * {
-  margin-left: var(--space-small);
+  margin-left: var(--space-small) !important;
```
chris-at-jobber, line 112: @jdeichert I'm glad we landed a fix, but is there a version of this where we actually rebuild the modal actions using our own layout components to avoid needing custom CSS in the Modal?

PR #2579, opened 2025-05-30
feat(components): add ref to Checkbox v2

ericchernuka:
Adds the forwarding of a ref to v2 to enable programatic control

## Motivations
With the move to v2, we lost the ability to programmatically focus/interact with the input directly.

## Changes

1. Adds a `ref` of `HTMLInputElement` to be forwarded through the shim

## Testing
1. Adding a ref should not have a type error by accurately exposing an HTMLInputElement.
2. Should be able to set `.focus()` the ref.

Changes can be

---

### DIFF
--- a/packages/components/src/Checkbox/Checkbox.rebuilt.tsx
+++ b/packages/components/src/Checkbox/Checkbox.rebuilt.tsx
@@ -1,24 +1,27 @@
-import React, { ChangeEvent } from "react";
+import React, { ChangeEvent, forwardRef } from "react";
 import classnames from "classnames";
 import styles from "./Checkbox.module.css";
 import { CheckboxRebuiltProps } from "./Checkbox.types";
 import { Icon } from "../Icon";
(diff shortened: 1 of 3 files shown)


edison-cy-yang (commented):
If the purpose is to programmatically focus, then you might want to add `.input:focus + .checkBox,` to the current css for focus, as it only has the `:focus-visible` pseudo class for now

edison-cy-yang (approved):
Discussed and confirmed focusing programmatically from RHF is treated as indirect user interaction and the `:focus-visible` pseudo class is applied correctly.

ericchernuka:
@edison-cy-yang wondering if we need an invalid border for our use case. I can add that. Won't have the message yet, but I think that's ok in this case.

JeffMarvin:
for the record, ex-head-of-atlantis-design, now general patron @chris-at-jobber approved the design: 
image

PR #2580, opened 2025-06-02
refactor: decouple shadow values from spacing values

chris-at-jobber:
## Motivations

[As a pattern](https://atlantis.getjobber.com/guides/frontend-styleguide#component-view-css), we want to avoid coupling spacing tokens to things that are not inherently spacing (such as size).

It stands to reason that shadows are also impacted by this logic. If padding/margin/gap increases or decreases, consumers and users should not inherently expect that shadows will increase or decrease.

image

### Changed

- No visual changes
- Shadow tokens internally reference px values (not spacing tokens) for blur, offset, spread, etc

image

## Testing

Test locally, run visual diffs, test in pre-release if desired

**LOCAL TESTING** ensure you run `npm run bootstrap` or the changes in the design package will not be picked up locally

Changes can be

---


chris-at-jobber:
Reverting the test fix because @JeffMarvin is solving in #2582

AutumnDaweBaillie (approved):

PR #2584, opened 2025-06-06
fix(components): add radius to focus states of FormatFile

chris-at-jobber:
## Motivations

When focused, FormatFile's focus indicator did not respect the rounded shape of the element.

### Fixed

- Radius of focus state follows the radius of the element being focused
image

## Testing

Locally, Cloudflare, in-product via pre-release if needed

- Notes and Quotes can be testing surfaces in-product

Changes can be

---

### DIFF
--- a/packages/components/src/FormatFile/FormatFile.module.css
+++ b/packages/components/src/FormatFile/FormatFile.module.css
@@ -58,6 +58,7 @@
 
 .compact .wrapper:focus {
   box-shadow: var(--shadow-focus);
+  border-radius: var(--radius-base);
   outline: none;
 }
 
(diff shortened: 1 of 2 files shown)


chris-at-jobber:
That's a good catch! It would make sense to fix in the same vein, but since it's not introduced in this PR (screenshot from current) I'd like to land this fix if that's OK? There are a bit more design considerations to make around that state to fix this issue (the small thumbnail also doesn't look great as the delete button takes up more than 1/4 of the space 😭 ) so I'd rather solve the "delete button design issues" separate from the focus state radius bug.

image
image

chris-at-jobber:
Chatted "Offline™" and going to land this one as-is

chris-at-jobber:
@nad182 working on the delete button cleanup here https://github.com/GetJobber/atlantis/pull/2593

PR #2590, opened 2025-06-19
fix(components-native): Cleanup form on submit success v2

davidr64:
## Motivations
We have had to make a few PRs into jobber-mobile to work around onSubmit being called before the form is fully submitted. 

https://github.com/GetJobber/jobber-mobile/pull/10437
https://github.com/GetJobber/jobber-mobile/pull/10214
https://github.com/GetJobber/jobber-mobile/pull/9974

Tickets resolved by above PRs. All talking about the cache data sticking around
https://jobber.atlassian.net/browse/JOB-126732
https://jobber.atlassian.net/browse/OPS-23103
https://jobber.atlassian.net/browse/OPS-22226

Outstanding ticket
Extra issues are caused when deleting line items if they are dirty
https://jobber.atlassian.net/browse/JOB-125243

We are calling onSubmitSuccess before react-hook-form thinks the form is submitted.

The function passed to handleSubmit(react-hook-form) returns a promise. If that promise succeeds then the the form is considered successfully submitted. We have been including onSubmitSuccess in that inner promise, meaning it is being called as part of the submission which also means if it fails then the form will tell the user that the save didn't happen. However, the save did happen because internalSubmit is already called.

I feel like this would be a more robust longer term fix to move `onSubmitSuccess` after the form is submitted.

### Added

- 

### Changed

Moved calling of the onSaveSuccess parameter to after the react-hook-form handle submit promise resolves. Meaning the form is considered submitted when the function is called.

## Testing

Undo the specific patch fixes in jobber-mobile in the PRs above and install the pre-release. The navigation and cache issues aren't reintroduced. 
eg:
Create a quote then create a second quote immediately after. You shouldn't see any cached data pre-populate in the form from the previous quote.

Dirty a line item before deleting it and you won't get confirmation before back.

Changes can be

---

### DIFF
--- a/packages/components-native/src/Form/Form.test.tsx
+++ b/packages/components-native/src/Form/Form.test.tsx
@@ -2,8 +2,10 @@ import React from "react";
 import { act, fireEvent, render, waitFor } from "@testing-library/react-native";
 import { Alert, Keyboard } from "react-native";
 import { Host } from "react-native-portalize";
+import { FieldValues, SubmitHandler } from "react-hook-form";
 import { Form, FormBannerMessage, FormBannerMessageType } from ".";
 import { FormBannerErrors, FormSubmitErrorType } from "./types";
+import * as useInternalFormModule from "./hooks/useInternalForm";
 import { atlantisContextDefaultValues } from "../AtlantisContext";
 import * as atlantisContext from "../AtlantisContext/AtlantisContext";
 import { Text } from "../Text";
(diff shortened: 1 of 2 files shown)


packages/components-native/src/Form/Form.test.tsx
```diff
@@ -300,6 +300,32 @@ describe("Form", () => {
...
+
+      fireEvent(getByLabelText(switchLabel), "onValueChange", true);
+      expect(onChangeSwitchMock).toHaveBeenCalled();
+
+      fireEvent.press(saveButton);
+
+      await waitFor(() => {
+        expect(onSubmitMock).toHaveBeenCalled();
+      });
+      expect(onSuccessMock).toHaveBeenCalledWith({
```
davidr64, line 324: I still want to look into a way I can spy on react-hook-form useForm to ensure that handleSubmit promise resolve calls onsuccessmock?

packages/components-native/src/Form/Form.tsx
```diff
@@ -247,6 +247,18 @@ function InternalForm<T extends FieldValues, S>({
...
 
+  async function callHandleSubmit() {
+    let result: S | undefined;
+
+    await handleSubmit(async data => {
+      const saveResult = await internalSubmit(data);
+      result = saveResult as S;
+    })();
+
+    removeListenerRef.current?.();
```
davidr64, line 258: Maybe move this into the above arrow function

PR #2591, opened 2025-06-19
fix(components): add filename sanitization to InputFile component

Jimmy-b36:
## Motivations

File uploads with ASCII characters outside the range 32-126 in filenames can cause issues when handing filenames to our backend.

### Added

- `scrubFileName` function that removes ASCII characters outside the range of 32-126 (space to tilde) from filenames
- Test coverage for filename scrubbing functionality

### Changed

- `InputFile` component now sanitizes filenames before creating `FileUpload` objects

## Testing

- Added test: "scrubs ASCII characters outside of the range of 32-126 from filename" - verifies that non-ASCII characters (emojis, accented characters) are removed
- Added test: "does not scrub ASCII characters within the range of 32-126 from filename" - verifies that valid ASCII characters are preserved
- Example: `"Screenshot 2025-06-18 at 10.24.19 AM🐦‍🔥.png"` becomes `"Screenshot 2025-06-18 at 10.24.19AM.png"`

---

### DIFF
--- a/packages/components/src/InputFile/InputFile.tsx
+++ b/packages/components/src/InputFile/InputFile.tsx
@@ -441,7 +441,13 @@ export function InputFile({
       httpMethod = "POST",
     } = params;
 
-    const fileUpload = getFileUpload(file, key, url, transformFilename);
+    const fileUpload = getFileUpload(
+      file,
+      key,
+      url,
+      transformFilename,
+      onUploadError,
+    );
     onUploadStart && onUploadStart({ ...fileUpload });
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -511,6 +517,7 @@ function getFileUpload(
   key: string,
   uploadUrl?: string,
   transformFilename?: (filename: string) => string,
+  onUploadError?: (error: Error) => void,
 ): FileUpload {
   let fileName = file.name;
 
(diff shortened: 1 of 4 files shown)


JeffMarvin (commented):
@nad182 A few questions about your approach... We're being pretty narrow in the function purpose and saying it is to "sanitize" but it's technically just manipulating the characters. The function could also be used to map certain characters over to others, or even do something like add a random UUID to the end. So, is there maybe room to consider the function name and approach more generally? 

Also, we should test and have some kind of fallback behavior if the function ends up returning an empty string. Or, in other words, what happens when there's no file name?

packages/components/src/InputFile/InputFile.tsx
```diff
@@ -469,10 +510,30 @@ function getFileUpload(
...
+  let fileName = file.name;
+
+  if (transformFilename) {
+    try {
+      const transformed = transformFilename(file.name);
+      // Use transformed name if it's not empty, otherwise fall back to original
+      fileName = transformed && transformed.trim() ? transformed : file.name;
+    } catch (error) {
+      // If transformation fails, silently fall back to original filename
+      fileName = file.name;
```
JeffMarvin, line 524: @nad182 I don't know the right answer off the top of my head, but silently falling back to the original could cause a lot of problems for our users. Can we think of a better alternative?

JeffMarvin: Even, resorting to the default below is better than silently returning to the original one.

nad182: We can sent some error back to the consumer for them to decide what to do with it. I pushed another commit addressing this: https://github.com/GetJobber/atlantis/pull/2591/commits/c77d5e82bdbd4d20a96721796aee660c44e5c2f9

JeffMarvin: I don't love that approach... Let me look at how this component and the props we're proposing would actually get used in reality. Want to think through what the best realistic user experience would be for this.

Jimmy-b36:
Hey Team, thank you for reviewing this. Our team has decided to move the solution to the backend, so I plan to close this PR unless you believe the solution would also be valuable on the front end.

nad182:
Will be "taken care of" on the backend (potentially in https://github.com/GetJobber/Jobber/pull/61173)

PR #2592, opened 2025-06-20
feat(components): Adding aria-busy to Button when in a loading state.

scotttjob:
## Motivations

1. Got a request to be able to validate the loading state on our button.

## Changes

1. Added aria-busy to Button when in a loading state.

### Added

- Added aria-busy to Button when in a loading state.

### Changed

- Button

### Fixed

- Button not being in a proper accessible loading state.

## Testing

- Button element should now have 'aria-busy' applied when Button has loading={true} as a prop.

Changes can be

---


nad182 (approved):

packages/components/src/Button/Button.tsx
```diff
@@ -61,6 +62,7 @@ function ButtonWrapper(props: ButtonProps) {
     ...(url === undefined && to === undefined && { type: buttonType }),
     "aria-controls": ariaControls,
     "aria-haspopup": ariaHaspopup,
+    "aria-busy": loading,
```
nad182, line 65: I approved this to unblock other work, but I'm wondering whether it makes sense to add tests for this new prop (and attribute).

PR #2593, opened 2025-06-20
fix(components): Clean up size and placement of FormatFile delete button

chris-at-jobber:
## Motivations

The delete button on a thumbnail in FormatFile is
- too big on the base size (covers way too much surface area) and arguably doesn't need to be as large as it is in the large size thumbnail either
- too close to the edge on the large size (focus state has a just-off overlap with the border)
- the radius of the button looks awkward against the corner of the thumbnail
- delete button was not visible on larger touch devices like an iPad

Base
image
image
Focus state does overlap but I think it's an acceptable trade-off for the reduction in image covered

Large
image
image

Touch emulator with a large viewport
image

Mouse with a small viewport
image

### Changed

- Delete button is always `small`
- On base size thumbnail, custom styling is applied to reduce size further
- No changes to Expanded state
- Added a shadow to give button slightly more distinction from background in always-visible (small-screen) states
- Query to show delete button is based on cursor, not viewport so that larger touch devices see the delete button

### Fixed

- focus state on large size of thumbnail no longer overlaps
- delete button on base size of thumbnail no longer covers >25% of surface area

## Testing

Can test locally/CF/in Notes or Quotes in Jobber

Changes can be

---

### DIFF
--- a/packages/components/src/FormatFile/FormatFile.module.css
+++ b/packages/components/src/FormatFile/FormatFile.module.css
@@ -83,6 +83,12 @@
   z-index: var(--elevation-base);
 }
 
+.compact .customDeleteButton {
+  max-width: 24px;
+  min-height: 24px;
+  border-radius: var(--radius-small);
+}
+
 .expanded .thumbnail {
   border-right: none;
   border-bottom-left-radius: var(--radius-base);
@@ -137,6 +143,7 @@
   position: absolute;
   top: var(--space-smaller);
   right: var(--space-smaller);
+  box-shadow: var(--shadow-base);
 }
 
 @media (--medium-screens-and-up) {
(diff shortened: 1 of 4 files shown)


chris-at-jobber:
Strictly my own opinion, I think it's an OK trade-off - in the previous version the radii of the two rectangles were inconsistent, and we've eliminated it on the large thumbnail here. On the base size, I explored 
- nudging the button further in, but that negates the benefits of the size reduction
- moving the button into a negative offset (ie -2 top, -2 right) but it felt a bit brittle in that we don't necessarily control where/how FormatFile is currently being used

(Previous overlap for reference)
image

AutumnDaweBaillie (approved):
Nice! Compared to what we have today, I think these updates are an improvement. Especially for that base smaller size, I get what you mean now by the button covering so much of it! haha. 

I'd agree with @chris-at-jobber that the bit of border overlap in the focused base case is a fair trade off, I'm good with it!

nad182 (approved):
Approving, since it was approved from on the design side!

PR #2594, opened 2025-06-20
fix(components): DataList Header Fixes

ZakaryH:
## Motivations

we had 2 somewhat related issues with DataList headers, both happening in the situation where your DataList has an `onSelect` and `selected` props implemented but NO `onSelectAll`

**the first issue**

because we aren't rendering the checkbox to select all, everything shifts to the left - causing misalignment in all your columns with the header. this is very hard, or awkward to solve at the consumer level because you don't have access to the checkbox. the best you can do is some not great workaround checking for `!item.id` in the `DataList.Layout` to give the header the exact amount of spacing as the Checkbox would occupy

image

**second issue**
on mobile/small screens, the headers would show up if you have a selection
image

this isn't desirable because the headers don't make any sense due to the content no longer being in columns that line up with the headers. also if you click on a header that has a dropdown it's just borked.
image

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

no longer showing headers on single-select + mobile. the logic wasn't account for this case, updated it.

the checkbox is now always being rendered as long as we have either `onSelect` or `onSelectAll`. admittedly it's a bit weird because in the `onSelect` case it is never used nor visible, but it'll provide the correct spacing. open to different ideas on how to approach this.

### Security

- 

## Testing

we should test a handful of cases around the header's visibility, wether that's the standard header or the select-all version of the header

- no onSelect nor onSelect all
- onSelect with no onSelectAll
- onSelect and no onSelectAll

then in each one, try out the "normal" screen sizes, mobile screen size, and manually passed in HeaderVisibility config

ALSO when trying out the select versions, make sure we have the select all UI when we expect to (1+ selections, onSelectAll present), and that it's not there for `onSelect` only implementations. 

alignment of the headers and columns should be correct for all cases too

Changes can be

---

### DIFF
--- a/packages/components/src/DataList/components/DataListHeader/DataListHeader.test.tsx
+++ b/packages/components/src/DataList/components/DataListHeader/DataListHeader.test.tsx
@@ -1,0 +1,207 @@
+import React from "react";
+import { render, screen } from "@testing-library/react";
+import { DataListHeader } from "./DataListHeader";
+import { useDataListContext } from "../../context/DataListContext";
+import { useBatchSelect } from "../../hooks/useBatchSelect";
+import { useActiveLayout } from "../../hooks/useActiveLayout";
+import { useResponsiveSizing } from "../../hooks/useResponsiveSizing";
+
+// Mock the hooks
+jest.mock("../../context/DataListContext", () => ({
+  useDataListContext: jest.fn(),
+}));
+
+jest.mock("../../hooks/useBatchSelect", () => ({
+  useBatchSelect: jest.fn(),
+}));
+
+jest.mock("../../hooks/useActiveLayout", () => ({
+  useActiveLayout: jest.fn(),
+}));
+
+jest.mock("../../hooks/useResponsiveSizing", () => ({
+  useResponsiveSizing: jest.fn(),
+}));
+
+describe("DataListHeader", () => {
+  const mockUseDataListContext = useDataListContext as jest.Mock;
+  const mockUseBatchSelect = useBatchSelect as jest.Mock;
+  const mockUseActiveLayout = useActiveLayout as jest.Mock;
+  const mockUseResponsiveSizing = useResponsiveSizing as jest.Mock;
+
+  const mockLayout = jest.fn(data => (
+    <div data-testid="mock-header">{data.label}</div>
+  ));
+  const mockHeaders = { label: "Test Header" };
+
+  beforeEach(() => {
+    // Default mock settings
+    mockUseResponsiveSizing.mockReturnValue({ xs: true });
+
+    mockUseDataListContext.mockReturnValue({
+      headerVisibility: { xs: true, sm: true, md: true, lg: true, xl: true },
+      headers: mockHeaders,
+      layoutBreakpoints: ["xs"],
+      data: [{ id: 1 }, { id: 2 }],
+      selected: [],
+    });
+
+    mockUseBatchSelect.mockReturnValue({
+      hasAtLeastOneSelected: false,
+      canSelect: false,
+      canSelectAll: false,
+    });
+
+    mockUseActiveLayout.mockReturnValue({
+      layout: mockLayout,
+    });
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe("when headerVisibility is true", () => {
+    it("should render the header when layout is defined", () => {
+      render(<DataListHeader />);
+      expect(screen.getByTestId("mock-header")).toBeInTheDocument();
+    });
+
+    it("should render header when layout is defined and no items are selected", () => {
+      mockUseBatchSelect.mockReturnValue({
+        hasAtLeastOneSelected: false,
+        canSelect: true,
+        canSelectAll: true,
+      });
+
+      render(<DataListHeader />);
+      expect(screen.getByTestId("mock-header")).toBeInTheDocument();
+    });
+
+    it("should render header when no items are selected and selection is disabled", () => {
+      mockUseBatchSelect.mockReturnValue({
+        hasAtLeastOneSelected: false,
+        canSelect: false,
+        canSelectAll: false,
+      });
+
+      render(<DataListHeader />);
+      expect(screen.getByTestId("mock-header")).toBeInTheDocument();
+    });
+
+    it("should not render when layout is undefined", () => {
+      mockUseActiveLayout.mockReturnValue({ layout: undefined });
+
+      render(<DataListHeader />);
+      expect(screen.queryByTestId("mock-header")).not.toBeInTheDocument();
+    });
+  });
+
+  describe("when headerVisibility is false", () => {
+    beforeEach(() => {
+      mockUseDataListContext.mockReturnValue({
+        headerVisibility: {
+          xs: false,
+          sm: false,
+          md: false,
+          lg: false,
+          xl: false,
+        },
+        headers: mockHeaders,
+        layoutBreakpoints: ["xs"],
+      });
+    });
+
+    it("should not render when no items are selected and selection is disabled", () => {
+      mockUseBatchSelect.mockReturnValue({
+        hasAtLeastOneSelected: false,
+        canSelect: false,
+        canSelectAll: false,
+      });
+
+      render(<DataListHeader />);
+      expect(screen.queryByTestId("mock-header")).not.toBeInTheDocument();
+    });
+
+    it("should not render when no items are selected, selection is enabled but canSelectAll is false", () => {
+      mockUseBatchSelect.mockReturnValue({
+        hasAtLeastOneSelected: false,
+        canSelect: true,
+        canSelectAll: false,
+      });
+
+      render(<DataListHeader />);
+      expect(screen.queryByTestId("mock-header")).not.toBeInTheDocument();
+    });
+
+    it("should render when at least one item is selected", () => {
+      mockUseBatchSelect.mockReturnValue({
+        hasAtLeastOneSelected: true,
+        canSelect: false,
+        canSelectAll: false,
+      });
+
+      render(<DataListHeader />);
+      expect(screen.getByTestId("mock-header")).toBeInTheDocument();
+    });
+
+    it("should not render when no items selected even if selection is enabled and canSelectAll is true", () => {
+      mockUseBatchSelect.mockReturnValue({
+        hasAtLeastOneSelected: false,
+        canSelect: true,
+        canSelectAll: true,
+      });
+
+      render(<DataListHeader />);
+      expect(screen.queryByTestId("mock-header")).not.toBeInTheDocument();
+    });
+
+    it("should not render when layout is undefined regardless of selection state", () => {
+      mockUseActiveLayout.mockReturnValue({ layout: undefined });
+      mockUseBatchSelect.mockReturnValue({
+        hasAtLeastOneSelected: true,
+        canSelect: true,
+        canSelectAll: true,
+      });
+
+      render(<DataListHeader />);
+      expect(screen.queryByTestId("mock-header")).not.toBeInTheDocument();
+    });
+  });
+
+  describe("with different responsive sizes", () => {
+    it("should use the first available breakpoint when current size is not in headerVisibility", () => {
+      // Mock a scenario where we're at a medium breakpoint but only small is defined in visibility
+      mockUseResponsiveSizing.mockReturnValue({
+        xs: false,
+        sm: false,
+        md: true,
+      });
+      mockUseDataListContext.mockReturnValue({
+        headerVisibility: { xs: true, sm: true }, // Only xs and sm defined
+        headers: mockHeaders,
+        layoutBreakpoints: ["xs", "sm", "md"],
+      });
+
+      render(<DataListHeader />);
+      expect(screen.getByTestId("mock-header")).toBeInTheDocument();
+    });
+
+    it("should respect headerVisibility for the current breakpoint", () => {
+      // Set medium breakpoint to false in visibility
+      mockUseResponsiveSizing.mockReturnValue({
+        xs: false,
+        sm: false,
+        md: true,
+      });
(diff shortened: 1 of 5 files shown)


packages/components/src/DataList/components/DataListHeader/DataListHeaderCheckbox.tsx
```diff
@@ -108,3 +103,45 @@ export function DataListHeaderCheckbox({ children }: DataListHeaderCheckbox) {
...
+  onSelect,
+}: {
+  readonly canSelectAll: boolean;
+  readonly children: ReactElement;
+  readonly hasAtLeastOneSelected: boolean;
+  readonly selectedCount: number;
+  readonly deselectText: string;
+  readonly onSelect:
+    | ((selected: DataListSelectedType<string | number>) => void)
+    | undefined;
```
nad182, line 122: I'm wondering if there was a specific reason you chose union over optional property? I.e.
```suggestion
  readonly onSelect?: ((selected: DataListSelectedType<string | number>) => void);
```

Not necessarily requesting a change, just curious.

ZakaryH: fair question. to be honest Claude auto filled that and I hit Tab 😅 

I think there can be reason to do one approach over the other in certain situations, but in this case there was no real intent behind it.

packages/components/src/DataList/components/DataListHeader/DataListHeader.tsx
```diff
@@ -44,3 +55,17 @@ export function DataListHeader() {
...
+function isHeaderHidden(
+  visible: boolean | undefined,
+  hasAtLeastOneSelected: boolean,
+  layout: LayoutRenderer<DataListObject> | undefined,
+  canSelect: boolean,
+  canSelectAll: boolean,
+) {
+  return (
+    (!visible && !hasAtLeastOneSelected) ||
+    !layout ||
```
nad182, line 68: This is mainly for me to understand, what does `!layout` mean? I can see in the code that `const layout = layouts[activeBreakpoint];`. But I'm just trying to understand in which cases it will be `falsy`.

packages/components/src/DataList/components/DataListHeader/DataListHeaderCheckbox.tsx
nad182: I also asked the AI to explain the benefits of using option 1 (your way) and option 2 (the suggested way above), and this is what it said:

> Practical Impact
>In most cases, Option 2 (optional property) is preferred because:
>- It's more idiomatic TypeScript
>- Allows cleaner object creation (can omit the property)
>- Better represents "this callback may or may not be provided"
>- More concise syntax
>
> Option 1 would be used when you specifically need to distinguish between "property exists but is undefined" vs "property doesn't exist" - which is rare in practice.
For callback properties like onSelect, Option 2 is the conventional choice.

packages/components/src/DataList/components/DataListHeader/DataListHeader.tsx
ZakaryH: that's a good question. I'm actually not sure when `layout` would ever be falsey.

I thought I understood but after digging into it it turns out I was wrong.

if you have 
```
<DataList.Layout size="md">
...
</DataList.Layout>
```
as your only defined layout, even at what would be the `xs` screen size, your `activeBreakpoint` is still `md` because of this 

```
  const activeBreakpoint =
    reversedLayoutBreakpoints.find(key => breakpoints[key]) ||
    layoutBreakpoints[0];
```

where it'll fallback to `layoutBreakpoints[0]` AKA `md`

then if you don't specify a size on your `DataList.Layout` it gets `xs` by default.

lastly, if you don't provide any `Layout` at all, well you won't have anything to show at all so that's not a real case.

**conclusion**
in reality, it doesn't seem like you can ever have `layout` be `undefined`. it seems to me that that's a result of our types that say "any one of these can be undefined" which is true
image

but what that type definition is missing is something that says "any of them can be undefined, but at least one of them will not be" and I believe that missing qualifier is what results in our types asserting that `layout` can be `undefined`

ZakaryH:
most of this PR is tests. I've outlined the conditions in the "testing" section of the PR description.

I've gone over the tests and pretty confident they are all asserting the right things. even still I'd appreciate a quick sanity check that I'm not asserting anything incorrect + making sure that my actual component logic isn't flawed.

packages/components/src/DataList/components/DataListHeader/DataListHeader.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe("when headerVisibility is true", () => {
+    it("should render the header when layout is defined", () => {
+      render(<DataListHeader />);
+      expect(screen.getByTestId("mock-header")).toBeInTheDocument();
```
nad182, line 67: I think using `toBeVisible` assertion might be a bit more useful when testing elements presence in the UI (so here, and in tests below):
```suggestion
      expect(screen.getByTestId("mock-header")).toBeVisible();
```

packages/components/src/DataList/components/DataListHeader/DataListHeader.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+    it("should not render when no items are selected and selection is disabled", () => {
+      mockUseBatchSelect.mockReturnValue({
+        hasAtLeastOneSelected: false,
+        canSelect: false,
+        canSelectAll: false,
+      });
+
+      render(<DataListHeader />);
+      expect(screen.queryByTestId("mock-header")).not.toBeInTheDocument();
```
nad182, line 123: Just wanted to comment that in this case I think `toBeInTheDocument` is perfectly reasonable!

packages/components/src/DataList/components/DataListHeader/DataListHeader.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+        canSelect: false,
+        canSelectAll: false,
+      });
+
+      render(<DataListHeader />);
+      expect(screen.getByTestId("mock-header")).toBeInTheDocument();
+    });
+
+    it("should not render when layout is undefined", () => {
+      mockUseActiveLayout.mockReturnValue({ layout: undefined });
```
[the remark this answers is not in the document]
ZakaryH: oh sorry it can't. not in real life.

I asked Claude to generate tests for all the code paths, and this is technically one due to our types.

packages/components/src/DataList/components/DataListHeader/DataListHeader.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+      });
+
+      render(<DataListHeader />);
+      expect(screen.queryByTestId("mock-header")).not.toBeInTheDocument();
+    });
+
+    it("should render when at least one item is selected", () => {
+      mockUseBatchSelect.mockReturnValue({
+        hasAtLeastOneSelected: true,
+        canSelect: false,
```
nad182, line 140: Shouldn't this be `true` if `hasAtLeastOneSelected: true`? I mean, is it theoretically possible to select one item without having the ability to select?

packages/components/src/DataList/components/DataListHeader/DataListHeaderCheckbox.test.tsx
```diff
@@ -17,76 +21,129 @@ beforeEach(() => {
...
+        </DataListContext.Provider>,
+      );
 
-    expect(screen.getByRole("checkbox")).toBeInTheDocument();
-    expect(
-      screen.getByRole("button", { name: deselectAllLabel }),
-    ).toBeInTheDocument();
-    expect(screen.queryByText(textToBeFound)).not.toBeInTheDocument();
+      expect(screen.getByTestId("child-element")).toBeInTheDocument();
+      expect(screen.getByRole("checkbox")).toBeInTheDocument();
```
nad182, line 83: A good candidate for having both `toBeInTheDocument` and `not.toBeVisible`, perhaps?

nad182 (commented):
As a general comment/question (not necessarily a request to make any changes), do we want to check the behaviour by not just mocking values, and instead performing interactions, like actually selecting/deselecting various checkboxes to assert the header behaviour?
Might be too big of an ask, but just wanted to throw it out there for your consideration.

packages/components/src/DataList/components/DataListHeader/DataListHeader.test.tsx
ZakaryH: with that, the tests are pretty exhaustive... but not representative of reality. 

honestly kinda weird situation where I feel like these can catch more bugs, but they're testing scenarios that aren't real.

I think the real solution would be to correct our types to make them more aligned with reality, but that's of course a lot more work.

for the purposes of this PR, I think I'm in favour of having more coverage even if it's not as realistic as it should be.

packages/components/src/DataList/components/DataListHeader/DataListHeaderCheckbox.test.tsx
ZakaryH: it's not a bad idea. unfortunately there are some limitations with `toBeVisible` that make it awkward

`toBeVisible`
```
     * An element is visible if **all** the following conditions are met:
     * * it does not have its css property display set to none
     * * it does not have its css property visibility set to either hidden or collapse
     * * it does not have its css property opacity set to 0
     * * its parent element is also visible (and so on up to the top of the DOM tree)
     * * it does not have the hidden attribute
     * * if `<details />` it has the open attribute
     
```

we are giving the parent `div` `hidden` via a classname

```
.selectAllCheckbox {
  display: flex;
  visibility: hidden;
...
```

it seems classnames aren't calculated though, and despite having that class present it fails the `not.toBeVisible`
image

maybe it does work with simpler class definitions? it could be our css modules causing it to not work.

then if I put that same logic inline via `style={{ visibility: canSelectAll ? "visible" : "hidden" }}` that works and our check that the element containing the Checkbox is not visible. that causes a new problem though where because the Checkbox is a child of something deemed not visible, it can't find the checkbox haha

image

so for the assertion on the checkbox directly to work, it seems we'd have to do some inline styles on the checkbox setting the `visibility`

ZakaryH:
>As a general comment/question (not necessarily a request to make any changes), do we want to check the behaviour by not just mocking values, and instead performing interactions, like actually selecting/deselecting various checkboxes to assert the header behaviour?

>Might be too big of an ask, but just wanted to throw it out there for your consideration.


that's a fair question. I did consider adding some integration tests that verify behavior. knowing how much that's goes to involve I didn't want to blow up this PR, so focused on unit tests instead.

lemme see what Claude can cook up real quick. if it's reasonable I'll commit it. if it turns this into a 1000+ lines PR, I'm not sure it's worth it.

packages/components/src/DataList/components/DataListHeader/DataListHeaderCheckbox.test.tsx
nad182: Thanks for looking into this! I don't feel strongly about this assertion, especially since yours seems to do the job well enough!

ZakaryH:
@nad182 added some integration tests. that should be all feedback addressed!

packages/components/src/DataList/components/DataListHeader/DataListHeaderCheckbox.tsx
```diff
@@ -29,15 +35,17 @@ export function DataListHeaderCheckbox({ children }: DataListHeaderCheckbox) {
     onSelect,
   } = useBatchSelect();
 
-  // If there's no onSelectAll or onSelect, we don't need to render the checkbox.
-  if (!canSelectAll) return children;
+  // If there's no onSelect, we don't need to render the checkbox.
+  // We'll still render the (invisible) checkbox even if only onSelect is provided for alignment.
+  if (!canSelect) return children;
```
nad182, line 40: I'm not 100% certain I fully understand the logic.
- If `canSelect` is false (i.e. `selected` and/or `onSelect` are falsy), we do not display the checkbox in the header.
- if `canSelectAll` is false (i.e. `onSelect` and/or `onSelectAll` are falsy), we also do not display the checkbox in the header.

Can we not combine the two here, and below return the original `AnimatedSwitcher`? I'm a bit confused why you moved `canSelectAll` check as a part of the newly created `ColumnHeaderContent`. Could you help me understand why it's needed?

ZakaryH: >If canSelect is false (i.e. selected and/or onSelect are falsy), we do not display the checkbox in the header.

we don't display anything other than the column headers (`children`), no checkbox, no select-all UI

>if canSelectAll is false (i.e. onSelect and/or onSelectAll are falsy), we also do not display the checkbox in the header.

for this case we **do** display the checkbox, sort of. we'll render it with its default style that is `visibility: hidden` meaning users and screen readers won't see it.

then there's a third case, where `canSelectAll` is true, that's when you get both the checkbox that can actually toggle between visible and not AND you get the switcher that can show the select-all UI


the simplest way I can describe it is:

- I don't have selections -> just heading and only headings
- I can ONLY select one or more items -> we still render the select-all checkbox for the space but you can't see or use it
- I can select all -> I have the checkbox at the top, it can toggle visibility AND I can switch to the select-all UI

ZakaryH: maybe some visuals would help

**No selecting at all**
image

**onSelect**
image

**onSelectAll**
image

nad182 (approved):
Tested locally and in Storybook (by stubbing some data here: https://job-126863-datalist-header-f.atlantis.pages.dev/storybook/?path=/code/components-lists-and-tables-datalist-web--custom-render-empty-state)

Thanks for adding comprehensive tests!

PR #2595, opened 2025-06-20
fix(components): Adding optional strategy prop to popovers to fix scrolling layout issue

ethanmillard:
## Motivations

There is an issue with the popovers when they scroll outside of the viewport in the sidenav, as it leads to the screen being larger and you can scroll down to this weird white space.

## Changes

- Introduced a new optional `strategy` prop to the Popover component, allowing users to specify the positioning strategy ('absolute' or 'fixed').
- Updated PopoverProvider and usePopover hooks to support the new strategy prop.

- Also added some padding so the popover wouldn't sit right on the bottom of the page.

https://github.com/user-attachments/assets/73d224d0-389d-46d5-92a1-699ff0f70863

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

This is the issue that's being fixed

https://github.com/user-attachments/assets/6bdbd58e-aa81-49de-bf67-3a1c089ae39b

### Security

- 

## Testing

Changes can be

---

### DIFF
--- a/packages/components/src/Popover/Popover.types.ts
+++ b/packages/components/src/Popover/Popover.types.ts
@@ -20,6 +20,12 @@ export interface PopoverProps {
   readonly open: boolean;
 
   /**
+   * The strategy to use for the Popover.
+   * @default 'absolute'
+   */
+  readonly strategy?: "absolute" | "fixed";
+
+  /**
    * Callback executed when the user wants to close/dismiss the Popover
    */
   readonly onRequestClose?: () => void;
@@ -54,7 +60,10 @@ export interface PopoverProps {
 }
 
 export type PopoverProviderProps = PropsWithChildren<
-  Pick<PopoverProps, "preferredPlacement" | "attachTo" | "open"> & {
+  Pick<
+    PopoverProps,
+    "preferredPlacement" | "attachTo" | "open" | "strategy"
+  > & {
     /**
      * **Use at your own risk:** Custom class names for specific elements. This should only be used as a
      * **last resort**. Using this may result in unexpected side effects.
--- a/packages/components/src/Popover/usePopover.tsx
+++ b/packages/components/src/Popover/usePopover.tsx
@@ -7,7 +7,11 @@ export const usePopover = ({
   preferredPlacement,
   attachTo,
   open,
-}: Pick<PopoverProps, "preferredPlacement" | "attachTo" | "open">) => {
+  strategy = "absolute",
+}: Pick<
+  PopoverProps,
+  "preferredPlacement" | "attachTo" | "open" | "strategy"
+>) => {
   const [popperElement, setPopperElement] = useState<HTMLElement | null>();
   const [arrowElement, setArrowElement] = useState<HTMLElement | null>();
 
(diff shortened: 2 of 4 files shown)


packages/components/src/Popover/Popover.types.ts
```diff
@@ -19,6 +19,12 @@ export interface PopoverProps {
    */
   readonly open: boolean;
 
+  /**
+   * The strategy to use for the Popover.
+   * @default 'absolute'
+   */
+  readonly strategy?: "absolute" | "fixed";
```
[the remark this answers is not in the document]
ethanmillard: thanks @ZakaryH 

and ya, I had the same thoughts, but the idea of changing the default for everyone was a little terrifying lol and I don't know `react-popper` well enough. Which is why I opted to just allow the option

packages/components/src/Popover/usePopover.tsx
```diff
@@ -23,6 +27,13 @@ export const usePopover = ({
           offset: [0, 10],
         },
       },
+      {
+        name: "preventOverflow",
+        options: {
+          boundary: "viewport",
+          padding: 8,
+        },
+      },
```
ZakaryH, line 36: couple thoughts on this change

first, we'd be applying it to Popovers across the board. I think I get what preventOverflow does, but I'd definitely like to make sure we're confident applying this to everything - otherwise it might also be something to opt in to.

second, I see 8 in the example from Popper but I wonder what that's based on. seems fairly arbitrary. I could see there being cases where you don't really mind your Popover getting very close to an edge.

oh one more is that if we did go ahead with this, it would be preferable to use a design spacing token like `space-small` (not necessarily that one, just an example)

ZakaryH:
I'm having trouble reproducing the happy path with this solution
image

I've packaged the local version of this branch, cleared all caches and installed it in product, then applied the `strategy: "fixed"` prop but I'm still seeing the bad behavior.

I did force the popups to show but I can't imagine that having any real effect on what I'm trying to test out here.

packages/components/src/Popover/usePopover.tsx
ethanmillard: those all make sense to me

should I make those changes?
or is this still in the decision phase of if this is the right path or not?

ZakaryH: I think this is the right path to go overall, with the `strategy` prop and its default.

I'll try to play around with the `preventOverflow` a bit more. I haven't really been able to get it to behave how I expected, likely due to me not setting up the right scenario.

I would like to know how you are testing these changes in the frontend project though, I'm getting some really bizzarre results when I try to use a local version of these changes. seems like Vite is doing something unexpected.

then there's one other thing I noticed that I'll make a separate comment about.

ethanmillard: I ended up just copying the entire popover folder into JFE, then fixed the couple of imports

I was able to play around with the changes really fast that way and just copied back everything I had done into atlantis

ZakaryH:
something I noticed we'll want to add is in the non-composed `Popover` ie. when you use `Popover` directly and not `Popover.Provider`, we'll want to actually implement the `strategy` there too 

since we're (rightfully) defining the `strategy` on  `PopoverProps` it's going tell you that you can do `<Popover strategy="fixed"...>` however we're not grabbing that value off the props, nor are we passing it to  the `Popover.Provider`

https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Popover/Popover.tsx#L12-L28

that way the non-composed invocations of `Popover` can also leverage this prop

PR #2600, opened 2025-06-25
fix(components-native): fix placement of assistive text on Checkbox [JOB-125934]

chris-at-jobber:
## Motivations

Assistive text on Checkbox does not match intended designs

**Actual**
image

**Design**
image

### Fixed

- Mobile checkbox now matches designs

**Short text**
image

**Wrapping text**
image

## Testing

Can test locally, limited usage in the app

Changes can be

---

### DIFF
--- a/packages/components-native/src/Checkbox/Checkbox.style.ts
+++ b/packages/components-native/src/Checkbox/Checkbox.style.ts
@@ -12,7 +12,7 @@ export const useStyles = buildThemedStyles(tokens => {
       width: "100%",
       justifyContent: "space-between",
       flexDirection: "row",
-      alignItems: "start",
+      alignItems: "flex-start",
       paddingVertical: tokens["space-small"],
     },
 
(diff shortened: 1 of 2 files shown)


nad182 (approved):
I only tested it in Web Chrome. It looks better and I'm approving this PR.
However, maybe it's just my eyes, since I didn't see anything in CSS signalling it, but it seems like the checkbox text and assistive text are ever so slightly misaligned on the left side (it looks like the assistive text is shifted a bit more to the left).

image

ZakaryH:
@nad182 I wouldn't really trust web's rendering of native components, it's very unreliable. are you able to test this on a device, or even a simulator?

ZakaryH:
@nad182 ah, yeah so it might work - the problem is that it won't  be rendering the component like it will actually be rendered in a RN app.

everything in storybook, since it's a normal React website, that renders a RN component will be using the `react-native-web` package to do so.

there are many things that are completely identical between the two, but in general I would suggest trying out components in the environment they'll actually be rendered in. `react-native-web` is really only a documentation tool, not a proper proxy for native development/testing.

PR #2601, opened 2025-06-25
fix(components-native): Do not render Select's label element when no label is provided [JOB-125934]

chris-at-jobber:
## Motivations

When no label is provided to Select, there's still an element rendered that occupies space
image

### Changed

- wrapped the Text for the label in a conditional so it only show up when present

### Fixed

- Select now looks nicer when you don't pass it a label

## Testing

Because of web differences, you have to test this one in the app (job forms or requests have examples of this use case)

Changes can be

---

### DIFF
--- a/packages/components-native/src/Select/Select.tsx
+++ b/packages/components-native/src/Select/Select.tsx
@@ -153,14 +153,15 @@ export function Select({
           <View
             style={[styles.container, (invalid || !!error) && styles.invalid]}
           >
-            <Text
-              level="textSupporting"
-              variation={textVariation}
-              hideFromScreenReader={true}
-            >
-              {label}
-            </Text>
-
+            {label && (
+              <Text
+                level="textSupporting"
+                variation={textVariation}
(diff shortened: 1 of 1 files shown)


nad182 (commented):
Tested in Web Chrome, so I assume it works fine on the mobile device (since it appears to be the same way on master):

image

The chevron is not placed correctly, but I assume it's because I'm not using a mobile device.

nad182 (commented):
I tested in BrowserStack (using Google Pixel 9 device), and I'm still seeing the same issue with the chevron (regardless of the presence of the label):

image

image

Kyrluckechuck (approved):
Testing via installing onto jobber-mobile, but had to use the last commit prior to the RN upgrade:
```bash
git checkout 0ea247a2d4f23b8f5732fcb7c9610c809ae1d6eb
dev up
npm i ../atlantis/jobber-components-native-0.81.1.tgz
npm run android
```
image

PR #2603, opened 2025-06-25
fix(components): Fix InputDate and InputTime v2 Issues

ZakaryH:
## Motivations

InputTime v2 has a problem where the input and label are not associated, because we are not giving the input the corresponding id

InputDate has a plethora of issues stemming from the way its types are defined with the immediate, blocking issue being that it is attempting to spread `readonly` and `showIcon` onto an html element

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

passed the id we have onto the InputTime

stopped spreading all of `props` onto the InputText inside InputDate, instead only applying the specific props we want to use. also stopped spreading the `activatorProps`/`newActivatorProps` since there are a handful of problems with those too.

### Security

- 

## Testing

Changes can be

---

### DIFF
--- a/packages/components/src/InputDate/InputDate.rebuilt.tsx
+++ b/packages/components/src/InputDate/InputDate.rebuilt.tsx
@@ -1,5 +1,4 @@
 import React, { forwardRef } from "react";
-import omit from "lodash/omit";
 import { useInputDateActivatorActions } from "./useInputDateActivatorActions";
 import { InputDateRebuiltProps } from "./InputDate.types";
 import { Suffix } from "../FormField";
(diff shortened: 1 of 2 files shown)


packages/components/src/InputDate/InputDate.rebuilt.tsx
```diff
@@ -61,8 +59,21 @@ export const InputDateRebuilt = forwardRef(function InputDateInternal(
       // We prevent the picker from opening on focus for keyboard navigation, so to maintain a good UX for mouse users we want to open the picker on click
       <div onClick={onClick}>
         <InputText
-          {...newActivatorProps}
-          {...props}
```
ZakaryH, line 65: I'm going to write a hefty ticket about these two sets of props. there's a ton to unpack that I won't go into for the purposes of this bugfix.

the shortest possible version of it is that spreading like this is hiding a lot. there are numerous invalid props, and overlapping props that will be stomped on by the second spread. Typescript isn't able to do its job when we spread

image

packages/components/src/InputDate/InputDate.rebuilt.tsx
```diff
@@ -83,3 +94,5 @@ export const InputDateRebuilt = forwardRef(function InputDateInternal(
     );
   }
 });
+
+InputDateRebuilt.displayName = "InputDateRebuilt";
```
ZakaryH, line 98: the nonstandard forward ref approach we had was causing stack traces to become hard to follow. this should address that.

mattander (approved):
Nice! Thanks for addressing this so quickly.

nad182 (approved):
Approving in case "Atlantis review" is also necessary.
Verified that warnings are no longer being shown in the console when using this branch.

PR #2607, opened 2025-06-28
docs: Add a Color Token Search component to the docs

taylorvnoj:
## Motivations

I've been hearing from Atlantis users that they've been wanting a color token search (I've also always wanted one) - so here it is. This is helpful in multiple scenarios: finding the correct semantic token to use instead of grabbing the base color, maybe a figma doesn't have the token name listed but has the hex value, etc.
I think this will encourage token usage & will save time finding the correct token to use.

### Added

 
- Added a ColorSearch component to the design documentation (Colors page)
  - Allows users to search for color tokens by entering a hex code, HSL, or RGB(A) value
  - Displays matching color tokens with a swatch, token name, and a copy-to-clipboard button
  - Shows a message when no matching color tokens are found

https://github.com/user-attachments/assets/0fbee1b0-7f9a-45c3-b634-ab956aa71535

## Testing

Enter a bunch of different color values and make sure you're getting a variety of tokens. 
When there is no matching token, subdued text under the input will inform you of that.
For example you can find color token values in:
- `packages/design/src/assets/tokens.all.colors.ts`
- `packages/design/src/tokens/`

Changes can be

---


packages/site/src/utils/colorUtils.ts
nad182, line 1: Is there, perhaps, a well-maintained and small size library that has all of those utils included? I'm not necessarily pushing for using an external dependency, but my concern is that none of those newly added utils are covered by tests. Moreover, it feels like to properly test everything, we'd need to write very extensive and robust test suites.

taylorvnoj: There are a few lightweight packages but I would also like to avoid using an external dependency, especially when it's just 1 utils file that gets us what we want.
I am happy to write a test file for the utils. Many other files in our new docs site do not have test files because we were just pushing to release but I can start that pattern here

packages/site/src/utils/colorTokenMatch/colorTokenMatch.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+            typeof colorValue === "string" &&
+            colorsAreEqual(normalizedInput, normalizeColor(colorValue), 2),
+        );
+
+        const matchTokens = matches.map(([token]) => token);
+
+        expectedTokens.forEach(expected => {
+          expect(matchTokens).toContain(expected);
+        });
+      });
```
nad182, line 115: Would it be possible to leverage `test.each` here? Similar to this [JFE example](https://github.com/GetJobber/jobber-frontend/blob/main/apps/jobber-online/app/features/invoices/invoicesOverview/__tests__/InvoicesOverview.test.tsx#L19-L33).

packages/site/src/utils/colorTokenMatch/colorTokenMatch.ts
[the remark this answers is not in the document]
taylorvnoj: oops! Good call!

packages/site/src/utils/colorTokenMatch/colorTokenMatch.test.tsx
taylorvnoj: updated!

nad182 (approved):
Nice addition to the docs site! I'm accepting the PR.

One thing I think would be nice to do (perhaps, in the next iteration) is to make search more flexible (by supporting partial matches), which could make it even faster/more useful to find tokens.
Otherwise, you need to be very precise (which is fine for the most part if colors are just being copied from Figma).

Screenshot_20250630-171507
Screenshot_20250630-171521

taylorvnoj:
@nad182 good call out! Thanks for approving - this way we can get it out and get some feedback. That change should be very minimal so happy to iterate on that soon!

PR #2610, opened 2025-07-03
feat(components): Popover implementation using floating-ui

JeffMarvin:
## Motivations

We have issues with Popover not repositioning itself when its `attachTo` element moves after the initial render. Implementing Popover with floating-ui resolves the issue with its [autoUpdate](https://floating-ui.com/docs/autoUpdate).
Everything should behave the same as today, except for the addition of shift.

## Changes

- floating-ui replaces react-popper 
- [shift](https://floating-ui.com/docs/shift) from floating-ui is introduced. When Popover reaches the edge of the container, it will shift first, then flip.
- Popover repositions itself when the `attachTo` element moves

### Added

- 

### Deprecated

- 

### Removed

- 

### Fixed

- The bugs in `Basic` and `Informational` stories where Popover is aligned to the center of the container are fixed

### Security

- 

## Testing

- `Moving Attach To Element` story can be used to confirm that Popover repositions itself when attachTo element moves

Changes can be

---


packages/components/src/Popover/Popover.test.tsx
```diff
@@ -47,6 +47,96 @@ describe("Non-composable Popover", () => {
     expect(screen.queryByText(content)).toBeNull();
   });
 
+  it("should handle attachTo element changes correctly", () => {
```
edison-cy-yang, line 50: Also don't think this is a valuable test:
1. it's not actually testing the popover is attached to the second element, just that it's visible
2. The goal of this test is pretty much "confirm we're passing referenceElement into useFloating hook from floating-ui, and that referenceElement will change if I change it"

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -255,6 +345,92 @@ describe("Composable Popover", () => {
...
+    act(() => {
+      container.appendChild(attachedElement);
+    });
+
+    // The popover should still be positioned relative to the moved element
+    expect(screen.getByText(content)).toBeInTheDocument();
+    expect(popover).toBeInTheDocument();
+  });
+
+  it("should handle attachTo element changes correctly (composable)", () => {
```
edison-cy-yang, line 385: Same for this one, don't think it's a valuable test

docs/components/Popover/Web.stories.tsx
```diff
@@ -84,6 +91,138 @@ const InformationalTemplate: ComponentStory<typeof Popover> = args => {
 
 export const Informational = InformationalTemplate.bind({});
 
+const MovingElementTemplate: ComponentStory<typeof Popover> = args => {
+  const divRef = useRef<HTMLDivElement>(null);
+  const [showPopover, setShowPopover] = useState(false);
+  const [showBanner, setShowBanner] = useState(false);
+  const [position, setPosition] = useState({ x: 0, y: 0 });
+
+  const moveElement = (direction: "up" | "down" | "left" | "right") => {
```
ZakaryH, line 100: while I do think this is pretty helpful while testing and developing, I'm not sure if this makes sense to keep around as a permanent webstory.

it's not exactly a common or realistic usage.

I wouldn't mind keeping the banner changing the layout example, that's fair and real. I'm not convinced about the rest though.

packages/components/src/Popover/Popover.module.css
```diff
@@ -35,47 +35,28 @@
...
-.popover[data-popper-placement^="left"] > .arrow::before {
   transform: rotate(135deg);
 }
 
 .popover[data-popper-placement^="right"] > .arrow {
-  left: var(--popover--position--offset);
-}
-.popover[data-popper-placement^="right"] > .arrow::before {
-  transform: rotate(-45deg);
+  transform: rotate(315deg);
```
ZakaryH, line 61: I get that these are equivalent. any reason for the change though?

packages/components/src/Popover/Popover.module.css
```diff
@@ -35,47 +35,28 @@
 }
 
 .arrow {
-  visibility: hidden;
-}
-
-.arrow,
-.arrow::before {
```
ZakaryH, line 42: testing it out the arrow is still doing what it is supposed to do and looks good.

what changed though? do we not need to build our our `::before` styles? or were they never necessary?

docs/components/Popover/Web.stories.tsx
edison-cy-yang: Yeah that's fair. Even the banner changing layout example is showcasing what floating-ui **is supposed to do** out of box.

packages/components/src/Popover/PopoverContext.tsx
```diff
@@ -85,7 +94,7 @@ function PopoverWrapper({
         ref={setPopperElement}
         style={{ ...popperStyles.popper, ...UNSAFE_style?.container }}
         className={classes}
-        {...attributes.popper}
+        data-popper-placement={placement}
```
ZakaryH, line 97: any reason we're keeping this name despite not using `popper` anymore?

I do see a couple references in a test in product looking for this selector, but I'd argue that test should be modified to not be so reliant on internal details, nor testing functionality that is tested by the component library - which is to say if the only reason is to not break that test, I wouldn't say it's a priority.

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -47,6 +47,65 @@ describe("Non-composable Popover", () => {
...
 
+  describe("preferredPlacement", () => {
+    it("should use bottom placement by default", () => {
+      renderPopover({
+        open: true,
+        children: content,
+      });
+
+      const popover = screen.getByTestId("ATL-Popover-Container");
+      expect(popover).toHaveAttribute("data-popper-placement", "bottom");
```
ZakaryH, line 58: I feel like this is bordering on testing the library's functionality since we're getting `placement` from floating ui, then passing it down and putting those attributes on a div....

though actually I guess the default bottom is unique to our setup. nevermind, I like it.

packages/components/src/Popover/Popover.test.tsx
```diff
@@ -47,6 +47,65 @@ describe("Non-composable Popover", () => {
...
+  describe("Accessibility", () => {
+    it("should have proper ARIA attributes when open", () => {
+      renderPopover({
+        open: true,
+        children: content,
+      });
+
+      const popover = screen.getByTestId("ATL-Popover-Container");
+      expect(popover).toHaveAttribute("role", "dialog");
+      expect(popover).toHaveAttribute("data-elevation", "elevated");
```
ZakaryH, line 82: I don't believe `data-elevation` is an accessibility attribute

packages/components/src/Popover/Popover.tsx
```diff
@@ -46,15 +47,39 @@ Popover.Arrow = function PopoverArrow({
...
 
+  // the arrow will get positioned opposite to the placement side
+  const staticSideMap: Record<Side, Side> = {
+    top: "bottom",
+    right: "left",
+    bottom: "top",
+    left: "right",
+  } as const;
+
+  // placement can include alignment (e.g. 'top-start') but we just want the side
```
ZakaryH, line 62: this would be the floating ui hook giving us a placement of `top-start` based off where it thinks it should be placed or something right?

we don't support all of the placement options in our Popover props for someone to pass that in

packages/components/src/Popover/PopoverContext.tsx
```diff
@@ -8,11 +8,19 @@ import { AtlantisPortalContent } from "../AtlantisPortalContent";
 
 interface PopoverContextProps {
   setArrowElement: (element: HTMLElement | null) => void;
-  popperStyles: { [key: string]: CSSProperties };
+  popperStyles: {
+    popper: CSSProperties;
+    arrow?: {
+      x?: number;
+      y?: number;
+      centerOffset?: number;
```
ZakaryH, line 16: what's this for?

I don't see us using it where we are applying the `popperStyles.x` and `y`

packages/components/src/Popover/Popover.tsx
```diff
@@ -46,15 +47,39 @@ Popover.Arrow = function PopoverArrow({
...
+  const staticSide = staticSideMap[placementKey];
+
+  const arrowStyles: React.CSSProperties = {
+    position: "absolute",
+    // only left or top will be defined at a time
+    left: popperStyles.arrow?.x != null ? `${popperStyles.arrow?.x}px` : "",
+    top: popperStyles.arrow?.y != null ? `${popperStyles.arrow?.y}px` : "",
+    right: "",
+    bottom: "",
+    [staticSide]: "-7px",
```
ZakaryH, line 73: I forget, what's this correspond to? the width of the base of the triangle for the arrow?

ZakaryH: oh no, it's for placing it on the edge of the popover. hmm I don't get how we get to `-7` when the width/height is 16px

if you understand it, plopping that value into a variable with a good name would make this easier to read.

packages/components/src/Popover/PopoverContext.tsx
```diff
@@ -8,11 +8,19 @@ import { AtlantisPortalContent } from "../AtlantisPortalContent";
...
+    };
+  };
+  placement?: string;
```
ZakaryH, line 19: this type feels a little loose. like yes it'll be a string, but it's a predefined set of strings. I wonder if we couldn't use `Placement` from floating UI?

packages/components/src/Popover/PopoverContext.tsx
```diff
@@ -60,14 +69,14 @@ export function PopoverProvider({
...
-  attributes,
   setPopperElement,
   UNSAFE_className,
   UNSAFE_style,
+  placement,
 }: {
   readonly children: React.ReactNode;
-  readonly attributes: Record<string, { [key: string]: string } | undefined>;
   readonly setPopperElement: (element: HTMLElement | null) => void;
+  readonly placement?: string;
```
ZakaryH, line 79: same thing here. bit of a loose type.

packages/components/src/Popover/usePopover.tsx
```diff
@@ -8,41 +16,50 @@ export const usePopover = ({
...
-      modifiers,
-      placement: preferredPlacement,
+  const referenceElement = isHTMLElement(attachTo)
+    ? attachTo
+    : attachTo.current;
+
+  const { refs, floatingStyles, middlewareData, placement } = useFloating({
+    placement: (preferredPlacement === "auto"
+      ? "bottom"
+      : preferredPlacement) as Placement,
```
ZakaryH, line 42: I think I get this? but it's still kinda weird to me.

I know in Popover we do `preferredPlacement = "auto",` so that's the default if you don't provide one, has that always just been mapped to "bottom"?

also I just noticed for non composed Popover we assign the preferred placement with the default. for the composed version that would have the Provider as the top level element, it doesn't assign "auto"

seems like it still manages to fallback to using bottom though? I wonder if our "auto" assignment and logic are even necessary? or alternatively if our Provider needs that default too.

packages/components/src/Popover/usePopover.tsx
```diff
@@ -8,41 +16,50 @@ export const usePopover = ({
...
+    strategy: "fixed",
```
ZakaryH, line 43: this is quite different from before!

https://floating-ui.com/docs/usefloating#strategy

>'absolute' — the floating element is positioned relative to its nearest positioned ancestor. With most layouts, this usually requires the browser to do the least work when updating the position.

this used to be our default, but now we're swapping to fixed which does behave differently

in the other Popover PR to fix positioning,  we added a strategy prop to allow opting in to `fixed` when you needed/wanted it rather than changing all Popover behavior across the board.

I do think this is a little spooky. do we understand the implications of swapping the strategy for all existing Popovers?

ZakaryH: also if we do go ahead with this, I'd expect this to fix the other Popover position bug we have open at the moment

packages/components/src/Popover/usePopover.tsx
```diff
@@ -8,41 +16,50 @@ export const usePopover = ({
...
+    middleware: modifiers,
+    elements: {
+      reference: referenceElement || null,
     },
-  );
+    whileElementsMounted: autoUpdate,
```
ZakaryH, line 48: >Warning
>Only use this option if your floating element is conditionally rendered, not hidden with CSS. Use an Effect to call and clean up autoUpdate in the latter scenario.

that is how our Popover works, so we are fine. hopefully we never change that 😅 

perhaps a comment would be nice here. what do you think?

ZakaryH (changes requested):
looking good! there are a couple minor things I feel we can improve

the biggest blocker or source of FUD for me is changing the `strategy` to be "fixed", the opposite of what we had before ("absolute") and the consequences of that

I could be convinced that's OK but I didn't see anything specifically mentioning it so I would like us to explore what that means and document why we feel it's fine to go ahead with

packages/components/src/Popover/Popover.test.tsx
edison-cy-yang: Removed in 63e17777f1b5df1479f4843fde73afe2a6932097

packages/components/src/Popover/Popover.tsx
edison-cy-yang: Yes, took this part out in b814fa9ca64b14e4d443c8ba66948d114ee9936e

packages/components/src/Popover/PopoverContext.tsx
edison-cy-yang: took this out in https://github.com/GetJobber/atlantis/pull/2610/commits/088cecbe6537f05ffdb616a86ffa4468906e063d

packages/components/src/Popover/PopoverContext.tsx
edison-cy-yang: Updated to Placement in https://github.com/GetJobber/atlantis/pull/2610/commits/8cb80036f3601e3cb1b013d04cace828dc40d676

packages/components/src/Popover/PopoverContext.tsx
edison-cy-yang: Same as above, https://github.com/GetJobber/atlantis/pull/2610/commits/8cb80036f3601e3cb1b013d04cace828dc40d676

packages/components/src/Popover/PopoverContext.tsx
edison-cy-yang: I'll get another ticket to clean it up

ZakaryH: what do you see as the benefit of doing it later vs now?

I ask because the PR is about swapping from popper, and I'd argue removing names that still reference would be part of that. 

if we wanted, it could be an additional subtask on this existing ticket that we make sure to do when pulling it into product.

packages/components/src/Popover/Popover.module.css
edison-cy-yang: No idea, I reverted it back

packages/components/src/Popover/usePopover.tsx
edison-cy-yang: You brought up a good point. With react-popper, placement: "auto" makes it pick the placement with the most space. The equivalent of that in floating-ui is `autoPlacement` (https://floating-ui.com/docs/autoPlacement), which is not tied to the `placement` argument passed into `useFloating`. It would be part of the middleware, and it's an alternative to `flip`, meaning they can't be used at the same time. 

So, we don't need "auto" placement anymore, and there won't be default preferredPlacement. A ton of the Popover usages in JFE are using default.

I am guessing our option is to use `autoPlacement` when default is used (with no `flip`), and `flip` with no `autoPlacement` when preferredPlacement is passed in?

packages/components/src/Popover/usePopover.tsx
edison-cy-yang: I'll be honest, I don't fully understand.

Do you think it makes sense to roll out this PR with strategy absolute first, make sure nothing goes wrong, then bring in that optional strategy prop? I do think it makes sense to keep it absolute unless fixed is needed

packages/components/src/Popover/usePopover.tsx
edison-cy-yang: Implemented what I just said in https://github.com/GetJobber/atlantis/pull/2610/commits/bbaf0ab164a7826171e13d998ee01f6762659627. It seems to work well.

edison-cy-yang: Also assigned default preferredPlacement on the provider for composed version in https://github.com/GetJobber/atlantis/pull/2610/commits/30a5552daca50e28649e8288eee2170c7f194cb0

packages/components/src/Popover/usePopover.tsx
edison-cy-yang: I don't think we would ever change it, but added it anyways https://github.com/GetJobber/atlantis/pull/2610/commits/20df3d7d27d240d7253b9256d23fd17a6c731f3a

docs/components/Popover/Web.stories.tsx
edison-cy-yang: I will remove this story after you're done with verifying the new changes

packages/components/src/Popover/usePopover.tsx
edison-cy-yang: Reverted it back to absolute for now, we can use the other ticket to support optional fixed strategy https://github.com/GetJobber/atlantis/pull/2610/commits/1aa7b023aa4f079555c2e5af38b873f179c8249f

packages/components/src/Popover/Popover.tsx
ZakaryH: gotcha. ok nice, I like us referring to that variable much better!

packages/components/src/Popover/usePopover.tsx
ZakaryH: interesting.... hmm so I think your new changes make a lot of sense. the only thing is that I worry they are going to result in different behavior?

if I'm understanding this right, previously if you gave us either nothing or `"auto"` as the preferredPlacement, we'd map that to `"bottom"`. now, we're properly applying an auto placer that won't always guarantee to put it on the bottom correct?

I'm trying to imagine where I'd expect to see a difference.... if you had an old one that got `"auto"` AND it did have enough room on both the top or bottom of it - it would be placed on the bottom. like this

image


now, with "auto" on this version with that same layout and props we get

image

packages/components/src/Popover/usePopover.tsx
ZakaryH: I do yeah.

that's what Ethan was doing in his PR to fix the Popover getting stuck in a weird place, and we both kinda felt it would be risky to change the default placement strategy across the board. opting in to `fixed` when you need it felt  safer.

like it's probably fine, maybe? the docs do suggest absolute is the one you most often want, that requires the least work from the browser

>With most layouts, this usually requires the browser to do the least work when updating the position.

though reading the warning

> the floating element is positioned relative to its nearest containing block (usually the viewport).  This is useful when the reference element is also fixed to reduce jumpiness with positioning while scrolling. It will in many cases also  "break” the floating element out of a clipping ancestor

I'm pretty sure that doesn't apply to us, since we're portalling it out to the body anyways. there are no reasonable ancestors above the `body` unless you've got some crazy iframe setup where you're rendering it.

I'm fine with exploring that separately. I do think it's the safer approach unless we want to take the time to truly understand the implications of swapping to `fixed` for all Popovers.

packages/components/src/Popover/usePopover.tsx
ZakaryH: oh! you're right, I missed that that was a new addition.

and the reason that example I posted from `master` is going below is because it still has the silly layout where the ref we're attaching it to spans the entire width, so of course it goes bottom because there's no space to the right nor left.

ZakaryH:
done reviewing the code. the only outstanding topic is the `strategy` bit. if we wanna be safe and finish this sooner, stickign with `absolute` is the way to go.

if we are willing to take the time to understand if `fixed` is actually better for us (it solves a bug we are actively trying to fix, and it does seem like having Popovers respond to scrolling would be good in general) then that's the other option.

or finally, we can do what we had been planning to do for a couple weeks now and make `strategy` a prop that people can opt into for the fixed behavior when needed. probably the best compromise between the two and why we were considering doing it that way to begin with 😄 

gonna spin up a prerelease build and check out some Popover instances in product to see if anything interesting happens with this version!

edison-cy-yang:
> nedone reviewing the code. the only outstanding topic is the `strategy` bit. if we wanna be safe and finish this sooner, stickign with `absolute` is the way to go.
> 
> if we are willing to take the time to understand if `fixed` is actually better for us (it solves a bug we are actively trying to fix, and it does seem like having Popovers respond to scrolling would be good in general) then that's the other option.
> 
> or finally, we can do what we had been planning to do for a couple weeks now and make `strategy` a prop that people can opt into for the fixed behavior when needed. probably the best compromise between the two and why we were considering doing it that way to begin with 😄
> 
> gonna spin up a prerelease build and check out some Popover instances in product to see if anything interesting happens with this version! nevermind, there is a prerelease version I just didn't see it at first.

I'd like to leave this one at absolute for now, then implement optional strategy prop next, since they're for different issues.

ZakaryH:
interesting! trying out the flow that was having issues on popper it's actually working as expected now even with `absolute`
Kapture 2025-07-10 at 10 30 53

tried this on Firefox
image

Safari

image


Chrome
image

so that's great that's working, we might not need the strategy prop at all!

**Issues**

though there are a couple things we should look into

first, why is the Safari Popover showing the reverse theme? I clearly have the light theme on for all of these.

second, the arrow goes just a bit too far and exceeds the area where it should be. I wonder if that's something we can address.

ZakaryH:
Kapture 2025-07-10 at 13 43 12
this is a bit janky

I'll have to see how that behaves vs the current popper version though. maybe it's the same.

ZakaryH:
Safari has some disturbing behavior if you use the trackpad specifically to keep scrolling past the available area. using the scrollbars and pageUp pageDown didn't do this.

Kapture 2025-07-10 at 15 43 03

probably not a blocker though. it's one browser doing something strange when you are doing something equally strange.

ZakaryH:
alright I'm pretty satisfied with the QA at this point. I've tested out a handful of different Popovers and everything is looking good to me.

the only thing worth us looking into before merging is the arrow positioning when it reaches the outer edges of the Popover

image

edison-cy-yang:
> alright I'm pretty satisfied with the QA at this point. I've tested out a handful of different Popovers and everything is looking good to me.
> 
> the only thing worth us looking into before merging is the arrow positioning when it reaches the outer edges of the Popover
> 
> image

@ZakaryH Turns out the padding on arrow did have an effect. I added it back and it serves as the padding between the arrow and the edge of the floating element. https://github.com/GetJobber/atlantis/pull/2610/commits/9d893fb5d09e039ccefdf309c302a9d6d42b522b
**Before**
Screenshot 2025-07-11 at 9 57 30 AM
**After**
Screenshot 2025-07-11 at 9 57 07 AM


Also removed the story https://github.com/GetJobber/atlantis/pull/2610/commits/9d893fb5d09e039ccefdf309c302a9d6d42b522b

ZakaryH (approved):
nice, that padding makes it line up perfectly

image


LGTM

PR #2611, opened 2025-07-04
feat(components): Add UNSAFE Props to Menu

ZakaryH:
## Motivations

we have a need to specify a max-height* on the Menu to allow scrolling. right now since the content is portalled out to the body, even CSS "drilling" won't work as a solution.

we might want to solve the problem of having various max-heights to trigger scrolling in a different way down the road, so that we don't have to rely on UNSAFE props, but even if we do that it would still be good to have these available for other needs that we can't or choose not to meet.

*to be clear we do have a max-height already. the issue is that the value for it is `72vh`. that's quite a bit of space. until we hit that point the Menu will keep growing, and if there are enough items inside the Menu Popper/Floating-UI won't be able to do anything about it even if it tries to position it above or to the sides so we'll have a very tall Menu that is outside the viewport, obstructing whatever menu items couldn't fit in the available space.

### Added

added UNSAFE props to Menu as per our established pattern.

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

give some UNSAFE props to a Menu, see that they apply.

if you want to properly test out this solution with the problem it seeks to resolve
- create a Menu with many items, let's say 12ish
- place your Menu within a Modal, so it's in the middle of a screen, having less vertical space to occupy
at this point you'll likely notice that not all your items can display (pending your screen size of course)
- to remedy this, apply an UNSAFE classname or style that applies to the menu, and give it an appropriate max-height, triggering the scroll
- you should now be able to access all the items in your Menu :^)

Changes can be

---

### DIFF
--- a/packages/components/src/Menu/Menu.tsx
+++ b/packages/components/src/Menu/Menu.tsx
@@ -1,4 +1,5 @@
 import React, {
+  CSSProperties,
   MouseEvent,
   ReactElement,
   RefObject,
@@ -47,6 +48,28 @@ export interface MenuProps {
    * Collection of action items.
    */
   readonly items: SectionProps[];
+
+  /**
+   * **Use at your own risk:** Custom class names for specific elements. This should only be used as a
+   * **last resort**. Using this may result in unexpected side effects.
+   * More information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).
+   */
+  readonly UNSAFE_className?: {
+    menu?: string;
+    header?: string;
+    action?: string;
+  };
+
+  /**
+   * **Use at your own risk:** Custom style for specific elements. This should only be used as a
+   * **last resort**. Using this may result in unexpected side effects.
+   * More information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).
+   */
+  readonly UNSAFE_style?: {
+    menu?: CSSProperties;
+    header?: CSSProperties;
+    action?: CSSProperties;
+  };
 }
 
 export interface SectionProps {
@@ -62,7 +85,12 @@ export interface SectionProps {
 }
 
 // eslint-disable-next-line max-statements
-export function Menu({ activator, items }: MenuProps) {
+export function Menu({
+  activator,
+  items,
+  UNSAFE_className,
+  UNSAFE_style,
+}: MenuProps) {
   const [visible, setVisible] = useState(false);
   const popperRef = useRef<HTMLDivElement>(null);
 
@@ -160,7 +188,7 @@ export function Menu({ activator, items }: MenuProps) {
               >
                 {items.length > 0 && (
                   <motion.div
-                    className={styles.menu}
+                    className={classnames(styles.menu, UNSAFE_className?.menu)}
                     role="menu"
                     data-elevation={"elevated"}
                     aria-labelledby={buttonID}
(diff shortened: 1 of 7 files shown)


packages/components/src/Menu/Menu.tsx
```diff
@@ -47,6 +48,28 @@ export interface MenuProps {
...
   readonly items: SectionProps[];
+
+  /**
+   * **Use at your own risk:** Custom class names for specific elements. This should only be used as a
+   * **last resort**. Using this may result in unexpected side effects.
+   * More information in the [Customizing components Guide](https://atlantis.getjobber.com/guides/customizing-components).
+   */
+  readonly UNSAFE_className?: {
+    menu?: string;
+    header?: string;
```
edison-cy-yang, line 59: to me, header here is really applying to the header container itself, instead of the Typography. I wonder if we can have the UNSAFE structure be 
```
{
    menu?: string;
    headerContainer?: {
         container?: string;
         header?: Typography["UNSAFE_className"];
    };
    action?: string;
}
```

Similar pattern can be applied to `action` as well since it does use `Icon` and `Typography`.

packages/site/src/content/Menu/MenuNotes.mdx
```diff
@@ -0,0 +1,26 @@
...
+[here](/guides/customizing-components).
+
+**Note**: Use of `UNSAFE_` props is **at your own risk** and should be
+considered a **last resort**. Future Menu updates may lead to unintended
+breakages.
+
+Menu has multiple elements that can be targeted with classes or styles:
+
+- `menu`: The actual menu element containing the actions
+- `section`: The container for each section
```
edison-cy-yang, line 15: This does not match the actual UNSAFE prop field name `header`

ZakaryH: good catch. I went back and forth on where to place the style and had it on the section at one point.

edison-cy-yang (commented):
Tested the storybook example and confirmed UNSAFE props are applied properly. 

The question on the structure of UNSAFE props is up for debate, considering `menu` seems to be the important prop for now.

packages/components/src/Menu/Menu.tsx
taylorvnoj: [This doc can be updated](https://jobber.atlassian.net/wiki/spaces/teamuxf/pages/3499885172/UNSAFE+props) - I should have done that after updating `Disclosure` which is a good example of bubbling up `UNSAFE_` props from `Typography` and `Icon`. The wrapper can then have it's own type defined in `Menu`.

Which leads me to my question - are we leaving out `container` and `overlay`, etc for a specific reason? For consistency we should have a `container` type but I can understand that maybe it's because we are only targeting what we need for the request?

imo `UNSAFE_` should be on as many elements if not all elements of a component because `UNSAFE_` should be a full escape hatch

Screenshot 2025-07-04 at 5 15 03 PM

edison-cy-yang: Button also does apply UNSAFE to ButtonLabel (Typography) and ButtonIcon (Icon) directly. But in those cases they don't have a wrapping container.

taylorvnoj: That's fair, someone can just implement the custom activator. 

Re: overlay, what if there's a want for a colored overlay like in my earlier screenshot?
I hear what you're saying and I do think that every component will be different. I just want to make sure we aren't making assumptions about what people may or may not need as far as `UNSAFE_` goes because then we're still coming from a place of being opinionated.
I'm happy to leave it as is, the great thing is that we can always add a type later on if the need presents itself.

edison-cy-yang (approved):

PR #2612, opened 2025-07-07
feat(components): Upgrading to Storybook9

scotttjob:
## Motivations

1. We need to upgrade to React 19, and part of that means upgrading Storybook.
2. Storybook 6 -> 7 was a nightmare, so we're skipping that process the second time and we're jumping directly to new instances of Storybook, in version 9 specifically.

## Changes

1. Added basic storybook9 instance to components
2. Added basic storybook9 instance to components-native
3. Moved Button Stories into each folder
4. Moved Button Docs into Docs site
5. Added proper `type` usage for items in use in new Storybook for vite compatibility.

## Testing

1. This is essentially spike code that I tossed over the fence. I imagine we'll need to test this well before we launch.

Changes can be

---


nad182 (commented):
Here are some things that I noticed after the first review pass:
- I believe we need to update story titles, by removing `Web`/`Mobile` from the "path": e.g. in the [Button.stories.tsx](https://github.com/GetJobber/atlantis/pull/2612/files#diff-b076422bfbbbc7c8b0b2c22cd8e9cea82141b607ef91c20352ac4b28ec5f7059R18), in `/components/`, we should just have `"Components/Actions/Button"`, instead of `"Components/Actions/Button/Web"` for a simpler structure/tree;
  - image

- We should avoid using `alert()` and `console.log()` in the stories, in favour of using [actions](https://storybook.js.org/docs/essentials/actions), which help us log not only the execution of event handlers, but also their params and number of executions (which will be seen in the "Actions" tab);
  - image

- Since we're setting up everything almost from scratch, it would be great to have some basic initial setup for:
  - Accessibility testing with [addon-a11y](https://storybook.js.org/docs/writing-tests/accessibility-testing#accessibility-checks-with-a11y-addon)
  - [play functions](https://storybook.js.org/docs/writing-stories/play-function), so that when people look at this first component as an example, they will follow suit and create those for other migrated components (in the future)
- It might be how your IDE is set up, but I noticed that in all of the new files the line breaks at the end of the files were missing.

If these sound good to you, I could make a commit to this existing branch, or branch off and submit a PR into this current branch, or if you'd like to tackle those by yourself, just let me know!

I don't think we need to go crazy in this initial PR, but setting some basic standards might be very useful for future migration work.

packages/components-native/src/Button/Button.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import type { Meta, StoryObj } from "@storybook/react-native";
+import { action } from "@storybook/addon-actions";
+import { Button } from "./Button";
+
+const meta: Meta<typeof Button> = {
+  title: "Components/Actions/Button/Mobile",
```
nad182, line 6: This is what I meant by removing "Web" or "Mobile" from the story title/path:

```suggestion
  title: "Components/Actions/Button",
```

packages/components/src/Button/Button.stories.tsx
```diff
@@ -23,21 +23,23 @@ export default {
...
-
-export const Basic = BasicTemplate.bind({});
-Basic.args = {
-  label: "New Job",
-  onClick: () => alert("👍"),
+export const Basic: StoryObj<typeof Button> = {
+  args: {
+    label: "New Job",
+    type: "primary",
+    onClick: () => alert("👍"),
```
nad182, line 38: This is what I meant by using "actions", instead of `alert()`:

```suggestion
import { fn } from "storybook/test";

    onClick: fn(),
```

PR #2614, opened 2025-07-09
feat(components): Autocomplete V2

scotttjob:
## Motivations

1. Needed a v2 automplete, along with the ability to trigger a 'setOptions' from the menu.

## Changes

1. Added support for v2 autocomplete, and the ability to trigger a 'setOptions' from the composed menu.

## What to look out for

1. `updateOptions` doesn't expect a string like getOptions does, it expects an arbitrary object which can contain just about anything. This allows a consumer to send specific requests up from the menu. This allows instructing a parent component to load a different list for instance to query the existing options against. Triggering 'updateOptions' is typically done in a header/footer within the Autocomplete.

## Testing

1. Everything should still work as-is with no changes across all apps.
2. Adding `version={2}` to an Autocomplete should now not render a react-hook-form controller.
3. There is a new `updateOptions` function prop that can be passed to Autocomplete which will be accessible via the menu.

Changes can be

---


packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -63,19 +65,34 @@ function AutocompleteInternal<
...
+          value: inputText,
+          onChange: handleInputChange,
+          validations,
+        };
+
   return (
     <div className={styles.autocomplete} ref={autocompleteRef}>
       <InputText
         ref={mergeRefs([ref, inputRef])}
         autocomplete={false}
```
ZakaryH, line 90: I think we've gotta prioritize getting InputText v2 to stop extending `React.InputHTMLAttributes<HTMLInputElement | HTMLTextAreaElement>` sooner than later. it's causing a lot of errors that Typescript can't catch.

using `version={2}` on an Autocomplete causes it to immediately get the error style with this in the console
image

there's 2 problems happening at the same time.

in Autocomplete we do

```
      <InputText
        ref={mergeRefs([ref, inputRef])}
        autocomplete={false}
        size={size}
        placeholder={placeholder}
        onFocus={handleInputFocus}
        onBlur={handleInputBlur}
        {...inputTextProps}
```

**Version 1**
on v1, it's using `FormFieldProps` which has

`readonly autocomplete?: boolean | AutocompleteTypes;`

so this is fine, and works.

**Version 2**

on v2, InputText it is both extending `InputHTMLAttributes` which is going to provide

```
   interface InputHTMLAttributes<T> extends HTMLAttributes<T> {
        autoComplete?: HTMLInputAutoCompleteAttribute | undefined;
```

where boolean is not an acceptable value

but v2 also has `readonly autocomplete?: boolean | AutocompleteTypes;`

meaning a v2 InputText will not complain about the following `<InputText version={2} autocomplete={false} autoComplete={"off"} ...>` as far as Typescript goes anyway.

**Result**

because in Autocomplete we are passing `autocomplete={false}`, inside InputText.rebuilt due to a bunch of spreading props Typescript gets confused and basically stops working, we will eventually do 

```
  return (
    <input
      {...fieldProps}
      ref={mergeRefs(inputRefs)}
      className={inputStyle}
      value={value}
    />
```

where `fieldProps` contains `autocomplete: false` - even though the types will tell you that doesn't exist

image

image

meaning we're spreading an invalid value onto `input` that expects `autoComplete: "off"` or something like that

😵

ZakaryH: to get around this we'll have to do something like

```
        {...(version === 1 && { autocomplete: false })}
        {...(version === 2 && { autoComplete: "off" })}
```

in `Autocomplete.tsx`

though even with that there's something still triggering the error state on the field. I'll dig into that next.

ZakaryH:
as an aside, the lines of code changed in this PR is really big. it's all coming from the package-lock where it looks like `react-aria-components` was added to `site`

it doesn't seem like we need that, right?

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -63,19 +65,34 @@ function AutocompleteInternal<
...
 
+  const inputTextProps =
+    version === 2
+      ? {
+          ...inputProps,
+          version: 2 as const,
+          value: inputText,
+          onChange: handleInputChange,
+          error: validations ? "" : undefined,
+          invalid: Boolean(validations),
```
ZakaryH, line 76: ah this is what is causing the immediate red border on the field!

if you don't have `validations` then of course `undefined` becomes `false`

if you do though, yeah we get `invalid: true` despite us not having run any of the validations, resulting it the red error state

actually.... hold on, what are we doing with `validations` on v2? I only see them included in the props for the v1.

we'll either want to fully remove those from v2's props if they don't do anything, or we need to implement them.

ZakaryH: unless I'm missing something, I think we have to remove `validations` from v2

the whole way that works with v1 is that you pass them to FormField who registers that field with a controller + `rules`

since v2 intentionally doesn't have that, we'd have to build out a different solution OR move that responsibility up to the consumer where the RHF instance will likely be

that would mean we need to expose `error`/`invalid` as `Autocomplete` props for someone to determine when to set on their own, and we need to provide an `onChange` that a consumer can listen to to validate the input.

the complication with that is `onChange` refers to the Autcomplete's selection changing, not typing in the input. it's only called on input change if `allowFreeForm` is true.

so we'd need to add another event callback to hook into. we could do like `react-aria` and wait until blur rather than on each keystroke, but that also needs a new prop of `onInputBlur`

packages/components/src/Autocomplete/Autocomplete.types.ts
```diff
@@ -88,6 +88,16 @@ export interface AutocompleteProps<
     | Array<GenericGetOptionsValue | GenericOption>
     | Promise<Array<GenericGetOptionsValue | GenericOption>>;
 
+  /**
+   * Called by advanced usage scenarios to update the options.
+   * @param newInputText
+   */
+  updateOptions?(
```
ZakaryH, line 95: I wonder if there's a way for us to have our types only include this if you are doing the custom render?

if not, we should communicate this only does something in the `customRenderMenu` flow and won't do anything at all for the default flow

I'd even be tempted to change the name too

`getOptions` and updateOptions` are so similar in name, and if you look behind the curtain even their implementations are almost identical

```
    const updatedOptions = await getOptions(inputText);
    const filteredOptions = updatedOptions.filter(option =>
      isOptionGroup(option) ? option.options.length > 0 : true,
    );

    setOptions(mapToOptions(filteredOptions));
  }
```

where we also call `setOptions` on the result of that callback

ZakaryH:
if you change the standard options, it works as I'd expect. however if you change the group section names at all, it causes the menu to close. the values will have been replaced though.

Kapture 2025-07-09 at 14 22 30

ZakaryH:
another couple scenarios that are behaving unexpectedly related to the above

if I keep the headers the same, but I click the footer button (both footer buttons have the same onclick calling handleUpdateOptions) in Invoices, it also closes the menu
Kapture 2025-07-09 at 14 24 52

then weridly, with that exact same data if I click the footer button in the Quotes section, it doesn't close 🤔 
Kapture 2025-07-09 at 14 27 37

packages/components/src/Autocomplete/Autocomplete.tsx
ZakaryH: one more thing to be mindful of when swapping this to v2, decoupled from RHF, is that the `name` prop on `Autocomplete` gets associated with the `InputText`

when `allowFreeForm` is `false` the flow is a bit different. the content being typed, that we might be passing back via an `onInputChange` isn't "real", it's only a search term that will get cleared on blur. meaning one should not perform any validation on it, you really might as well ignore it which is what our current `handleInputChange` does.

that's more of an implementation detail out at the consumer level, but still something that people will need to watch out for.

scotttjob: I'll admit this prop flip flop was Cursor's solution to a typing issue. I didn't see anything wrong at first glance because the linter was happy but I'm probably juggling too many things at once.

We can chat briefly on our coffee chat today, maybe there's some quick solves here.

packages/components/src/Autocomplete/Autocomplete.tsx
```diff
@@ -63,19 +65,34 @@ function AutocompleteInternal<
...
   }, [value]);
 
+  const inputTextProps =
+    version === 2
+      ? {
+          ...inputProps,
+          version: 2 as const,
+          value: inputText,
+          onChange: handleInputChange,
+          error: validations ? "" : undefined,
```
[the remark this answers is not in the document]
scotttjob: See my comment above. This was a last minute cursor solve to replace a cast I was doing, so there's probably room for improvement.

PR #2615, opened 2025-07-09
feat(components): add onSelectAll, onClear and onClick callbacks to Combobox

rjdestigter:
## Motivations

The `Combobox` component lacks callbacks for certain user interactions, forcing consumers to implement workarounds to react to events. This change aims to provide more control and flexibility by exposing callbacks for common actions. This will simplify implementations for consumers, allowing them to easily hook into the component's lifecycle to trigger side effects like analytics, data fetching, or other state updates.

### Added

-   `onSelectAll: (selection: ComboboxOption[]) => void` callback to `Combobox`. This is triggered when the "Select All" action is clicked in a multi-select `Combobox`.
-   `onClear: () => void` callback to `Combobox`. This is triggered when the "Clear" action is clicked.
-   `onClick: (option: ComboboxOption) => void` callback to `Combobox.Option`. This allows for attaching custom behavior when an individual option is clicked, separate from the selection event.

## Testing

Changes can be
[tested via Pre-release](https://github.com/GetJobber/atlantis/actions/workflows/trigger-qa-build.yml).

To test manually:

The story specific to this was removed, since the JSDoc is pretty self explanatory already. For testing purposes, you may refer to the commit that removed the story https://github.com/GetJobber/atlantis/pull/2615/commits/447b726f6a987a1ada4f03c5ab2c4fe091180760 to add the story back. 

1.  Navigate to the `Combobox` stories in Storybook.
2.  In a multi-select `Combobox`, verify that the `onSelectAll` callback is fired when you click "Select All".
3.  Verify that the `onClear` callback is fired when you click "Clear".
4.  Verify that the `onClick` callback on a `Combobox.Option` is fired when an option is clicked.

---

### DIFF
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -615,9 +615,6 @@ describe("ComboboxOption onClick", () => {
     expect(handleOptionClick).toHaveBeenCalledWith({
       id: "1",
       label: "Bilbo Baggins",
-      prefix: undefined,
-      customRender: undefined,
-      onClick: handleOptionClick,
     });
   });
 });
--- a/packages/components/src/Combobox/Combobox.types.ts
+++ b/packages/components/src/Combobox/Combobox.types.ts
@@ -27,13 +27,13 @@ export interface ComboboxProps {
   /**
    * Callback function invoked upon the selection of an option. Provides the selected option(s) as an argument.
    */
-  readonly onSelect: (selection: ComboboxOption[]) => void;
+  readonly onSelect: (selection: ComboboxSelection[]) => void;
 
   /**
    * Callback function invoked upon the selection of all options. Provides the selected option(s) as an argument.
    * This is only available when `multiSelect` is `true`.
    */
-  readonly onSelectAll?: (selection: ComboboxOption[]) => void;
+  readonly onSelectAll?: (selection: ComboboxSelection[]) => void;
 
   /**
    * Callback function invoked upon the clearing of all options.
@@ -155,6 +155,8 @@ export interface ComboboxOptionProps {
 }
 
 export type ComboboxOption = ComboboxOptionProps;
+
+export type ComboboxSelection = Pick<ComboboxOptionProps, "id" | "label">;
 
 export interface ComboboxContentProps {
   /**
--- a/packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.tsx
+++ b/packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.tsx
@@ -14,6 +14,12 @@ export function ComboboxOption(props: ComboboxOptionProps) {
     selection => selection.id.toString() === props.id.toString(),
   );
 
+  const handleClick = () => {
+    const { id, label } = props;
+    selectionHandler?.({ id, label });
+    onClick?.({ id, label });
+  };
+
   return (
     <li
       key={props.id}
@@ -21,10 +27,7 @@ export function ComboboxOption(props: ComboboxOptionProps) {
       data-selected={isSelected}
       role="option"
       aria-selected={isSelected}
-      onClick={() => {
-        selectionHandler?.({ id: props.id, label: props.label });
-        onClick?.(props);
-      }}
+      onClick={handleClick}
       className={classnames(styles.option)}
     >
       {customRender ? (
--- a/packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.test.tsx
+++ b/packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.test.tsx
@@ -223,9 +223,6 @@ describe("ComboboxOption", () => {
       expect(handleClick).toHaveBeenCalledWith({
         id: "1",
         label: "Michael",
-        prefix: undefined,
-        customRender: undefined,
-        onClick: handleClick,
       });
     });
   });
--- a/packages/components/src/Combobox/index.ts
+++ b/packages/components/src/Combobox/index.ts
@@ -1,3 +1,7 @@
 export * from "./Combobox";
 export { ComboboxContextProvider } from "./ComboboxProvider";
-export { ComboboxOption, ComboboxCustomActivatorProps } from "./Combobox.types";
+export {
+  ComboboxOption,
+  ComboboxCustomActivatorProps,
+  ComboboxSelection,
+} from "./Combobox.types";
--- a/packages/site/src/content/Combobox/Combobox.props.json
+++ b/packages/site/src/content/Combobox/Combobox.props.json
@@ -57,7 +57,7 @@
         },
         "required": true,
         "type": {
-          "name": "(selection: ComboboxOptionProps[]) => void"
+          "name": "(selection: ComboboxSelection[]) => void"
         }
       },
       "onSelectAll": {
@@ -70,7 +70,7 @@
         },
         "required": false,
         "type": {
-          "name": "(selection: ComboboxOptionProps[]) => void"
+          "name": "(selection: ComboboxSelection[]) => void"
         }
       },
       "onClear": {


edison-cy-yang (commented):
@rjdestigter Nice work getting this PR started! A few things I changed:
- Some cleanups around documentation
- Changing `onSelect` back to required. You're not actually using them.
- Move `onSelectAll` and `onClear` into ComboboxProvider to avoid props drilling
- Remove some redundant tests in Combobox.test and Combobox.test, leaving the ones testing the new callback props
- Cleanup Storybook example to just console log each handler
- Rename `onClick` to `onOptionClick` to better reflect the context

docs/components/Combobox/Web.stories.tsx
```diff
@@ -915,6 +915,68 @@ const ComboboxCustomRenderOptions: ComponentStory<typeof Combobox> = args => {
   );
 };
 
+const ComboboxCallbackProps: ComponentStory<typeof Combobox> = args => {
```
ZakaryH, line 918: not sure about this story. we have a lot of Combobox stories 😅 

image

there's no instructions about what to expect, so it doesn't really look like anything is happening until you open the dev tools.

even then, I'm not sure this is worthy of needing further explanation tbh. the JSDocs on the props already do a pretty good job describing these callbacks.

ZakaryH: I'm leaning towards no story for these props. open to discussion tho!

packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -135,6 +145,11 @@ export interface ComboboxOptionProps {
       isSelected: boolean;
     },
   ) => React.ReactNode;
+
+  /**
+   * Callback function invoked when the option is clicked.
+   */
+  readonly onOptionClick?: () => void;
```
ZakaryH, line 152: sorry ignore my other comment. I was reading the diff wrong and though this was somewhere else!

my remaining feedback is that `onOptionClick` feels redundant to me. we're going to have

```
<Combobox ...>
  <Combobox.Option onOptionClick={() => ...} />
</Combobox>
```

I'd argue it's obvious that it belongs to the `Option`

`onClick` should suffice.

packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -29,6 +29,16 @@ export interface ComboboxProps {
...
   readonly onSelect: (selection: ComboboxOption[]) => void;
 
+  /**
+   * Callback function invoked upon the selection of all options. Provides the selected option(s) as an argument.
+   */
+  readonly onSelectAll?: (selection: ComboboxOption[]) => void;
+
+  /**
+   * Callback function invoked upon the clearing of all options.
+   */
```
ZakaryH, line 39: for both of these, might be worth mentioning their dependency on `multiselect` you simply won't have a select-all/clear button for single-select

packages/components/src/Combobox/Combobox.tsx
```diff
@@ -21,6 +21,7 @@ export function Combobox(props: ComboboxProps): JSX.Element {
         label: option.props.label,
         prefix: option.props.prefix,
         customRender: option.props.customRender,
+        onClick: option.props.onClick,
```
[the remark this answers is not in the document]
ZakaryH: unless I'm missing something it should be as simple as 

```
 triggerElement?: ReactElement<ComboboxActivatorProps>;
  optionElements?: ReactElement<ComboboxOptionProps>[];
  actionElements?: ReactElement<ComboboxActionProps>[];
```

in `useComboboxValidation.ts`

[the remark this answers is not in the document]
ZakaryH: lastly, so far there's nothing that a `Combobox.Option` receives as a prop that we don't want this pseudo-copied data version of the component to have. in fact the whole point is to preserve their props.

I know it was already like this but what do you think about just spreading `option.props`?

```
  const options = useMemo(
    () =>
      optionElements?.map(option => ({
        ...option.props,
      })) || [],
    [optionElements],
  );

```

I've been adverse to spreading props lately because of some issues, but in this case I think if we tighten up the types returned from our validation hook, it'll be fine.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -10,10 +10,12 @@ import { useComboboxContent } from "../../hooks/useComboboxContent";
 import { useComboboxAccessibility } from "../../hooks/useComboboxAccessibility";
 import { ComboboxContentProps } from "../../Combobox.types";
 import { COMBOBOX_MENU_ID } from "../../constants";
+import { ComboboxContext } from "../../ComboboxProvider";
 
 export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
   const optionsExist = props.options.length > 0;
   const { optionsListRef } = useComboboxContent(props.open, props.selected);
+  const { onClear, onSelectAll } = useContext(ComboboxContext);
```
ZakaryH, line 18: so `useComboboxContent` is already a hook made specifically for the `ComboboxContent.tsx` to clean up and group the things it cares about.

inside there you'll see that we're already doing `const { shouldScroll } = useContext(ComboboxContext);`

so let's avoid having duplicate calls to the same Context.

instead, we can do

```
const { onClear, onSelectAll, shouldScroll } = useContext(ComboboxContext);
....
  return {
    onClear,
    onSelectAll,
    optionsListRef,
  };
```

then in ComboboxContent 

`const { onClear, onSelectAll, optionsListRef } = useComboboxContent(...)`

packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.tsx
```diff
@@ -21,10 +21,11 @@ export function ComboboxOption(props: ComboboxOptionProps) {
       data-selected={isSelected}
       role="option"
       aria-selected={isSelected}
-      onClick={() =>
+      onClick={() => {
         selectionHandler &&
-        selectionHandler({ id: props.id, label: props.label })
-      }
+          selectionHandler({ id: props.id, label: props.label });
```
ZakaryH, line 26: I like the more elegant `?.()` syntax

wanna swap the `selectionHandler?.(` to that too?

ZakaryH (changes requested):
biggest issue is that the `onClick` doesn't work 😅 and none of our tests told us that

just gotta fix the names (strong preference for `onClick`) and ideally tighten up the types!

packages/components/src/Combobox/Combobox.types.ts
ZakaryH: one other thing, we might as well pass the clicked item as an arg. there's no downside to getting that extra information. if you don't want to use it, just ignore it.

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -367,6 +367,65 @@ describe("Combobox Multiselect", () => {
...
+    beforeEach(() => {
+      handleSelectAll.mockClear();
+      handleClear.mockClear();
+    });
+
+    it("should call onSelectAll when clicking Select all", async () => {
+      render(
+        <Combobox
+          label={activatorLabel}
+          multiSelect={true}
```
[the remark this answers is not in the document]
edison-cy-yang: They were already in the multiselect test suite, just hidden in another layer of describe onClearAll and onSelectAll, which I think is redundant. I took out that extra layer https://github.com/GetJobber/atlantis/pull/2615/commits/35a215cbe754227ed357caf8d3fdea9719155a16

docs/components/Combobox/Web.stories.tsx
edison-cy-yang: Totally, I took out a big chunk of the story but still felt like they were targeting a very specific use case. 
I will leave it in for testing purposes until the end.

packages/components/src/Combobox/Combobox.tsx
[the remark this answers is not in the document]
edison-cy-yang: I forgot to change that onClick when I renamed it to onOptionClick earlier. Since it's renamed back to onClick this one will be fine.

Tighten up the typing of `useComboboxValidation` and spread option.props https://github.com/GetJobber/atlantis/pull/2615/commits/ce1586b597a702fa9ca1193bc687ba83fac836d9

Personally spreading here makes total sense to me. We have issues in other places because they're spread multiple layers down and some of the data were touched in the middle. We have straight mapping here.

packages/components/src/Combobox/Combobox.types.ts
edison-cy-yang: Good call, updated here https://github.com/GetJobber/atlantis/pull/2615/commits/9297bef9d04edd44210d324652ad50a4b34a9602

packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.tsx
```diff
@@ -24,7 +24,7 @@ export function ComboboxOption(props: ComboboxOptionProps) {
       onClick={() => {
         selectionHandler &&
           selectionHandler({ id: props.id, label: props.label });
-        onClick?.();
+        onClick?.(props);
```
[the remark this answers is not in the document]
edison-cy-yang: going to quickly spin up a pre release build to see how many changes are needed in JFE and JO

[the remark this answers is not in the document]
edison-cy-yang: Checked both JFE and JO, most of them will require just type changes (although I would have to go through each ComboboxOption usage as not all of them are used on the onSelect callback), except one typescript error in JO on `BillingTypeFilter` which is caused by incorrect implementation in the beginning. I can get them updated as part of bump process

PR #2617, opened 2025-07-16
feat(components): Add stretch type to Cluster align prop

taylorvnoj:
## Motivations

A use case has come up, that requires a vertical `Divider` to be used in a `Cluster`. However vertical `Divider` has auto height and if there is no height defined in the parent, the `Divider` will not be visible.

Screenshot 2025-07-16 at 6 08 00 PM

### Added

- `stretch` type to the `align` prop
  - gets mapped to the CSS align-items property via the CSS custom property `--public-cluster-align`

## Testing

I did not add a story as I don't think there needs to be one, but you can add 
```js
export const WithDivider = BasicTemplate.bind({});
WithDivider.args = {
  align: "stretch",
  children: (
    <>
      <Button label="Left Side" />
      <Divider direction="vertical" />
      <Button label="Right Side" type="secondary" />
    </>
  ),
};
```
to `docs/components/Cluster/Web.stories.tsx` to view

Changes can be

---

### DIFF
--- a/packages/site/src/content/Cluster/Cluster.props.json
+++ b/packages/site/src/content/Cluster/Cluster.props.json
@@ -29,7 +29,7 @@
         },
         "required": false,
         "type": {
-          "name": "\"start\" | \"end\" | \"center\""
+          "name": "\"start\" | \"end\" | \"center\" | \"stretch\""
         }
       },
       "gap": {


edison-cy-yang (approved):
Tested and confirmed stretch align works nicely with vertical Divider!

PR #2622, opened 2025-07-18
fix(components): Undo ComboboxSelection type for callback props

edison-cy-yang:
## Motivations

We deliberately chose to create a new ComboboxSelection type in this [PR](https://github.com/GetJobber/atlantis/pull/2615#discussion_r2205377698) in order to better reflect the combobox options being passed back to the consumers via callbacks (only `id` and `label` of ComboboxOption). However, digging deeper I found these callbacks (onSelect) to be passed through layers of hooks and internal components. Going through each usage and changing the type quickly becomes messy, although all CI jobs are passing. We will have to come back and tackle a full cleanup in a separate ticket. This PR will undo the addition of ComboboxSelection type and also only return `id` and `label` in each option of `onSelectAll`

### Added

- 

### Changed

- Undo the addition of ComboboxSelection and revert `onSelect` and `onSelectAll` type to `ComboboxOption[]`
- `onSelectAll` now only passes back `id` and `label` of each option

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

- Add `onSelectAll` callback prop to `ComboboxMultiSelection` story and confirm options only contain id and label

Changes can be

---

### DIFF
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -371,7 +371,9 @@ describe("Combobox Multiselect", () => {
     expect(handleSelect).toHaveBeenCalledWith([]);
   });
 
-  it("should call onSelectAll when clicking Select all", async () => {
+  it("should call onSelectAll with options with only id and label when clicking Select all", async () => {
+    const mockOnClick = jest.fn();
+
     render(
       <Combobox
         label={activatorLabel}
@@ -381,8 +383,8 @@ describe("Combobox Multiselect", () => {
         onSelectAll={handleSelectAll}
         onClear={handleClear}
       >
-        <Combobox.Option id="1" label="Bilbo Baggins" />
-        <Combobox.Option id="2" label="Frodo Baggins" />
+        <Combobox.Option id="1" label="Bilbo Baggins" onClick={mockOnClick} />
+        <Combobox.Option id="2" label="Frodo Baggins" onClick={mockOnClick} />
       </Combobox>,
     );
 
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -372,6 +372,7 @@ describe("Combobox Multiselect", () => {
   });
 
   it("should call onSelectAll with options with only id and label when clicking Select all", async () => {
+    // Not directly using mockOnClick, it is used to confirm its absence in the onSelectAll callback argument
     const mockOnClick = jest.fn();
 
     render(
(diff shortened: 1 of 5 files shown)


packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
```diff
@@ -27,6 +27,15 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
       props.wrapperRef,
     );
 
+  // options that are passed back to consumers via onSelectAll callback
+  // should only contain id and label
+  const consumerOptions = useMemo(() => {
```
edison-cy-yang, line 32: I feel weird about the naming of this, open to suggestion

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -371,7 +371,9 @@ describe("Combobox Multiselect", () => {
     expect(handleSelect).toHaveBeenCalledWith([]);
   });
 
-  it("should call onSelectAll when clicking Select all", async () => {
+  it("should call onSelectAll with options with only id and label when clicking Select all", async () => {
+    const mockOnClick = jest.fn();
```
ZakaryH, line 375: do we need this mock? I only see it being passed to the options, never checked if it was called.

ZakaryH: right right right

that makes a ton of sense. it's just hard to understand the connection or relevance of the onClick to what the test is asserting. I have all that context and still didn't get it immediately 😅 

maybe a comment? something saying while this isn't called or used directly, we want to verify its absence in the arguments of the selectAll?

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH: hmm yeah even though we talk about consumers in conversation, I don't love having it as a concept in the code.

maybe `selectedOptionData`? `optionSelectionData`? `optionData` / `optionsData` if we don't necessarily want to pair it with the concept of being selected since technically that only happens in conjunction with the `onSelectAll` that hasn't happened yet.

packages/components/src/Combobox/Combobox.test.tsx
edison-cy-yang: Makes sense, test description only covered half of the context. Added comment.

ZakaryH:
- onClick working as expected
- onSelectAll too
- onClear yup

the types are back to being a bit silly, but we can fix that later. it's been like this since we added new props to Option.
image

ZakaryH (approved):
lol didn't mean to close it

meant to approve it!

PR #2624, opened 2025-07-21
feat(components): Open Up Page Title to Composition

ZakaryH:
### Added

- 

### Changed

Page's title prop now accepts the significantly less restrictive `ReactNode` type, which includes the previously supported `string` type

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

### DIFF
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -24,7 +24,7 @@ interface PageFoundationProps {
   /**
    * Title of the page.
    */
-  readonly title: string;
+  readonly title: ReactNode;
 
   /**
    * TitleMetaData component to be displayed


ZakaryH:
here's an example with the requested use case

image

ZakaryH:
right now, with my changes this actually works just fine and is semantically valid. that's somewhat out of luck though because the `Tooltip` element (a div) is portalled out to the body resulting in the h1 only containing our phrasing content.

if that div were rendered inside the h1, that's not correct and could possible result in a bad accessibility/screen reader experience

related, I see we already have `titleMetaData` that's extremely similar. the use case for that is going to result in invalid HTML sadly. a `StatusLabel` isn't phrasing content.

so we kinda need to sort out a few things
- where does `title` go when it's not a string? we can't guarantee it's type is correct to use within an h1
- do we check it? how does that work? do we support the simple case?
- what does that mean for a consumer? do they have to pass in a Heading level=1 every time?
- do we still need titleMetaData? that's not being validated
- do we still take care of layout concerns? titleMetaData gets that currently but I'm not sure if we can with arbitrary ReactNodes if we don't place it in an h1
- how does that work or look like if everyone is responsible for the layout and how do we ensure consistency?

for example

```
<Page
 title={<div style={styles.myCustomTitleLayout}><ToolTip message="lol"><Heading level={1}>"Cool"</Heading><Icon name="skull"/><Tooltip>}
...
/>
```

ZakaryH:
not needed

PR #2627, opened 2025-07-23
feat(components): Support any react node for Page title prop

jdeichert:
## Motivations

This PR updates our `Page`'s `title` prop to support any react node, instead of just strings. 

This allows the consumer to override the title and wrap it with extra behaviour and styling, such as tooltips for example.

## Before

`Page` only supported string titles.

## After

`Page` now supports any react node with the `title` prop. 

Code |  Demo
:---:|:---:
Screenshot 2025-07-23 at 12 05 22 PM | Screenshot 2025-07-23 at 12 05 26 PM

### Changed

- The Page `title` prop now supports any react node instead of being limited to strings

## Testing

1. Pull this branch, bootstrap and run storybook
2. Navigate to the new [Custom Title story](http://localhost:6005/?path=/story/components-layouts-and-structure-page-web--custom-title)
3. Hover on the page title and confirm the tooltip appears like below

Screenshot 2025-07-23 at 12 13 31 PM

---

### DIFF
--- a/packages/components/src/Page/Page.test.tsx
+++ b/packages/components/src/Page/Page.test.tsx
@@ -1,9 +1,10 @@
 import React from "react";
-import { fireEvent, render } from "@testing-library/react";
+import { fireEvent, render, screen } from "@testing-library/react";
 import { StatusLabel } from "@jobber/components/StatusLabel";
 import { Page } from ".";
 import { getActionProps } from "./Page";
 import { SectionProps } from "../Menu";
+import { Heading } from "../Heading";
 
 jest.mock("@jobber/hooks", () => {
   return {
@@ -247,5 +248,32 @@ describe("When titleMetaData is provided", () => {
 
     expect(getByText(pageTitle)).toBeDefined();
     expect(getByTestId(titleMetaDataTestId)).toBeDefined();
+  });
+});
+
+describe("when title is a React node", () => {
+  it("renders a Page using that node as the title", () => {
+    render(
+      <Page title={<Heading level={1}>Custom Heading Title</Heading>}>
+        Page content
+      </Page>,
+    );
+
+    expect(
+      screen.getByRole("heading", { level: 1, name: "Custom Heading Title" }),
+    ).toBeVisible();
+  });
+
+  it("does not render titleMetaData", () => {
+    render(
+      <Page
+        title={<Heading level={1}>Custom Heading Title</Heading>}
+        titleMetaData={<StatusLabel label={"Success"} status={"success"} />}
+      >
+        Page content
+      </Page>,
+    );
+
+    expect(screen.queryByText("Success")).not.toBeInTheDocument();
   });
 });


packages/components/src/Page/Page.tsx
ZakaryH, line 1: minor detail but it seems like XOR causes the jsdocs to be unavailable?

image

VS when I'm typing the prop name out, my IDE tells me what it is for. interestingly it is combining both descriptions it looks like. perhaps because at this point it can't tell which to use and ReactNode can also satisfy the "string" type?

image


as soon as I give the prop a value I lose the description

packages/components/src/Page/Page.tsx
```diff
@@ -23,14 +23,11 @@ interface PageFoundationProps {
 
   /**
    * Title of the page.
+   *
+   * Supports any React node. If a string is provided, it will be rendered as an H1 heading.
+   * Otherwise it will be rendered as is.
```
ZakaryH, line 28: the reason I mention the IDE JSDocs is because I was thinking it would be good to have a very strong suggestion here detailing that although yes you can technically pass in whatever you want for accessibility and semantic reasons a Page must have exactly one h1 element

technically you could satisfy this requirement elsewhere, with an h1 in some other place but part of the value of Page is that when you go from page A to B to C, they all look pretty familiar so I'd advise against doing that

packages/components/src/Page/Page.tsx
ZakaryH: the customRender feels like a bit much. ideally I'd like to keep the `title` prop.

if we make a separate group of props for the custom title scenario, TS is able to provide the JS docs when `titleMetaData` is present because it knows there's only one `title` definition that meets that criteria. that's a slight improvement, but doesn't help the case with the custom title unfortunately.

hmmm what if we say OK sure JSDocs don't quite work, fair enough. clicking into the definition would still tell you.

instead, to prevent this not containing an `h1` element we parse the children of `title` recursively to see if a `Heading level={1}` was provided and console.error if we didn't get one. you know what,  just typing that out I already don't love that idea haha.

ZakaryH: >Hmm, the more obvious solution is to do away with the XOR completely. Just allow titleMetaData and block it at the react level.

yeah that might be the way to go. leave a note saying it doesn't work with the custom markup title and that's it.

packages/components/src/Page/Page.test.tsx
edison-cy-yang, line 1: I dont think we have a test case for when `title` is a string with no `titleMetaData`. I'm just going off of the nested ternary expression you added 
```
{typeof title === "string" && titleMetaData ? (
                <div className={styles.titleRow}>
                  <Heading level={1}>{title}</Heading>
                  {titleMetaData}
                </div>
              ) : typeof title === "string" ? (
                <Heading level={1}>{title}</Heading>
              ) : (
                title
              )}
```

edison-cy-yang (commented):
One suggestion on a new test case. Otherwise looks good!

I know we're not going to loop through the children and check for existence of Heading level 1 when title is ReactNode, but I hope there's still a mechanism for us to catch the offences should there be any

packages/components/src/Page/Page.test.tsx
edison-cy-yang: ah yeah that's true

edison-cy-yang (approved):
Tested the component with string and ReactNode title. LGTM!

PR #2629, opened 2025-07-24
chore: Fix test:visual command

jdeichert:
## Motivations

It was failing because some of the build scripts require npm deps that are properly compiled for the docker container's linux environment. Your local node_modules directory contains both js scripts as well as pre-compiled architecture-specific dependencies for macOS. When you run `npm install` inside a linux environment, you end up getting pre-compiled deps for that architecture instead, and that's exactly what we needed here.

This is actually something I had in place back in May (as part of Scott's original PR) but I [reverted it](https://github.com/GetJobber/atlantis/pull/2566/commits/b2fb538ea4c591408f0572de9e8b97bed70e9f91) because it started working without the npm install step.

I think something possibly recently changed regarding deps we have, or I was just hitting a weird cache at the time which made me think it was unnecessary and working fine.

### Fixed

- Fixed test:visual for our local environments

## Testing

1. cd packages/site
2. `npm run test:visual`
    * observe it fails
3. Pull this branch
4. `npm run test:visual`
    * observe it succeeds!

---

### DIFF
--- a/packages/site/.gitignore
+++ b/packages/site/.gitignore
@@ -5,6 +5,7 @@ public/foundation.css
 public/dark.mode.css
 public/styles.css
 node_modules/
+node_modules.e2e/
 test-results/
 playwright-report/
 blob-report/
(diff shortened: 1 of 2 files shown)


scotttjob (approved):
I remember this! Approveled

PR #2640, opened 2025-08-01
feat(components): Use FloatingUI for Dismissible Chips

ZakaryH:
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

### DIFF
--- a/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
+++ b/packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
@@ -1,6 +1,7 @@
 import React, { useEffect } from "react";
 import classNames from "classnames";
 import { useSafeLayoutEffect } from "@jobber/hooks/useSafeLayoutEffect";
+import { FloatingPortal } from "@floating-ui/react";
 import styles from "./InternalChipDismissible.module.css";
 import { ChipDismissibleInputProps } from "./InternalChipDismissibleTypes";
 import {
@@ -13,6 +14,8 @@ import { Text } from "../../Text";
 import { Button } from "../../Button";
 import { Spinner } from "../../Spinner";
 
+// Adding Portal moves it over the limit, but doesn't impact the readabiltiy
+// eslint-disable-next-line max-statements
 export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
   const {
     activator = <Button icon="add" type="secondary" ariaLabel="Add" />,
@@ -44,14 +47,13 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     handleDebouncedSearch,
   } = useInternalChipDismissibleInput(props);
 
-  const menuRef = useScrollToActive(activeIndex);
+  const scrollableRef = useScrollToActive(activeIndex);
   const { ref: visibleChildRef, isInView } = useInView<HTMLDivElement>();
 
   const {
-    styles: popperStyles,
-    attributes,
+    styles: floatingStyles,
     update,
-    setPositionedElementRef,
+    setFloatingRef,
   } = useRepositionMenu(attachTo);
   useSafeLayoutEffect(() => {
     update?.();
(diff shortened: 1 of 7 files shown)


packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -93,49 +138,7 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
...
-
-            {isLoadingMore && (
-              <div className={styles.loadingIndicator}>
-                <Spinner size="small" inline />
-              </div>
-            )}
-          </div>
-        </div>
-      )}
+      {shouldShowMenu && <FloatingPortal>{menuContent}</FloatingPortal>}
```
ZakaryH, line 141: not strictly necessary. no one has complained about this, but it's the classic setup for the problem of if you use this inside something with overflow:hidden, it will get cut off

the portal prevents that, and the Floating version plays nice with other Floating context

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
```diff
@@ -44,14 +47,13 @@ export function InternalChipDismissibleInput(props: ChipDismissibleInputProps) {
     handleDebouncedSearch,
   } = useInternalChipDismissibleInput(props);
 
-  const menuRef = useScrollToActive(activeIndex);
+  const scrollableRef = useScrollToActive(activeIndex);
   const { ref: visibleChildRef, isInView } = useInView<HTMLDivElement>();
 
   const {
-    styles: popperStyles,
-    attributes,
```
ZakaryH, line 52: worth double checking that `attributes` was nothing previously. I'm pretty sure it was.

packages/components/src/Chips/InternalChipDismissible/hooks/useInternalChipDismissible.ts
```diff
@@ -38,7 +38,9 @@ export function useInternalChipDismissible({
   onClick,
   onCustomAdd,
 }: InternalChipDismissibleProps) {
-  const ref = useRef<HTMLDivElement>(null);
```
ZakaryH, line 41: floating UI wants you to use React state for the "attachTo" so it can stay up to date. real refs don't trigger updates, so won't be able to keep up.

packages/components/src/Chips/InternalChipDismissible/hooks/useRepositionMenu.ts
```diff
@@ -1,17 +1,55 @@
...
+  readonly update: () => void;
+}
+
+export function useRepositionMenu(
+  attachTo: HTMLElement | null,
+): UseRepositionMenu {
+  const { refs, floatingStyles, update } = useFloating({
+    placement: "bottom",
+    middleware: [
+      offset(8),
```
edison-cy-yang, line 26: can we create a variable with self-explanatory name for this offset number?

packages/components/src/Chips/InternalChipDismissible/InternalChipDismissibleInput.tsx
edison-cy-yang: Seeing this makes me realize the Popover version still uses ReactDOM's portal.

packages/components/src/Chips/InternalChipDismissible/hooks/useRepositionMenu.ts
ZakaryH: lmk what you think of the name. I tried to communicate what it does and why it's there. maybe a bit wordy?

packages/components/src/Chips/InternalChipDismissible/hooks/useRepositionMenu.ts
```diff
@@ -1,17 +1,55 @@
...
+    },
+    whileElementsMounted: autoUpdate,
   });
 
-  const targetWidth = attachTo.current?.clientWidth;
+  const targetWidth = attachTo?.clientWidth;
 
-  return { ...popper, setPositionedElementRef, targetWidth };
+  return {
+    floatingElement: refs.floating.current,
```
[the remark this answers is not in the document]
ZakaryH: good catch!

packages/components/src/Chips/InternalChipDismissible/hooks/useRepositionMenu.ts
```diff
@@ -1,17 +1,56 @@
...
+): UseRepositionMenu {
+  const { refs, floatingStyles, update } = useFloating({
+    placement: "bottom",
+    middleware: [
+      offset(ROUNDED_BORDER_ARROW_EDGE_OFFSET),
+      flip({ fallbackPlacements: ["top"] }),
+      size({
+        apply({ availableHeight, elements }) {
+          // Simple max height - either 320px or available space, whichever is smaller
+          const maxHeight = Math.min(320, availableHeight);
```
edison-cy-yang, line 32: I know this 320px was just moved over from the css module. If the value lives in here let's create a variable for it as well.

packages/components/src/Chips/InternalChipDismissible/hooks/useRepositionMenu.ts
edison-cy-yang: It's wordy but clearly explains what it does. Now thinking back maybe the similar variable names I added for Popover isnt self explanatory enough

edison-cy-yang (approved):
Had a second pass and everything looks good from the perspective of replacing popper with floating-ui. I still need to become more familiar with Chips to say I fully understand what's happening, but the changes lgtm!

PR #2641, opened 2025-08-01
feat(components): Use Floating UI for Menu

ZakaryH:
## Motivations

Popper is deprecated, so we're moving to FloatingUI.

### Added

- 

size middleware allows for dynamically setting a max height on the floating element based on the available space.

previously our max-height, which will enable scrolling once it is met, was 72vh. that's quite tall meaning not all situations where you want it to trigger happen. the result is that a long menu might have inaccessible items.

now, our max height will still be 72vh if that much space is available, however if it's not then the new max height will be set to the available space.

Kapture 2025-08-01 at 07 37 23

### Changed

- swapped out popper with FloatingUI
- swapped our Portal with FloatingPortal for better compatibility with Floating tools and ecosystem
- renamed variables referencing popper

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

the menu should still be positioned either immediately below the button/activator, or if not enough space it will move above the button

scrolling and layout changes should correctly trigger the position changes

try the size setup. easiest way would be to put a Menu inside a Modal, and give it maybe 12-15 items depending on your screen size. more should work too. see that you can access the items with scrolling.

Changes can be

---


packages/components/src/Menu/Menu.tsx
```diff
@@ -111,34 +118,41 @@ export function Menu({
...
+      flip({ flipAlignment: false }),
+      size({
+        apply({ availableHeight, elements }) {
+          // The inner element is the scrollable menu that requires the max height
+          const menuElement = elements.floating.querySelector(
+            '[role="menu"]',
+          ) as HTMLElement;
+
+          if (menuElement) {
+            const maxHeight = Math.min(MENU_MAX_HEIGHT, availableHeight);
```
ZakaryH, line 137: on all these size middlewares, I think I wanna add just a bit more space. it's exactly flush against the edge and it feels, claustrophobic.

packages/components/src/Menu/Menu.tsx
```diff
@@ -111,34 +118,45 @@ export function Menu({
...
+      size({
+        apply({ availableHeight, elements }) {
+          // The inner element is the scrollable menu that requires the max height
+          const menuElement = elements.floating.querySelector(
+            '[role="menu"]',
+          ) as HTMLElement;
+
+          if (menuElement) {
+            const viewportHeight = window.innerHeight;
+            const maxHeightVh = (viewportHeight * MENU_MAX_HEIGHT_VH) / 100;
```
ZakaryH, line 138: so previously the max-height was 72vh, a relative unit

for this size behavior we need to do a Math.min, which of course you can't do on vh because it hasn't evaluated to a number yet. so, we're doing a baby bit of math ourselves to re-create what 72vh would have done.

packages/components/src/Menu/Menu.tsx
```diff
@@ -111,34 +118,45 @@ export function Menu({
...
+    elements: {
+      reference: referenceElement,
+    },
+    whileElementsMounted: autoUpdate,
   });
+
   const positionAttributes =
     width >= SMALL_SCREEN_BREAKPOINT
       ? {
-          ...attributes.popper,
```
ZakaryH, line 140: once again, we're "losing" this but it's just popper stuff

image

packages/components/src/Menu/Menu.tsx
ZakaryH: playing with this, my 72% logic is a bit flawed. 

it does something slightly different from what I want because it will keep updating as 72% of the **current** view height which is constantly getting recalculated. 

it's not the worst but also it's not the precise expected behavior because the goalposts keep moving and our Math.min trigger becomes difficult to understand

nad182: I've spent quite a bit of time trying to find a better solution and I haven't found one yet, unfortunately. I'm a bit sceptical about using `72%` vs `72vh` for several reasons:
1. dynamic address bars in mobile web browsers will affect the height of the menu (when users scroll and the address bar disappears, the menu height would suddenly change)
2. Menu will look shorter than originally (also on mobile, e.g. by ~70px in Safari)
3. the height of the Menu will look slightly different across different browsers.

If you think these differences/inconsistencies are negligible, then ignore my concern.

nad182 (changes requested):
Left 2 "concerns", which are not necessarily blocking.

packages/components/src/Menu/Menu.tsx
ZakaryH: >dynamic address bars in mobile web browsers will affect the height of the menu (when users scroll and the address bar disappears, the menu height would suddenly change)

that's pretty big edge case, and thinking about it if I'm in the URL bar, that's what I'm focused on. does it matter to me that the menu changed height?

do you have an example of this happening where it creates a bad UX?

>Menu will look shorter than originally (also on mobile, e.g. by ~70px in Safari)


it shouldn't?
Notification_Center


Components___Navigation___Menu___Web_-_Horizontal_⋅_Storybook_and_Components___Navigation___Menu___Web_-_Horizontal_⋅_Storybook


Components___Navigation___Menu___Web_-_Horizontal_⋅_Storybook_and_Components___Navigation___Menu___Web_-_Horizontal_⋅_Storybook

I misspoke about my math. it's actually perfectly equivalent to the vh version.

1 vh = 1% of viewport height, a relative unit so this changes as the screen dimensions change

mine does the exact same thing, just with the browser API and manual calculations

the reason it appears shorter is actually because of the `box-sizing: border-box` I added.

on prod, the dimensions of the menu are very difficult to know due to the box sizing model.

the viewport height is 488
image

so it should be 351px tall. however it's not

image

it's 18px taller due to the padding and borders not being taken into account for its height

I suppose it's not a bug, but it does make it hard to understand the expected height.

this didn't realllly matter before, but now it does because the `size` middleware needs to set heights based on the actual height of the element. we _could_  leave it the same and account for that 18px in our calculation but that's really messy. I'd rather just use `border-box`. I'm not too worried about that change. I can take a peek at some existing Menus and make sure they don't freak out if you change their box-sizing.


>e.g. by ~70px in Safari

this amount however is not expected. do you have an example of that? that should never happen.


>the height of the Menu will look slightly different across different browsers.


I don't follow. can you provide examples of that? 

I know there are some differences with the `vh` unit in Safari, meaning this would already be the cause 

aside from that though I would expect that the manual calculations wouldn't have that same problem unless for some reason `innerHeight` never updates which would be wild

ZakaryH: >dynamic address bars in mobile web browsers will affect the height of the menu (when users scroll and the address bar disappears, the menu height would suddenly change)

wait I just remembered. this will never happen.

Menu has a totally different UI/UX on mobile.

image

ZakaryH: true. I'm still having a hard time imagining how this might result in a bad experience, or how it would be any different from today since we're already using units that are relative to the viewport height

nad182: It's a bit hard to explain the difference between Safari on iOS device and Chrome on Android device. The address bars have different height, so the menu appears to look slightly different (if we calculate the height based on `innerHeight` vs `vh`, which takes all viewport into account). But I agree, this is an edge case, so I'm good with ignoring this difference. We can go back to it, in case we here some reports with this interfering with any "real" menus in our app.

image


image

PR #2644, opened 2025-08-01
feat(components): Use FloatingUI for Tooltip

ZakaryH:
## Motivations

Popper is deprecated. replaced by FloatingUI.

### Added

shift middleware allows it to sort of follow you along as you scroll for example

this will be impossible to do with the hover trigger but focusing can happen. also added the shiftLimiter to prevent it from showing up off screen.

### Changed

- replaced usePopper with useFloating
- renamed anything referencing popper
- swapped the ReactDOM portal for FloatingPortal

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

**shift + limitshift** 

since tooltips respond to focus, focus the activator element then scroll. the tooltip should shift along with the movement, and the arrow should move along with you BUT hit the edge at such a point that doesn't look bad with the rounded border radius.

furthermore the shift should not apply to things fully off screen. so if you have a tooltip on something, focus it, then scroll it completely out of view it shouldn't be stuck at the X/Y coordinates where it would have been if it were in view.

**arrow position**
should be correct and look good

**placement**
preferred placement should be respected

it should adjust based on available space and use sensible fallbacks

**portal**

if you put it inside something with `overflow:hidden` it should still show because it's out on the body

Changes can be

---

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.tsx
+++ b/packages/components/src/Tooltip/Tooltip.tsx
@@ -1,9 +1,8 @@
-import React, { ReactElement, ReactNode, useState } from "react";
+import React, { ReactElement, useState } from "react";
 import classnames from "classnames";
-import ReactDOM from "react-dom";
+import { FloatingPortal } from "@floating-ui/react";
 import { motion } from "framer-motion";
 import { useSafeLayoutEffect } from "@jobber/hooks/useSafeLayoutEffect";
-import { useIsMounted } from "@jobber/hooks/useIsMounted";
 import styles from "./Tooltip.module.css";
 import { useTooltipPositioning } from "./useTooltipPositioning";
 import { Placement } from "./Tooltip.types";
@@ -40,7 +39,7 @@ export function Tooltip({
     attributes,
     placement,
     shadowRef,
-    styles: popperStyles,
+    styles: floatingStyles,
     setArrowRef,
     setTooltipRef,
   } = useTooltipPositioning({ preferredPlacement: preferredPlacement });
@@ -59,13 +58,14 @@ export function Tooltip({
     <>
       <span className={styles.shadowActivator} ref={shadowRef} />
       {children}
-      <TooltipPortal>
+      <FloatingPortal>
         {show && Boolean(message) && (
           <div
             className={toolTipClassNames}
-            style={popperStyles.popper}
+            style={floatingStyles.popper}
             ref={setTooltipRef}
             role="tooltip"
+            data-placement={placement}
             {...attributes.popper}
           >
             <motion.div
@@ -83,13 +83,23 @@ export function Tooltip({
               <p className={styles.tooltipMessage}>{message}</p>
               <div
                 ref={setArrowRef}
-                style={popperStyles.arrow}
+                style={{
+                  position: "absolute",
+                  left:
+                    floatingStyles.arrow?.x != null
+                      ? `${floatingStyles.arrow?.x}px`
+                      : "",
+                  top:
+                    floatingStyles.arrow?.y != null
+                      ? `${floatingStyles.arrow?.y}px`
+                      : "",
+                }}
                 className={styles.arrow}
               />
             </motion.div>
           </div>
         )}
-      </TooltipPortal>
+      </FloatingPortal>
     </>
   );
 
@@ -146,18 +156,4 @@ export function Tooltip({
       };
     }, []);
   }
-}
-
-interface TooltipPortalProps {
-  readonly children: ReactNode;
-}
-
-function TooltipPortal({ children }: TooltipPortalProps) {
-  const mounted = useIsMounted();
-
-  if (!mounted?.current) {
-    return null;
-  }
-
-  return ReactDOM.createPortal(children, document.body);
 }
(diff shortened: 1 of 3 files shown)


packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -83,13 +83,23 @@ export function Tooltip({
...
+                  position: "absolute",
+                  left:
+                    floatingStyles.arrow?.x != null
+                      ? `${floatingStyles.arrow?.x}px`
+                      : "",
+                  top:
+                    floatingStyles.arrow?.y != null
+                      ? `${floatingStyles.arrow?.y}px`
+                      : "",
+                }}
```
ZakaryH, line 96: this is a bit different from how Popover manages the arrow

ideally they work the same way so you don't have to understand two systems. a shared hook might even be the way to go. I'll see about making them more similar.

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -147,17 +157,3 @@ export function Tooltip({
...
-
-function TooltipPortal({ children }: TooltipPortalProps) {
-  const mounted = useIsMounted();
-
-  if (!mounted?.current) {
-    return null;
-  }
-
-  return ReactDOM.createPortal(children, document.body);
-}
```
ZakaryH, line 163: I didn't have to get rid of this. it's not doing any harm to have it be SSR safe. we just don't explicitly need it anymore.

packages/components/src/Tooltip/useTooltipPositioning.ts
```diff
@@ -12,29 +21,41 @@ export function useTooltipPositioning({
...
 
   return {
-    ...popper,
-    placement: popper.state?.placement || DEFAULT_PLACEMENT,
+    styles: {
+      popper: floatingStyles,
+      arrow: middlewareData.arrow,
+    },
+    attributes: { popper: {} },
+    placement: placement || DEFAULT_PLACEMENT,
```
edison-cy-yang, line 56: This seems to be pre-existing, but we have set the default of `preferredPlacement` prop in `Tooltip` to be `top` already, so this `DEFAULT_PLACEMENT` is not needed

edison-cy-yang (commented):
3 small questions. Tested it in storybook and everything looks good

packages/components/src/Tooltip/Tooltip.tsx
ZakaryH: thinking on this more, it's both not very intuitive these values are possible nor is it obvious that the `!= null` is **very** important for this to work

maybe a named function to perform the comparison or return the value would help. or even a comment like the Popover version has.

packages/components/src/Tooltip/useTooltipPositioning.ts
ZakaryH: nice catch. you're right. I don't know how we'd ever hit the other side of the `||` unless something went wrong with `useFloating` returning the `placement` and at that point idk what sort of state we're in.

packages/components/src/Tooltip/Tooltip.tsx
ZakaryH: I've made it more like the Popover version, and swapped `!=` with `Number.isFinite` they _should_ behave the same but hopefully in a clearer way for a reader?

```
undefined != null
false
null != null
false
0 != null
true
```

and 

```
Number.isFinite(undefined)
false
Number.isFinite(null)
false
Number.isFinite(0)
true
﻿
```

do you think that's any better or nah?

and yeah I'll leave the shared hook, if we want to pursue that, until the cleanup phase.

ZakaryH: haha fair enough. 

do you have a preference between:

`x !== null && x !== undefined` - verbose but clear
`x != null` - a clever, succinct but sneaky way to check both `undefined` and `null` at the same time, leveraging the fact that `undefined == null` returns `true` 
`Number.isFinite(x)` - well, I guess we just said no thanks to this one

ZakaryH: good question. looking at the types I don't see how it would be null.... I dunno. I assume they have that for some reason?  they do directly mention the null check in the docs.

I'll just go with the first option and be done with it 😄

ZakaryH:
something I noticed that isn't ideal is the arrow can exceed the boundaries of the attached element

image

this is happening  because of the `padding: TOOLTIP_ARROW_PADDING,`

buut that's also what they say to solve this problem 🤷 

[Arrow does not avoid rounded corners](https://floating-ui.com/docs/FloatingArrow#arrow-does-not-avoid-rounded-corners)
The arrow doesn’t automatically avoid rounded corners of the floating element when it’s shifted. To ensure it doesn’t overflow the corners, you can use the padding prop in the arrow() middleware when calling useFloating() to account for this.


vs with it gone

image

packages/components/src/Tooltip/Tooltip.tsx
ZakaryH: decided to keep it. it wasn't hurting anyone.

edison-cy-yang (commented):
There's also still a reference to popper on line 75 of `Tooltip.tss`

PR #2649, opened 2025-08-05
fix(components): Don't import directly from the hooks barrel file

jdeichert:
## Motivations

This led to a production build error due to the way vite's rollup bundles our app. In vite dev mode, powered by esbuild, we didn't see this issue, but production builds use rollup which behaves differently.

The hooks package is lacking support for esm, so we suspect that's what caused rollup to build things differently.

### Changed

- Fix import to be direct

## Testing

* Make a prerelease
* Install that in the app
* Run `pnpm dedupe && pnpm run build && pnpm run preview`

---

### DIFF
--- a/packages/components/src/FormField/FormFieldTypes.ts
+++ b/packages/components/src/FormField/FormFieldTypes.ts
@@ -1,7 +1,7 @@
 import React, { ChangeEvent, ReactNode, RefObject } from "react";
 import { RegisterOptions } from "react-hook-form";
 import { XOR } from "ts-xor";
-import { type Clearable } from "@jobber/hooks/useShowClear";
+import { Clearable } from "@jobber/hooks";
 import { IconNames } from "../Icon";
 
 export type FormFieldTypes =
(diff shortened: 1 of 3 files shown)


edison-cy-yang (approved):
Tested in pre-release on a production build and I dont see that error.

packages/components/src/LightBox/LightBox.tsx
```diff
@@ -8,7 +8,7 @@ import { useFocusTrap } from "@jobber/hooks/useFocusTrap";
 import { useIsMounted } from "@jobber/hooks/useIsMounted";
 import { useBreakpoints } from "@jobber/hooks/useBreakpoints";
 import classNames from "classnames";
-import { useDebounce } from "@jobber/hooks";
+import { useDebounce } from "@jobber/hooks/useDebounce";
```
nad182, line 11: I'm too late for this PR, but I think the proper solution would be to export each hook in `package.json` in the `exports` block - the same way we do in `components`. But looks like Scott submitted a [PR](https://github.com/GetJobber/atlantis/pull/2653) to do this.

PR #2653, opened 2025-08-06
fix(hooks): ESM Update (Also Formatters)

scotttjob:
## Motivations

1. When we did our ESM upgrades last year, we (me) failed to update both hooks/formatters.
2. Due to a recent issue, this update is relevant again so I've re-applied the work from an old branch.

## Changes

1. Updates hooks/formatters to be built using ESM instead of CJS.
2. This shouldn't be an issue for any downstream consumers, as this is a correction to the proper path. If we do get reports of consumers that would like the CJS builds for formatters/hooks, we have options (secondary tsconfig and we build twice, introduce rollup to these folders).

### Removed

- CJS builds for Hooks/Formatters

### Fixed

- Inconsistent Atlantis Building

## Testing

1. Everything should still continue to work as-is, there should be no downstream effect to any application level concerns (this is a well under the hood change). If anything, we should expect to see marginal bundle size improvements (hooks isn't very big).
2. We've gone to the effort of including all the hooks in the package.json, so this should be backwards compatible with all existing imports of hooks. If something is broken, it means this PR should be adjusted to match, it should be compatible.

Changes can be

---


Aiden-Brine (approved):
The changes look good and I tested it out locally and in staging for JFE

PR #2655, opened 2025-08-07
feat(components): FloatingUI Cleanup

ZakaryH:
## Motivations

after doing a few FloatingUI migrations, I've noticed some things we missed previously and/or that could be improved

### Added

- util for the common `size` middleware max/min height logic + some padding to make sure it stays off the edge
f

### Changed

**FloatingPortal**

we missed a couple spots in Combobox and Popover, just swapping these over too since the rest use this now for portalling the content

**Combobox**
Combobox had a couple middleware implementations that weren't quite what we wanted

shift: was meant to solve a position problem but you don't really want your Combobo menu to be able to shift
instead: we'll just give it more acceptable placements to solve the case where it's placed on the far right hand side of a view

autoPlacement: there's a subtle distinction between this and `flip`. autoplacement prioritizes maximum available space whereas flip will use your preferred placement until it can't. with Combobox, the preferred placement is below

**size**

I really wanted to add the size middleware to Combobox but it's a bit more involved since there are multiple scrolling elements within the floating menu. we'll have to do this another time.

**variable changes**
some vars still referenced "popper". 

**Size maxHeight**

many of these were directly using `availableHeight` as the maxHeight we'd apply when we run out of space. the problem with that is it means the edge of your element is flush against the edge of teh viewport. it feels and looks weird.

now, if we are going to use the availableHeight - we subtract a bit of space off that so it has room to breathe 

before
image

after
image

### Deprecated

- 

### Removed

- 

### Fixed

missed it during Menu's conversion, or maybe it was just never working, but used FloatingUI's interaction instead of our `useKeyDown` to allow a Menu within a Modal to be closed on its own, meaning if you press ESC - it only closes the open Menu and not the Modal too

### Security

- 

## Testing

**Combobox**

should work as it did before when it's NOT inside a Modal

if you place it on the far right of a screen, opening it should place it in such a way that you can still see and interact with the menu

if you run out of space vertically below, it should move above

**maxHeight/size on: Menu, Chips, Autocomplete**

all of these you will just want to try having them open in a situation where there's no space above for everything to show, and then it'll do its best to show it in the preferred lower position. when you have enough content that it can't even do that, it should set the maxheight of the floating element such that you can now scroll through the items that won't fit in the view.

when it hits these conditions, it should have a bit of room against the edge of the screen..

**Menu within Modal**

put a Menu in a Modal, open the Modal, open the Menu, press ESC to close the Menu -> see that only the Menu closes

Changes can be

---

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -1,6 +1,7 @@
 import React, { useMemo } from "react";
 import classnames from "classnames";
-import { FloatingPortal } from "@floating-ui/react";
+import { FloatingNode, FloatingPortal, FloatingTree } from "@floating-ui/react";
+import ReactDOM from "react-dom";
 import styles from "./ComboboxContent.module.css";
 import { ComboboxContentSearch } from "./ComboboxContentSearch";
 import { ComboboxContentList } from "./ComboboxContentList";
@@ -17,13 +18,14 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
     props.selected,
   );
 
-  const { floatingRef, floatingStyles } = useComboboxAccessibility(
-    props.handleSelection,
-    props.options,
-    optionsListRef,
-    props.open,
-    props.wrapperRef,
-  );
+  const { popperRef, popperStyles, floatingProps, nodeId, parentNodeId } =
+    useComboboxAccessibility(
+      props.handleSelection,
+      props.options,
+      optionsListRef,
+      props.open,
+      props.wrapperRef,
+    );
 
   // options that are passed back to consumers via onSelectAll callback
   // should only contain id and label
@@ -36,7 +38,7 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
 
   const content = (
     <div
-      ref={floatingRef}
+      ref={popperRef}
       id={COMBOBOX_MENU_ID}
       data-testid={COMBOBOX_MENU_ID}
       data-elevation={"elevated"}
@@ -44,7 +46,8 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
       className={classnames(styles.content, {
         [styles.hidden]: !props.open,
       })}
-      style={floatingStyles}
+      style={popperStyles}
+      {...floatingProps}
     >
       <ComboboxContentSearch
         open={props.open}
@@ -96,9 +99,17 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
     </div>
   );
 
-  return globalThis?.document ? (
-    <FloatingPortal>{content}</FloatingPortal>
-  ) : (
-    content
-  );
+  if (parentNodeId) {
+    return (
+      <FloatingTree>
+        <FloatingNode id={nodeId}>
+          <FloatingPortal>{content}</FloatingPortal>
+        </FloatingNode>
+      </FloatingTree>
+    );
+  }
+
+  return globalThis?.document
+    ? ReactDOM.createPortal(content, document.body)
+    : content;
 }
--- a/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
+++ b/packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
@@ -18,7 +18,7 @@ export function ComboboxContent(props: ComboboxContentProps): JSX.Element {
     props.selected,
   );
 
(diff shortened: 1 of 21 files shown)


packages/components/src/Combobox/hooks/useComboboxAccessibility.ts
```diff
@@ -26,36 +23,30 @@ export function useComboboxAccessibility(
...
       reference: wrapperRef.current,
-      floating: popperRef.current,
+      floating: floatingRef.current,
     },
     open,
     onOpenChange: openState => {
       if (!openState) handleClose();
     },
     middleware: [
-      autoPlacement({ allowedPlacements: ["bottom-start", "top-start"] }),
```
ZakaryH, line 56: >flip() uses a fallback “no space” strategy. Ensures the preferred placement is kept unless there is no space left.

>autoPlacement() uses a primary “most space” strategy. Always chooses the placement with the most space available.

https://floating-ui.com/docs/autoPlacement#conflict-with-flip

a Combobox is somewhat like a dropdown, or a Select. so our preference is for the menu to be on the bottom. it doesn't really matter if there's more space on the top, we want it to be on the bottom unless it doesn't have room.

Autoplacement could look at a layout with 400 px above, and 350px below and say "you're going up top".

we have a separate issue with the size but I won't get into that here.

packages/components/src/Combobox/components/ComboboxContent/ComboboxContent.tsx
ZakaryH, line 1: I know this looks like a lot but all it is is removing FloatingTree, and FloatingNode.

those are more than what we need to solve this problem.

packages/components/src/Modal/Modal.rebuilt.tsx
ZakaryH, line 1: same thing here. it looks like a lot of changes. it's removing FloatingFocusManager, FloatingNode and FloatingTree in another Modal file

ZakaryH: ah, maybe I spoke too soon. there's some weirdness going on.

packages/components/src/Menu/Menu.tsx
```diff
@@ -154,6 +158,9 @@ export function Menu({
     whileElementsMounted: autoUpdate,
   });
 
+  const dismiss = useDismiss(context);
```
ZakaryH, line 161: missed this originally when moving Menu over

having Menu inside Modal, closing Menu with ESC would close the Modal too

this fixes it, so no more `useOnKeyDown` for ESC needed

ZakaryH:
I tried to make a common hook for Tooltip and Popover's arrow positioning + useFloating

however they are slightly different in a few ways that makes it annoying to share code. I'm gonna leave it as-is for now. if we ever have a 3rd, arrow-having floating element then we can revisit this

packages/components/src/Popover/Popover.tsx
```diff
@@ -64,8 +64,8 @@ Popover.Arrow = function PopoverArrow({
   const arrowStyles: React.CSSProperties = {
     position: "absolute",
     // only left or top will be defined at a time
-    left: popperStyles.arrow?.x != null ? `${popperStyles.arrow?.x}px` : "",
-    top: popperStyles.arrow?.y != null ? `${popperStyles.arrow?.y}px` : "",
+    left: floatingStyles.arrow?.x != null ? `${floatingStyles.arrow?.x}px` : "",
```
edison-cy-yang, line 67: should we do the same `!== null` and `!== undefined` check here?

packages/components/src/Combobox/hooks/useComboboxAccessibility.ts
edison-cy-yang: 👌

packages/components/src/Combobox/hooks/useComboboxAccessibility.ts
```diff
@@ -26,36 +23,30 @@ export function useComboboxAccessibility(
...
       offset(COMBOBOX_OFFSET),
-      shift({ padding: COMBOBOX_OFFSET }),
```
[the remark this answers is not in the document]
edison-cy-yang: Nice, even if we have shift it should have the options of `mainAxis: true, crossAxis: false`. But yeah, doesnt really make sense for a dropdown menu like component

edison-cy-yang (commented):
Mostly looks good! I found a weird behaviour with useDismiss and useInteractions. Theyre used according to the documentation though, maybe Im missing something.

edison-cy-yang (approved):
Good to go!

edison-cy-yang:
Just saw the CI failure, is that related to the npm blip from yesterday?

PR #2657, opened 2025-08-07
fix(components-native): Fix Select presses not always responding on Android [JOB-131539]

Aiden-Brine:
## Motivations

Back when we [allowed Text highlighting](https://github.com/GetJobber/jobber-mobile/pull/5162) we added the [`TypographyGestureDetector`](https://github.com/GetJobber/atlantis/blob/master/packages/components-native/src/Typography/TypographyGestureDetector.tsx) to essentially swallow up gestures on text to prevent highlighting text when scrolling. Unfortunately, this was interfering with gesture handling on text elements inside the Select component. The text components were occupying more space than just the visible text, resulting in clicks not always propagating correctly when users tried to interact with the selection area.

### Changed

When `selectable=false` is passed to `Text`/`Typography` the component will no longer be wrapped in `TypographyGestureDetector` so will allow gestures on text to propagate up. `selectable=false` was set on the text within the selectable component. 

### Fixed

The bug where the Select component seemed laggy or unresponsive on Android

### Follow up

[Follow up ticket](https://jobber.atlassian.net/browse/JOB-132306?atlOrigin=eyJpIjoiNmEyMGRmODNjMWI0NDc0Yjg2NTU1YzMwNThmNDRmOGMiLCJwIjoiaiJ9)

## Testing

Try using Select on an Android device with this build or with an emulator and the menu should consistently open regardless of where you click within the Select.

Changes can be

---

### DIFF
--- a/packages/components-native/src/Select/Select.tsx
+++ b/packages/components-native/src/Select/Select.tsx
@@ -159,7 +159,6 @@ export function Select({
                 variation={textVariation}
                 hideFromScreenReader={true}
                 selectable={false}
-                allowParentPress={true}
               >
                 {label}
               </Text>
@@ -170,7 +169,6 @@ export function Select({
                   variation={disabled ? "disabled" : "base"}
                   hideFromScreenReader={true}
                   selectable={false}
-                  allowParentPress={true}
                 >
                   {getValue()}
                 </Text>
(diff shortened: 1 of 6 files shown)


Aiden-Brine:
@jdeichert thanks for the reminder, I meant to do that and forgot. PR description is now updated.

Aiden-Brine:
I tested this out on a physical Android device and Select is now responsive to all clicks. I also tested scrolling on text and it is working well, does not seem to be impacted.

PR #2659, opened 2025-08-07
chore(components): Remove Popper Library

ZakaryH:
## Motivations

getting rid of the dependency of popper

### Added

- 

### Changed

fixed a typo that said "popper" that was showing up in my search results

### Deprecated

- 

### Removed

removed the packages from our direct dependencies

though the package will continue to be included  due to `react-datepicker` still using `@popperjs/core` and `react-popper`

that's also the only remaining real reference to `popper` is react datepicker props, css rules and tests. can't get rid of these until we upgrade, and presumably they've gotten off popper too.

removed a mock for popper in tests

removed the rollupconfig entry for popper too

### Fixed

- 

### Security

- 

## Testing

build should work

React-datepicker should still work

Changes can be

---


ZakaryH:
something I ran into uninstalling popper that we'll need to sort out. for `react-router-dom`

we have a ^5 `dependency` in components

we also have a ^6 `peerDependency` in components

running `npm uninstall _____` caused npm to try and resolve this issue, resulting in a whole bunch of version changes including moving moving `react-router-dom` up to ^6

having both is not great for a few reasons: bundle size, compatibility with contexts, multiple contexts etc.

but going to 6 has some extra complexity. `@types/react-router-dom` doesn't exist anymore. versions of react-router-dom >6 ship with their own types, so a bunch of side effects from that. gotta uninstall the types, and of course update anything that changed. we don't really want react-router-dom in the package anymore though, so the real solution here is fully detaching from that and getting rid of the dep altogether!

edison-cy-yang (approved):
Installed pre-release in JO and JFE. Verified a few floating-ui components and DatePicker are working fine. LGTM

PR #2666, opened 2025-08-12
fix(components): Improve Floating Usages

ZakaryH:
## Motivations

Combobox manages the menu visibility with CSS, not conditional rendering (though we should do that!)

```
  const content = (
    <div
      ref={floatingRef}
      id={COMBOBOX_MENU_ID}
      data-testid={COMBOBOX_MENU_ID}
      data-elevation={"elevated"}
      tabIndex={0}
      className={classnames(styles.content, {
        [styles.hidden]: !props.open,
      })}
```

we shouldn't use the whileMounted + autoUpdate in this scenario

```
import {autoUpdate} from '@floating-ui/react';
 
useFloating({
  whileElementsMounted: autoUpdate,
});
```
>Warning
>Only use this option if your floating element is conditionally rendered, not hidden with CSS. Use an Effect to call and clean up autoUpdate in the latter scenario.

https://floating-ui.com/docs/useFloating#whileelementsmounted

### Added

- 

### Changed

using a "manual" update useEffect rather than the updateWhileMounted version from useFloating

moved the placement config up higher to prevent an issue I believe I'm seeing related to the fallbacks happening earlier than needed

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

not entirely sure how to reproduce the top placement, it's only happening in one codebase oddly enough

as for the updating, it should be identical to before. changes in the available space from scrolling, or layout changes should update the floating state and it should respond to it AKA if you scroll and it no longer has space below to display, it'll move up to the top during that scroll (and it is open)

if you scroll and then open, it should still open in a place where it can

Changes can be

---


ZakaryH:
thought: "hide" middleware might be a good fit for Popover

PR #2668, opened 2025-08-13
fix(components): Provide Additional Fallback Placements

ZakaryH:
## Motivations

Menu had a strict fallback strategy, disallowing it from flipping when on the far right of a screen.

Tooltip didn't have "top" defined as a possible fallback, and since the initial placement is configurable it's conceivable that it could start on the bottom and need to flip to top.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

Menu and Tooltip have more fallbacks for when they can't be in their preferred placement

### Security

- 

## Testing

Changes can be

---


ZakaryH:
then for trying out the Tooltip change, just give it a preferred placement of bottom, then focus the trigger element and scroll such that it has no space on the bottom

it should be able to go to the top

PR #2670, opened 2025-08-13
fix(components): Use type import specifier in packages/components

jdeichert:
## Motivations

This PR cleans up our type-related imports to use the `type` specifier. This is the modern approach for importing types and informs bundlers like vite that they should ignore these.

Scott started this work as part of [this PR](https://github.com/GetJobber/atlantis/pull/2492). We had to separate these changes out from that work however. 

This will unblock my current storybook v9 work as well! Storybook uses vite and enforces all type imports are correctly defined with `type`.

This PR also configures 2 new typescript eslint rules to enforce this:

* @typescript-eslint/consistent-type-exports
* @typescript-eslint/consistent-type-imports

Check out [packages/components/.eslintrc.cjs](https://github.com/GetJobber/atlantis/pull/2670/files#diff-72a2a01e552e9452de82824c9fd3641d3725574ae82eb2215e3b02787e2dc540) to see these new rules.

⚠️ NOTE: this PR is focused on `packages/components`. It's already massive, so I want to avoid cleaning up other packages here and instead I'll break those out into separate PRs.

### Changed

- Updated type imports to use the type specifier
- Added eslint rules for type imports

## Testing

Just import syntax changes, so it should only affect our typescript type checking and not our bundling process.

I've triggered a pre-release and will test against JO/JFE 👍 

---

### DIFF
--- a/packages/components/src/AtlantisThemeContext/AtlantisThemeContext.tsx
+++ b/packages/components/src/AtlantisThemeContext/AtlantisThemeContext.tsx
@@ -8,13 +8,13 @@ import React, {
   useState,
 } from "react";
 import merge from "lodash/merge";
-import { THEME_CHANGE_EVENT } from "./types";
 import type {
   AtlantisThemeContextProviderProps,
   AtlantisThemeContextValue,
-  type Theme,
-  type ThemeChangeDetails,
+  Theme,
+  ThemeChangeDetails,
 } from "./types";
+import { THEME_CHANGE_EVENT } from "./types";
 import styles from "./AtlantisThemeContext.module.css";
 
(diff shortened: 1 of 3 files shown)


packages/components/src/Autocomplete/Menu/DefaultMenu.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import { BaseAutocompleteMenuWrapper } from "./MenuWrapper";
-import { AnyOption, Option } from "../Autocomplete.types";
+import { type AnyOption, type Option } from "../Autocomplete.types";
```
edison-cy-yang, line 3: `type` can be moved up for this as well

PR #2677, opened 2025-08-15
fix(components-native): Use type import specifier in packages/components-native

jdeichert:
## Motivations

This is like the components PR I made a few days ago: https://github.com/GetJobber/atlantis/pull/2670

This PR cleans up our type-related imports to use the `type` specifier. This is the modern approach for importing types and informs bundlers like vite that they should ignore these.

Scott started this work as part of [this PR](https://github.com/GetJobber/atlantis/pull/2492). We had to separate these changes out from that work however. 

This will unblock my current storybook v9 work as well! Storybook uses vite and enforces all type imports are correctly defined with `type`.

This PR also configures 2 new typescript eslint rules to enforce this:

* @typescript-eslint/consistent-type-exports
* @typescript-eslint/consistent-type-imports

### Changed

- Updated type imports to use the type specifier
- Added eslint rules for type imports

## Testing

Just import syntax changes, so it should only affect our typescript type checking and not our bundling process.

---

### DIFF
--- a/packages/components-native/tsconfig.json
+++ b/packages/components-native/tsconfig.json
@@ -28,6 +28,7 @@
   "exclude": [
     "node_modules",
     "dist/**/*",
+    "tsconfig.eslint.json",
     "**/*.test.ts",
     "**/*.test.tsx",
   ],


Aiden-Brine (approved):
The code looks good and I smoke tested the changes in an iOS emulator

PR #2687, opened 2025-08-22
feat(components): React 19 support

jdeichert:
## Motivations

This PR is just to verify any remaining integration blockers with React 19 compatibility.

Things to note and verify:

* React 19 is defined at the root of the repo, all subpackages depend on this
  * Storybook v7 is in its own subdirectory, still must use react 18
* `@types/react` is still on React 18
  * We likely need to be compatible with both React 18 and React 19 for some time
  * Using React 18 types should help ensure we don't use any new React 19 functionality
  * The downside is that we may be using deprecated APIs that don't exist in React 19
  * TODO: discuss this tradeoff
* All packages should compile without errors: react is a peerDep so it doesn't affect the code we actually ship
* Docs site should now use React 19
  * `unmountComponentAtNode` was removed in v19
* Storybook v9 (both web and mobile) is running on React 19
* All tests should pass against React 19

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

---


packages/components/package.json
```diff
@@ -531,9 +528,9 @@
     "filesize": "^6",
     "framer-motion": "^11.11.12",
```
ZakaryH: did I forget to update this? whoops!

ZakaryH (approved):
only ran the code in mobile, storybook and new site. everything built fine, nothing obviously broken.

really like the test improvements. a couple minor things for us to follow up with, but aside from that LGTM

PR #2698, opened 2025-08-28
fix(hooks): Revert ESM Update (Also Formatters)

jdeichert:
Reverts GetJobber/atlantis#2653

Shipping that to our consumers turned out to be quite a large effort. We'll have to plan for that.


ZakaryH (approved):
revert looks good to me

tried building and using a few components that interact with hooks all working as expected

PR #2701, opened 2025-08-28
feat(components): Provide Default Content to Combobox CustomRender

ZakaryH:
## Motivations

there are times when you want to enhance or wrap a Combobox option, yet you don't want to change the layout.

additionally, there is complexity added from the portalling meaning that if you try to wrap the `<Combobox>` invocation, since the menu itself ends up elsewhere on the DOM - it might not behave as expected

to deal with both problems, we are going to provide what we would have rendered as an arg to the custom render render prop

this way a consumer can wrap the option, it will be directly around the option sidestepping the portal issue, and there's no need to make a fragile re-creation of the default content that may change

### Added

DefaultContent arg provided to customRender on Combobox 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

### DIFF
--- a/docs/components/Combobox/Web.stories.tsx
+++ b/docs/components/Combobox/Web.stories.tsx
@@ -536,6 +536,9 @@ const ComboboxCustomRenderOptions: ComponentStory<typeof Combobox> = args => {
   const [selectedLineItems, setSelectedLineItems] = useState<ComboboxOption[]>(
     [],
   );
+  const [selectedLineItems2, setSelectedLineItems2] = useState<
+    ComboboxOption[]
+  >([]);
 
   const teamMemberOptions = [
     {
@@ -899,6 +902,33 @@ const ComboboxCustomRenderOptions: ComponentStory<typeof Combobox> = args => {
                     </Box>
                   </Flex>
                 );
+              }}
+            />
+          );
+        })}
+
+        <Combobox.Action
+          label="+ Add new line item"
+          onClick={() => {
+            alert("Added a new line item ✅");
+          }}
+        />
+      </Combobox>
+      <Combobox
+        label="Add line item with default content"
+        subjectNoun="line items"
+        {...args}
+        onSelect={setSelectedLineItems2}
+        selected={selectedLineItems2}
+      >
+        {lineItemOptions.map(o => {
+          return (
+            <Combobox.Option
+              key={o.id}
+              id={`${o.id}`}
+              label={o.label}
+              customRender={({ DefaultContent }) => {
+                return <DefaultContent />;
               }}
             />
           );
--- a/packages/components/src/Combobox/Combobox.test.tsx
+++ b/packages/components/src/Combobox/Combobox.test.tsx
@@ -157,7 +157,9 @@ describe("Combobox", () => {
 
   it("should fire the onSelect when clicking an option", async () => {
     await userEvent.click(screen.getByRole("combobox"));
-    await userEvent.click(screen.getByText("Bilbo Baggins"));
+    await userEvent.click(
+      screen.getByRole("option", { name: "Bilbo Baggins" }),
+    );
 
     expect(handleSelect).toHaveBeenCalledTimes(1);
     expect(handleSelect).toHaveBeenCalledWith([
@@ -245,7 +247,9 @@ describe("Combobox Multiselect", () => {
     await userEvent.click(screen.getByRole("combobox"));
     expect(screen.getByTestId(MENU_TEST_ID)).not.toHaveClass("hidden");
 
-    await userEvent.click(screen.getByText("Bilbo Baggins"));
+    await userEvent.click(
+      screen.getByRole("option", { name: "Bilbo Baggins" }),
+    );
     expect(screen.getByTestId(MENU_TEST_ID)).not.toHaveClass("hidden");
   });
 
@@ -315,7 +319,9 @@ describe("Combobox Multiselect", () => {
 
     await userEvent.click(screen.getByRole("combobox"));
     await userEvent.type(searchInput, "Bilbo");
-    await userEvent.click(screen.getByText("Bilbo Baggins"));
+    await userEvent.click(
+      screen.getByRole("option", { name: "Bilbo Baggins" }),
+    );
 
     expect(searchInput).toHaveValue("Bilbo");
   });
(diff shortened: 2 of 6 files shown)


packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -145,6 +145,7 @@ export interface ComboboxOptionProps {
   readonly customRender?: (
     option: Omit<ComboboxOptionProps, "customRender"> & {
       isSelected: boolean;
+      DefaultContent: () => ReactElement;
```
ZakaryH, line 148: undecided on the syntax & name

`defaultContent` vs `DefaultContent`

`defaultComponent` vs `DefaultComponent`

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -593,6 +599,65 @@ describe("Infinite scroll", () => {
...
+        onSelect={handleSelect}
+      >
+        <Combobox.Option
+          id="1"
+          label="Bilbo Baggins"
+          customRender={({ DefaultContent }) => <DefaultContent />}
+        />
+        <Combobox.Option id="2" label="Frodo Baggins" />
+      </Combobox>,
+    );
```
nad182, line 644: I know that currently we're rendering this only once for this test, but do we want to move this to a separate function, similar to already existing `renderCombobox`, `renderInfiniteScrollCombobox`, `renderCustomOnSearchCombobox` etc.? I think ideally, we should move all those functions to the [POM file](https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Combobox/Combobox.pom.tsx), including various selectors (that use `getByRole`, `getByTestId` etc.), so we could just call them as util functions in this test file. But I understand that this is a bigger ask, so I'm not going to push for that now.

packages/components/src/Combobox/Combobox.types.ts
```diff
@@ -145,6 +145,7 @@ export interface ComboboxOptionProps {
   readonly customRender?: (
     option: Omit<ComboboxOptionProps, "customRender"> & {
```
nad182, line 146: This is a question/suggestion: should we also omit `onClick`? It seems to be "stripped" in the `ComboboxOption` component: https://github.com/GetJobber/atlantis/blob/master/packages/components/src/Combobox/components/ComboboxOption/ComboboxOption.tsx#L9-L10
But in this case it is kind of implied that `onClick` would be present.

nad182: Since this is not part of your change, perhaps, we can leave it be. Just trying to understand (and make sure) that we are typing things correctly.

nad182: I was thinking of a type like: `Omit<ComboboxOptionProps, "customRender" | "onClick">`.

docs/components/Combobox/Web.stories.tsx
```diff
@@ -904,6 +907,33 @@ const ComboboxCustomRenderOptions: ComponentStory<typeof Combobox> = args => {
...
+        selected={selectedLineItems2}
+      >
+        {lineItemOptions.map(o => {
+          return (
+            <Combobox.Option
+              key={o.id}
+              id={`${o.id}`}
+              label={o.label}
+              customRender={({ DefaultContent }) => {
+                return <DefaultContent />;
```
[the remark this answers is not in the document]
nad182: Perhaps, we could add a section in https://github.com/GetJobber/atlantis/blob/master/docs/components/Combobox/Combobox.stories.mdx

Something like:

```mdx
### Wrapping default option content with DefaultContent

Some use cases (e.g., masking option text with a privacy component) need to wrap the option’s default UI inside the portalled listbox. Use `customRender` with `DefaultContent` to wrap without re-implementing option visuals.

<Combobox label="Team" selected={[]} onSelect={() => {}}>
  <Combobox.Option
    id="1"
    label="Jane Doe"
    customRender={({ DefaultContent }) => (
      <PrivacyMask>
        <DefaultContent />
      </PrivacyMask>
    )}
  />
  <Combobox.Option id="2" label="John Smith" />
</Combobox>

Note: Do not call `event.stopPropagation()` or `event.preventDefault()` in the wrapper, as selection and menu behavior are handled by the option element.
```

nad182 (changes requested):
Finished the first review pass. I didn't find any major issues so far, but I left several suggestions and questions. Let me know what you think.

packages/components/src/Combobox/Combobox.types.ts
ZakaryH: sounds good, we'll go with `defaultContent`

reason for the type was because we do want it to be more restrictive. `ReactNode` has things like string, but we know this is an actual component/element

I'll see about changing the type but initially that was intentional since we know it's never going to be most of the possible types of ReactNode

docs/components/Combobox/Web.stories.tsx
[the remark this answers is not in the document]
nad182: We don't need to add that component, since this is just an example/code snippet in the docs. But that's not a "hard requirement" to add it.

packages/site/src/content/Combobox/Combobox.props.json
```diff
@@ -306,7 +306,7 @@
         },
         "required": false,
         "type": {
-          "name": "(option: Omit<ComboboxOptionProps, \"customRender\"> & { isSelected: boolean; }) => ReactNode"
+          "name": "(option: Omit<ComboboxOptionProps, \"customRender\"> & { isSelected: boolean; DefaultContent: () => ReactElement<any, string | JSXElementConstructor<any>>; }) => ReactNode"
```
nad182, line 309: This is "old" version of the prop name. Should the docs be re-generated?

nad182 (changes requested):
Left minor suggestions. 

Also CI fails:

```
Found uncommitted changes which is unexpected. This likely means you forgot to commit generated files.
Here's the diff:
On branch JOB-133583/provide-default-component
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   packages/site/src/content/Combobox/Combobox.props.json

no changes added to commit (use "git add" and/or "git commit -a")
If the diff includes *.props.json files, run 'npm run generate' inside packages/site, review and commit the changes.

Exited with code exit status 1
```

I think it might be related to regenerating the docs.

packages/components/src/Combobox/Combobox.test.tsx
```diff
@@ -593,6 +599,45 @@ describe("Infinite scroll", () => {
...
+      multiSelect: true,
+      selected: [{ id: "1", label: "Bilbo Baggins" }],
+    });
+
+    await userEvent.click(screen.getByRole("combobox"));
+    // Default content reflects selection state
+    expect(
+      screen.getByRole("option", { name: "Bilbo Baggins" }),
+    ).toHaveAttribute("aria-selected", "true");
+    expect(screen.getByTestId("checkmark")).toBeInTheDocument();
```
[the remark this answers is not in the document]
ZakaryH: unfortunately there's a bit more juggling and moving things around to share it, let's save this for a future test/POM cleanup.

PR #2707, opened 2025-09-02
fix(hooks): Prevent mockViewportWidth from accessing window until necessary

jdeichert:
## Motivations

This was causing some problems with vitest suites that use `node` env instead of the default jsdom environment. During the test suite bootup process (before tests run), imports of `@jobber/hooks` barrel file cause this file to load and in the vitest `node` environment, `window` doesn't exist so it crashes the suite.

### Fixed

- Only access window when necessary

---


edison-cy-yang (approved):
The change is straight forward and makes sense. LGTM

PR #2718, opened 2025-09-05
feat(components): Update Combobox CSS to limit its width to the parent/container width

nad182:
## Motivations

Combobox activator could keep growing in width and overflow its parent container. We should make sure it respects its parent width.

### Added

- 

### Changed

- Set max-width to 100% for the wrapper and combobox role (Chip inside of Combobox) to ensure proper responsiveness.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

### Before
image

### After
image

## Testing

- Set max-width on the parent `div` (e.g. `300px` or `400px`)
- Select multiple items (NOTE: `Combobox` should be `multiSelect`).
- Make sure the activator doesn't grow beyond the parent's width and the content is truncated
- Test in the actual product (e.g. [TeamMemberSelect](https://github.com/GetJobber/jobber-frontend/blob/main/apps/jobber-online/app/shared/users/TeamMemberSelect.tsx#L42-L51), which is used in the RequestNewAssessment form in [RequestNewAssessmentFields](https://github.com/GetJobber/jobber-frontend/blob/main/apps/jobber-online/app/features/requests/new/requestForm/assessment/RequestNewAssessmentFields.tsx#L81-L86))

Changes can be

---


packages/components/src/Combobox/Combobox.module.css
```diff
@@ -10,4 +10,9 @@
 
 .wrapper {
   display: inline-block;
+  max-width: 100%;
+}
+
+.wrapper > [role="combobox"] {
```
ZakaryH, line 16: I'd prefer to not have to also apply this to the child. if we could set it solely on the parent wrapper, that would be great.

ZakaryH:
can we add a visual test for this please? can just be something that has the described scenario. a constrained parent width, and then we have a whole bunch of selections that would cause the default activator chip to be much wider than that. I expect the screenshot would should truncated chip, that doesn't exceed the parent width

PR #2720, opened 2025-09-09
fix(components): Improve breakpoint handling for ContentBlock, Cluster and Sidekick 

Aiden-Brine:
## Motivations

I was trying to use `ContentBlock` and I noticed some small bugs with the breakpoints on `ContentBlock`, `Cluster` and `SideKick`. One of the options in `AtlantisBreakpoints` for this component was "xs" = 0px which doesn't make sense for a `maxWidth` or `collapseBelow` prop. Furthermore, you couldn't actually set something to `xs` because the following logic was used:
```
  if (AtlantisBreakpoints[maxWidth as keyof typeof AtlantisBreakpoints]) {
    return (
      AtlantisBreakpoints[maxWidth as keyof typeof AtlantisBreakpoints] + "px"
    );
```
Since `AtlantisBreakpoints[maxWidth as keyof typeof AtlantisBreakpoints] = 0` and 0 is falsey this conditional wouldn't actually be entered for `xs`. 

### Removed

- Removed the `xs` breakpoint option

### Fixed

- The conditional so that falsey breakpoint values wouldn't prevent the conditional from being entered

### Changed

- The default value `maxWidth` on `ContentBlock` from `Breakpoints.smaller` which has a value of 265 and is not a valid `AtlantisBreakpoints` value to `"sm"`. I went through the JFE and JO code to ensure this default value change wouldn't impact any implementations

## Testing

Play around with these components in Storybook to ensure they work as expected. 

Changes can be

---

### DIFF
--- a/packages/components/src/ContentBlock/ContentBlock.tsx
+++ b/packages/components/src/ContentBlock/ContentBlock.tsx
@@ -1,6 +1,5 @@
 import React from "react";
 import classNames from "classnames";
-import { Breakpoints } from "@jobber/hooks/useResizeObserver";
 import styles from "./ContentBlock.module.css";
 import type { ContentBlockProps } from "./types";
 import { getMappedAtlantisSpaceToken } from "../sharedHelpers/getMappedAtlantisSpaceToken";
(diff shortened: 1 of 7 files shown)


scotttjob (approved):
Changes look good to me from a high level fly by. Appreciate the cleanup + bug fixes.

Looks like you have some failing e2e tests, but I imagine you just need to regenerate with these changes (although I'm curious what the failing visual tests have caught from the layout updates)

PR #2724, opened 2025-09-10
feat(components): accessible naming and a11y updates for Modal v1/v2

nad182:
## Motivations

- Ensure `Modal` is accessibly named for assistive technologies and easier testing.
- Align naming precedence with accessibility guidance: prefer visible heading via `aria-labelledby`, fall back to `aria-label` when no visible title.

### Before

image

### After
 
image

### Added

- Modal v2 (composable):
  - `accessibleName` on `Modal.Provider`. Dialog now has `aria-modal="true"` and accessible name: uses `aria-labelledby` when a header/title exists; otherwise `aria-label` via `accessibleName`.
- Modal v1 (legacy):
  - `accessibleName` on `Modal`. Dialog now has `aria-modal="true"`. Names via `aria-labelledby` when `title` exists; otherwise `aria-label` via `accessibleName`.
- Tests (v1 & v2):
  - Assertions for naming precedence (title/header vs `accessibleName`).
- Docs (site):
  - Accessibility section describing role, `aria-modal`, naming precedence, custom header id, and testing tip.

### Changed

- Modal v2: Respect `dismissible={false}` by preventing close on Escape and outside click. (Keeps keyboard+overlay dismissal disabled only when explicitly non-dismissible.)

### Deprecated

- N/A

### Removed

- N/A

### Fixed

- N/A

### Security

- N/A

## Testing

- Unit tests cover both versions:
  - Title/header present → `screen.getByRole('dialog', { name: /title/i })`.
  - No title/header but with `accessibleName` → `screen.getByRole('dialog', { name: /accessible name/i })`.
- Updated snapshots where dialog attributes changed (e.g., `aria-modal`, `aria-labelledby`, header `id`).

---

## Follow-ups for Reviewer Consideration (not included to keep scope focused)

- v2: Only set `aria-labelledby` when a header is actually rendered
  - Track header presence (register id in context on mount) and set `aria-labelledby` conditionally; set `aria-label` only when no header exists. (Current implementation also sets `aria-label` when `accessibleName` is provided; compliant, but could be cleaner.)

- v1: Optionally gate Escape-to-close behind `dismissible={false}`
  - v2 already respects `dismissible={false}` for Escape and outside click. Matching v1 would be more consistent, but could be behavior-changing for some consumers.

- `aria-describedby` support
  - Allow passing a description id (or slot) to set `aria-describedby` for longer descriptive text.

## Notes on Backward Compatibility

- v1: No breaking behavior changes; accessible naming added, `aria-modal` added.
- v2: `dismissible={false}` now blocks Escape/outside click close; this aligns with the prop’s intent. Consumers relying on Escape/overlay to close while using `dismissible={false}` may notice the change (uncommon case).


packages/components/src/Modal/useModal.ts
```diff
@@ -37,8 +39,9 @@ export function useModal({
 
   const click = useClick(floatingContext);
   const dismiss = useDismiss(floatingContext, {
+    outsidePress: dismissible,
     outsidePressEvent: "click",
-    escapeKey: true,
+    escapeKey: dismissible,
     bubbles: false,
   });
   const role = useRole(floatingContext);
```
ZakaryH, line 47: we do already have this `useRole`, though  I see that we're not using `useModal` in the legacy modal which is unfortunate

nad182: The intent is that in v2, `dismissible={false}` means “don’t allow user-initiated dismissal,” so Escape and outside click are disabled (we now set `escapeKey: dismissible` and `outsidePress: dismissible`). For v1, I’m leaving Escape behavior as-is to avoid a breaking change. If you’d like parity, I can follow up with either:
- optional flags (closeOnEscape, closeOnOutsideClick) defaulting to current behavior, or
- gate Escape by `dismissible={false}` in v1 (noting this would be behavior-changing).

packages/components/src/Modal/Modal.rebuilt.tsx
```diff
@@ -128,12 +128,12 @@ export function ModalContent({ children }: ModalContainerProps) {
...
                   <div
                     ref={floatingRefs?.setFloating}
-                    role="dialog"
-                    tabIndex={0}
-                    aria-modal="true"
-                    aria-labelledby={modalLabelledBy}
-                    aria-label={accessibleName}
-                    {...getFloatingProps()}
+                    {...getFloatingProps({
+                      tabIndex: 0,
```
ZakaryH, line 132: do we even need to set the tabIndex ourselves?

floating has opinions on this and by using all their hooks, we've kinda put them in the driver's seat when it comes to focus management.

packages/components/src/Modal/useModal.ts
ZakaryH: can I ask why we're modifying that behavior in this PR? to me that seems like a much larger scope than adding some accessibility attributes.

ZakaryH: gotcha, fair enough. 

yeah I feel like this is worthy of some thought and design input.

in other libraries `dismissible` means that clicking outside doesn't close
https://react-spectrum.adobe.com/react-aria/Modal.html#dismissable

then a separate prop for opting out of esc press

whereas for us, it simply means that the little X icon isn't in the UI on the top right
image

disabling both clicking outside and esc presses with `dismissible={false}` in addition to our UI removal can put people into an impossible situation. there's no guarantee of a "close" or "cancel" button like a ConfirmationModal for example. so a user may literally have no way to get rid of the Modal once it's shown if dismissible encompasses all 3 of those behaviors

ZakaryH:
just noticed something on the other modal PR

https://github.com/GetJobber/atlantis/pull/2723#issuecomment-3276332077

basically our setup for v2 at least, is a bit odd. we have both FloatingOverlay and ModalOverlay. we only need one.

then we have our role='dialog'  assigned to something that is really not a dialog. it's a container for both the 2nd overlay and the actual modal. the ModalOverlay has no need to get the floating position. it's just going to ignore that anyways.

I wonder if this will have, or currently has, strange results if you query for the modal and expect it to immediately contain you r content but it's actually inside a child of that "modal". I'm pretty sure this is causing focus behavior to be weird, where it should actually go into the modal but it doesn't because we're effectively telling floating ui that some other element is the modal by spreading the floating props onto it and assigning it the floating ref

I think something like this is all we need 
https://codesandbox.io/p/sandbox/stoic-bas-frzus0?file=%2Fsrc%2FApp.tsx

if this turns into too much, then don't worry about it. we can make a ticket to solve this later. 

for the purposes of this task, I guess just make sure that when we write some tests that try to target these new attributes, that they are able to pass and make their assertions without getting confused by the fact that the thing they've been told is a modal, isn't strictly speaking actually a modal 😅

packages/components/src/Modal/Modal.types.ts
```diff
@@ -30,6 +30,16 @@ export interface ModalProviderProps {
    */
   readonly activatorRef?: MutableRefObject<HTMLElement | null> | null;
   readonly dismissible?: boolean;
+  /**
+   * Id to provide aria-labelledby to the modal. If you are using Modal.Header with children or another header you will need to provide this as the id of the header.
+   * @default "ATL-Modal-Header"
+   */
+  readonly modalLabelledBy?: string;
```
ZakaryH, line 37: I'm confused. why do we need to add this? this already exists and works no?

packages/components/src/Modal/Modal.rebuilt.test.tsx
```diff
@@ -86,6 +86,32 @@ describe("Composable Modal", () => {
...
+      <Modal.Provider open={true} accessibleName="Add Customer">
+        <Modal.Content>{/* no Modal.Header on purpose */}</Modal.Content>
+      </Modal.Provider>,
+    );
+
+    expect(
+      await screen.findByRole("dialog", { name: "Add Customer" }),
+    ).toBeInTheDocument();
+  });
+
```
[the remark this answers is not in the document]
taylorvnoj: > what happens if you have both a Header and an accessible label provided?

using voiceover:

1. `ariaLabel` & `Heading` - reads `ariaLabel`
2. no `ariaLabel`, `Heading` present - reads `Heading`
3. `ariaLabel` & `Heading` NOT present - reads `ariaLabel`

So when both are present, `ariaLabel` wins with the current code, no enforcement.

We can also do 

```
 ...(ariaLabel
 ? { "aria-label": ariaLabel }
 : { "aria-labelledby": modalLabelledBy }),
```

to enforce the intended behaviour. Then we know we are testing something that is sure to happen, because we said so.

packages/components/src/Modal/Modal.tsx
```diff
@@ -37,6 +39,9 @@ export function ModalLegacy({
           role="dialog"
           className={styles.container}
           tabIndex={0}
+          aria-modal="true"
+          aria-labelledby={title != undefined ? headerId : undefined}
+          aria-label={title != undefined ? undefined : accessibleName}
```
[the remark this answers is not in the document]
taylorvnoj: > if you gave us both a title and an aria-label, we'll assume you meant to do that and the aria label takes priority

Going to suggest we don't do this, as it would go against standards and what the browser behaviour does on it's own ([mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby#description)) - which I think we're on the same page about

> we could enforce the presence of the ariaLabel if no title is provided...that's probably more than we need to do here though.

Not saying we should never do that, but agree it's not something we should do here. Considerations would be that's a potential breaking change for existing `Modals` that don't provide either

packages/components/src/Modal/Modal.types.ts
taylorvnoj: removed `ModalProviderProps` from the types file here 073b526ffbe92c729f8d0ecff79cc5a2a2a11114

I can make a follow up task to move types and export properly

packages/components/src/Modal/Modal.rebuilt.test.tsx
```diff
@@ -76,41 +76,153 @@ describe("Composable Modal", () => {
...
+  describe("accessibility", () => {
+    it('modal contains aria role of "dialog"', async () => {
+      render(
+        <Modal.Provider open={true}>
+          <Modal.Content>
+            <Modal.Header title="Modal Title" />
+          </Modal.Content>
+        </Modal.Provider>,
+      );
+      expect(await screen.findByRole("dialog")).toBeInTheDocument();
```
ZakaryH, line 108: we actually did this in the test above, but since it's just always open  we don't need await/findBy 

it's not really doing anything to use findBy but let's just handle it consistently. my preference would be no findBy if we have nothing to wait for.

ZakaryH: since we're dealing with animations, `toBeVisible` is preferable

packages/components/src/Modal/Modal.rebuilt.test.tsx
```diff
@@ -76,41 +76,153 @@ describe("Composable Modal", () => {
...
+    });
 
-    expect(
-      await screen.findByRole("dialog", { name: "Billing Settings" }),
-    ).toBeInTheDocument();
-  });
+    describe("dialog is named from title prop", () => {
+      it("when ariaLabel is not provided", async () => {
```
ZakaryH, line 112: nitpick: the describe and it should be flipped

"when" and other qualifiers are setting the context. "it should..." is the test title, that way it reads as a natural sentence and serves as a layer of free documentation

packages/components/src/Modal/Modal.rebuilt.test.tsx
```diff
@@ -76,41 +76,153 @@ describe("Composable Modal", () => {
...
+            </Modal.Content>
+          </Modal.Provider>,
+        );
+
+        const dialog = await screen.findByRole("dialog");
+
+        expect(dialog).toBeInTheDocument();
+        expect(
+          screen.getByRole("dialog", { name: "Billing Settings" }),
+        ).toBeInTheDocument();
```
ZakaryH, line 143: finding the dialog twice in 2 different ways doesn't really tell us anything. let's just query for it with the method that is most relevant to this test, that it has the precise name we are looking for

packages/components/src/Modal/Modal.rebuilt.test.tsx
```diff
@@ -76,41 +76,153 @@ describe("Composable Modal", () => {
...
-    ).toBeInTheDocument();
+        const dialog = await screen.findByRole("dialog");
+
+        expect(
+          screen.getByRole("dialog", { name: "Add Customer" }),
+        ).toBeInTheDocument();
+        expect(dialog).toHaveAttribute("aria-label", "Add Customer");
+      });
+
+      it("when header lacks id/modalLabelledBy", async () => {
```
ZakaryH, line 164: our "test" or "it" should be a helpful description of the behavior we are asserting. that way when a test fails it tells the reader something about what went wrong

ie. "it should be identifiable by ariaLabel"

"when header lacks id/modalLabelledBy" only tells me the conditions, not what to expect. I think this would be better suited for a "describe"

packages/site/src/content/Modal/ModalNotes.mdx
```diff
@@ -131,3 +131,67 @@ function ModalWithCustomFocus() {
...
+`aria-modal="true"`.
+
+**Naming precedence**:
+
+- If a visible heading/title is provided, the dialog is named via
+  `aria-labelledby` that references the heading element.
+- If there is no visible heading/title, you can provide `accessibleName` to set
+  `aria-label` on the dialog.
+
+### Composable Modal (v2)
```
ZakaryH, line 147: we haven't been referring to the composable versions of components as v2 so far. let's stick to that convention.

packages/site/src/content/Modal/ModalNotes.mdx
```diff
@@ -131,3 +131,67 @@ function ModalWithCustomFocus() {
...
+// No title → fallback to ariaLabel
+<Modal.Provider open ariaLabel="Add Customer">
+  <Modal.Content>{/* custom content without heading */}</Modal.Content>
+</Modal.Provider>
+
+// Custom header → provide matching id via modalLabelledBy
+<Modal.Provider open modalLabelledBy="custom-header-id">
+  <Modal.Content>
+    <Modal.Header>
+      <h2 id="custom-header-id">My Custom Header</h2>
```
ZakaryH, line 171: we just added id to our Heading component didn't we? let's use that as an example to show it off.

packages/site/src/content/Modal/ModalNotes.mdx
```diff
@@ -131,3 +131,67 @@ function ModalWithCustomFocus() {
...
+    </Modal.Header>
+  </Modal.Content>
+</Modal.Provider>
+```
+
+### Legacy Modal (v1)
```
ZakaryH, line 177: similar to the v2 feedback we haven't been calling the existing version v1/legacy - at least not externally. if we wanna use that terminology inside the components and tests, that fine but let's avoid calling it legacy publicly

we have plans for that term, and it suggests that someone shouldn't use this version. I don't think that's our recommendation as far as I'm aware.

packages/components/src/Modal/Modal.tsx
```diff
@@ -60,6 +65,7 @@ export function ModalLegacy({
           >
             {title != undefined && (
               <Header
+                id={headerId}
```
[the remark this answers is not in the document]
ZakaryH: right. ok then the only weird thing is that we have a different test-id for the old one vs new. unfortunate but we can fix that some other time.

packages/site/src/content/Modal/ModalNotes.mdx
ZakaryH: ah wait I see we are doing that in here.... I'm not sure I agree with this. I'm going to tweak the copy so we don't have to refer to the old one as legacy.

it's also not really accurate to say that original Modal is not composable and the new one is. it's more composable, for sure but even it has restrictions ie. the actions.

docs/content/formatting.stories.mdx
ZakaryH, line 1: these are odd, but harmless changes. looks like formatting on save/commit is all.

packages/components/src/Modal/Modal.types.ts
```diff
@@ -118,8 +118,8 @@ export interface ModalLegacyProps {
   onRequestClose?(): void;
   readonly version?: 1;
   /**
-   * Accessible name override applied to the modal via aria-label. When provided, it takes
-   * precedence over using the title.
+   * Accessible name for the Modal.
+   * Only required if no title is provided. Title takes precedence over ariaLabel.
```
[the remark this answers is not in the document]
ZakaryH: fair. it's required in the sense that it's a best practice. we could enforce that with some XOR types, though that's fairly aggressive. most libraries I've seen that have these sorts of requirements either leave that in documentation, or do a console.warn

I'd be open to a console.warn

though we won't be able to do that with the composable version so easily, or at all in one scenario. we can know if they gave us the basic Header, but we have no idea if their custom content has what it needs.

so at best, we can warn for 2/3 scenarios (old missing title, new missing Header title)

ZakaryH (approved):
LGTM

PR #2729, opened 2025-09-12
fix(components): InputGroup horizontal outer borders should remain rounded

nad182:
## Motivations

Border radius on `InputGroup` with `flowDirection="horizontal"` became square when siblings existed before/after it or when the parent used `flex-direction: column`. Nested horizontal groups could also show rounded inner seams.

### Before

image

### After

image

### Changed
- `packages/components/src/InputGroup/InputGroup.module.css`:
  - Always restore outer rounding for horizontal groups via first/last child rules.
  - Preserve existing vertical first/last rounding.
  - Add minimal nested-horizontal flattening to avoid inner rounding when a nested group sits mid-row.

### Fixed
- Horizontal `InputGroup` outer corners stay rounded regardless of surrounding siblings or parent layout (row/column).
- Nested horizontal groups no longer show rounded inner seams.

## Testing

- Storybook → Components/InputGroup:
  - Verify DateRange (horizontal) with elements before/after retains rounded outer corners.
  - Wrap in a parent with `display: flex; flex-direction: column;` and re-verify.
  - Add a nested horizontal `InputGroup` as first/middle/last child and confirm only true outer edges are rounded.

No runtime/prop changes. Pure CSS; scoped to `InputGroup` only.


packages/components/src/InputGroup/InputGroup.module.css
```diff
@@ -30,22 +30,27 @@
...
+/* If the nested group is not at the left edge, flatten its first field's left corners */
+.horizontal > .horizontal:not(:first-child) > :first-child {
+  --public-field--top-left-radius: 0;
+  --public-field--bottom-left-radius: 0;
 }
 
-.horizontal:last-child > :last-child {
-  --public-field--bottom-right-radius: var(--radius-base);
+/* If the nested group is not at the right edge, flatten its last field's right corners */
+.horizontal > .horizontal:not(:last-child) > :last-child {
```
taylorvnoj, line 53: I think we're going to need more rules. If you look in the Nested story - a vertical group containing horizontal groups)  it seems like the css is giving rounded corners to the inner fields of the horizontal groups.

Screenshot 2025-09-17 at 8 43 31 PM

I think we need...
- if a horizontal group is the first child in a vertical group, give first field rounded top left corner/if horizontal group is last child give last field rounded bottom right corner
  - and if horizontal group is in the middle of a vertical group, its fields get no rounded corners

taylorvnoj (changes requested):
Looking at the nested example exposed some css holes unfortunately - problem being vertical groups that have horizontal groups

docs/components/InputGroup/Web.stories.tsx
```diff
@@ -85,3 +90,31 @@ export const DateRange = DateRangeTemplate.bind({});
 DateRange.args = {
   flowDirection: "horizontal",
 };
+
+const VerticalWithHorizontalRowsTemplate: ComponentStory<
```
taylorvnoj, line 94: Appreciate the story for testing but I don't think we need this added, the existing Nested story covers a vertical group with horizontal groups inside. I did put a Button before and after the wrapping `InputGroup` though and all looked good 👌🏻

taylorvnoj:
@nad182 should we update `VisualTestInputGroupPage.tsx`- I feel like having the vertical group with two horizontal groups should be defined in that file because maybe that would have been caught?

taylorvnoj (approved):
I tried to update the visual tests and maybe it's my machine but it took a looong time and then stalled out. If you can update the visual tests faster than me great - but I'm going to approve and we can add visual tests as a follow up

PR #2732, opened 2025-09-16
fix(components): ComboboxAction visibility logic to properly handle false values

taylorvnoj:
## Motivations

The `ComboboxAction` component's `visible` prop was not working correctly when set to `false`. The visibility logic had a bug where it only checked visibility when `props.visible` was truthy, causing `visible={false}` to be completely ignored and actions to always render.

### Added

- 

### Changed

Fixed visibility logic in `ComboboxAction.tsx` to properly handle boolean false values
✅ visible={true} → Shows action
✅ visible={false} → Hides action
✅ visible={function} → Shows/hides based on function result
✅ No visible prop → Shows action by default

## Testing

- Render `visible={false]` in any story with `Combobox.Action` and see that the action is not visible
- Render `visible={true}` in any story and confirm the action is visible
- Render without visible prop and confirm the action is visible by default
- Test with `visible={() => false}` (function returning false)
- Test with `visible={() => true}` (function returning true)
- Check that the `Dynamic Action` story still works as intended
- Check that the `Keep Open On Click` story still works as intended

Changes can be

---

### DIFF
--- a/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.test.tsx
+++ b/packages/components/src/Combobox/components/ComboboxAction/ComboboxAction.test.tsx
@@ -66,4 +66,49 @@ describe("ComboboxAction", () => {
 
     expect(handleClose).toHaveBeenCalledTimes(1);
   });
+
+  describe("visible prop", () => {
+    it("renders the action when visible is not provided", () => {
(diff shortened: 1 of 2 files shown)


taylorvnoj:
Looked through product with these changes and there is no impact to current instances 👍🏻

PR #2739, opened 2025-09-17
feat(components): Upgrade react-markdown to 10.1.0

edison-cy-yang:
## Motivations

Upgrade react-markdown to the latest version 10.1.0 which would
- Unblock React 19
- Get rid of deprecation warning about `defaultProps` when running tests

### Added

- 

### Changed

- Upgraded react-markdown to 10.1.0

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,10 +38,7 @@ references:
   npm_install: &npm_install
     run:
       name: Install Dependencies
-      command: |
-        npm ci
-        cd packages/storybook-v7
-        npm ci
+      command: npm ci
 
   check_git_status: &check_git_status
     run:
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 coverage
 dist
 node_modules
+node_modules.e2e/
 **/tsconfig.tsbuildinfo
 **/tsconfig.*.tsbuildinfo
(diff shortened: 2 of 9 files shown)


packages/components/src/Markdown/Markdown.tsx
```diff
@@ -63,7 +63,11 @@ export function Markdown({
     <Tag>
       <ReactMarkdown
         {...props}
-        linkTarget={externalLink ? "_blank" : undefined}
+        urlTransform={url => {
```
edison-cy-yang, line 66: react-markdown 9.0.0 started sanitizing protocols in URL. Allowed protocols are `http`, `https`, `irc`, `ircs`, `mailto`, and `xmpp`. 

I believe allowing tel and sms this way is sufficient. The [implementation ](https://github.com/remarkjs/react-markdown/blob/main/lib/index.js#L425) of `defaultUrlTransform` really just checks that:
- it has a colon (it is a protocol)
- question mark, number sign, and slash are not before the first colon
- it is one of the 6 listed protocols

Checking with `url.startsWith("tel:")` should cover the above cases. But we could also take inspiration from their implementation as well.

PR #2742, opened 2025-09-19
feat(components): Upgrade RAC to 1.11.0

ZakaryH:
## Motivations

In order to use RAC's Menu component with Tanstack Router, we need to be on a higher version
https://react-spectrum.adobe.com/react-aria/routing.html#tanstack-router
image

we know that people will want to use Menu to navigate, and the way to do this with client side routing is as shown. we do not want to have our Menu know about TSR, so that `createLink` component will live in the project that requires it.

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

we only have 1 component that uses RAC currently, the InputNumber v2. so all we're looking for is making sure the build builds, and that the features in InputNumber v2 are all good. that said this is a minor version bump, so nothing too concerning to watch out for.

Changes can be

---


ZakaryH:
technically we're actually going to 

```
    "packages/components/node_modules/react-aria-components": {
      "version": "1.12.2",
```

because of the caret. not a bad thing, there are some neat new features that we can use and bug fixes etc. just my PR title isn't exactly true.

edison-cy-yang (approved):
skimmed through package-lock changes and nothing stood out. LGTM

PR #2743, opened 2025-09-19
feat(components): Upgrade Framer Motion Version

ZakaryH:
## Motivations

to use RAC's Menu, and have the animations that we currently have on our Menu, that are already provided by framer-motion, we need to upgrade our framer-motion version to properly integrate with RAC

https://react-spectrum.adobe.com/react-aria/styling.html#framer-motion
https://motion.dev/docs/react-motion-component#custom-components

we don't necessarily need to go to ^11.11.12 (resolving to 11.18.2), it appears to have been added in 11.5.0 

https://github.com/motiondivision/motion/blob/main/CHANGELOG.md#1150-2024-09-04

but we also need to upgrade framer-motion for React 19 compatibility anyway, and that starts from 11.13.5 onwards

https://github.com/motiondivision/motion/blob/main/CHANGELOG.md#fixed-94

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

motion animations should still work. we use these in quite a few places. unfortunately they are a visual thing, you've gotta take a peek at a few of them to see how they're looking.

Changes can be

---

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23505,13 +23505,6 @@
       "integrity": "sha512-49Kt+HKjtbJKLtgO/LKj9Ld+6vw9BjH5d9sc40R/kVyH8GLAXgT42M2NnuPcJNuA3s9ZfZBUcwIgpmZWGEE+hA==",
       "license": "MIT"
     },
-    "node_modules/mri": {
-      "version": "1.2.0",
-      "license": "MIT",
-      "engines": {
-        "node": ">=4"
-      }
-    },
     "node_modules/ms": {
       "version": "2.1.3",
       "resolved": "https://registry.npmjs.org/ms/-/ms-2.1.3.tgz",


package-lock.json
```diff
@@ -22975,6 +22977,39 @@
         "url": "https://www.patreon.com/infusion"
       }
     },
+    "node_modules/framer-motion": {
+      "version": "11.18.2",
```
ZakaryH, line 22981: I had to manually delete the package-lock's `framer-motion` version for this to pick up. it just gets confused and unable to resolve the deps, resulting in no change.

we've got some problems with our package dependencies right now. I've got a separate PR to deal with that that should unblock this properly if we're not comfortable with this approach.

while inadvisable, the package lock is generated on `npm i` so this should be in the correct state now and you'd never know I manually edited it if I hadn't said so

packages/components/src/DataList/components/DataListItem/DataListItem.tsx
```diff
@@ -81,6 +81,7 @@ export function DataListItem<T extends DataListObject>({
   }
 
   function handleHideMenu() {
+    if (isContextMenuVisible) return;
```
ZakaryH, line 84: this is to address the `should not close the menu when hovering out of the target` becoming flakey

it is a real code change to DataList. not entirely ideal. however it seems pretty reasonable to me to do.


>Context menu is a persistent UI triggered by a deliberate user action (right-click) and requires coordinates. Therefore, contextPosition is the durable “open” intent. Hover is transient and should not close a persistent menu.
The flake came from event ordering: right-click often fires a mouseleave immediately after, which set showMenu=false and hid the menu while Motion was mid-enter/exit. Guarding handleHideMenu when the menu is open prevents this race without changing visible behavior.
Close semantics remain explicit and covered: Escape, overlay click, and action click all clear contextPosition (closing the menu). After close, the next blur/mouseleave will set showMenu=false, so hover actions don’t remain sticky.
In short: contextPosition is the source of truth for “context menu open” because it encodes both the user intent and the placement. showMenu remains the source of truth for “hover actions” intent. The guard simply ensures the transient hover signal can’t override the persistent context menu while it’s open.

ZakaryH: as further evidence to support this theory, that the actual code is the problem - if you go onto master and edit the `DataListItems.test.tsx` to this

```
it("should not close the menu when hovering out of the target", async () => {
      renderComponent();
      const listItem = screen.getByText(mockData[0].label);
      await userEvent.hover(listItem);
      fireEvent.contextMenu(listItem);

      await userEvent.unhover(listItem);
      const menuElement = await screen.findByRole("menu");
      await waitForElementToBeRemoved(menuElement);
      expect(menuElement).toBeInTheDocument();
    });
```

you'll see that it fails. this is because the menu is animated and we are immediately making this assertion

`expect(screen.getByRole("menu")).toBeInTheDocument(); `

so if we don't wait for anything, then we are in that period of time where the animation is still in progress and the element is there. if we waited just a moment later, or used the `waitForElementToBeRemoved` like this then we'd see that this test has been erroneously passing for its entire existence

scotttjob: Okay to summarize to help clarify my understanding:

Due to the framer motion update, handling when to Hide the DataList (in a test) became flakey.

To prevent the flake, you're now saying if we've explicitly said to keep the context menu open, to not hide the context menu when a transient 'hide' is triggered (like a hover out).

Is this a change we should communicate to users ahead of time, or will it be something that most users won't notice?

edison-cy-yang (approved):
lgtm, good work on making sure unit test are not flakey and steered away from unrealistic code paths!

PR #2748, opened 2025-09-22
build: Auto install storybook v7

jdeichert:
## Motivations

Before this change, if you `npm run storybook` from the root, it would auto install. This however isn't clear and it could be easier. Right now it's causing some problems locally because of this.. the site package fails to build when storybook-v7 isn't installed.

Now with `npm install` at the root, postinstall will auto install storybook-v7 so we never have to think about this.

### Fixed

- Auto install storybook-v7 

## Testing

1. Pull this branch
2. `rm -rf packages/storybook-v7/node_modules`
3. `npm install` at the root
4. `cd packages/site && npm run dev`
    * make sure the docs site runs
5. `npm run test:visual:update`
    * make sure these are successful (no missing deps errors)
6. `cd ../..` go to the root
7. `npm run storybook`
    * make sure storybook v7 runs

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 coverage
 dist
 node_modules
+node_modules.e2e/
 **/tsconfig.tsbuildinfo
 **/tsconfig.*.tsbuildinfo
 .DS_Store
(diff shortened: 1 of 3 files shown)


taylorvnoj (approved):
LGTM - thanks for the help with visual tests!

PR #2754, opened 2025-09-25
fix(components-native): Keep the zindex stable between both regular and mini label styles

jhorvat:
## Motivations

McMobile is working on migrating Jobber Mobile to [React Native's New Architecture](https://reactnative.dev/architecture/landing-page). The new arch brings with it a new rendering engine, [Fabric](https://reactnative.dev/architecture/fabric-renderer). It is mostly a drop in replacement, but in some edge cases behaviours have changed and we need to adapt. 

In this case the `zIndex` in the `miniLabel` style was causing focus to be lost on `InputText` fields when transitioning from 0 characters to 1 on iOS.

https://github.com/user-attachments/assets/e8e7b277-8f9d-4538-9111-d22bd07a67db

### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

- Moved the `zIndex` to the base `Placeholder` style to keep it stable regardless of whether its in its big or mini form 

### Security

- 

## Testing

I tested this change on both the new and old architectures of RN, both behave as expected

#### New Arch

https://github.com/user-attachments/assets/98069347-6288-40bf-be67-740d926fa766

#### Old Arch

https://github.com/user-attachments/assets/9dd53d17-2824-4dc6-8ad8-76f24736edbd

Changes can be

GetJobber/jobber-mobile#11113 will have a prerelease with this version integrated. 

---

### DIFF
--- a/packages/components-native/src/InputFieldWrapper/InputFieldWrapper.style.ts
+++ b/packages/components-native/src/InputFieldWrapper/InputFieldWrapper.style.ts
@@ -43,6 +43,7 @@ export const useStyles = buildThemedStyles(tokens => {
       right: 0,
       bottom: 0,
       left: 0,
+      zIndex: 1,
     },
 
     miniLabel: {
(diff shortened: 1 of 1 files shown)


andrewsjobber (commented):
Looking good on our side, great finding!
If Atlantis is good with that change, I'm good to come back and approve 👀

PR #2765, opened 2025-10-03
feat(components): Swap Menu Navigation to Arrow Keys

ZakaryH:
… test

## Motivations

because the new Menu uses arrow keys like we've always wanted, it would be good for the old version to work the same way. it would be a bit strange, and annoying for them to be different when they look identical visually and conceptually.

also it's something we've wanted as mentione

### Added

- 

### Changed

Menu is now navigated with arrow keys instead of tabs. arrow keys also open the Menu, but do not close them. only ESC, selections, or clicking outside will close.

we no longer use the focus trap because that's for Tabs.

the first item continues to receive focus on open when using keyboard.

### Deprecated

- 

### Removed

- 

### Fixed

- 

### Security

- 

## Testing

Changes can be

---

### DIFF
--- a/packages/components/src/Menu/Menu.tsx
+++ b/packages/components/src/Menu/Menu.tsx
@@ -1,4 +1,4 @@
-import type { MouseEvent, ReactElement, RefObject } from "react";
+import type { MouseEvent, ReactElement } from "react";
 import React, {
   createContext,
   useContext,
@@ -9,7 +9,6 @@ import React, {
 import classnames from "classnames";
 import { AnimatePresence, motion } from "framer-motion";
 import {
-  useFocusTrap,
   useIsMounted,
   useRefocusOnActivator,
   useWindowDimensions,
@@ -23,6 +22,7 @@ import {
   useDismiss,
   useFloating,
   useInteractions,
+  useListNavigation,
 } from "@floating-ui/react";
 import {
   Header as AriaHeader,
@@ -127,6 +127,8 @@ export function MenuLegacy({
   const [visible, setVisible] = useState(false);
   const [referenceElement, setReferenceElement] =
     useState<HTMLDivElement | null>(null);
+  const [activeIndex, setActiveIndex] = useState<number | null>(null);
+  const listRef = useRef<Array<HTMLElement | null>>([]);
 
   const { width } = useWindowDimensions();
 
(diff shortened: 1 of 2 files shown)


packages/components/src/Menu/__tests__/Menu.test.tsx
```diff
@@ -308,7 +308,7 @@ describe("Menu", () => {
   });
 });
 
-it("should focus first action item from the menu when activated", async () => {
+it("should focus first action item from the menu when activated by keyboard", async () => {
```
ZakaryH, line 311: with focus trap gone, we no longer focus the first item when opening with clicks, only keyboard interactions cause that

packages/components/src/Menu/Menu.tsx
```diff
@@ -232,7 +274,13 @@ export function MenuLegacy({
               <div
                 ref={refs.setFloating}
                 className={styles.floatingContainer}
-                {...getFloatingProps()}
+                {...getFloatingProps({
+                  onKeyDown: event => {
```
ZakaryH, line 278: debatable if this is the right way for tabs to behave when it's open. I've seen a few different ways this works; however RAC does this essentially. when the Menu is open, tabs do precisely nothing, so for parity let's just match until a compelling argument to do something else presents itself.

packages/components/src/Menu/Menu.tsx
```diff
@@ -336,30 +356,45 @@ function SectionHeader({
...
-  label,
-  sectionLabel,
-  icon,
-  iconColor,
-  destructive,
-  UNSAFE_style,
-  UNSAFE_className,
-  onClick,
-}: ActionProps) {
-  const actionButtonRef = useRef() as RefObject<HTMLButtonElement>;
```
ZakaryH, line 349: this ref was a leftover from when we used to use a `useEffect` to focus the first item

we haven't used these in a very long time. I just forgot to remove them when I made the changes back then.

packages/components/src/Menu/Menu.tsx
```diff
@@ -183,7 +184,16 @@ export function MenuLegacy({
   });
 
   const dismiss = useDismiss(context);
-  const { getFloatingProps } = useInteractions([dismiss]);
+  const listNavigation = useListNavigation(context, {
+    listRef,
+    activeIndex,
+    onNavigate: setActiveIndex,
+    loop: true,
+    focusItemOnOpen: true,
```
edison-cy-yang, line 192: This is supposed to focus on the first item on open, right? Just tested this in storybook by clicking to open menu, and the focus styling is not applied, although pressing enter triggers the first item's onClick. New Menu seems to have the equivalent of `focusItemOnOpen: "auto"`, wonder if legacy Menu should just do that.

ZakaryH: ah good catch. yeah `"auto"` sounds like what we want

it should only focus if you use keyboard. mouse doesn't need that.

PR #2766, opened 2025-10-03
feat(components-native): allow nested children in mobile Text/Typography

nad182:
## Motivations

React Native `Text` supports nested children, enabling inline formatting (e.g., mixing bold/italic within a sentence) and inline-like layouts that are difficult with flexbox. We have multiple usages in Jobber Mobile where developers had to reach for `RNText`, which conflicts with our linting rules. If Atlantis mobile `Text` supports nesting, consumers won’t need to import `RNText`, code becomes clearer, and we align our API with the underlying platform behavior.

This PR enables nested children in Atlantis mobile `Typography` and `Text`, and propagates support to other text-based components.

### Added
- Nested text support in mobile `Typography` and `Text` by allowing `ReactNode` children and rendering them directly inside `RN Text`.
- Mobile Storybook stories for `Text` showcasing nested usage:
  - NestedBold
  - NestedMixedStyles
  - MultiLevelNesting

### Changed
- `Typography` transform now applies only to string children of that specific component. Nested elements preserve their own content and may set their own transform independently.
- Widened `children` types for mobile `Heading` and `ActionLabel` to `ReactNode`, enabling nested content inline.
- Updated mobile docs (`TextNotes.mdx`) to document nested text support and transform scoping, and updated `Text.props-mobile.json` to reflect `ReactNode` children.
- Expanded tests to cover nested behavior, transform precedence (parent vs child), multi-level nesting, presence assertions, and key accessibility/ellipsis props.

### Deprecated
- None

### Removed
- None

### Fixed
- None

### Security
- N/A

## Testing

Automated:
- Mobile test suite passes locally with new coverage for nested content and transform behavior.
  - Run only mobile tests:
    - `npm test -- --selectProjects "Atlantis Mobile"`

Manual:
- Storybook mobile for `Text` (Nested stories):
  - Start Storybook: `cd packages/storybook-v7 && npm run storybook`
  - Navigate to Text (Mobile) stories and verify:
    - Mixed inline formatting renders as expected
    - Multi-level nesting renders as expected
- In JM app, confirm:
  - Parent transform (e.g., uppercase) applies to parent strings only
  - Child `Text` with its own transform (e.g., lowercase) renders correctly within a parent transformed `Text`
  - `maxLines` and `selectable` behave as before (no regressions)
  - `hideFromScreenReader` continues to set expected accessibility props

Notes:
- This is a non-breaking update; types are widened. Existing string usages continue to work.
- The transform semantics clarification is now documented: only string children of a component are transformed; nested elements retain control over their own transform.


ZakaryH:
playing around with various combinations, working nicely on both iOS and Android sims

image

not that I think this makes any sense to do, but interestingly if you have your `maxLines` on an inner nested Text-like element, it doesn't work.

I don't know why you'd do that. it should be on the parent, or a separate paragraph altogether if that 's what you wanted so I don't think it's an issue, nor is it likely we can even do anything about it since that's all RN

docs/components/Text/Mobile.stories.tsx
```diff
@@ -42,3 +42,26 @@ ReverseTheme.args = {
...
+export const NestedMixedStyles = () => (
+  <Text>
+    Start, <Text italic>italic segment</Text>, and a
+    <Text emphasis="strong"> bold segment</Text>.
+  </Text>
+);
+
+export const MultiLevelNesting = () => (
+  <Text>
+    {"Level 1 "}
```
ZakaryH, line 61: I did find this a bit awkward, adding `{" "}` for it to work. though we largely leverage translations anyways so I'm not too worried. it's rare to use strings directly as children like this

packages/components-native/src/ActionLabel/ActionLabel.tsx
```diff
@@ -12,9 +12,9 @@ type ActionLabelType = "default" | "cardTitle";
 
 interface ActionLabelProps {
   /**
-   * Text to display
+   * Text to display. Supports strings, numbers, and nested text nodes.
```
ZakaryH, line 15: Im not sure if we need to call out the types when the IDE will already tell us, and we're just extending ReactNode, which includes a bunch of other stuff too like the last values that always make me laugh a bit

```
    type ReactNode =
        | ReactElement
        | string
        | number
        | Iterable<ReactNode>
        | ReactPortal
        | boolean
        | null
        | undefined
        | DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES[
            keyof DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES
        ];
```

might be fine to leave the types to our type system, and here we just say something like "supports nesting text elements"?

packages/components-native/src/Heading/Heading.tsx
```diff
@@ -19,9 +19,9 @@ export type HeadingLevel = "title" | "subtitle" | "heading" | "subHeading";
 interface HeadingProps<T extends HeadingLevel>
   extends Pick<TypographyProps<"base">, "selectable"> {
   /**
-   * Text to display.
+   * Text to display. Supports strings, numbers, and nested text nodes.
```
ZakaryH, line 22: same idea here

packages/components-native/src/Text/Text.test.tsx
```diff
@@ -219,3 +229,14 @@ describe("TypographyGestureDetector", () => {
     expect(textElement.props.collapsable).toBeUndefined();
   });
 });
+
+it("works when nested inside a native RN Text wrapper", () => {
```
ZakaryH, line 233: thinking about this more, let's remove the test. it's not something we want to support or people to do. 


plus if this failed it wouldn't tell us much about us. it would most likely be RNText failing, which isn't our responsibility.

I can see some instances in the app of people doing this and making links using this method, but I probably wouldn't encourage that over some of our more standardized ways of presenting and creating interactive text in the mobile space

packages/components-native/src/Text/Text.tsx
```diff
@@ -53,9 +53,9 @@ export interface TextProps
   readonly align?: TextAlign;
 
   /**
-   * Text to display
+   * Text to display. Supports strings, numbers, and nested text nodes.
```
ZakaryH, line 56: same idea

packages/components-native/src/Typography/Typography.tsx
```diff
@@ -41,9 +41,9 @@ export interface TypographyProps<T extends FontFamily> {
   readonly size?: TextSize;
 
   /**
-   * Text to display
+   * Text to display. Supports strings, numbers, and nested text nodes.
```
ZakaryH, line 44: also here

packages/components-native/src/Typography/Typography.tsx
```diff
@@ -286,6 +286,22 @@ function getTransformedText(text?: string, transform?: TextTransform) {
   }
 }
 
+function transformChildren(
+  children: React.ReactNode,
+  transform?: TextTransform,
+): React.ReactNode {
+  if (children == null) return children;
+
+  return React.Children.map(children, child => {
```
ZakaryH, line 295: not super worried about performance, but we might be iterating over children when we don't need to

`transform` isn't always present, and that seems to be the only thing we're handling with this iteration

what if we do something like

```
function transformChildren(children: React.ReactNode, transform?: TextTransform): React.ReactNode {
  if (children == null || !transform || transform === "none") return children;

  return React.Children.map(children, child =>
    typeof child === "string" ? getTransformedText(child, transform) : child,
  );
}
```

it'll save us a bit of work we know we don't need to do

packages/components-native/src/Typography/Typography.tsx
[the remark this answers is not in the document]
nad182: Good catch! Thanks!

packages/components-native/src/Typography/Typography.tsx
nad182: I like this optimization!

docs/components/Text/Mobile.stories.tsx
ZakaryH: yeah exactly, in theory we could do it, but in practice prettier is always going to break it into multiple lines, and then also remove trailing whitespaces

nad182: I pushed a commit which slightly changed the parenthesis and quote situation: https://github.com/GetJobber/atlantis/pull/2766/commits/79827d11914583ef1cdefb2d903d8cf010b85ec1#diff-b25e629c9e8667d2460bec216423a09eb0eaa9c5e25505849d7e5fdee93cc356R59-R67

Still not ideal, but, hopefully, a bit cleaner.

PR #2917, opened 2026-02-18
fix(components): Fix Autocomplete Single Char Delete Issue

ZakaryH:
## Motivations

while looking into another issue I noticed that Autocomplete was misbehaving when I deleted characters one by one rather than clearing all with a select-all and delete
### Added

- 

### Changed

- 

### Deprecated

- 

### Removed

- 

### Fixed

also cancel the debounced setter so that when we are removing characters, it's both instantaneous and doesn't get stuck in an invalid state where the UI is empty but our debounce sets the filter value to be the last character we had, resulting in a filtered list that doesn't make sense visually

### Security

- 

## Testing

pretty straightforward. with a debounce, the default is fine, type something - then delete the characters with consecutive backspaces. on master you'll see that the list becomes filtered. on this branch, it should correctly return to the default, unfiltered state.

the test I wrote was failing, so you can try that against master - and then see that it now passes!

Changes can be

---

### DIFF
--- a/packages/components/src/Autocomplete/Autocomplete.pom.ts
+++ b/packages/components/src/Autocomplete/Autocomplete.pom.ts
@@ -1,7 +1,16 @@
 import { screen, waitFor } from "@testing-library/react";
 import { userEvent } from "@testing-library/user-event";
 
-const user = userEvent.setup();
+let user = userEvent.setup();
+
+/**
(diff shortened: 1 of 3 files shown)


ZakaryH:
fixed in the multiselect PR, no longer needed

PR #3027, opened 2026-03-31
feat(components): Add sticky variation to Modal Actions and Header

taylorvnoj:
## Motivations

Add sticky header/actions variants to composable Modals so long content scrolls in the middle while key controls stay visible, without changing default behaviour.

### Added

- `variant="sticky" | "inline" (default)` support on `Modal.Header` and `Modal.Actions`
- Visual test cases and snapshots for sticky header/actions/regions

When sticky is enabled:
Header and actions are extracted from the scrollable body
The modal switches to a flex column layout
Only the modal body scrolls, while header and actions remain fixed

This avoids issues with position: sticky (e.g. content appearing underneath the footer) and results in more predictable layout and visual behaviour.

NOTE: The implementation relies on identifying `ModalHeader` / `ModalActions` directly (no wrapper support), which reflects current usage patterns in production. If we decide we want to support that, we can iterate on this foundation.

### Changed

- Composable Modal layout now separates sticky regions from the scrollable body when variants are used
- Modal stories updated to demonstrate sticky header/actions
- Visual test page updated with new sticky region examples

## Testing

- test the variations in the "Sticky Regions" story (header only, actions only, both)
  - when Header is sticky, it stays fixed at the top
  - when Actions are sticky they stay fixed at the bottom
  - only middle content scrolls
- put `variant="sticky"` in the header of "Modal With Provider Example" story to check that tabbing through content is still behaving as normal

https://github.com/user-attachments/assets/12c5a4e3-5492-46ec-8da7-01b76fe8ef20

---


taylorvnoj:
@edison-cy-yang @julia-wjw and I'll write a ticket for spacing followup. ready for review
 
Screenshot 2026-04-10 at 2 42 00 PM

edison-cy-yang (approved):
Thanks for adding the sticky fade! LGTM. Treat this as an approval once @julia-wjw gives a design approval

julia-wjw:
@edison-cy-yang @taylorvnoj looks good to me! TY!

PR #3040, opened 2026-04-06
build: Swapping NPM for PNPM Step 2 of 5

scotttjob:
## Motivations

1. We've long wanted to explore pnpm, and between some of the new features they're rolling out, the fact that Storybook is embracing it fully, and some of our internal packages have as well.
2. This PR Chain is in 5 steps, to make it easier to review.
3. This second PR is making all the local package change updates that we've gotten away with for a long time because of how npm hoists packages. pnpm doesn't do that work magically, so we have to be much more explicit about how our packages get used.
4. This PR needs to be reviewed the closest along with the 4th in the chain (the actual PNPM upgrade).

## Changes

I've attached codex comments to code directly, for why certain things are changing.

## Testing

The application should still work as-is.

Changes can be

---


packages/formatters/package.json
```diff
@@ -12,6 +12,11 @@
       "require": "./dist/index.cjs"
     }
   },
+  "devDependencies": {
```
scotttjob: formatters was building only because Rollup plugins were available via hoisting. pnpm requires them to be declared where they’re used.

packages/hooks/src/useCollectionQuery/test-utilities/mocks.tsx
```diff
@@ -1,5 +1,8 @@
 import { DocumentNode } from "@apollo/client";
-import { MockedProvider, MockedResponse } from "@apollo/react-testing";
+// eslint-disable-next-line import/no-internal-modules
+import { MockedProvider } from "@apollo/client/testing";
```
scotttjob: @apollo/react-testing was the old/indirect path. With pnpm’s stricter graph, we now import Apollo testing helpers from the package that is actually declared and used.

packages/hooks/src/useOnKeyDown/useOnKeyDown.ts
```diff
@@ -12,7 +11,7 @@ interface VerboseKeyComparator {
   readonly [index: string]: boolean | string | undefined;
 }
 
-type KeyComparator = XOR<VerboseKeyComparator, SimpleKeyComparator>;
```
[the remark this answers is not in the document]
MichaelParadis: I really like this especially because I have noticed that XOR is really silly when dealing with array types which  `useOnKeyDown` uses in its parameter `keys: KeyComparator[] | KeyComparator`

PR #3047, opened 2026-04-07
feat(components): Integrate BaseUI and expose unstyled primitives

jdeichert:
## Why Is This Changing?

We're going to start using BaseUI in various parts of our `components` package. This is step 1, integrating the BaseUI dependency into our package structure so consumers can start using it.

## What Is Changing?

* Our `components` package now includes `@base-ui/react` as a dep
    * This is hardcoded to a specific version (no caret) to ensure consumers don't accidentally install a newer version that we haven't tested ourselves
* Our package now exposes 2 new import paths:
    * `/primitives`
        * Not used yet. We will be pre-styling BaseUI's components with our design system and exposing them directly with no logical changes on top. See sugared components below.
    * `/unstyledPrimitives`  
        * Raw, unstyled BaseUI. Only for advanced/experimental usage. We'll expose new components on an as-needed basis.
* Sugared components: when we determine we need to provide a higher-level API on top of BaseUI, we make _sugared_ versions of BaseUI's components with additional props, subcomponents, etc, as necessary. TBD on how/where we expose these (i.e. import paths). Might be case by case.
* A few floating-ui deps were updated to meet BaseUI's requirements. See package-lock.json for changes.

## Consumer Impact

- Impact: None or Low (floating-ui minor updates)
- Action required: None
- Breaking change: No

## Validation

* Visual regression tests pass. Good indicator that the FloatingUI dependency updates are fine.
* Tested the 1 usage of Combobox in our main consumer, imported now from `/unstyledPrimitives`, continues to work as expected.

## Reviewer Notes

This should be a pretty low risk change.

---


packages/components/package.json
```diff
@@ -15,6 +15,11 @@
     },
     "./styles": "./dist/styles.css",
     "./dist/styles.css": "./dist/styles.css",
+    "./unstyledPrimitives": {
+      "types": "./dist/unstyledPrimitives/index.d.ts",
+      "import": "./dist/unstyledPrimitives/index.mjs",
+      "require": "./dist/unstyledPrimitives/index.cjs"
+    },
```
jdeichert: We're not re-exporting from our root barrel file. Only exporting from this path: `@jobber/components/unstyledPrimives`

scotttjob (approved):
Perfect! Everything exactly where it should be. Thanks Jacob. This is a fantastic first step.

PR #3060, opened 2026-04-13
refactor(design): Update color tokens blue--700 and text--secondary

julia-wjw:
## Why Is This Changing?
- `text--secondary` does not pass contrast standards against some backgrounds. 

## What Is Changing?
- Updated `color.base.blue-.700` to `hsl(197, 21%, 36%)` (equivalent to #49646F)
- Point `color.text-.secondary` at `{color.base.blue-.700}` for light theme semantic tokens

## Consumer Impact
- Impact: Low
- Action required: None
- Breaking change: No

## Validation
- [ ] Visual review of secondary text in Storybook/site if applicable

Made with [Cursor](https://cursor.com)


jdeichert (changes requested):
Hey! Just did a quick review and caught some things the LLM probably shouldn't have touched. Unclear why the LLM decided to do that.. I'd have to dig in but don't have time right now.

See comments, reverting those files will unblock this PR 👍

jdeichert (approved):
Looks great, thanks for updating all the screenshots!

PR #3064, opened 2026-04-15
docs: Generate markdown files from docs content

MichaelParadis:
## Why Is This Changing?

UXF is looking into how we can generate & ship our docs in a more LLM friendly way. As part of this we want to be able to generate a single markdown file for each component that provides the code examples inlined. The first step in this is to be able to generate this file. This PR adds that script for generating the single markdown file for each component

This also starts restructuring how we do examples on the docs site so that storybook and the docs site can share code examples instead of having to update 2 different locations with the same code.

## What Is Changing?

Added script to look at the Markdown files in the docs site content and resolves the source code for the examples. It will then inject that content into the LLM markdown file.

I used AST parsing libraries instead of regexes to make it more robust and to minimize the chance a new docs story silently failed because our regex failed.

I also updated the `vite.config` on the docs site to support importing code examples from the `components` and `components-native` package. 

## Consumer Impact

- Impact: None 
- Action required: None
- Breaking change: No 

## Validation

in the sites package run `npm run generate:llm-docs -- --out ./out Autocomplete` and verify two markdown files for Autocomplete v1 and v2 exist. Verify that the code examples are inlined in the markdown

## Reviewer Notes

Changes can be

---


packages/site/vite.config.ts
edison-cy-yang: let's pull this one out as it will come from the code example migration branch, unless it's needed for something here

edison-cy-yang:
If the code example changes for Button and Autocomplete are not needed for this branch, let's pull them out.

packages/site/.gitignore
```diff
@@ -10,3 +10,4 @@ playwright-report/
 blob-report/
 playwright/.cache/
 core
+llm-docs/
```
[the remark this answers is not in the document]
MichaelParadis: That is the default directory if you just run `npm run generate:llm-docs` without a specified output dir

packages/site/scripts/generate-llm-docs.ts
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        const outPath = join(outputDir, `${componentName}.md`);
+        writeFileSync(outPath, markdown.trimEnd() + "\n", {
+          encoding: "utf-8",
+          flag: "w",
+        });
+        converted++;
+
+        if (process.env.VERBOSE) {
```
edison-cy-yang: judging by the commit message, is this env var for Cloudflare? Do we need to add it to CF page?

edison-cy-yang (approved):
lgtm!

edison-cy-yang (approved):

edison-cy-yang (approved):

PR #3068, opened 2026-04-15
build(components): suppress "use client" module directive warning in Rollup

edison-cy-yang:
## Why Is This Changing?

The `@base-ui/react` dependency includes `"use client"` directives in its ESM modules. Rollup treats these as invalid module-level directives and emits noisy warnings during the `packages/components` build, e.g.:

> Module level directives cause errors when bundled, "use client" in "…/ComboboxPortalContext.js" was ignored.

These warnings are harmless — `"use client"` is a React Server Components convention only consumed by RSC-aware bundlers (Next.js). Atlantis is a client-side component library and does not use SSR or RSC, so the directive has no effect.

## What Is Changing?

Added an `onwarn` handler to the Rollup config that silently drops warnings with code `MODULE_LEVEL_DIRECTIVE`. All other warnings continue to surface normally.

## Consumer Impact

- Impact: None
- Action required: None
- Breaking change: No

## Validation

Ran `npm run build` from `packages/components` and confirmed the `"use client"` warnings no longer appear, while other warnings (if any) are still reported.

## Reviewer Notes

The `onwarn` filter is intentionally narrow — it only suppresses `MODULE_LEVEL_DIRECTIVE`, not all warnings. If additional directives from other dependencies appear in the future they'll be covered by the same filter.

Changes can be

---

Made with [Cursor](https://cursor.com)


jdeichert (approved):
I'm fine with this, but not great we're filtering out this directive warning for _all_ cases.

Would prefer if we could limit to base-ui paths.

Up to you if you can get that to work. Unsure if `warning` or `warn` have metadata like file paths, perhaps could check for "base-ui" ?

jdeichert (approved):
Awesome, thank you for catching and fixing this!

PR #3093, opened 2026-04-20
feat(components): add open-in-new-tab action to LightBox toolbar

nad182:
## Why Is This Changing?

Voice of Customer feedback: `Gallery` and `Lightbox` don't let users read fine details on attached images (e.g. equipment serial numbers on technician photos). The image fits the viewport with no way to zoom in, and the previous workflow - clicking an attachment inside a Note's edit view to open the raw image in a new tab via the `InputFile` alternative in JO - is no longer available.

This PR is an interim fix that puts the "view at full size" workflow back without reintroducing it in `InputFile` (the permission concerns there are one of the reasons we're not putting it back in; e.g. in the Note edit state). A proper in-Lightbox zoom/pan solution is tracked separately in [JOB-160833].

## What Is Changing?

- Every Lightbox now renders an open-in-new-tab action on the left side of the toolbar. Clicking it opens the current image URL in a new browser tab, where the browser's native image viewer handles zoom/pan.
- The action tracks the currently-visible image: navigating with thumbnails, arrow keys, swipe, or prev/next updates the link target.
- A tooltip on hover/focus reads "Open image in new tab" (also the button's accessible name).
- The action is also exported as a composable leaf, `LightBox.OpenInNewTabAction`, for consumers who assemble a custom Lightbox tree via `LightBox.Provider` and want to keep the action.

Before | After
:---:|:---:
image | image

## Consumer Impact

- Impact: Low
- Action required: None
- Breaking change: No

Visually, every Lightbox in Atlantis gains a single small icon button on the left side of the toolbar. The counter and close button are unchanged. All existing `<LightBox>`, `<LightBox.Provider>`, and sub-component props keep their current contracts. No prop additions on existing entry points.

## Validation

- Unit tests: `cd packages/components && npx jest LightBox` - 28/28 pass (25 existing, 3 new for the action: renders with correct target, tracks the current image during navigation, returns null when the current image has no URL).
- Gallery tests still pass (Gallery is not touched): `cd packages/components && npx jest Gallery` - 20/20 pass.
- Storybook: the existing `LightBox > Basic` story now shows the new button in the toolbar. No story changes needed since the button is included by default.
- Manual: hover the button to confirm the tooltip, click it to confirm it opens in a new tab, navigate between images to confirm the link target follows the current image.
- Regenerated 9 visual snapshots under packages/site/tests/visual/ to include the new toolbar button; ran via `test:visual:update`.

## Reviewer Notes

- Scope of this PR is deliberately minimal: `Gallery` is untouched, no new public API on existing components. The only new surface is the `LightBox.OpenInNewTabAction` composable leaf.
- Design agreed with this approach with the tooltip addition; a11y is covered by the button's accessible name and the tooltip's `aria-description` injection via shadow activator.
- In-Lightbox zoom/pan (the richer fix) is intentionally out of scope here and has its own ticket and [POC branch].

[JOB-160712]: https://jobber.atlassian.net/browse/JOB-160712
[JOB-160833]: https://jobber.atlassian.net/browse/JOB-160833
[POC branch]: https://github.com/GetJobber/atlantis/tree/JOB-voc/lightbox-zoom-pan

Changes can be


edison-cy-yang (commented):
One comment on translation, otherwise the code looks good to me

julia-wjw (approved):

edison-cy-yang (approved):
lgtm from the dev side

PR #3095, opened 2026-04-22
feat(components): export isPreviewableImage util [JOB-161235]

nad182:
## Summary

Adds `isPreviewableImage` to the public `@jobber/components` barrel so it can be imported by consumers:

```ts
import { isPreviewableImage } from "@jobber/components";
```

No behavior change - the util itself is unchanged; this PR only adds a re-export in `packages/components/src/index.ts`.

## Why

[#3077](https://github.com/GetJobber/atlantis/pull/3077) fixed DWG files opening in preview by centralizing the preview-type check into `isPreviewableImage`. While validating that fix in the consumer app, we found that several surfaces bypass `<Gallery>` / `<Thumbnail>` and wire up `<LightBox>` manually, using a local duplicate of `isPreviewableImage` that had already drifted out of sync (it still returns `true` for `image/vnd.dwg`):

- Quotes attachments
- Media gallery
- Client communications modal
- Line items (read-only and input)
- Message center

Making the Atlantis util importable lets those teams dedupe to a single source of truth and prevents the same class of bug from being re-introduced.

## Scope

- Adds one line: `export * from "./utils/isPreviewableImage";` in the main barrel.
- Does not change `isPreviewableImage` or add new utils.
- No new tests - the util already has coverage in `isPreviewableImage.test.ts` and via `Gallery` / `Thumbnail` behavior tests.

## Follow-ups (not in this PR)

- Migrating the consumers listed above to the exported util.

## Ticket

[JOB-161235](https://jobber.atlassian.net/browse/JOB-161235)

[JOB-161235]: https://jobber.atlassian.net/browse/JOB-161235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scotttjob (approved):
Looks okay to me.

Looking at the underlying isPreviewableImage though, I most likely would have had the feedback on the original PR to not require `file` as the wrapping object if we're only concerned about getting the type from the object to compare on strings (the file object is irrelevant, and more of a detail about the caller than the function itself)

But, not sure if that's worth the refactor at this point! :)

nad182:
> Looks okay to me.
> 
> Looking at the underlying isPreviewableImage though, I most likely would have had the feedback on the original PR to not require `file` as the wrapping object if we're only concerned about getting the type from the object to compare on strings (the file object is irrelevant, and more of a detail about the caller than the function itself)
> 
> But, not sure if that's worth the refactor at this point! :)

@scotttjob, this is a good point. I didn't think there was a need to narrow down the prop originally, since this was an internal util, which was already defined like that - it was just moved to a shared file. Refactoring wasn't really a goal in the original ticket/PR. 

But I am more than happy to refactor this function for the following reasons: better ergonomics, safety and simplicity/readability. Done in https://github.com/GetJobber/atlantis/pull/3095/commits/2abaff13bc31c52655df04b3c6c45c2282f380db.

scotttjob (approved):
Oh I'm sorry! I approved the original, I definitely don't think we need to layer in this refactor just to get this out the door (was more commenting for the future, talking out loud).

Approved this one as well.

PR #3074, opened 2026-04-16
chore: Component Upgrade Menu POC

taylorvnoj:
## Consumer Impact

- Impact: 
- Action required: 
- Breaking change: 

## Reviewer Notes

Changes can be
[tested via Pre-release](https://github.com/GetJobber/atlantis/actions/workflows/trigger-qa-build.yml)

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

PR #3096, opened 2026-04-22
feat(components): Updating Chip for Upcoming Menu Updates

Beccaneer:
## Why Is This Changing?

The upcoming `ComposableMenu` work ([#3074](https://github.com/GetJobber/atlantis/pull/3074)) uses Base UI's render-prop composition pattern, where the menu injects handlers, refs, and aria attributes onto whatever trigger element the consumer provides. For `Chip` to act as a menu trigger, it needs to forward those injected props to the rendered element and accept a standard event-first `onClick` signature.

This PR splits the Chip-specific groundwork out of #3074 so it can land and be reviewed independently of the larger menu proposal.

## What Is Changing?

- **`Chip.onClick` is now event-first**: `(ev) => void`, matching standard DOM semantics and Base UI's prop-injection contract.
- **New `onClickValue` prop**: retains the previous `(value, ev) => void` behavior for consumers that need it (e.g. a shared handler reacting to any chip in a list). Marked `@deprecated` as a migration path — prefer `onClick` + closure over the value. When both are provided, `onClickValue` wins.
- **`Chip` forwards arbitrary DOM attributes** (event handlers, `aria-*`, `data-*`, `style`, `title`, etc.) to the rendered element via a single `...rest` spread, so it composes with render-prop APIs.
- **`className` is merged** with Chip's internal class list instead of being ignored.
- `InternalChip`'s `(_, ev) => onClick(ev)` adapter is gone — it just passes `onClick` through now.

## Consumer Impact

- **Impact**: Medium
- **Action required**: Migration needed for direct `Chip` consumers that rely on `onClick(value, ev)`
- **Breaking change**: Yes — `Chip.onClick` signature

### Migration

```
// Before
<Chip value="a" onClick={(value, ev) => handleClick(value)} />

// After — option 1 (preferred): closure
<Chip value="a" onClick={ev => handleClick("a")} />

// After — option 2: drop-in replacement via deprecated prop
<Chip value="a" onClickValue={(value, ev) => handleClick(value)} />
```

Internal consumers (`InternalChip`, `ChipSelectable`, `ChipDismissible`) are updated in this PR. No public Atlantis component changes its own API.

## Validation

- [ ] `jest packages/components/src/Chip packages/components/src/Chips packages/components/src/Combobox` passes
- [ ] `tsc` passes for `@jobber/components`
- [ ] Storybook renders Chip stories unchanged
- [ ] Spot-checked callers: `InternalChip`, `ChipSelectable`, `ChipDismissible`, `ComboboxActivator` (`Chip` as combobox activator)
- [ ] Changeset added

## Reviewer Notes

- `testID` remains the canonical test-id prop (not `data-testid`).
- This is intentionally a smaller diff than the original in #3074: no hand-enumeration of event handlers, no dual `aria-label`/`ariaLabel` or `testID`/`data-testid` accept-both logic, and no "both click handlers fire" semantic.
- Looking ahead: if `Button` ends up getting the same widening in #3074, a shared `<Slot>` / `asChild` helper would centralize this rather than every triggerable component re-implementing DOM passthrough. Not in scope here.

---


taylorvnoj (changes requested):
Some comments for this PR, but also saw failures putting the pre-release through prod. Will DM ya!

taylorvnoj (dismissed):
I think when we removed the updates from AI we also lost some code from the original commit (which matched Scott's PR). Let's go back to what's in Scott's branch/the first commit I think they are identical...

taylorvnoj (approved):
Thanks @Beccaneer - approving but let's wait to hear from @MichaelParadis before we hit merge. He's just dealing with the current version bump

PR #3097, opened 2026-04-22
refactor(components): widen Button DOM props and forward ref for composition

Beccaneer:
## Why Is This Changing?

The upcoming `ComposableMenu` work ([#3074](https://github.com/GetJobber/atlantis/pull/3074)) uses Base UI's render-prop composition pattern, where the menu injects handlers, refs, and aria attributes onto whatever trigger element the consumer provides. For `Button` to act as a menu trigger, it needs to forward those injected props to the rendered element and expose a ref so Base UI can manage focus and positioning.

This PR splits the Button-specific groundwork out of #3074 so it can land and be reviewed independently of the larger menu proposal.

## What Is Changing?

- **`Button` now uses `React.forwardRef`.** The rendered `<Link>` / `<a>` / `<button>` element receives the ref, typed as `HTMLAnchorElement | HTMLButtonElement`. Required for Base UI composition; also useful for any consumer that needs direct DOM access.
- **`Button` accepts arbitrary DOM attributes** (`onKeyDown`, `onFocus`, `aria-expanded`, `aria-haspopup`, `aria-controls`, `data-*`, `tabIndex`, etc.) via a single `...rest` spread on the rendered element. Atlantis-specific props are explicitly destructured so they don't leak onto the DOM.
- **`className` is now merged** with `useButtonStyles` output and `UNSAFE_className.container`. **`style` is now merged** with `UNSAFE_style.container` (shallow, UNSAFE wins on conflict, preserves `undefined` when neither is set).
- **Atlantis camelCase aria props take precedence** over kebab-case equivalents injected from a parent. `ariaControls` / `ariaHaspopup` / `ariaExpanded` / `ariaLabel` still drive the DOM `aria-*` attributes, but Base UI can now inject `aria-expanded` via `...rest` when the Atlantis prop is unset.
- The dynamic `Tag = url ? \"a\" : \"button\"` render is split into three explicit branches so ref typing is correctly narrowed per element.

### Non-DOM props stripped from `...rest`

The following Atlantis-specific props are destructured out and intentionally not forwarded to the DOM element: `fullWidth`, `icon`, `iconOnRight`, `label`, `size`, `type` (\"primary\" / \"secondary\" / \"tertiary\" — not the HTML `type`), `variation`. Leaving them in `...rest` would trigger React unknown-prop warnings.

## Consumer Impact

- **Impact**: Low
- **Action required**: None
- **Breaking change**: No

`Button.types.ts` is unchanged. Existing consumers continue to work with no API changes. The new capabilities (ref forwarding, DOM prop passthrough, `className` / `style` merging) are purely additive.

## Validation

- [ ] `jest packages/components/src/Button packages/components/src/Combobox` passes (Combobox tests use `<Button />` as an activator)
- [ ] Button snapshot tests match — DOM output was preserved for all existing cases (`disabled`, `url` + `external`, `submit`, `to`, `loading`, each `variation` / `type`, etc.)
- [ ] `tsc` passes for `@jobber/components`
- [ ] Storybook renders Button stories unchanged
- [ ] Changeset added

## Reviewer Notes

- **Deliberately narrower than the same work in #3074.** That version hand-enumerates ~15 event handlers (`onKeyDown`, `onFocus`, `onBlur`, `onPointerDown`, …) alongside a `...rest` spread, which is redundant — `...rest` already carries them. It also accepts both `ariaLabel` and `aria-label` with a `??` fallback. This PR keeps only the Atlantis camelCase form (`ariaLabel`) since Base UI doesn't inject `aria-label`; for `aria-controls` / `aria-haspopup` / `aria-expanded` the dual behaviour is preserved via conditional overrides because Base UI **does** inject those.
- **Snapshot fidelity** was the main preservation target. `mergedStyle` returns `undefined` (not `{}`) when neither `style` prop nor `UNSAFE_style.container` is set, matching the original `style: UNSAFE_style.container` shape. `disabled`, `role`, `id`, and the conditional `onClick` / `onMouseDown` guard behaviour are unchanged.
- **Forward-looking:** now that `Button` and `Chip` ([previously extracted PR](https://github.com/GetJobber/atlantis/compare/master...claude/inspiring-hofstadter)) have nearly identical DOM passthrough logic, a shared `<Slot>` / `asChild` helper would consolidate this rather than every trigger-capable component reimplementing it. Out of scope here.


Beccaneer:
moving PR out of worktree structure to proper branch: https://github.com/GetJobber/atlantis/pull/3103

PR #3099, opened 2026-04-23
fix(components-native): Tie Toast a11y announcement to render lifecycle

edison-cy-yang:
<!--
## Important notes

* Atlantis is a PUBLIC repo. Please avoid sharing internal details/context.
* If you haven't yet, please create a slack thread to ensure UXF is aware of this PR.

## PR title formatting

Atlantis uses Conventional Commits to track versions.
Pull request titles should follow the following format.

For help creating your pull request, you can [use this tool](https://atlantis.getjobber.com/guides/pull-request-title-generator)

<TYPE>(<optional SCOPE>): <conditionally BREAKING CHANGE:> <description>

eg.
fix(SCOPE): stop graphite breaking when too much pressure applied — Patch Release
feat(SCOPE): add 'graphiteWidth' option — (Minor) Feature Release
feat(SCOPE): BREAKING CHANGE: remove graphiteWidth option — (Major) Breaking Release

TYPE should consist of:
- fix: a commit of the type fix patches a bug in your codebase
- feat: a commit of the type feat introduces a new feature to the codebase
- docs: documentation only changes
- build: improvements to the build system
- refactor: a change that neither fixes a bug nor introduces a feature
- chore: other changes that don't modify src or test files

SCOPE should be one of:
- components
- components-native
- deps
- deps-dev
- design
- docx
- eslint
- formatters
- generators
- hooks
- stylelint


If your pull request introduces a breaking change please append `BREAKING CHANGE:` following type / scope.

Further Reading:
- https://www.conventionalcommits.org
- https://github.com/commitizen/conventional-commit-types/blob/master/index.json
-->

## Why Is This Changing?
                                                                                                                                                                                                                                                                                                                                      
  `AccessibilityInfo.announceForAccessibility` was being called inside the `onShow`                                                                                                                                                                                                                                                   
  animation callback of `react-native-toast-message`, behind a 100ms `setTimeout`.                                                                                                                                                                                                                                                    
  This created a two-step async chain (animation callback → timer) that was                                                                                                                                                                                                                                                           
  unreliable under CI load, causing intermittent failures:                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                      
  Error: expect(jest.fn()).toHaveBeenCalledWith(...expected)                                                                                                                                                                                                                                                                          
  Expected: "Notify message"
  Number of calls: 0                                                                                                                                                                                                                                                                                                                  
                  
  The root cause: `onShow` is called synchronously before React flushes the
  `setIsVisible(true)` state update, so the timer is scheduled at an earlier,                                                                                                                                                                                                                                                         
  less predictable point relative to the test's `waitFor` polling window. On                                                                                                                                                                                                                                                          
  slower CI machines, the timer races against `waitFor`'s timeout.                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                      
  ## What Is Changing?                                                                                                                                                                                                                                                                                                                
                  
  - **`Toast.tsx`**: Moves `announceForAccessibility` from `showToast`'s `onShow`                                                                                                                                                                                                                                                     
    callback into a `useEffect` inside `DefaultToast`. The component mounting is
    the correct signal that the toast is visible — no animation callback required.                                                                                                                                                                                                                                                    
    The 100ms delay is preserved to avoid conflicting with the entrance animation                                                                                                                                                                                                                                                     
    on device.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
                  
  ## Consumer Impact                                                                                                                                                                                                                                                                                                                  
                  
  - Impact: None — internal implementation change only, no public API affected.                                                                                                                                                                                                                                                       
  - Action required: None.
  - Breaking change: No.                                                                                                                                                                                                                                                                                                              
                  
  ## Validation                                                                                                                                                                                                                                                                                                                       
                  
  - I tried adding a regression test that strips `onShow` from `Toast.show` to simulate the race
    condition. It would fail on the original code (`Number of calls: 0`) and passes
    with the fix. It is not included in the diff since it was only used to confirm the root cause of the flake.
  - Existing tests continue to pass.

  ## Reviewer Notes

  The key insight is that tying the announcement to `useEffect` (React's render
  lifecycle) rather than an animation callback makes RNTL's `waitFor`/`findByText`
  the right tool — both respond to the same render flush, so there's no longer a
  hidden async dependency that the test framework can't see.
  
Changes can be

---


edison-cy-yang:
I tested this change on device and somehow the voiceover isn't working

PR #3111, opened 2026-04-29
chore: Fix e2e docker setup with pnpm

jdeichert:
## Why Is This Changing?

Our e2e docker test runner is running into issues where pnpm's installation strategy differs from npm's. This PR improves the docker setup to ensure pnpm's deps are installed and cached _within_ the container's volumes and avoids conflicts/overwrites with our local node_modules directories.

## What Is Changing?

Adjusted the e2e script to store each node_modules directory in a named volume. Also configured the pnpm global store directory to be cached as well.
Also simplified the inline bash script into a separate file to make modifications easier + documentation.

## Consumer Impact

- Impact: None
- Action required: None
- Breaking change: No

## Validation

* Re-run e2e tests and they should work
* Re-run again and you'll notice it skips the install process, due to the cached modules being available


scotttjob (approved):
I haven't run it directly, but reading the code, it all makes sense.

Appreciate you Jacob!

MichaelParadis (approved):
LGTM!

PR #3115, opened 2026-04-29
feat(components): Add step prop to InputTime

edison-cy-yang:
## Why Is This Changing?

`InputTime` only displays hours and minutes. This causes UX issue when there are two InputTime instances with time-elapsed computation. Their underlying Date values had non-zero seconds (e.g. 12:29:45 → 14:08:00), but the input only displayed HH:MM, so the visible duration didn't match the math (1h 38m vs the expected 1h 39m). We need a way to show seconds so the displayed times align with the computed duration.

Today, V2 InputTime exposes no way to control sub-minute precision — you get HH:MM and that's it. This PR adds an opt-in step prop that mirrors the [native HTML <input type="time"> step attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/time#step), defaulted to 60 so existing usages are unchanged.

## What Is Changing?

InputTime (V2 / rebuilt) accepts a new optional `step?: number `prop:

`step` defaults to 60 → input displays HH:MM (current behavior).
`step < 60` (e.g. 1, 15) → input displays HH:MM:SS and seconds round-trip cleanly through onChange.
`step > 60 `→ input still displays HH:MM; the browser uses step as the granularity for arrow-key increment and HTML5 stepMismatch validation.
The value is forwarded directly to the native `<input type="time">`, so the browser owns display granularity, arrow-key stepping, and validity. Our serialization layer (`dateToTimeString` / `timeStringToDate`) is updated to emit and parse `HH:MM:SS` when `step < 60`, and to default seconds to 0 when parsing `HH:MM`.

## Consumer Impact

- Impact: Low — purely additive.
- Action required: None — default of 60 preserves existing display and behavior. Consumers wanting seconds opt in via step={1} (or any value < 60).
- Breaking change: No.

## Validation

- New unit tests in InputTime.rebuilt.test.tsx:
- New dedicated unit tests for `utils/input-time-utils.ts` covering `dateToTimeString` (with/without seconds, padding, undefined input) and `timeStringToDate` (HH:MM, HH:MM:SS, base-date preservation, invalid/out-of-range guards).
- New Storybook story `WithSeconds (step: 1)` in `InputTimeV2.stories.tsx`.
- New visual test sections in `VisualTestInputTimePage.tsx`

## Reviewer Notes

Changes can be

---


MichaelParadis (approved):
LGTM! I am so sorry for the super long response on the review

PR #3117, opened 2026-04-30
feat(components): Adding HelperText pritimitve

Beccaneer:
## Why Is This Changing?

The upgraded `Select` (and other field-like components down the line) needs a small block of helper or error text below the trigger, matching the [Figma HelperText design](https://www.figma.com/design/ZSjVtQjTlF2iDudlt6AnBp/Select-%F0%9F%AB%B5?node-id=673-26429). Today the equivalent visuals live inside `FormField` as two separate sub-components (`FormFieldDescription` for neutral, `InputValidation` for critical), tightly coupled to the legacy form-field stack. They cannot be reused outside `FormField` without dragging that wrapper along.

This PR introduces `HelperText` as a standalone primitive so the new `Select` and any future field-like component can render the same neutral / critical text treatment without re-implementing it or pulling in `FormField`.

## What Is Changing?

- New `HelperText` primitive at `packages/components/src/primitives/HelperText/`, exported from `@jobber/components/primitives`.
- Props:
  - `message: string` — required, the helper text. Named `message` to match `InputValidation`'s API and to signal that this is a leaf component, not a composition surface.
  - `status: "neutral" | "critical"` — defaults to `"neutral"`. `critical` renders an `alert` icon followed by error-coloured text; `neutral` renders subdued text.
  - `id` — auto-generated via `useId` when omitted, so parent form fields can wire `aria-describedby` on their control.
- Reuses the existing `Text` and `Icon` atoms — no BaseUI dependency. HelperText is a visual leaf; live-region semantics, animation, and `aria-describedby` wiring stay with the parent component (see proposal).
- Proposal: [docs/component_upgrades/HelperText.proposal.md](docs/component_upgrades/HelperText.proposal.md).

Screenshot 2026-04-30 at 1 21 59 PM

Screenshot 2026-04-30 at 1 22 06 PM

Out of scope:

- Lint / Danger carve-out for `@jobber/components/primitives` so `HelperText` is importable by product teams without firing the rule that warns on consumer imports of the styled BaseUI Compound layer — flagged as an open question in the proposal.
- A native (mobile) equivalent — handled separately per the methodology.

## Consumer Impact

- Impact: Low
- Action required: None
- Breaking change: No

Net-new export. No existing components change behaviour. The component is currently consumed only by the in-progress `Select` upgrade; product teams can adopt it directly once it lands but are not required to.

## Validation

- HelperText unit tests pass: `pnpm test -- --testPathPattern="primitives/HelperText"` (5 tests covering both statuses, auto-id, custom id, ref forwarding).
- Lint clean: `pnpm exec eslint packages/components/src/primitives/HelperText`.
- Storybook stories (`Components/Primitives/HelperText`) render `Neutral` and `Critical` with a `status` control.

## Reviewer Notes

- The component lives under `primitives/` matching the existing `BottomSheet` and `OverlaySeparator` layout, but it's an Atom rather than a styled BaseUI mirror. The proposal calls out that the existing lint/Danger guard on this barrel will need to be updated before product teams import it directly.
- The CSS keeps both `status` variants on the same flex row (`align-items: center`); only `gap` differs between `neutral` (0) and `critical` (`--space-smaller`).
- `id` is auto-generated even in standalone use. That's intentional — cheap, and avoids a footgun for parent components that forget to pass one through. Consumers who need a specific id can override.

Changes can be


taylorvnoj:
👀

taylorvnoj (commented):
Thanks Becca! Some questions around naming and UI questions, lmk if you want to chat!

taylorvnoj (approved):
Thanks for ironing out all the fine details!

PR #3135, opened 2026-05-05
perf: generate docs faster

MichaelParadis:
## Why Is This Changing?

Our docs gen script was super slow. It would take ~50 seconds to run which added friction when working locally and it would also take a long time during our CI run. The main change was updating the docs gen to parse a list of file paths instead of doing `parse(singleFile)` which would cause a typescript compiler to be instantiated for every file. The batching means the typescript compiler is only instantiated once per batch

I also updated the script to leverage typescript to help with future maintainance

## What Is Changing?

Changed `generateDocs` script to batch all component paths based on the platform. 

The paths in the props files to be absolute paths instead of relative paths for the files. 
Before: `../components-native/src/BottomSheet/BottomSheet.tsx`
After: "packages/components-native/src/BottomSheet/BottomSheet.tsx"
This is a result of batching the parsing with `react-docgen-typescript`

### Known output differences

Batch compilation produces cosmetically different output in a small number
of cases:

- **Union type member ordering** may differ (e.g. `"base" | "large"` vs
  `"large" | "base"`). TypeScript resolves symbols in a different order when
  processing many files together. This is semantically equivalent.
- **`parent.fileName` format** may change from `../components/src/…` to
  `packages/components/src/…` for props defined in the same file as the
  component. The site's `normalizeGithubFilePath` handles both formats and
  produces the same GitHub URL.

## Consumer Impact

- Impact: None 
- Action required: None 
- Breaking change: No 

## Validation

Run `pnpm generate` in the sites package and observe how much faster the docs gen is. Your git status should also be empty

This was the differences I observed on my machine:

```
Benchmarking generateDocs across 139 component files
  web: 86  mobile: 45  v2: 8

Per-file parsing (OLD)
  runs : 1
  avg  : 48.98s
  min  : 48.98s
  max  : 48.98s

Batch parsing    (NEW)
  runs : 1
  avg  : 2.08s
  min  : 2.08s
  max  : 2.08s

Speedup: 23.55x  (96% faster)

```
## Reviewer Notes

Changes can be

---


MichaelParadis:
Current CI failure on check_for_manual_release seems like a network blip. I will rerun after the e2e finishes so the whole build doesn't need to run
image

scotttjob (approved):
Tests are passing, I reviewed by code. I didn't run it myself manually.

It's a low-risk change, that we can roll back if there's weird issues.

Approving.

PR #3137, opened 2026-05-06
feat(components): add InputNumberExperimental primitive on Base UI

nad182:
## Why Is This Changing?

We are deprecating the v1 `<InputNumber> and are actively replacing its instances in the consumer apps. We need a numeric input build on top of **Base UI** as a path forward, but the v2 `<InputNumber>` rebuild will undergo its own improvement/migration. 
This PR adds a styled, "experimental" primitive on top of `@base-ui/react/number-field` so those consumers can stop relying on v1 ahead of the full v2 migration/upgrade. Some of my decisions were cognizant of existing usages and patterns of the current V2 component - so that (hopefully) less migration work would be required.

## What Is Changing?

1. Fix: a "hidden" build-pipeline bug where styled primitives' CSS module class hashes ended up in the JS bundle but not in the extracted `styles.css`. Affects every styled primitive, including the existing `OverlaySeparator`, `BottomSheet`, and `HelperText`. I noticed it when the stylesheet wasn't getting applied to the `InputNumber` instances in the consumer apps. We didn't catch the bug earlier because no consumers have been using any of the other "styled" primitives yet. 
2. Add: the new component (source, types, CSS module, tests, barrel) and export it from `@jobber/components/primitives`.
3. Add: the Storybook stories for  easier visual QA and comparisons between the V1 and the replacement "Experimental" component.

### `<InputNumberExperimental>` API surface: supports composition with "sensible defaults"

- Wraps `@base-ui/react/number-field` with Atlantis design tokens
- Default render covers an animated mini-label, hover/focus-revealed stepper, optional prefix/suffix (label and/or icon), and `Field.Root`-composed description and error rendering so labels, descriptions, and errors wire up through Base UI's accessibility primitives
- Compound API (`<.Input>`, `<.Affix>`, `<.Stepper>`) lets consumers compose the parts manually with sensible context-driven defaults
- Controlled API: `value: number | null` paired with `onChange: (value: number | undefined) => void`. `onChange` is **commit-only** (fires on blur, Enter, stepper press, or arrow step) so it matches V1 / V2 timing — consumers wiring `onChange` to side effects like network requests fire on completed input rather than on every keystroke
- `allowOutOfRange` is on by default so typed values outside `min`/`max` stay visible and the consumer's validation can show a message; stepper interactions still clamp
- Composes inside `InputGroup`: the wrapper reads `--public-field--*-radius` via `var(name, fallback)` so adjacent siblings in a horizontal `InputGroup` continue to flatten their inner corners

## Consumer Impact

- Impact: Low to Medium: there no existing API changes, so this component should serve as a "direct substitute" for the V1
- Action required: ready for replacing consumers migrating off v1
- Breaking change (compared to V1): No

The build-pipeline fix retroactively makes the existing primitives' styles flow into `dist/styles.css` correctly. Nothing else in the public API changes.

## Validation

- 50 interaction tests covering rendering, affixes, controlled value, `onChange` commit-timing contract, stepper bounds, keyboard (Enter / modifier handling), focus / blur via ref, accessibility (label wiring, stepper aria-labels), and the compound pattern
- Storybook stories: `Basic`, `WithSuffixLabel`, `WithPrefixLabelAndSuffixLabel`, `InvoiceTotal_V1Parity`, `PrefixIconOnly`, `PrefixAndSuffixIcons`, `ClickableSuffixIcon`, `WithDescriptionAndError`, `AllSizes`, `Compound`, `V1ComparisonGrid`
- The `V1ComparisonGrid` story renders the v1 `<InputNumber>` and the new primitive side-by-side across the four affix variants (label+icon prefix, icon-only prefix, prefix+suffix icons, clickable suffix icon)
- Smoke-tested in a downstream consumer across 10 real call sites covering all sizes, controlled patterns, and suffix-label usage

## Reviewer Notes

I left targeted review comments on a few specific lines that capture intentional decisions worth flagging (mini-label position vs v1 with prefix labels, controlled-only API, `allowOutOfRange` default, the compound-pattern detection logic).


MichaelParadis (approved):
LGTM!

PR #3131, opened 2026-05-05
feat(components): Menu component upgrade

taylorvnoj:
## Why Is This Changing?

The composable `Menu` is being moved onto Base UI primitives so we can standardize menu behaviour on a more robust accessibility and interaction foundation. At the same time, we still need Atlantis-specific behaviour for cases where a Base UI primitive does not directly map to the UX we want, especially on small screens where the composable menu is presented as a bottom sheet rather than a desktop-style popup menu.

## What Is Changing?

The composable `Menu` implementation now routes through Base UI-backed desktop and mobile shells, while the legacy items API remains supported. On desktop, this improves the composable menu structure and submenu behaviour. On mobile, the bottom-sheet presentation keeps using drawer-style primitives, with Atlantis layering in the menu-specific focus and keyboard behaviour we want.

This change also tightens several visual and interaction details discovered during the migration, including trigger alignment, focus return on close, menu item text sizing, and mobile bottom-sheet navigation/highlight behaviour.

## Consumer Impact

- Impact: Low (but I will install a pre-release into product)
- Action required: Backwards compatible but I will investigate any test regressions in product
- Breaking change: No

## Validation

- Added/updated unit coverage for Atlantis-owned mobile bottom-sheet behaviour
- Added visual coverage for new composable menu states
- Updated page/menu visual snapshots where the accepted composable behavior changed

## Reviewer Notes

This is the original POC https://github.com/GetJobber/atlantis/pull/3074

Changes can be
[tested via Pre-release](https://github.com/GetJobber/atlantis/actions/workflows/trigger-qa-build.yml)

---

[In Atlantis we use Github's built in pull request reviews](https://help.github.com/en/articles/about-pull-request-reviews).

PR #3138, opened 2026-05-07
docs: Fix menu component prop types are exported and fixed union order when generating docs

MichaelParadis:
## Why Is This Changing?
I noticed after the merge of the Menu improvements https://github.com/GetJobber/atlantis/pull/3131 the subcomponent props were no longer present. This PR fixes that by leveraging our existing pattern. 
The source of this problem is the fact that the `Object.assign` is a runtime assignment making it impossible for `react-docgen-typescript` to infer the actual types. 

Another issue was that the performance improvements to doc gen (https://github.com/GetJobber/atlantis/pull/3135) there were the following issues that caused the generation to sometimes be inconsistent across runs meaning git changes would be observed across runs.
 1. The resolution of union types to be inconsistent sometimes causing git changes when there shouldn't be. This PR makes it so the resolution of these unions is stable by sorting them to be in alphabetical order.
 1. When several files use `React.forwardRef<…>`, the type checker reuses the
first file's resolution of the ref generic for every later file in the same
program. The visible symptom: V2 inputs all reported the *same* ref type, e.g.
`InputNumberRebuilt`, `InputPhoneNumberRebuilt`, and `InputTextSPAR` all
claimed `Ref<HTMLInputElement | HTMLTextAreaElement>` after the InputText
file was parsed first.
1. Subcomponent ordering wouldn't be consistent across runs

## What Is Changing?
 - Update our Menu component to use our existing patterns for exporting subcomponents
 - Update the doc gen so that the resolution of unions is always in alphabetical order
 - Update the doc gen so that the resolution of subcomponents is always in alphabetical order
 - Update the doc gen so that the resolution of props so they are always in alphabetical order

- Prevent forward refs from being incorrect. This is done by:
  - Creating a shared tsscript compiler instance for each group of components (mobile, web and v2)
  - Switching from `parse([files])` to `parseWithProgramProvider` and running each file against a *shared* TypeScript Program. Per-file parsing isolates the checker's symbol resolution; the shared program keeps program creation a one-time cost rather than per-file.

Benchmark result (script [here](https://github.com/GetJobber/atlantis/pull/3138/changes/01b2cd46b0d66305798f1f74bc09a5a219776e00)):
```
Running each strategy 3x per section, best-of-N

=== web (86 files) ===
 round 1:
  OLD parse([files])           : 1437ms
  NEW shared program + per-file: 1015ms
 round 2:
  OLD parse([files])           : 911ms
  NEW shared program + per-file: 1038ms
 round 3:
  OLD parse([files])           : 997ms
  NEW shared program + per-file: 998ms
  best of 3: OLD 911ms, NEW 998ms (+9.5%)

=== mobile (45 files) ===
 round 1:
  OLD parse([files])           : 749ms
  NEW shared program + per-file: 685ms
 round 2:
  OLD parse([files])           : 658ms
  NEW shared program + per-file: 648ms
 round 3:
  OLD parse([files])           : 650ms
  NEW shared program + per-file: 648ms
  best of 3: OLD 650ms, NEW 648ms (-0.3%)

=== v2 (7 files) ===
 round 1:
  OLD parse([files])           : 533ms
  NEW shared program + per-file: 510ms
 round 2:
  OLD parse([files])           : 531ms
  NEW shared program + per-file: 520ms
 round 3:
  OLD parse([files])           : 524ms
  NEW shared program + per-file: 521ms
  best of 3: OLD 524ms, NEW 510ms (-2.7%)
```

## Consumer Impact

- Impact: None 
- Action required: None 
- Breaking change: No 

## Validation

Verify that Menu's subcomponents' props are visible on the docs site
Verify that generate doesn't cause git status changes

## Reviewer Notes

Changes can be

---


taylorvnoj (approved):
Approved to unblock other work - however we do need to log a ticket to investigate the forwardRef issue where the props page on the docs site doesn't show MenuItem props, etc

PR #3139, opened 2026-05-07
feat(components): Add getObjectUrl() prop to InputFile

edison-cy-yang:
## Why Is This Changing?

- `FileUpload.src()` returns a `data:` URL, which browsers **block** from top-frame navigation (`window.open`, `<a target="_blank">`) for security reasons.
- This breaks "open in new tab" previews for non-image attachments (PDFs, Word docs, etc.) produced by `<InputFile>`.
- The most visible symptom is in `<Gallery>`: clicking a non-image thumbnail silently fails with `Not allowed to navigate top frame to data URL`.
- Consumers currently work around this by wrapping `src()` and substituting a different URL — the fix belongs in the design system.

## What Is Changing?

- New method `FileUpload.getObjectUrl()` returns `{ url: string; revoke: () => void }` — a navigation-safe `blob:` URL paired with its cleanup function.
- `<Gallery>` uses it internally on the non-image thumbnail click path; revoke is scheduled via `requestAnimationFrame` after `window.open`.
- `src()` is unchanged. New code should prefer `getObjectUrl()`; moving away from the `src()` pattern over time without formally deprecating it.
- Docs updated in `InputFileNotes.mdx` and `Gallery.stories.mdx`.

## Consumer Impact

- **Impact:** Low
- **Action required:** None — additive change. Opt into `getObjectUrl()` when navigation is needed.
- **Breaking change:** No

- Existing `src()` callers continue to work.
- `getObjectUrl` is optional on `FileUpload` — external consumers constructing `FileUpload`-shaped objects don't need to provide it.
- `GalleryFile` picks `getObjectUrl` from `FileUpload` (non-breaking TS change).
- `<Gallery>` consumers passing `FileUpload` directly now get working non-image previews — fix, not regression.

## Validation

- New unit tests in `InputFile.test.tsx` and `Gallery.test.tsx`.
- All existing `InputFile` and `Gallery` tests pass unchanged (51 tests, 5 snapshots).
- New Storybook story `Gallery > With InputFile` demonstrates the full flow.
- Manual QA: PDF thumbnail → opens in new tab; image thumbnail → LightBox.

## Reviewer Notes

- **API shape**: `{ url, revoke }` returned together vs. separate sibling methods on `FileUpload`. Coupled-return prevents getting the URL without seeing the cleanup handle.
- **Optionality**: `getObjectUrl?` is optional so external consumers / mocks aren't forced to implement it. Always present on `<InputFile>`-produced objects.
- **`GalleryFile` Pick**: extends the existing `Pick<FileUpload, ...>` instead of redeclaring, so JSDoc and shape stay in sync.
- **`requestAnimationFrame(revoke)` timing**: gives the new tab a frame to start loading before the URL is invalidated. Revoking too early = blank tab; not revoking = leaks until page unload.

---


jdeichert (commented):
Great work! Tested in storybook.

Left a few comments about test patterns. Otherwise after that, happy to approve 👍

jdeichert (changes requested):

PR #3143, opened 2026-05-11
feat(components): Remove react 18 peerDep

jdeichert:
## Why Is This Changing?

Removing the react 18 peerDep so we officially only support v19+

This allows us to start removing `forwardRef` in favour of the newer `ref` as a prop pattern.

## What Is Changing?

Removing some peerDep ranges that allowed v18 to be used

## Consumer Impact

- Impact: High -- only if you were on react v18
- Action required: Migration required to react v19 if you were on v18
- Breaking change: Yes 

## Validation

Nothing to test, just a peerDep range change.

## Reviewer Notes

Changes can be

---


nad182 (approved):
I think the reason we would need this peerDep is for compatibility with the Jobber repo, right? but since we've sunset atlantis updates there, there should be no risks and disadvantages of landing this now.

PR #3147, opened 2026-05-12
feat(components): add onValueChange prop to InputNumberExperimental

nad182:
## Why Is This Changing?

`InputNumberExperimental.onChange` deliberately commits on blur / Enter / stepper / arrow step to mirror V1 / V2 `InputNumber` timing. That keeps side effects like network requests, URL updates, and form-state writes from firing on every keystroke, which is the right default for the vast majority of consumers.

It's wrong for one legitimate case: UI that needs to reflect in-progress input as the user types (e.g. a subtotal that updates live as a discount is typed). Today there's no first-party way to opt in, so consumers reach for raw `onKeyUp` and parse `event.currentTarget.value` themselves, losing the parsed-number / empty-input contract that `onChange` provides.

## What Is Changing?

New opt-in `onValueChange` prop. Same `(newValue: number | undefined) => void` signature as `onChange`, fires per keystroke (also on stepper / paste). Wires straight through to Base UI's `NumberField.onValueChange`. `onChange`'s commit-only timing is unchanged.

Also tightened the verbose `onChange` JSDoc and added a sibling JSDoc on `onValueChange` that points consumers to the right callback for their use case.

## Consumer Impact

- Impact: Low — purely additive, opt-in
- Action required: None
- Breaking change: No

## Validation

- 4 new tests in `InputNumberExperimental.test.tsx` under a new `onValueChange contract` describe:
  - fires per keystroke with the parsed number while the user types
  - emits `undefined` when the field is cleared
  - does not bleed into `onChange`'s commit-only timing when both callbacks are wired
  - fires on stepper interactions with the new value
- Existing 51 tests remain green (55 total pass)
- ESLint clean — `max-statements` was already at the threshold for this function; added an inline disable matching the existing precedent in `Menu.Legacy.tsx`

## Reviewer Notes

- Naming follows Base UI's underlying prop name (`onValueChange`) so the wiring is mechanical and the API matches what most React devs reading the source would expect
- No new files, no public type renames, no behaviour change to `onChange`


MichaelParadis (approved):
LGTM!

PR #3158, opened 2026-05-14
feat(design): Tailwind CSS plugin [POC]

taylorvnoj:
## Summary

- Adds `atlantisPlugin` to `@jobber/design` as a published subpath export (`@jobber/design/tailwind`), built via a new rollup entry and exported with proper ESM/CJS/types paths. `tailwindcss` is listed as an optional peer dependency so consumers who don't use Tailwind aren't affected.
- Wires Tailwind into the docs site (`packages/site`): `tailwind.config.ts`, PostCSS config, a `build:tailwind` script that compiles to `public/tailwind.css`, and loads it into the preview iframe skeleton.
- Adds a **"Using Tailwind CSS"** guide page at `/guides/using-tailwind` with live Canvas examples (token-based colors/spacing/radius, responsive order switching, `Menu.Item` direct className support) and a token reference table.
- Fixes `RenderComponentShowCode` to correctly display native HTML element names (e.g. `div`) instead of "Anonymous" in the Show Code panel.

Go here to see in the docs: https://taylor-tailwind-poc.atlantis.pages.dev/guides/using-tailwind?isLegacy=false&minimal=false

## Decision: why `@jobber/design/tailwind` and not a new package

We considered extracting the plugin into its own package (e.g. `@jobber/tailwind`) to keep `@jobber/design` focused. A standalone package would be cleanly opt-in and easy to version independently.

We landed on a subpath export of `@jobber/design` instead for a few reasons:

- The plugin's entire job is mapping design tokens to Tailwind utilities. The tokens already live in `@jobber/design` — putting the plugin anywhere else means it has to take a dependency on `@jobber/design` anyway, adding a hop for no real benefit.
- Consumers who already import `@jobber/design` (which is most product code) get the plugin with no additional install — just a new import path.
- Token sync is automatic: when `@jobber/design` ships new or updated tokens, the Tailwind classes update in the same release. A separate package would need coordinated releases.
- `tailwindcss` as an optional peer dep means teams not using Tailwind install nothing extra.

The tradeoff is that `@jobber/design`'s surface area grows slightly, but the coupling is honest — the plugin *is* a design token concern.

## Why this belongs in Atlantis

Keeping a Tailwind plugin in a product codebase works until it doesn't — token values drift, the plugin goes stale, and the team maintaining it isn't the same team who owns the design system. 

By living in Atlantis:

- **Single source of truth.** Token changes land in `@jobber/design` and the Tailwind classes update automatically in the same release. No manual token syncing across repos.
- **Versioned and auditable.** Changes go through the same PR process, changelog, and semver releases as every other Atlantis package. Teams can pin, upgrade, and track breaking changes like any other dependency.
- **Shared by default.** Any product repo that installs `@jobber/design` can opt into Tailwind support with one import. There's no copy-paste, no forking, no "which repo has the latest version of this."
- **Maintained by the right people.** The design system team owns the tokens and the plugin together. When a token is renamed or removed, the impact on Tailwind classes is caught at the source.
- **Documented where developers look.** The guide lives on the Atlantis docs site alongside the components it describes, not buried in a product repo's internal docs.

## Test plan

- [ ] Visit `/guides/using-tailwind` on the docs site — guide renders with working Canvas examples and Show Code toggles
- [ ] Confirm token-based classes (`bg-success--surface`, `p-base`, `rounded-base`, etc.) apply correctly in the Canvas previews
- [ ] Confirm the breakpoints example shows order switching at the `sm` (490px) breakpoint
- [ ] Run `pnpm run build` in `packages/design` — confirm `dist/tailwind.mjs`, `dist/tailwind.cjs`, and `dist/tailwind/atlantisPlugin.d.ts` are generated
- [ ] Confirm no TypeScript or ESLint errors across the repo

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scotttjob (commented):
Incredible work Taylor! Thank you!

Overall I'm close to just hitting approve (pending the team seeing this of course). I have one minor hesitation, and it has nothing to do with the work, just thinking ahead of how to maximize the impact of this change.

